PR #289, opened 2016-09-12
WIP: All HTTP docs

ktoso:
Created by some cowboy git commands:

```bash
git subtree split --prefix akka-docs # ... onto a branch
git remote add that-repo-with-the-split
git merge --allow-unrelated-histories that-split-branch
git rm # anything that wasn't akka-http related
```

This did not delete non-http files from the history of the docs, just the last commit does.
Do we care enough to really delete them from history?

It's a bit painful to list every non-http file for branching. Thus I propose to only make sure we did not pull in anything "heavy" (like a jar or zip file).

The PR is un-reviewable in github.
It effectively is all commits which touched akka-docs + notice last 3 commits, which are the cleanup I did.


ktoso:
PLS BUILD

ktoso:
Hmmm weird

```
+ git checkout -f master
error: pathspec 'master' did not match any file(s) known to git.
Build step 'Execute shell' marked build as failure
```

ktoso:
Well, at least got build status tagging to work ;)
The rest tomorrow morning

jonas:
The [docs/rst/scala](https://github.com/akka/akka-http/tree/wip-add-docs/docs/rst/scala) directory still contains several non-http related files.

ktoso:
Yes it does still, but wanted to make progress - as you see we also need to fix the PR Validator as it fails for weird reasons ;)

Thanks for sharing your WIP in the other issue.

jrudolph:
Actually for the docs I would prefer to merge everything into a single commit (that points to the akka/akka commit from where we copied) if the history cannot be preserved in a meaningful way.

ktoso:
You mean squash and merge? Sure, can do.

ktoso:
(From UI even)

jrudolph:
Resolved by doing another filter-branch of the whole history from akka/akka that now also includes relevant documentation for akka-http so that we can keep attribution for those as well.

ktoso:
Uff, finally done with those I hope :-)

PR #321, opened 2016-09-16
Error logging verbosity is not configurable for routing and used to determine log content of ErrorHandler (#16).

rbudzko:
Reused already existing property type. Set to full logging in tests as event filters asserts on that.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/settings/RoutingSettingsImpl.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/settings/RoutingSettingsImpl.scala
@@ -29,5 +29,4 @@ object RoutingSettingsImpl extends SettingsCompanion[RoutingSettingsImpl]("akka.
     c getBytes "range-coalescing-threshold",
     c getIntBytes "decode-max-bytes-per-chunk",
     c getString "file-io-dispatcher")
-
 }
--- a/akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/ExecutionDirectivesSpec.scala
+++ b/akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/ExecutionDirectivesSpec.scala
@@ -14,7 +14,7 @@ import akka.testkit.EventFilter
 import org.scalatest.matchers.Matcher
 
 class ExecutionDirectivesSpec extends RoutingSpec {
-  object MyException extends RuntimeException
+  object MyException extends RuntimeException("Boom")
   val handler =
     ExceptionHandler {
       case MyException ⇒ complete((500, "Pling! Plong! Something went wrong!!!"))
@@ -54,7 +54,10 @@ class ExecutionDirectivesSpec extends RoutingSpec {
         }
       }
     }
-    "not interfere with alternative routes" in EventFilter[MyException.type](occurrences = 1).intercept {
+    "not interfere with alternative routes" in EventFilter.error(
+      occurrences = 1,
+      message = "Error during processing of request: 'Boom'. Completing with 500 Internal Server Error response."
+    ).intercept {
       Get("/abc") ~>
         get {
           handleExceptions(handler)(reject) ~ { ctx ⇒
@@ -65,7 +68,10 @@ class ExecutionDirectivesSpec extends RoutingSpec {
           responseAs[String] shouldEqual "There was an internal server error."
         }
     }
-    "not handle other exceptions" in EventFilter[RuntimeException](occurrences = 1, message = "buh").intercept {
+    "not handle other exceptions" in EventFilter.error(
+      occurrences = 1,
+      message = "Error during processing of request: 'buh'. Completing with 500 Internal Server Error response."
+    ).intercept {
       Get("/abc") ~>
         get {
           handleExceptions(handler) {
@@ -76,7 +82,10 @@ class ExecutionDirectivesSpec extends RoutingSpec {
           responseAs[String] shouldEqual "There was an internal server error."
         }
     }
-    "always fall back to a default content type" in EventFilter[RuntimeException](occurrences = 2, message = "buh2").intercept {
+    "always fall back to a default content type" in EventFilter.error(
+      occurrences = 2,
+      message = "Error during processing of request: 'buh2'. Completing with 500 Internal Server Error response."
+    ).intercept {
       Get("/abc") ~> Accept(MediaTypes.`application/json`) ~>
         get {
           handleExceptions(handler) {
@@ -123,6 +132,32 @@ class ExecutionDirectivesSpec extends RoutingSpec {
           response should haveContentEncoding(gzip)
           status shouldEqual StatusCodes.NotFound
         }
+    }
+  }
+
+  "Default handler" should {
+    "handle `IllegalRequestException` with appropriate block of `ErrorHandler`" in EventFilter.warning(
+      occurrences = 1,
+      message = "Illegal request: 'Some summary.'. Completing with 409 Conflict response."
+    ).intercept {
+      Get("/abc") ~>
+        get {
+          throw new IllegalRequestException(ErrorInfo(summary = "Some summary."), StatusCodes.Conflict)
+        } ~> check {
+          status shouldEqual StatusCodes.Conflict
+        }
+    }
+
+    "handle exceptions other than `IllegalRequestException` with appropriate block of `ErrorHandler`" in EventFilter.error(
+      occurrences = 1,
+      message = "Error during processing of request: 're'. Completing with 500 Internal Server Error response."
+    ).intercept {
+      Get("/abc") ~>
+        get {
+          throw new RuntimeException("re")
+        }
+    } ~> check {
+      status shouldEqual StatusCodes.InternalServerError
     }
   }
 
--- a/akka-http/src/main/scala/akka/http/scaladsl/server/ExceptionHandler.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/server/ExceptionHandler.scala
@@ -1,7 +1,6 @@
 /*
  * Copyright (C) 2009-2016 Lightbend Inc. <http://www.lightbend.com>
  */
-
 package akka.http.scaladsl.server
 
 import scala.util.control.NonFatal
@@ -40,13 +39,11 @@ object ExceptionHandler {
   def default(settings: RoutingSettings): ExceptionHandler =
     apply(knownToBeSealed = true) {
       case IllegalRequestException(info, status) ⇒ ctx ⇒ {
-        ctx.log.warning(
-          "Illegal request {}\n\t{}\n\tCompleting with '{}' response",
-          ctx.request, info.formatPretty, status)
+        ctx.log.warning("Illegal request: '{}'. Completing with {} response.", info.summary, status)
         ctx.complete((status, info.format(settings.verboseErrorMessages)))
       }
       case NonFatal(e) ⇒ ctx ⇒ {
-        ctx.log.error(e, "Error during processing of request {}", ctx.request)
+        ctx.log.error("Error during processing of request: '{}'. Completing with {} response.", e.getMessage, InternalServerError)
         ctx.complete(InternalServerError)
       }
     }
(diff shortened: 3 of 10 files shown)


ktoso:
OK TO TEST

rbudzko:
@ktoso, I see the mistake, but would like to understand why do you keep additional set of tests under doc/rst?

ktoso:
> @ktoso, I see the mistake, but would like to understand why do you keep additional set of tests under doc/rst?

Because we're not done migrating things yet 😉

rbudzko:
> Because we're not done migrating things yet :wink:
Sounds legit :shipit: 

I think it failed due to the timeout you mentioned last time, right?

rbudzko:
PLS BUILD

jrudolph:
PLS BUILD

rbudzko:
> java.util.concurrent.TimeoutException: Futures timed out after [500 milliseconds]

@ktoso believes it is timing issue. I'm not sure this is related to this PR 😢. What do we do?

akka-http/src/main/scala/akka/http/scaladsl/server/ExceptionHandler.scala
```diff
@@ -37,16 +37,29 @@ object ExceptionHandler {
...
+
   def default(settings: RoutingSettings): ExceptionHandler =
     apply(knownToBeSealed = true) {
       case IllegalRequestException(info, status) ⇒ ctx ⇒ {
-        ctx.log.warning(
-          "Illegal request {}\n\t{}\n\tCompleting with '{}' response",
-          ctx.request, info.formatPretty, status)
+        log(
+          settings.errorLoggingVerbosity,
+          ctx.log.warning("Illegal request completing with '{}' response", status),
```
2beaucoup: Why did you remove the newline?

akka-http/src/main/scala/akka/http/scaladsl/server/ExceptionHandler.scala
```diff
@@ -37,16 +37,29 @@ object ExceptionHandler {
...
+          ctx.log.warning("Illegal request {}\n\t{}\n\tCompleting with '{}' response", ctx.request, info.formatPretty, status))
+
         ctx.complete((status, info.format(settings.verboseErrorMessages)))
       }
       case NonFatal(e) ⇒ ctx ⇒ {
-        ctx.log.error(e, "Error during processing of request {}", ctx.request)
+        log(
+          settings.errorLoggingVerbosity,
+          ctx.log.error("Error during processing one of requests occurred due to '{}'", e.getMessage),
```
2beaucoup:
```scala
ctx.log.error("Error during processing of request: '{}'", e.getMessage)
```
?

akka-http/src/main/resources/reference.conf
```diff
@@ -13,6 +13,15 @@ akka.http.routing {
   # (Note that akka-http will always produce log messages containing the full error details)
   verbose-error-messages = off
 
+  # Configures the verbosity with which message (request or response) parsing
+  # errors are written to the application log.
+  #
+  # Supported settings:
+  # `off`   : no log messages are produced
+  # `simple`: a condensed single-line message is logged
+  # `full`  : the full error details (potentially spanning several lines) are logged
```
2beaucoup: s/potentially spanning several lines/including stack traces?

akka-http-core/src/main/scala/akka/http/impl/settings/RoutingSettingsImpl.scala
```diff
@@ -28,6 +30,6 @@ object RoutingSettingsImpl extends SettingsCompanion[RoutingSettingsImpl]("akka.
     c getInt "range-count-limit",
     c getBytes "range-coalescing-threshold",
     c getIntBytes "decode-max-bytes-per-chunk",
-    c getString "file-io-dispatcher")
-
-}
+    c getString "file-io-dispatcher",
+    ErrorLoggingVerbosity(c getString "error-logging-verbosity"))
```
2beaucoup: Feels weird to reuse `ParserSettings.ErrorLoggingVerbosity` as this has no relation to parsing. Should we make `ErrorLoggingVerbosity` more general?

akka-http/src/main/scala/akka/http/scaladsl/server/ExceptionHandler.scala
rbudzko: Because this warning will be used for Simple which is meant to be one liner. At least it is how this is used in other places of akka-http.

akka-http/src/main/resources/reference.conf
rbudzko: Yes. I'll upgrade this explanation in all reference.confs.

akka-http/src/main/scala/akka/http/scaladsl/server/ExceptionHandler.scala
rbudzko: Again, Simple is meant to be one liner. Are you unhappy with e.getMessage? How do we replace it for one liner?

akka-http-core/src/main/scala/akka/http/impl/settings/RoutingSettingsImpl.scala
rbudzko: I like this name, but people with more stars 😉 should decide on that.

akka-http/src/main/scala/akka/http/scaladsl/server/ExceptionHandler.scala
2beaucoup: just a change in wording

akka-http-core/src/main/scala/akka/http/impl/settings/RoutingSettingsImpl.scala
2beaucoup: I too like the name. My suggestion was moving it to another place. ;)

rbudzko: Ah you mean `ParserSettings`. All right, I agree. It is worth noting RoutingSettings and ServerSettings do exist as well. Those two seems better candidates for more general `ErrorLoggingVerbosity`.

akka-http/src/main/scala/akka/http/scaladsl/server/ExceptionHandler.scala
rbudzko: It seems my eyes replace string content with same object SomeStringLiteral...

akka-http/src/main/scala/akka/http/scaladsl/server/ExceptionHandler.scala
2beaucoup: I see: The previous version is now used for the `Full` case. I'd tweak this slightly to

     "Illegal request. Completing with {} response."

akka-http-core/src/main/scala/akka/http/impl/settings/RoutingSettingsImpl.scala
rbudzko: I don't think we can remove `ErrorLoggingVerbosity` from `ParserSettings`, because of `Public API but not intended for subclassing`. We could copy it into `RoutingSettings`, but it is kind of meh. We also could create similar object in `RoutingSettings`, but we cannot remove or move current `ErrorLoggingVerbosity` due to `*Settings` being public.

rbudzko:
`X-Real-Ip` failed. Passes locally 🐹 .

2beaucoup:
`Ok(X-Real-Ip: 1.2.3.4,List()) did not equal Ok(X-Real-Ip: 1.2.3.4,List()) (HttpHeaderSpec.scala:691)` :confused:

rbudzko:
Sounds like Polish government claiming black is not black 👺 !

rbudzko:
@2beaucoup @ktoso Can we re run this build? Should something still be improved?

akka-http-core/src/main/resources/reference.conf
```diff
@@ -330,7 +330,7 @@ akka.http {
     # Supported settings:
     # `off`   : no log messages are produced
     # `simple`: a condensed single-line message is logged
-    # `full`  : the full error details (potentially spanning several lines) are logged
+    # `full`  : the full error details (potentially spanning several lines/including stack traces) are logged
```
2beaucoup: With my former comment I actually suggested replacing "potentially spanning several lines" with "including stack traces".

rbudzko:
@2beaucoup Simplified it to just hiding request body 😢 .

akka-http/src/main/scala/akka/http/scaladsl/server/ExceptionHandler.scala
```diff
@@ -40,13 +40,11 @@ object ExceptionHandler {
...
       case IllegalRequestException(info, status) ⇒ ctx ⇒ {
-        ctx.log.warning(
-          "Illegal request {}\n\t{}\n\tCompleting with '{}' response",
-          ctx.request, info.formatPretty, status)
+        ctx.log.warning("Illegal request. Completing with {} response.", status)
         ctx.complete((status, info.format(settings.verboseErrorMessages)))
       }
       case NonFatal(e) ⇒ ctx ⇒ {
-        ctx.log.error(e, "Error during processing of request {}", ctx.request)
+        ctx.log.error(e, "Error during processing of request.", ctx.request)
```
2beaucoup: `ctx.request` seems unused.  Either remove or change to something else and add back `{}`.

akka-http/src/main/scala/akka/http/scaladsl/server/ExceptionHandler.scala
```diff
@@ -1,13 +1,13 @@
...
  */
-
 package akka.http.scaladsl.server
 
 import scala.util.control.NonFatal
-import akka.http.scaladsl.settings.RoutingSettings
+import akka.http.scaladsl.settings.{ ParserSettings, RoutingSettings }
 import akka.http.scaladsl.model._
 import StatusCodes._
+import akka.http.scaladsl.settings.ParserSettings.ErrorLoggingVerbosity
```
2beaucoup: All import changes can be reverted.

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/ExecutionDirectivesSpec.scala
```diff
@@ -126,6 +126,32 @@ class ExecutionDirectivesSpec extends RoutingSpec {
     }
   }
 
+  "Default handler" should {
+    "handle `IllegalRequestException` with appropriate block of `ErrorHandler`" in EventFilter.warning(
+      occurrences = 1,
+      start = "Illegal request"
```
2beaucoup: use `message` instead of `start` for an exact match?

akka-http/src/main/scala/akka/http/scaladsl/server/ExceptionHandler.scala
2beaucoup: or something like
```scala
ctx.log.error("Error during processing of request: '{}'. Completing with {} response.",
   e.getMessage, status)
```
as discussed earlier.

rbudzko:
@2beaucoup Thnx for your points. I've applied requested changes. The build failed on some timeout.

akka-http/src/main/scala/akka/http/scaladsl/server/ExceptionHandler.scala
```diff
@@ -40,13 +39,11 @@ object ExceptionHandler {
   def default(settings: RoutingSettings): ExceptionHandler =
     apply(knownToBeSealed = true) {
       case IllegalRequestException(info, status) ⇒ ctx ⇒ {
-        ctx.log.warning(
-          "Illegal request {}\n\t{}\n\tCompleting with '{}' response",
-          ctx.request, info.formatPretty, status)
+        ctx.log.warning("Illegal request. Completing with {} response.", status)
```
2beaucoup: What do you think of
```scala
ctx.log.warning("Illegal request: {}. Completing with {} response.", info.summary, status)
```
to streamline with processing errors?

rbudzko:
@2beaucoup You are right. It's better now 😉 .

akka-http-core/src/main/resources/reference.conf
```diff
@@ -330,7 +330,7 @@ akka.http {
     # Supported settings:
     # `off`   : no log messages are produced
     # `simple`: a condensed single-line message is logged
-    # `full`  : the full error details (potentially spanning several lines) are logged
+    # `full`  : the full error details (including stack traces) are logged
```
2beaucoup: I think this does no longer apply, now.

rbudzko: Good question. It's about parsing. I'll check.

ktoso: Ok, this seems to have been addressed

ktoso:
LGTM, thanks! Seems all of @2beaucoup's comments were addressed as well.

rbudzko:
Yay, I think I'll open some bottle after this one 😉 .

2beaucoup:
Well deserved! Cheers @rbudzko! 🍻

rbudzko:
Refs #16.

Issue #5, opened 2016-08-24
Feedback of assertHeaderExists can be improved


jypma:
Currently, `akka.http.javadsl.testkit.TestRouteResult.assertHeaderExists(expectedHeader)`, when the header has a different value, will simply say `"Header $expected was missing."`. That's not very useful if you're trying to diagnose what's wrong with your actual returned header

Much more useful would be `"Expected $expected, but instead got $actualHeadersWithThatName"`

rbudzko:
@ktoso Shall we close?

2beaucoup:
Done. Thanks again Robert!

PR #323, opened 2016-09-16
Both assertHeaderExists functions now behave in the same way.

rbudzko:
While not sure if it is a desired behavior, I'd like to propose such a change :innocent: .

### DIFF
--- a/akka-http-testkit/src/main/scala/akka/http/javadsl/testkit/TestRouteResult.scala
+++ b/akka-http-testkit/src/main/scala/akka/http/javadsl/testkit/TestRouteResult.scala
@@ -178,14 +178,6 @@ abstract class TestRouteResult(_result: RouteResult, awaitAtMost: FiniteDuration
     assertEqualsKind(expected, entity(unmarshaller), "entity")
 
   /**
-   * Assert that a given header instance exists in the response.
-   */
-  def assertHeaderExists(expected: HttpHeader): TestRouteResult = {
-    assertTrue(response.headers.exists(_ == expected), s"Header $expected was missing.")
-    this
-  }
-
-  /**
    * Assert that a header of the given type exists.
    */
   def assertHeaderKindExists(name: String): TestRouteResult = {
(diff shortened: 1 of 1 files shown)


rbudzko:
BTW, it's #5.

ktoso:
This does not solve #5 though? I was about making the error message show what headers were available in addition to saying what was missing.

ktoso:
OK TO TES

ktoso:
OK TO TEST

rbudzko:
> This does not solve #5 though? I was about making the error message show what headers were available in addition to saying what was missing.

Another `assertHeaderExists` (called inside) is stating that header was missing or that it was present but none of given values was compliant with the assertion: `header was found but had the wrong value. Found headers: ${headers.mkString(", ")}")`.

ktoso:
Please mention such things in PRs ;) It's non obvious to see from the changes made that now it's supposed to work as intended

ktoso:
LGTM

rbudzko:
> Please mention such things in PRs ;) It's non obvious to see from the changes made that now it's supposed to work as intended

Uh, sorry. Will do in future.

jypma:
Thanks for the PR! Even more important than the PR text is the commit message. If it resolves #5, mention it in the message. And be specific: rather than saying "in the same way", explain what that way is.

That way, anyone browsing the code at a future point in time can immediately see the intentions, without having to do further digging.

rbudzko:
Thanks for the feedback! I've improved the commit message.

jypma:
```
[info]   - the config setting (chunked entity) *** FAILED ***
[info]     null did not equal EntityStreamSizeException: actual entity size (None) exceeded content length limit (10 bytes)! You can configure this by setting `akka.http.[server|client].parsing.max-content-length` or calling `HttpEntity.withSizeLimit` before materializing the dataBytes stream.,
[info]     inside HttpResponse(200 OK,List(),HttpEntity.Chunked(application/octet-stream),HttpProtocol(HTTP/1.1)) (LowLevelOutgoingConnectionSpec.scala:564)
```
is that even related to the change?

jrudolph:
> @akka/akka-http-team Do we have a policy re. commit msgs yet?

I'd say we try to keep the akka one for consistency.

2beaucoup:
Thx @rbudzko!

2beaucoup:
> I'd say we try to keep the akka one for consistency.

see https://github.com/akka/akka-meta/issues/30

Issue #240, opened 2016-09-12
Add `reverseProxy` directive


ktoso:
<a href="https://github.com/jrudolph"><img src="https://avatars.githubusercontent.com/u/9868?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [jrudolph](https://github.com/jrudolph)**
_Tuesday Feb 10, 2015 at 16:33 GMT_
_Originally opened as https://github.com/akka/akka/issues/16844_

----

to enable simple proxying of single requests to an (internal) remote backend.

This has some security implications: What to do with cookies, authentication data, etc.? So, it may make sense to provide some safe-guards preventing people from simply relaying requests e.g. by requiring to whitelist potential targets in the configuration or something similar.

See the discussions in the original ticket spray/spray#145 and an existing PR spray/spray#764.

/cc @sirthias

ktoso:
<a href="https://github.com/rkuhn"><img src="https://avatars.githubusercontent.com/u/470469?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [rkuhn](https://github.com/rkuhn)**
_Tuesday Oct 20, 2015 at 08:38 GMT_

----

TBH: it should be called `reverseProxy`.

ktoso:
<a href="https://github.com/ktoso"><img src="https://avatars.githubusercontent.com/u/120979?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [ktoso](https://github.com/ktoso)**
_Monday Apr 04, 2016 at 12:31 GMT_

----

This should actually not take more than a few lines, contributions are welcome!

ktoso:
<a href="https://github.com/TobiasPfeifer"><img src="https://avatars.githubusercontent.com/u/483514?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [TobiasPfeifer](https://github.com/TobiasPfeifer)**
_Sunday Jun 26, 2016 at 00:19 GMT_

----

I'd like to help implementing this feature.
I experimented around a bit but I'm not sure if I'm on the right track. I think I need some guidance with this one.
Here is what I did so far (see comments for open questions): 6e06049f2f2dbbadc0763fda0103119b1f98fba4

ktoso:
[...]

PR #324, opened 2016-09-16
WIP Add `reverseProxy` directive #240

TobiasPfeifer:
Directives to enable proxying of single requests to an (internal) remote backend.
Previous work (before splitting off akka-http) was done in akka/akka#21185

### DIFF
--- a/akka-http/src/main/scala/akka/http/javadsl/server/directives/ReverseProxyDirectives.scala
+++ b/akka-http/src/main/scala/akka/http/javadsl/server/directives/ReverseProxyDirectives.scala
@@ -1,0 +1,163 @@
+/*
+ * Copyright (C) 2009-2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.http.scaladsl.server.directives
+
+import akka.actor.ActorSystem
+import akka.http.scaladsl.Http
+import akka.http.scaladsl.Http.{ HostConnectionPool, OutgoingConnection }
+import akka.http.scaladsl.model.Uri.Authority
+import akka.http.scaladsl.model.headers._
+import akka.http.scaladsl.model.{ HttpResponse, _ }
+import akka.http.scaladsl.server._
+import akka.http.scaladsl.server.directives.BasicDirectives._
+import akka.http.scaladsl.server.directives.FutureDirectives._
+import akka.http.scaladsl.server.directives.RouteDirectives._
+import akka.pattern.CircuitBreaker
+import akka.stream._
+import akka.stream.scaladsl.{ Flow, Keep, Sink, Source }
+
+import scala.collection.immutable.Seq
+import scala.concurrent.Future
+import scala.util.{ Failure, Success, Try }
+
+trait ReverseProxyDirectives {
+  type RequestExecutor = ((HttpRequest) ⇒ Future[HttpResponse])
+
+  def completeUnmatchedPathWithReverseProxy(authority: Authority, configOption: Option[ReverseProxyConfig] = None): Route = {
+    forward(authority, configOption).tapply(x ⇒ complete(x._1))
+  }
+
+  def forward(authority: Authority, configOption: Option[ReverseProxyConfig] = None): Directive1[Try[HttpResponse]] = {
+    extractUri.flatMap { uri ⇒
+      val forwardUri: Uri = uri.withAuthority(authority)
+      forwardToUri(forwardUri, configOption)
+    }
+  }
+
+  def forwardUnmatchedPath(authority: Authority, configOption: Option[ReverseProxyConfig] = None): Directive1[Try[HttpResponse]] = {
+    extractUnmatchedPath.flatMap { unmatched ⇒
+      extractUri.flatMap { uri ⇒
+        val forwardUri: Uri = uri.withAuthority(authority).withPath(unmatched)
+        forwardToUri(forwardUri, configOption)
+      }
+    }
+  }
+
+  def forwardToUri(uri: Uri, configOption: Option[ReverseProxyConfig] = None): Directive1[Try[HttpResponse]] = {
+    val requestUriInterceptor = ProxyUriRequestInterceptor(uri).andThen(ProxyHeaderRequestInterceptor)
+    mapRequest(requestUriInterceptor).tflatMap { _ ⇒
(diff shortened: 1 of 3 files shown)


akka-http/src/main/scala/akka/http/javadsl/server/directives/ReverseProxyDirectives.scala
```diff
@@ -0,0 +1,163 @@
...
+private[akka] case object ProxyHeaderRequestInterceptor extends ((HttpRequest) ⇒ HttpRequest) {
+  override def apply(httpRequest: HttpRequest): HttpRequest = {
+    val headers: Seq[HttpHeader] = httpRequest.headers
+    val remoteAddressHeaderOption = headers.collectFirst({ case h: `Remote-Address` ⇒ h })
+
+    val xRealIpPF: PartialFunction[HttpHeader, `X-Real-Ip`] = { case h: `X-Real-Ip` ⇒ h }
+    val xForwardedForPF: PartialFunction[HttpHeader, `X-Forwarded-For`] = { case h: `X-Forwarded-For` ⇒ h }
+
+    val xRealIpHeaderOption = headers.collectFirst(xRealIpPF)
+    val xForwardedForHeaderOption = headers.collectFirst(xForwardedForPF)
```
2beaucoup:
```scala
val realIpHeader = httpRequest.header[`X-Real-Ip`]
val forwardedForHeader =  httpRequest.header[`X-Forwarded-For`]
```

akka-http/src/main/scala/akka/http/javadsl/server/directives/ReverseProxyDirectives.scala
```diff
@@ -0,0 +1,163 @@
...
+
+    val ipOption = xRealIpOption.orElse(xForwardedForHeaderOption.flatMap(_.addresses.headOption).orElse(remoteAddressOption))
+
+    val updatedXRealIpHeaderOption = ipOption.map(`X-Real-Ip`(_))
+    val updatedXForwardedForHeaderOption = remoteAddressOption.map(remoteAddresss ⇒ xForwardedForHeaderOption.fold(`X-Forwarded-For`(remoteAddresss))(h ⇒ `X-Forwarded-For`(h.addresses ++ Seq(remoteAddresss))))
+
+    val updatedHeaders = headers.filterNot((xRealIpPF orElse xForwardedForPF).isDefinedAt(_)) ++ (updatedXRealIpHeaderOption.toSeq ++ updatedXForwardedForHeaderOption.toSeq)
+    httpRequest.withHeaders(updatedHeaders)
+
+    //TODO: This has some security implications: What to do with cookies, authentication data, etc.?
```
2beaucoup: :+1: Let's discuss this in the ticket first.

akka-http/src/main/scala/akka/http/javadsl/server/directives/ReverseProxyDirectives.scala
```diff
@@ -0,0 +1,163 @@
...
+          reverseProxyConfig.circuitBreakerOption match {
+            case Some(circuitBreaker) ⇒ onCompleteWithBreaker(circuitBreaker)(responseFuture)
+            case None                 ⇒ onComplete(responseFuture)
+          }
+        }
+      }
+    }
+  }
+
+  private[akka] def getOrCreateDefaultConfig(configOption: Option[ReverseProxyConfig]): Directive1[ReverseProxyConfig] = {
```
2beaucoup: `ReverseProxyConfig` needs to be added to `RoutingSettings`/`reference.conf` where it can be overridden by `mapSettings`.

derekwyatt:
I'm very excited to see this, as I had to write a rudimentary reverse http proxy myself.  I think you're missing a few header removals though, according to https://tools.ietf.org/html/rfc2616#section-13.5.1 (i.e. the hop-by-hop headers).

As well, `Location` rewrites might need to be done. If the proxied server doesn't have a reasonable redirect mechanism, and redirects to a server this thing is proxying to, then it needs to rewrite the `Location` header to point back at the reverse proxy.  This is probably a "bonus feature" though.

I probably went overboard with my server, but I enabled a REST api to allow for the addition of or changes to the proxy configuration.  This allows me to manipulate it while not taking it down.  However, this is a very specific solution for me and I wouldn't know how to make it a "generic" feature.

Do you anticipate needing to handle the `X-Forwarded-Proto` or is that not on the agenda?  (it's not on mine :D)

2beaucoup:
I suggest a user provided filter for the request headers that defaults to a hard coded whitelist of minimum necessary/required headers (and without cookies/auth).

akka-http/src/main/scala/akka/http/javadsl/server/directives/ReverseProxyDirectives.scala
```diff
@@ -0,0 +1,163 @@
...
+  }
+}
+
+case class SingleRequestExecutor(implicit val system: ActorSystem, implicit val materializer: Materializer) extends ((HttpRequest) ⇒ Future[HttpResponse]) {
+  override def apply(httpRequest: HttpRequest): Future[HttpResponse] = {
+    Http().singleRequest(httpRequest)
+  }
+}
+
+case class OutgoingRequestExecutor(implicit val system: ActorSystem, implicit val materializer: Materializer) extends ((HttpRequest) ⇒ Future[HttpResponse]) {
```
leachbj: Do the next two really add any value?  Since we can't re-use the flow I'm pretty sure they devolve to the same as `SingleRequestExecutor`.

raboof:
OK TO TEST

raboof:
PLS BUILD

raboof:
The compilation fails on jenkins, perhaps this needs a rebase?

nmarasoiu:
Hi, is the proxyTo directive close to being done? We have a proxying use case (doing auth/autz with AkkaHttp and then delegating internally the request to a backend)..Thank you

johanandren:
@nmarasoiu Given that it hasn't seen any action since 2016 and does not even compile right now I don't think it is close to done. Maybe the initial contributors ran into some roadblock they couldn't get around? Ping @TobiasPfeifer

johanandren:
Closing this as there is a new PR (#2020) trying to achieve the same.

Issue #216, opened 2016-09-09
[akka-testkit] JUnitRouteTest has dependency on scalatest, and is Serializable


ktoso:
<a href="https://github.com/jypma"><img src="https://avatars.githubusercontent.com/u/483519?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [jypma](https://github.com/jypma)**
_Thursday May 19, 2016 at 09:26 GMT_
_Originally opened as https://github.com/akka/akka/issues/20559_

---

Since 2.4.5, `JUnitRouteTest` pulls in the trait `JUnitSuiteLike` from scalatest. This has two (unintended?) side-effects:
1. Projects using `JUnitRouteTest` must declare a dependency on scalatest themselves, since it's not a declared compile dependency for akka-testkit (apparently).
2. All Java test classes now are marked `Serializable` (since `Suite` is), triggering compiler warnings by default.

ktoso:
<a href="https://github.com/johanandren"><img src="https://avatars.githubusercontent.com/u/666915?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [johanandren](https://github.com/johanandren)**
_Thursday May 19, 2016 at 09:27 GMT_

----

@ktoso ;(

richard-imaoka:
Is the warning "Serializable class without serialVersionUID"? Somehow I didn't see any warning when I did git clone akka-http, and sbt compile on my machine.

richard-imaoka:
ping @jypma

richard-imaoka:
As in my PR 325 the way I proposed would not work... I might revisit this later, or if anyone else has a better idea please go ahead.

jypma:
@richard-imaoka here's the base class we're currently using:
```
public class JUnitRouteTest extends akka.http.javadsl.testkit.RouteTest {
    private static final Config config = ConfigFactory.defaultReference();
    private static final ActorSystem _system = ActorSystem.create("route-test", config);
    private static final Materializer _materializer = ActorMaterializer.create(_system);
    
    @Override
    public TestRouteResult createTestRouteResult(HttpRequest arg0, RouteResult arg1) {
        return new TestRouteResult(arg1, awaitDuration(), _system.dispatcher(), _materializer) {
            @Override
            public void assertEquals(Object expected, Object actual, String message) {
                Assert.assertEquals(message, expected, actual);
            }

            @Override
            public void assertEquals(int expected, int actual, String message) {
[...]

PR #325, opened 2016-09-16
Add compile dependency to akka-http-testkit on scalatest

richard-imaoka:
For the first issue in #216. Work in progress as the second issue not fixed yet, or I can create another PR for the issue 2 later.

> 1. Projects using JUnitRouteTest must declare a dependency on scalatest themselves, since it's not a declared compile dependency for akka-testkit (apparently).

This is fixed as in before/after below.

#### Before 

```
<dependency org="org.scalatest" name="scalatest_2.11" rev="3.0.0" conf="provided->default(compile);test->default(compile)"/>
```
    
#### After
    
```
<dependency org="org.scalatest" name="scalatest_2.11" rev="3.0.0" conf="compile->default(compile)"/>   
```

However, I didn't fix the second issue yet.

> 2. All Java test classes now are marked Serializable (since Suite is), triggering compiler warnings by default.

 Actually I didn't see any warning in my sbt, so wondering what the warnining @jypma saw.

> Is the warning for 2 = "Serializable class without serialVersionUID"?

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -64,6 +64,8 @@ object Dependencies {
     val aeronDriver = "io.aeron"                      % "aeron-driver"                 % "1.0.1"       // ApacheV2
     val aeronClient = "io.aeron"                      % "aeron-client"                 % "1.0.1"       // ApacheV2
 
+    val scalatest    = Def.setting { "org.scalatest"  %% "scalatest"  % scalaTestVersion.value } // ApacheV2
+
     object Docs {
       val sprayJson   = "io.spray"                   %%  "spray-json"                  % "1.3.2"             % "test"
       val gson        = "com.google.code.gson"        % "gson"                         % "2.3.1"             % "test"
@@ -143,7 +145,7 @@ object Dependencies {
 
   lazy val httpTestkit = l ++= Seq(
     Test.junit, Test.junitIntf, Compile.junit % "provided", 
-    Test.scalatest.value.copy(configurations = Some("provided; test"))
+    Compile.scalatest.value
   )
 
   lazy val httpTests = l ++= Seq(Test.junit, Test.scalatest.value, Test.junitIntf)


richard-imaoka:
All tests passed but unsuccessfull ? :(

```
[info] All tests passed.
[info] Passed: Total 1108, Failed 0, Errors 0, Passed 1108, Pending 38
[error] (akka-http-core/test:test) sbt.TestsFailedException: Tests unsuccessful
```

jypma:
@richard-imaoka Thanks for looking in to this.

The Eclipse compiler, on default settings, will issue a [warning](http://stackoverflow.com/questions/146715/use-the-serialversionuid-or-suppress-warnings) when implementing `Serializable` without declaring an explicit `serialVersionUid`. And that warning makes sense: it's an unwanted default to want to serialize your object, yet not care about versioning.

Since test class instances aren't going to be persisted or sent over a network, `Serializable` shouldn't be in the hierarchy to begin with. I think I wrote an alternative `JUnitRouteTest` somewhere in Java, but can't find it right now. I'll try to dig it up on Monday. I think it simply implemented the abstract methods in terms of JUnit assertions, without needing to pull in scalatest at all.

ktoso:
> And that warning makes sense: it's an unwanted default to want to serialize your object, yet not care about versioning.

AFAIR for akka serialization to pick up things tehy should be Serializable, even if a diff serializer is used. This is documented:

> If this condition cannot be met, because e.g. java.io.Serializable and MyOwnSerializable both apply and neither is a subtype of the other, a warning will be issued

> Akka provides serializers for java.io.Serializable and protobuf com.google.protobuf.GeneratedMessage by default (the latter only if depending on the akka-remote module), so normally you don't need to add configuration for that; since com.google.protobuf.GeneratedMessage implements java.io.Serializable, protobuf messages will always be serialized using the protobuf protocol unless specifically overridden. In order to disable a default serializer, map its marker type to “none”:

http://doc.akka.io/docs/akka/snapshot/scala/serialization.html

---- 

> Since test class instances aren't going to be persisted or sent over a network, Serializable shouldn't be in the hierarchy to begin with.

Any case class is Serializable, so it could come from there.

jypma:
@ktoso this is for the `*Spec` class itself, not for any data it uses. Unless you're feeding your spec instance itself into akka as a message, I don't think akka serialization applies.

`JUnitRouteTest` pulls in `JUnitSuiteLike` which is `Suite` which, for reasons I don't understand, implements `Serializable`.

But don't forget the argument that if you really want to provide a JUnit base class, it should be rooted in JUnit, not scalatest.

ktoso:
I'm just explaining where a Serializable interface in here could come from

richard-imaoka:
Thanks both for the info ! 

@jypma 

I will wait for this - I think you send in a separate PR?

>I wrote an alternative JUnitRouteTest somewhere in Java

If that goes through, let's close this PR.

Just in case if not, I think it makes sense to [suppress the warning](http://stackoverflow.com/questions/15045865/how-to-remove-the-warning-the-serializable-class-classname-does-not-declare-a-st) in this PR.

jypma:
@richard-imaoka You unfortunately can't suppress the warning in this PR. The warning is created on every concrete subclass of `JUnitRouteTest`, if it implements `Serializable`. So that'll always be in user code.

Either way, the current PR does not make things worse, so may as well go forward :-)

ktoso:
I'm not sure this is the right thing to do.
Won't it break projects for people who were using specs2? It would pull in Scala TEST, which they don't want.

The real solution here might be that we make special modules for specs2 / scalatest / java, that have proper dependencies.

richard-imaoka:
Hmm that sounds like overkill just for a test module and other akka-xxx-test(kit) modules don't do that.

Let me close this PR, and revisit if I can come up with a better idea.

ktoso:
I think we should think about it before releasing a stable, reopening and will give it some thought over the week. 

cc @jrudolph

ktoso:
Meh, thought this is an issue not PR. Closing back

Issue #218, opened 2016-09-09
Great getting started ticket: write example snippets for Akka HTTP Java DSL


ktoso:
<a href="https://github.com/johanandren"><img src="https://avatars.githubusercontent.com/u/666915?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [johanandren](https://github.com/johanandren)**
_Friday May 06, 2016 at 14:06 GMT_
_Originally opened as https://github.com/akka/akka/issues/20466_

---

The last couple of months we have been hard at work with providing a DSL for Java that is (almost) as neat as the Scala DSL for Akka HTTP. We are now at the point where we have the API in place in master and basic docs but what is missing is examples for each of the directives. 

Providing an example for a directive is a perfect way to start contributing to Akka, each directive is sufficiently isolated, and there is a corresponding Scala example to base the sample on.

How to do it:
- **Select a directive below**, **write a comment** that you will be working on it
- **Read the docs**, [the corresponding Scala directive docs]file(http://doc.akka.io/docs/akka/2.4.4/scala/http/routing-dsl/directives/alphabetically.html#predefined-directives) and make sure you understand what the directive does
- Look at **the Scala example** in `akka-docs/rst/scala/code/docs/http/scaladsl/server/directives` and port it to a Java example in `akka-docs/rst/java/code/docs/http/javadsl/server/directives`. Look at the existing Java samples to see how the tests should be set up
- **Run the test using** `akka-docs/testOnly docs.http.javadsl.server.directives.YourDirective`
- **Include the test in the docs** by adding comments with the label from the `.rst` documentation of the directive. 
- **Generate the docs** using `akka-docs/sphinx:generateHtml` and see that the sample is included in the docs and looks nice (you will find the docs generated as HTML under akka-docs/target/sphinx/html/index.html` so that you can preview in your browser)
- **Make a PR** to have the sample reviewed and merged into Akka HTTP

**Directives to provide samples for**
(all paths relative to `akka-docs/rst/java/http/routing-dsl/directives`)
- [x] basic-directives/cancelRejection.rst thanks @Hawstein!
- [x] basic-directives/cancelRejections.rst thanks @Hawstein!
- [x] basic-directives/extract.rst thanks @Hawstein!
- [x] basic-directives/extractExecutionContext.rst thanks @Hawstein!
- [x] basic-directives/extractLog.rst thanks @Hawstein!
- [x] basic-directives/extractMaterializer.rst thanks @Hawstein!
[...]

PR #335, opened 2016-09-19
Add entity directive example

richard-imaoka:
Refs #218 entity directive

From entity.md, I removed mentions to spray-json and JsValue, but is that ok? I was assuming currently supported marshalling methods in Java is only via Jackson, as in [marshalling.html](http://doc.akka.io/docs/akka/2.4/java/http/routing-dsl/marshalling.html), but let me know if that is wrong.

image

### DIFF
--- a/docs/src/main/paradox/java/http/routing-dsl/directives/marshalling-directives/entity.md
+++ b/docs/src/main/paradox/java/http/routing-dsl/directives/marshalling-directives/entity.md
@@ -24,12 +24,8 @@ are mapped to rejections:
 
 ## Examples
 
-The following example uses `spray-json` to unmarshall a json request into a simple `Person` 
-class.  It utilizes `SprayJsonSupport` via the `PersonJsonSupport` object as the in-scope unmarshaller.
+The following example uses @ref[Json Support via Jackson](../../../common/json-support.md#json-jackson-support-java) to unmarshall a json request into a simple `Person`  
 
-TODO: Add example snippets.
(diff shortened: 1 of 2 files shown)


docs/src/test/java/docs/http/javadsl/server/directives/MarshallingDirectivesExamplesTest.java
```diff
@@ -0,0 +1,61 @@
...
+  }
+  //#person
+
+  @Test
+  public void testEntity() {
+    //#example-entity-with-json
+    final Unmarshaller<HttpEntity, Person> unmarshallar = Jackson.unmarshaller(Person.class);
+
+    final Route route = entity(unmarshallar, person ->
+      complete( "Person:" +  person.getName() + " - favoriteNumber:" + person.getFavoriteNumber() )
```
ktoso: weird formatting, why the spaces around here :)

docs/src/main/paradox/java/http/routing-dsl/directives/marshalling-directives/entity.md
```diff
@@ -24,12 +24,8 @@ are mapped to rejections:
 
 ## Examples
 
-The following example uses `spray-json` to unmarshall a json request into a simple `Person` 
-class.  It utilizes `SprayJsonSupport` via the `PersonJsonSupport` object as the in-scope unmarshaller.
```
ktoso: that removal is correct, thanks a lot!

PR #337, opened 2016-09-20
DO NOT MERGE: mute some error logging

2beaucoup:
Just trying to get some feedback. While cleaning up the test log I came across these.

``` java
new JavaTestKit(system()) {{
  new JavaTestKit.EventFilter<TestRouteResult>(SomeException.class) {
    protected TestRouteResult run() {
      return testRoute(route).run(...).assertXXX;
    }
  }.exec();
}};
```

looks rather verbose. But maybe there's no way around it?

### DIFF
--- a/docs/src/test/java/docs/http/javadsl/server/directives/FutureDirectivesExamplesTest.java
+++ b/docs/src/test/java/docs/http/javadsl/server/directives/FutureDirectivesExamplesTest.java
@@ -11,9 +11,11 @@ import akka.http.javadsl.model.HttpRequest;
 import akka.http.javadsl.marshalling.Marshaller;
 import akka.http.javadsl.server.Route;
 import akka.http.javadsl.testkit.JUnitRouteTest;
+import akka.http.javadsl.testkit.TestRouteResult;
 import akka.http.scaladsl.model.StatusCodes;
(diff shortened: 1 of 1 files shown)


2beaucoup:
PLS BUILD

jrudolph:
Not least because we are showing the code in the documentation, I think we need to come up with something less verbose and maybe put it into our testkit (e.g. into JUnitRouteTest or even into akka JavaTestKit).

A few ideas:
 * use Java 8 lambdas to make that wrapping less verbose, i.e. create a method `<T> T dontLogExceptionOfType(Class<?> exceptionType, Callable<T> body)`
 * add another resource similar to `ActorSystemResource` that allows to switch off certain kinds of logging by side-effects (but clean-up before running new tests), so that you can run `muteExceptionLogging(RuntimeException.class)` at the start of a test

ktoso:
Something similar to `Mute` (see `akka.testkit.TestEvent`) would probably be better. 
We could add our special test logger that would perhaps solve this and listen to muting?

ktoso:
I'll close this for now so we have a clean view on the list of outstanding PRs as we near the 2nd half of the week, hoping to release by the end of it.

Please feel free to contiune the discussion here or elsewhere though.

Issue #333, opened 2016-09-19
FAILED HttpExtensionApiSpec


jrudolph:
Failed here: https://jenkins.akka.io:8498/job/pr-validator-akka-http/58/consoleFull

```
[info] - should properly bind and handle a server with a flow (with four parameters) *** FAILED ***
[info]   java.util.concurrent.TimeoutException:
[info]   at java.util.concurrent.CompletableFuture.timedGet(CompletableFuture.java:1763)
[info]   at java.util.concurrent.CompletableFuture.get(CompletableFuture.java:1907)
[info]   at scala.concurrent.java8.FuturesConvertersImpl$CF.scala$concurrent$java8$FuturesConvertersImpl$CF$$super$get(FutureConvertersImpl.scala:84)
[info]   at scala.concurrent.java8.FuturesConvertersImpl$CF$$anonfun$get$2.apply(FutureConvertersImpl.scala:84)
[info]   at scala.concurrent.BlockContext$DefaultBlockContext$.blockOn(BlockContext.scala:53)
[info]   at scala.concurrent.package$.blocking(package.scala:123)
[info]   at scala.concurrent.java8.FuturesConvertersImpl$CF.get(FutureConvertersImpl.scala:84)
[info]   at akka.http.javadsl.HttpExtensionApiSpec.akka$http$javadsl$HttpExtensionApiSpec$$waitFor(HttpExtensionApiSpec.scala:472)
[info]   at akka.http.javadsl.HttpExtensionApiSpec$$anonfun$1$$anonfun$apply$mcV$sp$5.apply(HttpExtensionApiSpec.scala:125)
[info]   at akka.http.javadsl.HttpExtensionApiSpec$$anonfun$1$$anonfun$apply$mcV$sp$5.apply(HttpExtensionApiSpec.scala:115)
[info]   ...
```

PR #339, opened 2016-09-20
Remove `charset` from `Multipart` types #333

derekwyatt:
- `Multipart` now derives from `Binary` instead of `WithOpenCharset`
- Everything else is a ripple effect from there

### DIFF
--- a/akka-http-core/src/main/java/akka/http/javadsl/model/MediaTypes.java
+++ b/akka-http-core/src/main/java/akka/http/javadsl/model/MediaTypes.java
@@ -120,13 +120,13 @@ public final class MediaTypes {
     public static final MediaType.Binary MESSAGE_DELIVERY_STATUS = akka.http.scaladsl.model.MediaTypes.message$divdelivery$minusstatus();
     public static final MediaType.Binary MESSAGE_RFC822 = akka.http.scaladsl.model.MediaTypes.message$divrfc822();
 
-    public static final MediaType.WithOpenCharset MULTIPART_MIXED = akka.http.scaladsl.model.MediaTypes.multipart$divmixed();
-    public static final MediaType.WithOpenCharset MULTIPART_ALTERNATIVE = akka.http.scaladsl.model.MediaTypes.multipart$divalternative();
-    public static final MediaType.WithOpenCharset MULTIPART_RELATED = akka.http.scaladsl.model.MediaTypes.multipart$divrelated();
-    public static final MediaType.WithOpenCharset MULTIPART_FORM_DATA = akka.http.scaladsl.model.MediaTypes.multipart$divform$minusdata();
-    public static final MediaType.WithOpenCharset MULTIPART_SIGNED = akka.http.scaladsl.model.MediaTypes.multipart$divsigned();
-    public static final MediaType.WithOpenCharset MULTIPART_ENCRYPTED = akka.http.scaladsl.model.MediaTypes.multipart$divencrypted();
-    public static final MediaType.WithOpenCharset MULTIPART_BYTERANGES = akka.http.scaladsl.model.MediaTypes.multipart$divbyteranges();
+    public static final MediaType.Binary MULTIPART_MIXED = akka.http.scaladsl.model.MediaTypes.multipart$divmixed();
+    public static final MediaType.Binary MULTIPART_ALTERNATIVE = akka.http.scaladsl.model.MediaTypes.multipart$divalternative();
+    public static final MediaType.Binary MULTIPART_RELATED = akka.http.scaladsl.model.MediaTypes.multipart$divrelated();
+    public static final MediaType.Binary MULTIPART_FORM_DATA = akka.http.scaladsl.model.MediaTypes.multipart$divform$minusdata();
+    public static final MediaType.Binary MULTIPART_SIGNED = akka.http.scaladsl.model.MediaTypes.multipart$divsigned();
+    public static final MediaType.Binary MULTIPART_ENCRYPTED = akka.http.scaladsl.model.MediaTypes.multipart$divencrypted();
+    public static final MediaType.Binary MULTIPART_BYTERANGES = akka.http.scaladsl.model.MediaTypes.multipart$divbyteranges();
 
     public static final MediaType.WithOpenCharset TEXT_ASP = akka.http.scaladsl.model.MediaTypes.text$divasp();
     public static final MediaType.WithOpenCharset TEXT_CACHE_MANIFEST = akka.http.scaladsl.model.MediaTypes.text$divcache$minusmanifest();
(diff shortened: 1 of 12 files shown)


akka-http-core/src/main/scala/akka/http/scaladsl/model/Multipart.scala
```diff
@@ -56,14 +56,13 @@ sealed trait Multipart extends jm.Multipart {
   /**
    * Creates a [[akka.http.scaladsl.model.MessageEntity]] from this multipart object.
    */
-  def toEntity(
-    charset:  HttpCharset = HttpCharsets.`UTF-8`,
-    boundary: String      = BodyPartRenderer.randomBoundary())(implicit log: LoggingAdapter = NoLogging): MessageEntity = {
+  def toEntityWithLog(
```
derekwyatt: Not sure how you'd like to deal with my introduced `toEntityWithLog` naming issue, so I just left it as 'awful' for the moment

derekwyatt:
Bugger.  My commit message references the wrong issue number.  I'll fix that as I fix the other things y'all will need me to fix.

ktoso:
OK TO TEST

akka-http-core/src/main/java/akka/http/javadsl/model/MediaTypes.java
```diff
@@ -120,13 +120,13 @@ private MediaTypes() { }
...
     public static final MediaType.Binary MESSAGE_RFC822 = akka.http.scaladsl.model.MediaTypes.message$divrfc822();
 
-    public static final MediaType.WithOpenCharset MULTIPART_MIXED = akka.http.scaladsl.model.MediaTypes.multipart$divmixed();
-    public static final MediaType.WithOpenCharset MULTIPART_ALTERNATIVE = akka.http.scaladsl.model.MediaTypes.multipart$divalternative();
-    public static final MediaType.WithOpenCharset MULTIPART_RELATED = akka.http.scaladsl.model.MediaTypes.multipart$divrelated();
-    public static final MediaType.WithOpenCharset MULTIPART_FORM_DATA = akka.http.scaladsl.model.MediaTypes.multipart$divform$minusdata();
-    public static final MediaType.WithOpenCharset MULTIPART_SIGNED = akka.http.scaladsl.model.MediaTypes.multipart$divsigned();
-    public static final MediaType.WithOpenCharset MULTIPART_ENCRYPTED = akka.http.scaladsl.model.MediaTypes.multipart$divencrypted();
-    public static final MediaType.WithOpenCharset MULTIPART_BYTERANGES = akka.http.scaladsl.model.MediaTypes.multipart$divbyteranges();
+    public static final MediaType.Binary MULTIPART_MIXED = akka.http.scaladsl.model.MediaTypes.multipart$divmixed();
```
jrudolph: I wonder what we can do about compatibility constraints here.

akka-http-core/src/main/java/akka/http/javadsl/model/Multipart.java
```diff
@@ -39,7 +39,7 @@
     /**
      * Creates an entity from this multipart object.
      */
-    RequestEntity toEntity(HttpCharset charset, String boundary);
+    RequestEntity toEntity(String boundary);
```
jrudolph: We probably need to create a new method here as well and deprecate the old one.

akka-http-core/src/main/java/akka/http/javadsl/model/MediaTypes.java
jrudolph: akka-http-core is currently stable so we cannot change public API in a binary incompatible way. It might mean that we have to introduce new constants using the fixed types.

derekwyatt: Ah, I see.

derekwyatt: So, to fix this, do you think we need to create a `WithOpenCharset` adapter from `Binary` and then create _new_ values, such as `MULTIPART_MIXED_BINARY`?

akka-http-core/src/main/scala/akka/http/scaladsl/model/Multipart.scala
```diff
@@ -97,9 +96,9 @@ object Multipart {
      */
     def strictParts: immutable.Seq[Multipart.BodyPart.Strict]
 
-    override def toEntity(charset: HttpCharset, boundary: String)(implicit log: LoggingAdapter = NoLogging): HttpEntity.Strict = {
-      val data = BodyPartRenderer.strict(strictParts, boundary, charset.nioCharset, partHeadersSizeHint = 128, log)
-      HttpEntity(mediaType withBoundary boundary withCharset charset, data)
+    override def toEntityWithLog(boundary: String)(implicit log: LoggingAdapter = NoLogging): HttpEntity.Strict = {
```
derekwyatt: Is this now OK?  The base now has `toEntity` that delegates down to this override, eventually.  Does that satisfy bin-compat?

akka-http-core/src/main/scala/akka/http/javadsl/model/MediaType.scala
```diff
@@ -27,12 +27,35 @@ object MediaType {
...
+    def isApplication: Boolean = b.isApplication
+    def isAudio: Boolean = b.isAudio
+    def isImage: Boolean = b.isImage
+    def isMessage: Boolean = b.isMessage
+    def isMultipart: Boolean = b.isMultipart
+    def isText: Boolean = b.isText
+    def isVideo: Boolean = b.isVideo
+    def toRange: MediaRange = b.toRange
+    def toRange(qValue: Float): MediaRange = b.toRange(qValue)
+  }
```
derekwyatt: I'm not sure this is a really great idea, but I'm a bit stuck on how to keep everything truly happy.

ktoso:
@jrudolph @derekwyatt would it make sense to notice this in the migration guide? Let's include it in the 3.0.0-RC1 I'm about to cut 👍

jrudolph:
This PR currently contains part of the changes to actually make it binary compatible. We might want to revisit that now and maybe just make the plain change and access a bit of potential breakage if people relied on `Multipart` being a subclass of `WithOpenCharset`.

derekwyatt:
If there's an opportunity to remove the bincompat stuff, that'd be great.

ktoso:
So I propose to merge this as is and I'll remove the `binaryAdaptedToWithOpenCharset ` and the related values before releasing, sounds good?

derekwyatt:
It's a bit more than that. I added some deprecations and created a new method called `toEntityWithLog`.  It would be much better if we could avoid that poorly named variant and just modify the existing code to suit the change.

ktoso:
With the plan to release before evening hits, what's our best way to get this shipped? :-)
Shall I take over the PR?

derekwyatt:
That's perfectly fine with me. Alternatively, I can also try and fix it up myself, but I have a feeling that your "evening" is very soon :)

ktoso:
4 hours or so would still work ;)

jrudolph:
I'm just looking into it.

derekwyatt:
OK, I can probably jump on it soon if need be.  Just lemme know and I'll do whatever needs doing.

jrudolph:
Created a new PR in https://github.com/akka/akka-http/pull/398.

ktoso:
Thanks! Superseded by https://github.com/akka/akka-http/pull/398/

Issue #346, opened 2016-09-22
Directory Traversal Attack vulnerability in Akka HTTP's getFrom(Browseable)Directory directive


roikonen:
I noticed that when running Akka HTTP server in Windows machine I am able to access all files of the server machine.

**Example**

With route:
`
pass {
   getFromBrowseableDirectory("/some/directory/i/only/want/to/share")
}
`
or
`
pass {
   getFromDirectory("/some/directory/i/only/want/to/share")
}
`

I can request for example:
http://localhost:8080/%5c..%5c..%5c..%5c..%5c..%5c..%5c..%5c..%5c..%5cwindows/win.ini
...and download the actual win.ini file that exist in C:/windows

In the case of getFromBrowseableDirectory I can also see all the files there is in the server machine.

That is Directory Traversal Attack vulnerability: https://www.owasp.org/index.php/Path_Traversal

rbudzko:
`FileAndResourceDirective` check by tailrec if path try to escape base dir configured in routing by `(head.indexOf('/') >= 0 || head == "..")`. I don't think maintaining rules on server side will ever make it 100% safe.

drewhk:
hm, shouldn't it check after parsing, using the JDK File API? It can canonicalize paths, but I am not an expert here.

rbudzko:
Well, it will be more powerful than current solution, definitely. JDK `*absolutePath` seems to do a lot of magic (really magic 😨 ) to check it without calling OS.

jypma:
In order to do this proper, we'd have to go all-in and check on symlinks etc. as well. Like apache does.

jypma:
yes, maybe `absolutePath` will do the symlink dereffing as well?

drewhk:
Nope, you need CanonicalPath. Absolute Path means that it is not relative to the current directory, but it does not mean that it does not have `..` sequences in there. As far as I understood...

drewhk:
> A canonical pathname is both absolute and unique. The precise definition of canonical form is system-dependent. This method first converts this pathname to absolute form if necessary, as if by invoking the getAbsolutePath() method, and then maps it to its unique form in a system-dependent way. This typically involves removing redundant names such as "." and ".." from the pathname, resolving symbolic links (on UNIX platforms), and converting drive letters to a standard case (on Microsoft Windows platforms).

rbudzko:
Yeah, also it calls native canonicalize0 which could be checked if implemented correctly or if making sys calls.

[...]

PR #348, opened 2016-09-23
Defend against directory traversal

2beaucoup:
refs #346


2beaucoup:
The 2nd commit is the interesting change. If this looks OK then I'll
- change `getFromResourceDirectory` as well
- add the symlink follow setting

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectives.scala
```diff
@@ -127,14 +149,26 @@ trait FileAndResourceDirectives {
...
+    extractUnmatchedPath { unmatchedPath ⇒
       extractLog { log ⇒
-        fileSystemPath(base, path, log) match {
-          case ""       ⇒ reject
-          case fileName ⇒ getFromFile(fileName)
-        }
+        val path = if (unmatchedPath.startsWithSlash) unmatchedPath.tail else unmatchedPath
+        val file = base.resolve(path.toString)
+        if (exists(file, LinkOption.NOFOLLOW_LINKS) && file.toRealPath().startsWith(base)) // FIXME: optional symlinks
+          // TODO log
```
jypma: do we really need to log all? these could also be "honest" 404s to files that no longer exist.

jypma:
direction looks good to me.

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectivesSpec.scala
```diff
@@ -19,108 +21,136 @@ import akka.http.scaladsl.model.MediaTypes._
...
 
   // operations touch files, can be randomly hit by slowness
   implicit val routeTestTimeout = RouteTestTimeout(3.seconds)
 
+  val base = Paths.get(getClass.getClassLoader.getResource("").toURI)
+
   override def testConfigSource = "akka.http.routing.range-coalescing-threshold = 1"
 
   "getFromFile" should {
+    def _getFromFile(file: String) = getFromFile(base.resolve(file))
```
jrudolph: Nice one :+1:

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectivesSpec.scala
```diff
@@ -19,108 +21,136 @@ import akka.http.scaladsl.model.MediaTypes._
...
+      Get() ~> _getFromFile("sample.js.gz") ~> check {
+        mediaType shouldEqual `application/javascript`
+        header[`Content-Encoding`] shouldEqual Some(`Content-Encoding`(HttpEncodings.gzip))
+        responseAs[String] shouldEqual "456"
+      }
+    }
+  }
+
+  "getFromDirectory" should {
+    def getFromDir(directory: String) = getFromDirectory(base.resolve(directory))
```
jrudolph: `_getFromDirectory` for consistency with `_getFromFile`?

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectivesSpec.scala
```diff
@@ -19,108 +21,136 @@ import akka.http.scaladsl.model.MediaTypes._
...
+      Get("nonExistentFile") ~> getFromDir("subDirectory") ~> check { handled shouldEqual false }
+    }
+    "reject requests to directories" in {
+      Get("sub") ~> getFromDir("someDir") ~> check { handled shouldEqual false }
+    }
+    "reject path traversal attempts" in {
+      val route = getFromDir("someDir/sub")
+      Get("file.html") ~> route ~> check { handled shouldEqual true }
+      def shouldReject(prefix: String) = Get(prefix + "fileA.txt") ~> route ~> check { handled shouldEqual false }
+      shouldReject("../")
```
jrudolph: Did you come up with this list of potential targets yourself or is it based on some public advice?

jrudolph: In any case, it's already a nice set...

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectives.scala
jrudolph: Good point. We could alternatively just log paths that are outside the base directory.

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectivesSpec.scala
2beaucoup: Most of them are from https://www.owasp.org/index.php/Path_Traversal which is linked from the ticket description. If anyone has more I'll be happy to add them here.

2beaucoup:
@jrudolph Haven't verified against current impl yet. Will do.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectives.scala
2beaucoup: Agreed. Current impl does some logging. Comment was just a reminder for myself. ;)

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectivesSpec.scala
2beaucoup: I introduced this one first. Will consistify with above. :)

2beaucoup:
@jrudolph Current impl gives:
```
../: reject
%5c../: pass
../: reject
%2e%2e%2f: reject
%2e%2e/: reject
..%2f: reject
%2e%2e%5c: pass
%2e%2e\: pass
..%5c: pass
..%255c: reject
..%c0%af: reject
..%c1%9c: reject
```

jrudolph:
Just to be sure: you are running those on Windows?

2beaucoup:
yep

jrudolph:
Have you seen the failing test? It fails reproducibly for me as well:

```
[info] - should return the file content with the MediaType matching the file extension *** FAILED ***
[info]   Request was rejected (RoutingSpec.scala:19)
[info]   org.scalatest.exceptions.TestFailedException:
[info]   at akka.http.scaladsl.testkit.TestFrameworkInterface$Scalatest$class.failTest(TestFrameworkInterface.scala:24)
[info]   at akka.http.scaladsl.server.RoutingSpec.failTest(RoutingSpec.scala:19)
[info]   at akka.http.scaladsl.testkit.RouteTestResultComponent$RouteTestResult.rawResponse(RouteTestResultComponent.scala:56)
[info]   at akka.http.scaladsl.testkit.RouteTestResultComponent$RouteTestResult.entityRecreator$lzycompute(RouteTestResultComponent.scala:80)
[info]   at akka.http.scaladsl.testkit.RouteTestResultComponent$RouteTestResult.entityRecreator(RouteTestResultComponent.scala:79)
[info]   at akka.http.scaladsl.testkit.RouteTestResultComponent$RouteTestResult.entity(RouteTestResultComponent.scala:43)
[info]   at akka.http.scaladsl.testkit.RouteTest$class.responseEntity(RouteTest.scala:63)
[info]   at akka.http.scaladsl.server.RoutingSpec.responseEntity(RoutingSpec.scala:19)
[info]   at akka.http.scaladsl.testkit.RouteTest$class.contentType(RouteTest.scala:73)
[info]   at akka.http.scaladsl.server.RoutingSpec.contentType(RoutingSpec.scala:19)
[info]   at akka.http.scaladsl.testkit.RouteTest$class.mediaType(RouteTest.scala:74)
[info]   at akka.http.scaladsl.server.RoutingSpec.mediaType(RoutingSpec.scala:19)
[info]   at akka.http.scaladsl.server.directives.FileAndResourceDirectivesSpec$$anonfun$1$$anonfun$apply$mcV$sp$4$$anonfun$apply$5.apply(FileAndResourceDirectivesSpec.scala:48)
[info]   at akka.http.scaladsl.server.directives.FileAndResourceDirectivesSpec$$anonfun$1$$anonfun$apply$mcV$sp$4$$anonfun$apply$5.apply(FileAndResourceDirectivesSpec.scala:47)
[info]   at scala.util.DynamicVariable.withValue(DynamicVariable.scala:58)
[info]   at akka.http.scaladsl.testkit.RouteTest$$anonfun$check$1.apply(RouteTest.scala:57)
[info]   at akka.http.scaladsl.testkit.RouteTest$$anonfun$check$1.apply(RouteTest.scala:57)
[info]   at akka.http.scaladsl.testkit.RouteTestResultComponent$RouteTestResult.$tilde$greater(RouteTestResultComponent.scala:51)
[info]   at akka.http.scaladsl.server.directives.FileAndResourceDirectivesSpec$$anonfun$1$$anonfun$apply$mcV$sp$4.apply(FileAndResourceDirectivesSpec.scala:47)
[info]   at akka.http.scaladsl.server.directives.FileAndResourceDirectivesSpec$$anonfun$1$$anonfun$apply$mcV$sp$4.apply(FileAndResourceDirectivesSpec.scala:47)
```

jrudolph:
Ah, it's a case-sensitivity issue: the test is looking for `sample space.pdf` but it's called `sample space.PDF`.

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectivesSpec.scala
2beaucoup: BTW: `RouteTest` does `request.withEffectiveUri` which ends in `URI#collapseDotSegments` which in turn cuts off the dots. Will have to find a way around that for proper tests.

2beaucoup:
Stripped out the symlink stuff for now.

@jrudolph Any idea about https://github.com/akka/akka-http/pull/348#discussion_r80496956?

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectivesSpec.scala
jrudolph: I'll have a look.

jrudolph: Maybe just add a directive that rewrites the URI back to what you wanted it to look like.

2beaucoup: ```mapRequestContext(_.withUnmatchedPath(...))``` did the trick. :)

2beaucoup:
> Ah, it's a case-sensitivity issue: the test is looking for sample space.pdf but it's called sample space.PDF.

fixed

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectives.scala
2beaucoup: updated so that nothing is logged if `!exists(file)`

2beaucoup:
Cleaned up commits and corrected tests.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectives.scala
```diff
@@ -356,30 +398,30 @@ object DirectoryListing {
   def directoryMarshaller(renderVanityFooter: Boolean): ToEntityMarshaller[DirectoryListing] =
     Marshaller.StringMarshaller.wrapWithEC(MediaTypes.`text/html`) { implicit ec ⇒ listing ⇒
       val DirectoryListing(path, isRoot, files) = listing
```
2beaucoup: Can't change `DirectoryListing#files: Seq[File]` to `Seq[Path]` because it's public.

jrudolph: I guess, that's ok.

2beaucoup: Yeah, I guess so. Prevented me from getting rid of `java.io.File` completely, though.

jrudolph:
Great, @2beaucoup. Would you say it is ready to go?

jrudolph:
It LGTM.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectives.scala
```diff
@@ -127,14 +149,28 @@ trait FileAndResourceDirectives {
...
       extractLog { log ⇒
-        fileSystemPath(base, path, log) match {
-          case ""       ⇒ reject
-          case fileName ⇒ getFromFile(fileName)
-        }
+        val pathString = (if (unmatchedPath.startsWithSlash) unmatchedPath.tail else unmatchedPath).toString
+        val file = base.resolve(pathString)
+        if (!exists(file))
+          reject
+        else if (!file.toRealPath().startsWith(base)) {
```
johanandren: This is the actual fix?

jrudolph: Yes.

jrudolph:
@2beaucoup I'd try to remix your commits a bit to simplify backporting just the fix to all the places where we'll need it.

2beaucoup:
> Great, @2beaucoup. Would you say it is ready to go?

It's ready from my POV.

> @2beaucoup I'd try to remix your commits a bit to simplify backporting just the fix to all the places where we'll need it.

Putting the last one first? No prob. I always put the cleanup stuff first but that makes it harder to backport so 👍 .

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectives.scala
2beaucoup: Yup. We no longer rely on `fileSystemPath` here.

jrudolph:
@2beaucoup what do you think about this order of commits: https://github.com/akka/akka-http/compare/pull-348-reorder-commits?expand=1

It would allow us to release a minimal update that would also work on the Java < 7 branches by just using the first two commits.

jrudolph:
(Just pushed again, as I left some imports in the wrong commit)

2beaucoup:
Nice work @jrudolph! Just verified that testing the first commit fails while the second succeeds! 👍

jrudolph:
Thanks for verifying, @2beaucoup. I had @drewhk also check and now I'm doing another round of checks on a Windows VPS. If everything goes well, we'll try to release a fix version tomorrow.

jrudolph:
Hmm I wonder if the "fix" works for the wrong reason... It seems that `Uri.Path.toString` percent encodes slashes and backslashes so the whole comparison business gets quite weird...

E.g. adding a println to `getFromDirectory` like this:

```scala
        val pathString = (if (unmatchedPath.startsWithSlash) unmatchedPath.tail else unmatchedPath).toString
        val file = base.resolve(pathString)
        println(s"Requesting pathString: $pathString file: $file realPath: ${if (exists(file)) file.toRealPath() else "[does not exist]"} (base: '$base') startsWith: ${if (exists(file)) file.toRealPath().startsWith(base).toString else "doesn't exist"}")
        if (!exists(file)) reject
```

will print for one of the examples

```
Requesting pathString: ..%2FfileA.txt file: /home/johannes/git/opensource/akka-http/akka-http-tests/target/test-classes/someDir/sub/..%2FfileA.txt realPath: [does not exist] (base: '/home/johannes/git/opensource/akka-http/akka-http-tests/target/test-classes/someDir/sub') startsWith: doesn't exist
Requesting pathString: ..%5CfileA.txt file: /home/johannes/git/opensource/akka-http/akka-http-tests/target/test-classes/someDir/sub/..%5CfileA.txt realPath: [does not exist] (base: '/home/johannes/git/opensource/akka-http/akka-http-tests/target/test-classes/someDir/sub') startsWith: doesn't exist
```

So, it will reject, but not because of the new check but because of paths containing percent encoded parts.

Also, there's another vulnerability with `getBrowseableDirectory` still listing arbitrary directories using the same kind of uris.

2beaucoup:
Only the two test cases marked with `// resolved` trigger the warning. The others are rejected via the `exists` check (which is necessary cause otherwise `toRealPath` would throw.)

`getBrowseableDirectory` indeed still needs fixing as it also uses `fileSystemPath` with `File.separatorChar`.

jrudolph:
> Only the two test cases marked with // resolved trigger the warning. The others are rejected via the exists check (which is necessary cause otherwise toRealPath would throw.)

It's not about the warning, it's about the fact that `getFromDirectory` lost the ability to serve files with percent-encoded file names.

I pushed to https://github.com/akka/akka-http/compare/pull-348-reorder-commits?expand=1 adding a new commit with a new test that succeeds before and breaks with the fix.

```scala
    "return the file content with the MediaType matching the file extension (unicode chars in filename)" in {
      Get("sample%20sp%c3%a4ce.PDF") ~> _getFromDirectory("sübdir") ~> check {
        mediaType shouldEqual `application/pdf`
        charsetOption shouldEqual None
        responseAs[String] shouldEqual "This is PDF"
        val lastModified = readAttributes(testRoot.resolve("sübdir/sample space.PDF"), classOf[BasicFileAttributes]).lastModifiedTime().toMillis
        headers should contain(`Last-Modified`(DateTime(lastModified)))
      }
    }
```

Btw. please don't feel obliged to fix these things yourself, @2beaucoup :). I can take over and do the last bits of the fix if you don't feel like it.

2beaucoup:
I see. So we could resemble what `fileSystemPath` did or "fix" `Uri.Path#toString` to not do escaping.

We came across this problem already in #23 where I suggested to rename `toString` to something else and replace it with simple concatenation. That would simplify the usage of path directives where you currently have to use patterns in their escaped form.

I don't have capacity ATM but look into it later. Thx for checking @jrudolph!

jrudolph:
> We came across this problem already in #23 where I suggested to rename toString to something else and replace it with simple concatenation. That would simplify the usage of path directives where you currently have to use patterns in their escaped form.

Thanks for confirming. Yes, that's what I had in might as well. Working on a fix...

2beaucoup:
Superseded by #359.

PR #357, opened 2016-09-27
+ht2 implement 'GET request in one HEADERS and one CONTINUATION frame'

ktoso:
Implements CONTINUATION handling, accumulates data and then feeds parser.
I was not able to make it work streaming into the buffer directly - it would fail with invalid identifiers etc, so perhaps it is not as smart about decoding as we thought. This would be a workable solution actually perhaps anyway, and we'd add a max size limit onto the headers in total in any case.

2 minor commits accompanying this PR. The meat is https://github.com/akka/akka-http/commit/87cbd235b68fe0203f918a474f4902a71008842f


akka-http2-support/src/main/scala/akka/http/impl/engine/http2/FrameParser.scala
```diff
@@ -66,7 +65,7 @@ class FrameParser(shouldReadPreface: Boolean) extends ByteStringParser[FrameEven
           if (priority) payload.readByte() & 0xff
           else 0
 
-        // TODO: check that streamId != 0
+        Http2Compliance.requirePositiveStreamId(streamId)
```
ktoso: or maybe just `requireStreamId` and it could take a Frame

jrudolph: Actually, that's not even correct: for incoming HEADERS from client it needs to be an odd streamId. We won't see even streamIds as long as we don't support sending out PUSH_PROMISEs.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/parsing/HttpRequestHeaderHpackDecompression.scala
```diff
@@ -52,28 +52,26 @@ private[http2] final class HttpRequestHeaderHpackDecompression extends GraphStag
...
 
-          // we know frames should be empty here.
-          // but for sanity lets kill that stream anyway I guess (at least for now)
-          killFramesStream(httpSubStream)
-        } else {
-          processFrame(httpSubStream, httpSubStream.initialFrame)
-          // more data in following frames
+              // we know frames should be empty here.
+              // but for sanity lets kill that stream anyway I guess (at least for now)
+              killFramesStream(httpSubStream)
```
jrudolph: Maybe we can replace it by a assertion instead? Or, we could already put this assertion into Http2SubStream?

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/parsing/HttpRequestHeaderHpackDecompression.scala
```diff
@@ -121,15 +120,22 @@ private[http2] final class HttpRequestHeaderHpackDecompression extends GraphStag
...
           val is = ByteStringInputStream(h.headerBlockFragment)
 
-          beingBuiltRequest = zeroRequest.addHeader(Http2StreamIdHeader(h.streamId))
           decoder.decode(is, this) // this: HeaderListener (invoked synchronously)
-          if (h.endHeaders) decoder.endHeaderBlock()
+          beingBuiltRequest = beingBuiltRequest.addHeader(Http2StreamIdHeader(h.streamId))
+          pushBeingBuiltRequestIf(h.endHeaders)
+
+        case c: ContinuationFrame ⇒
+          // not checking streamId, I believe we're guaranteed here by construction to have the right one hmmm...?
```
jrudolph: yep

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/parsing/HttpRequestHeaderHpackDecompression.scala
```diff
@@ -172,10 +228,10 @@ private[http2] final class HttpRequestHeaderHpackDecompression extends GraphStag
         if (httpSubStream.frames == Source.empty) ()
         else {
           // it should be empty, but wasn't, let's kill it
-          // FIXME probably this should be removed and we should just fail?
+          // FIXME these situations should likely call for PROTOCOL_ERROR instead?
```
jrudolph: Yes, just kill the stream, this has to be a bug.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -130,24 +130,22 @@ class Http2ServerSpec extends AkkaSpec with WithInPendingUntilFixed {
...
         )
       }
-      "GET request in one HEADERS and one CONTINUATION frame" inPendingUntilFixed new TestSetup with RequestResponseProbes {
-        // FIXME: needs CONTINUATION frame parsing and actual support in decompression
-
+      "GET request in one HEADERS and one CONTINUATION frame" in new TestSetup with RequestResponseProbes {
         val headerBlock = HPackSpecExamples.C41FirstRequestWithHuffman
-        val fragment1 = headerBlock.take(5)
-        val fragment2 = headerBlock.drop(5)
+        val fragment1 = headerBlock.take(8) // must be grouped by octets
```
jrudolph: Hmm IIUC octet = byte, so it should work with any kind of fragmentation?

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -468,6 +466,7 @@ class Http2ServerSpec extends AkkaSpec with WithInPendingUntilFixed {
       Flow.fromSinkAndSource(Sink.fromSubscriber(requestIn), Source.fromPublisher(responseOut))
 
     def expectRequest(): HttpRequest = requestIn.requestNext().removeHeader("x-http2-stream-id")
+    def expectRequestRaw(): HttpRequest = requestIn.requestNext() // TODO, make it so that internal headers are not listed in `headers` etc?
```
jrudolph: Right, I also suspected that we would need this at some point.

jrudolph: On the other hand, as we know the stream id in this case, we can just construct the right header.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/FrameParser.scala
ktoso: Ah, so similar to websockets if i remember correctly, amending

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/parsing/HttpRequestHeaderHpackDecompression.scala
ktoso: yeah, blowing up makes sense here. I'll think about how

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
ktoso: Yeah, not sure why I was thinking on byte level

ktoso: // Current impl works with any amount taken

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/FrameParser.scala
ktoso: Hm, then we have this check in too many places right now... I'll remove it for now so we can merge and add those later

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/parsing/HttpRequestHeaderHpackDecompression.scala
ktoso: (assert kills the actor system btw since `Error`, so using require)

ktoso:
Addressed some of the comments but calling it a night for today.
Did not end up doing any more work on HTTP/2 sadly today.

Feel free to merge and base your work on top of this, I'll get some work squeezed in in-flights tomorrow.

PR #359, opened 2016-09-28
Fix directory traversal in Windows in getFromDirectory / listDirectoryContents

### LINKED ISSUE #346: Directory Traversal Attack vulnerability in Akka HTTP's getFrom(Browseable)Directory directive

jrudolph:
@2beaucoup here's a new fix without all your extra changes. AFAICS these should apply also to former versions for which we might have to issue a fix release.

WDYT about containing the changes inside of the new `safeDirectoryChildPath` (old `fileSystemPath`)?

Fixes #346.

### DIFF
--- a/akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectivesSpec.scala
+++ b/akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectivesSpec.scala
@@ -20,11 +20,14 @@ import akka.http.scaladsl.model._
 import akka.http.scaladsl.model.headers._
 import akka.http.impl.util._
 import akka.http.scaladsl.TestUtils.writeAllText
+import akka.http.scaladsl.model.Uri.Path
 
 class FileAndResourceDirectivesSpec extends RoutingSpec with Inspectors with Inside {
 
   // operations touch files, can be randomly hit by slowness
   implicit val routeTestTimeout = RouteTestTimeout(3.seconds)
+
+  val testRoot = new File(getClass.getClassLoader.getResource("").toURI)
 
   override def testConfigSource = "akka.http.routing.range-coalescing-threshold = 1"
 
@@ -121,6 +124,52 @@ class FileAndResourceDirectivesSpec extends RoutingSpec with Inspectors with Ins
           responseAs[String] shouldEqual "456"
         }
       } finally file.delete
+    }
+  }
+
+  "getFromDirectory" should {
+    def _getFromDirectory(directory: String) = getFromDirectory(new File(testRoot, directory).getCanonicalPath)
+
+    "reject non-GET requests" in {
+      Put() ~> _getFromDirectory("someDir") ~> check { handled shouldEqual false }
+    }
+    "reject requests to non-existing files" in {
+      Get("nonExistentFile") ~> _getFromDirectory("subDirectory") ~> check { handled shouldEqual false }
+    }
+    "reject requests to directories" in {
+      Get("sub") ~> _getFromDirectory("someDir") ~> check { handled shouldEqual false }
+    }
+    "reject path traversal attempts" in {
+      def route(uri: String) =
+        mapRequestContext(_.withUnmatchedPath(Path("/" + uri))) { _getFromDirectory("someDir/sub") }
+
+      Get() ~> route("file.html") ~> check { handled shouldEqual true }
+
+      def shouldReject(prefix: String) =
+        Get() ~> route(prefix + "fileA.txt") ~> check { handled shouldEqual false }
+      shouldReject("../") // resolved
+      shouldReject("%5c../")
+      shouldReject("%2e%2e%2f")
+      shouldReject("%2e%2e/") // resolved
+      shouldReject("..%2f")
+      shouldReject("%2e%2e%5c")
+      shouldReject("%2e%2e\\")
+      shouldReject("..\\")
+      shouldReject("\\")
+      shouldReject("%5c")
+      shouldReject("..%5c")
+      shouldReject("..%255c")
+      shouldReject("..%c0%af")
+      shouldReject("..%c1%9c")
+    }
+    "return the file content with the MediaType matching the file extension" in {
+      Get("fileA.txt") ~> _getFromDirectory("someDir") ~> check {
+        mediaType shouldEqual `text/plain`
+        charsetOption shouldEqual Some(HttpCharsets.`UTF-8`)
+        responseAs[String] shouldEqual "123"
+        val lastModified = new File(testRoot, "someDir/fileA.txt").lastModified()
+        headers should contain(`Last-Modified`(DateTime(lastModified)))
+      }
     }
   }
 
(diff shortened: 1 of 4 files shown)


jrudolph:
We should follow this up with your other changes for modernizing the infrastructure to java.nio.file.

2beaucoup:
Looks good @jrudolph. A bit sad that that we have to resort to basic string matching.

It's just the nio stuff that's missing here, right?

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectivesSpec.scala
```diff
@@ -27,7 +27,8 @@ class FileAndResourceDirectivesSpec extends RoutingSpec with Inspectors with Ins
   // operations touch files, can be randomly hit by slowness
   implicit val routeTestTimeout = RouteTestTimeout(3.seconds)
 
-  val testRoot = new File(getClass.getClassLoader.getResource("").toURI)
+  val testRoot = new File("akka-http-tests/src/test/resources")
+  require(testRoot.exists(), s"testRoot was not found at ${testRoot.getAbsolutePath}")
```
2beaucoup: Why hardcode the `testRoute` path?

jrudolph: Because, otherwise, files will be served from the `target` directory where sbt will have copied all files over, resolving symlinks in the process.

jrudolph:
> A bit sad that that we have to resort to basic string matching.

Yes, but it's only for this version. We should still fix it using the nio stuff in the end. I'd just like to keep the actual fix small for now and compatible to older versions.

> It's just the nio stuff that's missing here, right?

Yes, basically all your cleanups are still missing (but still highly appreciated). If you'd like you could port them over in a new PR, otherwise, I'd try to do it when I get to it (but this might take a while because of upcoming traveling next week).

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectives.scala
```diff
@@ -217,14 +217,33 @@ object FileAndResourceDirectives extends FileAndResourceDirectives {
       BasicDirectives.extractSettings
 
   private def withTrailingSlash(path: String): String = if (path endsWith "/") path else path + '/'
-  private def fileSystemPath(base: String, path: Uri.Path, log: LoggingAdapter, separator: Char = File.separatorChar): String = {
+
+  /**
+   * Given a base directory and a (Uri) path, returns a path to a location contained in the base directory,
+   * while checking that no path traversal is possible. Path traversable is prevented by two individual measures:
```
ktoso: Path traversable is prev => Path traversabl is prev

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectives.scala
```diff
@@ -217,14 +217,33 @@ object FileAndResourceDirectives extends FileAndResourceDirectives {
...
+   *  - A path segment must not be ".." and must not contain slashes or backslashes that may carry special meaning in
+   *    file-system paths. This logic is intentionally a bit conservative as it might also prevent legitimate access
```
ktoso: good

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectives.scala
```diff
@@ -217,14 +217,33 @@ object FileAndResourceDirectives extends FileAndResourceDirectives {
...
+
+  private def safeJoinPaths(base: String, path: Uri.Path, log: LoggingAdapter, separator: Char = File.separatorChar): String = {
     import java.lang.StringBuilder
     @tailrec def rec(p: Uri.Path, result: StringBuilder = new StringBuilder(base)): String =
       p match {
         case Uri.Path.Empty       ⇒ result.toString
         case Uri.Path.Slash(tail) ⇒ rec(tail, result.append(separator))
         case Uri.Path.Segment(head, tail) ⇒
-          if (head.indexOf('/') >= 0 || head == "..") {
+          if (head.indexOf('/') >= 0 || head.indexOf('\\') >= 0 || head == "..") {
```
ktoso: good

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectives.scala
```diff
@@ -233,6 +252,24 @@ object FileAndResourceDirectives extends FileAndResourceDirectives {
...
+   */
+  private def checkIsSafeDescendant(basePath: String, finalPath: String, log: LoggingAdapter): String = {
+    val baseFile = new File(basePath)
+    val finalFile = new File(finalPath)
+    val canonicalFinalPath = finalFile.getCanonicalPath
+
+    if (!canonicalFinalPath.startsWith(baseFile.getCanonicalPath)) {
+      log.warning(s"[$finalFile] points to a location that is not part of [$baseFile]. This might be a directory " +
+        "traversal attempt.")
+      ""
```
ktoso: why "" and not throwing here?

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectives.scala
jrudolph: Thanks ;)

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectives.scala
jrudolph: I just kept the previous behavior of just rejecting the request.

jrudolph: Otherwise, there might be a higher risk of revealing some information if the response is not the same as if the file is just missing.

ktoso: I see, that makes sense. Information exposure would be bad too

jrudolph:
Merged. Thanks a lot again @2beaucoup.

2beaucoup:
👍 I'll close the other one then.

jrudolph:
Backport to old repo (akka 2.4.x): https://github.com/akka/akka/pull/21584
Backport to stream-http-2.0: https://github.com/akka/akka/pull/21586

PR #371, opened 2016-10-05
#368 Capture user handler in a termination future

### LINKED ISSUE #368: Requests with closed upstreams are not handled with `max-connections`
Reproducer [here](https://github.com/svezfaz/akka-http-proxy-example)

The problem is that when `max-connections` is set to a lower value than gatling load test opens connections, some incoming connections are buffered in the OS kernel throughout all of the test. When the test finishes, these buffered connections are closed by gatling, but akka-http server still handles them  without respecting `max-connections` setting.

2m:
Fixes #368

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
@@ -17,7 +17,6 @@ import akka.http.impl.engine.client._
 import akka.http.impl.engine.server._
 import akka.http.impl.engine.ws.WebSocketClientBlueprint
 import akka.http.impl.settings.{ ConnectionPoolSetup, HostConnectionPoolSetup }
-import akka.http.impl.util.{ MapError, StreamUtils }
 import akka.http.scaladsl.model._
 import akka.http.scaladsl.model.headers.Host
 import akka.http.scaladsl.model.ws.{ Message, WebSocketRequest, WebSocketUpgradeResponse }
@@ -137,8 +136,8 @@ class HttpExt(private val config: Config)(implicit val system: ActorSystem) exte
 
     val fullLayer: Flow[ByteString, ByteString, Future[Done]] = Flow.fromGraph(Fusing.aggressive(
       Flow[HttpRequest]
-        .watchTermination()(Keep.right)
         .viaMat(handler)(Keep.left)
+        .watchTermination()(Keep.right)
         .joinMat(fuseServerLayer(settings, connectionContext, log))(Keep.left)))
 
     val connections: Source[Tcp.IncomingConnection, Future[Tcp.ServerBinding]] =
--- a/akka-http-tests/src/test/scala/akka/http/scaladsl/server/RespectConnectionLimitSpec.scala
+++ b/akka-http-tests/src/test/scala/akka/http/scaladsl/server/RespectConnectionLimitSpec.scala
@@ -1,0 +1,98 @@
+/**
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.http.scaladsl.server
+
+import akka.actor._
+import akka.http.scaladsl.Http
+import akka.http.scaladsl.model._
+import akka.http.scaladsl.server.Directives._
+import akka.stream._
+import akka.stream.scaladsl._
+import akka.testkit.TestKit
+import io.gatling.app.Gatling
+import io.gatling.core.scenario.{ Simulation ⇒ GatlingSim }
+import org.scalatest._
+import java.util.concurrent.atomic._
+import scala.concurrent._
+import scala.concurrent.duration._
+import scala.util._
+import com.typesafe.config.ConfigFactory
+
+class RespectConnectionLimitSpec extends AsyncWordSpec with Matchers with BeforeAndAfterAll {
+  import RespectConnectionLimitSpec._
+
+  implicit val system = ActorSystem("RespectConnectionLimitSpec", ConfigFactory.parseString(s"""
+    akka.http.server.max-connections = $MaxConnections
+  """))
+  implicit val materializer = ActorMaterializer()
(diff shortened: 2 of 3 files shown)


akka-http-tests/src/test/scala/akka/http/scaladsl/server/RespectConnectionLimitSpec.scala
```diff
@@ -0,0 +1,98 @@
...
+        }
+
+      Http().bindAndHandle(handler, "localhost", 0).map { binding ⇒
+        Simulation.port = binding.localAddress.getPort
+        Gatling.fromArgs(Array("--mute", "--no-reports"), Some(classOf[Simulation.TooManyUsers].asInstanceOf[Class[GatlingSim]]))
+        tooManyConnections.get shouldBe false
+      }
+    }
+
+  }
```
ktoso: how heavy is this test? Are we ok to run it always?

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
```diff
@@ -137,8 +136,8 @@ class HttpExt(private val config: Config)(implicit val system: ActorSystem) exte
 
     val fullLayer: Flow[ByteString, ByteString, Future[Done]] = Flow.fromGraph(Fusing.aggressive(
       Flow[HttpRequest]
-        .watchTermination()(Keep.right)
         .viaMat(handler)(Keep.left)
+        .watchTermination()(Keep.right)
```
ktoso: add a comment that the position of this here specifically _matters_?

ktoso:
So to my understanding it's technically still racy, but "less", as the termination after user handler right?

akka-http-tests/src/test/scala/akka/http/scaladsl/server/RespectConnectionLimitSpec.scala
2m: Currently it takes ~ 5 seconds to run. But that is tunable by changing gatling test setup.

ktoso: I'd want to make those run only when a flag is set then.
Plz :)

2m: Will anyone ever gonna run these tests then? :) I can move them to `it:test` as an alternative?

ktoso: jenkins will. same as multinode tests, no-one runs them in a distributed fashion when developing.

2m: Right. Fair enough. But I still think it makes more sense to move that to it:test instead of adding a flag. Cleaner build and more according to the sbt convetions.

ktoso: Yep, it:test sounds very good

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
johanandren: "THIS NEEDS TO BE HERE, TRUST ME"

jrudolph:
Should I try to write a test without gatling for that one or do you want to follow up, @2m?

ktoso:
Any chance to get a test rewrite from you here early next week @2m ?

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
ktoso: Please add a comment before we merge here :)

2m:
So my WIP branch where I swapped gatling with plain SocketChannels is here: https://github.com/2m/akka-http/commit/e387990ff4d945e553b5095408c989047cea78d6

The problem there is that even with the fix proposed in this PR (moving `watchTermination` under user handler) connection count gets larger that defined in configuration. So opening connections with socket channels exercises another codepath than gatling tests.

ktoso:
Didn't make it for RC1, but we should try to get this in soon :) Definitely for 3.0.0.

jrudolph:
@2m using `watchTermination` will not allow you to observe the events in sequence because there's another "dispatch hop" going on from `watchTermination` signalling the future and handling the future. So, the test wasn't really sound because it couldn't prove a happens-before relationship between completion of the user handler and materializing the handler for another connection.

jrudolph:
I think the real issue here is that it is impossible to observe the completion of the user handler flow (or any  flow component). What happens here is that completion and cancellation race from both sides into user handler and even if we'd kept `watchTermination` at both ends we would only observe that we send signals into the stage but not what the stage did with those signals:

 * it might support half-closing in which case it might just gulp down one of the signals and still wait for the other one
 * the stage could be slow so that it might take a while processing the completion/cancellation event. So, even if we know it should have received both the completion and the cancellation, we still don't know if it has executed its postStop method (or even processed all incoming requests before the completion).

I think the only way to actual improve this is defining an explicit close protocol (as always) where we prevent the race between completion and cancellation. I guess for the HTTP handler it could be this:

We never cancel immediately but either send completion (if the connection was closed expectedly) or an error (if it was closed during receiving a request). Then we leave a bit of headroom for completion (or error) to arrive from the user handler. In that case I think we can be sure that every processing is finished (though `postStop` may still run later?). If completion wasn't propagated after some time, we also cancel (just in case) and also mark processing as finished.

Didn't we talk about a general stage like this a while ago?

ktoso:
What shall we do about this PR?

jrudolph:
I created https://github.com/akka/akka-http/pull/473 to just document the guarantees instead of changing any code. Closing this one for now.

PR #372, opened 2016-10-06
Close blueprint on no body early complete

### LINKED ISSUE #367: Memory leak when posting requests to akka-http without consuming the response
Migrated from akka/akka#21621, reported by @jeroentervoorde 

Hi,

I think i discovered a memory leak when sending a lot of posts requests to a server without consuming the response and just closing the connection. 

I am using akka-http 2.4.11 and scala 2.11.8

We've ran into in issue because we're using a proxy which duplicates traffic to staging and dev and also does not read the response. Our application goes OOM after a few hours.

I created a project to reproduce the issue here: https://github.com/jeroentervoorde/akka-http-memory-leak.git

This consists of a Client and a Server application. These can be run from Idea

Client posts 10 requests to a socket and closes the socket. This is repeated 4242 times (i used this number to be easily recognizable in yourkit)

For each loop some objects remain allocated and are never garbage collected.

Here are some yourkit screenshots of Server after running the Client:

Initial: 
filtered on "Stage"

image

filtered on "Error"
image

After running Client:  

filtered on "Stage"
image

filtered on "Error"
image

filtered on "Actor"
image

johanandren:
Fixes #367, not sure about other potential consequences though.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
@@ -115,8 +115,8 @@ private[http] object HttpServerBluePrint {
           // application layer has cancelled or only partially consumed response entity:
           // connection will be closed
           entitySource.complete()
-          completeStage()
         }
+        completeStage()
       }
 
       override def onPush(): Unit = grab(in) match {


johanandren:
Ping @jrudolph @ktoso

johanandren:
PLS BUILD

ktoso:
LGTM I think, cc @jrudolph

jrudolph:
A test would be nice :)

johanandren:
I tried to write one, but since it is about the stream for one connection staying alive, but being shutdown when the materializer/akka-http stops I couldn't figure out a way to detect it staying in memory (assertAllStagesStopped wouldn't help here).

PR #374, opened 2016-10-06
Fix typo in the spray migration guide

theon:
First argument of MediaType.applicationWithFixedCharset is only the subType, not the full MediaType name.


theon:
@ktoso You mentioned there was a second occurrence of the typo on this page, but I couldn't find it. There is a `MediaType.custom("application/vnd.acme+json")`, but I think is valid spray code?

ktoso:
LGTM, thanks! You're right, I missed that there it was correct actually

PR #381, opened 2016-10-11
=htp #366 disable content-type negotiation for responses marshalled from status codes

### LINKED ISSUE #366: complete(StatusCodes.NoContent) fails with UnacceptedResponseContentTypeRejection for Accept(MediaTypes.`application/json`) request
When a route completes with `StatusCodes.NoContent` the marshalling will fail with `StatusCodes.NotAcceptable` if the request doesn't accept text/plain.

The following test fails with `Request was rejected with rejection UnacceptedResponseContentTypeRejection(Set(MediaType(text/plain)))`

```scala
import akka.http.scaladsl.model.headers.Accept
import akka.http.scaladsl.model.{MediaTypes, StatusCodes}
import akka.http.scaladsl.server.{Directives, Route}
import akka.http.scaladsl.testkit.ScalatestRouteTest
import org.scalatest.{Matchers, WordSpecLike}

class NoContentSpec extends WordSpecLike with Matchers with ScalatestRouteTest with Directives {

  val noContentRoute: Route = complete(StatusCodes.NoContent)

  "The noContentRoute" should {
    "return HTTP NoContent for accept header `application/json`" in {
      Get("/").withHeaders(Accept(MediaTypes.`application/json`)) ~> noContentRoute ~> check {
        status shouldBe StatusCodes.NoContent
      }
    }
  }
}
```

As a work-around `complete(HttpResponse(StatusCodes.NoContent, entity = HttpEntity.Empty))` can be used instead.

jrudolph:
RFC 7231 3.4.1 states:

```
   A user agent cannot rely on proactive negotiation preferences being
   consistently honored, since the origin server might not implement
   proactive negotiation for the requested resource or might decide that
   sending a response that doesn't conform to the user agent's
   preferences is better than sending a 406 (Not Acceptable) response.
```

So, especially for status code responses which usually won't contain any
actual "resource" at all, it makes more sense to return the status code
and a `text/plain` response with the status codes default message instead
of requiring the client to accept `text/plain` just to support receiving
status information.

Fixes #366.

/cc @leachbj
/cc @gosubpl (since you changed the code the last time)

### DIFF
--- a/akka-http-testkit/src/main/scala/akka/http/scaladsl/testkit/MarshallingTestUtils.scala
+++ b/akka-http-testkit/src/main/scala/akka/http/scaladsl/testkit/MarshallingTestUtils.scala
@@ -8,7 +8,8 @@ import scala.concurrent.duration._
 import scala.concurrent.{ Await, ExecutionContext }
 import akka.http.scaladsl.unmarshalling.{ FromEntityUnmarshaller, Unmarshal }
 import akka.http.scaladsl.marshalling._
-import akka.http.scaladsl.model.{ HttpEntity, HttpRequest, HttpResponse }
+import akka.http.scaladsl.model.headers.Accept
+import akka.http.scaladsl.model.{ HttpEntity, HttpRequest, HttpResponse, MediaRange }
 import akka.stream.Materializer
 
 import scala.util.Try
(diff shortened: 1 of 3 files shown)


leachbj:
This looks similar to a local hack I tested using Marshaller.opaque which worked for me.

gosubpl:
@jrudolph much better now. Should've considered that case. Thanks!

ktoso:
LGTM, thanks @jrudolph and @gosubpl for reviewing!

Issue #11, opened 2016-08-26
Create `@example` "include" pre-processor feature for java/scala-doc


ktoso:
We want to use this to power examples for all directives.
It will make it easier to maintain the directive docs and would also address https://github.com/akka/akka/issues/18660

jonas:
References #187 

I'm interested in working on this. Should I try to land the tooling upstream in Paradox or do we want to iterate on it in this project before suggesting a fix for lightbend/paradox#26

jonas:
In my investigation so far I didn't manage to get a Scaladoc preprocessor working. These are the main examples I've looked at so far:

- https://github.com/szeiger/extradoc
- https://github.com/todesking/sbt-custom-doclet and https://github.com/todesking/scala-json-doclet

I wonder if for the Scala directives it would be in scope to use scala.meta or [the approach of sbt-doctest](https://github.com/tkawachi/sbt-doctest/blob/master/src/main/scala/com/github/tkawachi/doctest/ScaladocExtractor.scala) to extract the directive scaladoc and the directive signature? That is we would be able to copy paste the markdown to the scaladoc but keep the directives documentation more or less as it currently is. This was why I created lightbend/paradox#26 although I should probably have discussed this here first.

jrudolph:
I think we should do the processing directly on the source files (probably through sbt). It is important that the published source artifact also contains all the examples because many if not most people will read the scaladoc either through an IDE (which will pick up the documentation from the source files) or by reading the source code to which the IDE will link to.

jonas:
First prototype for the scala side is [here](https://jonas.github.io/akka-http/index.html#akka.http.scaladsl.server.Directives) using jonas/akka-http@89a8a3ab07c56f80313ace4069b7be44ad4acdbd

2m:
It would be interesting to see if the new scaladoc (example of the [new output here](http://www.scala-lang.org/api/2.12.0-RC2/scala/collection/immutable/List.html)) might fix some usability issues.

jonas:
I will try. That would be a good excuse to configure cross building for Scala 2.12.

jrudolph:
:+1: An old version of the akka docs generated with 2.12 can be found here: akka-docs.virtual-void.net

jonas:
OK, got a bit further using scalameta to extract signatures and comments. This would require to have the tooling separate from sbt (until version 1.0) since scalameta only supports 2.11+.

2m:
[...]

PR #389, opened 2016-10-12
=doc create new `@signature` directive to support directive documentation

jrudolph:
This is basically a port from the old `includecode2.py` script customization we built for spray.

I guess that's as good as it gets without spending too much time on it if #11 is the way to go eventually.

WDYT, @jonas?

### DIFF
--- a/project/ParadoxSignatureDirective.scala
+++ b/project/ParadoxSignatureDirective.scala
@@ -1,0 +1,55 @@
+package akka
+
+import sbt._
+import java.io.{File, FileNotFoundException}
+
+import com.lightbend.paradox._
+import markdown.Writer
+import com.lightbend.paradox.sbt.ParadoxPlugin.autoImport._
+import com.lightbend.paradox.markdown.{LeafBlockDirective, _}
+import org.pegdown.Printer
+import org.pegdown.ast.{DirectiveNode, VerbatimNode, Visitor}
+
+import scala.collection.JavaConverters._
+import scala.io.Source
+
+object ParadoxSupport {
+  val paradoxWithSignatureDirective = Seq(
+    paradoxProcessor in Compile ~= { _ =>
+      // FIXME: this is a HACK so far that copies stuff over from paradox
+      // it would be better if the plugin has a way of specifying extra directives through normal sbt mechanisms
+      // see https://github.com/lightbend/paradox/issues/35
+      new ParadoxProcessor(writer =
+        new Writer(serializerPlugins = context =>
+          Seq(
+          new ActiveLinkSerializer,
+          new AnchorLinkSerializer,
+          new DirectiveSerializer(Writer.defaultDirectives(context) :+
+            new SignatureDirective(context.location.tree.label)
+      ))))
+    }
+  )
+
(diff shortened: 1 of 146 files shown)


project/ParadoxSignatureDirective.scala
```diff
@@ -0,0 +1,55 @@
+package akka
+
+import sbt._
+import java.io.{File, FileNotFoundException}
+
+import com.lightbend.paradox._
+import markdown.Writer
```
jonas: Looks like this is imported by `com.lightbend.paradox.markdown.{LeafBlockDirective, _}` below

project/ParadoxSignatureDirective.scala
```diff
@@ -0,0 +1,55 @@
...
+        //println(s"Looking for signature regex '$Signature'")
+        val text =
+          Source.fromFile(file).getLines.collect {
+            case line@Signature(full, l, _*) if labels contains l =>
+              //println(s"Found label '$l' with sig '$full' in line $line")
+              full
+          }.mkString("\n")
+
+        val lang = Option(node.attributes.value("type")).getOrElse(Snippet.language(file))
+        new VerbatimNode(text, lang).accept(visitor)
```
jonas: This should fail if text is empty

jonas:
FWIW, LGTM

project/ParadoxSignatureDirective.scala
jrudolph: Thanks, will do.

project/ParadoxSignatureDirective.scala
jrudolph: Yes, good point.

jrudolph: Ha, this directly catched a problem when referencing a type alias in `authenticateBasic.md`.

jrudolph:
Thanks for the comments, @jonas. I've now added a warning when the directive doesn't match at all (otherwise, if we'd threw an exception, we'd need paradox support to make it report multiple failures. Also we inject error text directly into the document.

We can now only include `type` definitions in which case the complete definition is included and not just the signature. Maybe this extra handling could be replaced by more generic parameter to the directive whether to include the complete line but as we still essentially want to get rid of the directive pages here, I guess it doesn't matter too much.

jonas:
👍 Paradox' embrace of "let it crash" on first failure was not fun during the doc migration.

project/ParadoxSignatureDirective.scala
```diff
@@ -0,0 +1,65 @@
...
+            case line@Signature(signature, kind, l, definition) if labels contains l.toLowerCase() =>
+              //println(s"Found label '$l' with sig '$full' in line $line")
+              if (kind == "type") signature + definition
+              else signature
+          }.mkString("\n")
+
+        if (text.trim.isEmpty) {
+          logWarn(
+            s"Did not find any signatures with one of those names [${labels.mkString(", ")}] in ${node.source} " +
+            s"(was referenced from [${page.path}])")
```
ktoso: Ok, though in final versions I'd prefer to fail (accumulate many errors and fail) but paradox can't do that right now.

ktoso:
LGTM, nice - thanks @jrudolph !

Issue #396, opened 2016-10-16
Cannot push port (requestPrepOut) twice on request timeout


kstokoz:
Consider following code:

``` scala

import akka.actor.ActorSystem
import akka.http.scaladsl.Http
import akka.http.scaladsl.server.Directives._
import akka.http.scaladsl.server.Route
import akka.stream.ActorMaterializer

import scala.concurrent.duration._
import scala.concurrent.{Await, Promise}

object ReproducerApp extends App {

  implicit val actorSystem = ActorSystem()
  implicit val ec = actorSystem.dispatcher
  implicit val materializer = ActorMaterializer()

  private val timeout = 1.second
  private val waitTime = 2.second

  val route: Route =
    (path("testPost") & post) {
      withRequestTimeout(timeout) {
        val promise = Promise[String]()

        actorSystem.scheduler.scheduleOnce(waitTime) {
          promise.success("Timeout didn't work!")
        }

        complete {
          promise.future
        }
      }
    }

  val bindingFuture = Http().bindAndHandle(route, "localhost", 12345)

  System.in.read()
  Await.ready(actorSystem.terminate(), 5.seconds)
}
```

Executing: 

```
curl -vvv --data @fileAtLeast10Kb "http://localhost:12345/testPost"
```

Would produce:

```
[ERROR] [10/16/2016 22:20:57.381] [default-akka.actor.default-dispatcher-24] [akka://default/user/StreamSupervisor-0/flow-15-0-unknown-operation] Error in stage [akka.http.impl.engine.server.HttpServerBluePrint$ControllerStage@5c8aff50]: requirement failed: Cannot push port (requestPrepOut) twice
java.lang.IllegalArgumentException: requirement failed: Cannot push port (requestPrepOut) twice
    at scala.Predef$.require(Predef.scala:224)
    at akka.stream.stage.GraphStageLogic.push(GraphStage.scala:459)
    at akka.http.impl.engine.server.HttpServerBluePrint$ControllerStage$$anon$12$$anon$13.onPush(HttpServerBluePrint.scala:400)
    at akka.stream.impl.fusing.GraphInterpreter.processPush(GraphInterpreter.scala:747)
    at akka.stream.impl.fusing.GraphInterpreter.execute(GraphInterpreter.scala:649)
    at akka.stream.impl.fusing.GraphInterpreterShell.runBatch(ActorGraphInterpreter.scala:471)
[...]

PR #405, opened 2016-10-18
WIP =htp #396 minimal guard against double pushing on 100-continue pending

ktoso:
WIP on #396

Solves one part of the issue.
However timing the request timeout is rather different... Note that if reproducer gets:

```
    (path("testPost") & post) {
      extractRequestEntity { ent: RequestEntity ⇒
        ent.discardBytes()
```

then everything works. Needs more research, I'll head home and work more on it there.

// cc @jrudolph

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
@@ -391,11 +391,15 @@ private[http] object HttpServerBluePrint {
               push(requestPrepOut, rs)
             case MessageEnd ⇒
               messageEndPending = false
-              push(requestPrepOut, MessageEnd)
+              if (!oneHundredContinueResponsePending) push(requestPrepOut, MessageEnd)
             case MessageStartError(status, info) ⇒ finishWithIllegalRequestError(status, info)
             case x: EntityStreamError if messageEndPending && openRequests.isEmpty ⇒
               // client terminated the connection after receiving an early response to 100-continue
               completeStage()
+            case x if oneHundredContinueResponsePending ⇒
+              log.debug("Dropping incoming entity since `oneHundredContinueResponsePending` still active")
+              pull(requestParsingIn)
+
             case x ⇒
               push(requestPrepOut, x)
           }
--- a/akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/TimeoutDirectivesSpec.scala
+++ b/akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/TimeoutDirectivesSpec.scala
@@ -4,8 +4,10 @@
 
 package akka.http.scaladsl.server.directives
 
-import akka.http.scaladsl.model.{ HttpResponse, StatusCodes }
+import akka.http.scaladsl.model.{ ContentTypes, HttpEntity, HttpResponse, StatusCodes }
 import akka.http.scaladsl.server.IntegrationRoutingSpec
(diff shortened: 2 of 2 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -391,11 +391,15 @@ private[http] object HttpServerBluePrint {
...
               messageEndPending = false
-              push(requestPrepOut, MessageEnd)
+              if (!oneHundredContinueResponsePending) push(requestPrepOut, MessageEnd)
             case MessageStartError(status, info) ⇒ finishWithIllegalRequestError(status, info)
             case x: EntityStreamError if messageEndPending && openRequests.isEmpty ⇒
               // client terminated the connection after receiving an early response to 100-continue
               completeStage()
+            case x if oneHundredContinueResponsePending ⇒
+              log.debug("Dropping incoming entity since `oneHundredContinueResponsePending` still active")
+              pull(requestParsingIn)
```
jrudolph: Will this work? What kind of elements do we discard here? We might end up in an illegal state so it might make sense to just kill the connection?

jypma:
The reproducer in https://github.com/akka/akka-http/issues/448 still fails with 
```
13:40:27,251 ERROR [a.s.Materializer] akka.stream.Log(akka://http-integration-spec/user/StreamSupervisor-0) - [counted] Upstream failed.
java.lang.IllegalArgumentException: requirement failed: Cannot pull port (requestParsingIn) twice
	at scala.Predef$.require(Predef.scala:224)
	at akka.stream.stage.GraphStageLogic.pull(GraphStage.scala:355)
	at akka.http.impl.engine.server.HttpServerBluePrint$ControllerStage$$anon$12$$anon$15.onPush(HttpServerBluePrint.scala:436)
```
when ran against this branch.

jypma:
And my more complex server, for which I haven't been able to write a minimal reproducer yet, also still fails with
```
info] java.lang.IllegalArgumentException: requirement failed: Cannot push port (requestPrepOut) twice
[info] 	at scala.Predef$.require(Predef.scala:224)
[info] 	at akka.stream.stage.GraphStageLogic.push(GraphStage.scala:459)
[info] 	at akka.http.impl.engine.server.HttpServerBluePrint$ControllerStage$$anon$12$$anon$13.onPush(HttpServerBluePrint.scala:394)
```

jypma:
Looking at the PR code, that's also kinda expected, since in my case, it looks (?) like the double-push occurs _after_ both the `100 Continue` _and_ the `500 Internal server error` responses have already been written.

ktoso:
@jypma, this is addressing directly the #396 reproducer, not #448. Baby steps... Happy to take help in addressing the other ones :)

ktoso:
I guess I need to get back to this one on a bugfixing day

jypma:
@ktoso I totally get that, good progress either way. In order for me to more directly contribute on this one, would however require me studying the state machine of `HttpServerBluePrint` more closely, and seeing if I can make it misbehave directly by sending it exact sequences of messages.

I may not get to that in the short term, since for us the problem only shows when both (A) manual testing with `curl`, since it insists on using `100-Continue`, and (B) the request itself is failing asynchronously mid-stream. Hence, fairly low priority. But I'll let you know if I do find something.

jrudolph:
I'm closing this as I have a simpler patch that seems to work in #869. I think the problem with this fix was that it would disallow a use case where the request entity is read at the same time as the response entity is already sent out.

PR #409, opened 2016-10-19
Move common settings to an autoplugin 

### LINKED ISSUE #408: Published artifacts don't have scala cross version
3.0.0-RC1, noticed that the artifacts are missing the scala suffix.
Need to fix this of course.

2m:
Fixes #408

`inThisBuild` sets project scope to the current build, but settings set in the `/build.sbt` are added to the root project only.

Instead moved common settings to an AutoPlugin that is applied to every project.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,36 +1,6 @@
 import com.typesafe.sbt.pgp.PgpKeys._
 import akka._
 
-inThisBuild(Def.settings(
-  organization := "com.typesafe.akka",
-  organizationName := "Lightbend",
-  organizationHomepage := Some(url("https://www.lightbend.com")),
-  homepage := Some(url("http://akka.io")),
-  scmInfo := Some(
-    ScmInfo(url("https://github.com/akka/akka-http"), "git@github.com:akka/akka-http.git")),
-  developers := List(
-    Developer("contributors", "Contributors", "akka-user@googlegroups.com",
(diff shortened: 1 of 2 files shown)


ktoso:
LGTM, thanks for the fix. Very weird that this is how inThisBuild works there... Oh well

ktoso:
I'll release RC1 again, so it has the proper version suffix.

jrudolph:
I don't think this was necessary. We set `crossPaths := false` in Publish.scala which is what disables the version suffixes.

ktoso:
Why did we have crossPaths false anywhere in the first place...? Weird.
I think you're right.

I published the correct artifacts now (@2m's patch also works, but we should remove that `crossPaths := false` I think)

jrudolph:
Yes, we need to remove that. Otherwise, we might rely on the ordering of loading plugins.

2m:
`crossPaths := false` makes sure that there is only one target directory even when cross building. It is required during the release process. So everything ends up in the same directory, which then can be uploaded to S3.

ktoso:
S3 upload what? We don't do a distribution here hm.

2m:
Yes, you are right. Akka Http does not publish to S3. Akka does. And that is where `crossPaths := false` came from. So from release point of view having `crossPaths := false` is fine.

ktoso:
Uhm, you mean removing it is fine, since we don't use it - we don't have a "release" (zip) here,

PR #413, opened 2016-10-20
Fix SBT's ambiguous keys error for crossVersion

jonas:
Remove the crossVersion setting since SBT will automatically set it to
`CrossVersion.binary` when `crossPaths` is `true` which is the default.
As for the cause, defining it inside `inThisPlugin` creates a conflict
with SBT's default which is automatically added to each subproject.

Fixes the following error reported for `show crossVersion`:

    akka-http-root > show crossVersion
    [error] Ambiguous keys: {.}/*:crossVersion, *:crossVersion
    [error] show crossVersion
    [error]                  ^

---

With this I get the following:
```
akka-http-root > show crossVersion
[info] akka-http-testkit/*:crossVersion
[info] 	Binary
[info] akka-http-core/*:crossVersion
[info] 	Binary
[info] akka-http-jackson/*:crossVersion
[info] 	Binary
[info] akka-http-marshallers-java/*:crossVersion
[info] 	Binary
[info] docs/*:crossVersion
[info] 	Binary
[info] akka-http-spray-json/*:crossVersion
[info] 	Binary
[info] akka-http-xml/*:crossVersion
[info] 	Binary
[info] akka-http-marshallers-scala/*:crossVersion
[info] 	Binary
[info] akka-http/*:crossVersion
[info] 	Binary
[info] akka-parsing/*:crossVersion
[info] 	Binary
[info] akka-http-tests/*:crossVersion
[info] 	Binary
[info] akka-http-root/*:crossVersion
[info] 	Binary
```

### DIFF
--- a/project/Publish.scala
+++ b/project/Publish.scala
@@ -49,8 +49,8 @@ object NoPublish extends AutoPlugin {
 
   override def projectSettings = Seq(
     publishArtifact := false,
-    publish := (),
-    publishLocal := ()
+    publish := {},
+    publishLocal := {}
   )
 
 }
--- a/build.sbt
+++ b/build.sbt
@@ -16,7 +16,6 @@ inThisBuild(Def.settings(
   test in assembly := {},
   licenses := Seq("Apache License 2.0" -> url("https://opensource.org/licenses/Apache-2.0")),
   scalaVersion := "2.11.8",
-  crossVersion := CrossVersion.binary,
   scalacOptions ++= Seq(
     "-deprecation",
     "-encoding", "UTF-8", // yes, this is 2 args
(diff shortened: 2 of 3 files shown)


jonas:
Failure due to #399

project/Publish.scala
```diff
@@ -49,8 +49,8 @@ object NoPublish extends AutoPlugin {
 
   override def projectSettings = Seq(
     publishArtifact := false,
-    publish := (),
-    publishLocal := ()
+    publish := {},
+    publishLocal := {}
```
ktoso: () is refered to as the "unit literal" so technically := () IMO is nicer... :-) 
I don;t mind merging as is though, thanks :)

ktoso:
Nice cleanup, thanks :)

project/Publish.scala
jonas: True, but it would require `(())` to avoid the auto-tupling warning.

ktoso: did it really? Weird hm... Thanks for fixing then :)

PR #414, opened 2016-10-20
Fix test timeouts

jonas:
With a small cleanup to make future timeouts more explicit in `LowLevelOutgoingConnectionSpec`.

### DIFF
--- a/akka-http-core/src/test/scala/akka/http/impl/engine/client/ConnectionPoolSpec.scala
+++ b/akka-http-core/src/test/scala/akka/http/impl/engine/client/ConnectionPoolSpec.scala
@@ -388,7 +388,7 @@ class ConnectionPoolSpec extends AkkaSpec("""
               case x ⇒ fail(x.toString)
             }.toMat(Sink.fold(0)(_ + _))(Keep.both).run()
 
-        Await.result(idSum, 30.seconds) shouldEqual N * (N + 1) / 2
+        Await.result(idSum, 35.seconds) shouldEqual N * (N + 1) / 2
       } catch {
         case thr: Throwable ⇒
           throw new RuntimeException(s"Failed at pipeliningLimit=$pipeliningLimit, poolFlow=$poolFlow", thr)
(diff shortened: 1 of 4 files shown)


akka-http-core/src/test/scala/akka/http/impl/engine/client/ConnectionPoolSpec.scala
```diff
@@ -388,7 +388,7 @@ class ConnectionPoolSpec extends AkkaSpec("""
               case x ⇒ fail(x.toString)
             }.toMat(Sink.fold(0)(_ + _))(Keep.both).run()
 
-        Await.result(idSum, 30.seconds) shouldEqual N * (N + 1) / 2
+        Await.result(idSum, 35.seconds) shouldEqual N * (N + 1) / 2
```
jrudolph: Hmm, we should try to speed up this test, instead. Any idea why it takes so long?

jonas: True, that spec contains several tests with long await thresholds. We saw it fail at 30.4s so we could bump it to 31. Else I would prefer to remove that commit from the PR and look at the issue in a separate PR.

I'm also wondering if it would be possible to get rid of most of the test timeout failures. In most tests I've seen the encoded timeouts are AFAIK not related with correctness, for example request/response rendering.

jrudolph: > I'm also wondering if it would be possible to get rid of most of the test timeout failures. In most tests I've seen the encoded timeouts are AFAIK not related with correctness, for example request/response rendering.

You mean that the actual length of the timeout is not related to correctness? So, that we could just the increase the values a lot because in the happy case it will succeed in a finite amount of time and it will only take a long time for the failure case?

jrudolph:
PLS BUILD

jrudolph:
LGTM.

Thanks, @jonas. I'm going to merge this as it will improve PR validation instantly.

PR #421, opened 2016-10-22
Reenable sbt-unidoc settings

jonas:
Adapts the Unidoc auto plugins from akka/akka to work in akka/akka-http with minimal changes.

### DIFF
--- a/akka-http-core/src/main/java/akka/http/javadsl/model/HttpEntity.java
+++ b/akka-http-core/src/main/java/akka/http/javadsl/model/HttpEntity.java
@@ -145,7 +145,7 @@ public interface HttpEntity {
     /**
      * Discards the entities data bytes by running the {@code dataBytes} Source contained in this entity.
      *
-     * Note: It is crucial that entities are either discarded, or consumed by running the underlying [[Source]]
+     * Note: It is crucial that entities are either discarded, or consumed by running the underlying [[akka.stream.javadsl.Source]]
      * as otherwise the lack of consuming of the data will trigger back-pressure to the underlying TCP connection
      * (as designed), however possibly leading to an idle-timeout that will close the connection, instead of
      * just having ignored the data.
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/Multipart.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/Multipart.scala
@@ -105,8 +105,17 @@ sealed trait Multipart extends jm.Multipart {
   def toStrict(timeoutMillis: Long, materializer: Materializer): CompletionStage[_ <: jm.Multipart.Strict] =
     toStrict(FiniteDuration(timeoutMillis, concurrent.duration.MILLISECONDS))(materializer).toJava
 
-  /** Java API */
-  @deprecated def toEntity(charset: jm.HttpCharset, boundary: String): jm.RequestEntity =
+  /**
+   * Java API
+   *
+   * @deprecated This variant of `toEntity` is not supported any more. The charset parameter will be ignored. Please use
+   *             one of the other overloads instead.
+   */
+  @deprecated(
+    message = "This variant of `toEntity` is not supported any more. The charset parameter will be ignored. " +
+    "Please use the variant without specifying the charset.",
+    since = "3.0.0")
+  def toEntity(charset: jm.HttpCharset, boundary: String): jm.RequestEntity =
     toEntity(boundary)
 }
 
--- a/build.sbt
+++ b/build.sbt
@@ -36,7 +36,11 @@ lazy val root = Project(
     id = "akka-http-root",
     base = file(".")
   )
-  .enablePlugins(NoPublish)
+  .enablePlugins(UnidocRoot, NoPublish)
+  .settings(
+    // Unidoc doesn't like macros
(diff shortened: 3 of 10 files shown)


jonas:
I also tried to use `LocalProject("akka-parsing")` to exclude it from the projects considered by unidoc. It allows to contain everything related to `UnicodeRoot` however it is less "safe" than using a project reference from the top-level `build.sbt` so I went with using a setting.

jonas:
Created #423

project/Doc.scala
```diff
@@ -91,3 +91,57 @@ object ScaladocNoVerificationOfDiagrams extends AutoPlugin {
...
+    // genjavadoc needs to generate synthetic methods since the java code uses them
+    scalacOptions += "-P:genjavadoc:suppressSynthetic=false",
+    // FIXME: see #18056
+    sources in(JavaUnidoc, unidoc) ~= (_.filterNot(_.getPath.contains("Access$minusControl$minusAllow$minusOrigin")))
+  )).getOrElse(Nil)
+
+  val settings = inTask(unidoc)(Seq(
+    unidocProjectFilter in ScalaUnidoc := inAnyProject -- inProjects(unidocProjectExcludes.value: _*),
+    unidocProjectFilter in JavaUnidoc := inAnyProject -- inProjects(unidocProjectExcludes.value: _*),
+    apiMappings in ScalaUnidoc := (apiMappings in (Compile, doc)).value
```
2m: I know that this is the old code, but this setting might not be needed anymore after https://github.com/sbt/sbt-unidoc/pull/10

But to be sure, we need to test, if scaladoc generated by `unidoc` has apiMappings to external libs.

akka-http-core/src/main/scala/akka/http/scaladsl/model/Multipart.scala
```diff
@@ -105,8 +105,17 @@ sealed trait Multipart extends jm.Multipart {
...
 
-  /** Java API */
-  @deprecated def toEntity(charset: jm.HttpCharset, boundary: String): jm.RequestEntity =
+  /**
+   * Java API
+   *
+   * @deprecated This variant of `toEntity` is not supported any more. The charset parameter will be ignored. Please use
+   *             one of the other overloads instead.
+   */
+  @deprecated(
```
jrudolph: Thanks for adding!

build.sbt
```diff
@@ -34,7 +34,11 @@ lazy val root = Project(
     id = "akka-http-root",
     base = file(".")
   )
-  .enablePlugins(NoPublish)
+  .enablePlugins(UnidocRoot, NoPublish)
+  .settings(
+    // Unidoc doesn't like macros
```
jrudolph: :+1:

project/Doc.scala
```diff
@@ -31,7 +31,30 @@ object Scaladoc extends AutoPlugin {
...
+        val IvyRegex = """^.*[/\\]([\.\-_\w]+)[/\\]([\.\-_\w]+)[/\\](?:jars|bundles)[/\\]([\.\-_\w]+)\.jar$""".r
+
+        for {
+          jar <- (dependencyClasspath in Compile in doc).value.toSet ++ (dependencyClasspath in Test in doc).value
+          fullyFile = jar.data
+          url <- fullyFile.getCanonicalPath match {
+            case IvyRegex(organization, name, jarBaseFile) if jarBaseFile.startsWith(s"$name-") => {
+              val version = jarBaseFile.substring(name.length + 1, jarBaseFile.length)
+              organization match {
+                case "com.typesafe.akka" =>
```
jonas: This ensures we at least have `apiMappings` for Akka core. It would be better to make a PR in akka/akka to configure apiURL so sbt will automatically add this for us.

akka-http-core/src/main/java/akka/http/javadsl/model/HttpEntity.java
```diff
@@ -145,7 +145,7 @@
     /**
      * Discards the entities data bytes by running the {@code dataBytes} Source contained in this entity.
      *
-     * Note: It is crucial that entities are either discarded, or consumed by running the underlying [[Source]]
+     * Note: It is crucial that entities are either discarded, or consumed by running the underlying [[akka.stream.javadsl.Source]]
```
ktoso: thanks 👍

akka-http-core/src/main/scala/akka/http/scaladsl/model/Multipart.scala
ktoso: good!

project/Doc.scala
ktoso: Awesome :)
Can we do the same rigth away for Scala?

ktoso: And `org.reactivestreams` => http://www.reactive-streams.org/reactive-streams-1.0.0-javadoc/

jonas: I don't see a way to add mappings to Javadoc without a lot of hacks (post-processing). Also reactive streams is not used in any public APIs.

jonas:
Squashed the fixup commit

ktoso:
Oh I see it can't cross link to javadoc, ofc.. Okey, LGTM thanks!

jonas:
Test failure is#423

ktoso:
LGTM, thanks!

Issue #429, opened 2016-10-25
Replace Akka with Akka-http in Contributing file


jlprat:
Currently the Contributing file is the same one from the main Akka project.
Instructions still refer to the Akka project instead of the akka-http project in some points. For example:

- [ ] Branches summary
- [ ] Project name under Tags
- [ ] Contributing guidelines headline
- [ ] General workflow project name to fork
- [ ] TL;DR; section under General workflow refers to fork akka project
- [ ] Pull Request project mention
- [ ] JavaDoc project mention
- [ ] Reference to akka-contrib

I know it's still understandable, but maybe it's worth fixing it.

ktoso:
Would you be able to help us out on this please?
We're trying to polish other things needed for the as-soon-as-possible release right now, thus any help would be very welcome - thanks!

jlprat:
I'm on it

ktoso:
Awesome, thanks a lot :)

jlprat:
Didn't modify the Test section, as I don't know which are the guidelines for it

ktoso:
Thanks, testing I think does not really change hm

PR #430, opened 2016-10-25
Replace Akka with Akka-http when it refers to the project name #429

jlprat:
* Replace Akka project name with akka-http project name
* Update github URLS to right project
* Removed not present branches
* Removed Contrib module as it doesn't exist in akka-http
* Updated tags to more meaningful ones in akka-http


jlprat:
I just realized, I submitted the wrong branch. I close it and create a new PR

PR #431, opened 2016-10-25
Replace Akka with Akka-http when it refers to the project name #429

jlprat:
Issue #429
* Replace Akka project name with akka-http project name
* Update github URLS to right project
* Removed not present branches
* Removed Contrib module as it doesn't exist in akka-http
* Updated tags to more meaningful ones in akka-http


ktoso:
OK TO TEST

ktoso:
Nothing to test, just muscle memory kicking in ;-)
LGTM, thanks a lot!

PR #447, opened 2016-10-31
Cross compile to Scala 2.12.0-RC2 (#437)

jonas:
Fixes two tests to work with Scala 2.12:
- SprayJsonPrettyMarshalSpec defined a trait inside a test scope, which
  resulted in a ClassFormatError:
  
  [info] docs.http.scaladsl.SprayJsonPrettyMarshalSpec **\* ABORTED ***
  [info]   java.lang.ClassFormatError: Duplicate field name&signature in class file docs/http/scaladsl/SprayJsonPrettyMarshalSpec$MyJsonService$1
  [info]   at java.lang.ClassLoader.defineClass1(Native Method)
- The fix for https://issues.scala-lang.org/browse/SI-9019 causes
  DebuggingDirectivesSpec to fail by printing `ByteString.ByteString1C`
  instead of `ByteString` in Scala 2.11.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -16,7 +16,6 @@ inThisBuild(Def.settings(
   startYear := Some(2014),
   //  test in assembly := {},
   licenses := Seq("Apache License 2.0" -> url("https://opensource.org/licenses/Apache-2.0")),
-  scalaVersion := "2.11.8",
   scalacOptions ++= Seq(
     "-deprecation",
     "-encoding", "UTF-8", // yes, this is 2 args
--- a/docs/src/main/paradox/scala/http/common/json-support.md
+++ b/docs/src/main/paradox/scala/http/common/json-support.md
@@ -22,11 +22,7 @@ To enable automatic support for (un)marshalling from and to JSON with [spray-jso
 
 Next, provide a `RootJsonFormat[T]` for your type and bring it into scope. Check out the [spray-json] documentation for more info on how to do this.
 
-Finally, mix in the `akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport` trait as shown in the example below or import the `FromEntityUnmarshaller[T]` and `ToEntityMarshaller[T]` implicits directly from `SprayJsonSupport`
-
-```scala
-import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport._
-```
+Finally, import the `FromEntityUnmarshaller[T]` and `ToEntityMarshaller[T]` implicits directly from `SprayJsonSupport` as shown in the example below or mix the `akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport` trait into your JSON support module.
 
 Once you have done this (un)marshalling between JSON and your type `T` should work nicely and transparently.
 
--- a/docs/src/test/scala/docs/http/scaladsl/SprayJsonPrettyMarshalSpec.scala
+++ b/docs/src/test/scala/docs/http/scaladsl/SprayJsonPrettyMarshalSpec.scala
@@ -3,7 +3,6 @@
  */
 package docs.http.scaladsl
 
-import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport
 import akka.http.scaladsl.server.Directives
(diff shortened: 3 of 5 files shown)


jonas:
The second workaround is a bit ugly, splitting DebuggingDirectivesSpec in a 2.11 and 2.12 flavour. A simpler solution would be to add `override def stringPrefix = "ByteString1C"` or something similar to the [definition of ByteString1C](https://github.com/akka/akka/blob/v2.4.12/akka-actor/src/main/scala/akka/util/ByteString.scala#L123)

jrudolph:
Yep, I'd vote for trying to keep the `ByteString.toString` result the same for both 2.11 and 2.12. See https://github.com/akka/akka/pull/21775/files

jonas:
Ref akka/akka#21775

jonas:
Sounds good. My understanding is also that `ByteString1C` is mainly for internal representation. Then we won't be releasing Akka HTTP before 2.4.13?

jrudolph:
No, let's not depend on such a change. Couldn't we just wrap all expected strings and replace `"ByteString.ByteString1C("` with `"ByteString("`?

docs/src/test/scala/docs/http/scaladsl/SprayJsonPrettyMarshalSpec.scala
```diff
@@ -21,6 +20,15 @@ class SprayJsonPrettyMarshalSpec extends server.RoutingSpec {
...
+    //#setup
+  }
+}
+
+class SprayJsonPrettyMarshalSpec extends server.RoutingSpec {
+
+  "spray-json example" in {
+    import SprayJsonPrettyMarshalSpec.setup._
+    //#example
     // use it wherever json (un)marshalling is needed
```
jrudolph: I'd prefer something like this:

```scala
    // use it wherever json (un)marshalling is needed
    object PrettyJsonFormatSupport1 {
      import DefaultJsonProtocol._
      implicit val printer = PrettyPrinter
      implicit val prettyPrintedItemFormat = jsonFormat2(PrettyPrintedItem)
    }

    class MyJsonService extends Directives {
      import PrettyJsonFormatSupport1._
      import SprayJsonSupport._
```

which also follows the best practice not to inherit from `DefaultJsonProtocol`.

build.sbt
```diff
@@ -17,6 +17,7 @@ inThisBuild(Def.settings(
   //  test in assembly := {},
   licenses := Seq("Apache License 2.0" -> url("https://opensource.org/licenses/Apache-2.0")),
   scalaVersion := "2.11.8",
+  crossScalaVersions := Seq("2.11.8", "2.12.0-RC2"),
```
schmitch: well I guess you can delay it till 2.12 can be used.

jonas:
PR updated to address comments. It also fixes double definition of `crossScalaVersion` and `scalaVersion` as well as the JSON support docs.

docs/src/main/paradox/scala/http/common/json-support.md
```diff
@@ -22,11 +22,7 @@ To enable automatic support for (un)marshalling from and to JSON with [spray-jso
 
 Next, provide a `RootJsonFormat[T]` for your type and bring it into scope. Check out the [spray-json] documentation for more info on how to do this.
 
-Finally, mix in the `akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport` trait as shown in the example below or import the `FromEntityUnmarshaller[T]` and `ToEntityMarshaller[T]` implicits directly from `SprayJsonSupport`
-
-```scala
-import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport._
-```
+Finally, import the `FromEntityUnmarshaller[T]` and `ToEntityMarshaller[T]` implicits directly from `SprayJsonSupport` as shown in the example below or mix the `akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport` trait into your JSON support module.
```
jrudolph: Thanks, for fixing the docs as well.

jrudolph:
Great, thanks, @jonas. Seems we are already there for 2.12 support.

PR #470, opened 2016-11-03
Prepare for Scala 2.12 cross building

jrudolph:


ktoso:
LGTM :)

project/Dependencies.scala
```diff
@@ -6,7 +6,7 @@ import sbt.Keys._
 object Dependencies {
   import DependencyHelpers._
 
-  val akkaVersion = "2.4.11"
+  val akkaVersion = "2.4.12"
```
ktoso: Yes please, let's keep a nice stream of PRs running/merging :)

akka-http-core/src/main/scala/akka/http/impl/util/StageLoggingWithOverride.scala
```diff
@@ -11,11 +11,13 @@ import akka.event.LoggingAdapter
 import akka.stream.ActorMaterializer
 import akka.event.NoLogging
 
-// TODO this can be removed when https://github.com/akka/akka/issues/18793 has been implemented
+// TODO Try to reconcile with what Akka provides in StageLogging.
+// We thought this could be removed when https://github.com/akka/akka/issues/18793 had been implemented
+// but we need a few more changes to be able to override the default logger. So for now we keep it here.
```
jrudolph: It's that `logOverride` below. I introduced it to allow to specify a logger from the outside and take the one from the GraphStage otherwise.

PR #473, opened 2016-11-03
=htc #368 document the guarantees of `max-connections`

### LINKED ISSUE #368: Requests with closed upstreams are not handled with `max-connections`

jrudolph:
Actually providing better guarantees turned out to be difficult, so
being clear about the limit being an approximate one seems to be the
best solution for now.

See also the discussions in https://github.com/akka/akka-http/pull/371.

Fixes #368.

@2m WDYT?

### DIFF
--- a/akka-http-core/src/main/resources/reference.conf
+++ b/akka-http-core/src/main/resources/reference.conf
@@ -38,6 +38,13 @@ akka.http {
     #
     # This setting doesn't apply to the `Http().bind` method which will still
     # deliver an unlimited backpressured stream of incoming connections.
+    #
+    # Note, that this setting limits the number of the connections on a best-effort basis.
+    # It does *not* strictly guarantee that the number of established TCP connections will never
+    # exceed the limit (but it will be approximately correct) because connection termination happens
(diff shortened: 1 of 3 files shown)


2m:
This paragraph is great.

Should be move `watchTermination` below user handler as well? That improved the behavior at least when testing with gatling.

jrudolph:
@2m I agree that we can try to improve the situation. So, I added another commit so that we will now have two `watchTermination`s around the user handler and wait for both futures to be completed. This still doesn't solve it perfectly (so the added comment in the Scaladoc is still valid) but it hopefully improves the situation for gatling.

WDYT?

ktoso:
LGTM, good docs and the additional watch. Certainly an improvement I think

2m:
LGTM!

PR #477, opened 2016-11-03
New authors script

johanandren:


scripts/authors.scala
```diff
@@ -0,0 +1,69 @@
...
+!#
+
+/*
+ * Usage:
+ *    authors.scala tag1 tag2
+ *
+ * or if on non unixy os:
+ *    scala authors.scala tag1 tag2
+ *
+ * requires scala 2.11.x+ and command line git on path
```
jrudolph: If it's in Scala, couldn't we just include it in sbt?

johanandren: Yeah, I guess, what would we gain from it though (unless we make something grander for releasing where it is a part)?

jrudolph: I thought it might be simpler to compile if there's no scala on the command line but on the other hand, if you don't need it a lot it might make sense to just keep it here.

johanandren: Oh, I didn't even think of that since I always have. Good point. Hmm.

ktoso: Do you want to make that change in this PR @johanandren ? :)
Sounds like a good idea 👍

johanandren: nah, lets merge this and I can do that later, given time, this is useful until then

ktoso: Okey 👍

ktoso:
Thanks @johanandren  :)

Issue #416, opened 2016-10-20
Response entity failures kill client pool


jrudolph:
It was really hard to find out how problems with a response are fed back into the pool so that at the end the complete pool is shutdown.

I changed GraphStageLogic.failStage to wrap all exceptions with more information to get at the trace of how the error propagated like this:

``` scala
  final def failStage(ex: Throwable): Unit = {
    def stageName: String = try {
      val outerField = getClass.getDeclaredField("$outer")
      outerField.setAccessible(true)
      outerField.get(this) match {
        case MapAsyncUnordered(_, func) ⇒ s"MapAsyncUnordered(${func.getClass})"
        case Map(func)                  ⇒ s"Map(${func.getClass})"
        case Collect(func)              ⇒ s"Collect(${func.getClass})"
        case _                          ⇒ getClass.toString
      }
    } catch {
      case x ⇒ getClass.toString
    }

    val newEx = new RuntimeException(s"Failing stage in $stageName", ex).fillInStackTrace()
    // use newEx instead of ex
  }
```

You then get a stack trace like this (stripped the actual code lines) if while receiving a response entity the connection breaks:

```
java.lang.RuntimeException: Failing stage in class akka.http.impl.engine.client.PoolSlot$SlotProcessor$$anon$3
Caused by: java.lang.RuntimeException: Failing stage in class akka.stream.impl.fusing.Log$$anon$23
Caused by: java.lang.RuntimeException: Failing stage in class akka.http.impl.engine.client.PoolConductor$Route$$anon$4
Caused by: java.lang.RuntimeException: Failing stage in class akka.http.impl.engine.client.PoolConductor$SlotSelector$$anon$1
Caused by: java.lang.RuntimeException: Failing stage in MapAsyncUnordered(class akka.http.impl.engine.client.PoolConductor$$anonfun$apply$2$$anonfun$4)
Caused by: java.lang.RuntimeException: Failing stage in class akka.http.scaladsl.model.HttpEntity$Limitable$$anon$1
Caused by: java.lang.RuntimeException: Failing stage in Collect(class akka.http.impl.engine.parsing.HttpMessageParser$$anonfun$defaultEntity$1$$anonfun$1)
Caused by: java.lang.RuntimeException: Failing stage in class akka.stream.impl.fusing.SubSource$$anon$4
Caused by: java.lang.RuntimeException: Failing stage in class akka.stream.impl.fusing.StatefulMapConcat$$anon$25
[...]

PR #481, opened 2016-11-04
=htc #416 fix for connection pool crash

gosubpl:
This is to fix connection pool crashing entirely in case of response entity future failing.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolConductor.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolConductor.scala
@@ -117,31 +117,31 @@ private object PoolConductor {
   private class SlotSelector(slotSettings: PoolSlotsSetting, pipeliningLimit: Int, log: LoggingAdapter)
     extends GraphStage[FanInShape2[RequestContext, SlotEvent, SwitchSlotCommand]] {
 
-    private val ctxIn = Inlet[RequestContext]("requestContext")
-    private val slotIn = Inlet[SlotEvent]("slotEvents")
-    private val out = Outlet[SwitchSlotCommand]("slotCommand")
+    private val requestContextsIn = Inlet[RequestContext]("requestContexts.In")
+    private val slotEventsIn = Inlet[SlotEvent]("slotEvents.In")
+    private val slotCommandsOut = Outlet[SwitchSlotCommand]("slotCommands.Out")
 
     override def initialAttributes = Attributes.name("SlotSelector")
 
-    override val shape = new FanInShape2(ctxIn, slotIn, out)
+    override val shape = new FanInShape2(requestContextsIn, slotEventsIn, slotCommandsOut)
 
     override def createLogic(effectiveAttributes: Attributes) = new GraphStageLogic(shape) {
       val slotStates = Array.fill[SlotState](slotSettings.maxSlots)(Unconnected)
       var nextSlot = 0
 
-      setHandler(ctxIn, new InHandler {
+      setHandler(requestContextsIn, new InHandler {
         override def onPush(): Unit = {
-          val ctx = grab(ctxIn)
+          val ctx = grab(requestContextsIn)
           val slot = nextSlot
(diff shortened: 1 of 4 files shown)


gosubpl:
Refs #416

gosubpl:
**Please do not merge before reading this note**
As per the discussion on the issue ticket (https://github.com/akka/akka-http/issues/416) I have implemented only a simple fix that results in that in most cases the state will end up in `Idle` instead of more proper `Disconnect`. While functionally this is not a problem, the connection slot scheduler prefers `Idle` slots to `Disconnected` ones as they are supposedly hot.
I have submitted this PR to not block any progress on `FIRST_STABLE` release.

- If the release train needs to depart, this PR can be merged and I will register an issue to fix the terminal state of the connection pool, as suggested by @leachbj (that can be solved with doing the below)
- If the release will not happen anytime soon, please let me know and I will extend connection pool state machine logic with additional processing based on idea inspired by `SlotEvent.NoOp` (https://github.com/leachbj/akka-http/commit/516f997b94fb980a1a4b314ed3bd4c5b49a9597d) to get proper handling for the terminal state of the pool in this case.

@jrudolph @leachbj many thanks for the insightful discussion leading to this PR. Also @jrudolph many thanks for your reproducer with tracing - investigating that issue with this was a breeze (relatively speaking :) )!

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolConductor.scala
```diff
@@ -150,28 +150,28 @@ private object PoolConductor {
...
 
         // eagerly start at least slotSettings.minSlots connections
         (0 until slotSettings.minSlots).foreach { connect }
       }
 
       def connect(slotIx: Int): Unit = {
-        emit(out, SwitchSlotCommand(ConnectEagerlyCommand, slotIx))
+        emit(slotCommandsOut, SwitchSlotCommand(ConnectEagerlyCommand, slotIx))
         slotStates(slotIx) = Idle
       }
```
leachbj: I would recommend leaving these changes out as they don't contribute to the fix (they're good changes but unrelated to the fix).

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolSlot.scala
```diff
@@ -112,7 +113,24 @@ private object PoolSlot {
...
+          //    SlotEvent.RequestCompleted(slot); transitions from Loaded(1) => Idle
+          // 2. Fail after the response headers (Future failed and then onUpstreamFailure);
+          //    SlotEvent.RequestCompleted(slot); transitions from Loaded(1) => Idle
+          //    SlotEvent.Disconnected(slot, 0); transitions from Idle => Unconnected
+          // the result is that the final state of the slot is usually Idle which results in this slot interpreted as hot
+          // and being preferred by the scheduler instead of more correct Unconnected
+          // FIXME: enhance the event & processing in the connection pool state machine to allow proper transition
+          val completed = whenCompleted.map(_ ⇒ SlotEvent.RequestCompleted(slotIx))
+            .recoverWith { case _ ⇒ FastFuture.successful(SlotEvent.RequestCompleted(slotIx)) }
+          push(eventsOut, SlotEvent.RequestCompletedFuture(completed))
```
leachbj: 👍

ktoso:
Disabled the troublesome node (a7).

ktoso:
PLS BUILD

gosubpl:
@leachbj many thanks for reviewing. I have moved the `PoolConductor` logging improvements to a separate ticket (https://github.com/akka/akka-http/issues/482 ).

jrudolph:
I think the fix is more important than perfect sync as long as we keep in mind that we might want to improve the situation (or that we might have missed cases). I created #490 to track this and also added the link to the FIXME. No waiting for PR validation and then I'll merge it.

jrudolph:
Thanks, @gosubpl and @leachbj for great discussion and help with this one.

PR #495, opened 2016-11-07
#21172 module as part of http

hveiga:
Hi @ktoso @patriknw @drewhk here is the PR with all the requested changes. Please let me know if you need anything else and I can add it to this PR.


ktoso:
OK TO TEST

akka-http-cluster-management/src/main/resources/reference.conf
```diff
@@ -0,0 +1,4 @@
+akka.http.cluster.management {
```
ktoso: ok, good namespace

akka-http-cluster-management/src/main/scala/akka/cluster/http/management/ClusterHttpApiSettings.scala
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.cluster.http.management
+
+import com.typesafe.config.Config
+
+final class ClusterHttpManagementSettings(val config: Config) {
+  private val cc = config.getConfig("akka.http.cluster.management")
+  val clusterHttpManagementPort = cc.getInt("port")
```
ktoso: uppercase perhaps? `ClusterHttpManagementPort`?

akka-http-cluster-management/src/main/scala/akka/cluster/http/management/ClusterHttpManagement.scala
```diff
@@ -0,0 +1,236 @@
...
+        routesMember(cluster)
+    }
+
+  def apply(cluster: Cluster, asyncAuthenticator: AsyncAuthenticator[String]): Route = {
+    authenticateBasicAsync[String](realm = "secured", asyncAuthenticator) { _ ⇒
+      apply(cluster)
+    }
+  }
+
+  def apply(cluster: Cluster, pathPrefixName: String, asyncAuthenticator: AsyncAuthenticator[String]): Route = {
```
ktoso: needs scaladoc that explains what this will provide

akka-http-cluster-management/src/main/scala/akka/cluster/http/management/ClusterHttpManagement.scala
```diff
@@ -0,0 +1,236 @@
...
+        case Some(member) ⇒
+          routeGetMember(cluster, member) ~ routeDeleteMember(cluster, member) ~ routePutMember(cluster, member)
+        case None ⇒
+          complete(StatusCodes.NotFound → ClusterHttpManagementMessage(s"Member [$memberAddress] not found"))
+      }
+    }
+
+  def apply(cluster: Cluster): Route = apply(cluster, "members")
+
+  def apply(cluster: Cluster, pathPrefixName: String): Route =
```
ktoso: (all of these need scaladoc)

akka-http-cluster-management/src/main/scala/akka/cluster/http/management/ClusterHttpManagement.scala
```diff
@@ -0,0 +1,236 @@
...
+  def apply(cluster: Cluster, pathPrefix: String, https: ConnectionContext) =
+    new ClusterHttpManagement(cluster, Some(pathPrefix), None, Some(https))
+
+  def apply(cluster: Cluster, asyncAuthenticator: AsyncAuthenticator[String], https: ConnectionContext): ClusterHttpManagement =
+    new ClusterHttpManagement(cluster, None, Some(asyncAuthenticator), Some(https))
+
+  def apply(cluster: Cluster, pathPrefix: String, asyncAuthenticator: AsyncAuthenticator[String], https: ConnectionContext): ClusterHttpManagement =
+    new ClusterHttpManagement(cluster, Some(pathPrefix), Some(asyncAuthenticator), Some(https))
+
+  def create(cluster: Cluster): ClusterHttpManagement =
```
ktoso: good, thanks for all java api - it could just delegate to it's sibling apply I think

akka-http-cluster-management/src/main/scala/akka/cluster/http/management/ClusterHttpManagement.scala
```diff
@@ -0,0 +1,236 @@
...
+  https:              Option[ConnectionContext]          = None) {
+
+  private val settings = new ClusterHttpManagementSettings(cluster.system.settings.config)
+  private implicit val system = cluster.system
+  private implicit val materializer = ActorMaterializer()
+  import system.dispatcher
+
+  private var bindingFuture: Future[Http.ServerBinding] = _
+
+  def start(): Unit = {
```
ktoso: this should be safe from people calling it accidentally multiple times.
I'd suggest making `bindingFuture` an `AtomicReference` and bail if server already started.
Perhaps a return type that includes info on what address we started would be useful here too.

akka-http-cluster-management/src/main/scala/akka/cluster/http/management/ClusterHttpManagement.scala
```diff
@@ -0,0 +1,236 @@
...
+      case None ⇒
+        bindingFuture = Http().bindAndHandle(
+          routes,
+          settings.clusterHttpManagementHostname,
+          settings.clusterHttpManagementPort)
+    }
+  }
+
+  def stop() = {
+    bindingFuture.flatMap(_.unbind())
```
ktoso: and here we'd swap bad the atomic `bindingFuture` back to empty

akka-http-cluster-management/src/test/java/akka/cluster/http/management/ClusterHttpManagementJavaCompileTest.java
```diff
@@ -0,0 +1,14 @@
+package akka.cluster.http.management;
+
+import akka.actor.ActorSystem;
+import akka.cluster.Cluster;
+
+public class ClusterHttpManagementJavaCompileTest {
```
ktoso: I think this will need an empty test method or Junit will freak out with "no test method found"
Just add `@Test public void compileOnly() {}`

project/Dependencies.scala
```diff
@@ -155,6 +158,8 @@ object Dependencies {
 
   lazy val docs = l ++= Seq(Docs.sprayJson, Docs.gson)
 
+  lazy val httpClusterManagement = l ++= Seq(akkaCluster, Test.mockito)
```
ktoso: yeap, structure is good like that 👍

build.sbt
```diff
@@ -101,6 +101,12 @@ lazy val httpMarshallersJava = project("akka-http-marshallers-java")
 lazy val httpJackson =
   httpMarshallersJavaSubproject("jackson")
 
+lazy val httpClusterManagement = project("akka-http-cluster-management")
```

ktoso: cc @patriknw @jrudolph how do you think we should mark this as experimental or somehow "fresh"...?

ktoso:
Will review the spec and routes in more detail soon :)
All in all very excited about this.

PS: I'm in US timezone now for a week, so likely swifter review while you're also not asleep @hveiga 👍

hveiga:
Cool :) Thanks for taking the time reviewing this @ktoso . I have fixed few of the things you talked about. I implemented the `AtomicReference` thing but I am not completely sure if that is what you had in mind. Let me know if there is anything else! Thanks again.

akka-http-cluster-management/src/main/scala/akka/cluster/http/management/ClusterHttpManagement.scala
```diff
@@ -0,0 +1,366 @@
...
+import akka.http.scaladsl.server.Directives._
+import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport
+import akka.http.scaladsl.model.StatusCodes
+import akka.http.scaladsl.server.{ Route, RouteResult }
+import akka.stream.ActorMaterializer
+import spray.json._
+
+import scala.concurrent.Future
+
+final case class ClusterUnreachableMember(node: String, observedBy: Seq[String])
```
patriknw: can this `Seq` be an `immutable.Seq`?

akka-http-cluster-management/src/main/scala/akka/cluster/http/management/ClusterHttpManagement.scala
```diff
@@ -0,0 +1,366 @@
...
+final case class ClusterMember(node: String, nodeUid: String, status: String, roles: Set[String])
+final case class ClusterMembers(selfNode: String, members: Set[ClusterMember], unreachable: Seq[ClusterUnreachableMember])
+final case class ClusterHttpManagementMessage(message: String)
+final case class ClusterHttpManagementServerInfo(hostname: String, port: Int, alreadyStarted: Boolean)
+
+sealed trait ClusterHttpManagementOperation
+case object Down extends ClusterHttpManagementOperation
+case object Leave extends ClusterHttpManagementOperation
+case object Join extends ClusterHttpManagementOperation
```
patriknw: can we make more things `private[akka]` so that we can change these without having to worry about binary compatibility?

akka-http-cluster-management/src/main/scala/akka/cluster/http/management/ClusterHttpManagement.scala
```diff
@@ -0,0 +1,366 @@
...
+    bindingFuture.set(serverFutureBinding)
+
+    ClusterHttpManagementServerInfo(
+      settings.ClusterHttpManagementHostname,
+      settings.ClusterHttpManagementPort,
+      false
+    )
+  }
+
+  def stop() = {
```
patriknw: explicit return type, `: Unit`

patriknw: or `Future[Done]`

akka-http-cluster-management/src/main/scala/akka/cluster/http/management/ClusterHttpManagement.scala
```diff
@@ -0,0 +1,366 @@
...
+
+  private val settings = new ClusterHttpManagementSettings(cluster.system.settings.config)
+  private implicit val system = cluster.system
+  private implicit val materializer = ActorMaterializer()
+  import system.dispatcher
+
+  private val bindingFuture = new AtomicReference[Future[Http.ServerBinding]]()
+
+  def start(): ClusterHttpManagementServerInfo = {
+    if (bindingFuture.get() != null) {
```
patriknw: use compare and swap of the `AtomicReference` to start it only once, something like:

```scala
val serverBindingPromise = Promise[Http.ServerBinding]()
if (bindingFuture.compareAndSet(null, serverBindingPromise.future)) {
  ...
  serverBindingPromise.completeWith(serverFutureBinding)
}  
```

but I'm not sure it makes sense to support `stop` followed by `start` again. Then things will become complicated, because you can't really start until the `unbind` is completed, and I see no use case for that.

akka-http-cluster-management/src/main/scala/akka/cluster/http/management/ClusterHttpManagement.scala
```diff
@@ -0,0 +1,366 @@
...
+  https:              Option[ConnectionContext]          = None) {
+
+  private val settings = new ClusterHttpManagementSettings(cluster.system.settings.config)
+  private implicit val system = cluster.system
+  private implicit val materializer = ActorMaterializer()
+  import system.dispatcher
+
+  private val bindingFuture = new AtomicReference[Future[Http.ServerBinding]]()
+
+  def start(): ClusterHttpManagementServerInfo = {
```
patriknw: I don't see the purpose of this `ClusterHttpManagementServerInfo`. I think a `Future[Done]` would be enough

jrudolph:
Hmm, I missed the discussion at akka/akka#21772. I see that it might be simpler to maintain this module here in the short-term. Content-wise it doesn't seem to belong here. It is not an extension to akka-http but just an application of akka-http. I see a few problems with this:

 * It becomes part of the wrong release cycle. We will be forced to make a new akka-http release when akka cluster things change.
 * The module name should more strongly emphasize that this is an extension to akka-cluster and not to akka-http. Maybe "akka-cluster-management-http"?
 * Where would the documentation be published to?
 * People who subscribed to akka-http will also get notified about events for akka-http-cluster.
 * There's a potential future extension to akka-http to distribute HTTP APIs or services in a cluster which people could confuse with this module.
 * The cost of maintaining this module (new PR validation failures, additional tests, etc) are accrued at the wrong place.

From experience we know that there's a huge cost to clean up later when a module is maintained at the wrong place. Having just carved out a clean space for akka-http here, I would suggest to find another solution for this module.

Wouldn't it be the easier to just create a new repository for things like this?

jrudolph:
Great work, @hveiga btw :)

patriknw:
Good points, @jrudolph. We discussed this in the team back and forth and it's not easy to foresee future similar contributions, but we think this is important enough to motivate a separate repository.  I have created https://github.com/akka/akka-cluster-management

We might rename the repository if more similar things need a home in the future.

Sorry for the inconvenience of directing your to new places all the time, @hveiga 
We appreciate your work and initiative around this very much.

@2m Can you please fill that repository with the minimal build stuff, and then @hveiga can open a new PR there.

ktoso:
Yep, agreed with @jrudolph and @patriknw - just had a call on this.
Terribly sorry about the dance around the repositories, we really wanted to give it it's proper place @hveiga :)

Getting it into the new repo should be very smooth though!
Thanks again

hveiga:
New repo sounds like the best idea here. I will also do the few changes @patriknw suggested and create a new PR once https://github.com/akka/akka-cluster-management is ready. I will check in a few days.

2m:
@hveiga a PR for initial repo structure: https://github.com/akka/akka-cluster-management/pull/1

ktoso:
The new repo is ready, please PR there: github.com/akka/akka-cluster-management @hveiga 
Thanks a lot and sorry for all the jumping around between repositories :)

hveiga:
Thanks @2m and @ktoso . I will do the latest changes we talked about and create a new PR. Hopefully before the end of the week :) No worries about the back and forth!

PR #506, opened 2016-11-10
Initial steps for HTTP/2 incoming Entity support

### LINKED ISSUE #384: HTTP2: read requests with entities
Currently, only requests without an entity are supported. We should support also reading DATA frames for reading entity data.

ktoso:
Initial step to resolve #384 

I'll continue to work on this during the next days I think.
@jrudolph's suggestion to put stuff into the Demuxer was very good!

Review as entire thing, not individual commits - I'll squash later.

### DIFF
--- a/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/BufferedOutletSupport.scala
+++ b/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/BufferedOutletSupport.scala
@@ -37,17 +37,21 @@ trait BufferedOutletSupport { logic: GraphStageLogic ⇒
      */
     protected def doPush(elem: T): Unit = outlet.push(elem)
 
-    override def onPull(): Unit =
-      if (!buffer.isEmpty) doPush(buffer.pop())
-
+    override def onPull(): Unit = {
+      if (!buffer.isEmpty) {
+        val popped = buffer.pop()
+        doPush(popped)
+      }
+    }
     outlet.setHandler(this)
 
     def push(elem: T): Unit =
       if (outlet.canBePushed && buffer.isEmpty) doPush(elem)
       else buffer.addLast(elem)
 
-    def tryFlush(): Unit =
+    def tryFlush(): Unit = {
       if (outlet.canBePushed && !buffer.isEmpty)
         doPush(buffer.pop())
+    }
   }
 }
--- a/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/FrameParser.scala
+++ b/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/FrameParser.scala
@@ -72,7 +72,6 @@ class FrameParser(shouldReadPreface: Boolean) extends ByteStringParser[FrameEven
       case DATA ⇒
         val pad = Flags.PADDED.isSet(flags)
         val endStream = Flags.END_STREAM.isSet(flags)
-        println("parse endStream = " + endStream)
 
         val paddingLength =
           if (pad) payload.readByte() & 0xff
--- a/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/FrameParser.scala
+++ b/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/FrameParser.scala
@@ -23,7 +23,7 @@ class FrameParser(shouldReadPreface: Boolean) extends ByteStringParser[FrameEven
       }
 
       object ReadPreface extends Step {
-        def parse(reader: ByteReader): ParseResult[FrameEvent] =
+        override def parse(reader: ByteReader): ParseResult[FrameEvent] =
           if (reader.remainingSize < 24) throw NeedMoreData
           else if (reader.take(24) == Http2Protocol.ClientConnectionPreface)
             ParseResult(None, ReadFrame, acceptUpstreamFinish = false)
@@ -32,7 +32,7 @@ class FrameParser(shouldReadPreface: Boolean) extends ByteStringParser[FrameEven
       }
 
       object ReadFrame extends Step {
-        def parse(reader: ByteReader): ParseResult[FrameEvent] = {
+        override def parse(reader: ByteReader): ParseResult[FrameEvent] = {
           val length = reader.readShortBE() << 8 | reader.readByte()
           val tpe = reader.readByte() // TODO: make sure it's valid
           val flags = new ByteFlag(reader.readByte())
@@ -72,6 +72,7 @@ class FrameParser(shouldReadPreface: Boolean) extends ByteStringParser[FrameEven
       case DATA ⇒
         val pad = Flags.PADDED.isSet(flags)
         val endStream = Flags.END_STREAM.isSet(flags)
+        println("parse endStream = " + endStream)
 
         val paddingLength =
           if (pad) payload.readByte() & 0xff
--- a/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
+++ b/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
@@ -5,6 +5,7 @@
 package akka.http.impl.engine.http2
 
 import akka.NotUsed
+import akka.event.Logging
 import akka.http.impl.engine.http2.parsing.HttpRequestHeaderHpackDecompression
 import akka.http.impl.engine.http2.rendering.HttpResponseHeaderHpackCompression
 import akka.http.scaladsl.model.HttpRequest
@@ -31,10 +32,11 @@ private[http2] final case class Http2SubStream(initialFrame: HeadersFrame, frame
 
 object Http2Blueprint {
   // format: OFF
-  def serverStack(): BidiFlow[HttpResponse, ByteString, ByteString, HttpRequest, NotUsed] =
+  def serverStack(): BidiFlow[HttpResponse, ByteString, ByteString, HttpRequest, NotUsed] = {
     httpLayer() atop
     demux() atop
     framing()
+  }
   // format: ON
 
   def framing(): BidiFlow[FrameEvent, ByteString, ByteString, FrameEvent, NotUsed] =
@@ -59,7 +61,7 @@ object Http2Blueprint {
    */
   def httpLayer(): BidiFlow[HttpResponse, Http2SubStream, Http2SubStream, HttpRequest, NotUsed] = {
     val incomingRequests = Flow[Http2SubStream].via(new HttpRequestHeaderHpackDecompression)
-      .log(getClass.getSimpleName)
+      .log(Logging.simpleName(getClass)) //FIXME replace with a proper `atop logging()`
     val outgoingResponses = Flow[HttpResponse].via(new HttpResponseHeaderHpackCompression)
     BidiFlow.fromFlows(outgoingResponses, incomingRequests)
   }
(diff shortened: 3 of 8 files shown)


ktoso:
Wow, build just hang for 7 hours... :O https://jenkins.akka.io:8498/job/pr-validator-akka-http/259/console
restarting...

I don't think it's related to my changes, locally things pass.

ktoso:
PLS BUILD

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -74,9 +75,12 @@ class Http2ServerDemux extends GraphStage[BidiShape[Http2SubStream, FrameEvent,
...
 
   def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
-    new GraphStageLogic(shape) with BufferedOutletSupport { logic ⇒
-      case class SubStream(
+    new GraphStageLogic(shape) with BufferedOutletSupport {
+      logic ⇒
+
+      final case class SubStream(
         streamId:                  Int,
+        headers:                   HeadersFrame,
```
jrudolph: We might not need to keep it here. There can only be one HEADERS frame ongoing at any point. So, at the point we start reading a `HEADERS` frame with `endHeaders = false` we can just move into a special state where we accept only `CONTINUATION` frames and err-out otherwise.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -24,44 +24,45 @@ import akka.stream.stage.OutHandler
...
  *
  * (An alternative API would just push a BidiHttp2SubStream(subStreamFlow: Flow[StreamFrameEvent, StreamFrameEvent])
- *  similarly to IncomingConnection. This would more accurately model the one-to-one relation between incoming and
- *  outgoing Http2Substream directions but wouldn't stack so nicely with other BidiFlows.)
+ * similarly to IncomingConnection. This would more accurately model the one-to-one relation between incoming and
+ * outgoing Http2Substream directions but wouldn't stack so nicely with other BidiFlows.)
  *
  * Backpressure logic:
  *
- *  * read all incoming frames without applying backpressure
```
jrudolph: Somehow the indentation got lost here?

ktoso: Oh weird... perhaps auto format in intellij broke things :~

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
ktoso: I'll try out that idea 👍

akka-http2-support/src/main/scala/akka/http/scaladsl/model/http2/Http2StreamIdHeader.scala
```diff
@@ -6,6 +6,7 @@ package akka.http.scaladsl.model.http2
 
 import akka.http.impl.engine.ws.InternalCustomHeader
 
+// TODO should the streamId not be a Long? I may be mis-remembering the spec
```
gosubpl: I thought so too... basically I think we are having to do with Unsigned 4 octet long numbers. These don't have good representation in Scala world... as the '>>' vs. '>>>' shows. This goes for all 32-bit stuff extracted from the frames, so not only streamId.

schmitch: according to the spec:

> Stream Identifier:  A stream identifier (see Section 5.1.1) expressed as an unsigned 31-bit integer

31-bit is really akward, and since it's so small it's actually possible to exhaust it.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -204,7 +223,33 @@ class Http2ServerDemux extends GraphStage[BidiShape[Http2SubStream, FrameEvent,
...
+        moreCompleteHandler
+      }
+
+      def makeHttp2SubStream(handler: SubStream): Http2SubStream = {
+        val headers = handler.headers
+        val subStream = handler.outlet
+        if (headers.endStream && headers.endHeaders) {
+          Http2SubStream(headers, Source.empty)
+        } else {
+          // FIXME a bit naive but correct I think -- todo check the spec
```
schmitch: well that's not enterely true since one *can* actually send a header/continuation at the end (after the last data frame). This actually replaces the chunked response, that send trailers at the end.

akka-http2-support/src/main/scala/akka/http/scaladsl/model/http2/Http2StreamIdHeader.scala
ktoso: If in doubt we could introduce our value class representing these numbers.

jrudolph: I think they took 31 bits by purpose because that still fits into a signed 32 bit integer.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
jrudolph: @schmitch, yes we just ignored the possibility of trailing headers so far. I'd suggest not to implement anything for them yet. We might add some code that actively ignores them.

Not sure what you mean with the push stream examples. Where did you get them from?

As I understand, potential sequences can only be like this:

PUSH_STREAM -> (CONTINUATION) -> HEADERS -> (CONTINUATION) -> DATA -> (trailing headers)

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/BufferedOutletSupport.scala
```diff
@@ -37,17 +37,21 @@ trait BufferedOutletSupport { logic: GraphStageLogic ⇒
      */
     protected def doPush(elem: T): Unit = outlet.push(elem)
 
-    override def onPull(): Unit =
-      if (!buffer.isEmpty) doPush(buffer.pop())
-
+    override def onPull(): Unit = {
```
jrudolph: This file seems basically unchanged.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/FrameParser.scala
```diff
@@ -72,6 +72,7 @@ class FrameParser(shouldReadPreface: Boolean) extends ByteStringParser[FrameEven
       case DATA ⇒
         val pad = Flags.PADDED.isSet(flags)
         val endStream = Flags.END_STREAM.isSet(flags)
+        println("parse endStream = " + endStream)
```
jrudolph: println

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
```diff
@@ -51,6 +59,7 @@ object Http2Blueprint {
    */
   def httpLayer(): BidiFlow[HttpResponse, Http2SubStream, Http2SubStream, HttpRequest, NotUsed] = {
     val incomingRequests = Flow[Http2SubStream].via(new HttpRequestHeaderHpackDecompression)
+      .log(getClass.getSimpleName)
```
jrudolph: Do we want to keep this?

viktorklang: No matter what, getSimpleName is not safe for Scala classes, hence Logging.simpleName

ktoso: I think we do want to keep it, but as Logging.simpleName, yeah

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
schmitch: Actually that potential sequence happens on the related stream not on the *actual* stream.
I actually made a mistake by calling it PUSH_STREAM, it's in fact `PUSH_PROMISE`.

Basically the PUSH_PROMISE can be sent after the data frame, basically I guess what you described is more what happens on the client them however the PUSH_PROMISE from the server is sent with the stream id of the connection anywhere between the headers and the end_stream and promises a stream id and then it will sequence it like you described.

jrudolph: I see. I wonder why they did it that way and not have the PUSH_PROMISE start a new stream, putting the referenced stream id in the extra field. 

But I don't think it really matters for us currently. We only have server side support and don't support push_promise so far.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
ktoso: Hm, right on the entire blueprint would be better perhaps hm...
Reason to have this is that otherwise errors (onError) literally get swallowed making debugging harder.

jrudolph: Yes, we should make sure that errors are not swallowed but are surfaced in user level streams. Maybe add a pending test case stub for it?

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
ktoso: Would you be able to re-word this into a ticket @jrudolph or @schmitch ?
I also think that we don't care _right now_ but would be good to collect such things as tickets.

ktoso:
This is now mostly cleaned up, can I merge? There's some FIXME's left, to be addressed as we go.

ktoso:
PLS BUILD

ktoso:
Build was stuck, weird.

ktoso:
PLS BUILD

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -445,6 +472,11 @@ class Http2ServerSpec extends AkkaSpec with WithInPendingUntilFixed {
       encode(":path", request.uri.path.toString)
       encode(":authority", request.uri.authority.toString.drop(2) /* Authority.toString prefixes two slashes */ )
 
+      encode("content-type", request.entity.contentType.render(new StringRendering).get)
```
jrudolph: :+1:

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
jrudolph: Hmm the comment seems still broken.

ktoso: Ok, copy pasted current state (from master).

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
jrudolph: Added #530, and #531 for (ignoring) trailing header support. On the server side it doesn't seem necessary to support receiving PUSH_PROMISE frames.

ktoso: Great, thanks! (please mark such as t:http2 / triaged)

ktoso:
Merged this step!

jrudolph:
:+1:

PR #508, opened 2016-11-10
=htc add LogByteStringTools to impl.util to have it ready during debugging

jrudolph:
I use this all the time. It's not http specific but it's very useful here.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/util/LogByteStringTools.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/util/LogByteStringTools.scala
@@ -1,0 +1,74 @@
+package akka.http.impl.util
+
+import akka.NotUsed
+import akka.stream.TLSProtocol._
+import akka.stream.scaladsl
+import akka.stream.scaladsl.{ BidiFlow, Flow }
+import akka.util.ByteString
+
+/**
+ * INTERNAL API
+ *
+ * Flow and BidiFlow stages to log streams of ByteString.
+ */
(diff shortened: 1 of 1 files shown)


jrudolph:
Next, this could be used to implement #509.

akka-http-core/src/main/scala/akka/http/impl/util/LogByteStringTools.scala
```diff
@@ -0,0 +1,74 @@
+package akka.http.impl.util
```
ktoso: make sure to have a license header configured ;)

akka-http-core/src/main/scala/akka/http/impl/util/LogByteStringTools.scala
```diff
@@ -0,0 +1,74 @@
...
+      case n: NegotiateNewSession ⇒ n.toString
+    })
+
+  def logTlsInbound(name: String, maxBytes: Int = MaxBytesPrinted): Flow[SslTlsInbound, SslTlsInbound, NotUsed] =
+    Flow[SslTlsInbound].log(name, {
+      case s @ SessionTruncated         ⇒ s
+      case SessionBytes(session, bytes) ⇒ "SessionBytes " + printByteString(bytes, maxBytes)
+    })
+
+  def printByteString(bytes: ByteString, maxBytes: Int = MaxBytesPrinted): String = {
```
ktoso: I love this one, let's get it in.

PR #511, opened 2016-11-10
Fix newlines to pass test on windows

2beaucoup:


docs/src/test/scala/docs/http/scaladsl/SprayJsonPrettyMarshalSpec.scala
```diff
@@ -43,10 +42,10 @@ class SprayJsonPrettyMarshalSpec extends server.RoutingSpec {
...
     Get("/") ~> service.route ~> check {
       responseAs[String] shouldEqual
-        """{
-          |  "name": "akka",
-          |  "id": 42
-          |}""".stripMargin
+        "{" + "\n" +
+        "  \"name\": \"akka\"," + "\n" +
+        "  \"id\": 42" + "\n" +
+        "}"
```
jrudolph: Hmm, this is in the documentation, so we either need to add comments to explain what's going on but it would of course be better to find some less verbose alternative we could offer to Windows users.

jonas:
Does it happen with `git config core.autoclf false`?

2beaucoup:
> Does it happen with git config core.autoclf false?

I guess no. But we try to make tests newline agnostic. Unfortunately `stripMarginWithNewline()` isn't available in samples. See https://github.com/akka/akka-http/blob/d0b7829dc6d7091cf106dd9b51c4980c82729e70/docs/src/test/scala/docs/http/scaladsl/server/directives/JsonStreamingExamplesSpec.scala#L107-L109. I could use triple quotes to make it a bit less verbose though.

docs/src/test/scala/docs/http/scaladsl/SprayJsonPrettyMarshalSpec.scala
```diff
@@ -43,10 +42,10 @@ class SprayJsonPrettyMarshalSpec extends server.RoutingSpec {
...
     Get("/") ~> service.route ~> check {
       responseAs[String] shouldEqual
-        """{
-          |  "name": "akka",
-          |  "id": 42
-          |}""".stripMargin
+        """{""" + "\n" +
+        """  "name": "akka",""" + "\n" +
+        """  "id": 42""" + "\n" +
+        """}"""
```
2beaucoup: @jrudolph:
```scala
List(
  """{""",
  """  "name": "akka",""",
  """  "id": 42""",
  """}"""
).mkString("\n")
```
would be another (still kind of ugly) option.

ktoso:
It's an OK improvement, thanks @2beaucoup .

Issue #245, opened 2016-09-12
Model HTTP/2 frame events


jrudolph:
Part of #244 

Frame types of the HTTP/2 specification should be modeled as case classes.

These events have or should been defined:
- [x] DataFrame
- [x] HeadersFrame
- [x] PriorityFrame https://github.com/akka/akka-http/pull/515#issuecomment-261204895
- [x] RstStreamFrame
- [x] SettingsFrame and SettingsAckFrame
- [x] PushPromiseFrame
- [x] PingFrame
- [x] GoAwayFrame https://github.com/akka/akka-http/pull/528
- [x] WindowUpdateFrame 
- [x] ContinuationFrame

ktoso:
Adding the remaining models should be rather simple - feel free to help us out on these.

gosubpl:
@ktoso @jrudolph is that still valid? If yes, I'd like to take on PriorityFrame. Will that be ok?

ktoso:
Absolutely, please hack away :-)

-- 
Konrad `ktoso` Malawski
Akka <http://akka.io> @ Lightbend <http://lightbend.com>

On 25 October 2016 at 14:14:06, gosubpl (notifications@github.com) wrote:

@ktoso <https://github.com/ktoso> @jrudolph <https://github.com/jrudolph>
is that still valid? If yes, I'd like to take on PriorityFrame. Will that
be ok?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/akka/akka-http/issues/245#issuecomment-256017837>, or mute
the thread

ktoso:
Go away worked on by @schmitch here https://github.com/akka/akka-http/pull/528

jrudolph:
Support for the last one (`PUSH_PROMISE`) was added in 3ab873a.

PR #515, opened 2016-11-11
+ht2 #245 parse PriorityFrame but without rendering

gosubpl:
PriorityFrame HTTP2 model, without renderer atm

### DIFF
--- a/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/FrameEvent.scala
+++ b/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/FrameEvent.scala
@@ -41,6 +41,12 @@ final case class ContinuationFrame(
   endHeaders: Boolean,
   payload:    ByteString) extends StreamFrameEvent
 
+final case class PriorityFrame(
+  streamId:         Int,
+  exclusiveFlag:    Boolean,
+  streamDependency: Int,
+  weight:           Int) extends StreamFrameEvent
+
 final case class Setting(
   identifier: SettingIdentifier,
   value:      Int)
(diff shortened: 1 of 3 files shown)


gosubpl:
Refs #245

gosubpl:
@jrudolph @ktoso please give me some guidance how to add remaining frame models. I have started with `PriorityFrame` in this PR. What is done - is parsing the frame. What is not done - is rendering the frame back. There is interesting comment: https://github.com/akka/akka-http/blob/master/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/FrameRenderer.scala#L31 
Does it mean that we are not aiming at rendering `PriorityFrames` atm or you think that this should be done in a separate renderer?

ktoso:
AFAICS this is a good step anyway. 
As for rendering PRIORITY frames I don't think we'll support that in the first versions simply, so we don't need to care about rendering I think.

gosubpl:
This just needs one more fix around the bit-twiddling

ktoso:
LGTM from my side, thanks a lot :)

jrudolph:
Thanks, @gosubpl.

jrudolph:
I think adding rendering makes a lot of sense to allow testing of round trips which helps to find problems with edge cases. Apart from that more tests for parsing (for the edge conditions) would be nice as well.

gosubpl:
Yup, I will get around to that. The priority thing should probably be consumed in the Demux and asynchronously change scheduler settings, is there any scaffolding for this to try looking into placing that logic somewhere?
Thanks @jrudolph !

Issue #500, opened 2016-11-09
`Server` string is still `akka-http/2.4.12`


jrudolph:
```
# curl -v http://localhost:8080

* Connected to localhost (127.0.0.1) port 8080 (#0)
> GET / HTTP/1.1
> User-Agent: curl/7.35.0
> Host: localhost:8080
> Accept: */*
> 
< HTTP/1.1 200 OK
* Server akka-http/2.4.12 is not blacklisted
< Server: akka-http/2.4.12
< Date: Wed, 09 Nov 2016 10:56:04 GMT
< Content-Type: text/plain; charset=UTF-8
< Content-Length: 11
< 
* Connection #0 to host localhost left intact
Hello world%  
```

jrudolph:
Tested with 10.0.0-RC2

ktoso:
Ah...! Good catch... Since we don't have our previous release script which
changed those things I think.
I can work on this and the other 2 build things in my morning in addition
to the entity btw.

-- 
Konrad `ktoso` Malawski
Akka <http://akka.io> @ Lightbend <http://lightbend.com>

On 9 November 2016 at 02:57:35, Johannes Rudolph (notifications@github.com)
wrote:

Tested with 10.0.0-RC2

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/akka/akka-http/issues/500#issuecomment-259387562>, or mute
the thread

PR #518, opened 2016-11-12
Fix server header and user agent strings to show the Akka HTTP version (#500)

jonas:


project/Version.scala
```diff
@@ -0,0 +1,44 @@
...
+      """|akka.http.version = "%s"
+         |"""),
+    sourceGenerators <+= generateVersion(sourceManaged, _ / "akka" / "http" / "Version.scala",
+      """|package akka.http
+         |
+         |import com.typesafe.config.Config
+         |
+         |object Version {
+         |  val current: String = "%s"
+         |  def check(config: Config): Unit = {
```
jonas: If it's preferred to not have this code in a generated file I can move it to HttoExt. Also not sure if this check makes sense for Akka HTTP.

akka-http-core/src/main/resources/reference.conf
```diff
@@ -5,14 +5,18 @@
...
+
 akka.http {
 
   server {
     # The default value of the `Server` header to produce if no
     # explicit `Server`-header was included in a response.
     # If this value is the empty string and no header was included in
     # the request, no `Server` header will be rendered at all.
-    server-header = akka-http/${akka.version}
+    server-header = akka-http/${akka.http.version}
```
ktoso: ah right :)

akka-http-core/build.sbt
```diff
@@ -1,4 +1,3 @@
 import akka._
 
 //OSGi.httpCore
-Dependencies.httpCore
```
ktoso: I think that's good, wasn't it?

build.sbt
```diff
@@ -60,6 +60,7 @@ lazy val parsing = project("akka-parsing")
 
 lazy val httpCore = project("akka-http-core")
   .settings(Dependencies.httpCore)
+  .settings(Version.versionSettings)
```

project/Version.scala
ktoso: Yeah, check is good - imagine if people would force different versions of akka-http and akka-http core - it could break (we are binary compatible externally, but those modules may evolve "together" internally and need exact the same version)

ktoso:
Resolves #500

akka-http-core/build.sbt
jonas: It was added both here and in the root build.sbt file.

Issue #462, opened 2016-11-02
Failed: ConnectionPoolSpec 


jrudolph:
Seems it failed for the first time here https://jenkins.akka.io:8498/job/pr-validator-akka-http/199/testReport/junit/akka.http.impl.engine.client/ConnectionPoolSpec/The_host_level_client_infrastructure_should_not_open_a_second_connection_if_there_is_an_idle_one_available/

```
java.lang.AssertionError: Expected OnNext(_), yet no element signaled during 3 seconds
	at akka.stream.testkit.TestSubscriber$ManualProbe.expectNext(StreamTestKit.scala:268)
	at akka.stream.testkit.TestSubscriber$ManualProbe.expectNext(StreamTestKit.scala:259)
	at akka.http.impl.engine.client.ConnectionPoolSpec$$anonfun$1$$anonfun$apply$mcV$sp$4$$anon$7.<init>(ConnectionPoolSpec.scala:132)
	at akka.http.impl.engine.client.ConnectionPoolSpec$$anonfun$1$$anonfun$apply$mcV$sp$4.apply(ConnectionPoolSpec.scala:121)
	at akka.http.impl.engine.client.ConnectionPoolSpec$$anonfun$1$$anonfun$apply$mcV$sp$4.apply(ConnectionPoolSpec.scala:121)
	at org.scalatest.OutcomeOf$class.outcomeOf(OutcomeOf.scala:85)
	at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
	at org.scalatest.Transformer.apply(Transformer.scala:22)
```

ktoso:
Merged a timeout improvement, if does not happen then close on next cleanup day

jrudolph:
Since I disabled a7 we almost had no issues any more.

ktoso:
I see, that's a waste though... hoped we could use it. What issue were we stuck on?

jrudolph:
You mean with a7? That it failed all tests ;)

ktoso:
Well it had other failure behaviours (ssh etc), so only that it failed tests? I see... that's tough to act on :(

PR #519, opened 2016-11-13
=ht #462 introduce timeout control setting and increase timeout from …

gosubpl:
…3s to 5s in ConnectionPoolSpec


gosubpl:
Not sure if this is going to resolve #462 but I have seen some failures of `ConnectionPoolSpec` also locally. As a workaround the timeout is increased from default 3s to 5s.

ktoso:
LGTM

Issue #454, opened 2016-11-01
Add Introduction is Akka-Http documentation section for Java


jlprat:
Scala documentation contains a section called introduction where the project is introduced, this section is missing in the Java counterpart.

ktoso:
Thanks, good point - could up submit a PR adding it? Thanks in advance.
Can be a copy adjusted eith java wording.

jlprat:
I guess also with Java code snippets, right?

ktoso:
Yes if it has samples those should be made java. Should be fairly simple with the new javadsl it's very similar to the scala one.

Thanks a lot!

jlprat:
For the code samples that involve some `case class`, should I add them as inner classes?

ktoso:
Yeah, precisely :) Thanks!

jlprat:
I have 3 out of 6 snippets translated to Java, do you want me to do a WIP pull request, so those files can already be checked if you have any comment?

ktoso:
Sure :) might as well catch something early :)
Thanks a lot!

Issue #493, opened 2016-11-07
FAILED HttpServerBug21008Spec


jrudolph:
Probably related to #396 or #448. Probably doesn't make sense to work on it as long as those have not been fixed.

https://jenkins.akka.io:8498/job/pr-validator-akka-http/238/testReport/junit/akka.http.impl.engine.server/HttpServerBug21008Spec/The_HttpServer_should_not_cause_internal_graph_failures_when_consuming_a__100_Continue__entity_triggers_a_failure/

```
akka.http.impl.engine.server.HttpServerBug21008Spec.The HttpServer should not cause internal graph failures when consuming a `100 Continue` entity triggers a failure

org.scalatest.exceptions.TestFailedException: HttpServerBluePrint.ControllerStage: requirement failed: Cannot pull closed port (requestParsingIn),
inside HttpRequest(HttpMethod(POST),http://example.com/,List(Host: example.com, Expect: 100-continue, Timeout-Access: <function1>),HttpEntity.Chunked(application/octet-stream),HttpProtocol(HTTP/1.1))
	at org.scalatest.Assertions$class.newAssertionFailedException(Assertions.scala:528)
	at akka.testkit.AkkaSpec.newAssertionFailedException(AkkaSpec.scala:57)
	at org.scalatest.Assertions$class.fail(Assertions.scala:1089)
	at akka.testkit.AkkaSpec.fail(AkkaSpec.scala:57)
	at akka.http.impl.engine.server.HttpServerBug21008Spec$$anonfun$1$$anonfun$apply$mcV$sp$1$$anonfun$apply$1$$anon$1$$anonfun$2.applyOrElse(HttpServerBug21008Spec.scala:70)
	at akka.http.impl.engine.server.HttpServerBug21008Spec$$anonfun$1$$anonfun$apply$mcV$sp$1$$anonfun$apply$1$$anon$1$$anonfun$2.applyOrElse(HttpServerBug21008Spec.scala:38)
	at scala.runtime.AbstractPartialFunction.apply(AbstractPartialFunction.scala:36)
	at org.scalatest.Inside$class.inside(Inside.scala:128)
	at akka.http.impl.engine.server.HttpServerBug21008Spec.inside(HttpServerBug21008Spec.scala:18)
	at akka.http.impl.engine.server.HttpServerBug21008Spec$$anonfun$1$$anonfun$apply$mcV$sp$1$$anonfun$apply$1$$anon$1.<init>(HttpServerBug21008Spec.scala:38)
[...]

PR #521, opened 2016-11-14
Add Intro documentation for Java API (#454)

jlprat:
Issue: #454
Add Java snippets for:

- Minimal Server (#minimal-routing-example)
- Spray JSON example (#second-spray-json-example)
- Streaming Random Numbers (#stream-random-numbers)

Missing Java snippets for:
- Actor Interaction (#actor-interaction)
- Low Level Server Example (#actor-interaction)
- Single Client Request (#single-request-example)

### DIFF
--- a/docs/src/main/paradox/java/http/index.md
+++ b/docs/src/main/paradox/java/http/index.md
@@ -31,6 +31,7 @@ akka-http-jackson
 
 @@@ index
 
+* [introduction](introduction.md)
 * [configuration](configuration.md)
 * [common/index](common/index.md)
 * [http-model](http-model.md)
--- a/docs/src/main/paradox/java/http/introduction.md
+++ b/docs/src/main/paradox/java/http/introduction.md
@@ -21,6 +21,8 @@ However, if your application is not primarily a web application because its core
 
 Akka HTTP was designed specifically as “not-a-framework”, not because we don’t like frameworks, but for use cases where a framework is not the right choice. Akka HTTP is made for building integration layers based on HTTP and as such tries to “stay on the sidelines”. Therefore you normally don’t build your application “on top of” Akka HTTP, but you build your application on top of whatever makes sense and use Akka HTTP merely for the HTTP integration needs.
 
+On the other hand, if you prefer to build your applications with the guidance of a framework, you should give Play Framework or Lagom a try, which both use Akka internally.
+
 ## Using Akka HTTP
 
 Akka HTTP is provided in a separate jar file, to use it make sure to include the following dependency:
--- a/docs/src/main/paradox/java/http/introduction.md
+++ b/docs/src/main/paradox/java/http/introduction.md
@@ -52,8 +52,7 @@ done separately from the route declarations, in marshallers, which are pulled in
 This means that you can `complete` a request with any kind of object a as long as there is an implicit marshaller
 available in scope.
 
-Default marshallers are provided for simple objects like String or ByteString, and you can define your own for example
-for JSON. An additional module provides JSON serialization using the spray-json library (see @ref[JSON Support](common/json-support.md#akka-http-spray-json)
+JSON support is possible in `akka-http` by the use of Jackson, an external artifact (see @ref[JSON Support](common/json-support.md#json-support-via-jackson)
 for details).
 
 The `Route` created using the Route DSL is then "bound" to a port to start serving HTTP requests:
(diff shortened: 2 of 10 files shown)


jlprat:
Note:
As all examples should be complete (including imports) I put each example as a separated file so it can be included nicely.
Unfortunately, in Java I can't add imports in the middle of the class ;)

ktoso:
OK TO TEST

jonas:
It is possible to include multiple snippets in Paradox, either using:
```java
// Java
//#imports
import akka.actor.ActorSystem;
//#imports
...
class ExampleTest {
  //#usage
  //#usage
}

// Docs
@@snip [ExampleSpec](.../ExampleSpec.java) { #imports #usage }
```
Or using the same snippet ID:
```java
// Java
//#some-example
import akka.actor.ActorSystem;
//#some-example
...
class ExampleTest {
  //#some-example
  //#some-example
}

// Docs
@@snip [ExampleSpec](.../ExampleSpec.java) { #imports #usage }
```

For example see the [HttpClientExampleDocTest.java](https://github.com/akka/akka-http/blob/master/docs/src/test/java/docs/http/javadsl/HttpClientExampleDocTest.java) used [here](http://doc.akka.io/docs/akka-http/current/java/http/implications-of-streaming-http-entity.html#consuming-the-http-response-entity-client-).

One improvement we might want to do in Paradox is to have an option to "undent" all snippets to the same level so the example above doesn't show:

```java
import akka.actor.ActorSystem;
``

**Comment by jlprat** [comment]
That's a cool feature, I didn't know!
However, I don't know if I would be wise to merge those examples in the same file as somebody hitting "reorganize imports" might change the order of the imports causing bad examples.

**Comment by jonas** [comment]
True, I've been thinking of adding Paradox support for `@@@ snip` blocks which would generate a test file and compile/run it for this sort of thing and to make it simpler to write small testable examples.

**Comment by jlprat** [comment]
Added 3 remaining snippets.
Couldn't test paradox generation due to some strange error. Will do tomorrow

File: docs/src/main/paradox/java/http/index.md [hunk 1]
```diff
@@ -31,6 +31,7 @@ akka-http-jackson
 
 @@@ index
 
+* [introduction](introduction.md)
```
ktoso: 👍

jlprat:
This PR is not a WIP anymore :)

docs/src/main/paradox/java/http/introduction.md
```diff
@@ -0,0 +1,137 @@
...
+
+However, if your application is not primarily a web application because its core is not browser-interaction but some specialized maybe complex business service and you are merely trying to connect it to the world via a REST/HTTP interface a web-framework might not be what you need. In this case the application architecture should be dictated by what makes sense for the core not the interface layer. Also, you probably won’t benefit from the possibly existing browser-specific framework components like view templating, asset management, JavaScript- and CSS generation/manipulation/minification, localization support, AJAX support, etc.
+
+Akka HTTP was designed specifically as “not-a-framework”, not because we don’t like frameworks, but for use cases where a framework is not the right choice. Akka HTTP is made for building integration layers based on HTTP and as such tries to “stay on the sidelines”. Therefore you normally don’t build your application “on top of” Akka HTTP, but you build your application on top of whatever makes sense and use Akka HTTP merely for the HTTP integration needs.
```
jlprat: Sounds good, this should be added to both docs (Java and Scala) then. Would it be better to have a separated issue for this?

ktoso: In both I think :) Thanks! Will review the rest in a bit.

ktoso:
OK TO TEST

ktoso:
PLS BUILD

jlprat:
Did I break any test?

jonas:
Some of the tests are rather timing sensitive which means that they sometimes fail if the Jenkins servers are overloaded. When this happens a ticket with the `failure` label is created to track it. In this case there's already a ticket: #493.

docs/src/main/paradox/java/http/introduction.md
jlprat: So, should I add this text under this PR? or better have a new issue + PR for better traceability?

ktoso: Here is fine, thanks!

jlprat: Just pushed, let's see if you like the wording

docs/src/test/java/docs/http/javadsl/SprayJsonExampleTest.java
```diff
@@ -0,0 +1,123 @@
...
+
+  private Route createRoute() {
+
+    return route(
+      get(() ->
+        pathPrefix("item", () ->
+          path(longSegment(), (Long id) -> {
+            final CompletionStage<Optional<Item>> futureMaybeItem = fetchItem(id);
+            return onSuccess(() -> futureMaybeItem, maybeItem ->
+              maybeItem.map(item -> completeOK(item, Jackson.marshaller()))
```
ktoso: This is not using spray 😉  It's using Jackson, all our docs examples for java (and support) is for jackson

docs/src/main/paradox/java/http/introduction.md
```diff
@@ -0,0 +1,139 @@
...
+akka-http-core
+: A complete, mostly low-level, server- and client-side implementation of HTTP (incl. WebSockets)
+Details can be found in sections @ref[Low-Level Server-Side API](server-side/low-level-server-side-api.md#http-low-level-server-side-api) and @ref[Consuming HTTP-based Services (Client-Side)](client-side/index.md#http-client-side)
+
+akka-http-testkit
+: A test harness and set of utilities for verifying server-side service implementations
+
+akka-http-spray-json
+: Predefined glue-code for (de)serializing custom types from/to JSON with [spray-json](https://github.com/spray/spray-json)
+Details can be found here: @ref[JSON Support](common/json-support.md#akka-http-spray-json)
```
ktoso: This should be removed - java does not care about spray :-)
See: See https://github.com/akka/akka-http/pull/521/files#r88125838

Toe example we want to point to here is jackson

ktoso:
One small thing to fix about jackson/spray-json and then LGTM :-)
Thanks a lot for converting these!

jlprat:
I will change it later today. I didn't even think about renaming the examples to Jackson, even if it was quite obvious...

ktoso:
Thanks, and just in time for tomorrow's release :)

Issue #516, opened 2016-11-11
Cannot pull port (requestParsingIn) twice


rgmccaw:
I'm using Akka HTTP (3.0.0-RC1) and I am getting an Exception in certain cases.
I can reproduce it by sending my application a request through a Burp Suite proxy. Below are three curl requests; two which do not cause the exception and one that does.

For the following CURLs the following is true:
* `test` is an empty file
* There is Burp Suite proxy listening at `127.0.0.1:8282`
* My application is `localhost:9090`

POST directly to the application
```
# Note that -F adds the 'Expect: 100-Continue' header
curl -v "http://localhost:9090/upload" -F "file=@/tmp/test"
```
```
12:56:23.128 [default-akka.actor.default-dispatcher-13] DEBUG - Request is: HttpRequest(HttpMethod(POST),http://localhost:9090/upload,List(Remote-Address: 127.0.0.1:64301, Host: localhost:9090, User-Agent: curl/7.43.0, Accept: */*, Expect: 100-continue, Timeout-Access: <function1>),HttpEntity.Default(multipart/form-data; boundary=------------------------287d87847c39d574,196 bytes total),HttpProtocol(HTTP/1.1))
```
POST through Burp Suite then to the application
```
curl -v -x 127.0.0.1:8282 "http://localhost:9090/upload" -F "file=@/tmp/test"
```
```
[...]

Issue #514, opened 2016-11-11
Special handling of Expect: 100-Continue may not be appropriate in a proxy


derekwyatt:
I'm using Akka HTTP (3.0.0-RC1) to create a reverse-http-proxy.  The special handling of `Expect: 100-Continue` looks like it's causing weirdness in this case.  In wireshark, here's what I see:
```
Proxy  -> RSP: 100 Continue -> Client
Client -> REQ -> Proxy
Server -> RSP: 100 Continue -> Proxy
Proxy  -> REQ -> Server
Server -> 200 -> Proxy
Proxy  -> 200 -> Client
```
In the production logs (not in an attempt at a clean room reproduction) we see a number of these messages:
```
Sending an 2xx 'early' response before end of request was received...
```
It's not 100% clear that these are related to the special handling of 100, but they do seem to coincide a lot.
It _feels_ like the special handling should be avoided when `akka-http` is operating as a proxy, leaving the handling logic up to the server to which we're proxying.

ktoso:
Early response is when you send a reply, before having read the entire request entity AFAIR.
Technically not all clients survive this under some scenarios so we log this warning.
Perhaps it could be made less noisy.

Are you seeing actual problems with it or just the warning and things work?

derekwyatt:
We're still investigating.  A) it's not 100% clear that the special handling and the warning are related; but it certainly seems that way, B) we do see a strong correlation between this warning, and our proxy "pausing".

We're handling, say, 50-100 request per second.  Then we see this warning, and there is no req/rsp activity for 2-4 seconds.  It doesn't seem as though we're dropping anything on the floor but there are definitely no requests logged for 2-4 seconds.  So it seems as though our proxy just goes into a global "wait state" for a bit.

That effect seems quite real, and the correlation seems quite real as well.  We haven't gone deep enough to investigate a cause, however.

ktoso:
No idle-timeouts triggering?
Idle timeout is 60s which seems to match up... if you'd not consume some
entity to the end, reply right away, then idle timeout would eventually
kill the connection and resume operation perhaps...

How many connections do you have?

-- 
Konrad `ktoso` Malawski
Akka <http://akka.io> @ Lightbend <http://lightbend.com>

On 11 November 2016 at 10:44:16, Derek Wyatt (notifications@github.com)
wrote:

We're still investigating. A) it's not 100% clear that the special handling
and the warning are related; but it certainly seems that way, B) we do see
[...]

PR #522, opened 2016-11-14
Fixes for #514 and #516

derekwyatt:
- Introduces a new configuration `akka.http.server.proxy-mode`
- When `proxy-mode` is `on` the special handling of the `Expect:
  100-Continue` behaviour will be disabled.
- This alleviates the problem illustrated by
  https://github.com/derekwyatt/redirect-and-continue


akka-http-core/src/main/resources/reference.conf
```diff
@@ -87,6 +87,11 @@ akka.http {
     # error details.
     verbose-error-messages = off
 
+    # Puts the server into a reverse-proxy mode
+    # This, effectively, disables special handling of the Expect: 100-continue
+    # behaviour
+    proxy-mode = off
+
```
derekwyatt: I expect that this will be a fairly contentious issue. The problem is that a "correct" fix (e.g. a new directive) would have a big challenge ahead of it. The places where the change needs to be made do not have all of the contextual information required in order to make that change.  Hence, the configuration setting.

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -411,16 +411,23 @@ private[http] object HttpServerBluePrint {
...
+            val close = requestStart.closeRequested ||
+              (requestStart.expect100Continue && oneHundredContinueResponsePending) ||
+              (isClosed(requestParsingIn) && openRequests.isEmpty) || early
+            (early, close)
+          }
+
+          if (isEarlyResponse && response.status.isSuccess && !requestStart.expect100Continue)
             log.warning(
-              "Sending an 2xx 'early' response before end of request was received... " +
+              "Sending a {} 'early' response before end of request was received... " +
```
derekwyatt: If nothing else, this change is valuable; it took us quite a while to figure out that the true response was a 301.

jrudolph: I think we should fix it properly by not looking for `status.isSuccess` but by checking if it is a subclass of `StatusCode.Success`.

jrudolph:
Let's close this PR for now. I think we've come to the conclusion that fixing both #514 and #516 might be possible without having to introduce special cases.

PR #528, opened 2016-11-15
+ht2 #245 adds support for GoAwayFrame

schmitch:
See below


ktoso:
OK TO TEST

ktoso:
And adding to whitelist :-) Thanks a lot @schmitch !
Message is fine, we usually do a `refs #23232` in the commit - so github links PR to issue, the rest is not important.

akka-http-core/src/main/scala/akka/http/impl/util/package.scala
```diff
@@ -63,6 +63,13 @@ package object util {
         case Failure(ex) ⇒ throw new RuntimeException("Trying to await result of failed Future, see the cause for the original problem.", ex)
       }
     }
+    def awaitWithException(atMost: Duration): T = {
```
jrudolph: `awaitResult` also keeps the exception as the cause but does also include an exception to find out at which place it was called. So, no need to introduce another method.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/FrameParser.scala
```diff
@@ -46,10 +46,20 @@ class FrameParser(shouldReadPreface: Boolean) extends ByteStringParser[FrameEven
       }
     }
 
+  def checkCompliance(streamId: Int): Unit = {
+    // stream id's needs to follow http://httpwg.org/specs/rfc7540.html#rfc.section.5.1.1
+    // actually if we add support for PUSH_PROMISE we might need to change the server compliance
```
jrudolph: Does a server need to parse a PUSH_PROMISE frame?

schmitch:
I addressed the comments and merged it with my GoAwayFrame support, excluding the Connection close.

Basically this handle's *most* stream id checks, basically what I still dislike is the pattern match stuff, maybe I change it and check the things before we pattern match, not sure what is prefered?

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/FrameRenderer.scala
```diff
@@ -18,6 +18,20 @@ object FrameRenderer {
 
   def render(frame: FrameEvent): ByteString =
     frame match {
+      case GoAwayFrame(lastStreamId, errorCode, debug) ⇒
+        val bb = new ByteStringBuilder
+        // TODO: render this more correct at the moment it will just kill the reserved bit
+        bb.putInt(0x0FFFFFFF & lastStreamId)
```
jrudolph: This will kill more than the reserved bit. Maybe just add a requirement in `GoAwayFrame` to prevent negative stream ids  in the first place?

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -306,15 +306,31 @@ class Http2ServerSpec extends AkkaSpec with WithInPendingUntilFixed {
     }
 
     "respect the substream state machine" should {
-      "reject other frame than HEADERS/PUSH_PROMISE in idle state with connection-level PROTOCOL_ERROR (5.1)" in pending
+      abstract class SimpleRequestResponseRoundtripSetup extends TestSetup with RequestResponseProbes
+
+      "reject other frame than HEADERS/PUSH_PROMISE in idle state with connection-level PROTOCOL_ERROR (5.1)" in new SimpleRequestResponseRoundtripSetup {
+        sendDATA(9, endStream = true, HPackSpecExamples.C41FirstRequestWithHuffman)
+        expectGOAWAY(0)
```
jrudolph: We should also check the any other kind of necessary connection closure procedure here (and in the other tests).

schmitch: well the connection closing procedure still needs to be done, so at the moment it won't close the connection, so maybe I just add pending and check if the connection is closed here after a while (which than will fail) until the connection close is fixed? I guess closing the connection is a PR by it's own. WDYT?

ktoso:
Needs a rebase now though - sorry about the conflict in Demux :)

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -115,17 +142,20 @@ class Http2ServerDemux extends GraphStage[BidiShape[Http2SubStream, FrameEvent,
...
-                  if (updatedHandler.headers.endHeaders) dispatchSubstream(sub)
-                case None ⇒
-                  throw ContinuationFrameForUnknownStreamEncounteredException(streamId)
-              }
+              val streamHandler = incomingStreams(streamId)
+              val updatedHandler = concatContinuationIntoHeaders(streamHandler, cont)
+              val sub = makeHttp2SubStream(updatedHandler)
+              if (updatedHandler.headers.endHeaders) dispatchSubstream(sub)
+
+            case e: StreamFrameEvent if !incomingStreams.contains(e.streamId) ⇒
```
schmitch: ups, that needs to be above ContinuationFrame

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/FrameParser.scala
```diff
@@ -49,6 +49,11 @@ class FrameParser(shouldReadPreface: Boolean) extends ByteStringParser[FrameEven
   def parseFrame(tpe: FrameType, flags: ByteFlag, streamId: Int, payload: ByteReader): FrameEvent = {
     // TODO: add @switch? seems non-trivial for now
     tpe match {
+      case GOAWAY ⇒
+        Http2Compliance.requireZeroStreamId(streamId)
```
ktoso: Ok, that's probably ok; and we'd handle the 

> The GOAWAY frame applies to the connection, not a specific stream. An endpoint MUST treat a GOAWAY frame with a stream identifier other than 0x0 as a connection error (Section 5.4.1) of type PROTOCOL_ERROR.

on the error handling level, though perhaps specific exception that it was a GOAWAY would be better then?

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/FrameParser.scala
```diff
@@ -49,6 +49,11 @@ class FrameParser(shouldReadPreface: Boolean) extends ByteStringParser[FrameEven
...
+        GoAwayFrame(payload.readIntBE(), ErrorCode.byId(payload.readIntBE()),
+          if (payload.hasRemaining) Some(payload.takeAll().utf8String) else None)
```
ktoso: I'm not sure it's a string, could be just ByteString IMO - keep as BS?

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/FrameEvent.scala
```diff
@@ -14,6 +14,7 @@ sealed trait StreamFrameEvent extends FrameEvent {
   def streamId: Int
 }
 
+final case class GoAwayFrame(lastStreamId: Int, errorCode: ErrorCode, debug: Option[String] = None) extends FrameEvent
```
ktoso: ByteString instead of String for the debug data? I don't see in the spec where it said that it's a string.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Compliance.scala
```diff
@@ -16,11 +16,10 @@ private[akka] object Http2Compliance {
...
 
-  // FIXME: this should be checking client/server ids instead (odd / even)
-  @Deprecated final def requirePositiveStreamId(id: Int): Unit =
-    if (id < 1) throw new IllegalHttp2StreamIdException(id, "MUST be > 0.")
-
   final def requireZeroStreamId(id: Int): Unit =
     if (id != 0) throw new IllegalHttp2StreamIdException(id, "MUST BE == 0.")
 
+  /** checks if the stream id was client initiated, by checking if the stream id was odd-numbered */
+  final def clientInitiatedStreamId(id: Int): Boolean = id % 2 != 0
```
ktoso: `isClientInitiatedStreamId`?

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -98,15 +106,34 @@ class Http2ServerDemux extends GraphStage[BidiShape[Http2SubStream, FrameEvent,
         bufferedFrameOut.push(SettingsFrame(Nil)) // server side connection preface
       }
 
-      var incomingStreams = Map.empty[Int, SubStream]
+      var closedAfter: Option[Int] = None
```
ktoso: add a comment that this is for handling the "we should not handle streams later than the GOAWAY told us about with lastStreamId"

ktoso: For good measure let's add `private` to all these `var`s btw

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -98,15 +106,34 @@ class Http2ServerDemux extends GraphStage[BidiShape[Http2SubStream, FrameEvent,
...
+      var incomingStreams = mutable.Map.empty[Int, SubStream]
       var totalOutboundWindowLeft = Http2Protocol.InitialWindowSize
       var streamLevelWindow = Http2Protocol.InitialWindowSize
 
+      def lastStreamId: Int = {
+        incomingStreams.lastOption.map(_._1).getOrElse(0)
+      }
+
+      def goAwayFrame(errorCode: ErrorCode = Http2Protocol.ErrorCode.PROTOCOL_ERROR): Unit = {
```
ktoso: Can we call it `pushGOAWAY`? I really liked that convention we had in tests about the frame types when we push/pull etc. Here it'd also help to know it performs a push.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
ktoso: Pending test would be good, would you follow up on the closing handling afterwards?
I think it'd be outside of the Demux by reacting to the GOAWAY

ktoso:
Looks very good, had some nitpicks and questions :)
LGTM after those addressed, the close behaviour we can add in the next PR, WDYT?

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
schmitch: yeah I probably will handle that aswell, well my first idea was just completionStage the demuxer, also there are cases where you don't have a GOAWAY frame and still want to close the connection, i.e. malicious clients.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Compliance.scala
```diff
@@ -16,11 +16,10 @@ private[akka] object Http2Compliance {
...
   // require methods use `if` because `require` allocates
 
-  // FIXME: this should be checking client/server ids instead (odd / even)
-  @Deprecated final def requirePositiveStreamId(id: Int): Unit =
-    if (id < 1) throw new IllegalHttp2StreamIdException(id, "MUST be > 0.")
-
   final def requireZeroStreamId(id: Int): Unit =
     if (id != 0) throw new IllegalHttp2StreamIdException(id, "MUST BE == 0.")
 
+  /** checks if the stream id was client initiated, by checking if the stream id was odd-numbered */
```
schmitch: well probably the odd-numbered is only needed if we do the client, but I guess we can just add a requiredPositiveStreamId on **every** frame and handle the `Illegalhttp2StreamIdException` inside the demuxer as a goaway

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -512,6 +546,11 @@ class Http2ServerSpec extends AkkaSpec with WithInPendingUntilFixed {
     def expectRequestRaw(): HttpRequest = requestIn.requestNext() // TODO, make it so that internal headers are not listed in `headers` etc?
     def emitResponse(streamId: Int, response: HttpResponse): Unit =
       responseOut.sendNext(response.addHeader(Http2StreamIdHeader(streamId)))
+
+    def expectGracefulCompletion(): Unit = {
+      toNet.expectComplete()
```
schmitch: is there a better way to check the test probe's for completion/close, maybe I address that in the connection close handle pr..?

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/FrameParser.scala
schmitch: I actually added `onUpstreamFailure` to the demuxer to clarify what I meant, basically now we might re-add `requirePositiveStreamId` (https://github.com/akka/akka-http/pull/528/files#r88233191) and handle these also in the `FrameParser`, if it is fine to handle parser exceptions in the demuxer.

schmitch:
I addressed most/all (?) comments.
I actually will take on after that to make the disconnect logic.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
ktoso: This looks OK IMO

ktoso:
LGTM from my side, thanks a lot @schmitch ! Super cool to have you collaborating here.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/FrameEvent.scala
jrudolph: Hmm... Why would someone put sensitive data in there? If we go by the logic that someone could send us sensitive data in any kind of field we couldn't log anything.

In cases where *we* want to log the GoAwayFrame we also want to see its contents so it doesn't seem useful for me to hide it here. Does anyone know the logic behind that paragraph in the spec?

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/FrameEvent.scala
```diff
@@ -14,6 +14,9 @@ sealed trait StreamFrameEvent extends FrameEvent {
   def streamId: Int
 }
 
+final case class GoAwayFrame(lastStreamId: Int, errorCode: ErrorCode, debug: Option[ByteString] = None) extends FrameEvent {
```
jrudolph: `debug` can be just a `ByteString` with `ByteString.Empty` as the default.

ktoso: True, good idea.

ktoso:
Having trouble rebasing this on top of recent changes so it still works, would you have some time to try @schmitch ?

ktoso:
Yay! Thanks for the rebase.

ktoso:
#245

Issue #464, opened 2016-11-02
HTTP/2: implement upgrade switch between Http/2 Http/1 based on TLS ALPN


jrudolph:
Currently, we have the Http/2 stack hard-wired regardless of whether a client actually support Http/2. We need to support fallback to the Http/1.1 stack depending on what the client sends out via ALPN.

ktoso:
I refer to this as "ALPN Upgrade support", wrong wording in my vocabulary?

jrudolph:
Yep, agree will change the title.

2beaucoup:
Would this make it possible to have generic `bindAndHandleX` and friends?

jrudolph:
Generic `bindAndHandle` could be the next step. The change here is a mandatory requirement to be able support Http/2 and Http/1 clients on a single port.

ktoso:
This is in progress right?

PR #529, opened 2016-11-15
ALPN switch between Http/1 vs Http/2

jrudolph:
This will still contain a serious hack to transport the chosen ALPN protocol to the switch stage because the switch is done based on the engine (which we would never actually see) and the streaming TLS API only gives access to the SSLSession which doesn't seem to offer any backreferences to the engine...

/cc @ktoso

### DIFF
--- a/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/AlpnSwitch.scala
+++ b/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/AlpnSwitch.scala
@@ -1,0 +1,122 @@
+/**
+ * Copyright (C) 2009-2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.http.impl.engine.http2
+
+import javax.net.ssl.{ SSLContext, SSLException }
+
+import akka.NotUsed
+import akka.http.scaladsl.model.{ HttpRequest, HttpResponse }
+import akka.stream.TLSProtocol.{ SessionBytes, SessionTruncated, SslTlsInbound, SslTlsOutbound }
+import akka.stream.scaladsl.{ BidiFlow, Flow }
+import akka.stream.stage.{ GraphStage, GraphStageLogic, InHandler, OutHandler }
+import akka.stream.{ Attributes, BidiShape, Inlet, Outlet }
+
+import scala.concurrent.Future
+
+object AlpnSwitch {
+  type HttpServerBidiFlow = BidiFlow[HttpResponse, SslTlsOutbound, SslTlsInbound, HttpRequest, NotUsed]
+
+  def apply(
+    chosenProtocolAccessor: () ⇒ String,
+    http1Stack:             HttpServerBidiFlow,
+    http2Stack:             HttpServerBidiFlow): HttpServerBidiFlow =
+    BidiFlow.fromGraph(
+      new GraphStage[BidiShape[HttpResponse, SslTlsOutbound, SslTlsInbound, HttpRequest]] {
+        val netIn = Inlet[SslTlsInbound]("AlpnSwitch.netIn")
+        val netOut = Outlet[SslTlsOutbound]("AlpnSwitch.netOut")
+
+        val requestOut = Outlet[HttpRequest]("AlpnSwitch.requestOut")
+        val responseIn = Inlet[HttpResponse]("AlpnSwitch.responseIn")
+
+        val shape: BidiShape[HttpResponse, SslTlsOutbound, SslTlsInbound, HttpRequest] =
+          BidiShape(responseIn, netOut, netIn, requestOut)
+
+        def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new GraphStageLogic(shape) {
+          override def preStart(): Unit = pull(netIn)
+
+          setHandler(netIn, new InHandler {
+            def onPush(): Unit =
+              grab(netIn) match {
+                case first @ SessionBytes(session, bytes) ⇒
+                  val chosen = chosenProtocolAccessor()
+                  chosen match {
+                    case "h2" ⇒ install(http2Stack, first)
+                    case _    ⇒ install(http1Stack, first)
+                  }
+                case SessionTruncated ⇒ failStage(new SSLException("TLS session was truncated (probably missing a close_notify packet)."))
+              }
+          })
+
+          val ignorePull = new OutHandler { def onPull(): Unit = () }
+          val failPush = new InHandler { def onPush(): Unit = throw new IllegalStateException("Wasn't pulled yet") }
+
+          setHandler(netOut, ignorePull)
+          setHandler(requestOut, ignorePull)
+          setHandler(responseIn, failPush)
+
+          def install(serverImplementation: HttpServerBidiFlow, firstElement: SslTlsInbound): Unit = {
+            val serverDataIn = new SubSinkInlet[SslTlsOutbound]("ServerImpl.netIn")
+            val serverDataOut = new SubSourceOutlet[SslTlsInbound]("ServerImpl.netOut")
+
+            val serverRequestIn = new SubSinkInlet[HttpRequest]("ServerImpl.serverRequestIn")
+            val serverResponseOut = new SubSourceOutlet[HttpResponse]("ServerImpl.serverResponseOut")
+
+            val networkSide = Flow.fromSinkAndSource(serverDataIn.sink, serverDataOut.source)
+            val userSide = Flow.fromSinkAndSource(serverRequestIn.sink, serverResponseOut.source)
+
+            connect(netIn, serverDataOut, Some(firstElement))
+            connect(responseIn, serverResponseOut, None)
+
+            connect(serverDataIn, netOut)
+            connect(serverRequestIn, requestOut)
+
+            serverImplementation
+              .join(networkSide)
+              .join(userSide)
+              .run()(interpreter.subFusingMaterializer)
+          }
+
+          // helpers to connect inlets and outlets
+          def connect[T](in: Inlet[T], out: SubSourceOutlet[T], initialElement: Option[T]): Unit = {
+            val propagatePull =
+              new OutHandler {
+                def onPull(): Unit = pull(in)
+              }
+
+            val firstHandler =
+              initialElement match {
+                case Some(ele) if out.isAvailable ⇒
+                  out.push(ele)
+                  propagatePull
+                case Some(ele) ⇒
+                  new OutHandler {
+                    def onPull(): Unit = {
+                      out.push(initialElement.get)
+                      out.setHandler(propagatePull)
+                    }
+                  }
+                case None ⇒ propagatePull
+              }
+
+            out.setHandler(firstHandler)
+            setHandler(in, new InHandler {
+              def onPush(): Unit = out.push(grab(in))
+            })
+
+            if (out.isAvailable) pull(in) // to account for lost pulls during initialization
+          }
+          def connect[T](in: SubSinkInlet[T], out: Outlet[T]): Unit = {
+            in.setHandler(new InHandler {
+              def onPush(): Unit = push(out, in.grab())
+            })
+            setHandler(out, new OutHandler {
+              def onPull(): Unit = in.pull()
+            })
+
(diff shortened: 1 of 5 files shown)


akka-http2-support/src/main/scala/akka/http/impl/engine/http2/AlpnSwitch.scala
```diff
@@ -0,0 +1,121 @@
+package akka.http.impl.engine.http2
```
ktoso: configure copyright header in your IDE please 😉 
Or... we add the license header plugin?

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/WrappedSslContextSPI.scala
```diff
@@ -6,49 +6,87 @@ package akka.http.impl.engine.http2
...
+import scala.collection.immutable
+import scala.collection.JavaConverters._
 import akka.http.scaladsl.HttpsConnectionContext
 import org.eclipse.jetty.alpn.ALPN
+import org.eclipse.jetty.alpn.ALPN.ServerProvider
+
+object ALPNRegistry {
+  def registerServerEngine(engine: SSLEngine)(decider: immutable.Set[String] ⇒ Option[String]): Unit =
+    ALPN.put(engine, new ServerProvider {
+      def select(protocols: ju.List[String]): String =
```
ktoso: is this `override def`? Please mark as such if yes.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/WrappedSslContextSPI.scala
```diff
@@ -6,49 +6,87 @@ package akka.http.impl.engine.http2
...
+            else ""
+        }
+        finally ALPN.remove(engine)
+    })
+
+  def getChosenProtocolAndRemove(engine: SSLEngine): Option[String] = synchronized {
+    val selected = selections.getOrElse(engine, None)
+    selections -= engine
+    selected
+  }
```
ktoso: Hm, doesn't this need to synchronize (I'd synchronize on selections instance btw, a bit more clear what we're protecting) as well? Synchronized itself probably fine for this use case (we benched similar in BufferPool).

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/WrappedSslContextSPI.scala
```diff
@@ -6,49 +6,87 @@ package akka.http.impl.engine.http2
...
-            throw new SSLException("No supported protocol found. Only 'h2' supported right now.")
-        finally ALPN.remove(engine)
-
-      def unsupported(): Unit = {
-        println("ALPN not supported by client!") // FIXME: remove
-        ALPN.remove(engine)
+  def engineCreateSSLEngine(): SSLEngine =
+    if (theEngine == null) {
+      val engine = underlying.createSSLEngine()
+      ALPNRegistry.registerServerEngine(engine) { protocols ⇒
```
ktoso: For my understanding - so we use this kind of like a side channel to communicate back into the stream, as here we're in a different place completely (someone else creates the engine)?

ktoso:
Refs https://github.com/akka/akka-http/issues/223

ktoso:
Refs also https://github.com/akka/akka-http/issues/464

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/AlpnSwitch.scala
```diff
@@ -0,0 +1,121 @@
...
+
+            def connect[T](in: Inlet[T], out: SubSourceOutlet[T], initialElement: Option[T]): Unit = {
+              var needToFlushFirst =
+                initialElement match {
+                  case Some(ele) if out.isAvailable ⇒
+                    out.push(ele)
+                    false
+                  case Some(ele) ⇒ true
+                  case None      ⇒ false
+                }
```
jrudolph: It's a tiny one-time cost so emit would be fine. Unfortunately Outlet and SubSourceOutlet don't have a common API, so you cannot use `emit` anyways.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/WrappedSslContextSPI.scala
```diff
@@ -6,49 +6,87 @@ package akka.http.impl.engine.http2
...
+import scala.collection.JavaConverters._
 import akka.http.scaladsl.HttpsConnectionContext
 import org.eclipse.jetty.alpn.ALPN
+import org.eclipse.jetty.alpn.ALPN.ServerProvider
+
+object ALPNRegistry {
+  def registerServerEngine(engine: SSLEngine)(decider: immutable.Set[String] ⇒ Option[String]): Unit =
+    ALPN.put(engine, new ServerProvider {
+      def select(protocols: ju.List[String]): String =
+        choose(protocols.asScala.toSet)
```
jrudolph: But it's still an API and it's not important for performance so I'd rather keep it that way.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/WrappedSslContextSPI.scala
jrudolph: Yep, good point. Changed `selections` into a `mutable.WeakHashMap` and changed to use `selections.synchronize`

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/WrappedSslContextSPI.scala
jrudolph: Yes, we need some kind of side-channel as the ALPN handler thing. But right now I figured that this is a much more complicated side-channel than we actually need. I'll update with something much simpler based on a volatile var.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/AlpnSwitch.scala
```diff
@@ -0,0 +1,121 @@
...
+import akka.stream.stage.{ GraphStage, GraphStageLogic, InHandler, OutHandler }
+import akka.util.ByteString
+
+object AlpnSwitch {
+  type HttpServerBidiFlow = BidiFlow[HttpResponse, SslTlsOutbound, SslTlsInbound, HttpRequest, NotUsed]
+
+  def apply(
+    context:   SSLContext,
+    http1Flow: HttpServerBidiFlow,
+    http2Flow: HttpServerBidiFlow): HttpServerBidiFlow =
```
jrudolph: done

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/AlpnSwitch.scala
```diff
@@ -0,0 +1,121 @@
...
+                      install(http2Flow, first)
+                    case _ ⇒
+                      println("Selected Http/1")
+                      install(http1Flow, first)
+                  }
+                case SessionTruncated ⇒
+                  println("Got truncated. Ignoring....")
+              }
+          })
+          setHandler(netOut, new OutHandler {
```
jrudolph: done

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/WrappedSslContextSPI.scala
```diff
@@ -6,49 +6,87 @@ package akka.http.impl.engine.http2
...
-import javax.net.ssl.TrustManager
+import javax.net.ssl._
 
+import scala.collection.immutable
+import scala.collection.JavaConverters._
 import akka.http.scaladsl.HttpsConnectionContext
 import org.eclipse.jetty.alpn.ALPN
+import org.eclipse.jetty.alpn.ALPN.ServerProvider
+
+object ALPNRegistry {
```
jrudolph: done

jrudolph: I got rid of it actually ;)

jrudolph:
Updated with a simpler and cleaner version.

akka-http2-support/src/main/scala/akka/http/scaladsl/Http2.scala
```diff
@@ -60,36 +60,41 @@ class Http2Ext(private val config: Config)(implicit val system: ActorSystem) ext
 
     val effectivePort = if (port >= 0) port else 443
 
-    val serverLayer: BidiFlow[HttpResponse, ByteString, ByteString, HttpRequest, NotUsed] = {
-      val wrappedContext = WrappedSslContextSPI.wrapContext(httpsContext)
-      val tls = http.sslTlsStage(wrappedContext, TLSRole.server)
+    def http2Layer(): BidiFlow[HttpResponse, SslTlsOutbound, SslTlsInbound, HttpRequest, NotUsed] =
```
jrudolph: We cannot reuse the flows any more because of the volatile var. With the akka-stream TLS change, we can get rid of the WrappedSslContext hack but we still have a problem the problem of the sidechannel between TLS and AlpnSwitch that we probably won't be able to solve cleanly.

ktoso: Yeah I see, this triggered the discussion about "stram local place to put things" I assume as well?
So be it for now, but add a comment about this - so someone isn't tempted to "oh, let's make it a val" 😉

ktoso:
LGTM, it's a bit simpler: +1.
Let's add a comment warning why we can't reuse the flows though

akka-http2-support/src/main/scala/akka/http/scaladsl/Http2.scala
```diff
@@ -60,36 +60,41 @@ class Http2Ext(private val config: Config)(implicit val system: ActorSystem) ext
...
-          case SessionBytes(_, bytes) ⇒ bytes
-        })
+    def unwrapTls(): BidiFlow[ByteString, SslTlsOutbound, SslTlsInbound, ByteString, NotUsed] =
+      BidiFlow.fromFlows(Flow[ByteString].map(SendBytes), Flow[SslTlsInbound].collect {
+        case SessionBytes(_, bytes) ⇒ bytes
+      })
 
-      Http2Blueprint.serverStack() atop
-        unwrapTls atop
+    @volatile var chosenProtocol: String = "h1"
```
jrudolph: This is a local variable, no need for private. In fact, it doesn't even need to be volatile because there should already be a happens-before relationship enforced by sending out the first `SessionBytes`. I updated the code with assertions to make sure that it is not written more than once. Also added the comment that the flow cannot be reused.

akka-http2-support/src/main/scala/akka/http/scaladsl/Http2.scala
```diff
@@ -60,36 +60,53 @@ class Http2Ext(private val config: Config)(implicit val system: ActorSystem) ext
...
-        unwrapTls atop
+        unwrapTls
+
+    // Flow is not reusable because we need a side-channel to transport the protocol
+    // chosen by ALPN from the SSLEngine to the switching stage
+    def serverLayer(): BidiFlow[HttpResponse, ByteString, ByteString, HttpRequest, NotUsed] = {
+      // Mutable cell to transport the chosen protocol from the SSLEngine to
+      // the switch stage.
+      // Doesn't need to be volatile because there's a happens-before relationship (enforced by memory barriers)
+      // between the SSL handshake and sending out the first SessionBytes, and receiving the first SessionBytes
```
ktoso: great thanks

akka-http2-support/src/main/scala/akka/http/scaladsl/Http2.scala
jrudolph: Actually, this didn't trigger the discussion (the original trigger of the discussion was about being able to tag a graph with metadata like the localAdress/remoteAddress to use that in log message) but I agree that it would be useful here as well.

ktoso:
Ok, awesome :) Getting away without the volatile also made this less scary 😉

ktoso:
Resolves https://github.com/akka/akka-http/issues/223

Issue #509, opened 2016-11-10
Introduce flag for wire logging to help with debugging problems


jrudolph:
That's especially important for problems with HTTPS as it is hard to log plain text data otherwise.

jrudolph:
#508 introduces a stage for easy logging of streams of ByteStrings (and SslTlsInbound / SslTlsOutbound). We should introduce a config setting which should trigger the addition of those stages to the server and client-side stacks in `Http`.

PR #540, opened 2016-11-17
=htc #509 add security note to log-unencrypted-network-bytes setting

jrudolph:


ktoso:
👍  Yeah, thanks for following up on this.

PR #542, opened 2016-11-17
Provide HttpApp API for Scala and Java

jlprat:
Issue: #156
Add a bootstrap trait that does all the boilerplate for you. Just write as much as you need, not more.

I converted some of the examples under the docs project to Objects extending `HttpApp` to see how the trait works and how much code it saves.

Feedback and additional hooks proposal are highly welcomed!

### DIFF
--- a/akka-http-tests/src/test/java/akka/http/javadsl/server/MinimalHttpApp.java
+++ b/akka-http-tests/src/test/java/akka/http/javadsl/server/MinimalHttpApp.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright (C) 2009-2017 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.http.javadsl.server;
+
+import akka.Done;
+import akka.actor.ActorSystem;
+import scala.runtime.BoxedUnit;
+
+import java.util.concurrent.CompletionStage;
+import java.util.concurrent.CompletableFuture;
+
+public class MinimalHttpApp extends HttpApp {
+
+  CompletableFuture<Done> shutdownTrigger = new CompletableFuture<>();
+
+  @Override
+  protected Route route() {
+    return route(path("foo", () ->
+        complete("bar")
+      ),
+      path("shutdown", () -> {
+        if (shutdownTrigger.complete(Done.getInstance())) {
+          return complete("Shutdown request accepted");
+        } else {
+          return complete("Shutdown is already in progress");
+        }
+      }));
+  }
+
+  @Override
+  protected CompletionStage<Done> waitForShutdownSignal(ActorSystem system) {
+    return shutdownTrigger;
+  }
+}
--- a/akka-http-tests/src/test/scala/akka/http/javadsl/server/HttpAppSpec.scala
+++ b/akka-http-tests/src/test/scala/akka/http/javadsl/server/HttpAppSpec.scala
@@ -1,0 +1,156 @@
+/*
+ * Copyright (C) 2009-2017 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.http.javadsl.server
+
+import java.util.concurrent.TimeUnit
+
+import akka.http.javadsl.settings.ServerSettings
+import akka.http.scaladsl.{ Http, TestUtils }
+import akka.http.scaladsl.client.RequestBuilding
+import akka.http.scaladsl.model.{ HttpRequest, StatusCodes }
+import akka.stream.ActorMaterializer
+import akka.testkit.AkkaSpec
+import com.typesafe.config.ConfigFactory
+import org.scalatest.concurrent.Eventually
+
+import scala.concurrent.duration.Duration
+import scala.concurrent.{ Await, Future }
+import scala.runtime.BoxedUnit
+
+class HttpAppSpec extends AkkaSpec with RequestBuilding with Eventually {
+  import system.dispatcher
+
+  "HttpApp" should {
+
+    "start without ActorSystem" in {
+      val (_, host, port) = TestUtils.temporaryServerHostnameAndPort()
+
+      val minimal = new MinimalHttpApp()
+
+      val server = Future {
+        minimal.startServer(host, port, ServerSettings.create(ConfigFactory.load))
+      }
+
+      callAndVerify(host, port, "foo")
+
+      // Requesting the server to shutdown
+      callAndVerify(host, port, "shutdown")
+
+      Await.ready(server, Duration(1, TimeUnit.SECONDS))
+      server.isCompleted should ===(true)
+
+    }
+
+    "start providing an ActorSystem" in {
+      val (_, host, port) = TestUtils.temporaryServerHostnameAndPort()
+
+      val minimal = new MinimalHttpApp()
+
+      val server = Future {
+        minimal.startServer(host, port, ServerSettings.create(system), system)
+      }
+
+      callAndVerify(host, port, "foo")
+
+      // Requesting the server to shutdown
+      callAndVerify(host, port, "shutdown")
+
+      Await.ready(server, Duration(1, TimeUnit.SECONDS))
+      server.isCompleted should ===(true)
+      system.isTerminated should ===(false)
+    }
+
+    "provide binding if available" in {
+      val (_, host, port) = TestUtils.temporaryServerHostnameAndPort()
+
+      val minimal = new MinimalHttpApp()
+
+      intercept[IllegalStateException] {
+        minimal.binding()
+      }
+
+      val server = Future {
+        minimal.startServer(host, port, ServerSettings.create(ConfigFactory.load))
+      }
+
+      callAndVerify(host, port, "foo")
+
+      minimal.binding().localAddress.getPort should ===(port)
+      minimal.binding().localAddress.getHostName should ===(host)
+
+      // Requesting the server to shutdown
+      callAndVerify(host, port, "shutdown")
+      Await.ready(server, Duration(1, TimeUnit.SECONDS))
+      server.isCompleted should ===(true)
+    }
+
+    "let get notified" when {
+
+      "shutting down" in {
+        val (_, host, port) = TestUtils.temporaryServerHostnameAndPort()
+
+        val sneaky = new SneakHttpApp()
+
+        val server = Future {
+          sneaky.startServer(host, port, ServerSettings.create(ConfigFactory.load))
+        }
+
+        sneaky.postServerShutdownCalled.get() should ===(false)
+
+        // Requesting the server to shutdown
+        callAndVerify(host, port, "shutdown")
+        Await.ready(server, Duration(1, TimeUnit.SECONDS))
+        server.isCompleted should ===(true)
+        eventually {
+          sneaky.postServerShutdownCalled.get() should ===(true)
+        }
+      }
+
+      "after binding is successful" in {
+        val (_, host, port) = TestUtils.temporaryServerHostnameAndPort()
+
+        val sneaky = new SneakHttpApp()
+
+        val server = Future {
+          sneaky.startServer(host, port, ServerSettings.create(ConfigFactory.load))
+        }
+
+        callAndVerify(host, port, "foo")
+
+        sneaky.postBindingCalled.get() should ===(true)
+
+        // Requesting the server to shutdown
+        callAndVerify(host, port, "shutdown")
+        Await.ready(server, Duration(1, TimeUnit.SECONDS))
+        server.isCompleted should ===(true)
+      }
+
+      "after binding is unsuccessful" in {
+        val sneaky = new SneakHttpApp()
+
+        sneaky.startServer("localhost", 1, ServerSettings.create(ConfigFactory.load))
+
+        eventually {
+          sneaky.postBindingFailureCalled.get() should ===(true)
+        }
+      }
+
+    }
+
+  }
+
+  private def callAndVerify(host: String, port: Int, path: String) = {
+
+    val request = HttpRequest(uri = s"http://$host:$port/$path")
+
+    implicit val mat = ActorMaterializer()
+
+    eventually {
+      val response = Http().singleRequest(request)
+      response.futureValue.status should ===(StatusCodes.OK)
+    }
+  }
+
+}
--- a/akka-http-tests/src/test/scala/akka/http/scaladsl/server/HttpAppSpec.scala
+++ b/akka-http-tests/src/test/scala/akka/http/scaladsl/server/HttpAppSpec.scala
@@ -1,0 +1,207 @@
+/*
+ * Copyright (C) 2009-2017 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.http.scaladsl.server
+
+import java.util.concurrent.TimeUnit
+import java.util.concurrent.atomic.AtomicBoolean
+
+import akka.Done
+import akka.actor.ActorSystem
+import akka.http.scaladsl.Http.ServerBinding
+import akka.http.scaladsl.{ Http, TestUtils }
+import akka.http.scaladsl.client.RequestBuilding
+import akka.http.scaladsl.model.{ HttpRequest, StatusCodes }
+import akka.http.scaladsl.settings.ServerSettings
+import akka.stream.ActorMaterializer
+import akka.testkit.AkkaSpec
+import com.typesafe.config.ConfigFactory
+import org.scalatest.concurrent.Eventually
+
+import scala.concurrent.duration.Duration
+import scala.concurrent.{ Await, ExecutionContext, Future, Promise }
+import scala.util.Try
+
+class HttpAppSpec extends AkkaSpec with Directives with RequestBuilding with Eventually {
+  import system.dispatcher
+
+  class Minimal extends HttpApp {
+
+    private val shutdownPromise = Promise[Done]()
+
+    override protected def route: Route =
+      path("foo") {
+        complete("bar")
+      } ~
+        path("shutdown") {
+          if (shutdownPromise.isCompleted) complete("Shutdown already in process")
+          else {
+            shutdownPromise.success(Done)
+            complete("Shutdown request accepted")
+          }
+        }
+
+    override protected def waitForShutdownSignal(system: ActorSystem)(implicit ec: ExecutionContext): Future[Done] = {
+      shutdownPromise.future
+    }
+  }
+
+  "HttpApp" should {
+
+    "start without ActorSystem" in {
+      val (_, host, port) = TestUtils.temporaryServerHostnameAndPort()
+
+      val minimal = new Minimal()
+
+      val server = Future {
+        minimal.startServer(host, port, ServerSettings(ConfigFactory.load))
+      }
+
+      callAndVerify(host, port, "foo")
+
+      // Requesting the server to shutdown
+      callAndVerify(host, port, "shutdown")
+      Await.ready(server, Duration(1, TimeUnit.SECONDS))
+      server.isCompleted should ===(true)
+
+    }
+
+    "start providing an ActorSystem" in {
+      val (_, host, port) = TestUtils.temporaryServerHostnameAndPort()
+
+      val minimal = new Minimal()
+
+      val server = Future {
+        minimal.startServer(host, port, ServerSettings(system), system)
+      }
+
+      callAndVerify(host, port, "foo")
+
+      // Requesting the server to shutdown
+      callAndVerify(host, port, "shutdown")
+      Await.ready(server, Duration(1, TimeUnit.SECONDS))
+      server.isCompleted should ===(true)
+      system.isTerminated should ===(false)
+    }
+
+    "provide binding if available" in {
+      val (_, host, port) = TestUtils.temporaryServerHostnameAndPort()
+
+      val minimal = new Minimal()
+
+      minimal.binding().isFailure should ===(true)
+
+      val server = Future {
+        minimal.startServer(host, port, ServerSettings(ConfigFactory.load))
+      }
+
+      callAndVerify(host, port, "foo")
+
+      minimal.binding().isSuccess should ===(true)
+      minimal.binding().get.localAddress.getPort should ===(port)
+      minimal.binding().get.localAddress.getHostName should ===(host)
+
+      // Requesting the server to shutdown
+      callAndVerify(host, port, "shutdown")
+      Await.ready(server, Duration(1, TimeUnit.SECONDS))
+      server.isCompleted should ===(true)
+    }
+
+    "let get notified" when {
+      class SneakSever extends HttpApp {
+
+        private val shutdownPromise = Promise[Done]()
+
+        val postBindingCalled = new AtomicBoolean(false)
+        val postBindingFailureCalled = new AtomicBoolean(false)
+        val postShutdownCalled = new AtomicBoolean(false)
+
+        override protected def route: Route =
+          path("foo") {
+            complete("bar")
+          } ~
+            path("shutdown") {
+              if (shutdownPromise.isCompleted) complete("Shutdown already in process")
+              else {
+                shutdownPromise.success(Done)
+                complete("Shutdown request accepted")
+              }
+            }
+
+        override protected def waitForShutdownSignal(system: ActorSystem)(implicit ec: ExecutionContext): Future[Done] = {
+          shutdownPromise.future
+        }
+
+        override protected def postHttpBindingFailure(cause: Throwable): Unit = postBindingFailureCalled.set(true)
+
+        override protected def postHttpBinding(binding: ServerBinding): Unit = postBindingCalled.set(true)
+
+        override protected def postServerShutdown(attempt: Try[Done], system: ActorSystem): Unit = postShutdownCalled.set(true)
+      }
+
+      "shutting down" in {
+        val (_, host, port) = TestUtils.temporaryServerHostnameAndPort()
+
+        val sneaky = new SneakSever()
+
+        val server = Future {
+          sneaky.startServer(host, port, ServerSettings(ConfigFactory.load))
+        }
+
+        sneaky.postShutdownCalled.get() should ===(false)
+
+        // Requesting the server to shutdown
+        callAndVerify(host, port, "shutdown")
+        Await.ready(server, Duration(1, TimeUnit.SECONDS))
+        server.isCompleted should ===(true)
+        eventually {
+          sneaky.postShutdownCalled.get() should ===(true)
+        }
+      }
+
+      "after binding is successful" in {
+        val (_, host, port) = TestUtils.temporaryServerHostnameAndPort()
+
+        val sneaky = new SneakSever()
+
+        val server = Future {
+          sneaky.startServer(host, port, ServerSettings(ConfigFactory.load))
+        }
+
+        callAndVerify(host, port, "foo")
+
+        sneaky.postBindingCalled.get() should ===(true)
+
+        // Requesting the server to shutdown
+        callAndVerify(host, port, "shutdown")
+        Await.ready(server, Duration(1, TimeUnit.SECONDS))
+        server.isCompleted should ===(true)
+      }
+
+      "after binding is unsuccessful" in {
+        val sneaky = new SneakSever()
+
+        sneaky.startServer("localhost", 1, ServerSettings(ConfigFactory.load))
+
+        eventually {
+          sneaky.postBindingFailureCalled.get() should ===(true)
+        }
+      }
+
+    }
+
+  }
+
+  private def callAndVerify(host: String, port: Int, path: String) = {
+
+    val request = HttpRequest(uri = s"http://$host:$port/$path")
+
+    implicit val mat = ActorMaterializer()
+
+    eventually {
+      val response = Http().singleRequest(request)
+      response.futureValue.status should ===(StatusCodes.OK)
+    }
+  }
+}
--- a/akka-http/src/main/java/akka/http/javadsl/server/HttpApp.java
+++ b/akka-http/src/main/java/akka/http/javadsl/server/HttpApp.java
@@ -1,0 +1,155 @@
+/*
+ * Copyright (C) 2009-2017 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.http.javadsl.server;
+
+import akka.Done;
+import akka.actor.ActorSystem;
+import akka.event.Logging;
+import akka.http.javadsl.ConnectHttp;
+import akka.http.javadsl.Http;
+import akka.http.javadsl.ServerBinding;
+import akka.http.javadsl.settings.ServerSettings;
+import akka.stream.ActorMaterializer;
+
+import java.io.IOException;
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.atomic.AtomicReference;
+
+/**
+ * API MAY CHANGE - EXPERIMENTAL
+ * Bootstrap trait for Http Server. It helps booting up an akka-http server by only defining the desired routes.
+ * It offers additional hooks to modify the default behavior.
+ */
+// @akka.annotation.ApiMayChange // FIXME replace with real annotation once Akka dependency bumped
+public abstract class HttpApp extends AllDirectives {
+
+  private AtomicReference<ServerBinding> serverBinding = new AtomicReference<>();
+  /**
+   * Holds a reference to the {@link ActorSystem} used to start this server. Stopping this system will interfere with the proper functioning condition of the server.
+   */
+  protected AtomicReference<ActorSystem> systemReference = new AtomicReference<>();
+
+  /**
+   * Start a server on the specified host and port, using provided settings.
+   * Note that this method is blocking.
+   */
+  public void startServer(String host, int port, ServerSettings settings) throws ExecutionException, InterruptedException {
+    startServer(host, port, settings, Optional.empty());
+  }
+
+  /**
+   * Start a server on the specified host and port, using provided settings and [[ActorSystem]].
+   * Note that this method is blocking.
+   */
+  public void startServer(String host, int port, ServerSettings settings, ActorSystem system) throws ExecutionException, InterruptedException {
+    startServer(host, port, settings, Optional.of(system));
+  }
+
+  /**
+   * Start a server on the specified host and port, using provided settings and [[ActorSystem]] if present.
+   * Note that this method is blocking.
+   * This method may throw an {@link ExecutionException} or {@link InterruptedException} if the future that signals that
+   * the server should shutdown is interrupted or cancelled.
+   */
+  public void startServer(String host, int port, ServerSettings settings, Optional<ActorSystem> system) throws ExecutionException, InterruptedException {
+
+    final ActorSystem theSystem = system.orElseGet(() -> ActorSystem.create(Logging.simpleName(this).replaceAll("\\$", "")));
+    systemReference.set(theSystem);
+    final ActorMaterializer materializer = ActorMaterializer.create(theSystem);
+
+    CompletionStage<ServerBinding> bindingFuture = Http
+      .get(theSystem)
+      .bindAndHandle(route().flow(theSystem, materializer),
+        ConnectHttp.toHost(host, port),
+        settings,
+        theSystem.log(),
+        materializer);
+
+    bindingFuture.handle((binding, exception) -> {
+      if (exception != null) {
+        postHttpBindingFailure(exception);
+      } else {
+        //setting the server binding for possible future uses in the client
+        serverBinding.set(binding);
+        postHttpBinding(binding);
+      }
+      return null;
+    });
+
+    try {
+      bindingFuture
+        .thenCompose(ignore -> waitForShutdownSignal(theSystem)) // chaining both futures to fail fast
+        .toCompletableFuture()
+        .exceptionally(ignored -> Done.getInstance()) // If the future fails, we want to complete normally
+        .get(); // It's waiting forever because maybe there is never a shutdown signal
+    } finally {
+      bindingFuture.thenCompose(ServerBinding::unbind).handle( (success, exception) -> {
+        postServerShutdown(Optional.ofNullable(exception), theSystem);
+        if (!system.isPresent()) {
+          // we created the system. we should cleanup!
+          theSystem.terminate();
+        }
+        return null;
+      });
+    }
+
+  }
+
+  /**
+   * It tries to retrieve the {@link ServerBinding} if the server has been successfully started. It throws an {@link IllegalStateException} otherwise.
+   * You can use this method to attempt to retrieve the {@link ServerBinding} at any point in time to, for example, stop the server due to unexpected circumstances.
+   */
+  ServerBinding binding() {
+    if (serverBinding.get() == null)
+      throw new IllegalStateException("Binding not yet stored. Have you called startServer?");
+    return serverBinding.get();
+  }
+
+  /**
+   * Hook that will be called just after the server termination. Override this method if you want to perform some cleanup actions after the server is stopped.
+   * The {@code failure} parameter contains a {@link Throwable} only if there has been a problem shutting down the server.
+   */
+  protected void postServerShutdown(Optional<Throwable> failure, ActorSystem system) {
+    systemReference.get().log().info("Shutting down the server");
+  }
+
+  /**
+   * Hook that will be called just after the Http server binding is done. Override this method if you want to perform some actions after the server is up.
+   */
+  protected void postHttpBinding (ServerBinding binding) {
+    systemReference.get().log().info("Server online at http://" + binding.localAddress().getHostName() + ":" + binding.localAddress().getPort() + "/");
+  }
+
+  /**
+   * Hook that will be called in case the Http server binding fails. Override this method if you want to perform some actions after the server binding failed.
+   */
+  protected void postHttpBindingFailure(Throwable cause) {
+    systemReference.get().log().error(cause, "Error starting the server: " + cause.getMessage());
+  }
+
+  /**
+   * Hook that lets the user specify the future that will signal the shutdown of the server whenever completed.
+   */
+  protected CompletionStage<Done> waitForShutdownSignal (ActorSystem system) {
+    return CompletableFuture.supplyAsync( () -> {
+      System.out.println("Press RETURN to stop...");
+      try {
+        System.in.read();
+      } catch (IOException e) {
+        systemReference.get().log().error(e, "Problem occurred! " + e.getMessage());
+      }
+      return Done.getInstance();
+    });
+  }
+
+  /**
+   * Override to implement the route that will be served by this http server.
+   */
+  protected abstract Route route();
+
+
+}
--- a/akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
@@ -1,0 +1,136 @@
+/*
+ * Copyright (C) 2009-2017 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.http.scaladsl.server
+
+import java.util.concurrent.atomic.AtomicReference
+
+import akka.Done
+import akka.actor.ActorSystem
+import akka.event.Logging
+import akka.http.scaladsl.Http
+import akka.http.scaladsl.Http.ServerBinding
+import akka.stream.ActorMaterializer
+import akka.http.scaladsl.settings.ServerSettings
+
+import scala.concurrent.duration.Duration
+import scala.concurrent.{ Await, ExecutionContext, Future }
+import scala.io.StdIn
+import scala.util.{ Failure, Success, Try }
+
+/**
+ * API MAY CHANGE - EXPERIMENTAL
+ * Bootstrap trait for Http Server. It helps booting up an akka-http server by only defining the desired routes.
+ * It offers additional hooks to modify the default behavior.
+ */
+// @akka.annotation.ApiMayChange // FIXME replace with real annotation once Akka dependency bumped
+trait HttpApp {
+
+  private val serverBinding = new AtomicReference[ServerBinding]()
+  /**
+   * [[ActorSystem]] used to start this server. Stopping this system will interfere with the proper functioning condition of the server.
+   */
+  protected val systemReference = new AtomicReference[ActorSystem]()
+
+  /**
+   * Start a server on the specified host and port, using provided settings.
+   * Note that this method is blocking.
+   */
+  def startServer(host: String, port: Int, settings: ServerSettings): Unit = {
+    startServer(host, port, settings, None)
+  }
+
+  /**
+   * Start a server on the specified host and port, using provided settings and [[ActorSystem]].
+   * Note that this method is blocking.
+   */
+  def startServer(host: String, port: Int, settings: ServerSettings, system: ActorSystem): Unit = {
+    startServer(host, port, settings, Some(system))
+  }
+
+  /**
+   * Start a server on the specified host and port, using provided settings and [[ActorSystem]] if present.
+   * Note that this method is blocking.
+   */
+  def startServer(host: String, port: Int, settings: ServerSettings, system: Option[ActorSystem]): Unit = {
+    implicit val theSystem = system.getOrElse(ActorSystem(Logging.simpleName(this).replaceAll("\\$", "")))
+    systemReference.set(theSystem)
+    implicit val materializer = ActorMaterializer()
+    implicit val executionContext = theSystem.dispatcher
+
+    val bindingFuture = Http().bindAndHandle(
+      handler = route,
+      interface = host,
+      port = port,
+      settings = settings)
+
+    bindingFuture.onComplete {
+      case Success(binding) ⇒
+        //setting the server binding for possible future uses in the client
+        serverBinding.set(binding)
+        postHttpBinding(binding)
+      case Failure(cause) ⇒
+        postHttpBindingFailure(cause)
+    }
+
+    Await.ready(
+      bindingFuture.flatMap(_ ⇒ waitForShutdownSignal(theSystem)), // chaining both futures to fail fast
+      Duration.Inf) // It's waiting forever because maybe there is never a shutdown signal
+
+    bindingFuture
+      .flatMap(_.unbind())
+      .onComplete(attempt ⇒ {
+        postServerShutdown(attempt.map(_ ⇒ Done), theSystem)
+        // we created the system. we should cleanup!
+        if (system.isEmpty) theSystem.terminate()
+      })
+  }
+
+  /**
+   * It tries to retrieve the [[ServerBinding]] if the server has been successfully started. It fails otherwise.
+   * You can use this method to attempt to retrieve the [[ServerBinding]] at any point in time to, for example, stop the server due to unexpected circumstances.
+   */
+  def binding(): Try[ServerBinding] = {
+    if (serverBinding.get() == null) Failure(new IllegalStateException("Binding not yet stored. Have you called startServer?"))
+    else Success(serverBinding.get())
+  }
+
+  /**
+   * Hook that will be called just after the server termination. Override this method if you want to perform some cleanup actions after the server is stopped.
+   * The `attempt` parameter is represented with a [[Try]] type that is successful only if the server was successfully shut down.
+   */
+  protected def postServerShutdown(attempt: Try[Done], system: ActorSystem): Unit = {
+    systemReference.get().log.info("Shutting down the server")
+  }
+
+  /**
+   * Hook that will be called just after the Http server binding is done. Override this method if you want to perform some actions after the server is up.
+   */
+  protected def postHttpBinding(binding: Http.ServerBinding): Unit = {
+    systemReference.get().log.info(s"Server online at http://${binding.localAddress.getHostName}:${binding.localAddress.getPort}/")
+  }
+
+  /**
+   * Hook that will be called in case the Http server binding fails. Override this method if you want to perform some actions after the server binding failed.
+   */
+  protected def postHttpBindingFailure(cause: Throwable): Unit = {
+    systemReference.get().log.error(cause, s"Error starting the server ${cause.getMessage}")
+  }
+
+  /**
+   * Hook that lets the user specify the future that will signal the shutdown of the server whenever completed.
+   */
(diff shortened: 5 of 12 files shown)


akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
```diff
@@ -0,0 +1,78 @@
...
+
+  /**
+   * Your class should provide an implementation for the routes. These will be bound to the [[Http]]
+   */
+  val route: server.Route
+
+  /**
+   * Provides configuration values for the host, port and [[ServerSettings]] on which the server will run
+   */
+  val config = Config("localhost", 8080, None)
```
jonas: If the goal is to remove boilerplate, would it make sense to read these from a Typesafe Config or env variable so they can be overridden from the command line, e.g. `-Dakka.http.server.app.port=3000`?

jlprat: I was also thinking about reading it from there yes.
The only "problem" is that ha simple HelloWorld snippet that wants to run on a different port needs 2 files or explanation on ho to run it

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
```diff
@@ -0,0 +1,78 @@
...
+    postStartActorSystem(system)
+
+    val bindingFuture = Http().bindAndHandle(
+      handler = route,
+      interface = config.host,
+      port = config.port,
+      settings = config.serverSettings.getOrElse(ServerSettings(system)))
+
+    println(s"Server online at http://${config.host}:${config.port}/\nPress RETURN to stop...")
+    StdIn.readLine()
```
drewhk: Can you make this overridable? I imagine some implementations wanting to wait on a different signal.

jlprat: Sure! I was thinking about making overridable via a hook that would return a Future or something

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
```diff
@@ -0,0 +1,78 @@
...
+      .onComplete(_ ⇒ {
+        preTerminateActorSystem(system)
+        system.terminate()
+      })
+  }
+
+  /**
+   * Provides the [[ActorSystem]] from where the routes will be served. Override this method, if the default [[ActorSystem]] doesn't suit your needs. You don't need to store the created instance of the [[ActorSystem]].
+   */
+  def actorSystem(): ActorSystem = ActorSystem("my-system")
```
jlprat: True!

jlprat:
Let me know what you think about it.
I renamed some hooks names so it's clearer what they do.
I added a termination hook (so now the user can decide how and when the server should be terminated)
Also refactored the Config providing, now using a proper Config, turns out it's quite straight-forward to provide custom values

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
```diff
@@ -0,0 +1,106 @@
...
+  val route: Route
+
+  /**
+   * It loads the config provided by `configProvider` and add the needed defaults just in case they are not present
+   */
+  private def retrieveConfig(): Config = {
+    val config = createConfig()
+    val defaults = ConfigFactory.parseString(
+      """
+        |akka.http.server{
```
jypma: consider just having these in `reference.conf`

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
```diff
@@ -0,0 +1,106 @@
...
+    implicit val system = createActorSystem()
+    implicit val materializer = ActorMaterializer()
+    implicit val executionContext = system.dispatcher
+    postStartActorSystem(system)
+
+    val config = retrieveConfig()
+    val host = config.getString("akka.http.server.host")
+    val port = config.getInt("akka.http.server.port")
+
+    val bindingFuture = Http().bindAndHandle(
```
jypma: does it make sense to already model for multiple (port,settings,route) combinations? In my applications I often have both an HTTP _and_ an HTTPS server, serving different routes.

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
jlprat: It's here to ensure that some values are really provided and to keep the PoC simple. If this gets over the PoC phase, yes, they should be in there

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
jlprat: For the port, the intention behind it is that if you just want to boot up a server, but 8080 is already taken by any other server running, it should be easy to specify an alternate port using `HttpApp`.

jlprat:
@jypma I'll add also the Java counterpart following the same structure as this one. Once this gets "green light".
About the command line args idea, sounds really cool. I'll hack something :+1:

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
```diff
@@ -0,0 +1,106 @@
...
+      handler = route,
+      interface = host,
+      port = port,
+      settings = ServerSettings(config))
+
+    println(s"Server online at http://$host:$port/")
```
hseeberger: If binding fails, this is not true! You better do the following:
``` scala
Http().bindAndHandle(...).onComplete {
  case Success(address) =>
    log.info(...) // not println!
  case Failure(cause) =>
    log.error(...)
    // terminate the system
}
```

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
```diff
@@ -0,0 +1,106 @@
...
+      .onComplete(_ ⇒ {
+        preTerminateActorSystem(system)
+        system.terminate()
+      })
+  }
+
+  /**
+   * Provides the [[ActorSystem]] from where the routes will be served. Override this method, if the default [[ActorSystem]] doesn't suit your needs. You don't need to store the created instance of the [[ActorSystem]].
+   */
+  def createActorSystem(): ActorSystem = ActorSystem("my-system")
```
hseeberger: Use `protected`

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
```diff
@@ -0,0 +1,106 @@
...
+
+  /**
+   * Hook that will be called once the actor system is created. Override this method if you want to do something with this [[ActorSystem]] like creating actors.
+   */
+  def postStartActorSystem(actorSystem: ActorSystem): Unit = {
```
hseeberger: Use `protected`.

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  /**
+   * Hook that will be called just before the termination of the actor system. Override this methods if you want to perform some actions on the current actorSystem, like gracefully shutting down any created actor.
+   */
+  def preTerminateActorSystem(actorSystem: ActorSystem): Unit = {
```
hseeberger: Use `protected`.

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
```diff
@@ -0,0 +1,106 @@
...
+    Future {
+      println("Press RETURN to stop...")
+      StdIn.readLine()
+    }
+  }
+
+  /**
+   * Your class should provide an implementation for the routes. These will be bound to the [[Http]]
+   */
+  val route: Route
```
hseeberger: Use a `def`, because abstract `val`s are evil.

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
```diff
@@ -0,0 +1,106 @@
...
+        | port = 8080
+        |}
+      """.stripMargin)
+    config.withFallback(defaults)
+  }
+
+  /**
+   * Enables the user the possibility to override the `Config` that should be used for the application. By default it loads the config in the path
+   */
+  def createConfig(): Config = {
```
hseeberger: Use `protected`.

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
```diff
@@ -0,0 +1,106 @@
...
+      port = port,
+      settings = ServerSettings(config))
+
+    println(s"Server online at http://$host:$port/")
+
+    //It's waiting forever because maybe there is never a shutdown signal
+    Await.result(waitForShutdownSignal(system), Duration.Inf)
+
+    bindingFuture
+      .flatMap(_.unbind())
```
hseeberger: Is this necessary at all?

hseeberger: OK, now I see ;-)

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
hseeberger: Don't use a fancy name ("my-system"), just use `ActorSystem()`.

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
hseeberger: s/actorSystem/system

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
hseeberger: s/actorSystem/system

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
```diff
@@ -0,0 +1,106 @@
...
+
+  /**
+   * Your class should provide an implementation for the routes. These will be bound to the [[Http]]
+   */
+  val route: Route
+
+  /**
+   * It loads the config provided by `configProvider` and add the needed defaults just in case they are not present
+   */
+  private def retrieveConfig(): Config = {
```
hseeberger: No type annotation for private method necessary.

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
hseeberger: Please move these to the `reference.conf`! This will certainly also ensure that these settings are defined.

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
```diff
@@ -0,0 +1,106 @@
...
+        |}
+      """.stripMargin)
+    config.withFallback(defaults)
+  }
+
+  /**
+   * Enables the user the possibility to override the `Config` that should be used for the application. By default it loads the config in the path
+   */
+  def createConfig(): Config = {
+    ConfigFactory.load()
```
analytically: Why not system.settings.config?

hseeberger: Oops! Here's a severe bug! Well, it's up there where the actor system is created ... I'll add a comment up there.

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
```diff
@@ -0,0 +1,106 @@
...
+   */
+  def preTerminateActorSystem(actorSystem: ActorSystem): Unit = {
+  }
+
+  /**
+   * Hook that lets the user specify the future that will signal the shutdown of the server whenever completed
+   */
+  def waitForShutdownSignal(actorSystem: ActorSystem)(implicit executionContext: ExecutionContext): Future[Unit] = {
+    Future {
+      println("Press RETURN to stop...")
```
analytically: Press any key to exit ... ?

hseeberger: - `protected`
- s/executionContext/ec

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
```diff
@@ -0,0 +1,106 @@
...
+      StdIn.readLine()
```
hseeberger: Move text to print into `readLine("...")`.

hseeberger:
After thinking about this for a while I have arrived at the conclusion that I don't like all these hooks. No realy world Akka HTTP application I have seen has these. Instead thinks which are supposed to be done in these hooks are done inline.

I suggest to get rid of the hooks and instead add comments like `// Code to be executed after creating the actor system goes here` at the respective locations.

jlprat:
I incorporated some of the feedback: 
* protected methods
* rename parameters 
* reorder creation of config and system
* use of reference.conf
* cover case when binding fails

 Still using prints, not happy with termination of server and termination signal handling.

jlprat:
@hseeberger Thank you very much for your feedback! Really appreciated.
I don't pretend to have this class used in real world applications, but in PoC, tutorials, Hello World examples and similar areas.
Why I have chosen hooks is because I wanted to offer the quick shot experience where only the `route` has to be implemented, but at the same time, only if the user really wants (or needs), change some behavior by overriding hooks.

hseeberger:
We agree on the usage scenarios which are cases where the approach has to be easy to understand and modify. Don't you think that hooks make the solution too complex? Why not simply inline these defaults. Tweaking would then also happen inline. Another advantage is that any tweaking would be possible, not just what's offered by the hooks.

jlprat:
I will try to come up with a different approach, but as a user of the `HttpApp` I only need to define `route` and it works, if I need something more, I can override a hook. i also converted some of the docs examples to `HttpApp` and looked super easy to me (of course I'm biased).
Let's see if I can come up with something better.

hseeberger:
Defining route is fine and it's really nice if that's all you need. Just get rid of the hooks, they are neither easy to use nor flexible.

jlprat:
Then I'm going to a simplified version of my first draft, and for that I got feedback to include more hooks so people could redefine behavior ;)

akka-http/src/main/resources/reference.conf
```diff
@@ -4,40 +4,52 @@
...
-  # for an entity data stream.
-  decode-max-bytes-per-chunk = 1m
-
-  # Fully qualified config path which holds the dispatcher configuration
-  # to be used by FlowMaterialiser when creating Actors for IO operations.
-  file-io-dispatcher = ${akka.stream.blocking-io-dispatcher}
-}
+akka.http {
+
+  # Default settings for the HtppApp bootstrap
```
ktoso: typo: `HttpApp`

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
```diff
@@ -0,0 +1,98 @@
...
+   * Your class should provide an implementation for the routes. These will be bound to the [[Http]]
+   */
+  protected def route: Route
+
+  /**
+   * Enables the user the possibility to override the `Config` that should be used for the application. By default it loads the config in the path
+   */
+  protected def createConfig(): Config = {
+    ConfigFactory.load()
+  }
```
ktoso: good

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
```diff
@@ -0,0 +1,98 @@
...
+      .onComplete(_ ⇒ {
+        preTerminateActorSystem(system)
+        system.terminate()
+      })
+  }
+
+  /**
+   * Provides the [[ActorSystem]] from where the routes will be served. Override this method, if the default [[ActorSystem]] doesn't suit your needs. You don't need to store the created instance of the [[ActorSystem]].
+   */
+  protected def createActorSystem(config: Config): ActorSystem = ActorSystem()
```
ktoso: shouldn't this be passing in the config?
Give the Actor System a name - via `Logging.simpleName(getClass)`?

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  /**
+   * Hook that lets the user specify the future that will signal the shutdown of the server whenever completed
+   */
+  protected def waitForShutdownSignal(system: ActorSystem)(implicit ec: ExecutionContext): Future[Unit] = {
+    Future {
+      StdIn.readLine("Press RETURN to stop...\n")
+    }
+  }
```
ktoso: Yeah good API, could be implemented by external trigger completing the future... good.

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
```diff
@@ -0,0 +1,98 @@
...
+import scala.concurrent.duration.Duration
+import scala.concurrent.{ Await, ExecutionContext, Future }
+import scala.io.StdIn
+import scala.util.{ Failure, Success }
+
+/**
+ * Bootstrap trait. It helps booting up an akka-http by only defining the desired routes.
+ * It offers additional hooks to modify the default behavior
+ */
+trait HttpApp {
```
ktoso: Just a note: This will need JavaDSL

ktoso:
> Defining route is fine and it's really nice if that's all you need. Just get rid of the hooks, they are neither easy to use nor flexible.

I'm not sure entirely dropping them is good. They're still managable in the current proposal I think, we'd just document in 3 steps: 
1) "i don't care, just override routes" 
2) "ok, need s little bit of tweating, starting some other things when this starts, and already bound"
3) "ok, show me the full gory details" (current APIs)

ktoso:
I think this is shaping up well. 
I'd like to invite @lihaoyi for a quick review, since this revival of this trait came out of a series of tweets you did - please chime in if this seems useful enough if you have a second to spare, thanks!

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
```diff
@@ -0,0 +1,98 @@
...
+
+  def main(args: Array[String]) = {
+    val config = createConfig()
+    implicit val system = createActorSystem(config)
+    implicit val materializer = ActorMaterializer()
+    implicit val executionContext = system.dispatcher
+    postStartActorSystem(system)
+
+    val host = config.getString("akka.http.server.host")
+    val port = config.getInt("akka.http.server.port")
```
lihaoyi: My personal preference would be for this not be a JVM-`main(args: Array[String])` method, but instead some kind of 
```scala
def startServer(host: String, port: Int, settings: ServerSettings, system: ActorSystem = null)
```

Having it take in the `system` as an optional argument-with-default would also make this more useful for people who want to start up servers in an existing Akka project. Also, I often like converting config into to type-safe settings objects as soon as possible rather than waiting for some library to try (and possibly fail) on my behalf.

Letting people deal with the `ActorSystem` externally will also remove the need for the hooks related to it, since they would already have the option of instantiating their own system and running whatever code they wish before or after.

Generally I think people will want more flexibility when starting the server: e.g. setting the port based on some dynamically-loaded configuration, or running multiple servers on different ports in the same JVM. Even for my current tiny project, i have a bunch of custom initialization I need to do before starting the server, and I assume it'll only grow as projects get larger. There is no inherent reason why starting the server should be tied to starting the JVM process, and we can provide the helper-ness of dealign with all the boilerplate while still providing the flexibility to start servers where-ever someone wants to

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
```diff
@@ -0,0 +1,98 @@
...
+
+    val bindingFuture = Http().bindAndHandle(
+      handler = route,
+      interface = host,
+      port = port,
+      settings = ServerSettings(config))
+
+    bindingFuture.onComplete {
+      case Success(_) ⇒
+        println(s"Server online at http://$host:$port/")
```
lihaoyi: These two should probably be moved into hooks so they can be overriden; someone somewhere is going to want a different error logger (more verbose? logback?) or difference success banner

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
jlprat: Yes, for sure! I wanted to start with the Scala one and then either make it work for Java or create a dedicated class for it

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
jlprat: Sounds reasonable :)

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
ktoso: Good input @lihaoyi, thanks a lot! :)

FYI:

> Also, I often like converting config into to type-safe settings objects as soon as possible rather than waiting for some library to try (and possibly fail) on my behalf.

FYI, we're working on something that'll allow more programatic and typesafe configuration: https://github.com/akka/akka/pull/21922

ktoso:
Seems we're almost there :)

jlprat:
I'll try to find some time to work on this one soon :smile:

jlprat:
Next round of feedback applied!

jlprat:
Ooops, reverting this back, I rebased the wrong branch. I will fix it ASAP

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
```diff
@@ -0,0 +1,94 @@
...
+/**
+ * Bootstrap trait. It helps booting up an akka-http by only defining the desired routes.
+ * It offers additional hooks to modify the default behavior
+ */
+trait HttpApp {
+
+  /**
+   * Start a server on the specified host and port, using provided settings and [[ActorSystem]] if present
+   */
+  def startServer(host: String, port: Int, settings: ServerSettings, system: ActorSystem = null): Unit = {
```
jlprat: I used null instead of a default initialization here, because I want to know when it's not provided, so I can terminate the actor system in that case

ktoso:
Re-reviewing!

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
```diff
@@ -0,0 +1,94 @@
...
+import scala.concurrent.duration.Duration
+import scala.concurrent.{ Await, ExecutionContext, Future }
+import scala.io.StdIn
+import scala.util.{ Failure, Success, Try }
+
+/**
+ * Bootstrap trait. It helps booting up an akka-http by only defining the desired routes.
+ * It offers additional hooks to modify the default behavior
+ */
+trait HttpApp {
```
ktoso: Please add here ( https://github.com/akka/akka/pull/22110 )

`@ApiMayChange` let's have people try it for one or two releases.
We need to dance around it so:

```
// @akka.annotation.ApiMayChange // FIXME replace with real annotation once Akka dependency bumped
/** API MAY CHANGE - EXPERIMENTAL 
... rest of docs
*/
```

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
ktoso: default params are bad for binary compatibility. I'd rather add an overloaded method:

```scala
def startServer(host: String, port: Int, settings: ServerSettings, system: Option[ActorSystem]): Unit = 
def startServer(host: String, port: Int, settings: ServerSettings): Unit = 
  startServer(host, port, settings, None)

```

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
jlprat: Sure!

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
ktoso: If we really want we can provide

```
def startServer(host: String, port: Int, settings: ServerSettings, system: ActorSystem): Unit = 
  startServer(host, port, settings, Option(system))

```

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
```diff
@@ -0,0 +1,94 @@
...
+   * Start a server on the specified host and port, using provided settings and [[ActorSystem]] if present
+   */
+  def startServer(host: String, port: Int, settings: ServerSettings, system: ActorSystem = null): Unit = {
+    // Having a null is suboptimal, but [[Option]] might be too verbose?.
+    // Having null or [[Option]] is also useful to shutdown the system if we created it
+    implicit val theSystem = if (system == null) ActorSystem(getClass.getSimpleName) else system
+    implicit val materializer = ActorMaterializer()
+    implicit val executionContext = system.dispatcher
+
+    val bindingFuture = Http().bindAndHandle(
```
ktoso: binding must be stored somewhere - as we want to be able to cause a shutdown.

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
jlprat: Funny, yesterday I was thinking about the exact same approach, I wasn't happy with the default parameter and  I didn't like `Option[ActorSystem]`

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
```diff
@@ -0,0 +1,94 @@
...
+      handler = route,
+      interface = host,
+      port = port,
+      settings = settings)
+
+    bindingFuture.onComplete {
+      case Success(binding) ⇒
+        postHttpBinding(binding)
```
ktoso: hm i see... you pass it around but I'd also store it such that user can call `binding().unbind()`
`binding()` should throw if it wasn't initialized yet (use an AtomicRef to store it) "binding not yet stored, have you called startServer?"

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
```diff
@@ -0,0 +1,94 @@
...
+    Await.result(
+      bindingFuture.flatMap(_ ⇒ waitForShutdownSignal(system)), // chaining both futures to fail fast
+      Duration.Inf) // It's waiting forever because maybe there is never a shutdown signal
+
+    bindingFuture
+      .flatMap(_.unbind())
+      .onComplete(attempt ⇒ {
+        postServerShutdown(attempt, system)
+        // we created the system. we should cleanup!
+        if (system == null) system.terminate()
```
ktoso: good

ktoso:
Had a final round of comments, let's address those, add javadsl and docs and merge :) 
This is another example of an API we'll want to release as MayChange

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
jlprat: Calling `unbind()` will stop the server, but the `Await` for the `waitForShutdownSignal` won't be completed, so it will still wait for that `Future` completion

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
```diff
@@ -0,0 +1,94 @@
...
+
+  /**
+   * Start a server on the specified host and port, using provided settings and [[ActorSystem]] if present
+   */
+  def startServer(host: String, port: Int, settings: ServerSettings, system: ActorSystem = null): Unit = {
+    // Having a null is suboptimal, but [[Option]] might be too verbose?.
+    // Having null or [[Option]] is also useful to shutdown the system if we created it
+    implicit val theSystem = if (system == null) ActorSystem(getClass.getSimpleName) else system
+    implicit val materializer = ActorMaterializer()
+    implicit val executionContext = system.dispatcher
```
matsluni: Shouldn't this be `theSystem`here? The given `system` might be null, right?

ktoso: Good catch!

jlprat: Good catch!!!

ktoso: One more reason to stick to Option :) https://github.com/akka/akka-http/pull/542#discussion_r94954359

jlprat: I know, I know...

matsluni: There some more places where this mixup is also happening, see `postServerShutdown` and `system.terminate`.

jlprat: `theSystem` was introduced later, and I forgot to refactor in some places

jlprat:
Status update:
* Last feedback for Scala API is done, but not pushed
* Java API is done, but not pushed
* Examples for Scala API are done
* Examples for Java API are in progress
* Tests for Scala and Java API are missing

Let me know if you want to have some of these in a separate PR

docs/src/test/scala/docs/http/scaladsl/HttpAppExampleSpec.scala
```diff
@@ -56,16 +56,25 @@ class HttpAppExampleSpec extends WordSpec with Matchers
           }
         }
     }
-    //#minimal-routing-example
+    val system = ActorSystem("example2")
+    WebServer.startServer("localhost", 80, ServerSettings(system), system)
+    //#failed-binding-example
```
ktoso: very fun few-liner, will be a nice hello world experience 👍

akka-http-tests/src/test/scala/akka/http/scaladsl/server/HttpAppSpec.scala
```diff
@@ -0,0 +1,185 @@
...
+
+  }
+
+  private def verifyRequest(host: String, port: Int) = {
+
+    val request = HttpRequest(uri = s"http://$host:$port/foo")
+
+    implicit val mat = ActorMaterializer()
+
+    eventually {
```
jlprat: Using it to ensure the server is correctly up and running. Without it, I got some test errors due to races

akka-http/src/main/java/akka/http/javadsl/server/HttpApp.java
```diff
@@ -0,0 +1,150 @@
...
+    if (serverBinding.get() == null)
+      throw new IllegalStateException("Binding not yet stored. Have you called startServer?");
+    return serverBinding.get();
+  }
+
+  /**
+   * Hook that will be called just after the server termination. Override this method if you want to perform some cleanup actions after the server is stopped
+   * When problems shutting down {@code possibleFailure} contains the failure occurred. It is null otherwise
+   */
+  protected void postServerShutdown(Throwable possibleFailure, ActorSystem system) {
```
jlprat: In Scala world this `possibleFailure` would be and `Option` but Java discourages the user of `Optional` in parameters and only recommending it for return types

akka-http/src/main/java/akka/http/javadsl/server/HttpApp.java
```diff
@@ -0,0 +1,150 @@
...
+      }
+      return null;
+    });
+
+    try {
+      bindingFuture
+        .thenCompose(ignore -> waitForShutdownSignal(theSystem)) // chaining both futures to fail fast
+        .toCompletableFuture()
+        .exceptionally(ignored -> BoxedUnit.UNIT) // If the future fails, we want to complete normally
+        .get(); // It's waiting forever because maybe there is never a shutdown signal
```
jlprat: Combination of exceptionally and get to wait for the future result no matter if successful or not

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
```diff
@@ -0,0 +1,128 @@
...
+    bindingFuture.onComplete {
+      case Success(binding) ⇒
+        //setting the server binding for possible future uses in the client
+        serverBinding.set(binding)
+        postHttpBinding(binding)
+      case Failure(cause) ⇒
+        postHttpBindingFailure(cause)
+    }
+
+    Await.ready(
```
jlprat: Using ready instead of result because we want to stop waiting in case of any `Future` completing with a failure

jlprat:
PLS BUILD

jlprat:
This PR is not WIP anymore

docs/src/main/paradox/java/http/routing-dsl/httpApp.md
```diff
@@ -0,0 +1,69 @@
...
+@@@ warning { title="API may change" }
+This is still experimental and the API is subjected to change in futures releases of Akka HTTP.
+@@@
+
+@@toc { depth=1 }
+
+## Introduction
+
+The objective of `HttpApp` is to reduce the boilerplate needed to start a server and handle its shutdown in a proper way.
+The current version of it, enables to create a new server just by extending `HttpApp` and implementing the `route()` method.
```
ktoso: I think I'd rather explain that this is a class to quickly get started, not talk about boilerplate at all, since with that "boilerplate" you get way more control and IMO in many prod apps you should be doing that (the more explicit way) anyway.

docs/src/main/paradox/java/http/routing-dsl/index.md
```diff
@@ -20,9 +20,29 @@ To use the high-level API you need to add a dependency to the `akka-http` module
...
 
+## Minimal Example
+
+This is a complete, very basic Akka HTTP application relying on the Routing DSL:
+
+@@snip [HttpServerMinimalExampleTest.java](../../../../../test/java/docs/http/javadsl/HttpServerMinimalExampleTest.java) { #minimal-routing-example }
+
+It starts an HTTP Server on localhost and replies to GET requests to `/hello` with a simple response.
+
+@@@ warning { title="API may change" }
```
ktoso: Please link to https://github.com/akka/akka/blob/master/akka-docs/rst/common/binary-compatibility-rules.rst#the-donotinherit-and-apimaychange-markers in the docs (thats akka docs, so direct link to /current/ will work well here IMO) so people can read up what this means.

docs/src/main/paradox/java/http/routing-dsl/httpApp.md
```diff
@@ -0,0 +1,69 @@
+<a id="http-app"></a>
+# HttpApp Bootstrap
+
+@@@ warning { title="API may change" }
+This is still experimental and the API is subjected to change in futures releases of Akka HTTP.
```
ktoso: Please link to https://github.com/akka/akka/blob/master/akka-docs/rst/common/binary-compatibility-rules.rst#the-donotinherit-and-apimaychange-markers in the docs (thats akka docs, so direct link to /current/ will work well here IMO) so people can read up what this means.

akka-http-tests/src/test/java/akka/http/javadsl/server/MinimalHttpApp.java
```diff
@@ -0,0 +1,28 @@
...
+
+import akka.actor.ActorSystem;
+import scala.runtime.BoxedUnit;
+
+import java.util.concurrent.CompletionStage;
+import java.util.concurrent.CompletableFuture;
+
+public class MinimalHttpApp extends HttpApp {
+
+  CompletableFuture<BoxedUnit> shutdownTrigger = new CompletableFuture<>();
```
ktoso: show in the app that you trigger this somehow, add a route that does `/shutdown` and triggers this.
Otherwise some people will certainly be confused ;)

akka-http/src/main/java/akka/http/javadsl/server/HttpApp.java
ktoso: there's an overload with a timeout rihght? Please use that instead, we never want to show get() without timeouts

akka-http/src/main/java/akka/http/javadsl/server/HttpApp.java
ktoso: I personally would not really care about that java guideline... We want APIs to be closely mirroring one another

akka-http/src/main/java/akka/http/javadsl/server/HttpApp.java
```diff
@@ -0,0 +1,150 @@
...
+      } catch (IOException e) {
+        System.out.println("Problem occurred! " + e.getMessage());
+        e.printStackTrace(System.err);
+      }
+      return BoxedUnit.UNIT;
+    });
+  }
+
+  /**
+   * Your class should provide an implementation for the routes. These will be bound to the [[Http]]
```
ktoso: Try to avoid "you" in scaladoc. Say "Override to implement route that will be served by this http server.

akka-http/src/main/java/akka/http/javadsl/server/HttpApp.java
```diff
@@ -0,0 +1,150 @@
...
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.atomic.AtomicReference;
+
+/**
+ * API MAY CHANGE - EXPERIMENTAL
+ * Bootstrap trait. It helps booting up an akka-http by only defining the desired routes.
+ * It offers additional hooks to modify the default behavior
```
ktoso: missing `.`

akka-http/src/main/java/akka/http/javadsl/server/HttpApp.java
```diff
@@ -0,0 +1,150 @@
...
+ */
+// @akka.annotation.ApiMayChange // FIXME replace with real annotation once Akka dependency bumped
+public abstract class HttpApp extends AllDirectives {
+
+  private AtomicReference<ServerBinding> serverBinding = new AtomicReference<ServerBinding>();
+
+  /**
+   * Start a server on the specified host and port, using provided settings
+   * Note that this method is blocking
```
ktoso: Explain that it may throw. Missing `.`

akka-http/src/main/java/akka/http/javadsl/server/HttpApp.java
```diff
@@ -0,0 +1,150 @@
...
+   */
+  protected void postServerShutdown(Throwable possibleFailure, ActorSystem system) {
+    System.out.println("Shutting down the server");
+  }
+
+  /**
+   * Hook that will be called just after the Http server binding is done. Override this method if you want to perform some actions after the server is up
+   */
+  protected void postHttpBinding (ServerBinding binding) {
+    System.out.println("Server online at http://" + binding.localAddress().getHostName() + ":" + binding.localAddress().getPort() + "/");
```
ktoso: I'd be tempted to put the system in a value in this class then these logs can be made to system.log

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
```diff
@@ -0,0 +1,128 @@
...
+   */
+  def binding(): Try[ServerBinding] = {
+    if (serverBinding.get() == null) Failure(new IllegalStateException("Binding not yet stored. Have you called startServer?"))
+    else Success(serverBinding.get())
+  }
+
+  /**
+   * Hook that will be called just after the server termination. Override this method if you want to perform some cleanup actions after the server is stopped
+   */
+  protected def postServerShutdown(attempt: Try[Unit], system: ActorSystem): Unit = {
```
ktoso: attempt's type is not nice. Why unit. Should be Try[Done] right?

ktoso: Explain what the param means too

ktoso:
Some cleanups in added content need to be fixed, but then this is ready to go

ktoso:
Misclick

jlprat:
I will address the feedback later today

akka-http/src/main/java/akka/http/javadsl/server/HttpApp.java
jlprat: Following the same structure as `binding`, it could make sense.

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
jlprat: If I'm not mistaken, `unbind` returns a `Future[Unit]` [ServerBinding#unbind](http://doc.akka.io/api/akka-http/current/akka/http/scaladsl/Http$$ServerBinding.html#unbind():scala.concurrent.Future[Unit])

akka-http/src/main/java/akka/http/javadsl/server/HttpApp.java
jlprat: Regarding using `get` without a timeout, what if the server should run forever. The timeout we specify here will limit the longest time to live for this server.

akka-http-tests/src/test/scala/akka/http/javadsl/server/HttpAppSpec.scala
```diff
@@ -0,0 +1,148 @@
...
+
+      val minimal = new MinimalHttpApp()
+
+      val server = Future {
+        minimal.startServer(host, port, ServerSettings.create(ConfigFactory.load))
+      }
+
+      verifyRequest(host, port)
+
+      minimal.shutdownTrigger.complete(BoxedUnit.UNIT)
```
jlprat: >show in the app that you trigger this somehow, add a route that does /shutdown and triggers this.
>Otherwise some people will certainly be confused ;)
Here is how we trigger the shutdown. In this case, we decide to shutdown the server from the "outside". I can add some comments in here. Is that what you mean?

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
ktoso: Yeah and that's a mistake that we slipped through when making it stable sadly.
Let's use the proper types whenever we can, so Done here

docs/src/main/paradox/java/http/routing-dsl/httpApp.md
patriknw: Good. I'd remove the word "still"

akka-http/src/main/java/akka/http/javadsl/server/HttpApp.java
```diff
@@ -0,0 +1,155 @@
...
+      });
+    }
+
+  }
+
+  /**
+   * It tries to retrieve the {@link ServerBinding} if the server has been successfully started. It throws an {@link IllegalStateException} otherwise.
+   * You can use this method to attempt to retrieve the {@link ServerBinding} at any point in time to, for example, stop the server due to unexpected circumstances.
+   */
+  ServerBinding binding() {
```
jonas: Did you consider returning `Optional<ServerBinding>` here to avoid the throw?

docs/src/main/paradox/scala/http/routing-dsl/index.md
```diff
@@ -36,6 +37,18 @@ This is a complete, very basic Akka HTTP application relying on the Routing DSL:
 
 It starts an HTTP Server on localhost and replies to GET requests to `/hello` with a simple response.
 
+@@@ warning { title="API may change" }
+The following code example uses an experimental feature and its API is subjected to change in futures releases of Akka HTTP.
+For further information about this marker, see [The @DoNotInherit and @ApiMayChange markers](http://doc.akka.io/docs/akka/current/common/binary-compatibility-rules.html#The_@DoNotInherit_and_@ApiMayChange_markers)
```
jonas: Using `@extref:[The @DoNotInherit and @ApiMayChange markers](akka-docs:common/binary-compatibility-rules.html#The_@DoNotInherit_and_@ApiMayChange_markers)` will make sure we link to the version of Akka that Akka HTTP depends on. Not that this is likely to change, but still.

docs/src/main/paradox/scala/http/routing-dsl/index.md
```diff
@@ -36,6 +37,18 @@ This is a complete, very basic Akka HTTP application relying on the Routing DSL:
 
 It starts an HTTP Server on localhost and replies to GET requests to `/hello` with a simple response.
 
+@@@ warning { title="API may change" }
+The following code example uses an experimental feature and its API is subjected to change in futures releases of Akka HTTP.
```
jonas: s/futures/future/

docs/src/main/paradox/scala/http/routing-dsl/index.md
```diff
@@ -36,6 +37,18 @@ This is a complete, very basic Akka HTTP application relying on the Routing DSL:
...
+For further information about this marker, see [The @DoNotInherit and @ApiMayChange markers](http://doc.akka.io/docs/akka/current/common/binary-compatibility-rules.html#The_@DoNotInherit_and_@ApiMayChange_markers)
+in the Akka documentation.
+@@@
+
+In order to help getting started with setting up a server Akka HTTP comes with an experimental helper class called `HttpApp`.
```
jonas: Nitpicking: `To help start a server Akka HTTP provides an experimental helper class called `HttpApp`.

docs/src/main/paradox/scala/http/routing-dsl/index.md
```diff
@@ -36,6 +37,18 @@ This is a complete, very basic Akka HTTP application relying on the Routing DSL:
...
+This is the same example as before rewritten using `HttpApp`:
+
+@@snip [HttpAppExampleSpec.scala](../../../../../test/scala/docs/http/scaladsl/HttpAppExampleSpec.scala) { #minimal-routing-example }
+
```
jonas: Maybe add a link to the HttpApp docs here.

```
See [HttpApp Bootstrap](...) for more details about setting up a server using this approach.
```

docs/src/main/paradox/scala/http/routing-dsl/index.md
```diff
@@ -25,6 +25,7 @@ from a background with non-"streaming first" HTTP Servers.
 * [case-class-extraction](case-class-extraction.md)
 * [source-streaming-support](source-streaming-support.md)
 * [testkit](testkit.md)
+* [http-app](httpApp.md)
```
jonas: Nitpicking: Using upper case letters in the doc file name

docs/src/main/paradox/scala/http/routing-dsl/httpApp.md
```diff
@@ -0,0 +1,71 @@
+<a id="http-app"></a>
+# HttpApp Bootstrap
+
+@@@ warning { title="API may change" }
+This is experimental and the API is subjected to change in futures releases of Akka HTTP.
```
jonas: s/futures/future/

docs/src/main/paradox/scala/http/routing-dsl/httpApp.md
```diff
@@ -0,0 +1,71 @@
...
+For further information about this marker, see [The @DoNotInherit and @ApiMayChange markers](http://doc.akka.io/docs/akka/current/common/binary-compatibility-rules.html#The_@DoNotInherit_and_@ApiMayChange_markers)
+in the Akka documentation.
+@@@
+
+@@toc { depth=1 }
+
+## Introduction
+
+The objective of `HttpApp` is to quickly get you started starting a Http Server with just a few lines of code.
```
jonas: Nitpicking to remove `started starting`: ... is to help you start an HTTP server with just ...

docs/src/main/paradox/scala/http/routing-dsl/httpApp.md
```diff
@@ -0,0 +1,71 @@
...
+This is accomplished just by extending `HttpApp` and implementing the `route()` method.
+If desired, `HttpApp` provides different hook methods that can be overridden to change its default behavior.
+
+## Minimal Example
+
+The following example shows how succinct is to start a server:
```
jonas: Something is missing here, maybe `s/succinct is//` instead?

docs/src/main/paradox/scala/http/routing-dsl/httpApp.md
```diff
@@ -0,0 +1,71 @@
...
+
+@@snip [HttpAppExampleSpec.scala](../../../../../test/scala/docs/http/scaladsl/HttpAppExampleSpec.scala) { #minimal-routing-example }
+
+Firstly we define an `object` (it can also be a `class`) that extends `HttpApp` and we just implement the routes this server will handle.
+After that, we can start a server just by providing a `host`, `port` and `ServerProperties`. Calling `startSever` blocks the current thread until the server is signaled for termination.
```
jonas: s/startSever/startServer/

docs/src/main/paradox/scala/http/routing-dsl/httpApp.md
```diff
@@ -0,0 +1,71 @@
...
+The default behavior of `HttpApp` is to start a server, and shut it down after `ENTER` is pressed. Once this happens, the server is properly shutdown.
```
jonas: A possible clarification:

```
When the call to `startServer` returns the server is properly shut down.
```

docs/src/main/paradox/scala/http/routing-dsl/httpApp.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Reacting to Bind Failures
+
+`HttpApp` provides different hooks that will be called after a successful and unsuccessful bindings. For example, the server
```
jonas: Maybe: s/bindings/initialization/ at least `bindings` should loose an `s`

docs/src/main/paradox/scala/http/routing-dsl/httpApp.md
```diff
@@ -0,0 +1,71 @@
...
+
+Here you can see an example server that overrides the `postHttpBindingFailure` hook and prints the error to the console (this is also the default behavior)
+
+@@snip [HttpAppExampleSpec.scala](../../../../../test/scala/docs/http/scaladsl/HttpAppExampleSpec.scala) { #failed-binding-example }
+
+So if the port 80 would be already taken by another app, the call to `startServer` returns immediately and the `postHttpBindingFailure` hook will be called.
+
+## Providing your own Actor System
+
+`HttpApp` creates its own `ActorSystem` instance when not provided by the user. This is pretty convenient, but more often than not,
```
jonas: I suggest switching to 1st person:

```
`HttpApp` creates its own `ActorSystem` instance when one is not provided.
In case you already created an `ActorSystem` in your application you can
pass it to `startServer` as illustrated in the following example:

@@snip [HttpAppExampleSpec.scala](../../../../../test/scala/docs/http/scaladsl/HttpAppExampleSpec.scala) { #with-actor-system }
```

docs/src/main/paradox/scala/http/routing-dsl/index.md
jlprat: I wanted to add it, but I forgot! Thanks

docs/src/main/paradox/scala/http/routing-dsl/index.md
jlprat: Cool, will do. Didn't know that feature

akka-http/src/main/java/akka/http/javadsl/server/HttpApp.java
jlprat: Ideally I would return a Try as well, but as there is none in the Std. lib, I was doubting was was the closest, if throwing or Optional.
I picked throwing because this way the caller of this method doesn't lose any information, like which was the exception thrown. If it would return `Optional<ServerBinding>` this information would be lost.

jlprat:
@jonas feedback applied. Thanks!

akka-http/src/main/java/akka/http/javadsl/server/HttpApp.java
ktoso: Historically and personally we'd use Try here. Once every 6 months though we get a user or customer who completely panics on the sight of a single scala type in javadsl ;-) Let's keep the throws for now, thanks

akka-http/src/main/java/akka/http/javadsl/server/HttpApp.java
```diff
@@ -0,0 +1,155 @@
...
+
+    bindingFuture.handle((binding, exception) -> {
+      if (exception != null) {
+        postHttpBindingFailure(exception);
+      } else {
+        //setting the server binding for possible future uses in the client
+        serverBinding.set(binding);
+        postHttpBinding(binding);
+      }
+      return null;
```
ktoso: completion stage and lambdas are such a mess that you have to return here...

akka-http/src/main/java/akka/http/javadsl/server/HttpApp.java
```diff
@@ -0,0 +1,155 @@
...
+
+  /**
+   * Start a server on the specified host and port, using provided settings and [[ActorSystem]] if present.
+   * Note that this method is blocking.
+   * This method may throw an {@link ExecutionException} or {@link InterruptedException} if the future that signals that
+   * the server should shutdown is interrupted or cancelled.
+   */
+  public void startServer(String host, int port, ServerSettings settings, Optional<ActorSystem> system) throws ExecutionException, InterruptedException {
+
+    final ActorSystem theSystem = system.orElseGet(() -> ActorSystem.create(Logging.simpleName(this).replaceAll("\\$", "")));
```
ktoso: good

ktoso:
Great, I think this is ready for being merged as experimental API. 
Let's make sure to mark it as such on a new release.

akka-http/src/main/java/akka/http/javadsl/server/HttpApp.java
jlprat: yes... I don't like the Java APIs either...

ktoso:
Thanks a ton @jlprat !

jlprat:
It was a pleasure!

ktoso:
For such a long debated PR you deserve a special gif I think 😉 

god job

alias1:
<3 Thanks for your hard work! Keen to check this out when I get a chance :3

PR #544, opened 2016-11-18
=pro #499 prepare for publishing Akka HTTP X, dont publish tests

### LINKED ISSUE #499: akka-http-tests should not be released

ktoso:
Resolves #499
Also some smaller build cleanup

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -30,12 +30,19 @@ inThisBuild(Def.settings(
   ),
   testOptions += Tests.Argument(TestFrameworks.JUnit, "-q", "-v"),
   Dependencies.Versions,
-  Formatting.formatSettings
+  Formatting.formatSettings,
+  shellPrompt := { s => Project.extract(s).currentProject.id + " > " }
 ))
+
+val dontPublishSettings = Seq(
+  publishSigned := (),
+  publish := (),
+  publishArtifact in Compile := false
+)
 
 
 lazy val root = Project(
-    id = "akka-http-root",
+    id = "root",
     base = file(".")
   )
   .enablePlugins(UnidocRoot, NoPublish)
@@ -44,6 +51,7 @@ lazy val root = Project(
     // Unidoc doesn't like macros
     unidocProjectExcludes := Seq(parsing)
   )
+  .settings(dontPublishSettings)
   .aggregate(
     parsing,
     httpCore,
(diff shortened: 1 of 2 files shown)


ktoso:
Resolves https://github.com/akka/akka-http/issues/437

build.sbt
```diff
@@ -4,6 +4,7 @@ import akka._
 inThisBuild(Def.settings(
   organization := "com.typesafe.akka",
   organizationName := "Lightbend",
+  crossScalaVersions := Seq("2.11.8", "2.12.0"),
```
jonas: Note, `crossScalaVersions` is currently defined in [`project/Dependencies.scala`](https://github.com/akka/akka-http/blob/master/project/Dependencies.scala#L19) and added via `Dependencies.Versions` below. So if you want to define it here then I suggest the one in `Dependencies.Versions` is removed.

build.sbt
```diff
@@ -44,6 +52,7 @@ lazy val root = Project(
     // Unidoc doesn't like macros
     unidocProjectExcludes := Seq(parsing)
   )
+  .settings(dontPublishSettings)
```
jonas: This should not be needed because of the `.enablePlugins(..., NoPublish)` above.

build.sbt
```diff
@@ -30,12 +31,19 @@ inThisBuild(Def.settings(
   ),
   testOptions += Tests.Argument(TestFrameworks.JUnit, "-q", "-v"),
   Dependencies.Versions,
-  Formatting.formatSettings
+  Formatting.formatSettings,
+  shellPrompt := { s => Project.extract(s).currentProject.id + " > " }
 ))
 
+val dontPublishSettings = Seq(
```
jonas: The [`NoPublish`](https://github.com/akka/akka-http/blob/master/project/Publish.scala#L8) plugin achieves the same thing, so this should not be needed.

build.sbt
```diff
@@ -80,6 +89,7 @@ lazy val httpTestkit = project("akka-http-testkit")
 
 lazy val httpTests = project("akka-http-tests")
   .settings(Dependencies.httpTests)
+  .settings(dontPublishSettings)
```
jonas: With the plugin, this becomes `.enablePlugins(NoPublish)`

build.sbt
ktoso: Heh, I see though the plugin is wrong it seems - needs to cancel publishSigned - the one we actually care about. Will change the PR

build.sbt
ktoso: Thanks, fixing

build.sbt
ktoso: Oh, there's no publish signed now... I see

PR #545, opened 2016-11-18
handle too long entity via setting limitable on it

ktoso:
I had this WIP outstanding, but actually not sure if it's the correct way we want to handle this (as in, if it's the right place).
Though on the other hand, we do the counting anyway in `Limitable` nowadays too.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
@@ -605,11 +605,13 @@ object HttpEntity {
           // "no limit"
           case Some(SizeLimit(bytes, cl @ Some(contentLength))) ⇒
             if (contentLength > bytes) throw EntityStreamSizeException(bytes, cl)
-          // else we still count but never throw an error
+            maxBytes = bytes
+            bytesLeft = bytes
           case Some(SizeLimit(bytes, None)) ⇒
             maxBytes = bytes
             bytesLeft = bytes
           case None ⇒
+          // else we still count but never throw an error
         }
       }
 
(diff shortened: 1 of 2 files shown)


akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
```diff
@@ -605,19 +605,21 @@ object HttpEntity {
...
+          // else we still count but never throw an error
         }
       }
 
       override def onPush(): Unit = {
         val elem = grab(in)
         bytesLeft -= sizeOf(elem)
         if (bytesLeft >= 0) push(out, elem)
-        else failStage(EntityStreamSizeException(maxBytes))
+        else failStage(EntityStreamSizeException(maxBytes, Some(maxBytes - bytesLeft)))
```
schmitch: actually, I guess we should **never** failStage when something on our side misbehaves and instead start a graceful shutdown with an error (maybe `INTERNAL_ERROR`?). or better just push and let the client give us a `FRAME_SIZE_ERROR`. not sure what's better

schmitch: ah, well I got confused. And of course this stream needs to fail. I somehow was already inside the Http2ServerDemuxer...

ktoso:
Rebased once more..

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -207,6 +214,21 @@ class Http2ServerSpec extends AkkaSpec with WithInPendingUntilFixed with Eventua
           remainingToServerWindowFor(TheStreamId) shouldBe 0
           expectNoWindowUpdates(100.millis) // might fail here until all buffers have been filled
         }
+      "fail if advertised content-length doesn't match (actual entity is longer)" in new WaitingForRequest(
```
jrudolph: I think this also needs to check that `RST_STREAM` was sent for the stream and that later streams still work.

ktoso:
I think I'll kill this PR fow now..

PR #549, opened 2016-11-18
Transparent head handling moved from a blueprint to a directive (#40).

### LINKED ISSUE #40: Transparent HEAD request should not make it impossible to handle HEAD in route
<a href="https://github.com/ktoso"></img></a> **Issue by [ktoso](https://github.com/ktoso)**
_Monday Aug 01, 2016 at 10:17 GMT_
_Originally opened as https://github.com/akka/akka/issues/21085_

---

Given such route

``` scala
  val routes: Route =
    path("setting") {
      val value = system.settings.config.getString("akka.http.server.transparent-head-requests")
      complete(s"akka.http.server.transparent-head-requests = $value")
    } ~
    path("test") {
      head { complete(HttpResponse()) }
    }
```

people are very confused since the HEAD route never gets hit.
This is because of `akka.http.server.transparent-head-requests` which mutated it into a GET request and would strip the response entity from the response.

We should make this still work - if someone wants to handle a HEAD direcly they should be able to.

```
$ curl 127.0.0.1:8080/setting
akka.http.server.transparent-head-requests = on

ktoso @ 月~/code/akka [master*]
$ http -v HEAD 127.0.0.1:8080/test
HEAD /test HTTP/1.1
Accept: */*
Accept-Encoding: gzip, deflate
Connection: keep-alive
Host: 127.0.0.1:8080
User-Agent: HTTPie/0.9.3



HTTP/1.1 405 Method Not Allowed
Allow: HEAD
Content-Length: 48
Content-Type: text/plain; charset=UTF-8
Date: Mon, 01 Aug 2016 10:18:17 GMT
Server: akka-http/2.4-SNAPSHOT



```

rbudzko:
Resolution bases on moving transparent head setting to routing settings. Possibly resolves #40.


jrudolph:
@rbudzko interesting approach. Summarizing:

 * transparent-head-request feature is moved out of akka-http-core to akka-http
 * the feature would now work differently, not by rewriting requests but by changing method filters to let HEAD requests through where before only GET requests would be left through. It will only work if you use the `get` or the `method(GET)` directives.
 * You can always choose to handle HEAD requests manually by providing a matching route for HEAD requests (but it must be before the one with the also matching get filter)
 * as before response bodies are automatically held back by the low-level implementation for responses to GET requests (in HttpResponseRendererFactory)

I'm not yet completely sure what to think about it. It seems simple enough and will probably handle the common cases good enough.

The alternative would be a dedicated directive that would rewrite requests as before. The disadvantage would be that it would need to be manually applied when needed and couldn't just be enabled by configuration.

rbudzko:
@jrudolph, summary is correct - that was my intention. As per my knowledge, it does not affect anything in negative way, but I also understand that change might be controversial. When I was thinking about way of implementing it I was trying to think as a enterprise developer for whom this change will be transparent yet beneficial.

jrudolph:
PLS BUILD

rbudzko:
PLS BUILD

rbudzko:
@jrudolph :wink:

rbudzko:
@jrudolph @ktoso I believe we should close this PR. Do we like/feel confident with the approach to merge?

rbudzko:
PLS BUILD

rbudzko:
@gosubpl you have been discussing merging this on a conference with @jrudolph. Did I get it right?

akka-http/src/main/resources/reference.conf
```diff
@@ -19,6 +19,14 @@ akka.http.routing {
   # Enables/disables the rendering of the "rendered by" footer in directory listings
   render-vanity-footer = yes
 
+  # Enables/disables automatic handling of HEAD requests.
+  # If this setting is enabled the server dispatches HEAD requests as GET
```
2beaucoup: is this still valid?

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/MethodDirectivesSpec.scala
```diff
@@ -28,21 +29,60 @@ class MethodDirectivesSpec extends RoutingSpec {
...
+    val headRoute = head {
+      emptyComplete
+    }
+
+    val getRoute = get {
+      emptyComplete
+    }
+
+    def responseEntityEmpty() = {
+      val lengthF = response._3.dataBytes.runFold(0)((c, _) ⇒ c + 1)
```
2beaucoup: `response.entity` plz

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/MethodDirectivesSpec.scala
```diff
@@ -28,21 +29,60 @@ class MethodDirectivesSpec extends RoutingSpec {
...
+      implicit val settings = RoutingSettings.default.withTransparentHeadRequests(true)
+
       Head() ~> headRoute ~> check {
         status shouldEqual StatusCodes.OK
+        responseEntityEmpty()
+      }
+    }
+
+    "allow manual complete if transparent head is disabled" in {
+      implicit val settings = RoutingSettings.default.withTransparentHeadRequests(false)
```
2beaucoup: isn't relevant here, right?

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/MethodDirectivesSpec.scala
```diff
@@ -28,21 +29,60 @@ class MethodDirectivesSpec extends RoutingSpec {
...
 
-        val lengthF = response._3.dataBytes.runFold(0)((c, _) ⇒ c + 1)
-        val length = Await.result(lengthF, Duration(100, "millis"))
-        length shouldEqual 0
+      Head() ~> headRoute ~> check {
+        status shouldEqual StatusCodes.OK
+        responseEntityEmpty()
+      }
+    }
```
2beaucoup: do we really need these `Head() ~> headRoute` tests? maybe add a `checkHead(route: Route, transparentHeadRequests: Boolean)` helper?

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/MethodDirectivesSpec.scala
```diff
@@ -28,21 +29,60 @@ class MethodDirectivesSpec extends RoutingSpec {
...
-      ))
+      )
+    )
+
+    val headRoute = head {
+      emptyComplete
+    }
+
+    val getRoute = get {
+      emptyComplete
```
2beaucoup: should this complete something?

raboof:
@rbudzko sorry for leaving this for so long - it definitely looks interesting. would you be interested in rebasing and looking into the review comments above?

raboof:

PR #550, opened 2016-11-18
=ht2 use takeWhile(inclusive) instead of custom completeIf

### LINKED ISSUE #536: Replace internal completeAfter with takeUntil(..., inclusive) once on Akka 2.4.13
It's used in `HttpRequestHeaderHpackDecompression`, and was implemented in Akka now for 2.4.13

ktoso:
Resolves #536


drewhk:
LGTM

jrudolph:
Let's stash this one for now. I think we might be able to simplify this anyway by moving completion handling into the demux stage. I think we need to handle completion there anyway because we need to maintain the stream state machine.

ktoso:
Agreed, happy to close.

PR #487, opened 2016-11-06
[docs] Add missing import `akka.actor.ActorSystem`

jtwarren:

PR #553, opened 2016-11-18
Improve http client documentation

jtwarren:
Improves the examples by removing imports from the top of the file and explicitly importing inline within examples.

I'm happy to go through other parts of the documentation, just want to make sure I'm on the right track first!

cc @ktoso, per https://github.com/akka/akka-http/pull/487#issuecomment-258929762.

### DIFF
--- a/docs/src/test/scala/docs/http/scaladsl/HttpClientExampleSpec.scala
+++ b/docs/src/test/scala/docs/http/scaladsl/HttpClientExampleSpec.scala
@@ -7,18 +7,18 @@ package docs.http.scaladsl
 import docs.CompileOnlySpec
 import org.scalatest.{ Matchers, WordSpec }
 
-import scala.concurrent.{ ExecutionContextExecutor, Future }
+import scala.concurrent.Future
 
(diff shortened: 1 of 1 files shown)


ktoso:
OK TO TEST

ktoso:
Thanks a lot for the PR :-) 
Changes like this are very useful, thanks!
If you have the time going through other examples like this would be great - newcomers get lost if we don't have the imports so it's really good to have them properly inside the example snippets.

jtwarren:
@ktoso the test failure looks unrelated to my change, but let me know if I'm wrong:
```
HttpServerBluePrint.ControllerStage: requirement failed: Cannot pull closed port (requestParsingIn), inside HttpRequest(HttpMethod(POST),http://example.com/,List(Host: example.com, Expect: 100-continue, Timeout-Access: <function1>),HttpEntity.Chunked(application/octet-stream),HttpProtocol(HTTP/1.1))
```

jrudolph:
PLS BUILD

ktoso:
Thanks a lot :)

Issue #465, opened 2016-11-02
Make `Server-Side WebSocket Support` and `HTTP Model` first Level Topic for Scala Docs


jlprat:
Add `Server-Side WebSocket Support` and `HTTP Model` as main topic in Scala docs.
Remove them as subsections from the current place they are now.

PR #554, opened 2016-11-19
Scala docs reorg

jlprat:
Issue: #465
Make Http Model entry a first level topic
Make server side Websocket Support a first level topic
Additionally, fixed the Java docs, Http Model was supposed to be first level topic but it wasn't shown as it also belogned to `commons/index.md`


ktoso:
LGTM, thanks!

PR #562, opened 2016-11-21
=htc make code relying on ByteStringParser more resilient against exception wrapping

jrudolph:


ktoso:
I see, so simply that the exceptions are wrapped now. 
Thanks a lot @jrudolph, LGTM

PR #568, opened 2016-11-22
=spray-json #564 fix decoding emoji, port from spray-json patch 

### LINKED ISSUE #564: Problem with emoji?
I'm using "com.typesafe.akka" %% "akka-http-spray-json-experimental" % "2.4.7", and doing something like:

case class User(username: String)
val userFormat = jsonFormat1(User)

and whenever username: String has an emoji in it, it does not properly handle it. I get the question mark in a bounding box.

This is something new as I just recently upgraded to scala 2.11.8, sbt 0.13.13, and replaced spray.io with akka-http, where things were being handled properly.

Has anyone seen this and know what to do?

I've confirmed that the marshaller is the problem, as I wrote a custom marshaller and the problem goes away.

I've also written a reproducer that can be found here: 
https://www.dropbox.com/s/46fy46uojcnjs1u/echo.zip?dl=0

Just download the folder, cd into echo, type: sbt run and then open another terminal on the same machine echo is running and type:

curl -XPOST 'localhost:8080/echo' -d '{"msg":"😎"}' --header "Content-Type: application/json"

You will get something back like:

{"
"msg": "�"
}

Thanks!

ktoso:
Resolves #564 
I'm somewhat stuck on adding the scalatest dependency there though... any ideas, maybe @2m or @akka/akka-team ?

### DIFF
--- a/akka-http-marshallers-scala/akka-http-spray-json/src/main/scala/akka/http/scaladsl/marshallers/sprayjson/SprayJsonByteStringParserInput.scala
+++ b/akka-http-marshallers-scala/akka-http-spray-json/src/main/scala/akka/http/scaladsl/marshallers/sprayjson/SprayJsonByteStringParserInput.scala
@@ -20,7 +20,7 @@ final class SprayJsonByteStringParserInput(bytes: ByteString) extends DefaultPar
   import SprayJsonByteStringParserInput._
 
   private[this] val byteBuffer = ByteBuffer.allocate(4)
-  private[this] val charBuffer = CharBuffer.allocate(1)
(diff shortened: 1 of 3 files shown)


ktoso:
PLS BUILD

johanandren:
Ups, doesn't compile, missing dependency? `object scalatest is not a member of package org`

ktoso:
I'd actually need help making this compile. Please help @2m?

2m:
Sorry I missed this. Somehow Github did not notify me... :\

Anyway, added a commit that fixes dependency issue and fixes test compilation.

ktoso:
You rock, thanks a lot @2m !

PR #570, opened 2016-11-22
=htc #364 add test that backpressure will be fixed on akka 2.4.14

### LINKED ISSUE #364: WebSocket Backpressure not working
Hi Akka Team,

I think backpressure for (single) WebSockets might be broken. A test project shows strange behavior
see attached screenshots and reproducer repository. The strange thing is that if this is correct it has been broken during the transition from Akka Streams 1.0 to Akka Streams 2.0 and is broken since including the latest release Akka 2.4.10. So potentially I'm missing some important settings?

**echo server with red == data in, green == data out**

[observed backpressure with Akka 1.0](https://github.com/MichaelZinsmaier/backpressure-bug/blob/master/withBackpressure.png)
[no backpressure with Akka 2.x](https://github.com/MichaelZinsmaier/backpressure-bug/blob/master/broken.png)
[code](https://github.com/MichaelZinsmaier/backpressure-bug)

Assumption: if the server is not reading from a WebSocket stream after the HTTP upgrade this should result in TCP backpressure thus blocking the sender. If the server reads slowly the sender will be throttled (via TCP).

**Test Setup:**
A Netty based WebSocket client sending data as fast as possible. A Akka Http based WebSocket server receiving data and throttling the data stream (with sleep because the throttle stage didn't exist at the time of writing).

```
1000 bytes per WS object
10 milliseconds delay per WS object
20 MB test data => ~20.000 WS objects
no further configuration
```

(running with 200MB and 2 milliseconds yields similar [results](https://github.com/MichaelZinsmaier/backpressure-bug/blob/master/broken_200mb_35secSendBurst.png))

**reproducing without sockets**

[code see branch broken serverLayer](https://github.com/MichaelZinsmaier/backpressure-bug/tree/brokenServerLayer)
A small test without Netty / real Sockets / real IO yields similar results, the source seems to
complete very fast without being backpressured by the slow intermediate consumer. Interestingly
a test with a large streamed HTTP Post request does show backpressure.

Hope that helps if you have further questions don't hesitate

best Michael

jrudolph:
Fixes #364.

### DIFF
--- a/akka-http-core/src/test/scala/akka/http/impl/engine/ws/MessageSpec.scala
+++ b/akka-http-core/src/test/scala/akka/http/impl/engine/ws/MessageSpec.scala
@@ -8,7 +8,7 @@ import akka.NotUsed
 
 import scala.concurrent.duration._
 import scala.util.Random
-import org.scalatest.{ Matchers, FreeSpec }
+import org.scalatest.{ FreeSpec, Matchers }
 import akka.stream.scaladsl._
 import akka.stream.testkit._
 import akka.util.ByteString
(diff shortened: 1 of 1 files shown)


jrudolph:
Supposed to fail on Akka 2.4.12. Should be fixed with upcoming Akka 2.4.14.

ktoso:
PLS BUILD

PR #572, opened 2016-11-22
Update bintray package

2m:
Because this is the package name we got from bintray which is linked to JCenter and can not rename it.


project/Publish.scala
```diff
@@ -25,7 +25,7 @@ object Publish extends AutoPlugin {
 
   override def projectSettings = Seq(
     bintrayOrganization := Some("akka"),
-    bintrayPackage := "akka-http"
+    bintrayPackage := "com.typesafe.akka:akka-http_2.11"
```
ktoso: crossScalaVersion instead? since 2.12

PR #583, opened 2016-11-24
WIP Client-side proxy implementation

jypma:
Shared connection pools are maintained per proxy host, ignoring the target
host, since that proxy is the thing we're making actual TCP connections to.

If Http.outgoingConnection is invoked directly, proxy settings are always
applied and respected (this is in fact what the shared gateway also relies
on).

Implements https://github.com/akka/akka-http/issues/115 .


jypma:
@jrudolph Let me know if you think this is going in the right direction.

schmitch:
I would prefer to not have them on `ConnectionPoolSettings` or `ClientConnectionSettings`, since a client would need way more logic to actually have a client with a proxy on a request base, yes with that it's simple to generate a ws client with a global proxy.

jrudolph:
@schmitch I don't understand. Can you elaborate?

schmitch:
Well it's probably best to show it in code, this is the current way to create a akka-ws client:

https://github.com/schmitch/play-ws-akka/blob/master/src/main/scala/play/api/libs/ws/akkaws/AkkaWSClient.scala#L41 (far from perfect yet)

And the API also comes with a `withProxyServer` (this can change on a request base). https://github.com/schmitch/play-ws-akka/blob/master/src/main/scala/play/api/libs/ws/WS.scala#L280

When I now want to actually implement the withProxyServer I probably need to recrate the `ConnectionPoolSettings` and `ClientConnectionSettings` every time I send a request, not sure if that is the best thing to do.

Scalahansolo:
Wanted to ping this to see if there is still movement here? Also, is there anything others can contribute here to get this on track again?

jypma:
I can probably spend a bit of time to rebase it and get the tests green. I think @Scalahansolo 's concern should be coverable by just creating a specific connection pool with specific proxy settings; that's how I intended the config to work. So you'd only create the pool & settings once, hold on to the pool instance, and use that to make requests (`Http.newHostConnectionPool` and `Http.cachedHostConnectionPool`).

rest of @akka/akka-http-team OK?

schmitch:
TLDR; I'm okai with it however I think Proxy Settings should be in their own `case class`, so that we might add authentication later.

Well basically my concern is that some people don't want "global" proxy settings. So sometimes they need to route requests trough a proxy and sometimes not. 
That's basically what I meant. So I think it should be preferable to actually set the Proxy Params on a per Request Level.

Basically if it will be added to `ClientConnectionSettings` than one would need to recreate `ClientConnectionSettings` and `ConnectionPoolSettings`. (I guess the Connection Pool might stay the same even if I actually change these values, when passing it to `Http.singleRequest`)

The problem here might be that this is the easiest way to pass it trough all API methods. But I think it's overly complicated, but from a user perspective the API doesn't look convienient.

Another way would be to actually add the Params to `PoolGateway` and the `HttpRequest` (so that `sharedGateway` would actually extract them from the HttpRequest and pass it to PoolGateway, however than we need to change all the Pool Methods that you can pass a ProxySettigns or so) however that would mean we would need to change some public Apis and I guess that's worse?

P.S.: I'm not Scalahansolo and not in the akka-team. And I think any way for client side proxy is better than no way ;)

jypma:
The reason for putting them in `ClientConnectionSettings` / `ConnectionPoolSettings`, is that they affect which connection pool gets used for the connection (proxied connections go to a different host, after all). So any API that works on the connection pool level is going to see that change.

And `Http.singleRequest` can take a `ConnectionPoolSettings` argument, through which the proxy (and hence connection pool) can be chosen per request. @schmitch what would you rather see instead? having actual request arguments on `singleRequest itself for proxy host and port?

schmitch:
As already said I think it's important to have a case class for the ProxySettings, since at a later point it's probably useful to add authentication on top of it.

Well as written basically I think from a user perspective it would make sense to add proxy host and port to `singleRequest`. However that only applies to `singleRequest` I guess.

At the moment I do most calls like that (pseudo code):
```
@Singleton
class Client (....) {
   private val client = Http()
   ....
   private val connectionContext: HttpsConnectionContext = client.createClientHttpsContext(akkaSSLConfig)
  client.validateAndWarnAboutLooseSettings()
   private val clientConnectionSettings: ClientConnectionSettings = ClientConnectionSettings(system)
      .withConnectingTimeout(FiniteDuration(config.connectionTimeout._1, config.connectionTimeout._2))
      .withIdleTimeout(config.idleTimeout)
      .withUserAgentHeader(config.userAgent.map(`User-Agent`(_)))
   private val connectionPoolSettings: ConnectionPoolSettings = ConnectionPoolSettings(system)
      .withConnectionSettings(clientConnectionSettings)
      .withMaxRetries(0)
      .withIdleTimeout(config.idleTimeout)

def executeRequest[T](request: HttpRequest): Future[HttpResponse] = {
    client.singleRequest(request, connectionContext, connectionPoolSettings)
  }
}
```

With you change I would need to re-create the settings if I want to add proxy support:,
```
def executeRequest[T](request: HttpRequest, proxySettings: ProxySettings): Future[HttpResponse] = {
     val clientConnectionSettings: ClientConnectionSettings = ClientConnectionSettings(system)
.withProxySettings(proxySettings)
    val connectionPoolSettings: ConnectionPoolSettings = ConnectionPoolSettings(system)
      .withConnectionSettings(clientConnectionSettings)
      .withMaxRetries(0)
      .withIdleTimeout(config.idleTimeout)
    client.singleRequest(request, connectionContext, connectionPoolSettings)
  }
}
```

However as said from a user perpective this seems to be better (or put ProxySettings somehow into the HttpRequest):
```
def executeRequest[T](request: HttpRequest, proxySettings: ProxySettings): Future[HttpResponse] = {
    client.singleRequest(request, connectionContext, connectionPoolSettings, proxySettings)
}
```

Might just come down to a personal preference, but I think most settings won't change that much while proxy settings actually can change more often.

Well maybe we *can* add it to ConnectionPoolSettings **and** `singleRequest` (as `Option[ProxySettings]` of course) and than just overwrite it inside `sharedGateway` if present WDYT?

jypma:
@schmitch can you explain your use case more closely? I can see a system having two fixed `val` settings (one with proxy, one without), but you mention proxy settings change quite often. Are you working against multiple dynamic proxy servers?

schmitch:
Well basically in my case comes down to this:
- Internal Requests don't use a Proxy at all
- External Requests use Proxy X (reconfigurable at run time)
- Requests to one of our Partners use Proxy Y (also reconfigurable at run time)

jypma:
@schmitch Thanks a lot for explaining.

Would it be possible to model that as 3 methods that each keep a local `var` of the current proxy settings (as `ClientConnectionSettings`), and simply pass that to `singleRequest`? You'd then only need to construct a new `ClientConnectionSettings` when either of your proxy settings _change_, rather than on every request.

schmitch:
At the moment that won't work since that can lead to races (the class is a `Singleton` so it acts like a object).

Also keep in mind that I also speak from the `play-ws` API perspective where it is possible to set the Proxy on a per Request level (https://www.playframework.com/documentation/2.5.x/api/scala/index.html#play.api.libs.ws.WSRequest@withProxyServer(proxyServer:play.api.libs.ws.WSProxyServer):play.api.libs.ws.WSRequest) and I want to see a `play-ws-akka` client, that's another use case why I think that the API should be modeled to keep in mind that proxy settings can be something that people want to change in their runtime more often.

As said, maybe you should just refactor it and get it inside the code base. With just my feedback it's probably not really useful to change the API of `singleRequest`.

Scalahansolo:
For what it's worth, in our case the per request proxy would be useful as for our use case there are URLs inside our DC that we do not need a proxy for and some that we do. And in most cases, our apps hit both close together at run time.

ktoso:
This seems to be superseded by https://github.com/akka/akka-http/pull/921 ?

jypma:
Oh, interesting. Yes, possibly. Either way, good that someone is actually working on it; I hope #921 also comes with a "simple" `.singleRequest` ability to use proxies, since that's what the S3 library in alpakka is using right now.

jrudolph:
No, this is something different. HTTPS and HTTP proxy support are happening on different layers and need different treatment.

Scalahansolo:
Is this still on the radar to get done at some point?

raboof:
I don't think it has - #921 added support for HTTPS (`HTTP CONNECT`) proxies but that's a separate feature

PR #584, opened 2016-11-24
Fix compilation warnings

jonas:
I know the `() => (()))` are not super nice, but the warnings is somewhat distracting.

Note I didn't test the MultiJvm sbt changes. Is there anything specific I need to do to run those tests?


ktoso:
Good one, thanks!

Issue #492, opened 2016-11-07
FAILED ClientServerSpec


jrudolph:
https://jenkins.akka.io:8498/job/pr-validator-akka-http/243/testReport/junit/akka.http.scaladsl/ClientServerSpec/The_low_level_HTTP_infrastructure_should_timeouts_should_support_client_timeouts_should_close_connection_with_idle_server_after_idleTimeout__using_connection_level_client_API_/

```
akka.http.scaladsl.ClientServerSpec.The low-level HTTP infrastructure should timeouts should support client timeouts should close connection with idle server after idleTimeout (using connection level client API)

org.scalatest.exceptions.TestFailedException: 330564097 nanoseconds was not greater than or equal to 345 milliseconds
	at org.scalatest.MatchersHelper$.indicateFailure(MatchersHelper.scala:340)
	at org.scalatest.Matchers$ShouldMethodHelper$.shouldMatcher(Matchers.scala:6668)
	at org.scalatest.Matchers$AnyShouldWrapper.should(Matchers.scala:6704)
	at akka.http.scaladsl.ClientServerSpec$$anonfun$1$$anonfun$apply$mcV$sp$7$$anonfun$apply$mcV$sp$9$$anonfun$apply$mcV$sp$23.apply(ClientServerSpec.scala:281)
	at org.scalatest.OutcomeOf$class.outcomeOf(OutcomeOf.scala:85)
	at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
	at org.scalatest.Transformer.apply(Transformer.scala:22)
	at org.scalatest.Transformer.apply(Transformer.scala:20)
	at org.scalatest.WordSpecLike$$anon$1.apply(WordSpecLike.scala:1078)
	at org.scalatest.TestSuite$class.withFixture(TestSuite.scala:196)
	at org.scalatest.WordSpec.withFixture(WordSpec.scala:1881)
	at org.scalatest.WordSpecLike$class.invokeWithFixture$1(WordSpecLike.scala:1075)
	at org.scalatest.WordSpecLike$$anonfun$runTest$1.apply(WordSpecLike.scala:1088)
	at org.scalatest.WordSpecLike$$anonfun$runTest$1.apply(WordSpecLike.scala:1088)
	at org.scalatest.SuperEngine.runTestImpl(Engine.scala:289)
	at org.scalatest.WordSpecLike$class.runTest(WordSpecLike.scala:1088)
[...]

Issue #443, opened 2016-10-27
FAILED HttpHeaderSpec


jrudolph:
Failed several times now in PR validation:

https://jenkins.akka.io:8498/job/pr-validator-akka-http/178/testReport/akka.http.impl.model.parser/HttpHeaderSpec/The_HTTP_header_model_must_correctly_parse_and_render_the_headers_If_Match/
https://jenkins.akka.io:8498/job/pr-validator-akka-http/120/testReport/akka.http.impl.model.parser/HttpHeaderSpec/The_HTTP_header_model_must_correctly_parse_and_render_the_headers_If_Match/

```
Error Message

Ok(If-Match: *,List()) did not equal Ok(If-Match: *,List())
Stacktrace

org.scalatest.exceptions.TestFailedException: Ok(If-Match: *,List()) did not equal Ok(If-Match: *,List())
    at org.scalatest.MatchersHelper$.indicateFailure(MatchersHelper.scala:340)
    at org.scalatest.Matchers$ShouldMethodHelper$.shouldMatcher(Matchers.scala:6668)
    at org.scalatest.Matchers$AnyShouldWrapper.should(Matchers.scala:6716)
    at akka.http.impl.model.parser.HttpHeaderSpec$TestHeader.apply(HttpHeaderSpec.scala:691)
    at akka.http.impl.model.parser.HttpHeaderSpec$TestHeader.apply(HttpHeaderSpec.scala:688)
    at akka.http.impl.model.parser.HttpHeaderSpec$TestLine.$eq$bang$eq(HttpHeaderSpec.scala:680)
    at akka.http.impl.model.parser.HttpHeaderSpec$$anonfun$1$$anonfun$apply$mcV$sp$31.apply$mcV$sp(HttpHeaderSpec.scala:305)
    at akka.http.impl.model.parser.HttpHeaderSpec$$anonfun$1$$anonfun$apply$mcV$sp$31.apply(HttpHeaderSpec.scala:304)
    at akka.http.impl.model.parser.HttpHeaderSpec$$anonfun$1$$anonfun$apply$mcV$sp$31.apply(HttpHeaderSpec.scala:304)
    at org.scalatest.OutcomeOf$class.outcomeOf(OutcomeOf.scala:85)
    at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
    at org.scalatest.Transformer.apply(Transformer.scala:22)
    at org.scalatest.Transformer.apply(Transformer.scala:20)
[...]

PR #586, opened 2016-11-25
Match path maps in order of longest matching key prefix (#394)

### LINKED ISSUE #394: PathMatcher DSL: Map Matcher problem (?)
Given the following setup, describing two resources "/a" and "/aa", all requests to "/aa" get rejected. 

```
val route = path(Map("a" -> 1, "aa" -> 2)) {
  res => get {
    complete(res.toString)
  }
}
Get("/a") ~> route ~> check { println(responseAs[String]) } // OK
Get("/aa") ~> route ~> check { println(responseAs[String]) } // rejected
```

According to the documentation, the `Map[String, T]` matcher should match "any of the keys and extracts the respective map value for it". 
[See Documentation](http://doc.akka.io/docs/akka/2.4.11/scala/http/routing-dsl/path-matchers.html) 

I assume the cause for this issue is common prefix (in this case "a") of the two resources. 

If the order of the resources in the definition of the route is reversed, it works as expected:

```
val route = path(Map("aa" -> 2, "a" -> 1)) {
  res => get {
    complete(res.toString)
  }
}
Get("/a") ~> route ~> check { println(responseAs[String]) } // OK
Get("/aa") ~> route ~> check { println(responseAs[String]) } // OK
```

I have a unit test that reproduces this problem with akka-http-experimental 2.4.11

jonas:
Fixes #394 by ensuring order of the provided mappings instead of leaving it up to the undefined order provided by the map implementation.

It might be a bit overkill to include all of the test cases created by @gosubpl but I included them for completeness and to test with `ListMap`.

### DIFF
--- a/akka-http/src/main/scala/akka/http/scaladsl/server/PathMatcher.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/server/PathMatcher.scala
@@ -289,12 +289,13 @@ trait ImplicitPathMatcherConstruction {
    * Creates a PathMatcher from the given Map of path segments (prefixes) to extracted values.
    * If the unmatched path starts with a segment having one of the maps keys as a prefix
    * the matcher consumes this path segment (prefix) and extracts the corresponding map value.
+   * For keys sharing a common prefix the longest matching prefix is selected.
    *
    * @group pathmatcherimpl
    */
   implicit def _valueMap2PathMatcher[T](valueMap: Map[String, T]): PathMatcher1[T] =
     if (valueMap.isEmpty) PathMatchers.nothingMatcher
-    else valueMap.map { case (prefix, value) ⇒ _stringExtractionPair2PathMatcher((prefix, value)) }.reduceLeft(_ | _)
+    else valueMap.toSeq.sortWith(_._1 > _._1).map(_stringExtractionPair2PathMatcher).reduceLeft(_ | _)
 }
 
 /**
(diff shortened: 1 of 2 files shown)


jonas:
Failed test is #493

jonas:
Hmm looks like #443 and #492

ktoso:
PLS BUILD

akka-http/src/main/scala/akka/http/scaladsl/server/PathMatcher.scala
```diff
@@ -289,12 +289,13 @@ trait ImplicitPathMatcherConstruction {
...
    * If the unmatched path starts with a segment having one of the maps keys as a prefix
    * the matcher consumes this path segment (prefix) and extracts the corresponding map value.
+   * For keys sharing a common prefix the longest matching prefix is selected.
    *
    * @group pathmatcherimpl
    */
   implicit def _valueMap2PathMatcher[T](valueMap: Map[String, T]): PathMatcher1[T] =
     if (valueMap.isEmpty) PathMatchers.nothingMatcher
-    else valueMap.map { case (prefix, value) ⇒ _stringExtractionPair2PathMatcher((prefix, value)) }.reduceLeft(_ | _)
+    else valueMap.toSeq.sortWith(_._1 > _._1).map(_stringExtractionPair2PathMatcher).reduceLeft(_ | _)
```
2beaucoup: `sortyBy(_._1)`?

ktoso: checked that, actually would fail hm

ktoso:
Thanks a lot @jonas @gosubpl !

PR #558, opened 2016-11-20
Add rejectEmptyResponse example

jlprat:
Issue: #218 
Add example about `rejectEmptyResponse` for Java.
Add warning about completing a route with an Optional and its non expected output.

PR #590, opened 2016-11-28
Java Optional entity is now being marshalled to empty string - like in scala DSL (#559).

### LINKED ISSUE #559: Wrong Marshalling when Entity is `Optional`
When completing a route with a Java `Optional` the output is not the one I would expect:
## Expected Behavior
* If `Optional[T]` is empty: completed with an empty `Entity`
* If `Optional[T]` is not empty: completed with proper marshalling of `T`

## Actual Behavior
* If `Optional[T]` is empty: completed with `{"present": false}`
* If `Optional[T]` is not empty: completed with `{"present": true}`

I have a gist where one can reproduce this: https://gist.github.com/jlprat/5e1c27ce4b912e0a3eb0ad56237246d7

rbudzko:
Possibly resolves #559.

### DIFF
--- a/akka-http-marshallers-java/akka-http-jackson/src/main/java/akka/http/javadsl/marshallers/jackson/Jackson.java
+++ b/akka-http-marshallers-java/akka-http-jackson/src/main/java/akka/http/javadsl/marshallers/jackson/Jackson.java
@@ -3,22 +3,25 @@
  */
 package akka.http.javadsl.marshallers.jackson;
 
-import java.io.IOException;
-
+import akka.http.javadsl.marshalling.Marshaller;
 import akka.http.javadsl.model.HttpEntity;
 import akka.http.javadsl.model.MediaTypes;
 import akka.http.javadsl.model.RequestEntity;
-import akka.http.javadsl.marshalling.Marshaller;
 import akka.http.javadsl.unmarshalling.Unmarshaller;
-
 import akka.util.ByteString;
+import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.MapperFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.datatype.jdk8.Jdk8Module;
+
+import java.io.IOException;
 
 public class Jackson {
-  private static final ObjectMapper defaultObjectMapper =
-    new ObjectMapper().enable(MapperFeature.SORT_PROPERTIES_ALPHABETICALLY);
+  private static final ObjectMapper defaultObjectMapper = new ObjectMapper()
+    .enable(MapperFeature.SORT_PROPERTIES_ALPHABETICALLY)
+    .registerModule(new Jdk8Module())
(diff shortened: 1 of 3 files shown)


jlprat:
Once this is merged I will work on https://github.com/akka/akka-http/pull/558 to have the right wording and a working example.

rbudzko:
Well, not sure if it is appropriate solution tbh, but I hope to have wider audience criticizing it 😈 .

jlprat:
That's why I said once this is merged ;)

akka-http-marshallers-java/akka-http-jackson/src/main/java/akka/http/javadsl/marshallers/jackson/Jackson.java
```diff
@@ -35,23 +38,31 @@
...
     return Unmarshaller.sync(s -> fromJSON(mapper, s.utf8String(), expectedType));
   }
 
   private static String toJSON(ObjectMapper mapper, Object object) {
     try {
-      return mapper.writeValueAsString(object);
+      final String candidate = mapper.writeValueAsString(object);
+
+      // Jackson (if root unwrapped) pushes `null` String into output instead of empty String.
+      if ("null".equals(candidate)) {
```
jlprat: I guess some strings that render to `"null"` will get caught by this if and be transformed to `""`.
Maybe some type checking on `object`?

rbudzko: Yeah, you are right. It will work for null objects of all types as jackson renders empty root to `"null"`, but indeed `"null"` will be rendered incorrectly.

jlprat: Even if it's usually considered a bad practice to check the type of an object, I think in this particular case might be quite handy

rbudzko: I think there is yet another issue: Scala DSL produced unquoted strings if string marshaled. Jackson wrap strings with the quote. Need to dig into differences at some point more closely.

rbudzko:
Tweaked to, so root strings are de quoted to reflect scala dsl default serialization. Root null serializations are now translated to empty string as well. 

Unfortunately, I think there is more. For example scala dsl serializes `"\"a"` to `""a"`. It is illegal I believe and a lot of json parses will fail on that, right? Java dsl will make it `"\"a"`.

ktoso:
Thanks for the update!

> Unfortunately, I think there is more. For example scala dsl serializes "\"a" to ""a". It is illegal I believe and a lot of json parses will fail on that, right? Java dsl will make it "\"a".

Yes, it should render the `\"` correctly

// nice new avatar 😉  📷

rbudzko:
Not sure if spray serializer should be fixed here. New issue I think. I think it is uber rare to serialize strings as roots, so impact is very low I believe.

rbudzko:
#600

rbudzko:
@ktoso @jlprat Shall we decline in favor of approach described in #559?

jlprat:
Yes, I think it makes sense

Issue #604, opened 2016-12-01
JavaDSL Unmarshaller is leaking ScalaDSL classes


jlprat:
The current `akka.http.javadsl.unmarshalling.Unmarshaller` is leaking some scaladsl as return types. The current examples I found are:
* `entityToMultipartByteRanges()` that is returning `Unmarshaller<http.javadsl.model.HttpEntity, akka.http.scaladsl.model.Multipart.ByteRanges>`
* `entityToUrlEncodedFormData()` that is returning `Unmarshaller<http.javadsl.model.HttpEntity, akka.http.scaladsl.model.FormData>`

ktoso:
Thanks for catching it, that's a bug / missing API that needs ot be fixed.
It is possible that we're missing those datatypes in javadsl... :/

Would you give it an attempt to fix?

jlprat:
Sure!

ktoso:
Thanks @jlprat !

PR #609, opened 2016-12-03
Fix Java Unmarshaller to return only Java DSL classes

jlprat:
Issue: #604
Return types should belong to Java DSL only.
Add extra method `entityToMultipartFormData` to unmarshall multipart form data objects.

### DIFF
--- a/akka-http/src/main/scala/akka/http/javadsl/unmarshalling/Unmarshaller.scala
+++ b/akka-http/src/main/scala/akka/http/javadsl/unmarshalling/Unmarshaller.scala
@@ -10,10 +10,11 @@ import akka.http.impl.model.JavaQuery
 import akka.http.impl.util.JavaMapping
 import akka.http.impl.util.JavaMapping.Implicits._
 import akka.http.javadsl.model._
-import akka.http.scaladsl.model.{ ContentTypeRange, ContentTypes }
+import akka.http.scaladsl.model.{ ContentTypeRange, ContentTypes, FormData => SFormData }
+import akka.http.scaladsl.model.{ Multipart => SMultipart }
 import akka.http.scaladsl.unmarshalling
 import akka.http.scaladsl.unmarshalling.FromEntityUnmarshaller
-import akka.http.scaladsl.unmarshalling.Unmarshaller.{ EnhancedFromEntityUnmarshaller, UnsupportedContentTypeException }
+import akka.http.scaladsl.unmarshalling.Unmarshaller.{EnhancedFromEntityUnmarshaller, UnsupportedContentTypeException}
 import akka.http.scaladsl.util.FastFuture
 import akka.stream.Materializer
 import akka.util.ByteString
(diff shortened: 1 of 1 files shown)


jlprat:
Let me know if this approach is OK.

I added an extra Unmarshaller so website examples can make use of it.

akka-http/src/main/scala/akka/http/javadsl/unmarshalling/Unmarshaller.scala
```diff
@@ -45,13 +53,26 @@ object Unmarshaller {
   def entityToByteArray: Unmarshaller[HttpEntity, Array[Byte]]       = unmarshalling.Unmarshaller.byteArrayUnmarshaller
   def entityToCharArray: Unmarshaller[HttpEntity, Array[Char]]       = unmarshalling.Unmarshaller.charArrayUnmarshaller
   def entityToString: Unmarshaller[HttpEntity, String]               = unmarshalling.Unmarshaller.stringUnmarshaller
-  def entityToUrlEncodedFormData: Unmarshaller[HttpEntity, FormData] = unmarshalling.Unmarshaller.defaultUrlEncodedFormDataUnmarshaller
-  def entityToMultipartByteRanges: Unmarshaller[HttpEntity, Multipart.ByteRanges] = unmarshalling.MultipartUnmarshallers.defaultMultipartByteRangesUnmarshaller
+  def entityToUrlEncodedFormData: Unmarshaller[HttpEntity, FormData] = urlEncodedFormDataUnmarshaller(MediaTypes.`application/x-www-form-urlencoded`)
```
jlprat: I could not take the same approach as the other ones, as `FormData` from JavaDSL and ScalaDSL don't share hierarchy.

jrudolph: Even then, I think we should use the same implementation and adapt just the data types. Otherwise, it will be double work to maintain.

jrudolph: I.e. piggy-back on the Scala unmarshaller and then convert from the Scala `FormData` to the java one.

jlprat: Will do

jrudolph:
PLS BUILD

jrudolph:
It seems that this is technically a binary incompatible change. Let's see what newly-enabled mima says.

jlprat:
What does mima say?

jrudolph:
PLS REBUILD

jrudolph:
PLS BUILD

jrudolph:
Sounds good, thanks.

jlprat:
Reverted to old methods, added deprecation information and added the new ones with different names

ktoso:
Seems good, LGTM

ktoso:
new-years-merging/cleaning, this PR is good I believe. Merging.

PR #617, opened 2016-12-05
=htc #616 replace OutputTruncationException for the http client use case

### LINKED ISSUE #616: Improve One2OneBidiFlow OutputTruncation error for the client
Instead of the meaninglessly generic One2OneBidiFlow error message it should have a domain specific one. Also, it shouldn't throw in the stage to prevent that it gets logged with ERROR level.

jrudolph:
We will now send an `UnexpectedConnectionClosureException` which more
closely explains the problem at hand.

Also, the error is reported with `failStage` instead of throwing which
prevents the noisy logging.

Fixes #616.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
@@ -28,6 +28,8 @@ import akka.stream.stage.GraphStageLogic
 import akka.stream.stage.{ InHandler, OutHandler }
 import akka.http.impl.util.LogByteStringTools._
 
+import scala.util.control.NoStackTrace
+
 /**
  * INTERNAL API
  */
@@ -120,7 +122,10 @@ private[http] object OutgoingConnectionBlueprint {
         terminationFanout.out(1))
     })
 
-    One2OneBidiFlow[HttpRequest, HttpResponse](-1) atop
+    One2OneBidiFlow[HttpRequest, HttpResponse](
+      -1,
+      outputTruncationException = new UnexpectedConnectionClosureException(_)
+    ) atop
       core atop
       logTLSBidiBySetting("client-plain-text", settings.logUnencryptedNetworkBytes)
   }
@@ -352,4 +357,7 @@ private[http] object OutgoingConnectionBlueprint {
       override def preStart(): Unit = getNextMethod()
     }
   }
+
+  class UnexpectedConnectionClosureException(outstandingResponses: Int)
+    extends RuntimeException(s"The http server closed the connection unexpectedly before delivering responses for $outstandingResponses outstanding requests")
 }
--- a/akka-http-core/src/main/scala/akka/http/impl/util/One2OneBidiFlow.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/util/One2OneBidiFlow.scala
@@ -15,10 +15,10 @@ import akka.stream.stage.{ OutHandler, InHandler, GraphStageLogic, GraphStage }
  */
 private[http] object One2OneBidiFlow {
 
-  case class UnexpectedOutputException(element: Any) extends RuntimeException(element.toString) with NoStackTrace
-  case object OutputTruncationException
-    extends RuntimeException("Inner stream finished before inputs completed. Outputs might have been truncated.")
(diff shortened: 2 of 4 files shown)


jrudolph:
PLS BUILD

ktoso:
> We will now send an UnexpectedConnectionClosureException which more closely explains the problem at hand.

Great rename.

> Also, the error is reported with failStage instead of throwing which prevents the noisy logging.

Yep, awesome.

akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
```diff
@@ -352,4 +357,8 @@ private[http] object OutgoingConnectionBlueprint {
       override def preStart(): Unit = getNextMethod()
     }
   }
+
+  class UnexpectedConnectionClosureException(outstandingResponses: Int)
+    extends RuntimeException(s"The http server closed the connection unexpectedly before delivering responses for $outstandingResponses outstanding requests")
```
ktoso: can we add the following too?

```
[...]. This may have happened e.g. because of an connection idle-timeout or abrupt termination of the entity stream
```

akka-http-core/src/main/scala/akka/http/impl/util/One2OneBidiFlow.scala
```diff
@@ -15,9 +15,11 @@ import akka.stream.stage.{ OutHandler, InHandler, GraphStageLogic, GraphStage }
  */
 private[http] object One2OneBidiFlow {
 
-  case class UnexpectedOutputException(element: Any) extends RuntimeException(element.toString) with NoStackTrace
-  case object OutputTruncationException
-    extends RuntimeException("Inner stream finished before inputs completed. Outputs might have been truncated.")
+  case class UnexpectedOutputException(element: Any)
+    extends RuntimeException(s"Inner flow produced unexpected result element '$element' when no element was outstanding")
+    with NoStackTrace
```
ktoso: I was thinking again if it was a good idea to make them NoStackTrace, WDYT?

akka-http-core/src/main/scala/akka/http/impl/util/One2OneBidiFlow.scala
```diff
@@ -83,10 +91,10 @@ private[http] object One2OneBidiFlow {
               pullSuppressed = false
               if (!isClosed(in)) pull(in)
             }
-          } else throw new UnexpectedOutputException(element)
+          } else failStage(unexpectedOutputException(element))
```
ktoso: yes, great.

ktoso:
LGTM, had some questions in-line

akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
jrudolph: I started adding more info but I then figured that it isn't possible to provide all possible causes of this comprehensively. I wonder if we shouldn't create more detailed explanations for some errors in the documentation instead which might also be easier to maintain?

akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
```diff
@@ -352,4 +357,8 @@ private[http] object OutgoingConnectionBlueprint {
...
+    with NoStackTrace
```
jrudolph: I just kept it from what the default One2OneBidiFlow exceptions do. The stack trace will always be the same. But, I agree, it shouldn't happen to often to make it expensive, and having a stack trace will also help people look into the code (even if it's always the same stack trace). I'll change that.

akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
ktoso: Yeah true - good as is

ktoso: About errors I think the idea is good, let's start documenting them in some docs section as people keep asking  things - next time someone asks we document it ;)

Issue #596, opened 2016-11-29
Host: header is incorrectly required for HTTP/1.0 requests


TJC:
HTTP/1.0 does *not* require a Host: header.

However if you attempt to query an Akka HTTP server via HTTP/1.0 without a Host: header, it responds with:

```
400 Bad Request
Request is missing required `Host` header
```

And on the server side, logs:
```
Illegal request, responding with status '400 Bad Request': Request is missing required `Host` header: Cannot establish effective URI of request to `/ping`, request has a relative URI and is missing a `Host` header
```

# To reproduce

- Run up a simple "hello world" akka http server
- telnet localhost 8080
- paste in `GET / HTTP/1.0`

jonas:
From my understanding, an Akka HTTP server needs to have `akka.http.server.default-host-header` configured to accept HTTP/1.0 requests without a `Host` header:

https://github.com/akka/akka-http/blob/ebfbdc22d64da42307d3c6951847e3f3cd02dfae/akka-http-core/src/main/resources/reference.conf#L102-L110

jrudolph:
Yes, I agree with @jonas. We could try to improve the server-side log message to mention how to fix it. WDYT?

jonas:
Yes, I can make a PR

TJC:
Just a comment, but, I don't think documenting the default-host-header option is really the right way to fix this.
If you're not compatible with HTTP/1.0 then just reject messages immediately as such.
If you are accepting HTTP/1.0 requests, then you should not require the Host: header.

My preference would be for your HTTP server library to operate like every other HTTP server out there, and to Just Work (without workarounds or hacks) with both 1.0 and 1.1 of the HTTP standard.

As someone who writes software, I love good quality libraries that work easily and without surprises, and I'd like Akka to aspire to being one of those.

ktoso:
Akka certainly is known for striving for high quality, and here I'd argue it was a design choice; I would not call a well documented option a hack - we may disagree but let's not get personal and offensive please - thanks.

Improving the logs is a simple first step, and we can consider additional steps if really needed.

jrudolph:
> If you're not compatible with HTTP/1.0 then just reject messages immediately as such.
> If you are accepting HTTP/1.0 requests, then you should not require the Host: header.

> My preference would be for your HTTP server library to operate like every other HTTP server out there, and to Just Work (without workarounds or hacks) with both 1.0 and 1.1 of the HTTP standard.

[...]

PR #619, opened 2016-12-06
Improve the server-side log message when default-host-header is not set

jonas:
When the server receives a request it will try to reconstruct the
effective request URI as defined in HTTP/1.1 [1]. Requests using
HTTP/1.0 will use the same code path but since the `Host` header is not
required in HTTP/1.0 Akka HTTP requires that a default `Host` header
value is configured in `akka.http.server.default-host-header`. In
addition to the documentation in `reference.conf` make this more clear
by also hinting to the setting in the error log as discussed in #596.

 [1]: https://tools.ietf.org/html/rfc7230#section-5.5


jrudolph:
LGTM. Thanks, @jonas.

jrudolph:
PLS BUILD

ktoso:
LGTM, thanks 👍

Issue #623, opened 2016-12-07
Add links to ScalaDoc and JavaDoc to main documentation site


eswenson1:
It is currently difficult (with Google) to locate the ScalaDoc (and JavaDoc) for Akka HTTP. Without much effort, you can find the documentation here:  http://doc.akka.io/docs/akka-http/current/, but neither that page, or the subordinate Scala or Java pages, include links to http://doc.akka.io/api/akka-http/current/akka/http/scaladsl/index.html and http://doc.akka.io/api/akka-http/current/akka/http/javadsl/index.html. 

These pages should include links to the Scala/JavaDoc for Akka HTTP.

ktoso:
Thanks for the ticket :)

PR #624, opened 2016-12-07
Add a links to ScalaDoc and JavaDoc on main documentation pages #623

eswenson1:


typesafehub-validator:
Hi @eswenson1,

eswenson1:
Received email from Lightbend asking me to add a comment to the PR saying I had signed the CLA. Not sure why this was necessary because the typesafe-cla-validator check passed as soon as I acknowledged the signing via GitHub, but doing it anyway.

jonas:
In order for the links to work for each release they should point to the API docs of a specific version. The easiest way to do this will probably be to use the `@scaladoc` and `@javadoc` directives as documented here http://developer.lightbend.com/docs/paradox/latest/features/linking.html#scaladoc-directive.

This will require that the Paradox plugin is bumped to version 0.2.7.

eswenson1:
Bumping the Paradox plugin to 0.2.7 results in these errors loading the project:

```
[info] Compiling 1 Scala source to /Users/eswenson/Projects/akka/ws/akka-http/project/target/scala-2.10/sbt-0.13/classes...
[error] /Users/eswenson/Projects/akka/ws/akka-http/project/ParadoxSignatureDirective.scala:38: overloaded method constructor File with alternatives:
[error]   (x$1: java.io.File,x$2: String)java.io.File <and>
[error]   (x$1: String,x$2: String)java.io.File
[error]  cannot be applied to (java.io.File, org.pegdown.ast.DirectiveNode.Source)
[error]         val file = new File(page.file.getParentFile, node.source)
[error]                    ^
[error] one error found
[error] (compile:compileIncremental) Compilation failed
```

eswenson1:
And after putting in some of-unknown-correctness code to convert an org.pegdown.ast.DirectiveNode.Source to a String, the project fails later on with

```
[info] Compiling 1 Scala source to /Users/eswenson/Projects/akka/ws/akka-http/akka-http-core/target/scala-2.11/classes...
[trace] Stack trace suppressed: run last docs/compile:paradoxMarkdownToHtml for the full output.
[error] (docs/compile:paradoxMarkdownToHtml) com.lightbend.paradox.markdown.SnipDirective$LinkException: Unknown snippet [/Users/eswenson/Projects/akka/ws/akka-http/docs/src/main/paradox/scala/http/routing-dsl/directives/basic-directives/() (No such file or directory)] referenced from [scala/http/routing-dsl/directives/basic-directives/cancelRejection.html]
[error] Total time: 11 s, completed Dec 6, 2016 5:33:59 PM
```

Seems like switching to Paradox plugin 0.2.7 is non-trivial.

jonas:
The following will make the `@signature` directive work:
```
diff --git i/project/ParadoxSignatureDirective.scala w/project/ParadoxSignatureDirective.scala
index f9ac99d..80a750e 100644
--- i/project/ParadoxSignatureDirective.scala
+++ w/project/ParadoxSignatureDirective.scala
@@ -35,7 +35,10 @@ object ParadoxSupport {
     def render(node: DirectiveNode, visitor: Visitor, printer: Printer): Unit =
       try {
         val labels = node.attributes.values("identifier").asScala.map(_.toLowerCase())
-        val file = new File(page.file.getParentFile, node.source)
+        val file = node.source match {
+          case direct: DirectiveNode.Source.Direct => new File(page.file.getParentFile, direct.value)
+          case _                                   => sys.error("Source references are not supported")
+        }

         val Signature = """\s*((def|val|type) (\w+)(?=[:(\[]).*)(\s+\=.*)""".r // stupid approximation to match a signature
         //println(s"Looking for signature regex '$Signature'")
```

The related change in Paradox is commit lightbend/paradox@69d3e8b38c74d2d056822c58a83ee02c2a9169c5

ktoso:
Thanks! Please submit that change as PR @jonas, I'd like to attribute it to you properly. :)

ktoso:
Refs https://github.com/akka/akka-http/issues/623

jonas:
OK, I can do a follow up PR.

eswenson1:
@ktoso Should I wait for @jonas's PR to be merged before I move forward with this one?  I can't switch from explicit URLs to @scaladoc/@javadoc-based URLs until the change to both the paradox plugin version and the fix to ParadoxSignatureDirective.scala are made.

jonas:
I've rebased my PR to only upgrade Paradox. Unless there are any concerns I will merge it tomorrow to unblock you.

eswenson1:
Thanks, @jonas

ktoso:
The other PR was merged 👍

eswenson1:
@jonas I can't get @scaladoc to work.  It appears to work fine with fragments, but the URL I'm trying to make it link to is http://doc.akka.io/api/akka-http/current/akka/http/scaladsl/index.html.   I looked at the source for @scaladoc and compared it to that of @javadoc, and whereas @javadoc uses the path, @scaladoc appears to ignore it (except in the case where scaladoc.version is 2.12).  Also, since I wanted to make the corresponding change to the Java docs, I tried using @javadoc, and it complains that javadoc.base_url is not defined.  It shouldn't care if that is defined, should it, since we're trying to reference the package akka.http, and javadoc.akka.http.base_url *is* defined in build.sbt.  When you fixed paradox for 2.12 support, is is possible that you dropped the use of the path?  The @scaladoc implementation now does "        baseUrl.withEndingSlash.withFragment(link)", whereas the @javadoc implementation does "baseUrl.withEndingSlash.withQuery(path).withFragment(url.getFragment)".

jonas:
To link to a package `@scaladic:[...](akka.http.scaladsl.index)`. If you have ideas how to improve it in Paradox to better support linking to packages in 2.12 we can try to come up with a patch.

eswenson1:
If you use @scaladoc[ScalaDoc](akka.http.scaladsl.index), the resulting URL is

http://doc.akka.io/api/akka-http/current/#akka.http.scaladsl.index

In other words, it treats the parameter as a fragment and always uses a path of /.  We need to have it resolve to: http://doc.akka.io/api/akka-http/current/akka/http/scaladsl/index.html.

jonas:
Is this after rebasing on top of the commit that upgraded Paradox to 0.2.7? If you push your change to the PR branch I can take a look later.

eswenson1:
Yes. Well, I haven't pushed my change yet to my PR branch because I haven't gotten it to work yet.  The @javadoc doesn't work (missing define javadoc.base_url in build.sbt (which shouldn't be necessary), and my change to add the @scaladoc link results in a bad url (fragment). I didn't want it accidentally merged.

But if you look at the 0.2.7 com.lighttbend.paradox.markdown.Directive.scala, you'll see the definition for scaladoc (for non-2.12), ignores the path and only generates a fragment.

jonas:
The only thing that comes to mind then is whether the scala version is set to Scala 2.12:

    sbt ++2.12.0 docs/paradox

eswenson1:
I guess I'm not communicating very well.  Take a look at this code, from Paradox 0.2.7:

```
case class ScaladocDirective(page: Page, variables: Map[String, String])
    extends ApiDocDirective("scaladoc", page, variables) {

  def resolveApiLink(baseUrl: Url, link: String): Url = {
    variables.getOrElse("scaladoc.version", "???").split('.').take(2) match {
      case Array("2", "12") =>
        val url = Url(link).base
        val path = url.getPath.replace('.', '/') + ".html"
        baseUrl / path
      case _ =>
        baseUrl.withEndingSlash.withFragment(link)
    }
  }
}
```
Note that if the scaladoc.version is not 2.12, we don't use (or compute) the path at all.  We generate a URL based on the baseUrl with only the fragment.  This will always produce URLs of the form <base_url>#link.  We need a <base_url>/path#link.

jonas:
>  if the scaladoc.version is not 2.12, we don't use (or compute) the path at all

Yes, this is because Paradox supports linking to both Scaladoc for version 2.12 and 2.11/2.10 which are significantly different:

 - http://doc.akka.io/api/akka/2.4.11/#akka.http.scaladsl.Http$
 - http://doc.akka.io/api/akka-http/10.0.0/akka/http/scaladsl/Http$.html

The Scaladoc for Akka HTTP is generated with 2.12 which as you point out uses a path. Since Paradox slaps a `.html` at the end we need to include `.index` to link to the scala DSL package page. For example with:
```markdown
@javadoc:[Javadoc](akka.http.javadsl.package-summary)
@scaladoc:[Scaladoc](akka.http.scaladsl.index)
```
and
```
sbt ++2.12.0 docs/paradox
```
the generated links look good to me.

ktoso:
Whoa, this turned out to be more tricky than initially anticipated :)
Seems like the last comment by @jonas contains what we want here :)
Please amend and we'll merge :)

Thanks for your patience!

eswenson1:
I've updated my fork (patch-1 branch). I tested with "sbt ++2.12.0 docs/paradox" and it does seem to work. It doesn't work without the ++2.12.0 -- that was the missing piece of information.

docs/src/main/paradox/java/http/index.md
```diff
@@ -26,6 +26,7 @@ akka-http-testkit
 akka-http-jackson
 : Predefined glue-code for (de)serializing custom types from/to JSON with [jackson](https://github.com/FasterXML/jackson)
 
+Akka HTTP API - @javadoc:[Javadoc](akka.http.javadsl.index)
```
jonas: To link to the package page for `akka.http.javadsl` this should be `akka.http.javadsl.package-summary`, which would generate http://doc.akka.io/japi/akka-http/current/?akka/http/javadsl/package-summary.html

ktoso:
I applied @jonas's feedback, though I don't know if this still breaks on 2.12? Can we merge this?

ktoso:
Yay :) (Feel free to merge such if you want @jonas )

ktoso:
Thanks @eswenson1 !

ktoso:
Resolves #623

PR #626, opened 2016-12-07
Upgrade Paradox to version 0.2.7

jonas:
To enable #624 to use the `@scaladoc` and `@javadoc` directives.


build.sbt
```diff
@@ -146,11 +146,16 @@ lazy val docs = project("docs")
       "akka.version" -> Dependencies.akkaVersion,
       "scala.binaryVersion" -> scalaBinaryVersion.value,
       "scala.version" -> scalaVersion.value,
+      "scaladoc.version" -> scalaVersion.value,
```
jonas: This has to be here due to a bug in Paradox. Will try to get a fix in later today.

docs/src/main/paradox/scala/http/index.md
```diff
@@ -1,6 +1,8 @@
 <a id="http-scala"></a>
 # Akka HTTP
 
+Akka HTTP API - @scaladoc:[ScalaDoc](akka.http.scaladsl.index)
```
jonas: This won't work for Scala 2.10 and 2.11

ktoso:
LGTM, thanks! :)

PR #634, opened 2016-12-08
=doc make sure to install TestEventListener for docs tests

jrudolph:


docs/src/test/resources/application.conf
```diff
@@ -0,0 +1,3 @@
+akka {
+   loggers = ["akka.testkit.TestEventListener"]
```
ktoso: please add a comment why we're doing this

docs/src/test/resources/application.conf
```diff
@@ -0,0 +1,3 @@
...
+}
```
ktoso: Probably doesn't matter here, but to be safe always have an empty \n in conf files. 
A missing newline breaks multinode tests due to the way they concat stuff... (so it'd break the perf spec for example etc). Bette safe ;)

ktoso:
I added the newline I complained about ;) https://github.com/akka/akka-http/pull/634/files#r91552388

Issue #636, opened 2016-12-09
Docs. Missing snippet in Extract Directive (ScalaDSL)


jlprat:
The [extract](http://doc.akka.io/docs/akka-http/current/scala/http/routing-dsl/directives/basic-directives/extract.html) directive in ScalaDSL documentation is missing a snippet (typo in `@@snip` id)

jlprat:
Created an issue in Paradox, asking for support of numeric-prefixed-ids: https://github.com/lightbend/paradox/issues/79

jonas:
Fixed via be6e176

PR #640, opened 2016-12-11
Docs. Change snippet id to not start with a number

jlprat:
Issue: #636
Paradox doesn't seem to support snippets whose id starts with a number.
Change id format for offending ones.

### DIFF
--- a/docs/src/main/paradox/scala/http/routing-dsl/directives/basic-directives/extract.md
+++ b/docs/src/main/paradox/scala/http/routing-dsl/directives/basic-directives/extract.md
@@ -15,4 +15,4 @@ See @ref[Providing Values to Inner Routes](index.md#providedirectives) for an ov
 
 ## Example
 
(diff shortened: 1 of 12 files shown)


jrudolph:
LGTM

jrudolph:
I wonder: did the docs build fail with the old identifiers? We should make sure that things like that are checked.

jlprat:
Docs build was successful, I opened an issue in Paradox: https://github.com/lightbend/paradox/issues/79

jlprat:
I was playing around with it, and it looks like, if the `id` starts with 0, the whole identifier checking (if it exists in the source file linked) is not happening at all.

jrudolph:
Ok, thanks for th explanation.

jlprat:
Sure thing

ktoso:
Very weird paradox issue... thanks for noticing and reporting there.

Issue #505, opened 2016-11-09
Make idleTimeout not scare people with "No elements passed in the last 1 minute." errors


chauhraj:
In `akka-http v2.4.11`, I got this following exception:-

```
java.util.concurrent.TimeoutException: No elements passed in the last 1 minute.
        at akka.stream.impl.Timers$IdleTimeoutBidi$$anon$7.onTimer(Timers.scala:160)
        at akka.stream.stage.TimerGraphStageLogic.akka$stream$stage$TimerGraphStageLogic$$onInternalTimer(GraphStage.scala:1152)
        at akka.stream.stage.TimerGraphStageLogic$$anonfun$akka$stream$stage$TimerGraphStageLogic$$getTimerAsyncCallback$1.apply(GraphStage.scala:1141)
        at akka.stream.stage.TimerGraphStageLogic$$anonfun$akka$stream$stage$TimerGraphStageLogic$$getTimerAsyncCallback$1.apply(GraphStage.scala:1141)
        at akka.stream.impl.fusing.GraphInterpreter.runAsyncInput(GraphInterpreter.scala:691)
        at akka.stream.impl.fusing.GraphInterpreterShell.receive(ActorGraphInterpreter.scala:419)
        at akka.stream.impl.fusing.ActorGraphInterpreter.akka$stream$impl$fusing$ActorGraphInterpreter$$processEvent(ActorGraphInterpreter.scala:603)
        at akka.stream.impl.fusing.ActorGraphInterpreter$$anonfun$receive$1.applyOrElse(ActorGraphInterpreter.scala:618)
        at akka.actor.Actor$class.aroundReceive(Actor.scala:484)
        at akka.stream.impl.fusing.ActorGraphInterpreter.aroundReceive(ActorGraphInterpreter.scala:529)
        at akka.actor.ActorCell.receiveMessage(ActorCell.scala:526)
        at akka.actor.ActorCell.invoke(ActorCell.scala:495)
        at akka.dispatch.Mailbox.processMailbox(Mailbox.scala:257)
        at akka.dispatch.Mailbox.run(Mailbox.scala:224)
        at akka.dispatch.Mailbox.exec(Mailbox.scala:234)
        at scala.concurrent.forkjoin.ForkJoinTask.doExec(ForkJoinTask.java:260)
        at scala.concurrent.forkjoin.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1339)
[...]

PR #650, opened 2016-12-12
Idle timeout improvements

ktoso:
Resolves https://github.com/akka/akka-http/issues/505 though also needs a PR in Akka itself,
and then an update to this piece of code

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/HttpConnectionIdleTimeoutBidi.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/HttpConnectionIdleTimeoutBidi.scala
@@ -1,0 +1,44 @@
+/*
+ * Copyright (C) 2015-2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.http.impl.engine
+
+import java.net.InetSocketAddress
+import java.util.concurrent.TimeoutException
+
+import akka.NotUsed
+import akka.stream.scaladsl.{ BidiFlow, Flow }
+import akka.util.ByteString
+
+import scala.concurrent.duration.FiniteDuration
+import scala.util.control.NoStackTrace
+
+/** INTERNAL API */
+private[akka] object HttpConnectionIdleTimeoutBidi {
+  def apply(idleTimeout: FiniteDuration, remoteAddress: Option[InetSocketAddress]): BidiFlow[ByteString, ByteString, ByteString, ByteString, NotUsed] = {
+    val connectionToString = remoteAddress match {
+      case Some(addr) ⇒ s" on connection to [$addr]"
+      case _          ⇒ ""
+    }
+    val ex = new HttpIdleTimeoutException(
+      "HTTP idle-timeout encountered" + connectionToString + ", " +
+        "no bytes passed in the last " + idleTimeout + ". " +
+        "This is configurable by akka.http.[server|client].idle-timeout.", idleTimeout)
+
+    val mapError = Flow[ByteString].mapError({ case t: TimeoutException ⇒ ex })
+
+    val toNetTimeout: BidiFlow[ByteString, ByteString, ByteString, ByteString, NotUsed] =
+      BidiFlow.fromFlows(
+        mapError,
+        Flow[ByteString]
+      )
+    val fromNetTimeout: BidiFlow[ByteString, ByteString, ByteString, ByteString, NotUsed] =
+      toNetTimeout.reversed
+
+    fromNetTimeout atop BidiFlow.bidirectionalIdleTimeout[ByteString, ByteString](idleTimeout) atop toNetTimeout
+  }
+
+}
+
+class HttpIdleTimeoutException(msg: String, timeout: FiniteDuration) extends TimeoutException(msg: String) with NoStackTrace
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/HttpConnectionTimeoutException.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/HttpConnectionTimeoutException.scala
@@ -6,4 +6,5 @@ package akka.http.impl.engine
 
 import scala.util.control.NoStackTrace
 
+@deprecated("Not used. Will be replaced by `akka.stream.impl.io.TcpIdleTimeoutException` in Akka itself", since = "10.0.1")
 class HttpConnectionTimeoutException(msg: String) extends RuntimeException(msg) with NoStackTrace
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
@@ -20,7 +20,6 @@ import akka.stream.TLSProtocol._
 import akka.stream.scaladsl._
 import akka.stream.stage._
 import akka.http.scaladsl.settings.ServerSettings
-import akka.http.impl.engine.HttpConnectionTimeoutException
 import akka.http.impl.engine.parsing.ParserOutput._
 import akka.http.impl.engine.parsing._
 import akka.http.impl.engine.rendering.{ HttpResponseRendererFactory, ResponseRenderingContext, ResponseRenderingOutput }
@@ -231,16 +230,8 @@ private[http] object HttpServerBluePrint {
 
     val responseRendererFactory = new HttpResponseRendererFactory(serverHeader, responseHeaderSizeHint, log)
 
-    val errorHandler: PartialFunction[Throwable, Throwable] = {
-      // idle timeouts should not result in errors in the log. See 19058.
-      case timeout: HttpConnectionTimeoutException ⇒
-        log.debug(s"Closing HttpConnection due to timeout: ${timeout.getMessage}"); timeout
-      case t ⇒ log.error(t, "Outgoing response stream error"); t
-    }
-
     Flow[ResponseRenderingContext]
       .via(responseRendererFactory.renderer.named("renderer"))
-      .via(MapError[ResponseRenderingOutput](errorHandler).named("errorLogger"))
   }
 
   class RequestTimeoutSupport(initialTimeout: Duration)
--- a/akka-http-core/src/main/scala/akka/http/impl/util/package.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/util/package.scala
@@ -82,24 +82,6 @@ package util {
   import akka.stream.{ Attributes, Outlet, Inlet, FlowShape }
   import scala.concurrent.duration.FiniteDuration
 
-  /**
-   * Maps error with the provided function if it is defined for an error or, otherwise, passes it on unchanged.
-   */
-  private[http] final case class MapError[T](f: PartialFunction[Throwable, Throwable]) extends SimpleLinearGraphStage[T] {
-    override def createLogic(attr: Attributes) =
-      new GraphStageLogic(shape) with InHandler with OutHandler {
-        override def onPush(): Unit = push(out, grab(in))
-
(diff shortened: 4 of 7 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/HttpConnectionTimeoutException.scala
```diff
@@ -6,4 +6,5 @@ package akka.http.impl.engine
 
 import scala.util.control.NoStackTrace
 
-class HttpConnectionTimeoutException(msg: String) extends RuntimeException(msg) with NoStackTrace
+@deprecated("Not used. Will be replaced by `akka.stream.impl.io.TcpIdleTimeoutException` in Akka itself", since = "10.0.1")
+final class HttpConnectionTimeoutException(msg: String) extends RuntimeException(msg) with NoStackTrace
```
ktoso: We should enable MiMa here, so this will fail (and I'd whitelist it I guess)

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -385,27 +390,49 @@ private[http] object HttpServerBluePrint {
...
-        override def onUpstreamFinish() =
+
+        override def onUpstreamFinish =
           if (openRequests.isEmpty) completeStage()
           else complete(requestPrepOut)
+
+        override def onUpstreamFailure(ex: Throwable): Unit = {
+          ex match {
+            case t: TimeoutException ⇒
+              // FIXME: this must be `akka.stream.impl.io.TcpIdleTimeoutException` instead! Then improve below exception:
```
ktoso: This is step 2, and can only be done after Akka has my other PR and is released.

ktoso:
Example output:

`[DEBUG] [12/12/2016 17:32:15.932] [ServerTest-akka.actor.default-dispatcher-3] [akka.actor.ActorSystemImpl(ServerTest)] Timeout exception occured while replying to request: [GET /], shutting down connection. Cause: No elements passed in the last 10 seconds.`

ktoso:
The other PR this needs is https://github.com/akka/akka/pull/21989

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -385,27 +390,49 @@ private[http] object HttpServerBluePrint {
...
+              log.debug(
+                "Timeout exception occured while replying to request: [{}], shutting down connection. Cause: {}",
```
jrudolph: Why special case on TimeoutException here?

ktoso: Check the fixme

ktoso: It's because the lock-step with Akka but I wanted to have something here already. We'll handle the specific exception here instead of TimeoutException

jrudolph: I still don't understand the reasoning ;) Why are timeout exceptions special? We should strive for composability so that if we add context to exceptions, it should be added to any kind of exception, not just timeouts.

ktoso: Timeouts have settings, so we should tell people not "no elements passed" because they have no idea what that means, they should instead get "no bytes sent/received on connection during ..., change the setting ... if you don't care". This is impossible to achieve if we don't specialize the exception type.

Thus, the exception here must be more than just a plain timeout exception which could be anything.

ktoso: The timeout originates in Akka infrastructure, so there's no sensible way to attach "akka.http setting..." information in here, thus we need to know by exception type that it was the idle timeout, and not a random timeout, so we can log an useful message.

jrudolph: I totally agree with the result to achieve. But as you say the improvement of the message must be done in akka. The only thing we can do here is to add context.

Once the akka message says `TCP connection was closed because no data was transferred for the configured idle-timeout of 10 seconds.`, I don't think we need to add here anything more than the generic prefix `HTTP connection was closed during handling of the request GET /bla because of a connection error: ...`.

The total error message would then be:

`HTTP connection was closed during handling of the request GET /bla because of a connection error: TCP connection was closed because no data was transferred for the configured idle-timeout of 10 seconds.`

What would you add to that?

ktoso:
Yess, perfect:

`[DEBUG] [12/14/2016 00:38:31.722] [ServerTest-akka.actor.default-dispatcher-4] [akka://ServerTest/system/IO-TCP/selectors/$a/0] New connection accepted`

`[DEBUG] [12/14/2016 00:38:43.072] [ServerTest-akka.actor.default-dispatcher-4] [akka://ServerTest/user/StreamSupervisor-0/flow-1-0-unknown-operation] Aborting tcp connection to /127.0.0.1:65161 because of upstream failure: HTTP idle-timeout encountered, no bytes passed in the last 10 seconds. This is configurable by akka.http.[server|client].idle-timeout.`

`[DEBUG] [12/14/2016 00:38:43.079] [ServerTest-akka.actor.default-dispatcher-4] [akka.actor.ActorSystemImpl(ServerTest)] Timeout exception occurred while replying to request: [GET /], shutting down connection. Cause: HTTP idle-timeout encountered, no bytes passed in the last 10 seconds. This is configurable by akka.http.[server|client].idle-timeout.`

ktoso:
Note to self, check client side.

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
```diff
@@ -67,22 +68,14 @@ class HttpExt(private val config: Config)(implicit val system: ActorSystem) exte
...
-        case t: TimeoutException ⇒ throw new HttpConnectionTimeoutException(t.getMessage)
-      })
-
-      tls.out2 ~> http.in2
-      tls.in1 <~ http.out1
-
-      tls.out1 ~> timeouts.in
+    val serverBidiFlow =
+      settings.idleTimeout match {
+        case t: FiniteDuration ⇒ httpLayer atop tlsStage atop HttpConnectionIdleTimeoutBidi(t, None)
```
ktoso: the change is here

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
```diff
@@ -102,15 +95,17 @@ class HttpExt(private val config: Config)(implicit val system: ActorSystem) exte
...
-    )
+  private def tcpBind(interface: String, port: Int, settings: ServerSettings): Source[Tcp.IncomingConnection, Future[Tcp.ServerBinding]] = {
+    Tcp()
+      .bind(
+        interface,
+        port,
+        settings.backlog,
+        settings.socketOptions,
+        halfClose = false,
+        idleTimeout = Duration.Inf // we knowingly disable idle-timeout on TCP level, as we handle it explicitly in Akka HTTP itself
```
ktoso: and we don't use the TCP timeout, we have our own

akka-http-core/src/main/scala/akka/http/impl/engine/HttpConnectionIdleTimeoutBidi.scala
```diff
@@ -0,0 +1,49 @@
...
+      case Some(addr) ⇒ s" on connection to [$addr]"
+      case _          ⇒ ""
+    }
+    val ex = new HttpIdleTimeoutException(
+      "HTTP idle-timeout encountered" + connectionToString + ", " +
+        "no bytes passed in the last " + idleTimeout + ". " +
+        "This is configurable by akka.http.[server|client].idle-timeout.", idleTimeout)
+
+    // val mapError = Flow[ByteString].mapError { case t: TimeoutException ⇒ ex } // FIXME once Akka 2.4.15
+    val mapError = Flow[ByteString].via(MapError({ case t: TimeoutException ⇒ ex }))
```
viktorklang: Isn't this equivalent to a `recover + throw`?

ktoso: Part of the debate https://github.com/akka/akka/pull/21989#discussion_r92156907

ktoso: No problem, thanks for reviewing.

ktoso:
With https://github.com/akka/akka-http/pull/676 merged this will be totally complete for 10.0.1 :)

akka-http-core/src/main/scala/akka/http/impl/util/package.scala
```diff
@@ -85,6 +85,7 @@ package util {
   /**
    * Maps error with the provided function if it is defined for an error or, otherwise, passes it on unchanged.
    */
+  @deprecated("Replaced with mapError in Akka Streams.", since = "10.0.1")
   private[http] final case class MapError[T](f: PartialFunction[Throwable, Throwable]) extends SimpleLinearGraphStage[T] {
```
ktoso: removing MapError from akka http, it's built in now :)

ktoso:
This is now ready I think.

Would merge after MiMa changes are merged so I can add filters if any are needed. cc @jrudolph

ktoso:
Added mima.

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -385,27 +378,48 @@ private[http] object HttpServerBluePrint {
...
           }
-        override def onUpstreamFinish() =
+
+        override def onUpstreamFinish(): Unit =
           if (openRequests.isEmpty) completeStage()
           else complete(requestPrepOut)
+
+        override def onUpstreamFailure(ex: Throwable): Unit = {
+          ex match {
+            case t: TcpIdleTimeoutException ⇒
```
jrudolph: Is there still a reason to special case here? Or what's the reason to add logging here? Most people won't see debug login anyway, right?

ktoso: Yeah for some reason (check all the tickets with people panicing about the idle-timeout, which was on debug always) people do run with debug on and panic... So yeah, wanted to special case the error message here.

jrudolph: Or if we want to log, couldn't we now do it directly in the custom idle timeout stage?

ktoso: Actually now that I think of it this won't really be thrown since we don't use that timeout on TCP level... I'll have another look at it.

ktoso: Removing this for now

ktoso:
This now gives us `[DEBUG] [12/22/2016 16:40:21.086] [ServerTest-akka.actor.default-dispatcher-4] [akka://ServerTest/user/StreamSupervisor-0/flow-1-0-unknown-operation] Aborting tcp connection to /127.0.0.1:58291 because of upstream failure: HTTP idle-timeout encountered, no bytes passed in the last 6 seconds. This is configurable by akka.http.[server|client].idle-timeout.
`

ktoso:
LGTMing the changes ;-)

jrudolph:
Great. Created https://github.com/akka/akka-http/issues/681 to track adding more context.

PR #652, opened 2016-12-12
=pro replace bintray-sbt by plain maven publish solution

jrudolph:
/cc @2m


jrudolph:
PLS BUILD

2m:
I think this is great if we can publish straight to maven repository. Have you checked if a new version is created in the bintray web ui when published like this?

ktoso:
Works for me, though this means we must update the: https://github.com/akka/akka-http/blob/master/RELEASING.md with the actual steps to be performed for a release now.

ktoso:
Resolved conflicts again, yay or nay? I don't care too much ;)

jonas:
I can rebase it you'd like

ktoso:
Ping, yay or nay?

jrudolph:
Let's stash and close for now. Not using sbt-bintray would be nice but maybe currently not worth the extra effort required.

ktoso:
+1, I agree. We can come back to this in case of some more free time ;)

PR #654, opened 2016-12-13
Add documentation for Raw-Request-URI synthetic header

### LINKED ISSUE #451: S3 doesn't interpret URI paths with `+` signs correctly vs. Akka HTTP normalizing URIs
According to an [AWS forum post](https://forums.aws.amazon.com/thread.jspa?messageID=722673), S3 mistakenly interprets `+` signs in the *path* portion of an URI as a space character. Instead, a file with a `+` sign in its path needs to be accessed by using the percent encoded representation of the plus sign, `%2b`. This is in violation of [RFC 3986](https://tools.ietf.org/html/rfc3986#section-2.3) which mandates that a literal plus sign or its percent encoded form are to be treated as equivalent. (Note, how this is different from the interpretation of the query portion of an http uri in which `+` signs actually *are* to be interpreted as a space character.)

Akka HTTP will never send out the `+` sign as percent encoded because it normalizes URIs on creation. This means that it isn't possible to access a file in S3 that contains a `+` sign in its path using Akka HTTP's URIs. 

As a workaround, you can use the synthetic `Raw-Request-Uri` header to render the URI yourself and force Akka HTTP to send a plus sign in its percent encoded form like this:

```scala
HttpRequest(uri = "/ignored", headers = List(Raw-Request-URI("/a/b%2Bc/")))
```

I put the issue here to help people who also run into this problem. To resolve it we should document the issue more prominently.

easel:
Fixes #451

### DIFF
--- a/docs/src/main/paradox/scala/http/common/http-model.md
+++ b/docs/src/main/paradox/scala/http/common/http-model.md
@@ -52,6 +52,22 @@ All parameters of `HttpRequest.apply` have default values set, so `headers` for 
 if there are none. Many of the parameters types (like `HttpEntity` and `Uri`) define implicit conversions
 for common use cases to simplify the creation of request and response instances.
 
+### Synthetic Headers
+
+In some cases it may be necessary to deviate from fully RFC-Compliant behavior. For instance, Amazon S3 treats 
+the + character in the path part of the URL as a string, even though the RFC specifies that this behavior should
+be limited exclusively to the query portion of the URI.
+
+In order to work around these types of edge cases, Akka HTTP provides for the ability to provide extra, 
+non-standard information to the request via synthetic headers. These headers are not passed to the client
+but are instead consumed by the request engine and used to override default behavior.
+
+For instance, in order to provide a raw request uri, bypassing the default url normalization, you could do the
+following:
+
+    import akka.http.scaladsl.model.headers.`Raw-Request-URI`
+    val req = HttpRequest(uri = "/ignored", headers=List(`Raw-Request-URI`("/a/b%2Bc")))
+
 ## HttpResponse
 
 An `HttpResponse` consists of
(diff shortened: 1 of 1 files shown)


easel:
Throwing this up here for early feedback. I believe it's accurate, but also likely not complete and perhaps not in the right place. Thanks

jlprat:
Don't forget to add it also on the Java doc page :slightly_smiling_face:

easel:
Yeah I have to figure out what the code looks like for that, but will do assuming folks are happy with the content and location on the scala side. Thanks for the reminder @jlprat

ktoso:
OK TO TEST

docs/src/main/paradox/scala/http/common/http-model.md
```diff
@@ -52,6 +52,22 @@ All parameters of `HttpRequest.apply` have default values set, so `headers` for
...
+
+In order to work around these types of edge cases, Akka HTTP provides for the ability to provide extra, 
+non-standard information to the request via synthetic headers. These headers are not passed to the client
+but are instead consumed by the request engine and used to override default behavior.
+
+For instance, in order to provide a raw request uri, bypassing the default url normalization, you could do the
+following:
+
+    import akka.http.scaladsl.model.headers.`Raw-Request-URI`
+    val req = HttpRequest(uri = "/ignored", headers=List(`Raw-Request-URI`("/a/b%2Bc")))
```
ktoso: I thought we usually use ``` for code snippets?

Contents look good though!

ktoso:
Looks good, please port to java doc page :) It'll be similar, just `.create()` calls instead of `()` I think.

easel:
@ktoso great, will do. Let me see if I can figure out the @@@snip business =p

ktoso:
Welcome to 2017, happy new year :)
Looking forward to the updated PR, would like to merge soon 👍

ktoso:
Merging as it, will follow up separately. Thanks!

Issue #620, opened 2016-12-06
Provide directive composition capabilities for Java DSL 


jlprat:
Currently Java DSL only offers route composition. In order to bring the Java DSL at par with the Scala DSL, it would be good to be able to apply composition at `Directive` level.
At the end it should be possible to write:
```
val route =
  path("order" / IntNumber) { id =>
    (get | put) { 
      complete(s"Received request for order $id")
    }
  }
```
like 
```
return path(segment("order").slash(integerSegment()), id -> 
  orElse(Directives::get, Directives::put, () -> 
    complete("Received request for order " + id)
  )
)
```

jlprat:
I could dive into this one as well :)

Issue #438, opened 2016-10-26
Consider @Experimental / @ApiMayChange markers on APIs - LOOKING FOR NAMES


ktoso:

patriknw:
Experimental means so many different things for different people. Shall we try another name?

`@ApiMayChange` ?

ktoso:
Yeah, absolutely let's talk about the names to use here.
I think the Hadoop ecosystem has started using various annotations like that years ago as well, let's see if naming there could inspire us. Names that have good meaning would be best here than the `experimental` which in Akka up until this day even people are confused about

ktoso:
One I'd like to add is something among the lines of `@NotMeantToBeExtended` - specifically for the HTTP model.

jonas:
Here's a link to Spark's API annotations some of which looks to be overlapping: https://github.com/apache/spark/tree/master/common/tags/src/main/java/org/apache/spark/annotation

As for name Spark use `InterfaceStability`.

patriknw:
Something like that looks good. I think we should use same in Akka core. Perhaps Stable is the default if no annotation is given.

jlprat:
@ktoso for the "Not meant to be extended", I know JBoss Drools packs all those classes under the `internal` package. It's pretty clear that if you depend on something under the `internal` package, you can't expect stability, or actually relay in any way on this.

ktoso:
Due to the design of our HTTP Model it is "not meant to be extended" however it is very much public API. It can not be final, due to scaladsl is-a javadsl - this saves impl complexity and allocations :)

Will think and propose names soon, feel free to propose some :)

jlprat:
I see... then the internal makes no sense

ktoso:
Let's start simple:

- `@InternalAPI` - equivalent of current `/** INTERNAL API*/` comments, we'll work on MiMa to be aware of these (!) so no filtering out when touching such classes/methods will be needed
- `@APIMayChange` - To mark new "experimental" APIs, such as the #662 API

These annotations will land immediately in akka.actor, we'll use them from there in Akka HTTP.

Hope this sounds good, cheers!

ktoso:
This will be sovled in Akka itself .
PR for annotations in Akka: https://github.com/akka/akka/pull/22110

PR #662, opened 2016-12-15
Add Composing directives java (#620)

jlprat:
Issue: #620
Add composing capabilities for the Java DSL.
* `anyOf` is the equivalent of Scala's `|`
* `allOf` is the equivalent of Scala's `&`

### DIFF
--- a/akka-http/src/main/scala/akka/http/javadsl/server/Directives.scala
+++ b/akka-http/src/main/scala/akka/http/javadsl/server/Directives.scala
@@ -1,8 +1,10 @@
 /*
- * Copyright (C) 2009-2016 Lightbend Inc. <http://www.lightbend.com>
+ * Copyright (C) 2009-2017 Lightbend Inc. <http://www.lightbend.com>
  */
 
 package akka.http.javadsl.server
+
+import java.util.function.{ BiFunction, Function, Supplier }
 
 import akka.http.impl.util.JavaMapping
 import akka.http.javadsl.server.directives.{ FramedEntityStreamingDirectives, TimeoutDirectives }
@@ -26,4 +28,96 @@ object Directives extends AllDirectives {
 
   @varargs override def getFromBrowseableDirectories(directories: String*): Route =
     super.getFromBrowseableDirectories(directories: _*)
+
+  /**
+   * Composes two 0-argument directives which share the same inner Route. This is equivalent to
+   * `first(inner).orElse(second(inner))`
+   * Usage example:
+   * `anyOf(this::get, this::post, () -> complete("hi")))`
+   * @param first the first 0-argument directive
+   * @param second the second 0-argument directive
+   * @param inner the inner route Producer
+   * @return the resulting route
+   */
+  // @akka.annotation.ApiMayChange // FIXME use the real ones once Akka dependency bumped
+  def anyOf(first: Function[Supplier[Route], Route], second: Function[Supplier[Route], Route], inner: Supplier[Route]): Route = {
+    first.apply(inner).orElse(second.apply(inner))
+  }
+
+  /**
+   * Composes two 1-argument directives (directives whose inner route takes a parameter) which share the same inner route.
+   * This is the same as `first(inner).orElse(second(inner))`.
+   * This can be used also for directives that take an additional parameter if the directive is partially applied, ie. parameter
+   * Usage example:
+   * `anyOf(bindParameter(this::path, "bar"), bindParameter(this::path, "baz"), () -> complete("bar or baz"))`
+   * @param first the first 1-argument directive
+   * @param second the second 1-argument directive
+   * @param inner the inner route that takes one argument
+   * @tparam A the type of the parameter the directives extract and the inner route takes
+   * @return the resulting route
+   */
+  // @akka.annotation.ApiMayChange // FIXME use the real ones once Akka dependency bumped
+  def anyOf[A](first: Function[Function[A, Route], Route], second: Function[Function[A, Route], Route], inner: Function[A, Route]): Route = {
+    first.apply(inner).orElse(second.apply(inner))
+  }
+
+  /**
+   * Nests two 0-argument directives together
+   * Usage example:
+   * `allOf(bindParameter(this::pathPrefix, "alice"), bindParameter(this::path, "bob"), () -> complete("Charlie!"))`
+   * @param first the 0-argument directive (outer one)
+   * @param second the 0-argument directive (inner one)
+   * @param inner the inner route function
+   * @return the resulting route
+   */
+  // @akka.annotation.ApiMayChange // FIXME use the real ones once Akka dependency bumped
+  def allOf(first: Function[Supplier[Route], Route], second: Function[Supplier[Route], Route], inner: Supplier[Route]): Route = {
+    first.apply(new Supplier[Route] {
+      override def get(): Route =
+        second.apply(inner)
+    })
+  }
+
+  /**
+   * Nests two 1-argument directives together
+   * Usage example:
+   * `allOf(this::extractScheme, this::extractMethod, (scheme, method) -> complete("You did a " + method.name() + " using " + scheme))`
+   * @param first the first 1-argument directive (outer one)
+   * @param second the second 1-argument directive (inner one)
+   * @param inner the inner route function that takes 2 different parameters
+   * @tparam A the type extracted from the first directive
+   * @tparam B the type extracted from the second directive
+   * @return the resulting route
+   */
+  // @akka.annotation.ApiMayChange // FIXME use the real ones once Akka dependency bumped
+  def allOf[A, B](first: Function[Function[A, Route], Route], second: Function[Function[B, Route], Route], inner: BiFunction[A, B, Route]): Route = {
+    first.apply(new Function[A, Route] {
+      override def apply(a: A): Route =
+        second.apply(new Function[B, Route] {
+          override def apply(b: B): Route =
+            inner.apply(a, b)
+        })
+    })
+  }
+
+  /**
+   * Nests one 0-argument and 1-argument directives together
+   * Usage example:
+   * `allOf(bindParameter(this::pathPrefix, "guess"), this::extractMethod, method -> complete("You did a " + method.name()))`
+   * @param first the 0-argument directive (outer one)
+   * @param second the 1-argument directive (inner one)
+   * @param inner the inner route function that takes 1 parameter
+   * @tparam A the type extracted from the second directive
+   * @return the resulting route
+   */
+  // @akka.annotation.ApiMayChange // FIXME use the real ones once Akka dependency bumped
+  def allOf[A](first: Function[Supplier[Route], Route], second: Function[Function[A, Route], Route], inner: Function[A, Route]): Route = {
+    first.apply(new Supplier[Route] {
+      override def get(): Route =
+        second.apply(new Function[A, Route] {
+          override def apply(a: A): Route =
+            inner.apply(a)
+        })
+    })
+  }
 }
--- a/docs/src/main/paradox/java/http/routing-dsl/directives/index.md
+++ b/docs/src/main/paradox/java/http/routing-dsl/directives/index.md
@@ -96,11 +96,23 @@ serves as the inner route of the `path` directive. Let's rewrite it in the follo
 
 In this previous example, we combined the `get` and `put` directives into one composed directive and extracted it to its own method, which could be reused anywhere else in our code.
 
+Instead of extracting the composed directives to its own method, we can also use the available `anyOf` combinator. The following code is equivalent to the previous one:
+
+@@snip [DirectiveExamplesTest.java](../../../../../../test/java/docs/http/javadsl/server/DirectiveExamplesTest.java) { #getOrPutUsingAnyOf }
+
+The previous example, tries to complete the route first with a `GET` or with a `PUT` if the first one was rejected. 
+
 In case you are constantly nesting the same directives several times in you code, you could factor them out in their own method and use it everywhere:
 
 @@snip [DirectiveExamplesTest.java](../../../../../../test/java/docs/http/javadsl/server/DirectiveExamplesTest.java) { #composeNesting }
 
-Here we simple created our own combined directive that accepts either `GET` or `PUT` requests, then extracts the method and completes it with an inner route that takes this HTTP method as a parameter.
+Here we simple created our own combined directive that accepts `GET` requests, then extracts the method and completes it with an inner route that takes this HTTP method as a parameter.
+
+Again, instead of extracting own combined directives to its own method, we can make use of the `allOf` combinator. The following code is equivalent to the previous one:
+
+@@snip [DirectiveExamplesTest.java](../../../../../../test/java/docs/http/javadsl/server/DirectiveExamplesTest.java) { #composeNestingAllOf }
+
+In this previous example, the the inner route function provided to `allOf` will be called when the request is a `GET` and with the extracted client IP obtained from the second directive.
 
 As you have already seen in the previous section, you can also use the `route` method defined in `RouteDirectives` as an alternative to `orElse` chaining. Here you can see the first example again, rewritten using `route`:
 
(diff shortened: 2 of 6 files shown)


akka-http/src/main/scala/akka/http/javadsl/server/directives/RouteDirectives.scala
```diff
@@ -46,6 +60,83 @@ abstract class RouteDirectives extends RespondWithDirectives {
...
 
   /**
+   * Composes two 0-argument directives which share the same inner Route. This is equivalent to
+   * first(inner).orElse(second(inner))
+   * @param first the first 0-argument directive
+   * @param second the second 0-argument directive
+   * @param inner the inner route Producer
+   * @return the resulting route
+   */
+  def anyOf(first: JFunction[Supplier[Route], Route], second: JFunction[Supplier[Route], Route], inner: Supplier[Route]): Route = {
```
jlprat: Makes an `OR` with 2 directives that pass no parameters to the inner Route

akka-http/src/main/scala/akka/http/javadsl/server/directives/RouteDirectives.scala
```diff
@@ -46,6 +60,83 @@ abstract class RouteDirectives extends RespondWithDirectives {
...
+   * Composes two 1-argument directives (directives whose inner route takes a parameter) which share the same inner route.
+   * This is the same as first(inner).orElse(second(inner)).
+   * This can be used also for directives that take an additional parameter if the directive is partially applied, ie. parameter
+   * @param first the first 1-argument directive
+   * @param second the second 1-argument directive
+   * @param inner the inner route that takes one argument
+   * @tparam A the type of the parameter the directives extract and the inner route takes
+   * @return the resulting route
+   */
+  def anyOf[A](first: JFunction[JFunction[A, Route], Route], second: JFunction[JFunction[A, Route], Route], inner: JFunction[A, Route]): Route = {
```
jlprat: Makes an `OR` with 2 directives that pass 1 parameter to the inner Route

akka-http/src/main/scala/akka/http/javadsl/server/directives/RouteDirectives.scala
```diff
@@ -46,6 +60,83 @@ abstract class RouteDirectives extends RespondWithDirectives {
...
+  }
+
+  /**
+   * Chains two 0-argument directives together
+   * @param first the 0-argument directive (outer one)
+   * @param second the 0-argument directive (inner one)
+   * @param inner the inner route function
+   * @return the resulting route
+   */
+  def allOf(first: JFunction[Supplier[Route], Route], second: JFunction[Supplier[Route], Route], inner: Supplier[Route]): Route = {
```
jlprat: Makes an `AND` with 2 directives that pass no parameters to the inner Route

akka-http/src/main/scala/akka/http/javadsl/server/directives/RouteDirectives.scala
```diff
@@ -46,6 +60,83 @@ abstract class RouteDirectives extends RespondWithDirectives {
...
+
+  /**
+   * Chains one 0-argument and 1-argument directives together
+   * @param first the 0-argument directive (outer one)
+   * @param second the 1-argument directive (inner one)
+   * @param inner the inner route function that takes 1 parameter
+   * @tparam A the type extracted from the second directive
+   * @return the resulting route
+   */
+  def allOf[A](first: JFunction[Supplier[Route], Route], second: JFunction[JFunction[A, Route], Route], inner: JFunction[A, Route]): Route = {
```
jlprat: Makes an AND with 2 directives where the first passes no parameters to the inner Route and the second passes 1 parameter to it

docs/src/test/java/docs/http/javadsl/ComposeDirectivesExampleTest.java
```diff
@@ -0,0 +1,71 @@
...
+    BiFunction<PathMatcher1<Integer>, Function<Integer,Route>, Route> pathWithInteger = this::path;
+
+    return route(
+      //anyOf examples
+      path("hello", () ->
+        anyOf(this::get, this::put, () -> complete("<h1>Say hello to akka-http</h1>"))),
+      path("foo", () ->
+        anyOf(partial(this::parameter, "foo"), partial(this::parameter, "bar"), (String param) -> complete("param is " + param))
+      ),
+      anyOf(partial(this::path, "bar"), partial(this::path, "baz"), () -> complete("bar - baz")),
```
jlprat: Examples of using `anyOf`

docs/src/test/java/docs/http/javadsl/ComposeDirectivesExampleTest.java
```diff
@@ -0,0 +1,71 @@
...
+
+      //allOf examples
+      allOf(partial(this::pathPrefix, "alice"), partial(this::path, "bob"), () -> complete("Charlie!")),
+      allOf(partial(this::pathPrefix, "guess"), this::extractMethod, method -> complete("You did a " + method.name())),
+      path("two", () ->
+        allOf(this::extractScheme, this::extractMethod, (scheme, method) -> complete("You did a " + method.name() + " using " + scheme))
+      ),
+      allOf(partial(this::pathPrefix, "number"), partial(pathWithInteger, integerSegment()), x -> complete("Number is " + x))
```
jlprat: Examples of using `allOf`

jlprat:
There are two `anyOf` methods:
* one for directives that pass no parameter to the inner Route
* one for directives that pass exactly 1 parameter to the inner Route
There are three `allOf` methods:
* one that nest directives that pass no parameter to the inner Route
* one that nest directives that pass exactly 1 parameter to the inner Route
* one that nest one directive that passes no paramter to the inner Route and one directive that passes 1 parameter to the inner Route

akka-http/src/main/scala/akka/http/javadsl/server/Directives.scala
```diff
@@ -26,4 +28,96 @@ object Directives extends AllDirectives {
 
   @varargs override def getFromBrowseableDirectories(directories: String*): Route =
     super.getFromBrowseableDirectories(directories: _*)
+
+  /**
+   * Partially applying functions for the Java World!
```
ktoso: A bit too silly comment :)
We should explain what this does so people can understand why they'd use it.
Perhaps an in-line example too?

Since some Java developers may feel this is a very new concept.

akka-http/src/main/scala/akka/http/javadsl/server/Directives.scala
```diff
@@ -26,4 +28,96 @@ object Directives extends AllDirectives {
...
+   * @param second the 0-argument directive (inner one)
+   * @param inner the inner route function
+   * @return the resulting route
+   */
+  def allOf(first: Function[Supplier[Route], Route], second: Function[Supplier[Route], Route], inner: Supplier[Route]): Route = {
+    first.apply(new Supplier[Route] {
+      override def get(): Route =
+        second.apply(inner)
+    })
+  }
```
ktoso: Will be pretty decent to write once we can drop 2.11, with the lambda interop... ;)

akka-http/src/main/scala/akka/http/javadsl/server/Directives.scala
jlprat: Sure, I will give my best explaining what it means and the in-line example is a good idea :slightly_smiling_face:

ktoso: Make the 2 param function into a 1 param function by binding the 1st param already to the `a` argument.

Something like that probably :)

docs/src/test/java/docs/http/javadsl/ComposeDirectivesExampleTest.java
ktoso: Ok examples, would be good to add reference docs which would include this example where we can explain what's happening here. I'd also potentially break the line after `() -> ` to indicate when the "inside" is.

docs/src/test/java/docs/http/javadsl/ComposeDirectivesExampleTest.java
ktoso: Ok examples, would be good to add reference docs which would include this example where we can explain what's happening here. I'd also potentially break the line after `() -> ` to indicate when the "inside" is.

docs/src/test/java/docs/http/javadsl/ComposeDirectivesExampleTest.java
ktoso: I'm iffy on the naming here but example is good I guess. Needs reference docs so people can really grasp it hm.
I don't have a better name to propose though hm... I'd ponder it a second before we merge, anyone have ideas like/dislike the allOf/anyOf's? @akka/akka-http-team

ktoso:
I think this is indeed shaping up nicely!

akka-http/src/main/scala/akka/http/javadsl/server/Directives.scala
```diff
@@ -26,4 +28,96 @@ object Directives extends AllDirectives {
...
+   * Composes two 1-argument directives (directives whose inner route takes a parameter) which share the same inner route.
+   * This is the same as first(inner).orElse(second(inner)).
+   * This can be used also for directives that take an additional parameter if the directive is partially applied, ie. parameter
+   * @param first the first 1-argument directive
+   * @param second the second 1-argument directive
+   * @param inner the inner route that takes one argument
+   * @tparam A the type of the parameter the directives extract and the inner route takes
+   * @return the resulting route
+   */
+  def anyOf[A](first: Function[Function[A, Route], Route], second: Function[Function[A, Route], Route], inner: Function[A, Route]): Route = {
```
ktoso: I think these may be hard to grasp so showing an example (like in the examples in code below) would save people a ton of headaches, even if it's a pain for us to maintain :)

ktoso:
Very cool, needs reference docs now but otherwise looking fine

jlprat:
My idea was also add the docs in this PR with examples, after somebody looked at it.
I intend to convert the current Scala DSL examples to the Java one. Also add the type checking section about it.

akka-http/src/main/scala/akka/http/javadsl/server/Directives.scala
jlprat: I can add one example for each `anyOf` and `allOf`

docs/src/test/java/docs/http/javadsl/ComposeDirectivesExampleTest.java
jlprat: Yes, naming is hard, I'm open to suggestions.  My shortlist was:

* `or` `and`
* `orElse` `both`

akka-http/src/main/scala/akka/http/javadsl/server/directives/RouteDirectives.scala
```diff
@@ -35,6 +34,21 @@ abstract class RouteDirectives extends RespondWithDirectives {
...
   /**
+   * Partially applying functions for the Java World!
+   * @param f the function to partially apply
+   * @param a the first parameter to partially apply
+   * @tparam A the type of the applied parameter
+   * @tparam B the type of the second parameter
+   * @tparam R the type of the return
+   * @return the function partially applied
+   */
+  def partial[A, B, R](f: BiFunction[A, B, R], a: A): JFunction[B, R] = {
```
jypma: Since this is not a directive, and not even http-specific, it should probably go to a different class/package.

It also should be properly documented, since the average Java developer wouldn't know what "partially applying" a function is. Actually, isn't the act of partially applying called [_currying_](https://en.wikipedia.org/wiki/Currying) or am I mixing sth up? In that case, `curry` would be a more apt method name.

akka-http/src/main/scala/akka/http/javadsl/server/directives/RouteDirectives.scala
jypma: The arguments can take type bound `JFunction[_ >: A]` for both, but double-confirm that those bounds make it into Java; otherwise, we'll have to write these in `.java` files.

akka-http/src/main/scala/akka/http/javadsl/server/directives/RouteDirectives.scala
jypma: to me they're not really "chained", they're nested. That's also what the implementation does. So perhaps call it `nest` or `nestAll`?

docs/src/test/java/docs/http/javadsl/ComposeDirectivesExampleTest.java
jypma: Adding a suggestion to this: `allOf` -> `nest` (see my code comments above)

akka-http/src/main/scala/akka/http/javadsl/server/Directives.scala
jypma: Actually, isn't the act of partially applying called [_currying_](https://en.wikipedia.org/wiki/Currying) or am I mixing sth up? In that case, `curry` would be a more apt method name.

docs/src/test/java/docs/http/javadsl/ComposeDirectivesExampleTest.java
jlprat: `nest` sounds nice :)

jypma:
(updated my comments, originally github decided to show me an old version)

akka-http/src/main/scala/akka/http/javadsl/server/Directives.scala
jlprat: I would say it's partially application: https://en.wikipedia.org/wiki/Partial_application
>Intuitively, partial function application says "if you fix the first arguments of the function, you get a function of the remaining arguments"

jypma: ok, agreed.

jlprat: And about the comment about this is not the right place, I completely agree with you @jypma (this was done in an older version), Any idea where this could be placed?

jypma: Maybe in [`akka-http/akka.http.javadsl.common`](https://github.com/akka/akka-http/tree/master/akka-http/src/main/java/akka/http/javadsl/common)`.JavaFunctions`? @ktoso what do you think?

jlprat:
PLS BUILD

jlprat:
Still pending things:
* Decide on better names for `anyOf` and `allOf`
* Find a place to put `partial` function (maybe akka-http/akka.http.javadsl.common.JavaFunctions)

docs/src/main/paradox/java/http/routing-dsl/directives/index.md
```diff
@@ -118,3 +130,31 @@ is in fact the most readable one.
...
+When you combine directives with `anyOf` and `allOf` methods the routing DSL makes sure that all extractions work as
+expected and logical constraints are enforced at compile-time.
+
+For example you cannot call `anyOf`  with a directive producing an extraction with one that doesn't:
+
+```java
+anyOf(this::get, this::extractClientIP, routeProvider) // doesn't compile
+```
+
+Also the number of extractions and their types have to match up:
```
ktoso: please explain why we're using partial, it can be rather confusing at first look :)

docs/src/main/paradox/java/http/routing-dsl/directives/index.md
```diff
@@ -118,3 +130,31 @@ is in fact the most readable one.
 
 Still, the purpose of the exercise presented here is to show you how flexible directives can be and how you can
 use their power to define your web service behavior at the level of abstraction that is right for **your** application.
+
+## Type Safety of Directives
```
ktoso: Is it a section about type safety or about combining directives?

ktoso:
Looking good, small comments about the docs, they need to explain a bit more hand-holdingly.

anyOf / allOf I guess is OK, I can't think of very good alternatives, everything is somewhat of a tradeoff hm...

About class name, how about `PartialApplication`? We could then have multiple versions in there, like `PartialApplication.bindParameter(2paramthingy, "get"): 1 pram thingy` and even `bindParameter2()` and alternatives

jlprat:
Having a dedicated class for `PartialApplication` makes sense. I will put it in there. `bindParameter` might be also more understandable than bind.

jlprat:
PLS BUILD

jlprat:
PLS BUILD

jlprat:
PLS BUILD

jlprat:
One thing that crossed my mind, are these composer functions a good candidate for the "API-may-change" markers (#438)

ktoso:
Yeah good point, let's take on that discussion now then

ktoso:
PR for annotations in Akka: https://github.com/akka/akka/pull/22110

jlprat:
PLS BUILD

akka-http/src/main/scala/akka/http/javadsl/server/Directives.scala
```diff
@@ -26,4 +28,96 @@ object Directives extends AllDirectives {
...
+   * Nests two 0-argument directives together
+   * Usage example:
+   * `allOf(bindParameter(this::pathPrefix, "alice"), bindParameter(this::path, "bob"), () -> complete("Charlie!"))`
+   * @param first the 0-argument directive (outer one)
+   * @param second the 0-argument directive (inner one)
+   * @param inner the inner route function
+   * @return the resulting route
+   */
+  // @akka.annotation.ApiMayChange // FIXME use the real ones once Akka dependency bumped
+  def allOf(first: Function[Supplier[Route], Route], second: Function[Supplier[Route], Route], inner: Supplier[Route]): Route = {
```
jrudolph: So, 0 + 0, 0 + 1, and 1 + 1, arguments are supported so far. Maybe that's enough.

jlprat: Exactly, we can add as many other combinations as we want. Like 0 + 0 + 0, 1 + 1 + 1, 0 + 0 + 1, 0 + 1 + 1 ...

jlprat: However, this explodes quite soon (a lot of combinations)

jrudolph: Obviously, we have implicits in Scala for a reason ;)

ktoso: Yeah we can add more later. Would use the sbt-boilerplate plugin then - note that we do this in many places already.

I agree it's likely enough for now

jrudolph:
Thanks, @jlprat. The PR LGTM as it is. 

IMO the use for `anyOf` is pretty clear. I'm not so sure about why you would use `allOf` at all? In the Scala version, the main benefit of using `&` is that you can compose a directive and give it a name. With `allOf` you directly have to supply the inner route, so the benefit is not so clear. Would it make sense to provide `allOf` in curried form instead? I.e. returning a `Function[Supplier[Route], Route]` instead? That would allow the user to put the value into a variable? We could then call `Function[Supplier[Route], Route]` "`Directive0`". But maybe that's something for another PR...

jlprat:
My idea for the `allOf` would b for the cases when you continuously use the same pair of nested directives. And you only want to perform some action (inner route) when both are met. `allOf` lets you save one level of embedding:
```java
parameter("foo", foo -> 
  parameter("bar", bar -> 
    complete("I have both " + foo " and " + bar)
));
```
vs
```java
allOf(bindParameter(this::parameter, "foo"), bindParameter(this::parameter, "bar"), (foo, bar) -> 
    complete("I have both " + foo " and " + bar)
));
```
If you extract the inner route to its own method it is easier with the second approach.

jlprat:
I can create a new issue to provide (overloaded?) methods for `allOf` that return a function instead of the Route as well :)

akka-http/src/main/scala/akka/http/javadsl/server/Directives.scala
jrudolph: I don't think it's feasible to produce much more combinations. As you say, it explodes quite soon ;)

ktoso:
Overloading directly causes trouble for java - the error messages become really crazy (literally pages of text if 1 para did not match and there's 22 overloads - only 1 overload possibly makes sense but it will print all anyway).

LGTM, thanks!

PR #663, opened 2016-12-16
moves gzip/deflate encoding to akka-http-core

schmitch:
Actually all of these are useful even for a akka-http-core since gzip is a really essential feature of modern http.
Since the structure remains the same and everything in coding didn't need anything from the `akka-http` API this change is backwards compatible.


ktoso:
This is breaking. We can't do such a thing. For example the osgi artifacts would be broken now.
It is not quite worth breaking things to do this move.

If you want to move it regardless then we need to:
- create new classes in core
- mark the current ones deprecated
- make the current ones extend or use the ones in core
- we can remove the deprecated ones in 10.1.0 then

The cost of success is stability and these pains ;)

schmitch:
well no need for that, on akka 2.4.15 :)

jrudolph:
We can instead remove a few of these classes which by now have moved into akka proper.

schmitch:
Well not sure if necessary. The only thing, which is bad, is that at the moment you can't specify a Buffer size.
I actually have two PRs on Play! open, that actually convert do Gzip/Gunzip with the `Compression` class:

https://github.com/playframework/playframework/pull/6810/files#diff-34dcc5f7cb05dfce782beeff65924fc0R12
https://github.com/playframework/playframework/pull/6813/files

So I guess `Compression` is enough for the most people since the Classes in akka-http only checked for the compression header and removed it (and on the compression side they checked if the content is compressible) so I at least play does not need the stuff, since it's baked into akka-stream now.

jrudolph:
I'd be interested to know what you need to specify the buffer sizes for?

ktoso:
> We can instead remove a few of these classes which by now have moved into akka proper.

Yeah all the internal ones, @jrudolph or @schmitch would you PR that? :)

schmitch:
Well we had a configurable Buffer for gzip even before akka-stream. Maybe because we used (GZIPOutputStream.html)[https://docs.oracle.com/javase/7/docs/api/java/util/zip/GZIPOutputStream.html] where you *can* specify a buffer. Or maybe somebody wanted that to be configurable to reduce memory pressure, I'm not sure, since it's not clear from the commits.

Is it possible to remove things? A lot of stuff is public there (i.e. all Compression/Decompression classes) Shouldn't this be rather deprecated and a ticket created to remove it in 10.1.0?

ktoso:
Yes, first deprecation.

jrudolph:
> Well we had a configurable Buffer for gzip even before akka-stream.

Ok, that might make sense. The chunker you are using shouldn't have any effect on memory usage, though, other than creating even more garbage. After all, the data is already in the pipeline.

Issue #639, opened 2016-12-10
Extract a header value or return default


zettelmj:
I some situation it would be handy to for the headerValue* directives to allow to specify a default value that is used if the header is not present.

for headerValueByName this could look like this:

`def headerValueByName(headerName: String, default: String): Directive1[String] =
    headerValue(optionalValue(headerName.toLowerCase)) | provide(default)`

zettelmj:
Hi @jrudolph,

that looks way better than what i tried over the weekend. Passing a function down through the directives makes all of the look very convoluted. Additionally it changes a bunch of signatures. 
I'll give the "provide" directive a try. 

Cheers
Jens

ktoso:
Would you have some time to add this pattern to our documentation, in 

- http://doc.akka.io/docs/akka-http/current/scala/http/routing-dsl/directives/header-directives/headerValue.html
- https://github.com/akka/akka-http/blob/master/docs/src/main/paradox/java/http/routing-dsl/directives/header-directives/headerValue.md

That would be very nice of you :) A way of saying thanks in a way :-)

zettelmj:
@ktoso i will after implementing.

jlprat:
I would say for the Java DSL is not that straight forward, unless https://github.com/akka/akka-http/issues/620 is done

jrudolph:
@jlprat agreed.

zettelmj:
@ktoso just putting it into headerValue feels a bit strange cause it is effectively just an application of http://doc.akka.io/docs/akka-http/current/scala/http/routing-dsl/directives/index.html#composing-directives.

There are some examples in composing-directives but they aren't covering this scenario.
Maybe a sub page of examples what can be done by composing patterns would be good?!

Thank you
Jens

ktoso:
Page with patterns would also be good, but this one I think is basically a headerValue "or else..." which is why I think it makes sense to add to its page. We have examples of directives which have such additional documentation (simply as section inside their docs) :)

jrudolph:
In #666, @zettelmj added the example to the Scala documentation. Would make sense to add something similar to the Java documented, not least for consistency.

PR #666, opened 2016-12-18
Add an example that shows how headerValue and provide can be used to …

zettelmj:
…return the header or if the header is not present use a default

This should be the solution for #639


ktoso:
OK TO TEST

jrudolph:
Thanks, @zettelmj. LGTM.

PR #677, opened 2016-12-21
=pro #222 reenable mima

### LINKED ISSUE #222: Enable MiMa
Enable the mima plugin for binary compat checking

jrudolph:
The basic structure was copied over from akka. Now problems are not defined
in source any more but in separate files to be put into the `project/mima-filters`
directory. This should speed up building and updating the project.

It is still required to reload the project after such a file was added or
changed because the sbt key it provides is a `SettingKey` which is only
evaluated once.

Fixes #222.


project/Mima.scala
```diff
@@ -0,0 +1,65 @@
...
+    def parseFile(file: File): (String, Seq[ProblemFilter]) = {
+      val version = file.getName.dropRight(fileExtension.size)
+
+      val excludes = Source.fromFile(file).getLines().filterNot(str => str.trim.isEmpty || str.trim.startsWith("#")).flatMap {
+        case ExclusionPattern(className, target) => Some(ProblemFilters.exclude(className, target))
+        case FilterAnyProblemPattern(target) => Some(FilterAnyProblem(target))
+        case FilterAnyProblemStartingWithPattern(target) => Some(FilterAnyProblemStartingWith(target))
+        case x =>
+          println(s"Couldn't parse '$x'")
+          None
```
patriknw: shouldn't this fail the build?

project/mima-filters/10.0.0.excludes
```diff
@@ -0,0 +1,7 @@
+# Change in internal class
+FilterAnyProblemStartingWith("akka.http.impl.util.One2OneBidiFlow")
+
+# Issue #673, added extra methods to RejectionHandler trait. Would blow up if a third-party lib compiled against 10.0.0
+# would provide a RejectionHandler and someone would call `mapRejectionResponse` against it from 10.0.1
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.scaladsl.server.RejectionHandler#Handler.mapResponse")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.scaladsl.server.RejectionHandler.mapRejectionResponse")
```
ktoso: @2m will likely be interested in this too

project/mima-filters/10.0.0.excludes
```diff
@@ -0,0 +1,7 @@
+# Change in internal class
```
ktoso: I think it'd be good to explain with a comment here that this file means "changes from 10.0.0" as it may be confusing for people if they should add things to 10.0.1 or 10.0.0 sometimes?

ktoso:
Oh one thing, we should add a note about it in CONTRIBUTING.md

project/mima-filters/10.0.0.excludes
jrudolph: Yep, good point.

2m:
Nice. This will take care of the merge conflicts.

What do you think about having this file per subproject? Since akka-http is starting from almost a clean BC filters slate, it is possible to have per subproject filter files here.

Otherwise [pr validator](https://github.com/akka/akka-http/blob/master/project/ValidatePullRequest.scala#L159) will need to be changed to ignore this file.

jrudolph:
> What do you think about having this file per subproject? Since akka-http is starting from almost a clean BC filters slate, it is possible to have per subproject filter files here.

We could do that, but I also somewhat like having an overview over what was ignored per version.

> Otherwise pr validator will need to be changed to ignore this file.

Yes, that's right. (We are not using pr-validator so far, though.)

patriknw:
When this is done it would be great to port it back to akka/akka. The old filters can remain in MiMa.scala.

jrudolph:
@2m:

> What do you think about having this file per subproject? Since akka-http is starting from almost a clean BC filters slate, it is possible to have per subproject filter files here.

I changed that with a new commit.

@patriknw 

> When this is done it would be great to port it back to akka/akka. The old filters can remain in MiMa.scala.

The benefit of moving the filters to extra files would probably be faster compilation times for the project build. It could also be done incrementally.

project/Mima.scala
jrudolph: It now fails *building the project*. This is not as nice as failing only when actually running mima but cannot be helped as long as the key in question is a `settingKey` and not a `taskKey`. I proposed to change that in https://github.com/typesafehub/migration-manager/pull/157

akka-http-core/src/main/mima-filters/10.0.0.backwards.excludes
```diff
@@ -0,0 +1,4 @@
+# Mima filters needed to check newer versions against 10.0.0
+
+# Internal class change
+FilterAnyProblemStartingWith("akka.http.impl.util.One2OneBidiFlow")
```
ktoso: ok, good name and comment. Also great to have them in projects.

2m:
Good stuff. I think instead of porting this to akka/akka, we should add this feature to sbt mima plugin itself when the dust settles. In such way keeping custom duplicated build code at minimum and pushing good practice to other users of the plugin.

ktoso:
Yeah, putting it into MiMa itself sounds like an interesting idea.

jrudolph:
@2m, yes, I agree, would be nice to have this in MiMa proper.

Issue #658, opened 2016-12-14
Directive for extractScheme does not work with https


michael-smith:
The `extractScheme` directive always returns "http" even when "https" is being used.

A basic example using an https context results in "Scheme: http" ...
```scala
package org.example

import scala.collection.immutable.Seq
import javax.net.ssl._
import java.security.cert.X509Certificate
import akka.actor.ActorSystem
import akka.stream._
import akka.http._
import akka.http.scaladsl.server._
import akka.http.scaladsl._
import akka.http.scaladsl.model._
import akka.http.scaladsl.server._
import akka.http.scaladsl.server.Directives._

object Main {

  private val httpsConnectionContext = ConnectionContext.https(createSslContext)

  def main(args: Array[String]) {
    implicit val system = ActorSystem("my-system")
    implicit val materializer = ActorMaterializer()
    implicit val executionContext = system.dispatcher

    val route =
      extractScheme { scheme =>
        get {
          println("Scheme is: " + scheme)
          complete(HttpEntity(ContentTypes.`text/html(UTF-8)`, s"<h1>Scheme: ${scheme}</h1>"))
        }
      }

    val bindingFuture = Http().bindAndHandle(route, "localhost", 8443, connectionContext = httpsConnectionContext)

    println(s"Server online at https://localhost:8443/\nPress RETURN to stop...")
    scala.io.StdIn.readLine()
    bindingFuture
      .flatMap(_.unbind())
      .onComplete(_ => system.terminate())
  }

  private def createSslContext: SSLContext = {
    val context = SSLContext.getInstance("TLSv1")
    context.init(keyManagers(context), Array.empty, null)
    context
  }

  private final def keyManagers(sslContext: SSLContext): Array[KeyManager] = {
    import java.security._
    val kmf = KeyManagerFactory.getInstance("SunX509")
    val ks = KeyStore.getInstance("JKS")
    val ksPass = "changeit".toCharArray()
    var kms: Array[KeyManager] = null
    val is = getClass.getResourceAsStream("keystore.jks")
    try {
      ks.load(is, ksPass)
      kmf.init(ks, ksPass)
      kms = kmf.getKeyManagers
    } finally { is.close }
    kms
  }
}
```

zettelmj:
Looks like the information if the session is encrypted or not is lost here: https://github.com/akka/akka-http/blob/master/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpMessageParser.scala#L56
[...]

PR #678, opened 2016-12-21
[WIP] Pass secure from http to blueprint

zettelmj:


zettelmj:
This is an attempt to fix #658.
I am currently a bit lost as to how that is supposed to be tested.
It seems that https testing is done manually? Would it be enough to test isSecureConnection?

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
```diff
@@ -61,11 +61,16 @@ class HttpExt(private val config: Config)(implicit val system: ActorSystem) exte
...
+    case hctx: HttpsConnectionContext ⇒ true
+    case _                            ⇒ false
+  }
+
   private def fuseServerBidiFlow(
     settings:          ServerSettings,
     connectionContext: ConnectionContext,
     log:               LoggingAdapter)(implicit mat: Materializer): ServerLayerBidiFlow = {
-    val httpLayer = serverLayer(settings, None, log)
+    val httpLayer = serverLayer(settings, None, log, isSecureConnection(connectionContext))
```
jrudolph: You can use `connectionContext.isSecure`, instead.

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
```diff
@@ -252,10 +257,11 @@ class HttpExt(private val config: Config)(implicit val system: ActorSystem) exte
    * this layer produces if the `akka.http.server.remote-address-header` configuration option is enabled.
    */
   def serverLayer(
```
jrudolph: This is public API so we need to find a binary compatible solution for that.

jrudolph:
OK TO TEST

jrudolph:
Thanks, @zettelmj. This is looking good.

> It seems that https testing is done manually? Would it be enough to test isSecureConnection?

Would be nice to have a test, but as you say we haven't built the infrastructure to add such a test in a simple fashion. I'll think of something.

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
zettelmj: Not really nice but there could be a secureAwareServerLayer that has the extended signature and is called by serverLayer. But this would have to be exposed in the public api aswell.

zettelmj:
Hi @jrudolph,

 this test is failing: info] Passed: Total 1134, Failed 0, Errors 0, Passed 1134, Pending 38
java.lang.RuntimeException: akka-http-core: Binary compatibility check failed!

but i am unsure if that is: wooow wait we are binary incompatible now or if it is you woops that has changed.

Thank you
Jens

jrudolph:
Here are the incompatibilities as found:

```
[info] akka-http-core: found 3 potential binary incompatibilities while checking against com.typesafe.akka:akka-http-core_2.11:10.0.0  (filtered 18)
[error]  * method parsingRendering(akka.http.scaladsl.settings.ServerSettings,akka.event.LoggingAdapter)akka.stream.scaladsl.BidiFlow in object akka.http.impl.engine.server.HttpServerBluePrint does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.impl.engine.server.HttpServerBluePrint.parsingRendering")
[error]  * method apply(akka.http.scaladsl.settings.ServerSettings,akka.event.LoggingAdapter)akka.stream.scaladsl.BidiFlow in object akka.http.impl.engine.server.HttpServerBluePrint does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.impl.engine.server.HttpServerBluePrint.apply")
[error]  * method parsing(akka.http.scaladsl.settings.ServerSettings,akka.event.LoggingAdapter)akka.stream.scaladsl.Flow in object akka.http.impl.engine.server.HttpServerBluePrint does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.impl.engine.server.HttpServerBluePrint.parsing")
```

jrudolph:
So, these are only changes to internal code which are fine. You need to add the filters indicated in the error message to `akka-http-core/src/mima-filters/10.0.1.backwards.excludes`.

zettelmj:
@jrudolph thank you that works

ktoso:
Squashed manually and merged, thanks!
https://github.com/akka/akka-http/commit/45ce8ad260053ebb7450349e2e2650a763aa2776

PR #688, opened 2016-12-26
=doc Remove akka.pattern package from Scaladoc (#684)

### LINKED ISSUE #684: Investigate how akka.pattern ends up in ScalaDoc
See http://doc.akka.io/api/akka-http/10.0.1/akka/pattern/index.html

jonas:
Fix #684 by using -skip-packages to ignore the package from generated
Scaladoc.

### DIFF
--- a/project/Doc.scala
+++ b/project/Doc.scala
@@ -42,7 +42,13 @@ object Scaladoc extends AutoPlugin {
 
   def scaladocOptions(ver: String, base: File): List[String] = {
     val urlString = GitHub.url(ver) + "/€{FILE_PATH}.scala"
-    val opts = List("-implicits", "-groups", "-doc-source-url", urlString, "-sourcepath", base.getAbsolutePath)
+    val opts = List(
+      "-implicits",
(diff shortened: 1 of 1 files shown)


ktoso:
Hmm good fix. But I don't understand how it ended up here in the first place hmm. Did I miss an explanation somewhere?

jonas:
I've seen an even weirder case where a scalar package ended up in the circe Scaladoc and it depends on cats instead of scalaz. Anyway, I suggest any investigation should be done in the context of scaladoc itself. I can check later if this could somehow be caused by sbt-unidoc.

ktoso:
Yeah it seems so, thanks for explaining. Lets log a ticket in scaladoc

jonas:
Logged a comment here: https://issues.scala-lang.org/browse/SI-10028?focusedCommentId=76091&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-76091

jrudolph:
Thanks, @jonas.

PR #695, opened 2016-12-28
=htc content-disposition with charset get's parsed/rendered wrong

schmitch:
Actually there's a small mistake in the Content-Disposition parser, because of that it will actually render a Content-Disposition header with a filename and a charset wrong and will print `filename=utf-8''name` instead of `filename*=utf-8''name`.

Thanks @shkhln for catching that.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/model/parser/ContentDispositionHeader.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/model/parser/ContentDispositionHeader.scala
@@ -26,7 +26,7 @@ private[parser] trait ContentDispositionHeader { this: Parser with CommonRules w
 
   def `filename-parm` = rule(
     ignoreCase("filename") ~ OWS ~ ws('=') ~ push("filename") ~ word
-      | ignoreCase("filename*") ~ OWS ~ ws('=') ~ push("filename") ~ `ext-value`)
+      | ignoreCase("filename*") ~ OWS ~ ws('=') ~ push("filename*") ~ `ext-value`)
 
   def `disp-ext-parm` = rule(
     token ~ ws('=') ~ word
--- a/akka-http-core/src/test/scala/akka/http/impl/model/parser/HttpHeaderSpec.scala
+++ b/akka-http-core/src/test/scala/akka/http/impl/model/parser/HttpHeaderSpec.scala
@@ -676,6 +676,10 @@ class HttpHeaderSpec extends FreeSpec with Matchers {
       HeaderParser.parseFull("location", "http://example.org/?abc=def=ghi", HeaderParser.Settings(uriParsingMode = Uri.ParsingMode.Relaxed)) shouldEqual
         Right(Location(targetUri))
     }
+    "correctly parse a Content-Disposition header with and without charset" in {
+      HttpHeader.parse("Content-Disposition", "inline; filename=\"fn\"; filename*=utf-8''fn") shouldEqual
+        ParsingResult.Ok(`Content-Disposition`(ContentDispositionTypes.inline, Map("filename" → "fn", "filename*" → "utf-8''fn")), List())
+    }
   }
 
   implicit class TestLine(line: String) {


jrudolph:
I think in the end we would want to treat the difference between `filename` and `filename*` as an implementation detail which we wouldn't like to push into the model.

schmitch:
Well I think it's fine to push both into the model for several reasons.
First of all, people who don't care about the charset, actually *just* want to use the filename without further parsing. (that's not possible since the actual filename would be `utf-8''filename` so you actually need to do something anyway.

I think the preferable way would be that the Map holds **all real** values and we might add a special `filename` method to the `Content-Disposition` header which **can** actually returns the correct filename according to https://tools.ietf.org/html/rfc6266#section-4.3 and applies the charset so that the filename will always be without the charset.

I mean, else it would mean that when somebody parses a Content-Disposition header and rerenders it that values would be lost?!

jrudolph:
> I mean, else it would mean that when somebody parses a Content-Disposition header and rerenders it that values would be lost?!

No, it should be rendered the "best way" (whatever that is) ;)

Right, now, I haven't seen a client actually using the `filename*` format. The last time I checked browsers just put the UTF-8 data into the normal `filename` field.

schmitch:
Well I've seen client's putting utf-8 characters into the `filename`, yes. But they didn't prefixed it with `utf-8''`. however If a client would explicitly send a `filename*` and a `filename` than the `akka-http` server (or client) would not be able to access `filename` since it get's overriden by the `filename*` field. So the akka-http server/client would get `utf-8''filename`. Not sure if that is great.

shkhln:
Actually we don't care about request headers, this issue is about response headers. The reason that current behavior is (slightly) problematic is [this piece of code](https://github.com/playframework/playframework/blame/9d6d39911c37bdaa6890e7916b9177e2376974b7/framework/src/play-akka-http-server/src/main/scala/play/core/server/akkahttp/ModelConversion.scala#L218). Here Play tries to reparse already properly formed response headers under assumption that they will not be changed substantially. (I have no idea why it doesn't simply use RawHeader instead.)

> The last time I checked browsers just put the UTF-8 data into the normal filename field.

It looks like `inline/attachment` has different rules vs `form-data` (see the last paragraph at the section 1 of rfc6266). Browsers definitely do not tolerate non-ascii filenames for `inline/attachment` responses without the fancy encoding.

jrudolph:
> So the akka-http server/client would get utf-8''filename. Not sure if that is great.

Right. We should definitely improve there. My question was more about where we want to put it. Should it be preserved in the model or can we do it right in the parser / renderer.

If we keep it in the model in `params`, we already have a method `Multipart.BodyPart.filename` that could be used to get the right filename either from `filename` or `filename*` parameter. 

When creating such a body part, it's really hard to say which the right rendering should be. As browsers are still using the old form with UTF-8 data, there will probably be server implementations that don't know about the "new" (`filename*`) one.

@shkhln 

> It looks like inline/attachment has different rules vs form-data (see the last paragraph at the section 1 of rfc6266). Browsers definitely do not tolerate non-ascii filenames for inline/attachment responses without the fancy encoding.

That's a good data point. According to [RFC 6266](https://tools.ietf.org/html/rfc6266#section-5):

```
   Note: Those user agents that do not support the RFC 5987 encoding
   ignore "filename*" when it occurs after "filename".
```

which would also point to a potential rendering strategy: if non-ASCII characters are encountered, replace them with a placeholder for the `filename` parameter and then add a `filename*` using the RFC 5987 encoding.

schmitch:
Then the question would be how to resove it.
I actually read out from your answer that we keep that PR as is and merge it, since that PR actually only handle's the parsers.

But we need a ticket to actually discuss the renderer part or another PR. Where we replace non ASCII char's with a placeholder.
And we actually need to look into the `Multipart.BodyPart.filename` and ensure that it returns `filename*` with the charset applied or if not present `filename`.

I could actually also fix everything in this PR. it depends on what you actually prefer?
(P.S.: I actually also know that some browser happily accept UTF-8 filename's without `filename`, so I'm not sure if we actually should change the renderer.
Also the current Parser will actually accept `utf-8` Strings with `filename` as well.
According to the RFC it's wrong, according to how browsers and maybe other clients handle it, it seems to be okai... 👎 ) 
But what we should also change is probably: https://github.com/akka/akka-http/blob/master/akka-http-core/src/main/scala/akka/http/scaladsl/model/Multipart.scala#L472

jrudolph:
PLS BUILD

jrudolph:
I made up my mind that we should do it properly and decode those kinds of parameters and store them as `filename`. That would be consistent with what we do elsewhere in the model (which isn't a 1:1 representation of the wire format).

To make the play use case work we could do two things:
 * use `RawHeader` in play (I worked on that, it failed so far because play relies on the validation the parsing does automatically but, in fact, could validate without parsing)
 * we need to support a proper round-trip for these kinds of parameters. That doesn't mean they literally have to be the same (you can always use `RawHeader` if you need literal headers), but they should make sense. I created #1066 to implement that.

Right now, I'd call it a bug that we put the encoded value into the `filename` parameter, that doesn't make any sense, of course. A better strategy for now would be to just skip that parameter until #1066 is implemented.

jrudolph:
I'm closing this for now. The solution we want to achieve is documented in the above comment and in #1240.

Thanks anyway for this contribution, @schmitch.

PR #696, opened 2016-12-29
=pro manually maintain list of previous versions in MiMa.scala

jrudolph:
Otherwise, we would only check subprojects against versions for which a
`mima-filters/<version>.*` file would have already been created.


jrudolph:
PLS BUILD

PR #699, opened 2016-12-30
Fix spray-json unmarshalling

### LINKED ISSUE #691: Difference detected between akka-http_2.11-10.0.1 and akka-http_2.12-10.0.1
I'm seeing a bug in 10.0.1 for 2.11 that doesn't exist for 2.12. That is, unicode characters aren't properly being encoded/decoded. For example emoji.

This was a bug that existed in 10.0.0 that I reported, but was then fixed for 10.0.1... however it's still happening for scala 2.11.

Has anyone seen this or know what's going on? It's blocking a critical release for us and would like to get to the bottom of it sooner than later. Using 2.12 isn't an option for us right now as other dependencies aren't yet compatible with it.

Here's what my build.sbt looks like:
https://gist.github.com/troypayne/7da72ae6bd43363f13ea277d2927f098

Here are the results of sbt > show dependencyClasspath
https://gist.github.com/troypayne/13a3e1a6e8d6450f10355333ecc0d672

jrudolph:
Changes:

   * upgrade to spray-json 1.3.3 (which has working 4-byte UTF-8 character decoding for all Scala versions)
   * use new IndexedBytesParserInput to avoid copied code from spray-json
   * refactor unmarshallers to base all on a single implementation
   * fix tests to actually test more kinds of UTF-8 characters

Fixes #691.

### DIFF
--- a/akka-http-marshallers-scala/akka-http-spray-json/src/main/scala/akka/http/scaladsl/marshallers/sprayjson/SprayJsonByteStringParserInput.scala
+++ b/akka-http-marshallers-scala/akka-http-spray-json/src/main/scala/akka/http/scaladsl/marshallers/sprayjson/SprayJsonByteStringParserInput.scala
@@ -4,64 +4,22 @@
 
 package akka.http.scaladsl.marshallers.sprayjson
 
-import java.nio.{ ByteBuffer, CharBuffer }
-import java.nio.charset.{ Charset, StandardCharsets }
+import java.nio.charset.StandardCharsets
 
 import akka.util.ByteString
-import spray.json.ParserInput.DefaultParserInput
-import scala.annotation.tailrec
+import spray.json.ParserInput.IndexedBytesParserInput
 
 /**
+ * INTERNAL API
+ *
  * ParserInput reading directly off a ByteString. (Based on the ByteArrayBasedParserInput)
  * that avoids a separate decoding step.
+ *
+ * TODO: make private before next major version
  */
-final class SprayJsonByteStringParserInput(bytes: ByteString) extends DefaultParserInput {
-
-  import SprayJsonByteStringParserInput._
-
-  private[this] val byteBuffer = ByteBuffer.allocate(4)
-  private[this] val charBuffer = CharBuffer.allocate(2)
-
-  private[this] val decoder = Charset.forName("UTF-8").newDecoder()
-
-  override def nextChar() = {
-    _cursor += 1
-    if (_cursor < bytes.length) (bytes(_cursor) & 0xFF).toChar else EOI
-  }
-
-  override def nextUtf8Char() = {
-    @tailrec def decode(byte: Byte, remainingBytes: Int): Char = {
-      byteBuffer.put(byte)
-      if (remainingBytes > 0) {
-        _cursor += 1
-        if (_cursor < bytes.length) decode(bytes(_cursor), remainingBytes - 1) else ErrorChar
-      } else {
-        byteBuffer.flip()
-        val coderResult = decoder.decode(byteBuffer, charBuffer, false)
-        charBuffer.flip()
-        val result = if (coderResult.isUnderflow & charBuffer.hasRemaining) charBuffer.get() else ErrorChar
-        byteBuffer.clear()
-        if (!charBuffer.hasRemaining) charBuffer.clear()
-        result
-      }
-    }
-
-    if (charBuffer.position() > 0) {
-      val result = charBuffer.get()
-      charBuffer.clear()
-      result
-    } else {
-      _cursor += 1
-      if (_cursor < bytes.length) {
-        val byte = bytes(_cursor)
-        if (byte >= 0) byte.toChar // 7-Bit ASCII
-        else if ((byte & 0xE0) == 0xC0) decode(byte, 1) // 2-byte UTF-8 sequence
-        else if ((byte & 0xF0) == 0xE0) decode(byte, 2) // 3-byte UTF-8 sequence
-        else if ((byte & 0xF8) == 0xF0) decode(byte, 3) // 4-byte UTF-8 sequence
-        else ErrorChar
-      } else EOI
-    }
-  }
+@deprecated("Will be made private.", "10.0.2")
+final class SprayJsonByteStringParserInput(bytes: ByteString) extends IndexedBytesParserInput {
+  protected def byteAt(offset: Int): Byte = bytes(offset)
 
   override def length: Int = bytes.size
   override def sliceString(start: Int, end: Int): String =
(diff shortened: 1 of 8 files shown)


jrudolph:
Needed to mima around some things... ;)

akka-http-marshallers-scala/akka-http-spray-json/src/main/scala/akka/http/scaladsl/marshallers/sprayjson/SprayJsonByteStringParserInput.scala
```diff
@@ -70,6 +28,7 @@ final class SprayJsonByteStringParserInput(bytes: ByteString) extends DefaultPar
     StandardCharsets.UTF_8.decode(bytes.slice(start, end).asByteBuffer).array()
 }
 
+@deprecated("Not needed any more. Should have been private.", "10.0.2")
 object SprayJsonByteStringParserInput {
   private final val EOI = '\uFFFF'
```
jonas: Would it make sense to remove the private members?

akka-http-marshallers-scala/akka-http-spray-json/src/main/scala/akka/http/scaladsl/marshallers/sprayjson/SprayJsonSupport.scala
```diff
@@ -23,23 +23,25 @@ import scala.language.implicitConversions
 trait SprayJsonSupport {
   implicit def sprayJsonUnmarshallerConverter[T](reader: RootJsonReader[T]): FromEntityUnmarshaller[T] =
     sprayJsonUnmarshaller(reader)
+
   implicit def sprayJsonUnmarshaller[T](implicit reader: RootJsonReader[T]): FromEntityUnmarshaller[T] =
     sprayJsValueUnmarshaller.map(jsonReader[T].read)
-  implicit def sprayJsonByteStringUnmarshaller[T](implicit reader: RootJsonReader[T]): FromByteStringUnmarshaller[T] =
+  implicit def sprayJsValueUnmarshaller: FromEntityUnmarshaller[JsValue] =
+    Unmarshaller.byteStringUnmarshaller
+      .forContentTypes(`application/json`)
```
jonas: If I understand correctly this forces the content type to avoid errors but keeps the charset attribute, which the old code explicitly handled. With the move to use `SprayJsonByteStringParserInput` is this explicit decoding no longer needed?

akka-http-marshallers-scala/akka-http-spray-json/src/main/scala/akka/http/scaladsl/marshallers/sprayjson/SprayJsonByteStringParserInput.scala
```diff
@@ -4,64 +4,22 @@
...
-        val byte = bytes(_cursor)
-        if (byte >= 0) byte.toChar // 7-Bit ASCII
-        else if ((byte & 0xE0) == 0xC0) decode(byte, 1) // 2-byte UTF-8 sequence
-        else if ((byte & 0xF0) == 0xE0) decode(byte, 2) // 3-byte UTF-8 sequence
-        else if ((byte & 0xF8) == 0xF0) decode(byte, 3) // 4-byte UTF-8 sequence
-        else ErrorChar
-      } else EOI
-    }
-  }
+@deprecated("Will be made private.", "10.0.2")
```
ktoso: good 👍

akka-http-marshallers-scala/akka-http-spray-json/src/main/scala/akka/http/scaladsl/marshallers/sprayjson/SprayJsonSupport.scala
ktoso: Content type json includes it being UTF-8 actually so this forces that.

ktoso: (as according to the JSON spec)

akka-http/src/main/scala/akka/http/javadsl/unmarshalling/Unmarshaller.scala
```diff
@@ -104,7 +104,16 @@ abstract class Unmarshaller[-A, B] extends UnmarshallerBase[A, B] {
...
+  /**
+   * Apply this Unmarshaller to the given value.
+   */
+  def unmarshal(value: A, ec: ExecutionContext, mat: Materializer): CompletionStage[B] = asScala.apply(value)(ec, mat).toJava
+
+  /**
+   * Deprecated in favor of [[unmarshal]].
+   */
+  @deprecated("Use unmarshal instead.", "10.0.2")
+  def unmarshall(a: A, ec: ExecutionContext, mat: Materializer): CompletionStage[B] = unmarshal(a, ec, mat)
```
ktoso: lol ok :)

project/Dependencies.scala
```diff
@@ -57,7 +57,7 @@ object Dependencies {
     val alpnApi     = "org.eclipse.jetty.alpn"        % "alpn-api"                     % "1.1.3.v20160715" // ApacheV2
 
     object Docs {
-      val sprayJson   = "io.spray"                   %%  "spray-json"                  % "1.3.2"             % "test"
+      val sprayJson   = Compile.sprayJson % "test"
```
ktoso: align the `"test"` please?

akka-http-marshallers-scala/akka-http-spray-json/src/main/scala/akka/http/scaladsl/marshallers/sprayjson/SprayJsonSupport.scala
jonas: So do we need a test that it will re-encode the original content? I'm mainly concerned that this could introduce a regression in the sense that the original special case must have been put in place for a reason.

jrudolph: According to http://www.iana.org/assignments/media-types/application/json adding charset parameter should have no effect:

```
Note:  No "charset" parameter is defined for this registration.
   Adding one really has no effect on compliant recipients.
```

So, yes, we might introduce a regression for people relying on non-standard behavior. In reality, it might not be a big deal because almost everyone seems to be on UTF-8 nowadays. It is still possible to create a custom unmarshaller if you need to deal with non-standard data. If it turns out that this is a pain point for lots of people we can still revert to more relaxed behavior.

jonas: Thanks for the clarification.

PR #706, opened 2017-01-02
Limit the maximum number of bytes when strictifying an entity (#268)

catalin-ursachi:
Solving #268.


ktoso:
OK TO TEST

ktoso:
Thanks, will review soon :)

catalin-ursachi:
Ah, it's complaining about binary compatibility, because (as far as I can tell) `ToStrict` was `private[akka]`, and hence in principle publicly exposed. What should I do about that?

ktoso:
Here's a guide about binary compat filtering: https://github.com/akka/akka-http/blob/master/CONTRIBUTING.md#binary-compatibility
If it's complaining about things that are `private[akka]` you can safely add it to the filters (explained in the link).

akka-http-core/src/main/resources/reference.conf
```diff
@@ -404,4 +404,6 @@ akka.http {
     # HttpResponse on client side).
     tls-session-info-header = off
   }
+
+  strict-entity-max-bytes = 9223372036854775807
```
ktoso: please add a small comment above explaining what this does

akka-http-tests/src/test/resources/reference.conf
```diff
@@ -6,4 +6,5 @@ akka {
     default-dispatcher.throughput = 1
   }
   stream.materializer.debug.fuzzing-mode = on
+  strict-entity-max-bytes = 9223372036854775807
```
ktoso: no need to repeat this in http-tests or docs, the default one from http-core will be loaded already.

docs/src/test/resources/application.conf
```diff
@@ -1,4 +1,6 @@
 akka {
   # needed for ErrorFilter in RouteDirectivesExampleSpec.scala
   loggers = ["akka.testkit.TestEventListener"]
+
+  strict-entity-max-bytes = 9223372036854775807
```
ktoso: remove from here too

akka-http-core/src/main/resources/reference.conf
catalin-ursachi: Is that otherwise a sensible default value?

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
```diff
@@ -70,10 +70,14 @@ sealed trait HttpEntity extends jm.HttpEntity {
    * Collects all possible parts and returns a potentially future Strict entity for easier processing.
    * The Future is failed with an TimeoutException if the stream isn't completed after the given timeout.
    */
-  def toStrict(timeout: FiniteDuration)(implicit fm: Materializer): Future[HttpEntity.Strict] =
+  def toStrict(timeout: FiniteDuration)(implicit fm: Materializer): Future[HttpEntity.Strict] = {
+    val maxBytes = ActorMaterializerHelper.downcast(fm).system.settings.config.getLong("akka.strict-entity-max-bytes")
```
ktoso: wrong key is used though? It should be `akka.http.strict-entity-max-bytes`

akka-http-core/src/main/resources/reference.conf
ktoso: Please note that this is under ` akka.http {` https://github.com/akka/akka-http/pull/706/files#diff-b4fbbc15badf58a0b02c9fd961f0aa2fR12

akka-http-core/src/main/scala/akka/http/impl/util/package.scala
```diff
@@ -77,31 +74,30 @@ package object util {
...
-  import akka.http.scaladsl.model.{ ContentType, HttpEntity }
-  import akka.stream.impl.fusing.GraphStages.SimpleLinearGraphStage
-  import akka.stream.{ Attributes, Outlet, Inlet, FlowShape }
+  import akka.stream.{ Attributes, FlowShape, Inlet, Outlet }
+
   import scala.concurrent.duration.FiniteDuration
 
-  private[http] class ToStrict(timeout: FiniteDuration, contentType: ContentType)
-    extends GraphStage[FlowShape[ByteString, HttpEntity.Strict]] {
+  private[http] class AggregateBytes(timeout: FiniteDuration, maxBytes: Long)
```
ktoso: Could be a good direction anyway WDYT @jrudolph ?

akka-http-core/src/main/scala/akka/http/impl/util/package.scala
```diff
@@ -110,22 +106,25 @@ package util {
       setHandler(byteStringIn, new InHandler {
         override def onPush(): Unit = {
           bytes ++= grab(byteStringIn)
-          pull(byteStringIn)
+          if (bytes.length > maxBytes)
+            failStage(new Exception(s"AggregateBytes received more than the configured maximum $maxBytes bytes of data"))
```
ktoso: I'd propose introducing a special exception for this, `AggregateBytesLimitExceededException`?

akka-http-core/src/main/resources/reference.conf
ktoso: Yeah, I think so since we also want to be compatible with previous 10.x releases. 
So it should not start blowing up once people upgrade to 10.0.2.
We could decrease it in 10.1 perhaps.

ktoso:
Looking good, needs to be explained in reference documentation too so people can discover this feature. Shaping up really nice, thanks! 👍

jrudolph:
I will have to see how this will mix with my new repeatable toStrict approach. I think adding a configuration setting might be the right way to do it. On the other hand we already have a `max-content-length` setting which also applies.

catalin-ursachi:
Incidentally, should .toStrict fail outright if the Content-Length is specified, and greater than the max value? I guess this would be automatically satisfied if we used the max-content-length setting instead of a distinct one.

catalin-ursachi:
Also, `ActorMaterializerHelper.downcast` would fail if we were using a `SubFusingActorMaterializerImpl` instead of an `ActorMaterializer`, even though `SubFusingActorMaterializerImpl.delegate` could provide an `ActorMaterializer`. Is that nevertheless fine?

akka-http-core/src/main/resources/reference.conf
jonas: I suggest to express the value using a unit format, for example `8192 pebibytes` as documented here https://github.com/typesafehub/config/blob/master/HOCON.md#size-in-bytes-format

catalin-ursachi:
I do wonder about this; any Entity received over HTTP will already have had the 'max-content-length' limit applied; is it useful to apply a different (or identical) limit when strictifying the entity? If not, why not leave it as it is?

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
2beaucoup: Use `getBytes` instead with a nicer looking value for `strict-entity-max-bytes`?

jonas:
What is the status of this?

@catalin-ursachi Would it be possible for you to rebase the changes to fix the conflicts?

catalin-ursachi:
I've made changes as per the comments; sorry for the delay! 🙂 

That said, I'm still not convinced about the usefulness of this feature, seeing as it seems to be duplicating the more general `max-content-length` setting anyway?

catalin-ursachi:
To summarize:

* Do we want a `akka.http.strict-entity-max-bytes` limit to be imposed when strictifying entitites, thereby overriding `akka.http.parsing.max-content-length`, or any specific size limit set on the entity using `withSizeLimit`?
* If yes, then would it not be more consistent to simply call `withSizeLimit(strictEntityMaxBytes)` just before strictifying the entity, rather than implementing a separate check withing the strictification graph stage?
* If not, then does this serve any purpose?

catalin-ursachi:
@jonas, @ktoso - Any thoughts on whether this should be closed/merged/changed any further?

PR #708, opened 2017-01-03
=htp #459 prevent "Connection closed by peer" errors during connection closure

### LINKED ISSUE #459: TCP RST was sent with large response body
Akka HTTP version: 3.0.0-RC1

How to reproduce:

 - Use small MTU for network interface (for example 1500). You can set MTU for loopback interface with `sudo ifconfig lo0 mtu 1500`
 - Response with large body content (should be larger than 128KiB) using `complete` and `ToResponseMarshallable`
 - Make sure client send header `Connection: close`.

Expected result:

Client should receive normally

Actual result:

Client got `Recv failure: Connection reset by peer`

This causes problems with some proxy like nginx or kong because they send `Connection: close` to upstream by default.

jrudolph:
A race between completion and cancellation towards the TCP stream endpoints
might lead to the connection being canceled (= RST frames sent) when an HTTP
connection is regularly closed (e.g. when the client sets `Connection: close`
in the request).

We now prevent cancellation from reaching the TCP stream at all and rely
on closing the connection from the write side.

Fixes #459.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
@@ -233,6 +233,29 @@ private[http] object StreamUtils {
       def open(): Unit = promise.success(())
     }
   }
+
+  /**
+   * INTERNAL API
+   *
+   * Returns a flow that is almost identity but doesn't propagate cancellation from downstream to upstream.
+   *
+   * Note: This might create a resource leak if the upstream never completes. External measures
+   * need to be taken to ensure that the upstream will always complete.
+   */
+  def absorbCancellation[T]: Flow[T, T, NotUsed] = Flow.fromGraph(new AbsorbCancellationStage)
+  final class AbsorbCancellationStage[T] extends SimpleLinearGraphStage[T] {
+    def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new GraphStageLogic(shape) with InHandler with OutHandler with StageLogging {
+      setHandlers(in, out, this)
+
+      def onPush(): Unit = push(out, grab(in)) // using `passAlong` was considered but it seems to need some boilerplate to make it work
+      def onPull(): Unit = pull(in)
+
+      override def onDownstreamFinish(): Unit = {
+        // don't pass cancellation to upstream, and ignore eventually outstanding future element
+        setHandler(in, new InHandler { def onPush(): Unit = log.debug("Ignoring unexpected data received after cancellation was ignored.") })
+      }
+    }
+  }
 }
 
 /**
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
@@ -17,6 +17,7 @@ import akka.http.impl.engine.client._
 import akka.http.impl.engine.server._
 import akka.http.impl.engine.ws.WebSocketClientBlueprint
 import akka.http.impl.settings.{ ConnectionPoolSetup, HostConnectionPoolSetup }
+import akka.http.impl.util.StreamUtils
 import akka.http.scaladsl.model._
 import akka.http.scaladsl.model.headers.Host
 import akka.http.scaladsl.model.ws.{ Message, WebSocketRequest, WebSocketUpgradeResponse }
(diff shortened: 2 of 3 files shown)


akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
```diff
@@ -233,6 +233,24 @@ private[http] object StreamUtils {
...
   }
+
+  /**
+   * Returns a flow that is almost identity but doesn't pass cancellation from downstream to upstream.
+   *
+   * Note: This might create a resource leak if the upstream never completes. External measures
+   * need to be taken to ensure that the upstream will always complete.
+   */
+  def ignoreCancellation[T]: Flow[T, T, NotUsed] = Flow.fromGraph(new IgnoreCancellation)
+  class IgnoreCancellation[T] extends SimpleLinearGraphStage[T] {
```
viktorklang: @jrudolph  Apologies for the seemingly dumb question, but doesn't *ignoring cancellation* violate Reactive Streams?
(I'm more than likely missing something here?) Also public class and not final?

jrudolph: I guess the comment is not quite accurate. Cancellation is not ignored in the sense that the stage will sent additional elements after it received cancellation (that is already prevented by the GraphStage). It is just not propagated to the upstream side of this flow in this case. (The default behavior when receiving cancellation will call `completeStage` which is correct for most flow-like stages but for many others it isn't.)

Also, it's only a utility class. As explained (also in the other comment in Http.scala) it is only used in circumstances where the overall setup ensures that a flow (or complete graph setup) is closed from the other side.

And, yes, `final` would make sense, it already lives in a `private[http]` object.

ktoso: Yeah the GraphStage upholds all RS rules, so we won't be able to push a closed port - it'll throw "can not push closed port". We could avoid that risk, and make it explicit that this won't happen by doing

```
  // don't pass cancellation to upstream, and
  setHandler(in, new OutHandler { def onPush(): Unit = log.debug("received unexpected push after cancellation swallowed") } 
}

```

ktoso: Also, perhaps changing name slightly would be good? Like `BarrierCancellation` / `BarrierCancellation`?
I've had instances where I needed to barrier off completion too, so that would make it naming wise simple to add such one too once needed again

viktorklang: @jrudolph @ktoso As you say, the comment/explanation/name is slightly misleading. It's really about something like "AbsorbCancellationStage"?

viktorklang: And, great fix btw!

ktoso: Might as well revive the old `absorb` wording for these, why not 👍

jrudolph: Thanks, @ktoso and @viktorklang. Good comments. Fixed!

akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
```diff
@@ -233,6 +233,29 @@ private[http] object StreamUtils {
...
+  final class AbsorbCancellationStage[T] extends SimpleLinearGraphStage[T] {
+    def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new GraphStageLogic(shape) with InHandler with OutHandler with StageLogging {
+      setHandlers(in, out, this)
+
+      def onPush(): Unit = push(out, grab(in)) // using `passAlong` was considered but it seems to need some boilerplate to make it work
+      def onPull(): Unit = pull(in)
+
+      override def onDownstreamFinish(): Unit = {
+        // don't pass cancellation to upstream, and ignore eventually outstanding future element
+        setHandler(in, new InHandler { def onPush(): Unit = log.debug("Ignoring unexpected data received after cancellation was ignored.") })
```
ktoso: yup cool

ktoso:
LGTM, great - and great new names :)

PR #711, opened 2017-01-04
WIP: =ht2 revamp header decompression

jrudolph:
* new phase in which headers are decompressed and compressed with HPACK
   (reusable for server and (later) client)
 * in place of the old phase there's only a simple conversion from
   key-value pairs to HttpRequest

Unfortunately, seems I started from the wrong branch, so I will have some conflict resolution do to before this is ready.

Afterwards, the same kind of change needs to be done for rendering.


akka-http2-support/src/main/scala/akka/http/impl/engine/http2/FrameEvent.scala
```diff
@@ -61,4 +61,10 @@ final case class UnknownFrameEvent(
...
   flags:    ByteFlag,
   streamId: Int,
-  payload:  ByteString) extends StreamFrameEvent
\ No newline at end of file
+  payload:  ByteString) extends StreamFrameEvent
+
+final case class ParsedHeadersFrame(
+  streamId:      Int,
+  endStream:     Boolean,
+  keyValuePairs: Seq[(String, String)]
```
ktoso: hm I wonder about this type, but reading on.

jrudolph:
Reopened against correct branch: https://github.com/akka/akka-http/pull/718/files

PR #715, opened 2017-01-05
Basic auth fix according to rfc7617 - 'charset' auth-param

guntiso:


ktoso:
Hm indeed we can add this. It seems it's purely advisory though. Have you seen problems arise due dot lack of this?
Do we need 2 versions of this method perhaps?

ktoso:
Related section in RFC https://tools.ietf.org/html/rfc7617#section-2.1

ktoso:
Please open tickets and then PRs so the PR can resolve the ticket as outlined in https://github.com/akka/akka-http/blob/master/CONTRIBUTING.md#general-workflow

I'll open a ticket for this one.

ktoso:
Aims to resolve https://github.com/akka/akka-http/issues/716

guntiso:
If charset is not defined, browsers will stick to their (conflicting) historical implementations. I think only one version of this method should exist, if possible. A good summary of problems with non-ASCII in HTTP Basic is here - chromium issue 25790, comment 10 - https://bugs.chromium.org/p/chromium/issues/detail?id=25790#c10

ktoso:
That's very good context, thanks - please provide such when opening PRs if you have right away :-)
Right it seems to be one of those painful fuzzy HTTP issues... Defining using UTF-8 sounds good actually given what I've read in the issue. 

LGTM from my side.
Failure was likely unrelated from what I see.

ktoso:
PLS BUILD

ktoso:
Merged, thanks for the PR @guntiso  👍

PR #720, opened 2017-01-05
+ht2 #251 Proposal of h2spec "tck" integration

### LINKED ISSUE #251: Setup HTTP/2 conformancy test
Part of #244 

See tools recommended here: https://blog.cloudflare.com/tools-for-debugging-testing-and-using-http-2/

ktoso:
Sadly had to include the jar file as file... since we need to fork the tests off with the agent already loaded it can't be lazily loaded by the test itself (since the entire test jvm won't be able to start).

We're running the h2spec in sections, so we'd be able to not use one in case it's not as good.
Example output: 

1__java
1__java

This won't fail validation since not passing TCK tests currently are marked as Pending, but we should work on making them pass and then switch to error mode.

Resolves #251   

cc @johanandren @schmitch


ktoso:
Website: https://github.com/summerwind/h2spec

schmitch:
can't you just pass the -j flag and use the junit xml for the test check:
```
$ h2spec --help
Usage: h2spec [OPTIONS]

Options:
....
  -j:        Creates report also in JUnit format into specified file.
....
```

I mean it generates a xml so you could create one file per section/test i.e.
`-k -t -p $port -s 3.5 -j 3_5_http2_connection_preface.xml`
?

ktoso:
It's not about "just use the flag". We'll use the flag, yes for reporting
on jenkins. However for development I want to be able to run a specific set
of tests directly using sbt.

schmitch:
Ah so this is more for "personal use"? than for the test server?

ktoso:
Yeah, though you're right - for the build server we want the junit report :)

ktoso:
Now with super cool @jonas agent and spec download improvements.
And reporting as junit on jenkins.

ktoso:
So everything works nice now, but we (as expected) failed the build.
https://jenkins.akka.io:8498/job/pr-validator-akka-http/550/testReport/

I'll add post processing so we can have this merged and mark them as pending.

ktoso:
Nicely reporting now: https://jenkins.akka.io:8498/view/PR%20Validators/job/pr-validator-akka-http/557/testReport/h2spec-4/3%20Header%20Compression%20and%20Decompression/h2spec_Sends_invalid_header_block_fragment/

ktoso:
I'm going to merge and move forward

akka-http2-support/build.sbt
```diff
@@ -1,8 +1,36 @@
 import akka._
+import java.nio.file.Files
+import java.nio.file.attribute.{ PosixFileAttributeView, PosixFilePermission }
+import Dependencies.{ h2specName, h2specExe }
 
 Dependencies.http2
+Dependencies.http2Support
```
jrudolph: We should remove one of them.

build.sbt
```diff
@@ -79,6 +79,8 @@ lazy val http = project("akka-http")
   .dependsOn(httpCore)
 
 lazy val http2Support = project("akka-http2-support")
+  .enablePlugins(JavaAgent)
+  .settings(javaAgents += Dependencies.Compile.Test.alpnAgent)
```
jrudolph: Very nice!

PR #735, opened 2017-01-09
+ht2 emit GO_AWAY on failed header decompression

ktoso:
This fails all (most) other tests  other than the go_away thing, the reason is that I now don't send the settings eagerly (removed from preStart). Once back home will work on fixing all those. The "pre-buffer idea" seemed interesting but is likely wrong, as a pull could come early anyway.


akka-http2-support/src/main/scala/akka/http/impl/engine/http2/BufferedOutletSupport.scala
```diff
@@ -75,18 +75,27 @@ trait BufferedOutletSupport { logic: GraphStageLogic ⇒
       elem.trigger()
     }
 
+    /** Buffer an element to be used as initial element to be pushed once a second element arrives. */
+    final def buffer(element: T): Unit =
```
ktoso: these are just random ideas so far... Instead I'll likely manually push the Settings in the right places.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
```diff
@@ -70,6 +63,12 @@ object Http2Blueprint {
    * is to add a special (virtual) header containing the streamId (or any other kind of token) is added to the HttRequest
    * that must be reproduced in an HttpResponse. This can be done automatically for the bindAndHandleAsync API but for
    * bindAndHandle the user needs to take of this manually.
+   * Translation between substream frames and Http messages (both directions)
+   *
+   * To make use of parallelism requests and responses need to be associated (other than by ordering), suggestion
+   * is to add a special (virtual) header containing the streamId (or any other kind of token) is added to the HttRequest
+   * that must be reproduced in an HttpResponse. This can be done automatically for the bindAndHandleAsync API but for
+   * bindAndHandle the user needs to take of this manually.
```
ktoso: bad rebase, meh

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -97,7 +97,7 @@ class Http2ServerDemux extends GraphStage[BidiShape[Http2SubStream, FrameEvent,
         pull(frameIn)
         pull(substreamIn)
 
-        bufferedFrameOut.push(SettingsFrame(Nil)) // server side connection preface
+        bufferedFrameOut.buffer(SettingsFrame(Nil)) // server side connection preface // FIXME don't do this, since we want to GOAWAY right away
```
ktoso: So removing that push generally speaking is needed so the first frame MAY be GO_AWAY if we want to.
Technically there's no obligation for it to be the first one I guess (the go-away), though the h2spec wants this so in many tests, so would be good to be able to control this fully anyway. I'll read more about Settings semantics too - I could continue on those after this perhaps.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/FramesInFramesOutSpec.scala
```diff
@@ -0,0 +1,78 @@
...
+      netIn.sendNext(HeadersFrame(1, endStream = true, endHeaders = true, headerBlock))
+
+      netOut.request(100)
+      val frame = netOut.expectNext
+      val goAway = frame.asInstanceOf[GoAwayFrame]
+      // not asserting the lastStreamId, it's allowed to be racy (and we actually only later store it, so it would be 0 here)
+      goAway.errorCode should ===(ErrorCode.COMPRESSION_ERROR)
+
+      netIn.sendComplete()
+    }
```
ktoso: this reproduces the 4.3 tests in a manual way.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
jrudolph: This is the server-side handshake, so we might change it but if it is just to make h2spec work, I'm not sure I'm for it.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/hpack/HeaderDecompression.scala
```diff
@@ -42,13 +44,20 @@ private[http2] object HeaderDecompression extends GraphStage[FlowShape[FrameEven
...
-      push(eventsOut, ParsedHeadersFrame(streamId, endStream, headers.result()))
+      try {
+        decoder.decode(ByteStringInputStream(payload), Receiver)
+        decoder.endHeaderBlock() // TODO: do we have to check the result here?
+
+        push(eventsOut, ParsedHeadersFrame(streamId, endStream, headers.result()))
+      } catch {
+        case ex: IOException ⇒
+          // this is signalled by the decoder when it failed, we want to react to this by rendering a GO_AWAY frame
+          push(eventsOut, GoAwayFrame(streamId, ErrorCode.COMPRESSION_ERROR))
```
jrudolph: THE GOAWAY frame needs to be sent in the other direction.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/FramesInFramesOutSpec.scala
```diff
@@ -0,0 +1,78 @@
...
+import akka.stream.ActorMaterializer
+import akka.stream.scaladsl.{ BidiFlow, Flow, Keep }
+import akka.stream.testkit.scaladsl.{ TestSink, TestSource }
+import akka.stream.testkit.{ TestPublisher, TestSubscriber }
+import akka.testkit.AkkaSpec
+import org.scalatest.concurrent.Eventually
+
+import scala.concurrent.duration._
+
+class FramesInFramesOutSpec extends AkkaSpec with Eventually {
```
jrudolph: Would it make sense to convert Http2ServerSpec, instead, to not rely on network bytes but parsed frames instead? Otherwise, how do we determine which tests will live where? The good thing about Http2ServerSpec is that it really checks the whole stack (without TLS) and already has most of the utilities to render and parse frames to ByteStrings.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
ktoso: >  if it is just to make h2spec work, I'm not sure I'm for it.

yeah absolutely agreed - will check the spec in more detail. but if its just that testkit demanding to much stuff I'm happy to ignore that test.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/hpack/HeaderDecompression.scala
ktoso: I was toying with 2 ideas here, either we send this way and in the demux handle and "bounce it back" like in this PR. Or we make the compression stuff a full bidistage in which case it can by itself send the frame then. The second one was closer to my original branch - happy to go back to that style (didnt get there yet since rebase hell with the moving compression around PR ;-)).

So, 1 bidistage and push back directly you say? Can do.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/FramesInFramesOutSpec.scala
ktoso: Good point, can work on it to accept frame objects.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/hpack/HeaderDecompression.scala
jrudolph: In any case, we somehow need to distinguish a client-side GoAwayFrame from this one, right? In cases of errors, I'd rather use exceptions if we want to do the loop. Maybe let's try without the BidiStage as long as we can get away with it.

ktoso: Hm, can model this as `fail(port`, let's see then.

akka-http-core/src/main/scala/akka/http/impl/util/LogByteStringTools.scala
```diff
@@ -17,16 +17,25 @@ private[akka] object LogByteStringTools {
...
     BidiFlow.fromFlows(
       logByteString(s"$name DOWN", maxBytes),
-      logByteString(s"$name UP", maxBytes)
+      logByteString(s"$name UP  ", maxBytes)
+    )
+
+  def logToStringBidi[A](name: String, maxBytes: Int = MaxBytesPrinted): BidiFlow[A, A, A, A, NotUsed] =
+    BidiFlow.fromFlows(
+      logToString(s"$name DOWN", maxBytes),
+      logToString(s"$name UP  ", maxBytes)
```
ktoso: useful when debugging what event is flying where, would keep those utils if you don't mind

akka-http-core/src/main/scala/akka/http/impl/util/LogByteStringTools.scala
```diff
@@ -17,16 +17,25 @@ private[akka] object LogByteStringTools {
...
     Flow[ByteString].log(name, printByteString(_, maxBytes))
 
+  def logToString[A](name: String, maxBytes: Int = MaxBytesPrinted): Flow[A, A, NotUsed] =
+    Flow[A].log(name, _.toString().take(maxBytes))
+
   def logTLSBidi(name: String, maxBytes: Int = MaxBytesPrinted): BidiFlow[SslTlsOutbound, SslTlsOutbound, SslTlsInbound, SslTlsInbound, NotUsed] =
     BidiFlow.fromFlows(
       logTlsOutbound(s"$name DOWN", maxBytes),
-      logTlsInbound(s"$name UP", maxBytes))
+      logTlsInbound(s"$name UP  ", maxBytes))
```
ktoso: aligns them for easier reading

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -106,18 +105,25 @@ class Http2ServerDemux extends GraphStage[BidiShape[Http2SubStream, FrameEvent,
...
 
-      def lastStreamId: Int =
-        incomingStreams.lastOption.map(_._1).getOrElse(0) // FIXME: this is a Map which has no order, so lastStreamId will be an arbitrary value
+      /**
+       * The "last peer-initiated stream that was or might be processed on the sending endpoint in this connection"
+       * @see http://httpwg.org/specs/rfc7540.html#rfc.section.6.8
+       */
+      def lastStreamId: Int = {
+        incomingStreams.keys.toList.sortBy(-_).headOption.getOrElse(0) // FIXME should be optimised
+      }
```
ktoso: trying a bit more than before, in step-by-step tests like we have the value should be predictable.

ktoso:
Please review :)

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
schmitch: well basically I think we end up saving the last streamId as var at some point and even when not the incomingStreams list will probably never hold many keys (depends on stream settings and max stream settings)

ktoso: Yeah, this will  likely become a number in a var, didnt want to refactor that for now

jrudolph: Agree with both.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/hpack/HeaderDecompression.scala
```diff
@@ -42,13 +44,20 @@ private[http2] object HeaderDecompression extends GraphStage[FlowShape[FrameEven
...
+      try {
+        decoder.decode(ByteStringInputStream(payload), Receiver)
+        decoder.endHeaderBlock() // TODO: do we have to check the result here?
+
+        push(eventsOut, ParsedHeadersFrame(streamId, endStream, headers.result()))
+      } catch {
+        case ex: IOException ⇒
+          // this is signalled by the decoder when it failed, we want to react to this by rendering a GOAWAY frame
+          fail(eventsOut, new Http2Compliance.HeaderDecompressionFailed("Decompression failed."))
+        case ex: Throwable ⇒ throw ex // re-throw (unxepected exception)
```
jrudolph: Not needed ;)

ktoso: True, silly :)

akka-http-core/src/main/scala/akka/http/impl/util/LogByteStringTools.scala
```diff
@@ -17,16 +17,25 @@ private[akka] object LogByteStringTools {
   def logByteStringBidi(name: String, maxBytes: Int = MaxBytesPrinted): BidiFlow[ByteString, ByteString, ByteString, ByteString, NotUsed] =
     BidiFlow.fromFlows(
       logByteString(s"$name DOWN", maxBytes),
-      logByteString(s"$name UP", maxBytes)
+      logByteString(s"$name UP  ", maxBytes)
+    )
+
+  def logToStringBidi[A](name: String, maxBytes: Int = MaxBytesPrinted): BidiFlow[A, A, A, A, NotUsed] =
```
jrudolph: Yep, why not.

jrudolph:
LGTM.

PR #739, opened 2017-01-09
+doc update paradox and enable source url feature

ktoso:


ktoso:
Thanks a lot @dwijnand @eed3si9n!

Preview:

java_documentation_ _akka_http

ktoso:
FYI @jonas, you may like this :-)

jrudolph:
Wow nice, I though about building it myself a while ago. This way is nicer ;)

jonas:
I do like this ... and props to @gsechaud for adding this to Paradox.

docs/src/main/paradox/_template/page.st
```diff
@@ -72,7 +72,11 @@
             <div class="page-content row">
               <div class="small-12 large-9 column" id="docs">
                 $page.content$
-                <!-- FIXME copy the template again once paradox 0.2.8 is out, it has source inclusion here.-->
+                
+                <div>
+                  $source()$
```
ktoso: We can customise the text too but going with the simple thing now.

ktoso:
Merging, full steam ahead ;-)

PR #747, opened 2017-01-10
Docs: Fix indentation of snippets

jrudolph:


docs/src/main/paradox/_template/page.st
```diff
@@ -71,10 +71,10 @@
 
             <div class="page-content row">
               <div class="small-12 large-9 column" id="docs">
-                $page.content$
-                
+$page.content$
```
ktoso: add a comment that this is on purpose?

jrudolph: Good point, added a comment.

jrudolph: FYI: `$! comment !$` can be used to add comments on the ST4 level which won't be added to the output.

docs/src/main/paradox/_template/page.st
```diff
@@ -71,10 +71,15 @@
...
             <div class="page-content row">
               <div class="small-12 large-9 column" id="docs">
-                $page.content$
-                
+
+$!
+   Must not be indented, will otherwise ST4 will otherwise indent <pre> blocks in the
+   content as well, leading to extra indentation in the generated output.
+!$
+$page.content$
```
ktoso: cool, thanks!

Issue #748, opened 2017-01-10
WS: Optimize first empty frame


jrudolph:
For implementation reasons, the WS implementation sends out a streamed WS message as this succession of frames:

```
TEXT         fin = false empty
CONTINUATION fin = false data chunk 1 
CONTINUATION fin = false data chunk 2
...
CONTINUATION fin = true  empty
```

https://github.com/akka/akka-http/blob/5932237a86a432d623fafb1e84eeeff56d7485fe/akka-http-core/src/main/scala/akka/http/impl/engine/ws/MessageToFrameRenderer.scala#L25-L25:

```scala
      Source.single(FrameEvent.empty(opcode, fin = false)) ++
        data.map(FrameEvent.fullFrame(Opcode.Continuation, None, _, fin = false)) ++
        Source.single(FrameEvent.emptyLastContinuationFrame)
```

The first frame could be optimized away.

This was observed in #730 by @elseorand.

jrudolph:
Fixed by #749.

PR #749, opened 2017-01-10
=htc #748 optimize frame creation for streamed WS messages

jrudolph:
Previously, a single empty frame was created to start a WS message. Now,
a streamed message will only be begun when the first data comes in.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
@@ -256,6 +256,15 @@ private[http] object StreamUtils {
       }
     }
   }
+
+  /**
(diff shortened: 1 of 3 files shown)


jrudolph:
Seems this fails a few WS tests. Will have a look.

akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
```diff
@@ -256,6 +256,15 @@ private[http] object StreamUtils {
       }
     }
   }
+
+  /**
+   * Similar idea than [[FlowOps.statefulMapConcat]] but for a simple map.
+   */
+  def statefulMap[T, U](functionConstructor: () ⇒ T ⇒ U): Flow[T, U, NotUsed] =
```
ktoso: nice and simple one, cool

PR #751, opened 2017-01-11
improve h2spec compliance - fixes 5.3 (#245)

gosubpl:
Introduces self dependency checks for `HEADER` and `PRORITY` frames, making `5.3` pass. Also adds missing size check for `PRIORITY` frame

### DIFF
--- a/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/FrameParser.scala
+++ b/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/FrameParser.scala
@@ -69,6 +69,11 @@ class FrameParser(shouldReadPreface: Boolean) extends ByteStringParser[FrameEven
           if (priority) payload.readByte() & 0xff
           else 0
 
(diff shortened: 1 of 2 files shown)


gosubpl:
Refs #245

gosubpl:
PLS BUILD

gosubpl:
`RouteDirectivesSpec` has failed. BTW, I really like the idea of small PRs :)

jrudolph:
LGTM, thanks, @gosubpl.

ktoso:
LGTM, I was thinking that replicating those tests in our own unit tests in Http2ServerSpec would be a good idea since we have more control there (possibly could remove h2spec if we cover everything and more ourselves)

gosubpl:
@ktoso agree, good idea. I will include this in the next PR. Especially since we are getting some false *positives* in h2spec.

ktoso:
Yeah, I'm not too happy about it - let's keep porting tests into our own suite

PR #762, opened 2017-01-12
=ht2 parse and render priority information of HEADERS frames

jrudolph:
Small one propagating the priority info from frames.

Initially, I thought we might just render this as an additional PriorityFrame. I think, however, that it might make sense to model it accurately, so that priority information is complete from the start.


ktoso:
Nice, looks fine to include like so 👍

Issue #757, opened 2017-01-12
HTTP2: Respect SETTINGS_MAX_FRAME_SIZE


jrudolph:
SETTINGS_MAX_FRAME_SIZE (0x5):
Indicates the size of the largest frame payload that the sender is willing to receive, in octets.

The initial value is 214 (16,384) octets. The value advertised by an endpoint MUST be between this initial value and the maximum allowed frame size (224-1 or 16,777,215 octets), inclusive. Values outside this range MUST be treated as a connection error (Section 5.4.1) of type PROTOCOL_ERROR.

ktoso:
That's the setting I started working with yesterday, so marking as in progress.

ktoso:
Part of https://github.com/akka/akka-http/issues/757

ktoso:
We're respecting the setting if sent from the client, the server is not sending it / limiting incoming frames yet,

PR #764, opened 2017-01-13
+h2 handle MAX_FRAME_SIZE, merge framing stages into bidi

ktoso:
- Solves part of https://github.com/akka/akka-http/issues/757 - the DATA part of it, we buffer data until the entire thing was pushed out (chunked up into at-most-MAX_FRAME_SIZE).
- GOAWAYs wrong set frame sizes


ktoso:
Now also handling https://github.com/akka/akka-http/issues/776 SETTINGS_HEADER_TABLE_SIZE in hpack
Over there it would also be nice to be able to swallow it early but that would need a bidistage.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/FrameEvent.scala
```diff
@@ -71,3 +73,10 @@ final case class ParsedHeadersFrame(
   keyValuePairs: Seq[(String, String)],
   priorityInfo:  Option[PriorityFrame]
 ) extends FrameEvent
+
+// ------ internal frames, used for communicating settings between stages of our pipeline ----- 
+/**
+ * Special synthetic FrameEvent used to communicate enoder setting changes back to encoder.
+ * MUST NOT be rendered to network.
+ */
+private[http] final case class SyntheticHpackEncoderSettingFrame(setting: Setting) extends SyntheticFrameEvent
```
ktoso: Either we do this dance with sending back "down" the setting in a special one like that, or we need to make the HPack stuff a bidistage the same way like the framing was done now.

ktoso:
Best to review the 2 commits at the same time, not independently IMO. Notice that in hpack where I did not make it a bidistage we need to do the weird "bounce the setting back down for the encoder to apply it" which is basically what we've avoided in framing by making it a bidistage. 

I don't have a good test for setting the table size yet hm...

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Protocol.scala
```diff
@@ -191,7 +192,9 @@ object Http2Protocol {
     def byId(id: Int): SettingIdentifier = All(id - 1)
   }
 
-  sealed abstract class ErrorCode(val id: Int) extends Product
+  sealed abstract class ErrorCode(val id: Int) extends Product {
+    def name = Logging.simpleName(this).replaceAll("$", "")
```
jrudolph: Why not `productPrefix`?

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -114,11 +114,11 @@ class Http2ServerDemux extends GraphStage[BidiShape[Http2SubStream, FrameEvent,
         incomingStreams.keys.toList.sortBy(-_).headOption.getOrElse(0) // FIXME should be optimised
       }
 
-      def pushGOAWAY(errorCode: ErrorCode = Http2Protocol.ErrorCode.PROTOCOL_ERROR): Unit = {
+      def pushGOAWAY(errorCode: ErrorCode = Http2Protocol.ErrorCode.PROTOCOL_ERROR, debug: String = ""): Unit = {
```
jrudolph: Good one, we should annotate all occurrences with this debug info, not only for the peer but also to log it on our side.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -179,32 +179,51 @@ class Http2ServerDemux extends GraphStage[BidiShape[Http2SubStream, FrameEvent,
               incomingStreams(streamId).inlet.foreach(_.cancel())
 
             case WindowUpdateFrame(streamId, increment) ⇒
-              incomingStreams(streamId).outboundWindowLeft += increment
+              incomingStreams(streamId).outboundWindowLeft = incomingStreams(streamId).outboundWindowLeft + increment
```
jrudolph: Does that change anything? (Other than doing the map lookup twice...)

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/hpack/HeaderCompression.scala
```diff
@@ -5,39 +5,45 @@
...
 
     object Idle extends State {
-      val handleEvent: PartialFunction[FrameEvent, Unit] = {
+      override val handleEvent: PartialFunction[FrameEvent, Unit] = {
         case fr @ ParsedHeadersFrame(streamId, endStream, kvs, prioInfo) ⇒
           os.reset()
           kvs.foreach {
             case (key, value) ⇒
-              encoder.encodeHeader(os, key.getBytes(HeaderDecompression.UTF8), value.getBytes(HeaderDecompression.UTF8), false)
+              encoder.encodeHeader(os, key.getBytes(StandardCharsets.UTF_8), value.getBytes(StandardCharsets.UTF_8), false)
```
jrudolph: Nice, thanks. Forgot about these.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -500,23 +502,113 @@ class Http2ServerSpec extends AkkaSpec with WithInPendingUntilFixed with Eventua
         error should ===(ErrorCode.FRAME_SIZE_ERROR)
       }
 
-      "received SETTINGS_MAX_FRAME_SIZE" in pending
+      "received SETTINGS_MAX_FRAME_SIZE should cause outgoing DATA to be chunked up into at-most-that-size parts " in new TestSetup with RequestResponseProbes {
```
jrudolph: nice test

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -500,23 +502,113 @@ class Http2ServerSpec extends AkkaSpec with WithInPendingUntilFixed with Eventua
...
+        def setMaxFrameSize(size: Int) = SettingsFrame(Setting(SettingIdentifier.SETTINGS_MAX_FRAME_SIZE, size) :: Nil)
```
jrudolph: Maybe add `sendSettingsFrame(settingIdentifier, value)` to the TestSetup, instead. Will be needed more often than that.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/FrameEvent.scala
jrudolph: What if we change the `SettingsAckFrame` to carry all ack'd settings, instead? (We don't change the actual rendering, of course.)

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
ktoso: good point will revert back, I was debugging things here

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Protocol.scala
ktoso: Good one 👍

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/FrameEvent.scala
ktoso: Yeah could be nicer indeed 👍

jrudolph:
Regarding bouncing settings vs. handling it right where it is applied: I think there might be places where it could be worthwhile if all stages see a settings change even if not directly affected (e.g. demux would probably want to try to buffer full frames for outgoing data but not more).

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
ktoso: Yup, sure 👍

ktoso:
// Talked it over a bit on chat.
Yeah let's go back to the bouncing back style. It will work out simpler than initially expected.
I'll change the Ack to carry around the setting, such that we can apply it on the "way out".
Seems we'll still need the synthetic version though, but I'll give it some more thought.

The spec does not really require "immediate" applying however it does in-order, which we can keep using the bouncing. Passing around the acks could impact perf a bit but it'll in no way be a big part of perf so I was over worrying perhaps.

Moving to bouncing style then.

Also added some more fun tests, adding now that we must now overflow the demand window etc.

jrudolph:
I'm closing this one as I think should break the parts of it apart to resolve the conflicts and actually make it mergeable.

jrudolph:
I extracted #786 and #787 from this PR. Unfortunately, these only contained the small and / or cosmetic changes.

All the remaining changes are lumped together in this commit: https://github.com/jrudolph/akka-http/commit/a3214dfb821f7f6498b4d7b6bf0a1a0323f6cb22, let's start again from that state.

ktoso:
Ok, thanks for breaking it up like that!

PR #766, opened 2017-01-13
Add further explaination to `Route.seal` method

vossad01:
The existing text is technically clear on what the `seal` method does, but is not behaviorally clear to a user who does not know what "wrapping it with exception handling and rejection conversion" does. 

I believe the added sentence helps to communicate what a user should expect when they use the `seal` method.  The hope is that with an expanded description users will be less likely to need to "try it out" to see if the behavior matches what they think `seal` might do.

I'm not at all tied to the sentence I came up and would welcome better suggestions.

### DIFF
--- a/akka-http/src/main/scala/akka/http/scaladsl/server/Route.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/server/Route.scala
@@ -22,6 +22,8 @@ object Route {
 
   /**
    * "Seals" a route by wrapping it with exception handling and rejection conversion.
+   *
+   * A "sealed" route accepts all requests making any routes after it unreachable at runtime.
    */
   def seal(route: Route)(implicit
     routingSettings: RoutingSettings,


ktoso:
OK TO TEST

akka-http/src/main/scala/akka/http/scaladsl/server/Route.scala
```diff
@@ -22,6 +22,8 @@ object Route {
 
   /**
    * "Seals" a route by wrapping it with exception handling and rejection conversion.
+   *
+   * A "sealed" route accepts all requests making any routes after it unreachable at runtime.
```
ktoso: hm I find this being weird wording. It'd be better to explain that it converts rejections to responses etc

vossad01: > `seal` converts all rejections and exceptions in the wrapped route to responses.  Any any routes concatenated after it will be unreachable at runtime.

^?

I'm afraid I am not yet well enough versed in Akka HTTP to know if  the "etc" in your response confers something I don't even know about.

For me the important part was mainly the "causing any routes concatenated after it to be unreachable at runtime." (slightly modified from the above) since that seems to be the behavior user would see when reasoning about what it does.

jlprat:
Would it make sense to add the same clarification to the Java API? 
[Route#seal(ActorSystem, Materializer)](https://github.com/akka/akka-http/blob/master/akka-http/src/main/scala/akka/http/javadsl/server/Route.scala#L45)
[Route#seal(RoutingSettings, ParserSettings, RejectionHandler, ExceptionHandler, ActorSystem, Materializer)](https://github.com/akka/akka-http/blob/master/akka-http/src/main/scala/akka/http/javadsl/server/Route.scala#L50)

jrudolph:
@vossad01 what do you think about the alternative improvement in #822? 

Closing this one.

Issue #760, opened 2017-01-12
JavaDSL Docs. Http client decoding response example in docs


jlprat:
As a result of issue #685 there is now a Scala example of client decoding response. Please provide a similar example and documentation for the Java DSL.

jlprat:
PR for reference 
https://github.com/akka/akka-http/pull/686

ktoso:
Can be closed, was done 👍

jlprat:
Well, there is this PR... https://github.com/akka/akka-http/pull/771

ktoso:
Ah ok missed the relation between those :~

Issue #595, opened 2016-11-29
FAILED: Unexpected message in HttpServerSpec


ktoso:
https://jenkins.akka.io:8498/job/pr-validator-akka-http/371/testReport/junit/akka.http.impl.engine.server/HttpServerSpec/The_server_implementation_should_support_request_timeouts_which_are_programmatically_increased__not_expiring_/

ortigali:
This failure may occur when it is too late to [updateTimeout](https://github.com/akka/akka-http/blob/5932237a86a432d623fafb1e84eeeff56d7485fe/akka-http-core/src/test/scala/akka/http/impl/engine/server/HttpServerSpec.scala#L845). because of [here](https://github.com/akka/akka-http/blob/5932237a86a432d623fafb1e84eeeff56d7485fe/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala#L337) (maybe GC pause influence it)

I could not find how to know updateTimeout is successful or not. 
I propose adding method to TimeoutAccess class to get current timeout
``` scala
  def timeout: Future[Duration] // this is hard to use
  // or
  def timeout: Duration  // this is good, but with locking
```

After updateTimeout we can check timeout changed or not by TimeoutAccess.timeout method.

If this proposal is correct, may I try it?

jlprat:
Another example, if needed: https://jenkins.akka.io:8498/job/pr-validator-akka-http/745/testReport/junit/akka.http.impl.engine.server/HttpServerSpec/The_server_implementation_should_support_request_timeouts_which_are_programmatically_increased__not_expiring_/

jrudolph:
https://jenkins.akka.io:8498/job/pr-validator-akka-http/1221/testReport/junit/akka.http.impl.engine.server/HttpServerSpec/The_server_implementation_should_support_request_timeouts_which_are_programmatically_increased__not_expiring_/

jonas:
https://jenkins.akka.io:8498/job/pr-validator-akka-http/1226/testReport/junit/akka.http.impl.engine.server/HttpServerSpec/The_server_implementation_should_support_request_timeouts_which_are_programmatically_increased__not_expiring_/

raboof:
https://jenkins.akka.io:8498/job/pr-validator-akka-http/1255/

raboof:
[...]

PR #771, opened 2017-01-15
Provide Decoding Response example for Java #760

jlprat:
Issue: #760
Provide an example to show how to decode a response using the Java DSL
Add missing method in `HttpResponse` to retrieve he encoding

### DIFF
--- a/akka-http-core/src/main/mima-filters/10.0.0.backwards.excludes
+++ b/akka-http-core/src/main/mima-filters/10.0.0.backwards.excludes
@@ -13,4 +13,3 @@ ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.javadsl.model.Ur
 ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.javadsl.model.Uri.getUserInfo")
 ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.javadsl.model.Uri.getPort")
 ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.javadsl.model.Uri.getScheme")
-ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.javadsl.model.HttpMessage#MessageTransformations.transformEntityDataBytes")
--- a/akka-http-core/src/main/mima-filters/10.0.0.backwards.excludes
+++ b/akka-http-core/src/main/mima-filters/10.0.0.backwards.excludes
@@ -13,3 +13,4 @@ ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.javadsl.model.Ur
 ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.javadsl.model.Uri.getUserInfo")
 ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.javadsl.model.Uri.getPort")
 ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.javadsl.model.Uri.getScheme")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.javadsl.model.HttpMessage#MessageTransformations.transformEntityDataBytes")
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
@@ -4,6 +4,8 @@
 
 package akka.http.scaladsl.model
 
+import akka.stream.scaladsl.Flow
+import akka.stream.{ FlowShape, Graph }
 import java.io.File
 import java.nio.file.Path
 import java.lang.{ Iterable ⇒ JIterable }
@@ -270,6 +272,8 @@ final class HttpRequest(
   override def withUri(path: String): HttpRequest = withUri(Uri(path))
   def withUri(uri: Uri): HttpRequest = copy(uri = uri)
 
+  def transformEntityDataBytes[T](transformer: Graph[FlowShape[ByteString, ByteString], T]): HttpRequest = copy(entity = entity.transformDataBytes(Flow.fromGraph(transformer)))
+
   import JavaMapping.Implicits._
   /** Java API */
   override def getUri: jm.Uri = uri.asJava
@@ -369,7 +373,7 @@ object HttpRequest {
       }
     }
 
-  /* Manual Case Class things, to easen bin-compat */
+  /* Manual Case Class things, to ease bin-compat */
 
   def apply(
     method:   HttpMethod                = HttpMethods.GET,
@@ -417,7 +421,9 @@ final class HttpResponse(
 
   def mapEntity(f: ResponseEntity ⇒ ResponseEntity): HttpResponse = withEntity(f(entity))
 
-  /* Manual Case Class things, to easen bin-compat */
+  def transformEntityDataBytes[T](transformer: Graph[FlowShape[ByteString, ByteString], T]): HttpResponse = copy(entity = entity.transformDataBytes(Flow.fromGraph(transformer)))
+
+  /* Manual Case Class things, to ease bin-compat */
 
   def copy(
     status:   StatusCode                = status,
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
@@ -142,7 +142,7 @@ sealed trait HttpMessage extends jm.HttpMessage {
   def withEntity(contentType: jm.ContentType, file: File): Self = withEntity(HttpEntity(contentType.asInstanceOf[ContentType], file))
   def withEntity(contentType: jm.ContentType, file: Path): Self = withEntity(HttpEntity.fromPath(contentType.asInstanceOf[ContentType], file))
 
-  def transformEntityDataBytes[T](transformer: Graph[FlowShape[ByteString, ByteString], T]): Self
+  def transformEntityDataBytes[M](transformer: Graph[FlowShape[ByteString, ByteString], M]): Self
 
   import collection.JavaConverters._
   /** Java API */
@@ -274,7 +274,7 @@ final class HttpRequest(
   override def withUri(path: String): HttpRequest = withUri(Uri(path))
   def withUri(uri: Uri): HttpRequest = copy(uri = uri)
 
-  def transformEntityDataBytes[T](transformer: Graph[FlowShape[ByteString, ByteString], T]): HttpRequest = copy(entity = entity.transformDataBytes(Flow.fromGraph(transformer)))
+  def transformEntityDataBytes[M](transformer: Graph[FlowShape[ByteString, ByteString], M]): HttpRequest = copy(entity = entity.transformDataBytes(Flow.fromGraph(transformer)))
 
   import JavaMapping.Implicits._
   /** Java API */
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
@@ -142,6 +142,8 @@ sealed trait HttpMessage extends jm.HttpMessage {
   def withEntity(contentType: jm.ContentType, file: File): Self = withEntity(HttpEntity(contentType.asInstanceOf[ContentType], file))
   def withEntity(contentType: jm.ContentType, file: Path): Self = withEntity(HttpEntity.fromPath(contentType.asInstanceOf[ContentType], file))
 
+  def transformEntityDataBytes[T](transformer: Graph[FlowShape[ByteString, ByteString], T]): Self
+
   import collection.JavaConverters._
   /** Java API */
   def getHeaders: JIterable[jm.HttpHeader] = (headers: immutable.Seq[jm.HttpHeader]).asJava
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
@@ -31,7 +31,7 @@ import headers._
  * Common base class of HttpRequest and HttpResponse.
  */
 sealed trait HttpMessage extends jm.HttpMessage {
-  type Self <: HttpMessage
+  type Self <: HttpMessage { type Self = HttpMessage.this.Self }
   def self: Self
 
   def isRequest: Boolean
--- a/akka-http/src/main/java/akka/http/javadsl/coding/Coder.java
+++ b/akka-http/src/main/java/akka/http/javadsl/coding/Coder.java
@@ -6,9 +6,13 @@ package akka.http.javadsl.coding;
 
 import java.util.concurrent.CompletionStage;
 
+import akka.http.javadsl.model.HttpRequest;
+import akka.http.javadsl.model.HttpResponse;
+import akka.http.javadsl.model.headers.ContentEncoding;
 import akka.http.scaladsl.coding.Deflate$;
 import akka.http.scaladsl.coding.Gzip$;
 import akka.http.scaladsl.coding.NoCoding$;
+import akka.http.scaladsl.model.HttpMessage;
 import akka.stream.Materializer;
 import akka.util.ByteString;
 import scala.compat.java8.FutureConverters;
@@ -25,9 +29,40 @@ public enum Coder {
         this.underlying = underlying;
     }
 
+    public HttpResponse encodeMessage(HttpResponse message) {
+        boolean shouldApply = (boolean) underlying.messageFilter().apply((HttpMessage) message);
+        if (shouldApply && message.getHeader(ContentEncoding.class).isPresent()) {
+            return message.transformEntityDataBytes(underlying.newEncodeTransformer());
+        }
+        return message;
+    }
+
+    public HttpRequest encodeMessage(HttpRequest message) {
+        boolean shouldApply = (boolean) underlying.messageFilter().apply((HttpMessage) message);
+        if (shouldApply && message.getHeader(ContentEncoding.class).isPresent()) {
+            return message.transformEntityDataBytes(underlying.newEncodeTransformer());
+        }
+        return message;
+    }
+
     public ByteString encode(ByteString input) {
         return underlying.encode(input);
     }
+
+    public HttpResponse decodeMessage(HttpResponse message) {
+        if (message.getHeader(ContentEncoding.class).isPresent()) {
+            return message.transformEntityDataBytes(underlying.decoderFlow());
+        }
+        return message;
+    }
+
+    public HttpRequest decodeMessage(HttpRequest message) {
+        if (message.getHeader(ContentEncoding.class).isPresent()) {
+            return message.transformEntityDataBytes(underlying.decoderFlow());
+        }
+        return message;
+    }
+
     public CompletionStage<ByteString> decode(ByteString input, Materializer mat) {
         return FutureConverters.toJava(underlying.decode(input, mat));
     }
--- a/akka-http/src/main/java/akka/http/javadsl/coding/Coder.java
+++ b/akka-http/src/main/java/akka/http/javadsl/coding/Coder.java
@@ -30,19 +30,11 @@ public enum Coder {
     }
 
     public HttpResponse encodeMessage(HttpResponse message) {
-        boolean shouldApply = (boolean) underlying.messageFilter().apply((HttpMessage) message);
-        if (shouldApply && message.getHeader(ContentEncoding.class).isPresent()) {
-            return message.transformEntityDataBytes(underlying.newEncodeTransformer());
-        }
-        return message;
+        return (HttpResponse) underlying.encodeMessage((akka.http.scaladsl.model.HttpMessage) message);
     }
 
     public HttpRequest encodeMessage(HttpRequest message) {
-        boolean shouldApply = (boolean) underlying.messageFilter().apply((HttpMessage) message);
-        if (shouldApply && message.getHeader(ContentEncoding.class).isPresent()) {
-            return message.transformEntityDataBytes(underlying.newEncodeTransformer());
-        }
-        return message;
+        return (HttpRequest) underlying.encodeMessage((akka.http.scaladsl.model.HttpMessage) message);
     }
 
     public ByteString encode(ByteString input) {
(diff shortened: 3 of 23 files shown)


jlprat:
I'm not so sure, if this is the best way to decode a response, the Java version is pretty verbose compared to the Scala one

jlprat:
PLS BUILD

jrudolph:
PLS BUILD

jlprat:
Ready for review

docs/src/test/java/docs/http/javadsl/HttpClientDecodingExample.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.concurrent.CompletionStage;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+public class HttpClientDecodingExample {
+
+  public static void main(String[] args) throws Exception {
+
+    final ActorSystem system = akka.actor.ActorSystem.create();
```
jonas: Any reason this is using fully qualified class name?

jlprat: Absolutely not,  I'll change it

jlprat:
I think it's quite nice to have both APIs at the same level, so yes, I'm in. Should I do it in this same PR?

jrudolph:
Do it as you like :)

jlprat:
I'll do a deep dive into the transformations part in order to provide a solution for this, so it might take a little longer

jlprat:
I think I'm missing one tiny bit in here. I'll write down what puzzles me.

AFAIK, the purpose of `HttpMessage.MessageTransformations.transformEntityDataBytes` would be to delegate the Data bytes transformation to its `HttpEntity`. This method should take a `Flow`, my first question would be: which one (Java or Scala DSL)? Any choice feels wrong to me as this method will end up in both `HttpRequest` and `HttpResponse` for both APIs.

Once both `HttpRequest` and `HttpResponse` end up having the `transformEntityDataBytes` method, I guess `DataMapper` is no longer needed for the Scala `Coder`, right?

Thanks in advance!

jrudolph:
> AFAIK, the purpose of HttpMessage.MessageTransformations.transformEntityDataBytes would be to delegate the Data bytes transformation to its HttpEntity. This method should take a Flow, my first question would be: which one (Java or Scala DSL)? Any choice feels wrong to me as this method will end up in both HttpRequest and HttpResponse for both APIs.

You can use `Graph[FlowShape[In, Out], Mat]` which is extended by both Scala and Java `Flow`.

> Once both HttpRequest and HttpResponse end up having the transformEntityDataBytes method, I guess DataMapper is no longer needed for the Scala Coder, right?

We don't need it for `Decoder.decode` but we still need it for `Decoder.decodeData` if someone wants to run it on an `HttpEntity` directly. Not sure if that's an important use case but we would have to think about compatibility if we want to remove `decodeData`.

jlprat:
Forgot to update MiMa filters and now I see this PR has conflicts, I will rebase

docs/src/test/scala/docs/http/scaladsl/HttpClientDecodingExampleSpec.scala
```diff
@@ -47,6 +47,8 @@ class HttpClientDecodingExampleSpec extends WordSpec with Matchers with CompileO
 
     val futureResponses: Future[Seq[HttpResponse]] =
       Future.traverse(requests)(http.singleRequest(_).map(decodeResponse))
+
+    system.terminate()
```
jlprat: Scala example was not terminating the system making the example hang

akka-http/src/main/java/akka/http/javadsl/coding/Coder.java
```diff
@@ -25,9 +29,40 @@
         this.underlying = underlying;
     }
 
+    public HttpResponse encodeMessage(HttpResponse message) {
```
jlprat: My Java generic typing Foo could not make it work in a better way, like for example expressing a `T` that extends both `HttpMessage` and `MessageTransformations`.
Comment is valid for the other `encodeMessage` and `decodeMessage` methods.

akka-http/src/main/java/akka/http/javadsl/coding/Coder.java
```diff
@@ -25,9 +29,40 @@
...
+        boolean shouldApply = (boolean) underlying.messageFilter().apply((HttpMessage) message);
```
jlprat: `JavaMappings` is not so convenient to be used in Java code, so I fell back to the un-fancy cast.

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
```diff
@@ -417,7 +421,9 @@ final class HttpResponse(
 
   def mapEntity(f: ResponseEntity ⇒ ResponseEntity): HttpResponse = withEntity(f(entity))
 
-  /* Manual Case Class things, to easen bin-compat */
+  def transformEntityDataBytes[T](transformer: Graph[FlowShape[ByteString, ByteString], T]): HttpResponse = copy(entity = entity.transformDataBytes(Flow.fromGraph(transformer)))
```
jlprat: It's a `T` because some may pass a `Materializer`, others maybe not.

jlprat:
PLS BUID

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
jrudolph: Might also work with just `Any`.

akka-http/src/main/mima-filters/10.0.3.backwards.excludes
```diff
@@ -13,3 +13,7 @@ ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.scaladsl.marshal
 
 # make httpApp abstract class
 ProblemFilters.exclude[IncompatibleTemplateDefProblem]("akka.http.scaladsl.server.HttpApp")
+
+
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.scaladsl.coding.Encoder.encodeMessage")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.scaladsl.coding.Decoder.decodeMessage")
```
jrudolph: These ones shouldn't be needed. Only the 10.0.4 ones should be needed. MiMa aggregates them for previous versions.

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
jrudolph: I see if that's also used in Java it would be problem with `Any` as Java doesn't respect or know about variance declarations.

akka-http/src/main/java/akka/http/javadsl/coding/Coder.java
jrudolph: Hmm, might just not work. But let's not reimplement any stuff here. Why not just `(HttpResponse) underlying.encodeMessage(message)`?

akka-http/src/main/scala/akka/http/scaladsl/coding/Decoder.scala
```diff
@@ -5,18 +5,24 @@
...
 import headers.HttpEncoding
-import akka.stream.scaladsl.{ Sink, Source, Flow }
+import akka.stream.scaladsl.{ Flow, Sink, Source }
 
 import scala.concurrent.Future
 
 trait Decoder {
   def encoding: HttpEncoding
 
+  def decodeMessage[T <: HttpMessage with MessageTransformations[T]](message: T): T#Self =
```
jrudolph: Could this be `def decodeMessage(message: HttpMessage): message.Self`? And please no copy pasting ;) Just delegate to the below method. We might need to introduce another `DataMapper[HttpMessage]`.

docs/src/test/java/docs/http/javadsl/HttpClientDecodingExampleTest.java
```diff
@@ -0,0 +1,73 @@
+/*
```
jrudolph: Nice example.

akka-http/src/main/java/akka/http/javadsl/coding/Coder.java
jlprat: This will not work, because `underlying.encodeMessage` needs a type that extends both `HttpMessage` and `MessageTranformations`. So what I could do is something like: `(HttpResponse) underlying.encodeMessage((SHttpResponse) response)`
Where `SHttpResponse` is the Scala one.

akka-http/src/main/mima-filters/10.0.3.backwards.excludes
jlprat: This PR started long time ago, and it was not in sync. I will remove them

akka-http/src/main/scala/akka/http/scaladsl/coding/Decoder.scala
jlprat: OK

jlprat:
I will address the comments and add a test for the Java functionality. Thanks!

akka-http/src/main/java/akka/http/javadsl/coding/Coder.java
jrudolph: Ah, I see. You can add `def transformEntityDataBytes[T](transformer: Graph[FlowShape[ByteString, ByteString], T]): Self` to `scaladsl.HttpMessage` which will be implemented by the concrete implementations in `HttpRequest` and `HttpResponse`. Then

jlprat: Cool, I wasn't sure to add this method, then I can simplify the signature of the Scala's `encodeMessage` and `decodeMessage` to just take `HttpMessage` (instead of `HttpMessage with MessageTransformers`).
Then, I also question if the `transformEntityDataBytes` should be in `MessageTransformers` at all.

jrudolph:
> Then, I also question if the transformEntityDataBytes should be in MessageTransformers at all.

That's the way how it gets into the Java API.

jlprat:
> That's the way how it gets into the Java API.

Ah, OK I was speaking from the top of my head.

akka-http/src/main/scala/akka/http/scaladsl/coding/Decoder.scala
jlprat: Shouldn't be the other way round? `decode` calling `decodeMessage`, otherwise what is the purpose of the `transformEntityDataBytes` method?

akka-http/src/main/mima-filters/10.0.3.backwards.excludes
jlprat: Apparently, they are needed

akka-http/src/main/java/akka/http/javadsl/coding/Coder.java
jlprat: Now they rely on the underlying encoder/decoder

akka-http/src/main/scala/akka/http/scaladsl/coding/Decoder.scala
```diff
@@ -10,18 +10,21 @@ import akka.stream.{ FlowShape, Materializer }
...
-import akka.stream.scaladsl.{ Sink, Source, Flow }
+import akka.stream.scaladsl.{ Flow, Sink, Source }
 
 import scala.concurrent.Future
 
 trait Decoder {
   def encoding: HttpEncoding
 
-  def decode[T <: HttpMessage](message: T)(implicit mapper: DataMapper[T]): T#Self =
+  def decodeMessage(message: HttpMessage): HttpMessage#Self =
```
jlprat: Keeping one implementation `decodeMessage` and having the old one calling this.

akka-http/src/main/scala/akka/http/scaladsl/coding/Decoder.scala
```diff
@@ -10,18 +10,21 @@ import akka.stream.{ FlowShape, Materializer }
...
     if (message.headers exists Encoder.isContentEncodingHeader)
-      decodeData(message).withHeaders(message.headers filterNot Encoder.isContentEncodingHeader)
+      message.transformEntityDataBytes(decoderFlow).withHeaders(message.headers filterNot Encoder.isContentEncodingHeader)
     else message.self
 
+  def decode[T <: HttpMessage](message: T)(implicit mapper: DataMapper[T]): T#Self =
+    decodeMessage(message).asInstanceOf[T#Self]
```
jlprat: I didn't know how else I could make the compiler happy

jlprat:
Regarding the tests for the Java functionality, I could write them, but the current Java implementation is just delegating to the Scala implementation. Should I still add them?

akka-http-core/src/main/mima-filters/10.0.0.backwards.excludes
```diff
@@ -13,3 +13,4 @@ ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.javadsl.model.Ur
 ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.javadsl.model.Uri.getUserInfo")
 ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.javadsl.model.Uri.getPort")
 ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.javadsl.model.Uri.getScheme")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.javadsl.model.HttpMessage#MessageTransformations.transformEntityDataBytes")
```
jrudolph: Please don't add anything to old mima-filters. Adding to the 10.0.5 version should be enough.

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
```diff
@@ -140,6 +142,8 @@ sealed trait HttpMessage extends jm.HttpMessage {
   def withEntity(contentType: jm.ContentType, file: File): Self = withEntity(HttpEntity(contentType.asInstanceOf[ContentType], file))
   def withEntity(contentType: jm.ContentType, file: Path): Self = withEntity(HttpEntity.fromPath(contentType.asInstanceOf[ContentType], file))
 
+  def transformEntityDataBytes[T](transformer: Graph[FlowShape[ByteString, ByteString], T]): Self
```
jrudolph: :+1:

akka-http/src/main/java/akka/http/javadsl/coding/Coder.java
jrudolph: :+1:

akka-http/src/main/scala/akka/http/scaladsl/coding/Decoder.scala
jrudolph: I think this can still be improved. We need to fix the `Self` type in `HttpMessage` to make it work (props to @rkuhn who figured out how to tie the recursive self-type knot for `FlowOps` a while ago).

If we change the `Self` type in `HttpMessage` to this:

```scala
type Self <: HttpMessage { type Self = HttpMessage.this.Self }
```

the type signature for `decodeMessage` can become

```scala
def decodeMessage(message: HttpMessage): message.Self
```

which means that whatever `HttpMessage` type is passed in will also come out.

akka-http/src/main/scala/akka/http/scaladsl/coding/Decoder.scala
jrudolph: See above ;)

akka-http/src/main/scala/akka/http/scaladsl/coding/Decoder.scala
```diff
@@ -10,18 +10,21 @@ import akka.stream.{ FlowShape, Materializer }
...
   def encoding: HttpEncoding
 
-  def decode[T <: HttpMessage](message: T)(implicit mapper: DataMapper[T]): T#Self =
+  def decodeMessage(message: HttpMessage): HttpMessage#Self =
     if (message.headers exists Encoder.isContentEncodingHeader)
-      decodeData(message).withHeaders(message.headers filterNot Encoder.isContentEncodingHeader)
+      message.transformEntityDataBytes(decoderFlow).withHeaders(message.headers filterNot Encoder.isContentEncodingHeader)
     else message.self
 
+  def decode[T <: HttpMessage](message: T)(implicit mapper: DataMapper[T]): T#Self =
```
jrudolph: I think we can now deprecate this old type signature in favor of `decodeMessage`. Could you add the deprecation annotation?

akka-http/src/main/scala/akka/http/scaladsl/coding/Encoder.scala
```diff
@@ -18,11 +18,14 @@ trait Encoder {
 
   def messageFilter: HttpMessage ⇒ Boolean
 
-  def encode[T <: HttpMessage](message: T)(implicit mapper: DataMapper[T]): T#Self =
```
jrudolph: Same changes as for `Decoder`.

akka-http/src/main/scala/akka/http/scaladsl/coding/Decoder.scala
jlprat: Kudos to @rkuhn!! I will try this later and try to understand the trickery behind it :)

akka-http-core/src/main/mima-filters/10.0.0.backwards.excludes
jlprat: Actually it was complaining that needed to be added to this file, I don't know why. Maybe it's because this PR started long time ago and mima got confused...

akka-http/src/main/scala/akka/http/scaladsl/coding/Decoder.scala
jlprat: Agree, didn't want to deprecate until you agreed on the solution

akka-http/src/main/scala/akka/http/scaladsl/coding/Decoder.scala
rkuhn: Glad I could help! :-)

jlprat:
WOPS rebase went wrong...

jlprat:
Ready to review.
As `decoder.decode` and `encoder.encode` are now deprecated, I guess we should go through the docs and update any occurrence with the new alternatives. I could do it here or in a different PR.

docs/src/test/java/docs/http/javadsl/HttpClientDecodingExampleTest.java
```diff
@@ -0,0 +1,73 @@
...
+    final Function<HttpResponse, HttpResponse> decodeResponse = response -> {
+      // Pick the right coder
+      final Coder coder;
+      if (HttpEncodings.gzip().equals(response.encoding())) {
+        coder = Coder.Gzip;
+      } else if (HttpEncodings.deflate().equals(response.encoding())) {
+        coder = Coder.Deflate;
+      } else {
+        coder = Coder.NoCoding;
+      }
```
2beaucoup: We could do something like
```java
final List<Coder> decoders = Arrays.asList(GZip, Deflate, NoCoding)
final Coder = decoders.filter(d -> d.encoding.equals(response.encoding)).findFirst()
```
work here (same for Scala)?

akka-http/src/main/mima-filters/10.0.4.backwards.excludes
```diff
@@ -0,0 +1,3 @@
+# Provide Decoding Response example for Java #760
```
jrudolph: File can be removed.

docs/src/test/scala/docs/http/scaladsl/HttpClientDecodingExampleSpec.scala
jlprat: I didn't look much on the Scala example as it was already there and approved. I agree with you, maybe it makes sense to collect the answers and log them.

akka-http/src/main/scala/akka/http/scaladsl/coding/Decoder.scala
```diff
@@ -10,18 +10,21 @@ import akka.stream.{ FlowShape, Materializer }
...
   def encoding: HttpEncoding
 
-  def decode[T <: HttpMessage](message: T)(implicit mapper: DataMapper[T]): T#Self =
+  def decodeMessage(message: HttpMessage): message.Self =
     if (message.headers exists Encoder.isContentEncodingHeader)
-      decodeData(message).withHeaders(message.headers filterNot Encoder.isContentEncodingHeader)
+      message.transformEntityDataBytes(decoderFlow).withHeaders(message.headers filterNot Encoder.isContentEncodingHeader)
     else message.self
 
+  @deprecated("Use Decoder#decodeMessage instead. No need for implicit mapper.", since = "10.0.6")
```
jrudolph: :+1:

docs/src/test/java/docs/http/javadsl/HttpClientDecodingExampleTest.java
jlprat: It is definitely more concise, and I guess it's still clear enough for the Java mainstream reader.

akka-http/src/main/mima-filters/10.0.4.backwards.excludes
jlprat: I knew I would forget to clean up at least one Mima file!

docs/src/test/java/docs/http/javadsl/HttpClientDecodingExampleTest.java
jlprat: Not so easy for the Java code as `Coder` doesn't have access to `encoding`. Maybe this could be part of another PR.

jlprat:
I just realized I forgot to update all references to the deprecated `encode` and `decode` to the new `encodeMessage` and `decodeMessage`.

jlprat:
Failures are: #299 and #595 
PLS BUILD

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
ktoso: Yeah, so the `T` is very good here

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
```diff
@@ -274,7 +274,7 @@ final class HttpRequest(
   override def withUri(path: String): HttpRequest = withUri(Uri(path))
   def withUri(uri: Uri): HttpRequest = copy(uri = uri)
 
-  def transformEntityDataBytes[T](transformer: Graph[FlowShape[ByteString, ByteString], T]): HttpRequest = copy(entity = entity.transformDataBytes(Flow.fromGraph(transformer)))
+  def transformEntityDataBytes[M](transformer: Graph[FlowShape[ByteString, ByteString], M]): HttpRequest = copy(entity = entity.transformDataBytes(Flow.fromGraph(transformer)))
```
ktoso: made it `M`, like we do for Materialized values everywhere

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
```diff
@@ -369,7 +375,7 @@ object HttpRequest {
       }
     }
 
-  /* Manual Case Class things, to easen bin-compat */
+  /* Manual Case Class things, to ease bin-compat */
```
ktoso: thanks :)

jlprat:
Yeah it was way more than expected! :smiley:
It started as a simple Scala to Java example conversion

Issue #194, opened 2016-09-08
Increase timeouts and add dilated to HTTP test durations


ktoso:
<a href="https://github.com/drewhk"><img src="https://avatars.githubusercontent.com/u/932285?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [drewhk](https://github.com/drewhk)**
_Wednesday Oct 08, 2014 at 13:26 GMT_
_Originally opened as https://github.com/akka/akka/issues/16054_

---

ktoso:
<a href="https://github.com/rkuhn"><img src="https://avatars.githubusercontent.com/u/470469?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [rkuhn](https://github.com/rkuhn)**
_Wednesday Mar 23, 2016 at 16:11 GMT_

----

This is important and should be done soon—many HTTP tests have unsuitably short timeouts that lead to CI flakiness.

ktoso:
<a href="https://github.com/ktoso"><img src="https://avatars.githubusercontent.com/u/120979?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [ktoso](https://github.com/ktoso)**
_Thursday Mar 24, 2016 at 17:14 GMT_

----

I agree, will do it next week

jrudolph:
Closed by #781.

Issue #434, opened 2016-10-26
Try Appveyor to run automated tests on Windows


jrudolph:
It's free for OSS. See https://ci.appveyor.com/signup/free

PR #781, opened 2017-01-17
Add dilated to test durations

jonas:
No `.dilated` added in `docs/` test or test apps. When possible the application of `.dilated` is pushed into test abstractions to ensure that future updates automatically also use it.

I didn't track build failures explicitly recently, but didn't see any timeout failure issues being created so the part of #194 related with increasing timeouts we can leave for later, and we already did some back in November, e.g.see #414.

### DIFF
--- a/akka-http-core/src/test/scala/akka/http/impl/engine/client/ConnectionPoolSpec.scala
+++ b/akka-http-core/src/test/scala/akka/http/impl/engine/client/ConnectionPoolSpec.scala
@@ -21,7 +21,7 @@ import akka.stream.ActorMaterializer
 import akka.stream.TLSProtocol._
 import akka.stream.scaladsl._
 import akka.stream.testkit.{ TestPublisher, TestSubscriber }
-import akka.testkit.AkkaSpec
+import akka.testkit._
 import akka.util.ByteString
 
 import scala.collection.immutable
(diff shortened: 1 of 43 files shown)


jonas:
This should also make it easier to get more consistent result on AppVeyor (#434).

BTW, one note is that I didn't touch timeouts in the Java code.

jrudolph:
@jonas have you already tried on AppVeyor if that helps? Seems like quite a big change...

jonas:
I didn't try yet. Will do.

jonas:
I have fixed test that failed locally with timefactor=1.5 and ran a couple of test on AppVeyor yesterday which all end up hanging. In one of the build the coroner reported a deadlock in the HttpAppSpec. The two known failing tests on Windows still fail, however I didn't see any new timeout failures as before.

jlprat:
Did some investigation on the HttpAppSpec tests:
* Only one spec is check each run (either the Java or the Scala one)
* After https://github.com/akka/akka-http/pull/849 there is only a deadlock when some assertion fails
* Looks like when assertion fails, server is not shut down correctly, so thread is blocked at that point

jrudolph:
PLS BUILD

jrudolph:
Great, thanks @jonas. Actually I think we can close #194 for now and follow up should there be new failures.

jonas:
Thanks, there's one minor tweak from the AppVeyor work I didn't cherry pick
into this PR. Will make a follow-up.

On Feb 16, 2017 8:32 AM, "Johannes Rudolph" <notifications@github.com>
wrote:

Great, thanks @jonas <https://github.com/jonas>. Actually I think we can
close #194 <https://github.com/akka/akka-http/issues/194> for now and
follow up should there be new failures.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/akka/akka-http/pull/781#issuecomment-280330953>, or mute
the thread

Issue #485, opened 2016-11-04
Specs2 support


fokot:
Hi guys, do you plan to support specs2 for testing.
I created simple specs2 support for testing routes. I can create a pull request. I didn't do it now as I'm not sure you would like to have new sub project in sbt etc. This is it also with usage example. Right now I support matchers only on entity as string but doing unmarshalling and also support for http codes etc. It is here https://gist.github.com/fokot/3827d2f0fc2cb02b3c4f65d8344f2f18

ktoso:
That'd be cool, yes :-)
We're not that familiar with Specs2 (well Johannes is I think), but most of us not really.

Would be excellent if you could PR and maintain it in case of upgrades needed?

Yes, separate project is the way to go I think. Would also need docs :)

viktorklang:
Or, if may I suggest, having it completely stand-alone, since supporting a
lot of different test fwks is not a core concern of the library itself?

-- 
Cheers,
√

jonas:
Ref (or duplicate) #232

fokot:
@viktorklang having it in this repo might be easier for publishing to maven central.. I'm not familiar with that at all. But if you want I can create a stand-alone project. Or I can just create a blog post with this gist, so ppl know how to test routes with specs2.. So options are:
1) new project in akka-http
2) standalone project
3) blog
4) nothing (it's easy for anyone to figure it out)

jrudolph:
We had support for specs2 in spray-testkit which was just a few lines. I would not be concerned about maintaining it if someone could contribute it and if it would be the same amount of glue code.

In spray, we referenced specs2 as a provided dependency so it wouldn't be pulled in automatically.

Here's the Scalatest support in akka-http-testkit:

https://github.com/akka/akka-http/blob/master/akka-http-testkit/src/main/scala/akka/http/scaladsl/testkit/ScalatestUtils.scala

Here's the former specs2 support in spray:

https://github.com/spray/spray/blob/master/spray-testkit/src/main/scala/spray/testkit/Specs2Interface.scala

ktoso:
I agree with @jrudolph. It's something we had, and now don't, which makes upgrading harder for people (which we want, a lot), thus we should add it back.

Very happy to accept a PR if you'd have the time @fokot !

sergiuszkierat:
yes, it would be nice to have specs2 support again.

> //FIXME: trait Specs2RouteTest extends RouteTest with Specs2Interface

[...]

PR #782, opened 2017-01-17
Add support for Specs2 in testkit #485

tom-walford:
* Add support for Specs2 tests using the trait Specs2RouteTest in the same form as ScalatestRouteTest

This is just a wrapper similar to the ScalatestRouteTest trait - If there's anything missing or I've not done something properly please let me know :)


typesafehub-validator:
Hi @tom-walford,

akka-http-testkit/src/main/scala/akka/http/scaladsl/testkit/Specs2Utils.scala
```diff
@@ -0,0 +1,30 @@
+package akka.http.scaladsl.testkit
```
ktoso: please add copyright headers like all other files, thanks 👍

project/Dependencies.scala
```diff
@@ -124,7 +126,8 @@ object Dependencies {
 
   lazy val httpTestkit = l ++= Seq(
     Test.junit, Test.junitIntf, Compile.junit % "provided",
-    Test.scalatest.value.copy(configurations = Some("provided; test"))
+    Test.scalatest.value.copy(configurations = Some("provided; test")),
+    "org.specs2" %% "specs2-core" % specs2Version.value % "provided"
```
ktoso: I guess that's fine. Sanity check @jrudolph ?

akka-http-testkit/src/test/scala/akka/http/scaladsl/testkit/Specs2RouteTestSpec.scala
```diff
@@ -0,0 +1,81 @@
+/*
+ * Copyright (C) 2009-2016 Lightbend Inc. <http://www.lightbend.com>
```
jlprat: 2017 :)

jonas:
It would be nice if you could also mention support for Specs2 in `docs/src/main/paradox/scala/http/routing-dsl/testkit.md`, for example the following snippet added after the [scalatest example](https://github.com/akka/akka-http/blame/4ba9e61a05790bc38cf70a75bcbcdbf7df24dddf/docs/src/main/paradox/scala/http/routing-dsl/testkit.md#L19)

```
The testkit also supports testing routes with [specs2](http://etorreborre.github.io/specs2/). The equivalent of the above example can be found
@github[here](akka-http-testkit/src/test/scala/akka/http/scaladsl/testkit/Specs2RouteTestSpec.scala).
```

... with `docs/src/test/scala/docs/http/scaladsl/server/FullTestKitExampleSpec.scala` translated to specs2.

project/Dependencies.scala
```diff
@@ -20,6 +21,7 @@ object Dependencies {
     scalaVersion := crossScalaVersions.value.head,
     scalaCheckVersion := sys.props.get("akka.build.scalaCheckVersion").getOrElse("1.13.4"),
     scalaTestVersion := "3.0.0",
+    specs2Version := "3.8.5",
```
jonas: It [looks like](http://search.maven.org/#search%7Cgav%7C1%7Cg%3A%22org.specs2%22%20AND%20a%3A%22specs2-core_2.12%22) we have to use `3.8.6` to have Scala 2.12 supported.

project/Dependencies.scala
jonas: It would be nice to have it declared in the `Test` section with license info for completeness.

jonas:
OK TO TEST

project/Dependencies.scala
tom-walford: I've added the license info - which I believe is MIT (although I got that from maven-central so...)

project/Dependencies.scala
```diff
@@ -124,7 +127,8 @@ object Dependencies {
 
   lazy val httpTestkit = l ++= Seq(
     Test.junit, Test.junitIntf, Compile.junit % "provided",
-    Test.scalatest.value.copy(configurations = Some("provided; test"))
+    Test.scalatest.value.copy(configurations = Some("provided; test")),
+    Test.specs2.value
```
jonas: This will work as long as we don't want to use specs2 inside the `docs` project since it uses this dependency mapping `httpTestkit % "compile;test->test"`.

tom-walford:
@jonas Re: your comment about the example spec. I've updated the MD file, but i'm having trouble with the dependencies being inherited from the Testkit project to the docs project. It isn't picking up the specs2 library so i'm currently playing around with that trying to get that happy in a nice way

jonas:
Ah, OK then see [comment](https://github.com/akka/akka-http/pull/782#discussion_r96714659). If you use the same way of defining specs2 as scalatest this will work. Declare it in "test" scope instead of provided and then use it like this:

```
Test.specs2.value.copy(configurations = Some("provided; test"))
```

The problem is that httpTestkit doesn't map `provided` dependencies to the `docs`, using the same approach as for scalatest will fix this.

akka-http-testkit/src/main/scala/akka/http/scaladsl/testkit/Specs2Utils.scala
ktoso: Thanks :)

project/Dependencies.scala
ktoso: Yeah that's fine I think 👍

docs/src/main/paradox/scala/http/routing-dsl/testkit.md
```diff
@@ -18,6 +18,9 @@ Here is an example of what a simple test with the routing testkit might look lik
 
 @@snip [FullTestKitExampleSpec.scala](../../../../../test/scala/docs/http/scaladsl/server/FullTestKitExampleSpec.scala) { #source-quote }
 
+The testkit also supports testing routes with [specs2](http://etorreborre.github.io/specs2/). The equivalent of the above example can be found
+@github[here](akka-http-testkit/src/test/scala/akka/http/scaladsl/testkit/Specs2RouteTestSpec.scala).
```
jonas: The last nitpicking thing from me would be to change this to refer to `docs/src/test/scala/docs/http/scaladsl/server/FullSpecs2TestKitExampleSpec.scala`.

jonas:
LGTM, looking forward to using this!

project/Dependencies.scala
jrudolph: Can we keep the style consistent with scalatest here? I.e. keep the config as `"test"` in the `Test` object above and use `copy` here.

akka-http-testkit/src/main/scala/akka/http/scaladsl/testkit/Specs2Utils.scala
tom-walford: no worries :)

tom-walford:
Done :)

ktoso:
Hm, that's not squashed :-) Squshed would mean that this PR contains one single commit. You can get there by doing `git rebase --interactive master` and marking most commits as `s` (squash). 
I'll merge this manually in a bit, if you don't have the time to rebase, cheers and thanks for the PR!

ktoso:
Or I'll figure out a way to fix our script to understand the GitHubs weird (IMO broken, which we reported) squash and merge button hm.

tom-walford:
Ah sorry - I did debase -i but I guess j broke the push. I'll try and take a look tomorrow :-)

jrudolph:
Github squashing works fine when there's only a single author who opened the PR. We should just double check the history that attribution is still ok after we squash-merged.

jrudolph:
Thanks again, @tom-walford. Great first contribution :)

PR #787, opened 2017-01-17
A set of smaller changes related extracted from #764

jrudolph:
Here's a first set of actual changes I extracted from #764. I think they are good as they are so we can merge them after pr validation.

### DIFF
--- a/akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
+++ b/akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
@@ -513,7 +513,26 @@ class Http2ServerSpec extends AkkaSpec with WithInPendingUntilFixed with Eventua
     }
 
     "support low-level features" should {
-      "respond to PING frames" in pending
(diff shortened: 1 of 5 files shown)


ktoso:
Ok thanks

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -515,7 +513,26 @@ class Http2ServerSpec extends AkkaSpec with WithInPendingUntilFixed with Eventua
     }
 
     "support low-level features" should {
-      "respond to PING frames" in pending
+      "respond to PING frames (spec 6_7)" in new TestSetup with RequestResponseProbes {
+        sendFrame(FrameType.PING, new ByteFlag(0x0), 0, ByteString("data1234")) // ping frame data must be of size 8
```
jrudolph: @ktoso in your original PR this was broken afaics because ping frame data needs to have size 8.

ktoso: Oh, thanks!

PR #788, opened 2017-01-17
WIP Remaining changes from former #764

jrudolph:
@ktoso let's use this new PR instead of #746 to review changes and to decide what to keep and what to solve otherwise.

### DIFF
--- a/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/FrameEvent.scala
+++ b/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/FrameEvent.scala
@@ -12,6 +12,7 @@ import akka.util.ByteString
 import scala.collection.immutable
 
 sealed trait FrameEvent
+sealed trait SyntheticFrameEvent extends FrameEvent
 sealed trait StreamFrameEvent extends FrameEvent {
   def streamId: Int
 }
@@ -78,3 +79,11 @@ final case class ParsedHeadersFrame(
   keyValuePairs: Seq[(String, String)],
   priorityInfo:  Option[PriorityFrame]
 ) extends FrameEvent
+
+// ------ internal frames, used for communicating settings between stages of our pipeline ----- 
+/**
+ * Special synthetic FrameEvent used to communicate enoder setting changes back to encoder.
+ * MUST NOT be rendered to network.
+ */
+private[http] final case class SyntheticHpackEncoderSettingFrame(setting: Setting) extends SyntheticFrameEvent
+private[http] final case class AcknowladgeSettings(originalFrame: SettingsFrame) extends SyntheticFrameEvent
--- a/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Compliance.scala
+++ b/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Compliance.scala
@@ -4,16 +4,38 @@
 
 package akka.http.impl.engine.http2
 
+import scala.annotation.switch
+
 /** INTERNAL API */
 private[akka] object Http2Compliance {
+  def parseSettungsEnablePushValue(incomingValue: Int): Boolean =
+    (incomingValue: @switch) match {
+      case 0 ⇒ false
+      case 1 ⇒ true
+      case _ ⇒ throw new IllegalEnablePushValueSettingException(s"Illegal value in SETTINGS_ENABLE_PUSH settings header! MUST BE 0 or 1, but was: [$incomingValue]")
+    }
+
+  private val MinOutMaxFrameSize = Math.pow(2, 14).toInt // minimum allowed to be set
+  private val MaxOutMaxFrameSize = (Math.pow(2, 24) - 1).toInt // maximum allowed to be set
+
+  final class WindowOverflowException(currentWindow: Int, increment: Int)
+    extends ArithmeticException(s"WINDOW overflow, must not exceed 2^31-1 nor go below 0. " +
+      s"Attempt to add increment ${increment} is illegal.")
 
   final class IllegalHttp2StreamIdException(id: Int, expected: String)
     extends IllegalArgumentException(s"Illegal HTTP/2 stream id: [$id]. $expected!")
 
   final class MissingHttpIdHeaderException extends IllegalArgumentException("Expected `Http2StreamIdHeader` header to be present but was missing!")
 
+  final class IllegalPushPromiseAttemptException
+    extends IllegalStateException("Server attempted to write PUSH_PROMISE to connection which had disabled it. See spec 6.5.2")
+
   final class IllegalHttp2StreamDependency(id: Int)
     extends IllegalArgumentException(s"Illegal self dependency of stream for id: [$id]!")
+
+  final class IllegalFrameSizeSettingException(msg: String) extends IllegalArgumentException(msg)
+
+  final class IllegalEnablePushValueSettingException(msg: String) extends IllegalArgumentException(msg)
 
   final class HeaderDecompressionFailed(msg: String) extends IllegalStateException(msg)
 
@@ -35,9 +57,35 @@ private[akka] object Http2Compliance {
   /** checks if the stream id was client initiated, by checking if the stream id was odd-numbered */
   final def isClientInitiatedStreamId(id: Int): Boolean = id % 2 != 0
 
+  final def requireFrameSize(size: Int, required: Int, hint: String = ""): Unit =
+    if (size != required) throw new IllegalHttp2FrameSize(size, s"MUST BE == $required. $hint")
+
   final def requireFrameSize(size: Int, max: Int): Unit =
     if (size != max) throw new IllegalHttp2FrameSize(size, s"MUST BE == $max.")
 
+  final def requireFrameSizeLessOrEqualThan(size: Int, max: Int, hint: String = ""): Unit =
+    if (size > max) throw new IllegalHttp2FrameSize(size, s"MUST BE <= $max. $hint")
+
   final def requireNoSelfDependency(id: Int, dependency: Int): Unit =
     if (id == dependency) throw new IllegalHttp2StreamDependency(id)
+
+  /** Validate value of MAX_FRAME_SIZE setting. */
+  def validateMaxFrameSize(value: Int): Unit = {
+    if (value < MinOutMaxFrameSize) throw new IllegalFrameSizeSettingException(s"MAX_FRAME_SIZE MUST NOT be < than $MinOutMaxFrameSize, attempted setting to: $value!")
+    if (value > MaxOutMaxFrameSize) throw new IllegalFrameSizeSettingException(s"MAX_FRAME_SIZE MUST NOT be > than $MaxOutMaxFrameSize, attempted setting to: $value!")
+  }
+
+  final case class WindowSizeWithOverflowProtection(value: Int) extends AnyVal {
+    def +(increment: Int): WindowSizeWithOverflowProtection = {
+      val res = value.toLong + increment
+      val resi = res.toInt
+      if (res != resi) throw new Http2Compliance.WindowOverflowException(value, increment)
+      WindowSizeWithOverflowProtection(resi)
+    }
+
+    def -(decrement: Int): WindowSizeWithOverflowProtection = {
+      this + (-decrement)
+    }
+  }
+
 }
--- a/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Protocol.scala
+++ b/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Protocol.scala
@@ -191,7 +191,9 @@ object Http2Protocol {
     def byId(id: Int): SettingIdentifier = All(id - 1)
   }
 
-  sealed abstract class ErrorCode(val id: Int) extends Product
+  sealed abstract class ErrorCode(val id: Int) extends Product {
+    def name = productPrefix
+  }
   object ErrorCode {
     /**
      * NO_ERROR (0x0):  The associated condition is not a result of an
--- a/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
+++ b/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
@@ -5,8 +5,9 @@
 package akka.http.impl.engine.http2
 
 import akka.NotUsed
+import akka.http.impl.engine.http2.Http2Compliance.WindowSizeWithOverflowProtection
 import akka.http.impl.engine.http2.Http2Protocol.ErrorCode
-import akka.http.impl.engine.http2.Http2Protocol.ErrorCode.{ COMPRESSION_ERROR, FRAME_SIZE_ERROR }
+import akka.http.impl.engine.http2.Http2Protocol.ErrorCode.{ COMPRESSION_ERROR, FRAME_SIZE_ERROR, PROTOCOL_ERROR }
 import akka.stream.Attributes
 import akka.stream.BidiShape
 import akka.stream.Inlet
@@ -88,7 +89,7 @@ class Http2ServerDemux extends GraphStage[BidiShape[Http2SubStream, FrameEvent, 
         state:                     StreamState,
         outlet:                    Option[BufferedOutlet[ByteString]],
         inlet:                     Option[SubSinkInlet[ByteString]],
-        initialOutboundWindowLeft: Long
+        initialOutboundWindowLeft: WindowSizeWithOverflowProtection
       ) {
         var outboundWindowLeft = initialOutboundWindowLeft
       }
@@ -103,8 +104,8 @@ class Http2ServerDemux extends GraphStage[BidiShape[Http2SubStream, FrameEvent, 
       // we should not handle streams later than the GOAWAY told us about with lastStreamId
       private var closedAfter: Option[Int] = None
       private var incomingStreams = mutable.Map.empty[Int, SubStream]
-      private var totalOutboundWindowLeft = Http2Protocol.InitialWindowSize
-      private var streamLevelWindow = Http2Protocol.InitialWindowSize
+      private var totalOutboundWindowLeft = WindowSizeWithOverflowProtection(Http2Protocol.InitialWindowSize)
+      private var streamLevelWindow = WindowSizeWithOverflowProtection(Http2Protocol.InitialWindowSize)
 
       /**
        * The "last peer-initiated stream that was or might be processed on the sending endpoint in this connection"
@@ -114,11 +115,11 @@ class Http2ServerDemux extends GraphStage[BidiShape[Http2SubStream, FrameEvent, 
         incomingStreams.keys.toList.sortBy(-_).headOption.getOrElse(0) // FIXME should be optimised
       }
 
-      def pushGOAWAY(errorCode: ErrorCode = Http2Protocol.ErrorCode.PROTOCOL_ERROR): Unit = {
+      def pushGOAWAY(errorCode: ErrorCode = Http2Protocol.ErrorCode.PROTOCOL_ERROR, debug: String = ""): Unit = {
         // http://httpwg.org/specs/rfc7540.html#rfc.section.6.8
         val last = lastStreamId
         closedAfter = Some(last)
-        val frame = GoAwayFrame(last, errorCode)
+        val frame = GoAwayFrame(last, errorCode, ByteString(debug))
         bufferedFrameOut.push(frame)
         // FIXME: handle the connection closing according to the specification
       }
@@ -130,7 +131,7 @@ class Http2ServerDemux extends GraphStage[BidiShape[Http2SubStream, FrameEvent, 
           in match {
             case WindowUpdateFrame(0, increment) ⇒
               totalOutboundWindowLeft += increment
-              debug(f"outbound window is now $totalOutboundWindowLeft%10d after increment $increment%6d")
+              debug(f"outbound window is now ${totalOutboundWindowLeft.value}%10d after increment $increment%6d")
               bufferedFrameOut.tryFlush()
 
             case e: StreamFrameEvent if !Http2Compliance.isClientInitiatedStreamId(e.streamId) ⇒
@@ -159,7 +160,7 @@ class Http2ServerDemux extends GraphStage[BidiShape[Http2SubStream, FrameEvent, 
 
             case e: StreamFrameEvent if !incomingStreams.contains(e.streamId) ⇒
               // if a stream is invalid we will GO_AWAY
-              pushGOAWAY()
+              pushGOAWAY(debug = s"Incoming streamId:${e.streamId} not present in incomingStreams:${incomingStreams}")
 
             case h: ParsedHeadersFrame ⇒
               pushGOAWAY()
@@ -179,22 +180,38 @@ class Http2ServerDemux extends GraphStage[BidiShape[Http2SubStream, FrameEvent, 
               incomingStreams(streamId).inlet.foreach(_.cancel())
 
             case WindowUpdateFrame(streamId, increment) ⇒
-              incomingStreams(streamId).outboundWindowLeft += increment
-              debug(f"outbound window for [$streamId%3d] is now ${incomingStreams(streamId).outboundWindowLeft}%10d after increment $increment%6d")
-              bufferedFrameOut.tryFlush()
+              try {
+                incomingStreams(streamId).outboundWindowLeft += increment
+
+                println(s"incomingStreams(streamId) = ${incomingStreams(streamId)} ( added + $increment ) ")
+                debug(f"outbound window for [$streamId%3d] is now ${incomingStreams(streamId).outboundWindowLeft.value}%10d after increment $increment%6d")
+                bufferedFrameOut.tryFlush()
+              } catch {
+                case ex: ArithmeticException ⇒
+                  pushGOAWAY(ErrorCode.FLOW_CONTROL_ERROR, debug = ex.getMessage)
+              }
 
             case PriorityFrame(streamId, exclusiveFlag, streamDependency, weight) ⇒
               debug(s"Received PriorityFrame for stream $streamId with ${if (exclusiveFlag) "exclusive " else "non-exclusive "} dependency on stream $streamDependency and weight $weight")
 
             case SettingsFrame(settings) ⇒
-              debug(s"Got ${settings.length} settings!")
+              if (settings.nonEmpty) debug(s"Got ${settings.length} settings!")
+
               settings.foreach {
                 case Setting(Http2Protocol.SettingIdentifier.SETTINGS_INITIAL_WINDOW_SIZE, value) ⇒
                   debug(s"Setting initial window to $value")
-                  val delta = value - streamLevelWindow
-                  streamLevelWindow = value
-                  incomingStreams.values.foreach(_.outboundWindowLeft += delta)
-                case Setting(id, value) ⇒ debug(s"Ignoring setting $id -> $value")
+                  val delta = WindowSizeWithOverflowProtection(value) - streamLevelWindow.value
+                  streamLevelWindow = WindowSizeWithOverflowProtection(value)
+                  incomingStreams.values.foreach(_.outboundWindowLeft += delta.value)
+
+                case Setting(Http2Protocol.SettingIdentifier.SETTINGS_MAX_FRAME_SIZE, value) ⇒
+                  debug(s"Already set client max frame size to ${value} in framing stage...")
+
+                case tableSizeSetting @ Setting(Http2Protocol.SettingIdentifier.SETTINGS_HEADER_TABLE_SIZE, _) ⇒
+                  debug("Bouncing header table size update back to encoder...")
+                  bufferedFrameOut.push(SyntheticHpackEncoderSettingFrame(tableSizeSetting))
+                case Setting(id, value) ⇒
+                  debug(s"Ignoring setting $id -> $value")
               }
 
               bufferedFrameOut.push(SettingsAckFrame)
@@ -204,8 +221,15 @@ class Http2ServerDemux extends GraphStage[BidiShape[Http2SubStream, FrameEvent, 
             case PingFrame(false, data) ⇒
               bufferedFrameOut.push(PingFrame(ack = true, data))
 
+            case AcknowladgeSettings(originalFrame) ⇒
+              log.info("Downstream applied setting {}, sending ACK back", originalFrame)
+              bufferedFrameOut.push(SettingsAckFrame) // TODO re-think these bouncing things... but since we applied in parser, we need a way to push ACK back
+
+            case SettingsAckFrame ⇒
+              log.info("Got Settings ACK...")
+
             case e ⇒
-              debug(s"Got unhandled event $e")
+              debug(s"Got unhandled event in Demux $e")
             // ignore unknown frames
           }
           pull(frameIn)
@@ -215,13 +239,19 @@ class Http2ServerDemux extends GraphStage[BidiShape[Http2SubStream, FrameEvent, 
           ex match {
             // every IllegalHttp2StreamIdException will be a GOAWAY with PROTOCOL_ERROR
             case e: Http2Compliance.IllegalHttp2StreamIdException ⇒
-              pushGOAWAY()
+              pushGOAWAY(debug = e.getMessage)
 
             case e: Http2Compliance.HeaderDecompressionFailed ⇒
-              pushGOAWAY(COMPRESSION_ERROR)
+              pushGOAWAY(COMPRESSION_ERROR, e.getMessage)
 
             case e: Http2Compliance.IllegalHttp2FrameSize ⇒
-              pushGOAWAY(FRAME_SIZE_ERROR)
+              pushGOAWAY(FRAME_SIZE_ERROR, e.getMessage)
+            case e: Http2Compliance.IllegalFrameSizeSettingException ⇒
+              pushGOAWAY(FRAME_SIZE_ERROR, e.getMessage)
+            case e: Http2Compliance.IllegalEnablePushValueSettingException ⇒
+              pushGOAWAY(PROTOCOL_ERROR, e.getMessage)
+            case e: Http2Compliance.IllegalPushPromiseAttemptException ⇒
+              pushGOAWAY(PROTOCOL_ERROR, e.getMessage)
 
             case e: ParsingException ⇒
               e.getCause match {
@@ -263,7 +293,7 @@ class Http2ServerDemux extends GraphStage[BidiShape[Http2SubStream, FrameEvent, 
         override def doPush(elem: ElementAndTrigger): Unit = {
           elem.element match {
             case d @ DataFrame(streamId, _, pl) ⇒
-              if (pl.size <= totalOutboundWindowLeft && pl.size <= incomingStreams(streamId).outboundWindowLeft) {
+              if (pl.size <= totalOutboundWindowLeft.value && pl.size <= incomingStreams(streamId).outboundWindowLeft.value) {
                 super.doPush(elem)
 
                 val size = pl.size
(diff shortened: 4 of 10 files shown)


akka-http2-support/src/main/scala/akka/http/impl/engine/http2/FrameEvent.scala
```diff
@@ -78,3 +79,11 @@ final case class ParsedHeadersFrame(
   keyValuePairs: Seq[(String, String)],
   priorityInfo:  Option[PriorityFrame]
 ) extends FrameEvent
+
+// ------ internal frames, used for communicating settings between stages of our pipeline ----- 
+/**
+ * Special synthetic FrameEvent used to communicate enoder setting changes back to encoder.
+ * MUST NOT be rendered to network.
+ */
+private[http] final case class SyntheticHpackEncoderSettingFrame(setting: Setting) extends SyntheticFrameEvent
```
jrudolph: Not sure if we need those if we put the settings into the SettingsAck frame.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Compliance.scala
```diff
@@ -4,17 +4,39 @@
...
 /** INTERNAL API */
 private[akka] object Http2Compliance {
+  def parseSettungsEnablePushValue(incomingValue: Int): Boolean =
+    (incomingValue: @switch) match {
+      case 0 ⇒ false
+      case 1 ⇒ true
+      case _ ⇒ throw new IllegalEnablePushValueSettingException(s"Illegal value in SETTINGS_ENABLE_PUSH settings header! MUST BE 0 or 1, but was: [$incomingValue]")
+    }
+
+  private val MinOutMaxFrameSize = Math.pow(2, 14).toInt // minimum allowed to be set
```
jrudolph: Maybe put these together with the other spec related constants into Http2Protocol.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Compliance.scala
```diff
@@ -4,17 +4,39 @@
...
+  private val MaxOutMaxFrameSize = (Math.pow(2, 24) - 1).toInt // maximum allowed to be set
+
+  final class WindowOverflowException(currentWindow: Int, increment: Int)
```
jrudolph: I wonder if we don't go a bit crazy with all those exception types. Do we really need them? What if we just have one or two exceptions, one for connection-level problems, and one for stream-level ones, that contain a message and an error code. Could still be created in here, WDYT?

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Compliance.scala
```diff
@@ -35,9 +57,35 @@ private[akka] object Http2Compliance {
...
     if (id == dependency) throw new IllegalHttp2StreamDependency(id)
+
+  /** Validate value of MAX_FRAME_SIZE setting. */
+  def validateMaxFrameSize(value: Int): Unit = {
+    if (value < MinOutMaxFrameSize) throw new IllegalFrameSizeSettingException(s"MAX_FRAME_SIZE MUST NOT be < than $MinOutMaxFrameSize, attempted setting to: $value!")
+    if (value > MaxOutMaxFrameSize) throw new IllegalFrameSizeSettingException(s"MAX_FRAME_SIZE MUST NOT be > than $MaxOutMaxFrameSize, attempted setting to: $value!")
+  }
+
+  final case class WindowSizeWithOverflowProtection(value: Int) extends AnyVal {
+    def +(increment: Int): WindowSizeWithOverflowProtection = {
```
jrudolph: Interesting approach ;)

Might be a bit overkill, if we manage to centralize all window change handling code at a single place, instead? But I'm not strongly against keeping this.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Compliance.scala
```diff
@@ -35,9 +57,35 @@ private[akka] object Http2Compliance {
...
+      val res = value.toLong + increment
+      val resi = res.toInt
+      if (res != resi) throw new Http2Compliance.WindowOverflowException(value, increment)
```
jrudolph: I think a simpler comparison would just be if the result has become negative. Then, it's either an overflow if the increment was positive or an underflow the increment was negative.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Protocol.scala
```diff
@@ -191,7 +191,9 @@ object Http2Protocol {
     def byId(id: Int): SettingIdentifier = All(id - 1)
   }
 
-  sealed abstract class ErrorCode(val id: Int) extends Product
+  sealed abstract class ErrorCode(val id: Int) extends Product {
+    def name = productPrefix
```
jrudolph: I guess this is used somewhere, but where?

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -159,7 +160,7 @@ class Http2ServerDemux extends GraphStage[BidiShape[Http2SubStream, FrameEvent,
 
             case e: StreamFrameEvent if !incomingStreams.contains(e.streamId) ⇒
               // if a stream is invalid we will GO_AWAY
-              pushGOAWAY()
+              pushGOAWAY(debug = s"Incoming streamId:${e.streamId} not present in incomingStreams:${incomingStreams}")
```
jrudolph: If we push the debug message with the GOAWAY frame, we should describe it in implementation-indepedent terms, so the text should be more like "Received illegal frame for stream $streamId in state ..."

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -179,22 +180,38 @@ class Http2ServerDemux extends GraphStage[BidiShape[Http2SubStream, FrameEvent,
...
-              debug(f"outbound window for [$streamId%3d] is now ${incomingStreams(streamId).outboundWindowLeft}%10d after increment $increment%6d")
-              bufferedFrameOut.tryFlush()
+              try {
+                incomingStreams(streamId).outboundWindowLeft += increment
+
+                println(s"incomingStreams(streamId) = ${incomingStreams(streamId)} ( added + $increment ) ")
+                debug(f"outbound window for [$streamId%3d] is now ${incomingStreams(streamId).outboundWindowLeft.value}%10d after increment $increment%6d")
+                bufferedFrameOut.tryFlush()
+              } catch {
+                case ex: ArithmeticException ⇒
```
jrudolph: Not sure if I like the exception handling here. Let's try to centralize window handling first (multiplexer PR goes already in that direction) before checking it.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -179,22 +180,38 @@ class Http2ServerDemux extends GraphStage[BidiShape[Http2SubStream, FrameEvent,
...
+                  debug(s"Already set client max frame size to ${value} in framing stage...")
+
+                case tableSizeSetting @ Setting(Http2Protocol.SettingIdentifier.SETTINGS_HEADER_TABLE_SIZE, _) ⇒
+                  debug("Bouncing header table size update back to encoder...")
+                  bufferedFrameOut.push(SyntheticHpackEncoderSettingFrame(tableSizeSetting))
+                case Setting(id, value) ⇒
+                  debug(s"Ignoring setting $id -> $value")
               }
 
               bufferedFrameOut.push(SettingsAckFrame)
```
jrudolph: Add the received settings here to the ack frame instead of the specialized bouncing frames.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -204,8 +221,15 @@ class Http2ServerDemux extends GraphStage[BidiShape[Http2SubStream, FrameEvent,
             case PingFrame(false, data) ⇒
               bufferedFrameOut.push(PingFrame(ack = true, data))
 
+            case AcknowladgeSettings(originalFrame) ⇒
```
jrudolph: See above.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -215,13 +239,19 @@ class Http2ServerDemux extends GraphStage[BidiShape[Http2SubStream, FrameEvent,
...
-              pushGOAWAY()
+              pushGOAWAY(debug = e.getMessage)
 
             case e: Http2Compliance.HeaderDecompressionFailed ⇒
-              pushGOAWAY(COMPRESSION_ERROR)
+              pushGOAWAY(COMPRESSION_ERROR, e.getMessage)
 
             case e: Http2Compliance.IllegalHttp2FrameSize ⇒
-              pushGOAWAY(FRAME_SIZE_ERROR)
+              pushGOAWAY(FRAME_SIZE_ERROR, e.getMessage)
```
jrudolph: This looks as if it would benefit from the approach discussed above with removing all those specialized exception types.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/framing/Http2FrameParsing.scala
```diff
@@ -162,5 +182,28 @@ private[http2] class Http2FrameParsing(shouldReadPreface: Boolean) extends ByteS
             UnknownFrameEvent(tpe, flags, streamId, payload.remainingData)
         }
       }
+
+      // FIXME this should be smarter, we need to somehow manage when we send back the ACK basically...
+      private def applySetting(s: SettingsFrame): SettingsFrame = {
+        s.settings.foreach {
+          case Setting(SettingIdentifier.SETTINGS_MAX_FRAME_SIZE, value) ⇒
+            settings.updateMaxOutFrameSize(value)
```
jrudolph: I think this logic might actually be wrong. As discussed in https://github.com/akka/akka-http/pull/764#issuecomment-272897513 a received SETTINGS_MAX_FRAME_SIZE doesn't say something about the frames that our peer sends (and we receive here) but about the frames which our peer receives:

```
  SETTINGS_MAX_FRAME_SIZE (0x5):  Indicates the size of the largest
      frame payload that the sender is willing to receive, in octets.
```

so, as long as we ourselves don't change this setting, we cannot act on it here (and if, we would have to act on the settings ACK frame instead).

It's actually the same for `SETTINGS_ENABLE_PUSH`, so we might not need any handling of settings frames in the parser right now (apart from checking the limits against the default values.)

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/framing/Http2FrameParsing.scala
```diff
@@ -7,21 +7,43 @@ package framing
...
   import ByteStringParser._
 
   abstract class Step extends ParseStep[FrameEvent]
 
+  override def initialAttributes = Attributes.name(Logging.simpleName(getClass))
+
   def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
     new ParsingLogic {
+
+      private var _outMaxFrameSize: Int = 16384 // default
```
jrudolph: Protocol constants please into Http2Protocol.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/framing/Http2FrameParsing.scala
```diff
@@ -138,13 +160,11 @@ private[http2] class Http2FrameParsing(shouldReadPreface: Boolean) extends ByteS
             RstStreamFrame(streamId, ErrorCode.byId(payload.readIntBE()))
 
           case PRIORITY ⇒
-            Http2Compliance.requireFrameSize(payload.remainingSize, 5)
```
jrudolph: Why is that gone?

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/framing/Http2FrameParsing.scala
```diff
@@ -138,13 +160,11 @@ private[http2] class Http2FrameParsing(shouldReadPreface: Boolean) extends ByteS
...
             val streamDependency = payload.readIntBE() // whole word
             val exclusiveFlag = (streamDependency >>> 31) == 1 // most significant bit for exclusive flag
-            val dependencyId = streamDependency & 0x7fffffff // remaining 31 bits for the dependency part
+            val dependencyPart = streamDependency & 0x7fffffff // remaining 31 bits for the dependency part
             val priority = payload.readByte() & 0xff
-            Http2Compliance.requireNoSelfDependency(streamId, dependencyId)
```
jrudolph: Why is that gone?

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/framing/Http2FrameParsing.scala
jrudolph: I realize that this might be an artifact of the previous solution which would have communicated the peer intention with the renderer.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/framing/Http2FrameRendering.scala
```diff
@@ -0,0 +1,85 @@
...
+  override def initialAttributes = Attributes.name(Logging.simpleName(getClass))
+
+  override val shape = FlowShape[FrameEvent, ByteString](frameIn, netOut)
+
+  override def createLogic(inheritedAttributes: Attributes) = new GraphStageLogic(shape) with InHandler with OutHandler {
+    setHandlers(frameIn, netOut, this)
+
+    // TODO remove
+    object settings {
+      private var _outMaxFrameSize: Int = 16384 // default
```
jrudolph: Here we need these settings but they need to be changed when we send out the SettingAck. I think that might now have been lost somewhere?

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/framing/Http2FrameRendering.scala
```diff
@@ -0,0 +1,85 @@
...
+import akka.http.impl.engine.http2.Http2Protocol.SettingIdentifier
+import akka.http.impl.engine.http2._
+import akka.stream.{ Attributes, FlowShape, Inlet, Outlet }
+import akka.stream.stage.{ GraphStage, GraphStageLogic, InHandler, OutHandler }
+import akka.util.ByteString
+
+import scala.collection.immutable
+
+/** INTERNAL API */
+private[http] class Http2FrameRendering extends GraphStage[FlowShape[FrameEvent, ByteString]] {
```
jrudolph: I had the idea today that we could implement this whole stage using `statefulMapConcat`? Maybe let's try this.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/framing/Http2FrameRendering.scala
```diff
@@ -0,0 +1,85 @@
...
+    override def onPush(): Unit = {
+      val frame = grab(frameIn)
+
+      frame match {
+        case d @ DataFrame(_, _, payload) if payload.size > settings.maxOutFrameSize ⇒
+          // payload too large, we must split it into multiple frames:
+          val splitDataDrames = splitByPayloadSize(d, settings.maxOutFrameSize).iterator.map(FrameRenderer.render)
+          emitMultiple(netOut, splitDataDrames) // TODO would manual handling be more efficient? 
+
+        case p: PushPromiseFrame if settings.pushPromiseDisabled ⇒
```
jrudolph: We could check it here, but it would be an error from our own code. Fine for me though.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/hpack/HeaderCompression.scala
```diff
@@ -5,39 +5,45 @@
...
-import akka.stream.stage.{ GraphStage, GraphStageLogic, OutHandler }
+import akka.stream.stage.{ GraphStage, GraphStageLogic, OutHandler, StageLogging }
 import akka.util.ByteString
 
 /**
  * INTERNAL API
  */
 private[http2] object HeaderCompression extends GraphStage[FlowShape[FrameEvent, FrameEvent]] {
+
+  final val InitialMaxHeaderTableSize = 4096 // according to spec 6.5.2
```
jrudolph: Move to Http2Protocol

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/hpack/HeaderDecompression.scala
```diff
@@ -55,12 +54,24 @@ private[http2] object HeaderDecompression extends GraphStage[FlowShape[FrameEven
...
-          fail(eventsOut, new Http2Compliance.HeaderDecompressionFailed("Decompression failed."))
+          fail(eventsOut, new Http2Compliance.HeaderDecompressionFailed("Header decompression failed."))
       }
     }
 
+    // TODO optimise this
+    def handleSettings: PartialFunction[FrameEvent, Unit] = {
+      case frame @ SettingsFrame(settings) ⇒
+        settings.collect {
+          case Setting(SettingIdentifier.SETTINGS_HEADER_TABLE_SIZE, size) ⇒
```
jrudolph: That's again the wrong way round. We need to call `decoder.setMaxHeaderTableSize` when we receive the ack for this setting we sent ourselves (which we don't do currently).

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/hpack/HeaderCompression.scala
```diff
@@ -55,6 +61,15 @@ private[http2] object HeaderCompression extends GraphStage[FlowShape[FrameEvent,
               }
             })
           }
+        case e: SyntheticFrameEvent ⇒
+          e match {
+            case SyntheticHpackEncoderSettingFrame(Setting(SETTINGS_HEADER_TABLE_SIZE, value)) ⇒
+              log.warning("Set outgoing compression table size to {}", value)
+              encoder.setMaxHeaderTableSize(os, value)
+              push(eventsOut, SettingsAckFrame)
```
jrudolph: This wouldn't work if the setting came in together with another setting and both of them need to be ack'd together. => Let's just put the settings into the SettingsAck frame ;)

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/FrameEvent.scala
ktoso: Yes you're right - was in a rush pushing out the passing version today. Removing it

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Compliance.scala
ktoso: good idea

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Protocol.scala
ktoso: I used it in some point during testing to get better error messages. I see it's not used anymore, removing

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Compliance.scala
ktoso: I was somewhat thinking we could expand and collapse them, in the sense of look how exceptions need to be reacted on and then collapse them together into a minimum viable set or maybe some common super types.

For example some errors need to be rendered as PROTOCOL_ERRORS, some as FRAME_SIZE_ERROR and some as others... Perhaps we'll simply carry the error code to GOAWAY with in the exception, that would solve it as well :)

I can give it a shot at collapsing using that idea

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Compliance.scala
ktoso: Yeah, sounds good - if we centralize it let's not use this style

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Compliance.scala
ktoso: I actually skimmed to check what Guava does, perhaps falsely assuming it's somewhat optimal but perhaps not hm. Will move it to < checks

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
ktoso: Yes good point

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
ktoso: Right, good idea to move towards centralizing those. Shall I remove the handling in this PR or how to proceed here?

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
ktoso: Yeap 👍

jrudolph:
Btw. try looking at the output of

```
git diff a3214df..80aa92cab8d4de909e2d91a129bce29b96908366 -- akka-http2-support
```

for a diff between your original PR and this one.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -472,12 +475,26 @@ class Http2ServerSpec extends AkkaSpec with WithInPendingUntilFixed with Eventua
...
+      "multiple WINDOW_UPDATEs exceeding total demand above 2^31-1 MUST cause FLOW_CONTROL_ERROR (spec 6.9.1)" in new TestSetup with RequestResponseProbes {
+        override def safeUpdate(update: Int ⇒ Int): Int ⇒ Int = { oldValue ⇒ update(oldValue) }
+
+        sendHEADERS(1, endStream = false, endHeaders = true, HPackSpecExamples.C41FirstRequestWithHuffman)
+
+        sendWINDOW_UPDATE(1, Math.pow(2, 30).toInt)
+        sendWINDOW_UPDATE(1, Math.pow(2, 30).toInt)
+        sendWINDOW_UPDATE(1, Math.pow(2, 30).toInt)
+
+        val (lastStreamId, errorCode) = expectGOAWAY()
```
jrudolph: Shouldn't this only reset the stream?

From 6.9.1:

```
   A sender MUST NOT allow a flow-control window to exceed 2^31-1
   octets.  If a sender receives a WINDOW_UPDATE that causes a flow-
   control window to exceed this maximum, it MUST terminate either the
   stream or the connection, as appropriate. For streams, the sender
   sends a RST_STREAM with an error code of FLOW_CONTROL_ERROR
```

jrudolph: Also add a test for the connection-level one (which should then indeed be a GOAWAY).

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -499,16 +516,82 @@ class Http2ServerSpec extends AkkaSpec with WithInPendingUntilFixed with Eventua
         error should ===(ErrorCode.FRAME_SIZE_ERROR)
       }
 
-      "received SETTINGS_MAX_FRAME_SIZE" in pending
+      "received SETTINGS_MAX_FRAME_SIZE should cause outgoing DATA to be chunked up into at-most-that-size parts " in new TestSetup with RequestResponseProbes {
+        def setMaxFrameSize(size: Int) = SettingsFrame(Setting(SettingIdentifier.SETTINGS_MAX_FRAME_SIZE, size) :: Nil)
+
+        private val maxSize = Math.pow(2, 15).toInt // 32768, valid value (between 2^14 and 2^14 - 1)
+        sendFrame(setMaxFrameSize(maxSize))
```
jrudolph: Use the new `sendSetting` you introduced.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -499,16 +516,82 @@ class Http2ServerSpec extends AkkaSpec with WithInPendingUntilFixed with Eventua
...
+
+        expectSettingsAck()
+
+        sendWINDOW_UPDATE(0, maxSize * 3) // make sure we can receive such large response, on connection
+
+        sendHEADERS(1, endStream = false, endHeaders = true, HPackSpecExamples.C41FirstRequestWithHuffman)
+        sendWINDOW_UPDATE(1, maxSize * 3) // make sure we can receive such large response, on this stream
```
jrudolph: HEADERS don't count towards the window, so I guess you can remove this one.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -499,16 +516,82 @@ class Http2ServerSpec extends AkkaSpec with WithInPendingUntilFixed with Eventua
...
+        def setMaxFrameSize(size: Int) = SettingsFrame(Setting(SettingIdentifier.SETTINGS_MAX_FRAME_SIZE, size) :: Nil)
+
+        private val maxSize = Math.pow(2, 15).toInt // 32768, valid value (between 2^14 and 2^14 - 1)
+        sendFrame(setMaxFrameSize(maxSize))
+
+        expectSettingsAck()
+
+        sendWINDOW_UPDATE(0, maxSize * 3) // make sure we can receive such large response, on connection
+
+        sendHEADERS(1, endStream = false, endHeaders = true, HPackSpecExamples.C41FirstRequestWithHuffman)
```
jrudolph: `endStream` should be true, as there's no request body.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -499,16 +516,82 @@ class Http2ServerSpec extends AkkaSpec with WithInPendingUntilFixed with Eventua
...
+        private val theTooLargeByteString = ByteString("x" * (maxSize * 2))
+        val tooLargeEntity = Source.single(theTooLargeByteString)
+        val tooLargeResponse = HttpResponse(entity = HttpEntity(ContentTypes.`application/octet-stream`, tooLargeEntity))
+        emitResponse(1, tooLargeResponse)
+
+        expectHeaderBlock(1, endStream = false)
+        sendWINDOW_UPDATE(1, maxSize * 3) // make sure we can receive such large response, on this stream
+        // FIXME we should not need to send this window update, the stream already has enough window I think, see above ^^^
+        // we receive the DATA in 2 parts, since the ByteString does not fit in a single frame
+        val d1 = expectDATA(1, endStream = false, numBytes = maxSize)
```
jrudolph: This will work for now but might be overly restrictive: the implementation would be allowed to send any number of bytes not just exactly the max frame size.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -499,16 +516,82 @@ class Http2ServerSpec extends AkkaSpec with WithInPendingUntilFixed with Eventua
...
+      "receive SETTINGS_MAX_FRAME_SIZE with too small value (spec 4.2)" in new TestSetup with RequestResponseProbes {
+        def setMaxFrameSize(size: Int) = SettingsFrame(Setting(SettingIdentifier.SETTINGS_MAX_FRAME_SIZE, size) :: Nil)
 
-      "received SETTINGS_MAX_CONCURRENT_STREAMS" in pending
+        /*
+          An endpoint MUST send an error code of FRAME_SIZE_ERROR if a frame exceeds the size defined in SETTINGS_MAX_FRAME_SIZE, 
+          exceeds any limit defined for the frame type, or is too small to contain mandatory frame data. 
+         */
+
+        sendFrame(setMaxFrameSize(8))
```
jrudolph: `sendSetting`

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -499,16 +516,82 @@ class Http2ServerSpec extends AkkaSpec with WithInPendingUntilFixed with Eventua
...
+        private val maxSize = Math.pow(2, 15).toInt // 32768, valid value (between 2^14 and 2^14 - 1)
+        sendFrame(setHeaderTableSize(maxSize))
+
+        expectSettingsAck()
+      }
+
+      "received disable PUSH_PROMISE by received SETTINGS_ENABLE_PUSH (invalid 6.5)" in new TestSetup with RequestResponseProbes {
+        def setAllowPushPromise(value: Boolean) = SettingsFrame(Setting(SettingIdentifier.SETTINGS_ENABLE_PUSH, if (value) 1 else 0) :: Nil)
+
+        sendFrame(setAllowPushPromise(false))
```
jrudolph: `sendSetting`

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -499,16 +516,82 @@ class Http2ServerSpec extends AkkaSpec with WithInPendingUntilFixed with Eventua
...
+
+        expectSettingsAck()
+
+        // FIXME should send a PUSH_PROMISE from server side now, it should fail since it's not allowed
```
jrudolph: Actually, it shouldn't fail but the server push should just be ignored and maybe issue a warning on a low logging level. Maybe let's keep this test pending until it can check anything.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -533,8 +616,13 @@ class Http2ServerSpec extends AkkaSpec with WithInPendingUntilFixed with Eventua
         val (lastStreamId, errorCode) = expectGOAWAY()
         errorCode should ===(ErrorCode.PROTOCOL_ERROR)
       }
+      "acknowledge SETTINGS frames" in new TestSetup with RequestResponseProbes {
+        def setMaxFrameSize(size: Int) = SettingsFrame(Setting(SettingIdentifier.SETTINGS_MAX_FRAME_SIZE, size) :: Nil)
```
jrudolph: remove

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -796,7 +884,7 @@ class Http2ServerSpec extends AkkaSpec with WithInPendingUntilFixed with Eventua
       decodeHeaders(headerBlockBytes)
     }
 
-    val encoder = new Encoder(HeaderDecompression.maxHeaderTableSize)
+    val encoder = new Encoder(HeaderCompression.InitialMaxHeaderTableSize)
```
jrudolph: Should be from `Http2Protocol`.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
ktoso: yup, applied everywhere now

jrudolph:
Hmm, so we have now these features in this diff:
 * respect MAX_FRAME_SIZE from peer
 * respect MAX_HEADER_TABLE_SIZE from peer (would be nice to have tests for that but that's not so easy as one would have to fill the encoder table first and then verify that after the setting was applied, the high-index table entries wouldn't be used any more, maybe something for later...)
 * check that WINDOW_UPDATES don't overflow the Int
 * simplify exception tree (not yet in this PR but a learning from this one)

I think it would be easiest to take these features one by one, take the tests from this PR and then try to build a new PR for each one salvaging only the things really needed to make them pass. WDYT?

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/framing/Http2FrameParsing.scala
```diff
@@ -7,21 +7,43 @@ package framing
 
 import scala.collection.immutable
 
-import akka.stream.Attributes
```
jrudolph: To be clear, I don't think we need any changes in this file to make any of the tests pass.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/framing/Http2FrameParsing.scala
ktoso: I'd guess messed up diffs, not on purpose

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/framing/Http2FrameRendering.scala
```diff
@@ -0,0 +1,85 @@
...
+      }
+    }
+
+    override def onPull(): Unit = pull(frameIn)
+
+    private def splitByPayloadSize(d: DataFrame, size: Int): immutable.Seq[DataFrame] = {
+      println(s"Splitting up too large data-frame into smaller frames due to exceeding max frame size. Length: ${d.payload.length}, max: ${size}")
+      val parts = d.payload.grouped(size) // TODO optimise, splitAt would be better
+
+      if (d.endStream) {
```
jrudolph: For now you could alternatively require that the last frame (endStream = true) is empty because that's an invariant of the current infrastructure.

ktoso:
Killing this PR and re implementing (picking whats possible from here) to get cleaner impls of the solutions from here. Seems there's a lot of assumptions from previous way of doing things in the salvaged code

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/framing/Http2FrameRendering.scala
ktoso: yeah i can mark it as "this is a bug in akka", let's keep it while we develop actively

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/framing/Http2FrameRendering.scala
ktoso: I wonder if it wouldn't introduce overhead for the simple 1 frame case though?

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/hpack/HeaderDecompression.scala
ktoso: Hah right, thanks.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/framing/Http2FrameParsing.scala
ktoso: yup

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/framing/Http2FrameParsing.scala
ktoso: Related question: Not allowing push we'd handle in demux already?

jrudolph: Some place in our code will have to generate the PUSH_PROMISE frame so we would handle it there.

PR #792, opened 2017-01-18
+ht2 #757 handle incoming SETTINGS_MAX_FRAME_SIZE

ktoso:
Handles https://github.com/akka/akka-http/issues/757


akka-http2-support/src/main/scala/akka/http/impl/engine/http2/framing/Http2FrameRendering.scala
```diff
@@ -26,10 +26,25 @@ private[http] class Http2FrameRendering extends GraphStage[FlowShape[FrameEvent,
   override def createLogic(inheritedAttributes: Attributes) = new GraphStageLogic(shape) with InHandler with OutHandler {
     setHandlers(frameIn, netOut, this)
 
+    object settings {
```
jrudolph: Can we get rid of this extra object?

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/framing/Http2FrameRendering.scala
```diff
@@ -26,10 +26,25 @@ private[http] class Http2FrameRendering extends GraphStage[FlowShape[FrameEvent,
...
+      }
+    }
+
     override def onPush(): Unit = {
       val frame = grab(frameIn)
-
       frame match {
+        case d @ DataFrame(_, _, payload) if payload.size > settings.maxOutFrameSize ⇒
+          // payload too large, we must split it into multiple frames:
+          val splitDataDrames = splitByPayloadSize(d, settings.maxOutFrameSize).iterator.map(FrameRenderer.render)
```
jrudolph: `splitDataFrames`

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/framing/Http2FrameRendering.scala
```diff
@@ -26,10 +26,25 @@ private[http] class Http2FrameRendering extends GraphStage[FlowShape[FrameEvent,
...
+          emitMultiple(netOut, splitDataDrames) // TODO would manual handling be more efficient?
```
jrudolph: Fine with me for now performance-wise.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/framing/Http2FrameRendering.scala
```diff
@@ -44,11 +59,30 @@ private[http] class Http2FrameRendering extends GraphStage[FlowShape[FrameEvent,
 
     override def onPull(): Unit = pull(frameIn)
 
+    private def splitByPayloadSize(d: DataFrame, size: Int): immutable.Seq[DataFrame] = {
+      debug(s"Splitting up too large data-frame into smaller frames due to exceeding max frame size. Length: ${d.payload.length}, max: ${size}")
+      val parts = d.payload.grouped(size) // TODO optimise, splitAt would be better
+
+      if (d.endStream) {
```
jrudolph: Let's add a require here that all frames with d.endStream must be empty which is an invariant of the current impl.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -499,7 +500,32 @@ class Http2ServerSpec extends AkkaSpec with WithInPendingUntilFixed with Eventua
...
+        sendWINDOW_UPDATE(1, maxSize * 3) // make sure we can receive such large response, on this stream
+
+        val theTooLargeByteString = ByteString("x" * (maxSize * 2))
+        val tooLargeEntity = Source.single(theTooLargeByteString)
+        val tooLargeResponse = HttpResponse(entity = HttpEntity(ContentTypes.`application/octet-stream`, tooLargeEntity))
+        emitResponse(1, tooLargeResponse)
+
+        expectHeaderBlock(1, endStream = false)
+        sendWINDOW_UPDATE(1, maxSize * 3) // make sure we can receive such large response, on this stream
+        // FIXME we should not need to send this window update, the stream already has enough window I think, see above ^^^
```
jrudolph: Yes, please fix that here. I think we would only need two sendWINDOW_UPDATE, one for the connection and one for the stream.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -499,7 +500,32 @@ class Http2ServerSpec extends AkkaSpec with WithInPendingUntilFixed with Eventua
         error should ===(ErrorCode.FRAME_SIZE_ERROR)
       }
 
-      "received SETTINGS_MAX_FRAME_SIZE" in pending
+      "received SETTINGS_MAX_FRAME_SIZE should cause outgoing DATA to be chunked up into at-most-that-size parts " in new TestSetup with RequestResponseProbes {
+        private val maxSize = Math.pow(2, 15).toInt // 32768, valid value (between 2^14 and 2^14 - 1)
```
schmitch: the comment has a typo, since it's between 2^14 and 2^24 -1 (http://httpwg.org/specs/rfc7540.html#rfc.section.4.2)

ktoso: thanks

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/framing/Http2FrameRendering.scala
ktoso: Yes, absolutely.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/framing/Http2FrameRendering.scala
ktoso: thanks

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/framing/Http2FrameRendering.scala
ktoso: yeah probably ok, will remove todo

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/framing/Http2FrameRendering.scala
ktoso: Ok, will add; though generally speaking would we not want the last frame to carry the flag?

jrudolph: The point is that this should never be reached right now because we all `DATA` frames we generate with `endStream = true` are empty so it shouldn't reach this method anyway. So, we don't actually have to handle this case right now.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
ktoso: Working on this now

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/framing/Http2FrameRendering.scala
ktoso: Yeah right *now* sure, but it's not how we want to keep things forever I think, so I should mark it as such with a todo or so though

ktoso:
Fixed the flow control, I'll comment in line where the issue was

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -84,14 +85,12 @@ class Http2ServerDemux extends GraphStage[BidiShape[Http2SubStream, FrameEvent,
       logic ⇒
 
       final case class SubStream(
-        streamId:                  Int,
-        state:                     StreamState,
-        outlet:                    Option[BufferedOutlet[ByteString]],
-        inlet:                     Option[SubSinkInlet[ByteString]],
-        initialOutboundWindowLeft: Long
-      ) {
-        var outboundWindowLeft = initialOutboundWindowLeft
```
ktoso: so... copy() initialized this with the initial outbound value again :O whereas we actually wanted to keep the value we accumulated here, since the WINDOW_UPDATEs were writing into here.

ktoso: (this is what caused the weird need to send more WINDOW_UPDATE than we would have expected in the entity test below)

jrudolph: Aha, good catch. So, all WINDOW_UPDATEs were lost if they were received before the response came in from the user (because that's when `copy` is used).

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -499,32 +500,55 @@ class Http2ServerSpec extends AkkaSpec with WithInPendingUntilFixed with Eventua
...
       "received SETTINGS_HEADER_TABLE_SIZE" in new TestSetup with RequestResponseProbes {
-        sendSETTING(SettingIdentifier.SETTINGS_HEADER_TABLE_SIZE, Math.pow(2, 15).toInt) // 32768, valid value (between 2^14 and 2^14 - 1)
+        sendSETTING(SettingIdentifier.SETTINGS_HEADER_TABLE_SIZE, Math.pow(2, 15).toInt) // 32768, valid value (between 2^14 and 2^24 - 1)
 
         expectSettingsAck() // TODO check that the setting was indeed applied
       }
     }
 
     "support low-level features" should {
-      "respond to PING frames (spec 6_7)" in new TestSetup with RequestResponseProbes {
```
jrudolph: Btw. this was on purpose because the jenkins JUnit report file parser otherwise does stupid things and the test tree doesn't work properly in the UI. Let's keep it with the dot and think about how to fix it otherwise.

jrudolph:
LGTM.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/framing/Http2FrameRendering.scala
ktoso: Ok, agreed then 👍

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
ktoso: Yup, exactly.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
ktoso: Oh I see, will revert and use _ in my tests too then

PR #794, opened 2017-01-18
+ht2 validate that WINDOW_UPDATE must be > 0

ktoso:
Validating window value, small little addition that I added while I was looking at flow control handling.

### DIFF
--- a/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Compliance.scala
+++ b/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Compliance.scala
@@ -38,9 +38,15 @@ private[akka] object Http2Compliance {
   }
 
   class Http2ProtocolException(val errorCode: ErrorCode, message: String) extends IllegalStateException(message)
+  class Http2ProtocolStreamException(val streamId: Int, val errorCode: ErrorCode, message: String) extends IllegalStateException(message)
 
   final def requireZeroStreamId(id: Int): Unit =
     if (id != 0) throw new IllegalHttp2StreamIdException(id, "MUST BE == 0.")
+
+  final def requirePositiveWindowUpdateIncrement(streamId: Int, increment: Int): Unit =
+    if (increment <= 0)
+      if (streamId == 0) throw new Http2ProtocolException(ErrorCode.PROTOCOL_ERROR, "WINDOW_UPDATE MUST be > 0, was: " + increment) // cause GOAWAY
+      else throw new Http2ProtocolStreamException(streamId, ErrorCode.PROTOCOL_ERROR, "WINDOW_UPDATE MUST be > 0, was: " + increment) // cause RST_STREAM
 
   /** checks if the stream id was client initiated, by checking if the stream id was odd-numbered */
   final def isClientInitiatedStreamId(id: Int): Boolean = id % 2 != 0
--- a/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/framing/Http2FrameParsing.scala
+++ b/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/framing/Http2FrameParsing.scala
@@ -116,9 +116,9 @@ private[http2] class Http2FrameParsing(shouldReadPreface: Boolean) extends ByteS
           case WINDOW_UPDATE ⇒
             // TODO: check frame size
             // TODO: check flags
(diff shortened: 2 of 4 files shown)


akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -472,6 +472,22 @@ class Http2ServerSpec extends AkkaSpec with WithInPendingUntilFixed with Eventua
...
+        errorCode should ===(ErrorCode.PROTOCOL_ERROR)
+      }
+      "reject WINDOW_UPDATE for stream with zero increment with PROTOCOL_ERROR" in new TestSetup with RequestResponseProbes {
+        // making sure we don't handle stream 0 and others differently here
+        sendHEADERS(1, endStream = false, endHeaders = true, HPackSpecExamples.C41FirstRequestWithHuffman)
+        sendWINDOW_UPDATE(1, 0) // illegal
+        val (_, errorCode) = expectGOAWAY()
+
+        errorCode should ===(ErrorCode.PROTOCOL_ERROR)
+      }
```
ktoso: for sanity / future resilience checking both cases, even though we handle it in one spot.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/framing/Http2FrameParsing.scala
```diff
@@ -117,8 +117,9 @@ private[http2] class Http2FrameParsing(shouldReadPreface: Boolean) extends ByteS
             // TODO: check frame size
             // TODO: check flags
             // TODO: check reserved flag
```
jrudolph: You can remove this as well. The "reserved flag" is actually the sign when we read a complete `readIntBE`. So, technically the error message should then also be "Reserved flag in WINDOW_UPDATE" must be 0.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -472,6 +472,22 @@ class Http2ServerSpec extends AkkaSpec with WithInPendingUntilFixed with Eventua
...
+        sendWINDOW_UPDATE(0, 0) // illegal
+        val (_, errorCode) = expectGOAWAY()
+
+        errorCode should ===(ErrorCode.PROTOCOL_ERROR)
+      }
+      "reject WINDOW_UPDATE for stream with zero increment with PROTOCOL_ERROR" in new TestSetup with RequestResponseProbes {
+        // making sure we don't handle stream 0 and others differently here
+        sendHEADERS(1, endStream = false, endHeaders = true, HPackSpecExamples.C41FirstRequestWithHuffman)
+        sendWINDOW_UPDATE(1, 0) // illegal
+        val (_, errorCode) = expectGOAWAY()
```
jrudolph: Should be RST_STREAM here. No need to kill the connection when only one substream is affected.

6.9:

```
   A receiver MUST treat the receipt of a WINDOW_UPDATE frame with an
   flow-control window increment of 0 as a stream error (Section 5.4.2)
   of type PROTOCOL_ERROR; errors on the connection flow-control window
   MUST be treated as a connection error (Section 5.4.1).
```

ktoso: Oh, good catch thanks.

ktoso: Done, here we get a RST_STREAM now, handled rather nicely IMO with new super type for exceptions on stream level, which carry streamid

ktoso:
Should be ready, please review

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Compliance.scala
```diff
@@ -38,10 +38,16 @@ private[akka] object Http2Compliance {
   }
 
   class Http2ProtocolException(val errorCode: ErrorCode, message: String) extends IllegalStateException(message)
+  class Http2ProtocolStreamException(val streamId: Int, val errorCode: ErrorCode, message: String) extends IllegalStateException(message)
```
jrudolph: LGTM

PR #795, opened 2017-01-19
Fixes #784 exclude leading double slash from Uri.Authority.toString

### LINKED ISSUE #784: Uri.Authority.toString renders leading double slash
While the overall rendering of URIs is fine, `Uri.Authority.toString` will render leading double slashes (which will also fail the round-trip `Uri.Authority.parse(authority.toString)`):

```
scala> Uri("http://localhost/test").authority
res1: akka.http.scaladsl.model.Uri.Authority = //localhost

scala> Uri.Authority.parse(res1.toString)
akka.http.scaladsl.model.IllegalUriException: Illegal authority: Invalid input '/', expected authority or 'EOI' (line 1, column 1): //localhost
^
  at akka.http.scaladsl.model.IllegalUriException$.apply(ErrorInfo.scala:40)
  at akka.http.scaladsl.model.Uri$.fail(Uri.scala:750)
  at akka.http.impl.model.parser.UriParser.fail(UriParser.scala:77)
  at akka.http.impl.model.parser.UriParser.parseAuthority(UriParser.scala:72)
  at akka.http.scaladsl.model.Uri$Authority$.parse(Uri.scala:331)
  ... 42 elided
```

This doesn't seem to be covered by the spec.

 https://tools.ietf.org/html/rfc3986#section-3:

```
   The generic URI syntax consists of a hierarchical sequence of
   components referred to as the scheme, authority, path, query, and
   fragment.

      URI         = scheme ":" hier-part [ "?" query ] [ "#" fragment ]

      hier-part   = "//" authority path-abempty / [...]

  [...]

   The following are two example URIs and their component parts:

         foo://example.com:8042/over/there?name=ferret#nose
         \_/   \______________/\_________/ \_________/ \__/
          |           |            |            |        |
       scheme     authority       path        query   fragment
          |   _____________________|__
         / \ /                        \
         urn:example:animal:ferret:nose
```

So, the double slashes belong to the `hier-part` and not to the authority.

In URI we could move the rendering of the double slashes from `renderAuthority` (which is used in `Authority.toString`) to `renderUriWithoutFragment`. Hopefully making this change won't lead to much breakage.

ulasakdeniz:
Applied @jrudolph's comment in #784

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/Uri.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/Uri.scala
@@ -803,6 +803,7 @@ object UriRendering {
   def renderUriWithoutFragment[R <: Rendering](r: R, value: Uri, charset: Charset): r.type = {
     import value._
     if (isAbsolute) r ~~ scheme ~~ ':'
+    if (authority.nonEmpty) r ~~ '/' ~~ '/'
     renderAuthority(r, authority, path, scheme, charset)
     renderPath(r, path, charset, encodeFirstSegmentColons = isRelative)
     rawQueryString.foreach(r ~~ '?' ~~ _)
@@ -815,7 +816,6 @@ object UriRendering {
   def renderAuthority[R <: Rendering](r: R, authority: Authority, path: Path, scheme: String, charset: Charset): r.type =
     if (authority.nonEmpty) {
       import authority._
-      r ~~ '/' ~~ '/'
       if (!userinfo.isEmpty) encode(r, userinfo, charset, `userinfo-char`) ~~ '@'
       r ~~ host
       if (port != 0) r ~~ ':' ~~ port else r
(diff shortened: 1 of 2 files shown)


jonas:
OK TO TEST

jrudolph:
Thanks, @ulasakdeniz for getting to this so promptly. The fix is fine. We should probably have a quick discussion about the compatibility consequences. As the previous form of rendering has been in place for a long time ([dating back to spray](https://github.com/spray/spray/blob/master/spray-http/src/main/scala/spray/http/Uri.scala#L348)) some people might have taking precautions against it (e.g. doing a `drop(2)` on the result of `toString`.

I'm slightly in favor of fixing it even in the fix release as it is strictly speaking a bugfix and people should have filed a bug instead of relying on (somewhat obviously) strange behavior. Also, depending on the result of `toString` is brittle in most cases, as libraries usually don't give any compatibility guarantees on the value of `toString`.

@akka/akka-http-team WDYT?

jypma:
LGTM. It's a plain bugfix, so we should mention it in the release notes. Anyone relying on `drop(2)` will hopefully fairly quickly realize the mayhem caused by their upgrading :)

ktoso:
I agree that we should accept this PR and mention it explicitly in release notes of 10.0.2.

ulasakdeniz:
Thank you for the review

jrudolph:
Cool. Btw. I already have a PR open that uses the `drop(2)` thingy ;) But I agree, it's manageable mayhem.

PR #803, opened 2017-01-22
Allow Java bind API to specify port zero (#660)

### LINKED ISSUE #660: Java API for bindAndHandle does not allow port 0
We use port 0 during tests extremely frequently. As we do parallel tests, this allows Akka-HTTP to find me an open ephemeral port and bind there. We then read the actual port back from the resulting ServerBinding.

When writing Java API tests, we just found out it does not accept 0 for `bindAndHandle`. To be more specific, The `ConnectHttp.toHost("localhost", 0)` throws an exception - not the bindAndHandle itself. The culprit is in this code in `ConnectHttp`:

```scala
  def toHost(host: String, port: Int): ConnectHttp = {
    require(port > 0, "port must be > 0")
    val start = if (isHttpOrHttps(host)) host else s"http://$host"
    toHost(Uri.create(start).port(port))
  }
```

I think the second line should rather spell:

```scala
    require(port >= 0, "port must be 0 or higher")
```

jonas:
This fixes #660 by bringing feature parity with the Scala API and also reworks `ConnectHttp` to improve how connection data is extracted.

Tested manually in the Scala REPL using the Java HttpApp:
```
> sbt akka-http/console
[info] Starting scala interpreter...
[info]
Welcome to Scala 2.11.8 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_102).
Type in expressions for evaluation. Or try :help.

scala> :paste
// Entering paste mode (ctrl-D to finish)

import com.typesafe.config.ConfigFactory
import akka.http.javadsl.server.HttpApp
import akka.http.javadsl.settings.ServerSettings

object Server extends HttpApp {
  protected override def route = complete("hi")
  def run() = startServer("localhost", 0, ServerSettings.create(ConfigFactory.load()))
}

Server.run

// Exiting paste mode, now interpreting.

Press RETURN to stop...
[INFO] [01/22/2017 00:37:54.153] [ForkJoinPool.commonPool-worker-2] [akka.actor.ActorSystemImpl(readiwiwServer)] Server online at http://localhost:54855/
```

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/javadsl/ConnectHttp.scala
+++ b/akka-http-core/src/main/scala/akka/http/javadsl/ConnectHttp.scala
@@ -29,54 +29,78 @@ object ConnectHttp {
 
   // TODO may be optimised a bit to avoid parsing the Uri entirely for the known port cases
 
-  /** Extracts host data from given Uri. */
-  def toHost(uriHost: Uri): ConnectHttp = {
-    val s = uriHost.scheme.toLowerCase(Locale.ROOT)
-    if (s == "https") new ConnectHttpsImpl(uriHost.host.address, effectivePort(s, uriHost.port))
-    else new ConnectHttpImpl(uriHost.host.address, effectivePort(s, uriHost.port))
-  }
+  /** Extracts HTTP or HTTPS connection data from given Uri. */
+  def toHost(uriHost: Uri): ConnectHttp =
+    toHost(uriHost, uriHost.port)
 
-  def toHost(host: String): ConnectHttp = {
-    if (isHttpOrHttps(host)) toHost(Uri.create(host))
-    else toHost(Uri.create(s"http://$host"))
-  }
+  /**
+   * Extract HTTP or HTTPS connection data from given host.
+   *
+   * The host string may contain a URI or a <host>:<port> pair.
+   */
+  def toHost(host: String): ConnectHttp =
+    toHost(createUriWithScheme("http", host))
 
+  /**
+   * Extracts HTTP or HTTPS connection data from given host and port.
+   *
+   * The host string may contain a URI or a <host>:<port> pair. In both cases the
+   * port is ignored.
+   */
   def toHost(host: String, port: Int): ConnectHttp = {
     require(port > 0, "port must be > 0")
-    val start = if (isHttpOrHttps(host)) host else s"http://$host"
-    toHost(Uri.create(start).port(port))
+    toHost(createUriWithScheme("http", host), port)
+  }
+
+  private def toHost(uriHost: Uri, port: Int): ConnectHttp = {
+    val s = uriHost.scheme.toLowerCase(Locale.ROOT)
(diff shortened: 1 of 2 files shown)


akka-http-core/src/main/scala/akka/http/javadsl/ConnectHttp.scala
```diff
@@ -29,59 +29,89 @@ object ConnectHttp {
...
+  private def toHost(uriHost: Uri, port: Int): ConnectHttp = {
+    val s = uriHost.scheme.toLowerCase(Locale.ROOT)
+    if (s == "https") new ConnectHttpsImpl(uriHost.host.address, effectivePort(s, port))
+    else new ConnectHttpImpl(uriHost.host.address, effectivePort(s, port))
   }
 
   /**
-   * Extracts host data from given Uri.
-   * Forces an HTTPS connection to the given host, using the default HTTPS context and default port.
+   * Extracts HTTPS connection data from given host and port.
```
ktoso: `connection data`? host data

jonas: Yes, the phrasing can be improved. From my understanding `ConnectHttp` holds settings related to server binding, so maybe:

```
Extracts host settings {|from given host|from given host and port} used for the server binding.
```

jrudolph:
The fix looks good, though I have to admit I've seen this API for the first time.

akka-http-core/src/main/scala/akka/http/javadsl/ConnectHttp.scala
ktoso: Yeah that rewording would be good. We want to release today so let's merge and amend it

jrudolph: Actually, I think the current wording ("connection data") is consistent with the rest of the file. Let's keep it like that.

I think there are other issues with this API (like #823) and the overall naming that could be fixed at another time in a more wholesale fashion.

ktoso: Ok, sounds good. Let's clean up later on more completely.
That API came into being to avoid exponential explosion of method overloads we had to provide for bind() and all of its versions btw.

akka-http-core/src/main/scala/akka/http/javadsl/ConnectHttp.scala
```diff
@@ -29,59 +29,89 @@ object ConnectHttp {
...
+   *
+   * The host string may contain a URI or a <host>:<port> pair.
+   */
+  def toHost(host: String): ConnectHttp =
+    toHost(createUriWithScheme("http", host))
 
+  /**
+   * Extracts HTTP or HTTPS connection data from given host and port.
+   *
+   * The host string may contain a URI or a <host>:<port> pair. In both cases the
```
jrudolph: Thanks for documenting this. I think the API should be improved. I created https://github.com/akka/akka-http/issues/833 to track this.

jrudolph:
Thanks, @jonas.

Issue #773, opened 2017-01-16
HTTP2: Respect SETTINGS_INITIAL_WINDOW_SIZE


ktoso:
SETTINGS_INITIAL_WINDOW_SIZE (0x4):
Indicates the sender's initial window size (in octets) for stream-level flow control. The initial value is 216-1 (65,535) octets.

This setting affects the window size of all streams (see Section 6.9.2).

Values above the maximum flow-control window size of 231-1 MUST be treated as a connection error (Section 5.4.1) of type FLOW_CONTROL_ERROR.

jrudolph:
This setting is already handled. The only thing to do here is to check the received value against the maximum size.

jrudolph:
Remaining part fixed by #816.

PR #816, opened 2017-01-24
Enforce upper limit of SETTINGS_INITIAL_WINDOW_SIZE 

johanandren:
Refs #773

### DIFF
--- a/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
+++ b/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
@@ -183,11 +183,11 @@ class Http2ServerDemux extends GraphStage[BidiShape[Http2SubStream, FrameEvent, 
 
               settings.foreach {
                 case Setting(Http2Protocol.SettingIdentifier.SETTINGS_INITIAL_WINDOW_SIZE, value) ⇒
-                  if (value <= Http2Protocol.MaxInitialWindowSize) {
+                  if (value >= 0) {
                     debug(s"Setting initial window to $value")
                     multiplexer.updateDefaultWindow(value)
                   } else {
-                    pushGOAWAY(FLOW_CONTROL_ERROR, s"Too large SETTINGS_INITIAL_WINDOW_SIZE: $value")
+                    pushGOAWAY(FLOW_CONTROL_ERROR, s"Invalid value for SETTINGS_INITIAL_WINDOW_SIZE: $value")
                     settingsAppliedOk = false
                   }
                 case Setting(Http2Protocol.SettingIdentifier.SETTINGS_MAX_FRAME_SIZE, value) ⇒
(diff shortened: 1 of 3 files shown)


akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -560,6 +560,13 @@ class Http2ServerSpec extends AkkaSpec("" + "akka.loglevel = debug")
 
         expectSettingsAck() // TODO check that the setting was indeed applied
       }
+
+      "react on too large SETTINGS_INITIAL_WINDOW_SIZE with FLOW_CONTROL_ERROR" in new TestSetup with RequestResponseProbes {
+        sendSETTING(SettingIdentifier.SETTINGS_INITIAL_WINDOW_SIZE, Math.pow(2, 32).toInt) // valid values are below 2^32 - 1)
```
johanandren: Dang, it was supposed to be 2^31

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -178,10 +178,18 @@ class Http2ServerDemux extends GraphStage[BidiShape[Http2SubStream, FrameEvent,
...
             case SettingsFrame(settings) ⇒
               if (settings.nonEmpty) debug(s"Got ${settings.length} settings!")
+
+              var settingsAppliedOk = true
+
               settings.foreach {
                 case Setting(Http2Protocol.SettingIdentifier.SETTINGS_INITIAL_WINDOW_SIZE, value) ⇒
-                  debug(s"Setting initial window to $value")
-                  multiplexer.updateDefaultWindow(value)
+                  if (value <= Http2Protocol.MaxInitialWindowSize) {
```
jrudolph: Isn't the value `Int.MaxValue`? So, the condition will be trivially true. I think the sign needs to be checked in the parser instead.

johanandren: `Int.MaxValue - 1` actually

johanandren: I take that back, it is exactly Int.MaxValue, you are right

johanandren: Will we be able to reply with a FLOW_CONTROL_ERROR if the check is in the parser?

jrudolph: Yes, by throwing the right exception, there should be similar things checked there.

johanandren: When I think of it, can't I just as well check the sign here?

jrudolph: Yes, I guess so. I thought we had precedent of checking other settings in the parser but it seems we don't. Go for it ;)

jrudolph:
Thanks, @johanandren, LGTM.

Issue #827, opened 2017-01-26
HttpAppSpec incompatible with Akka 2.5


jrudolph:
As noted in the nightly build, https://jenkins.akka.io:8498/job/akka-http-nightly-against-akka-master-on-2.12/38//consoleFull

```
[error] /home/jenkins/workspace/workspace/akka-http-nightly-against-akka-master-on-2.12/akka-http-tests/src/test/scala/akka/http/javadsl/server/HttpAppSpec.scala:62: value isTerminated is not a member of akka.actor.ActorSystem
[error]       system.isTerminated should ===(false)
[error]              ^
[error] /home/jenkins/workspace/workspace/akka-http-nightly-against-akka-master-on-2.12/akka-http-tests/src/test/scala/akka/http/scaladsl/server/HttpAppSpec.scala:85: value isTerminated is not a member of akka.actor.ActorSystem
[error]       system.isTerminated should ===(false)
[error]              ^
```

jlprat:
Can be removed, it was just a sanity check, to ensure that HttpApp is not terminating the `ActorSystem`.

There is no replacement for this, right?

ktoso:
of course there is ;) `whenTerminated`

PR #828, opened 2017-01-26
Use whenTerminated instead of deprecated isTerminated #827

jlprat:
Issue: #827 
Replace deprecated (and soon to be removed) `isTerminated` in `HttpAppSpec` with `whenTerminated`


ktoso:
L to tha M

PR #829, opened 2017-01-26
Create virtual projects to run tests also against 2.5-M1

jrudolph:


build.sbt
```diff
@@ -60,7 +60,10 @@ lazy val root = Project(
     httpTests,
     httpMarshallersScala,
     httpMarshallersJava,
-    docs
+    docs,
+    httpCoreTestsAkka25,
+    httpTestsAkka25,
+    docsTestsAkka25
```
ktoso: great idea

ktoso:
LGTM (though may fail now?)

jrudolph:
With the HttpApp fixed I don't think it fails because it doesn't compile against 2.5 but only run the tests against.

jrudolph:
Hmm it adds 25% to PR validation time. Not sure if that it's worth it. I think I remove it from the aggregation and add it manually to the nightly build.

jonas:
Did you consider using a settings key to make the Akka version configurable? This would avoid having to maintain the virtual projects.

    lazy val akkaTestVersion = settingKey[String]("The version of Akka to use for tests.")

In the more exotic department there is also sbt synthetic projects which could be a fit? http://www.scala-sbt.org/0.13/docs/sbt-0.13-Tech-Previews.html#Synthetic+subprojects

jrudolph:
Good ideas in here but may not make sense to run in PR validation. Closing for now...

jonas:
Do you mind if I follow up with a PR to make it possible to run tests against Akka 2.5?

jrudolph:
No, that would be really nice. We can still decide, then, whether we want to run them in PR validation, only in nightlies, or if we want to introduce another build that tracks master.

Issue #758, opened 2017-01-12
Http2: Allow PRIORITY frames for streams in idle state


jrudolph:
Right now we send a GOAWAY frame because the frames are not yet in our map.

As firefox will use the first few streams as a priority tree to hang later resource requests into.

http://bitsup.blogspot.de/2015/01/http2-dependency-priorities-in-firefox.html

jrudolph:
This blog post from akamai shows it in a bit more detail what firefox does: https://blogs.akamai.com/2016/04/http2-enables-more-intelligent-resource-loading-via-stream-dependencies-but-its-not-as-simple-as-you.html

gosubpl:
I will look into that.

ktoso:
Hey @gosubpl, this is something we want to have in the release in this week so I'll start working on it. Hope you don't mind :-)

gosubpl:
Absolutely, please hack away. In the meantime - anything of a bit lower priority for me?

ktoso:
Various settings handlings would be good https://github.com/akka/akka-http/labels/t%3Ahttp2

ktoso:
I'll mark this one as resolved, actually handling them is a separate story, will add a ticket

Issue #531, opened 2016-11-16
Http2: Ignore trailing headers for now


jrudolph:
Before we get to supporting trailing headers (#530) we might just want to ignore them.

However, trailing HEADERS frames need to be "actively ignored" because they change the state of the header compression so we need to receive and decompress the headers even if we don't push them to the user.

ktoso:
I'll just add this now, should be simple.

ktoso:
Actually wasn't, leaving for now.

jrudolph:
Fixed by #531.

PR #830, opened 2017-01-26
+ht2 #758 allow early priotity frames

ktoso:
Refs #758, only the allowing part. 
Also added debug information to all GOAWAYs

### DIFF
--- a/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
+++ b/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
@@ -160,6 +160,7 @@ class Http2ServerDemux extends GraphStage[BidiShape[Http2SubStream, FrameEvent, 
               pushGOAWAY(ErrorCode.PROTOCOL_ERROR, "Unknown stream id: " + e.streamId)
 
             case h: ParsedHeadersFrame ⇒
+              // TODO: probably a trailing header frame here which we currently don't support but which we should ignore. 
               pushGOAWAY(ErrorCode.PROTOCOL_ERROR, "Unexpected internal frame reached Demux! Was: " + h)
 
             case DataFrame(streamId, endStream, payload) ⇒
--- a/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
+++ b/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
@@ -103,7 +103,6 @@ class Http2ServerDemux extends GraphStage[BidiShape[Http2SubStream, FrameEvent, 
       private var closedAfter: Option[Int] = None
       private var incomingStreams = mutable.Map.empty[Int, SubStream]
       private var maxConcurrentStreams: Option[Int] = None
-
       /**
        * The "last peer-initiated stream that was or might be processed on the sending endpoint in this connection"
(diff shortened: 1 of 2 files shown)


akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -152,12 +151,16 @@ class Http2ServerDemux extends GraphStage[BidiShape[Http2SubStream, FrameEvent,
 
               dispatchSubstream(Http2SubStream(frame, data))
 
+            case PriorityFrame(streamId, exclusiveFlag, streamDependency, weight) ⇒
+              // must be before the "unknown stream id" case, since can be sent eagerly (e.g. firefox does so)
+              debug(s"Received PriorityFrame for stream $streamId with ${if (exclusiveFlag) "exclusive " else "non-exclusive "} dependency on stream $streamDependency and weight $weight")
```
ktoso: this move is actually enough to keep firefox happy

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -152,12 +151,16 @@ class Http2ServerDemux extends GraphStage[BidiShape[Http2SubStream, FrameEvent,
...
+
             case e: StreamFrameEvent if !incomingStreams.contains(e.streamId) ⇒
               // if a stream is invalid we will GO_AWAY
-              pushGOAWAY()
+              pushGOAWAY(ErrorCode.PROTOCOL_ERROR, "Unknown stream id: " + e.streamId)
 
             case h: ParsedHeadersFrame ⇒
-              pushGOAWAY()
+              pushGOAWAY(ErrorCode.PROTOCOL_ERROR, "Unexpected internal frame reached Demux! Was: " + h)
```
jrudolph: The error message here is not quite correct. When we receive it here it is probably a trailing header frame here which we currently don't support but which we should ignore. It's ok to leave it like this but add a TODO comment to fix it in the course of #531.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -573,7 +573,7 @@ class Http2ServerSpec extends AkkaSpec("" + "akka.loglevel = debug")
         sendSETTING(SettingIdentifier.SETTINGS_MAX_CONCURRENT_STREAMS, 1)
         expectSettingsAck()
 
-        // TODO actually apply the limiting and verify it works
+        // TODO actually apply the limiting and verify it works
```
jrudolph: extra whitespace here

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
jrudolph: :+1:

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
ktoso: Will add a todo explaining that 👍

jrudolph:
LGTM (maybe add the small comment)

PR #837, opened 2017-01-26
Release notes for 10.0.3 (and stubs for previous versions)

jrudolph:


jrudolph:
Thanks, @ktoso good point. We can either retag or just fix the version before pushing the docs.

ktoso:
retag is fine here IMO

jonas:
This is great. I would suggest a more consistent use of capital letters to start each entry.

@ktoso You can alway build the docs forcing a specific version with something like:

    sbt 'project docs' 'set version in paradox := "10.0.3" paradox

ktoso:
thats lying about what they are though. Next time someone runs thigns you suddenly did wrong docs (missing pages). Builds must be repeatable.

jrudolph:
@jonas regarding capital letters, you mean the bullet points, right? Will do.

jonas:
@jrudolph Yes, the bullet points randomly start with or without capital letter. We should pick one way. ;)

jrudolph:
Yes, absolutely. I started with the log lines from the commits and then forget to go over it again.

jrudolph:
Forgot to push the capital letters which I now pushed to master directly.

ktoso:
ok ;)

Issue #761, opened 2017-01-12
Java Docs. Enrich Http Model documentation


jlprat:
The Java version of the Http Model documentation is slightly behind the Scala one.
Please provide documentation / samples about:

- [x] Limiting message entity length
- [x] Custom Headers
- [x] Registering Custom Media Types
- [x] Registering Custom Status Codes
- [x] The URI model (the small explanation in Http Model, not the dedicated page)

jlprat:
Will work on this one, as I need to go through this to improve https://github.com/akka/akka-http/pull/771

jlprat:
Only custom headers is left. PR is #888

PR #844, opened 2017-01-31
Provide Http Model documentation for Java #761

jlprat:
Issue: #761
Documentation for:
* Limiting message entity lenght
* Registering Custom Media Types
* Registering custom Status Codes
* The Uri Model

Additionally, fixed some typos and wordings

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/javadsl/Http.scala
+++ b/akka-http-core/src/main/scala/akka/http/javadsl/Http.scala
@@ -540,7 +540,7 @@ class Http(system: ExtendedActorSystem) extends akka.actor.Extension {
    * Fires a single [[HttpRequest]] across the (cached) host connection pool for the request's
    * effective URI to produce a response future.
    *
-   * The given [[HttpsConnectionContext]] will be used for encruption if the request is sent to an https endpoint.
+   * The given [[HttpsConnectionContext]] will be used for encryption if the request is sent to an https endpoint.
    *
    * Note that the request must have either an absolute URI or a valid `Host` header, otherwise
    * the future will be completed with an error.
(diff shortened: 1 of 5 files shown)


jlprat:
Custom Headers is missing from this PR as the full API as stated in Scala docs is missing. If you want, I will address it in a different PR.

jlprat:
Looking at the Custom Headers documentation for Scala [custom-headers](http://doc.akka.io/docs/akka-http/current/scala/http/common/http-model.html#custom-headers) I don't think we can provide a similar highlevel API for it. And the only way to supply a Custom Header would be with extending `CustomHeader` (which BTW is discouraged in the Scala API). What do you think?

ktoso:
What do you mean that we can't provide it? We totally can as far as I can see - what's the blocker in your opinion?

jlprat:
I was thinking about things like:
```scala
val RawHeader(k2, v2) = ApiTokenHeader("token")
k2 should ===("apiKey")
v2 should ===("token")

val ApiTokenHeader(v3) = RawHeader("APIKEY", "token")
v3 should ===("token")
```
Mainly the last one, in Java could be quite nasty. Meaning we create a `RawHeader` but then we treat it as `ApiTokenHeader`.

Maybe I'm looking at the problem from a too close distance and I need to step back. But I don't think this is easily doable for Java without some castings.

ktoso:
That's a niffty Scala feature which Java simply does not have in this syntax and would simply be:
`ApiTokenHeader.unapply(rawHeader): Optional<ApiTokenHeader>` in java

jlprat:
So do we want to offer this as a feature?

ktoso:
Yes, since I think this also is used by Http infra to create those instances from raw headers.

jlprat:
OK, what do you think about addressing this in a separate PR, as this one is purely about docs?

ktoso:
I'll review later today and chime in on that

jlprat:
I'll try to find some time to work on the pending feature. I will start in a different branch, and I can merge later if needed

jlprat:
Thanks for reviewing!

Issue #842, opened 2017-01-30
Docs. Include Search when screen is small too


jlprat:
Currently, when the screen is too small, and the template falls back to put the navigation as a side menu, the search field from Algolia disappears.
Please  provide the search field also on small screens

jlprat:
I can provide a PR for this as well

jlprat:
This can be closed :)

ktoso:
Thanks

PR #845, opened 2017-02-01
Show search input in small screens #842

jlprat:
Issue: #842 
Include search input when screen is small

### DIFF
--- a/docs/src/main/paradox/_template/page.st
+++ b/docs/src/main/paradox/_template/page.st
@@ -39,13 +39,17 @@
           <a href="#" class="off-canvas-toggle hide-for-medium" data-toggle="off-canvas-menu">$menu()$</a>
 
           <img src="$page.base$css/../mini-akka-logo.png" alt="Akka logo" style="float: left"/> $title()$
+
+          <div style="position: relative; right:0; min-width:25%" class="float-right">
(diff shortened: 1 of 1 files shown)


jlprat:
How it would look in big screens:
search_big
How it would look in mid screens:
search_mid
How it would look in small screens:
search_small

docs/src/main/paradox/_template/page.st
```diff
@@ -39,13 +39,17 @@
           <a href="#" class="off-canvas-toggle hide-for-medium" data-toggle="off-canvas-menu">$menu()$</a>
 
           <img src="$page.base$css/../mini-akka-logo.png" alt="Akka logo" style="float: left"/> $title()$
+
+          <div style="position: relative; right:0; min-width:25%" class="float-right">
+            <input type="search" id="search" class="form-control" style="margin: 10px 0;" placeholder="Search the docs..."/>
+          </div>
+
```
jonas: I think code in the footer be removed then:

```
$! Override Algolia's search wrapper styles to let the search box span the whole left panel !$
 <style>.algolia-autocomplete { display: block !important }</style>
```

jlprat: Checking it right now

jlprat: If this is removed the input field is smaller then...

jonas: OK, thanks for checking

PR #847, opened 2017-02-04
Remove duplicated code

jonas:


ktoso:
LGTM, seems good, thanks :)

jonas:
Cherry-picked another removal from my `add-dilated-to-timeout` PR

akka-http-core/src/test/scala/akka/stream/testkit/TestUtils.scala
```diff
@@ -1,44 +0,0 @@
-/**
```
ktoso: good

Issue #824, opened 2017-01-25
FAILED: akka.http.scaladsl.server.HttpAppSpec.HttpApp should start without ActorSystem


jrudolph:
https://jenkins.akka.io:8498/job/pr-validator-akka-http/704/testReport/junit/akka.http.scaladsl.server/HttpAppSpec/HttpApp_should_start_without_ActorSystem/

```
Error Message

The code passed to eventually never returned normally. Attempted 150 times over 5.030236196999999 seconds. Last failure message: The future returned an exception of type: akka.stream.StreamTcpException, with message: Tcp command [Connect(localhost:35085,None,List(),Some(10 seconds),true)] failed..

Stacktrace

org.scalatest.exceptions.TestFailedDueToTimeoutException: The code passed to eventually never returned normally. Attempted 150 times over 5.030236196999999 seconds. Last failure message: The future returned an exception of type: akka.stream.StreamTcpException, with message: Tcp command [Connect(localhost:35085,None,List(),Some(10 seconds),true)] failed..
	at org.scalatest.concurrent.Eventually$class.tryTryAgain$1(Eventually.scala:421)
	at org.scalatest.concurrent.Eventually$class.eventually(Eventually.scala:439)
	at akka.http.scaladsl.server.HttpAppSpec.eventually(HttpAppSpec.scala:26)
	at akka.http.scaladsl.server.HttpAppSpec.akka$http$scaladsl$server$HttpAppSpec$$callAndVerify(HttpAppSpec.scala:202)
	at akka.http.scaladsl.server.HttpAppSpec$$anonfun$1$$anonfun$apply$mcV$sp$2.apply(HttpAppSpec.scala:64)
	at akka.http.scaladsl.server.HttpAppSpec$$anonfun$1$$anonfun$apply$mcV$sp$2.apply(HttpAppSpec.scala:52)
	at org.scalatest.OutcomeOf$class.outcomeOf(OutcomeOf.scala:85)
	at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
	at org.scalatest.Transformer.apply(Transformer.scala:22)
	at org.scalatest.Transformer.apply(Transformer.scala:20)
	at org.scalatest.WordSpecLike$$anon$1.apply(WordSpecLike.scala:1078)
	at org.scalatest.TestSuite$class.withFixture(TestSuite.scala:196)
	at akka.testkit.AkkaSpec.withFixture(AkkaSpec.scala:57)
[...]

PR #849, opened 2017-02-06
Improve HttpApp stability #824

jlprat:
Issue: #824 
Attempt to fix this issue, and make tests more reliable.
reduce calls, to save threads

### DIFF
--- a/akka-http-tests/src/test/scala/akka/http/javadsl/server/HttpAppSpec.scala
+++ b/akka-http-tests/src/test/scala/akka/http/javadsl/server/HttpAppSpec.scala
@@ -6,6 +6,7 @@ package akka.http.javadsl.server
 
 import java.util.concurrent.TimeUnit
 
+import akka.Done
 import akka.http.javadsl.settings.ServerSettings
 import akka.http.scaladsl.{ Http, TestUtils }
 import akka.http.scaladsl.client.RequestBuilding
@@ -17,23 +18,37 @@ import org.scalatest.concurrent.Eventually
 
 import scala.concurrent.duration.Duration
 import scala.concurrent.{ Await, Future }
-import scala.runtime.BoxedUnit
 
 class HttpAppSpec extends AkkaSpec with RequestBuilding with Eventually {
   import system.dispatcher
 
-  "HttpApp" should {
+  def withMinimal(testCode: (MinimalHttpApp, String, Int) ⇒ Any): Unit = {
+    val (_, host, port) = TestUtils.temporaryServerHostnameAndPort()
+    val minimal = new MinimalHttpApp()
+    try testCode(minimal, host, port)
+    finally {
+      if (!minimal.shutdownTrigger.isDone()) minimal.shutdownTrigger.complete(Done)
+    }
+  }
 
-    "start without ActorSystem" in {
-      val (_, host, port) = TestUtils.temporaryServerHostnameAndPort()
+  def withSneaky(testCode: (SneakHttpApp, String, Int) ⇒ Any): Unit = {
+    val (_, host, port) = TestUtils.temporaryServerHostnameAndPort()
+    val sneaky = new SneakHttpApp()
+    try testCode(sneaky, host, port)
+    finally {
+      if (!sneaky.shutdownTrigger.isDone()) sneaky.shutdownTrigger.complete(Done)
+    }
+  }
 
-      val minimal = new MinimalHttpApp()
+  "HttpApp Java" should {
+
+    "start without ActorSystem" in withMinimal { (minimal, host, port) ⇒
 
       val server = Future {
         minimal.startServer(host, port, ServerSettings.create(ConfigFactory.load))
       }
 
-      callAndVerify(host, port, "foo")
+      minimal.bindingPromise.get(5, TimeUnit.SECONDS)
 
       // Requesting the server to shutdown
       callAndVerify(host, port, "shutdown")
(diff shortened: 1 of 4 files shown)


jlprat:
Test fails because it can't start an `ActorSystem` and the server within 2 seconds.

jlprat:
Increased timeout. I couldn't test on Win machine though...

akka-http-tests/src/test/scala/akka/http/javadsl/server/HttpAppSpec.scala
```diff
@@ -22,48 +23,55 @@ import scala.runtime.BoxedUnit
 class HttpAppSpec extends AkkaSpec with RequestBuilding with Eventually {
   import system.dispatcher
 
-  "HttpApp" should {
+  "HttpApp  Java" should {
```
ktoso: one space

akka-http-tests/src/test/scala/akka/http/javadsl/server/HttpAppSpec.scala
```diff
@@ -22,48 +23,55 @@ import scala.runtime.BoxedUnit
...
 
     "start without ActorSystem" in {
-      val (_, host, port) = TestUtils.temporaryServerHostnameAndPort()
 
       val minimal = new MinimalHttpApp()
 
       val server = Future {
-        minimal.startServer(host, port, ServerSettings.create(ConfigFactory.load))
+        minimal.startServer("localhost", 0, ServerSettings.create(ConfigFactory.load))
```
ktoso: Seems weird still that this would be more stable... why was that?

jonas: As mentioned in #824 there is a potential race condition when probing a port using `temporaryServerHostnameAndPort `.

akka-http-tests/src/test/scala/akka/http/javadsl/server/HttpAppSpec.scala
```diff
@@ -22,48 +23,55 @@ import scala.runtime.BoxedUnit
...
-      // Requesting the server to shutdown
-      callAndVerify(host, port, "shutdown")
+      try {
+        // Requesting the server to shutdown
+        callAndVerify(binding.localAddress.getHostName, binding.localAddress.getPort, "shutdown")
 
-      Await.ready(server, Duration(1, TimeUnit.SECONDS))
-      server.isCompleted should ===(true)
-      system.whenTerminated.isCompleted should ===(false)
+        Await.ready(server, Duration(1, TimeUnit.SECONDS))
```
jonas: Given the amount of additional code to ensure proper shutdown would it make sense to also change tests to either use test fixtures/contexts or before/after hooks to handle this?

akka-http-tests/src/test/scala/akka/http/scaladsl/server/HttpAppSpec.scala
```diff
@@ -50,143 +56,155 @@ class HttpAppSpec extends AkkaSpec with Directives with RequestBuilding with Eve
...
+        }
 
         override protected def postServerShutdown(attempt: Try[Done], system: ActorSystem): Unit = postShutdownCalled.set(true)
       }
 
       "shutting down" in {
-        val (_, host, port) = TestUtils.temporaryServerHostnameAndPort()
 
-        val sneaky = new SneakSever()
+        val sneaky = new SneakySever()
```
jonas: s/Sever/Server/

akka-http-tests/src/test/scala/akka/http/javadsl/server/HttpAppSpec.scala
jlprat: I think now I know why it failed that time, and the reason behind the deadlock in win.
It's not port collision (the stack trace we saw was the expected from binding failure test), it is that the server in rare occasions can't start in 5 seconds (start the actor system and the server). This causes an assertion failure, which prevents the server to be correctly shutdown, which is waiting for a `Future` to complete.
I could go back to the random port utility as it's simpler

akka-http-tests/src/test/scala/akka/http/javadsl/server/HttpAppSpec.scala
jlprat: Will look into it

akka-http-tests/src/test/scala/akka/http/scaladsl/server/HttpAppSpec.scala
```diff
@@ -42,23 +43,63 @@ class HttpAppSpec extends AkkaSpec with Directives with RequestBuilding with Eve
...
   }
 
-  "HttpApp" should {
+  class SneakyServer extends MinimalApp {
+
+    val postBindingCalled = new AtomicBoolean(false)
+    val postBindingFailureCalled = new AtomicBoolean(false)
+    val postShutdownCalled = new AtomicBoolean(false)
+
+    override protected def waitForShutdownSignal(system: ActorSystem)(implicit ec: ExecutionContext): Future[Done] = {
```
jonas: This `def` could be removed since it is inherited.

jlprat: True, will do in a sec

jrudolph:
Thanks a lot, @jlprat.

PR #850, opened 2017-02-07
add "or 2.12" in the introduction.md to make clear it's supported

emanresusername:
just to avoid the scenario of someone landing on this page, only seeing `2.11` and assuming there's no `2.12` support


jrudolph:
OK TO TEST

ktoso:
Good point, thanks!

PR #861, opened 2017-02-13
Add reference to SBT/Gradle/Maven artifact inclusion details

retrospectacus:


ktoso:
Good idea to link back I think. Not sure if this is the best place hmm...

cc @akka/akka-http-team

jlprat:
What about under `docs/src/main/paradox/scala/http/introduction.md` (and the Java one as well) just below the explanation of the different modules?

jrudolph:
OK TO TEST

jrudolph:
Hmm, yes, I agree with @jlprat that this rather belongs on the introduction page..

jrudolph:
Amended the commit to move the line around to the introductory page below the sbt config line.

jrudolph:
Thanks, @retrospectacus.

PR #869, opened 2017-02-16
=htc #516 fix pull/push twice in server ControllerStage

jrudolph:
Refs #516.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
@@ -417,7 +417,7 @@ private[http] object HttpServerBluePrint {
 
           emit(responseCtxOut, ResponseRenderingContext(response, requestStart.method, requestStart.protocol, close),
             pullHttpResponseIn)
-          if (!isClosed(requestParsingIn) && close && requestStart.expect100Continue) pull(requestParsingIn)
+          if (!isClosed(requestParsingIn) && close && requestStart.expect100Continue) maybePullRequestParsingIn()
         }
         override def onUpstreamFinish() =
           if (openRequests.isEmpty && isClosed(requestParsingIn)) completeStage()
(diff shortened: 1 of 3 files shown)


akka-http-core/src/main/scala/akka/http/impl/util/LogByteStringTools.scala
```diff
@@ -20,7 +20,7 @@ private[akka] object LogByteStringTools {
       logByteString(s"$name UP  ", maxBytes)
     )
 
-  def logToStringBidi[A](name: String, maxBytes: Int = MaxBytesPrinted): BidiFlow[A, A, A, A, NotUsed] =
+  def logToStringBidi[A, B](name: String, maxBytes: Int = MaxBytesPrinted): BidiFlow[A, A, B, B, NotUsed] =
```
ktoso: 👍

jrudolph:
PLS BUILD

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -461,6 +461,12 @@ private[http] object HttpServerBluePrint {
       def emitErrorResponse(response: HttpResponse): Unit =
         emit(responseCtxOut, ResponseRenderingContext(response, closeRequested = true), () ⇒ completeStage())
 
+      def maybePullRequestParsingIn(): Unit =
```
viktorklang: Since pull-suppression is related to `oneHundredContinueResponsePending` wouldn't it be more appropriate to make that tri-state (since I assume that oneHundredContinueResponsePending cannot be false while pullSuppressed is true)?

jrudolph: Yes, afaics this is correct. fwiw, we could also just remove `pullSuppressed` and use `!hasBeenPulled(requestParsingIn) && isAvailable(requestPrepOut)` instead.

I think the problem with that whole stage is that it is really hard to understand which states are mixed up there. I think at some point it needs a proper rewrite. Until then we try to only fix what's broken.

viktorklang: Fixing whats broken makes sense—increasing complexity needlessly seems to be at odds with BSR.
@jrudolph If `!hasBeenPulled(requestParsingIn) && isAvailable(requestPrepOut)` does the same job as the var, then I think the var loses. :)

Issue #883, opened 2017-02-20
Upgrade MiMa dependency


jlprat:
In order to be able to change MiMa exclusion filters without the need to reload the project, we should upgrade to version `0.1.14`

jrudolph:
Fixed by #884.

PR #884, opened 2017-02-20
Upgrade MiMa dependency #883

jlprat:
Issue: #883
In order to be able to modify filters without reloading the project


jrudolph:
Great, thanks for testing ;)

PR #901, opened 2017-02-25
Avoid redirect loops when `redirectToNoTrailingSlashIfPresent` is used

### LINKED ISSUE #878: Fix redirectToNoTrailingSlashIfPresent not to go into redirect loop for root path
In Http, the root path cannot be empty and must be at least a single slash. So, if you remove the slash for the root URL and redirect, it will automatically be added leading to the redirect loop.

/cc @lihaoyi

kubukoz:
Fixes #878.

I'm not sure if it's the right way to do it - if you have any suggestions on how to make it better, please tell ;)

### DIFF
--- a/akka-http/src/main/scala/akka/http/scaladsl/server/directives/PathDirectives.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/server/directives/PathDirectives.scala
@@ -177,13 +177,11 @@ trait PathDirectives extends PathMatchers with ImplicitPathMatcherConstruction w
   /**
    * If the request path ends with a slash, redirect to the same uri without trailing slash in the path.
    *
-   * '''Caveat''': [[pathSingleSlash]] directive will not match inside of this directive.
-   *
    * @group path
    */
   def redirectToNoTrailingSlashIfPresent(redirectionType: StatusCodes.Redirection): Directive0 =
     extractUri.flatMap { uri ⇒
-      if (uri.path.endsWithSlash) {
+      if (uri.path.endsWithSlash && uri.path != Path.SingleSlash) {
         val newPath = uri.path.reverse.tail.reverse
         val newUri = uri.withPath(newPath)
         redirect(newUri, redirectionType)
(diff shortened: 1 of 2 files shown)


kubukoz:
Sure thing.

kubukoz:
Should I squash, or can it be done automatically when merging?

akka-http/src/main/scala/akka/http/scaladsl/server/directives/PathDirectives.scala
```diff
@@ -177,13 +177,11 @@ trait PathDirectives extends PathMatchers with ImplicitPathMatcherConstruction w
   /**
    * If the request path ends with a slash, redirect to the same uri without trailing slash in the path.
    *
-   * '''Caveat''': [[pathSingleSlash]] directive will not match inside of this directive.
```
jrudolph: I think the cave at is still valid. It's just a reminder not to expect trailing slashes inside that directive.

jrudolph: Or maybe it needs to be amended to talk about the root case.

Let's change the whole docs to

```
If the request path ends with a slash, redirect to the same uri without trailing slash in the path.

Note, however, that his directive doesn't apply to the uri consisting of just a single slash. HTTP does not support empty target paths, so that browsers will convert a uri such as `http://example.org` to `http://example.org/` adding
 the trailing slash. Redirecting the single slash path uri would lead to a redirection loop.

'''Caveat''': [[pathSingleSlash]] directive only match on the root path level inside of this directive.
```

jrudolph:
LGTM once the slight scaladoc fix has been applied.

kubukoz:
@jrudolph I made some language/caps fixes to your suggestion, should be ready to go now ;)

kubukoz:
Oh and can you explain the caveat to me again? I'm not sure if I got its meaning.

jrudolph:
@kubukoz I think the idea was that if you write

```scala
redirectToNoTrailingSlashIfPresent {
  someInnerRoute
}
```

then `someInnerRoute` will never receive a request with an URI that has a trailing slash (which is the whole purpose of the route).

kubukoz:
Unless the URI is just `/`, right? Makes sense, thanks.

jrudolph:
Yep, exactly.

Issue #912, opened 2017-02-28
Improve connection pool logging


jrudolph:
At debug level it should log:

 * state (changes) of pool slots
 * regular and irregular events in the processing of a request
 * state of the queue

jrudolph:
Fixed in #913.

PR #913, opened 2017-02-28
=htc #912 add logging for pool events

jrudolph:
This adds lots of logging around pools, so that we hopefully find bugs in long-running pools more easily.

Refs #912.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolConductor.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolConductor.scala
@@ -12,6 +12,7 @@ import akka.stream.scaladsl._
 import akka.stream._
 import akka.http.scaladsl.util.FastFuture
 import akka.http.scaladsl.model.HttpMethod
+import akka.macros.LogHelper
 import akka.stream.stage.GraphStage
 import akka.stream.stage.GraphStageLogic
 import akka.stream.stage.InHandler
@@ -114,8 +115,8 @@ private object PoolConductor {
   private case class Busy(openRequests: Int) extends SlotState { require(openRequests > 0) }
   private object Busy extends Busy(1)
 
-  private class SlotSelector(slotSettings: PoolSlotsSetting, pipeliningLimit: Int, log: LoggingAdapter)
-    extends GraphStage[FanInShape2[RequestContext, SlotEvent, SwitchSlotCommand]] {
+  private class SlotSelector(slotSettings: PoolSlotsSetting, pipeliningLimit: Int, val log: LoggingAdapter)
+    extends GraphStage[FanInShape2[RequestContext, SlotEvent, SwitchSlotCommand]] with LogHelper {
 
     private val requestContextIn = Inlet[RequestContext]("SlotSelector.requestContextIn")
     private val slotEventIn = Inlet[SlotEvent]("SlotSelector.slotEventIn")
@@ -129,11 +130,20 @@ private object PoolConductor {
       val slotStates = Array.fill[SlotState](slotSettings.maxSlots)(Unconnected)
       var nextSlot = 0
 
+      def updateSlotState(idx: Int, f: SlotState ⇒ SlotState): Unit = {
+        val oldState = slotStates(idx)
+        val newState = f(oldState)
+
+        debug(s"[$idx] $oldState -> $newState")
+
+        slotStates(idx) = newState
+      }
+
       setHandler(requestContextIn, new InHandler {
         override def onPush(): Unit = {
           val ctx = grab(requestContextIn)
           val slot = nextSlot
-          slotStates(slot) = slotStateAfterDispatch(slotStates(slot), ctx.request.method)
+          updateSlotState(slot, slotStateAfterDispatch(_, ctx.request.method))
           nextSlot = bestSlot()
           emit(slotCommandOut, SwitchSlotCommand(DispatchCommand(ctx), slot), tryPullCtx)
         }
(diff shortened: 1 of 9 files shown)


jrudolph:
Fixed mima issues.

2m:
LGTM.

I also wonder would it make sense to have an integration with flight recorder. Storing events in circular buffers would be great for cases where unexpected behavior happen long after application has started and would otherwise require lots of logs to be saved.

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolConductor.scala
```diff
@@ -129,11 +130,20 @@ private object PoolConductor {
       val slotStates = Array.fill[SlotState](slotSettings.maxSlots)(Unconnected)
       var nextSlot = 0
 
+      def updateSlotState(idx: Int, f: SlotState ⇒ SlotState): Unit = {
+        val oldState = slotStates(idx)
+        val newState = f(oldState)
+
+        debug(s"[$idx] $oldState -> $newState")
```
ktoso: Why string interpolation? We advice against it since this will always build the string even if debug level is off.
Use `log.debug("... {}", ...)` instead

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolFlow.scala
```diff
@@ -70,9 +69,7 @@ private object PoolFlow {
   */
   def apply(
     connectionFlow: Flow[HttpRequest, HttpResponse, Future[Http.OutgoingConnection]],
-    settings:       ConnectionPoolSettings, log: LoggingAdapter)(
-    implicit
-    system: ActorSystem, fm: Materializer): Flow[RequestContext, ResponseContext, NotUsed] =
```
ktoso: cool, nice we don't need them here

akka-parsing/src/main/scala/akka/macros/LogHelper.scala
```diff
@@ -0,0 +1,62 @@
+/*
+ * Copyright (C) 2009-2017 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.macros
```
ktoso: oh I see. Hm... I guess I don't mind macros, good - let's keep it.

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolConductor.scala
ktoso: Oh it's the macro.

akka-parsing/src/main/scala/akka/macros/LogHelper.scala
ktoso: Hm this package breaks OSGi though. It has to be in the package that this module is exporting.
(we'd hear back that we broke people's code...)

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolInterfaceActor.scala
```diff
@@ -44,7 +47,12 @@ private object PoolInterfaceActor {
  *   (ActorPublisher) and response sink (ActorSubscriber).
  */
 private class PoolInterfaceActor(gateway: PoolGateway)(implicit fm: Materializer)
-  extends ActorSubscriber with ActorPublisher[RequestContext] with ActorLogging {
+  extends ActorSubscriber with ActorPublisher[RequestContext] with LogHelper {
+  override val log = {
+    val scheme = if (gateway.hcps.setup.connectionContext.isSecure) "https" else "http"
+    Logging.getLogger(context.system, s"${context.system.name}/Pool(${gateway.gatewayId.name}->$scheme://${gateway.hcps.host}:${gateway.hcps.port})")
```
ktoso: Yeah, good name

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolInterfaceActor.scala
```diff
@@ -127,7 +138,7 @@ private class PoolInterfaceActor(gateway: PoolGateway)(implicit fm: Materializer
       responsePromise.completeWith(gateway(request))
 
     case Shutdown ⇒ // signal coming in from gateway
-      log.debug("Shutting down host connection pool to {}:{}", hcps.host, hcps.port)
+      debug("Shutting down host connection pool")
```
ktoso: This is the shutdown of the entire pool? Ok

jrudolph:
@2m absolutely. Flight recorder would be nice. I guess if we see flight recorder as an internal general purpose tool we might want to move it into akka-actor.

akka-parsing/src/main/scala/akka/macros/LogHelper.scala
jrudolph: We don't currently support OSGI in akka-http. Shouldn't be a problem to move that internal class later on.

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolInterfaceActor.scala
jrudolph: Yep, exactly, this is the shutdown for this host connection pool.

jrudolph:
@ktoso any more comments or are we good to merge?

akka-parsing/src/main/scala/akka/macros/LogHelper.scala
ktoso: Hm, okey, if someone cares we'll hear from them

ktoso:
Looks good, let's merge

jrudolph:
Thanks, @ktoso.

jrudolph:
@2m created #918 to track using FlightRecorder in the future.

Issue #916, opened 2017-03-01
Support for pluggable transports on the client-side


jrudolph:
Right now it's possible to use `Http().clientLayer` to implement all kinds of transports below HTTP on a single connection level. However, in the pool infrastructure the transport is currently hard-wired.

It would be nice to be able to define a new custom transport and configure it for the pool so the pool will use it when creating pooled connections.

This could be used to implement transports for running directly over SOCKS proxies (https://github.com/akka/akka-http/issues/916), HTTPS proxies (#192), or over SSH tunneled connections.

(Note that, normal HTTP proxies need to be implemented differently as these are implemented on the HTTP layer, not on the transport layer.)

ktoso:
This was solved, please close / assign milestones when done with a PR :)

PR #917, opened 2017-03-01
+htc #916 add pluggable Transport API to switch out client-side transport layer

jrudolph:
For now, this is basically a refactoring of how the client-side graph is put together.

I had this already lying around for a while half-finished. Thought I could just finish it.

Java APIs are still missing.

Refs #916.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
@@ -289,7 +289,7 @@ class HttpExt(private val config: Config)(implicit val system: ActorSystem) exte
                          localAddress: Option[InetSocketAddress] = None,
                          settings:     ClientConnectionSettings  = ClientConnectionSettings(system),
                          log:          LoggingAdapter            = system.log): Flow[HttpRequest, HttpResponse, Future[OutgoingConnection]] =
-    _outgoingConnection(host, port, localAddress, settings, ConnectionContext.noEncryption(), log)
+    _outgoingConnection(host, port, settings, ConnectionContext.noEncryption(), ClientTransport.TCP(localAddress, settings), log)
 
   /**
    * Same as [[#outgoingConnection]] but for encrypted (HTTPS) connections.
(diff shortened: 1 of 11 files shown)


jrudolph:
Resolved conflicts and compile errors.

jrudolph:
PLS BUILD

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
```diff
@@ -305,35 +305,45 @@ class HttpExt(private val config: Config)(implicit val system: ActorSystem) exte
...
+  /**
+   * Similar to `outgoingConnection` but allows to specify a user-defined transport layer to run the connection on.
+   *
+   * Depending on the kind of `ConnectionContext` the implementation will add TLS between the given transport and the HTTP
+   * implementation
+   *
+   * To configure additional settings for requests made using this method,
+   * use the `akka.http.client` config section or pass in a [[akka.http.scaladsl.settings.ClientConnectionSettings]] explicitly.
+   */
+  def outgoingConnectionUsingTransport(
```
ktoso: good

akka-http-core/src/main/scala/akka/http/scaladsl/Transport.scala
```diff
@@ -0,0 +1,40 @@
...
+import akka.stream.scaladsl.{ Flow, Tcp }
+import akka.util.ByteString
+
+import scala.concurrent.Future
+
+/**
+ * Abstraction to allow the creation of alternative transports to run HTTP on.
+ */
+@ApiMayChange
+trait Transport {
```
ktoso: Should we mark it akka private for now as well?

ktoso: If someone were to implement a transport now they could do it in the akka namespace anyway

akka-http-core/src/main/scala/akka/http/scaladsl/Transport.scala
```diff
@@ -0,0 +1,40 @@
...
+  def connectTo(host: String, port: Int)(implicit system: ActorSystem): Flow[ByteString, ByteString, Future[OutgoingConnection]]
+}
+
+object Transport {
+  def TCP(localAddress: Option[InetSocketAddress], settings: ClientConnectionSettings): Transport =
+    new TCPTransport(localAddress, settings)
```
ktoso: ok

akka-http-core/src/main/scala/akka/http/scaladsl/Transport.scala
```diff
@@ -0,0 +1,40 @@
...
+
+  private case class TCPTransport(localAddress: Option[InetSocketAddress], settings: ClientConnectionSettings) extends Transport {
+    def connectTo(host: String, port: Int)(implicit system: ActorSystem): Flow[ByteString, ByteString, Future[OutgoingConnection]] =
```
ktoso: `override def` please

akka-http-core/src/main/scala/akka/http/scaladsl/Transport.scala
ktoso: Hmmm... this is only a **Client** Transport, isn't it?

jrudolph: I don't mind, we can move it to `impl` for now and make it `private[akka]`

jrudolph: On the other hand, then we cannot expose it in `ConnectionPoolSettings` which is the main purpose of the whole thing.

ktoso: Ah good point on exposing - ok let's keep it public then

jrudolph:
Renamed to ClientTransport and added Java API.

Issue #914, opened 2017-03-01
Default HttpApp shuts down immediately in sbt-revolver


shishkin:
The default shutdown signal in `HttpApp`, the `ENTER` key, is apparently getting triggered immediately after start when running in [sbt-revolver](https://github.com/spray/sbt-revolver).

Maybe `HttpApp` should be shutdown on `SIGTERM` by default? Something like this might work:

```scala
override def waitForShutdownSignal(actorSystem: ActorSystem)(implicit executionContext: ExecutionContext): Future[Done] = {
  val promise = Promise[Done]()
  sys.addShutdownHook {
    promise.complete(Success(Done))
  }
  promise.future
}
```

If it does make sense, I could make PR.

jrudolph:
I agree that HttpApp should work properly with sbt-revolver.

The reason it instantly completes is that there is no stdin when run with sbt-revolver. Couldn't we just check if `stdin` has already been closed, i.e. if `readLine` returns `null` and then log a message and keep the ActorSystem alive?

ktoso:
Sounds good - we definitely want to work well with revolver. 
Would you try to detect this and submit a PR @shishkin ?

jlprat:
Bonus point would be that it also works from the sbt console, so dummy demos are still possible

jrudolph:
What's the blocker with the sbt console?

jlprat:
Oh, no, just thinking about the convenience of hitting ENTER and shutting down the server without quitting the console.

shishkin:
> Would you try to detect this and submit a PR @shishkin ?

I will. Thanks for the suggestions everyone, will incorporate them into the PR.

PR #919, opened 2017-03-01
Shutdown HttpApp without STDIN #914

shishkin:
* Prevent immediate shutdown when STDIN is not present, e.g. under sbt-revolver;
* Shutdown the server upon JVM termination;

Couldn't find immediate way of writing a test for the change without introducing indirection of JVM shutdown hook and `StdIn`. Does anyone have any suggestions? I obviously verified the expected behavior in both sbt-revolver (shutdown via C-c) and console (via ENTER).

I will update the `javadsl` implementation as well if this is going in the right direction.

### DIFF
--- a/akka-http/src/main/java/akka/http/javadsl/server/HttpApp.java
+++ b/akka-http/src/main/java/akka/http/javadsl/server/HttpApp.java
@@ -145,15 +145,18 @@ public abstract class HttpApp extends AllDirectives {
    * Hook that lets the user specify the future that will signal the shutdown of the server whenever completed.
    */
   protected CompletionStage<Done> waitForShutdownSignal (ActorSystem system) {
-    return CompletableFuture.supplyAsync( () -> {
+    final CompletableFuture<Done> promise = new CompletableFuture<>();
+    Runtime.getRuntime().addShutdownHook(new Thread(() -> promise.complete(Done.getInstance())));
+    CompletableFuture.runAsync(() -> {
       System.out.println("Press RETURN to stop...");
       try {
-        System.in.read();
+        if (System.in.read() >= 0)
+          promise.complete(Done.getInstance());
       } catch (IOException e) {
         systemReference.get().log().error(e, "Problem occurred! " + e.getMessage());
       }
-      return Done.getInstance();
     });
+    return promise;
   }
 
   /**
--- a/akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
@@ -17,7 +17,7 @@ import akka.http.scaladsl.settings.ServerSettings
 import com.typesafe.config.ConfigFactory
 
 import scala.concurrent.duration.Duration
-import scala.concurrent.{ Await, ExecutionContext, Future }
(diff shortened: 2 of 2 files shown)


typesafehub-validator:
Hi @shishkin,

jlprat:
OK TO TEST

jlprat:
OK TO TEST

jlprat:
OK TO TEST

ktoso:
PLS BUILD

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
```diff
@@ -132,10 +132,15 @@ abstract class HttpApp extends Directives {
...
    */
   protected def waitForShutdownSignal(system: ActorSystem)(implicit ec: ExecutionContext): Future[Done] = {
+    val promise = Promise[Done]()
+    sys.addShutdownHook {
+      promise.success(Done)
+    }
     Future {
-      StdIn.readLine("Press RETURN to stop...\n")
-      Done
+      if (StdIn.readLine("Press RETURN to stop...\n") != null)
```
jlprat: Do you really need the if in this case? `StdIn.readline` is blocking anyway and no matter what it returns we want the promise to complete.

akka-http/src/main/java/akka/http/javadsl/server/HttpApp.java
```diff
@@ -145,15 +145,18 @@ protected void postHttpBindingFailure(Throwable cause) {
...
    */
   protected CompletionStage<Done> waitForShutdownSignal (ActorSystem system) {
-    return CompletableFuture.supplyAsync( () -> {
+    final CompletableFuture<Done> promise = new CompletableFuture<>();
+    Runtime.getRuntime().addShutdownHook(new Thread(() -> promise.complete(Done.getInstance())));
+    CompletableFuture.runAsync(() -> {
       System.out.println("Press RETURN to stop...");
       try {
-        System.in.read();
+        if (System.in.read() >= 0)
```
jlprat: Do you really need the if in this case? `System.in.read` is blocking anyway and no matter what it returns we want the promise to complete.

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
shishkin: In my testing with sbt-revolver is returns `null` when STDIN is not available.

akka-http/src/main/java/akka/http/javadsl/server/HttpApp.java
shishkin: In my testing with sbt-revolver is returns `-1` when STDIN is not available.

jlprat:
I have some minor questions about it, but it looks good! :smiley:

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
jlprat: OK, that makes sense :)

shishkin:
Thanks @jlprat

jrudolph:
I tried a quick test app and it works for me.

LGTM

Issue #192, opened 2016-09-08
Add support for proxying HTTPS server connections on the client side


ktoso:
<a href="https://github.com/sirthias"><img src="https://avatars.githubusercontent.com/u/135347?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [sirthias](https://github.com/sirthias)**
_Monday Oct 27, 2014 at 15:52 GMT_
_Originally opened as https://github.com/akka/akka/issues/16153_

---

Required for one connection:
1. Establish plain-text connection to proxy
2. Send plain-text `CONNECT target.host:443 HTTP/1.1` request to the proxy
3. Wait for `2xx` response (afterwards all communication on the connection will be directly tunneled through to the target host)
4. Start SSL handshake, i.e. upgrade the connection to TLS/SSL
5. Send the actual requests, read actual responses

This feature should likely be implemented on the level of host-level client-side API as in most cases you'll want a managed connection pool to the proxy.

Some pointers:
- http://tools.ietf.org/html/rfc2817#section-5.2 (The other "upgrading a connection" parts of that RFC appear to be rarely used)
- http://stackoverflow.com/questions/6594604/connect-request-to-a-forward-http-proxy-over-an-ssl-connection/6594880#6594880

ktoso:
<a href="https://github.com/sirthias"><img src="https://avatars.githubusercontent.com/u/135347?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [sirthias](https://github.com/sirthias)**
_Monday Oct 27, 2014 at 15:55 GMT_

----

/cc @jrudolph

ktoso:
<a href="https://github.com/mackler"><img src="https://avatars.githubusercontent.com/u/1840214?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [mackler](https://github.com/mackler)**
_Tuesday Oct 28, 2014 at 19:41 GMT_

----

+1

ktoso:
<a href="https://github.com/mlenner"><img src="https://avatars.githubusercontent.com/u/2656905?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [mlenner](https://github.com/mlenner)**
_Monday Nov 03, 2014 at 15:03 GMT_

----

+1

ktoso:
[...]

PR #921, opened 2017-03-02
[PREVIEW] HTTPS Proxy support

jrudolph:
@note, I took the liberty to create a PR from your current state which simplifies giving you a few comments already. You can continue pushing into that one or opening a clean one when you are working on the next version.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/client/ProxyGraphStage.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/client/ProxyGraphStage.scala
@@ -1,0 +1,160 @@
+/*
+ * Copyright (C) 2009-2017 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.http.impl.engine.client
+
+import akka.event.LoggingAdapter
+import akka.http.impl.engine.parsing.HttpMessageParser.StateResult
+import akka.http.impl.engine.parsing.ParserOutput.{ MessageEnd, NeedMoreData, RemainingBytes, ResponseStart }
+import akka.http.impl.engine.parsing.{ HttpHeaderParser, HttpResponseParser, ParserOutput }
+import akka.http.scaladsl.model.{ HttpMethods, StatusCodes }
+import akka.http.scaladsl.settings.ClientConnectionSettings
+import akka.stream.stage.{ GraphStage, GraphStageLogic, InHandler, OutHandler }
+import akka.stream.{ Attributes, BidiShape, Inlet, Outlet }
+import akka.util.ByteString
+
+object ProxyGraphStage {
+  sealed trait State
+  // Entry state
+  case object Starting extends State
+
+  // State after CONNECT messages has been sent to Proxy and before Proxy responded back
+  case object Connecting extends State
+
+  // State after Proxy responded  back
+  case object Connected extends State
+}
+
+final class ProxyGraphStage(targetHostName: String, targetPort: Int, settings: ClientConnectionSettings, log: LoggingAdapter)
+  extends GraphStage[BidiShape[ByteString, ByteString, ByteString, ByteString]] {
+
+  import ProxyGraphStage._
+
+  val bytesIn: Inlet[ByteString] = Inlet("OutgoingTCP.in")
+  val bytesOut: Outlet[ByteString] = Outlet("OutgoingTCP.out")
+
+  val sslIn: Inlet[ByteString] = Inlet("OutgoingSSL.in")
+  val sslOut: Outlet[ByteString] = Outlet("OutgoingSSL.out")
+
+  override def shape: BidiShape[ByteString, ByteString, ByteString, ByteString] = BidiShape.apply(sslIn, bytesOut, bytesIn, sslOut)
+
+  private val connectMsg = ByteString(s"CONNECT ${targetHostName}:${targetPort} HTTP/1.1\r\nHost: ${targetHostName}\r\n\r\n")
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new GraphStageLogic(shape) {
+    private var state: State = Starting
+
+    val parser = new HttpResponseParser(settings.parserSettings, HttpHeaderParser(settings.parserSettings, log)()) {
+      override def handleInformationalResponses = false
+      override protected def parseMessage(input: ByteString, offset: Int): StateResult = {
+        // hacky, we want in the first branch *all fragments* of the first response
+        if (offset == 0) {
+          super.parseMessage(input, offset)
+        } else {
+          if (input.size > offset) {
+            emit(RemainingBytes(input.drop(offset)))
+          } else {
+            emit(NeedMoreData)
+          }
+          terminate()
+        }
+      }
+    }
+    parser.setContextForNextResponse(HttpResponseParser.ResponseContext(HttpMethods.CONNECT, None))
+
+    setHandler(sslIn, new InHandler {
+      override def onPush() = {
+        state match {
+          case Starting ⇒
+            throw new IllegalStateException("inlet OutgoingSSL.in unexpectedly pushed in Starting state")
+          case Connecting ⇒
+            throw new IllegalStateException("inlet OutgoingSSL.in unexpectedly pushed in Connecting state")
+          case Connected ⇒
+            push(bytesOut, grab(sslIn))
+        }
+      }
+
+      override def onUpstreamFinish(): Unit = {
+        complete(bytesOut)
+      }
+
+    })
+
+    setHandler(bytesIn, new InHandler {
+      override def onPush() = {
+        state match {
+          case Starting ⇒
+          // that means that proxy had sent us something even before CONNECT to proxy was sent, therefore we just ignore it
+          case Connecting ⇒
+            val proxyResponse = grab(bytesIn)
+            parser.parseBytes(proxyResponse) match {
+              case NeedMoreData ⇒
+                pull(bytesIn)
+              case ResponseStart(_: StatusCodes.Success, _, _, _, _) ⇒
+                var pushed = false
+                val parseResult = parser.onPull()
+                require(parseResult == ParserOutput.MessageEnd, s"parseResult should be MessageEnd but was $parseResult")
+                parser.onPull() match {
+                  // NeedMoreData is what we emit in overriden `parseMessage` in case input.size == offset
+                  case NeedMoreData ⇒
+                  case RemainingBytes(bytes) ⇒
+                    push(sslOut, bytes) // parser already read more than expected, forward that data directly
+                    pushed = true
+                  case other ⇒
+                    throw new IllegalStateException(s"unexpected element of type ${other.getClass}")
+                }
+                parser.onUpstreamFinish()
+
+                state = Connected
+                if (isAvailable(bytesOut)) {
+                  pull(sslIn)
+                }
+                if (!pushed) {
+                  pull(bytesIn)
+                }
+              case ResponseStart(statusCode, _, _, _, _) ⇒
+                failStage(new ProxyConnectionFailedException(s"The HTTPS proxy rejected to open a connection to $targetHostName:$targetPort with status code: $statusCode"))
+              case other ⇒
+                throw new IllegalStateException(s"unexpected element of type $other")
+            }
+
+          case Connected ⇒
+            push(sslOut, grab(bytesIn))
+        }
+      }
+
+      override def onUpstreamFinish(): Unit = complete(sslOut)
+
+    })
+
+    setHandler(bytesOut, new OutHandler {
+      override def onPull() = {
+        state match {
+          case Starting ⇒
+            push(bytesOut, connectMsg)
+            state = Connecting
+          case Connecting ⇒
+          // don't need to do anything
+          case Connected ⇒
+            pull(sslIn)
+        }
+      }
+
+      override def onDownstreamFinish(): Unit = cancel(sslIn)
+
+    })
+
+    setHandler(sslOut, new OutHandler {
+      override def onPull() = {
+        pull(bytesIn)
+      }
+
+      override def onDownstreamFinish(): Unit = cancel(bytesIn)
+
+    })
+
+  }
+
+}
+
+final case class ProxyConnectionFailedException(msg: String) extends RuntimeException(msg)
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/ClientTransport.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/ClientTransport.scala
@@ -8,9 +8,10 @@ import java.net.InetSocketAddress
 
 import akka.actor.ActorSystem
 import akka.annotation.ApiMayChange
+import akka.http.impl.engine.client.ProxyGraphStage
 import akka.http.scaladsl.Http.OutgoingConnection
-import akka.http.scaladsl.settings.ClientConnectionSettings
-import akka.stream.scaladsl.{ Flow, Tcp }
+import akka.http.scaladsl.settings.{ ClientConnectionSettings }
+import akka.stream.scaladsl.{ BidiFlow, Flow, Keep, Tcp }
 import akka.util.ByteString
 
 import scala.concurrent.Future
@@ -36,5 +37,28 @@ object ClientTransport {
       Tcp().outgoingConnection(InetSocketAddress.createUnresolved(host, port), localAddress,
         settings.socketOptions, halfClose = true, settings.connectingTimeout, settings.idleTimeout)
         .mapMaterializedValue(_.map(tcpConn ⇒ OutgoingConnection(tcpConn.localAddress, tcpConn.remoteAddress))(system.dispatcher))
+  }
+
+  /**
+   * Returns [[ClientTransport]] that performs HTTP CONNECT tunnelling which is useful when you want to use HTTPS proxy.
+   *
+   * Using this kind of [[ClientTransport]] means that after the tunnel between client and target host has been
+   * established the subsequent communication is done on TCP level.
+   *
+   * To get know more about CONNECT tunnelling read https://tools.ietf.org/html/rfc7231#section-4.3.6
+   */
+  def proxy(localAddress: Option[InetSocketAddress], proxyAddress: InetSocketAddress, settings: ClientConnectionSettings): ClientTransport =
+    new ProxyTransport(localAddress, proxyAddress, settings)
+
+  private case class ProxyTransport(localAddress: Option[InetSocketAddress], proxyAddress: InetSocketAddress, settings: ClientConnectionSettings) extends ClientTransport {
+    def connectTo(host: String, port: Int)(implicit system: ActorSystem): Flow[ByteString, ByteString, Future[OutgoingConnection]] = {
+      val networkFlow = Tcp().outgoingConnection(proxyAddress, localAddress,
+        settings.socketOptions, halfClose = true, settings.connectingTimeout, settings.idleTimeout)
+        .mapMaterializedValue(_.map(tcpConn ⇒ OutgoingConnection(tcpConn.localAddress, tcpConn.remoteAddress))(system.dispatcher))
+
+      val proxyBidiFlow = BidiFlow.fromGraph(new ProxyGraphStage(host, port, settings, system.log))
+
+      proxyBidiFlow.joinMat(networkFlow)(Keep.right)
+    }
   }
 }
--- a/akka-http-core/src/test/scala/akka/http/impl/engine/client/ProxyGraphStageSpec.scala
+++ b/akka-http-core/src/test/scala/akka/http/impl/engine/client/ProxyGraphStageSpec.scala
@@ -1,0 +1,175 @@
+/*
+ * Copyright (C) 2009-2017 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.http.impl.engine.client
+
+import akka.http.scaladsl.settings.ClientConnectionSettings
+import akka.stream.scaladsl.{ BidiFlow, Flow, Keep, Sink, Source }
+import akka.stream.testkit.{ TestPublisher, TestSubscriber, Utils }
+import akka.stream.testkit.scaladsl.{ TestSink, TestSource }
+import akka.stream.{ ActorMaterializer, ActorMaterializerSettings }
+import akka.testkit.AkkaSpec
+import akka.util.ByteString
+
+import scala.concurrent.duration._
+
+class ProxyGraphStageSpec extends AkkaSpec {
+  implicit val materializer = ActorMaterializer(ActorMaterializerSettings(system).withFuzzing(true))
+
+  "A ProxyGraphStage" should {
+    "send CONNECT message and then forward incoming messages" in new Context {
+      testCase { (source, flowInProbe, flowOutProbe, sink) ⇒
+        source.sendNext(ByteString("anything"))
+        sink.request(1)
+
+        val connectMsg = flowInProbe.requestNext()
+        connectMsg.utf8String should startWith(s"CONNECT ${targetHostName}:${targetPort} HTTP/1.1")
+        flowOutProbe.sendNext(ByteString("HTTP/1.0 200 Connection established\r\n\r\n"))
+
+        // we deliberately wait here some time to see if ProxyGraphStage properly handles slow demand from Https Proxy
+        flowInProbe.expectNoMsg(300.millis)
+
+        source.sendNext(ByteString("something"))
+        flowInProbe.requestNext(ByteString("anything"))
+        flowOutProbe.sendNext(ByteString("anything"))
+        sink.expectNext(ByteString("anything"))
+
+        flowInProbe.requestNext(ByteString("something"))
+        flowOutProbe.sendNext(ByteString("something"))
+        sink.requestNext(ByteString("something"))
+
+        source.sendComplete()
+        flowOutProbe.sendComplete()
+        sink.expectComplete()
+      }
+    }
+
+    "treat any 2xx response for CONNECT message as successful" in new Context {
+      testCase { (source, flowInProbe, flowOutProbe, sink) ⇒
+        source.sendNext(ByteString("anything"))
+        sink.request(1)
+
+        flowInProbe.requestNext()
+        flowOutProbe.sendNext(ByteString("HTTP/1.0 201 Connection established\r\n\r\n"))
+
+        source.sendComplete()
+        flowInProbe.requestNext(ByteString("anything"))
+        flowOutProbe.sendComplete()
+        sink.expectComplete()
+      }
+    }
+
+    "treat fragmented 200 response for CONNECT message as successful" in new Context {
+      testCase { (source, flowInProbe, flowOutProbe, sink) ⇒
+        source.sendNext(ByteString("anything"))
+        sink.request(100)
+
+        flowInProbe.request(100)
+
+        flowInProbe.expectNext()
+        flowOutProbe.sendNext(ByteString("HTTP/1.0"))
+        flowOutProbe.sendNext(ByteString(" 200"))
+        flowOutProbe.sendNext(ByteString(" Connection established\r\n\r\n"))
+
+        val receivedByteString = flowInProbe.expectNext()
+        flowOutProbe.sendNext(receivedByteString)
+
+        sink.expectNext(ByteString("anything"))
+
+        source.sendComplete()
+        flowOutProbe.sendComplete()
+        sink.expectComplete()
+      }
+    }
+
+    "fail in case of non-2xx Proxy response for CONNECT message" in new Context {
+      testCase { (source, flowInProbe, flowOutProbe, sink) ⇒
+        source.sendNext(ByteString("anything"))
+        sink.request(100)
+
+        flowInProbe.request(100)
+
+        flowOutProbe.sendNext(ByteString("HTTP/1.0 501 Some Error\r\n\r\n"))
+
+        sink.expectError match {
+          case _: ProxyConnectionFailedException ⇒
+          case e ⇒
+            fail(s"should be ProxyConnectionFailedException, caught ${e.getClass.getName} instead")
+        }
+      }
+    }
+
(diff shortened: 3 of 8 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/client/ProxyGraphStage.scala
```diff
@@ -0,0 +1,91 @@
...
+        }
+      }
+    })
+
+    setHandler(bytesOut, new OutHandler {
+      override def onPull() = {
+        // TODO: rethink
+        state match {
+          case Starting ⇒
+            pullIfNeeded(sslIn)
```
jrudolph: Instead of collecting outgoing messages, you shouldn't pull from the user-level layer (`sslIn`) until the connection is established. You can just ignore the pull here and then when the proxy connection has been established, check `isAvailable(bytesOut)` and then `pull(sslIn)`..

akka-http-core/src/main/scala/akka/http/impl/engine/client/ProxyGraphStage.scala
```diff
@@ -0,0 +1,91 @@
...
+      }
+    })
+
+    setHandler(bytesIn, new InHandler {
+      override def onPush() = {
+        state match {
+          case Starting ⇒
+          // should not happen to be in Starting state
+          case Connecting(waitingElem) ⇒
+            val el = grab(bytesIn)
```
jrudolph: Hmm, wouldn't the first HTTP response you get be the `200 OK` from the proxy that would signal that the proxy is now ready? I think we need to invoke the response parser here to make sure we read the first response from the proxy before moving into `Connected` state. If the code seemed to be working the reason might be that the regular HTTP response parser just mistakes the response from the proxy as the response to the first regular outgoing request. After that all responses on this persistent connection will be wrongly associated so that each request will get the response from the previous request.

To get a first working version, instead of actually parsing the proxy response, it might be enough to just ignore everything received until the first `\r\n\r\n`.

jrudolph: If a basic version is working, parsing the proxy response can be done similarly as done in the WebSocket here: https://github.com/akka/akka-http/blob/5932237a86a432d623fafb1e84eeeff56d7485fe/akka-http-core/src/main/scala/akka/http/impl/engine/ws/WebSocketClientBlueprint.scala#L65-L65

akka-http-core/src/main/scala/akka/http/impl/engine/client/ProxyGraphStage.scala
note: 👍

akka-http-core/src/main/scala/akka/http/impl/engine/client/ProxyGraphStage.scala
note: `Hmm, wouldn't the first HTTP response you get be the 200 OK from the proxy that would signal that the proxy is now ready?` - Yes, we expect `200 OK` from the proxy as the first message going back. Because current version is very rough sketch we do not check if response was actually `200 OK` - need to be added. 

On the other hand not sure I understand your first comment completely - note that we are not mistaking proxy response with response to first regular outgoing request - we are not forwarding this method to the original caller - we would do that if we push it to `sslOut`, but the only thing we do is pushing `waitingElem` - namely a SSL's `Client Hello` to `bytesOut`.

Regarding link to WebSocket - thanks will take a look for sure

2beaucoup:
Nice work @note!

refs #192

akka-http-core/src/main/scala/akka/http/impl/engine/client/ProxyGraphStage.scala
jrudolph: > we are not forwarding this response to the original caller

Ok, I might be missing something (and I think I added this comment at the wrong place), so where do you discard the first response?

2beaucoup: It is correct as `el` is currently discarded (not pushed downstream). Just the response code checking is missing.

jrudolph: Ah, I see now. You hope that the `el` you read here contains the complete response from the proxy and then just ignore it. That's likely, indeed, but not guaranteed, so you need to check here at least if you have read the terminating `\r\n\r\n` and otherwise keep reading in the `Connecting` state.

2beaucoup: Feeding the bytes into the response parser like in the WS example @jrudolph posted is the right thing to do here. But manually parsing would work as well for now to get the ball rolling.

note: @jrudolph I see, I indeed used this incorrect assumption that I will read whole message at once, will fix it.

note:
@jrudolph I updated PR so it uses https://github.com/akka/akka-http/pull/917 and adjusted `ProxyGraphStage` according to your comments.

Now there's one problematic thing - how to make it works with HTTP `outgoingConnection`. Technically `CONNECT` tunneling works also for HTTP - once connection has been established Proxy just forwards TCP stream to and from the client. I've tried plain HTTP with current code with Charles Proxy and it works. However I don't think this is what we want to do - as I observed both Chrome and Play-WS don't use `CONNECT` tunneling for plain HTTP. 

How API for end user look like. So currently using proxy for HTTPS looks like following:

```
val proxySettings = ProxySettings("localhost", 8888, List.empty)
val transport = ClientTransport.proxy(None, proxySettings, ClientConnectionSettings(system))

val connectionFlow: Flow[HttpRequest, HttpResponse, Future[Http.OutgoingConnection]] =
	Http().outgoingConnectionUsingTransport("www.scala-lang.org", 443, transport, Http().createDefaultClientHttpsContext())
```

In case we want to create `outgoingConnection` for proxified HTTP we need to change `Http().createDefaultClientHttpsContext()` to `ConnectionContext.noEncryption()`.

If we want to have proxy without CONNECT tunnelling then we need different behavior for HTTP and HTTPS. Here are options: 

1. We can distinguish them by e.g. having `HttpsProxySettings` and `HttpProxySettings` but then end user would be able to call e.g. `outgoingConnectionUsingTransport` with `HttpsProxySettings` and `ConnectionContext.noEncryption()` which is not a valid combination of settings - namely HTTPS proxy and HTTP connectionContext. (It does not matter if we distinguish on `ProxySettings` or `ClientTransport` level - we end up with the same problem).

2. Another way is to have just one `ProxySettings` and just one proxy `ClientTransport` (at least from end user perspective) and in `outgoingConnectionUsingTransport` we have access to both `ProxySettings` and `ConnectionContext` so there we can create whatever internal representation based on those. The disadvantage of this is that we are breaking `ClientTransport` abstraction as we do some additional processing in `HttpExt` based `ClientTransport`.

3. Make HTTP/HTTPS distinction visible on API level, e.g. by adding third parameter to `ClientTransport.connectTo`.

Also, there's probably another problem to solve with HTTP proxy without tunnelling - I think we need to have slightly different `clientLayer` depending on whether we talking through proxy or not. As I saw current `clientLayer` is issuing requests like this: ```GET /some/path HTTP/1.1``` while when going through proxy we need to create them like this: ```GET http://host.name/some/path HTTP/1.1```. I checked and this is what Play-ws and Chrome do (namely - adding protocol+hostname depending on proxy/non-proxy configuration). I guess this would be quite hard to achieve strictly on `ClientTransport` level (on which we operates just on `ByteString`s). Then I guess some code depending on proxy/non-proxy will be needed in `HttpExt`...

What do you think?

michkhol:
Sorry, if I missing something from my brief code scan, but why not have `Http().outgoingConnectionUsingTransport` and `Http().outgoingConnectionHttpsUsingTransport`? From the user's POV it would make it consistent with the "transport-less" calls.

note:
@michkhol I think it's a good idea to have `Http().outgoingConnectionUsingTransport` and `Http().outgoingConnectionHttpsUsingTransport` but still it does not prevent user from calling e.g. `outgoingConnectionHttpsUsingTransport` with `PlainProxyTransport` (`PlainProxyTransport` does not exists yet - it's one that can be used for non-tunneling plain HTTP proxy).

@jrudolph I thought it over and maybe we can use such approach - narrow scope of this PR just to tunneling Proxy. It would fix completely https://github.com/akka/akka-http/issues/192 and also to great extend https://github.com/akka/akka-http/issues/115 - after all it's better to have tunneling Proxy for plain HTTP than not having it at all. Then even after plain HTTP proxy is implemented users would be able to choose between tunnelled one and plain for HTTP.

Still it may be beneficial to have at least rough idea on how we want to implement plain HTTP proxy - just to try to have API in eventual form as soon as possible - AFAIK akka stresses out binary compatibility so it may be hard to undo some implicit assumptions.

jrudolph:
@note:

> If we want to have proxy without CONNECT tunnelling then we need different behavior for HTTP and HTTPS.

I see regular HTTP proxies as a different issue completely. They need to be implemented on a different layer. For HTTPS proxy (or any other transport layer proxy) all the pool infrastructure can stay exactly the same because there's still a one-to-one association between each proxy connection and pool connections. For a HTTP proxy we only have one host pool we use to run all requests over. There's #583 which tries to attempt these things. I agree we need to figure out how to use the API to these cases.

For now, I'm glad we have a solution that is very simple that allows to specify alternative transports for `outgoingConnection` and more importantly for `singleRequest` at all.

@michkhol

> Sorry, if I missing something from my brief code scan, but why not have Http().outgoingConnectionUsingTransport and Http().outgoingConnectionHttpsUsingTransport? From the user's POV it would make it consistent with the "transport-less" calls.

All these APIs are still super low-level. We may introduce more API in the future but right now `outgoingConnectionUsingTransport` has all the necessary functionality to run both HTTP and HTTPS over custom transports.

@note 

> @michkhol I think it's a good idea to have Http().outgoingConnectionUsingTransport and Http().outgoingConnectionHttpsUsingTransport but still it does not prevent user from calling e.g. outgoingConnectionHttpsUsingTransport with PlainProxyTransport (PlainProxyTransport does not exists yet - it's one that can be used for non-tunneling plain HTTP proxy).

See above, there won't be something as a `PlainProxyTransport`, it's not a transport (in the sense of OSI layers which transports bytes). The plain proxy works on the level of HTTP transactions (request - response cycles) where the pool of proxy connections can be shared for any requests regardless of the host they are targetting.

> @jrudolph I thought it over and maybe we can use such approach - narrow scope of this PR just to tunneling Proxy. It would fix completely #192 and also to great extend #115 - after all it's better to have tunneling Proxy for plain HTTP than not having it at all. Then even after plain HTTP proxy is implemented users would be able to choose between tunnelled one and plain for HTTP.

:+1: I don't even want to introduce any more "sugar" to configure proxies or alternative transports e.g. in application.conf. That API-design just needs a bit more work.

akka-http-core/src/main/scala/akka/http/impl/engine/client/ProxyGraphStage.scala
```diff
@@ -0,0 +1,119 @@
...
+          terminate()
+        }
+      }
+    }
+    parser.setContextForNextResponse(HttpResponseParser.ResponseContext(HttpMethods.CONNECT, None))
+
+    setHandler(sslIn, new InHandler {
+      override def onPush() = {
+        state match {
+          case Starting   ⇒
```
jrudolph: Add `throw new IllegalStateExceptions` for all those cases so we know if there's a bug.

akka-http-core/src/main/scala/akka/http/impl/engine/client/ProxyGraphStage.scala
```diff
@@ -0,0 +1,119 @@
...
+          case Starting ⇒
+          // should not happen to be in Starting state
+          case Connecting ⇒
+            val proxyResponse = grab(bytesIn)
+            parser.parseBytes(proxyResponse) match {
+              case NeedMoreData ⇒ pull(bytesIn)
+              case ResponseStart(StatusCodes.OK, _, _, _, _) ⇒
+                state = Connected
+                pull(sslIn)
+                pull(bytesIn)
```
jrudolph: This will probably work because the proxy won't be sending more than the proxy response but in general, if the actual server behind the proxy sends more bytes, the parser might have consumed them. Therefore, just to be sure we should also add that bit of code to make sure that we see all data:

```scala
                    val parseResult = parser.onPull()
                    require(parseResult == ParserOutput.MessageEnd, s"parseResult should be MessageEnd but was $parseResult")
                    parser.onPull() match {
                      case NeedMoreData          ⇒ pull(in) // parser is basically done for our purpose
                      case RemainingBytes(bytes) ⇒ push(out, bytes) // parser already read more than expected, forward that data directly
                      case other ⇒
                        throw new IllegalStateException(s"unexpected element of type ${other.getClass}")
                    }
```

akka-http-core/src/main/scala/akka/http/impl/engine/client/ProxyGraphStage.scala
```diff
@@ -0,0 +1,119 @@
...
+              case ResponseStart(statusCode, _, _, _, _) ⇒
+                throw new IllegalStateException(s"unexpected proxy response status code: ${statusCode.intValue}")
```
jrudolph: This particular exception should use something like `failStage(new ProxyConnectionFailedException("The HTTPS proxy rejected to open a connection to <host>:<port> with $response"))`. This will fail less noisily than just throwing an exception and will give a better error to users.

akka-http-core/src/main/scala/akka/http/impl/engine/client/ProxyGraphStage.scala
```diff
@@ -0,0 +1,119 @@
...
+              case other ⇒
+                throw new IllegalStateException(s"unexpected element of type ${other.getClass}")
```
jrudolph: :+1:

akka-http-core/src/main/scala/akka/http/impl/engine/client/ProxyGraphStage.scala
```diff
@@ -0,0 +1,119 @@
...
+      }
+    })
+
+    setHandler(bytesOut, new OutHandler {
+      override def onPull() = {
+        state match {
+          case Starting ⇒
+            push(bytesOut, connectMsg)
+            state = Connecting
+          case Connecting ⇒
```
jrudolph: We need to make sure to balance pushs and pulls. If we don't note whether we got pulled here and we were not, the `push` in line 67 may fail. One simple fix is to pull in line 83 only `if( isAvailable(bytesOut))`. Might be the same for the other direction.

akka-http-core/src/main/scala/akka/http/impl/engine/client/ProxyGraphStage.scala
ktoso: `getClass.getName` would be nicer though

jrudolph: Make it just `$other`. This shouldn't happen in the first place so we should include as much info as possible when it happens.

akka-http-core/src/main/scala/akka/http/impl/engine/client/ProxyGraphStage.scala
note: Fixed that, and added test in https://github.com/akka/akka-http/pull/921/files#diff-f2ab7fa3256d9d613bdc4d73f5479281R21 - fails with previous version of code

akka-http-core/src/main/scala/akka/http/impl/engine/client/ProxyGraphStage.scala
note: That one is little bit problematic, we're touching a few things here:

1) We are not sending anything until 2xx response come back from server. According to RFC (https://tools.ietf.org/html/rfc2817#section-5.3) we can send content to be sent to target host right after CONNECT. That may result in overall faster response but wanted to start with simple approach. Is it ok?
2) If we stick to current approach to 1) then I see little chance of server sending us anything more than 2xx response and even if that happens I am not sure how to treat this. As we haven't even sent request it's hard to interpret this as a response.
3) It's more like an observation that may be useful if we want to be more diligent regarding parsing here. Actually we are not getting `MessageEnd` at all here. I guess that is because connection is not closed and there's not other mean for Parser to know that the 2xx message has ended. So for the first `pull` we are getting either `NeedMoreData` (for manual testing with Charles proxy or  `flowOutProbe.sendNext(ByteString("HTTP/1.0 200 Connection established\r\n\r\n"))` as in https://github.com/akka/akka-http/pull/921/files#diff-f2ab7fa3256d9d613bdc4d73f5479281R32) or even `EntityPart` if I code proxy response as `flowOutProbe.sendNext(ByteString("HTTP/1.0 200 Connection established\r\n\r\nSomethingMore))"` in test.

note:
Copying here as github has collapsed my response to @jrudolph comment https://github.com/akka/akka-http/pull/921#discussion_r104925836: 

That one is little bit problematic, we're touching a few things here:

1) We are not sending anything until 2xx response comes back from server. According to RFC (https://tools.ietf.org/html/rfc2817#section-5.3) we can send content to be sent to target host right after CONNECT. That may result in overall faster response but wanted to start with simple approach. Is it ok?
2) If we stick to current approach to 1) then I see little chance of server sending us anything more than 2xx response and even if that happens I am not sure how to treat this. As we haven't even sent request it's hard to interpret this as a response.
3) It's more like an observation that may be useful if we want to be more diligent regarding parsing here. Actually we are not getting `MessageEnd` at all here. I guess that is because connection is not closed and there's not other mean for Parser to know that the 2xx message has ended. So for the first pull we are getting either `NeedMoreData` (for manual testing with Charles proxy or `flowOutProbe.sendNext(ByteString("HTTP/1.0 200 Connection established\r\n\r\n"))` as in https://github.com/akka/akka-http/pull/921/files#diff-f2ab7fa3256d9d613bdc4d73f5479281R32) or even `EntityPart` if I code proxy response as `flowOutProbe.sendNext(ByteString("HTTP/1.0 200 Connection established\r\n\r\nSomethingMore))"` in test.

akka-http-core/src/main/scala/akka/http/impl/engine/client/ProxyGraphStage.scala
```diff
@@ -0,0 +1,133 @@
...
+          case Starting ⇒
+            throw new IllegalStateException("inlet OutgoingSSL.in unexpectedly pushed in Starting state")
+          case Connecting ⇒
+            throw new IllegalStateException("inlet OutgoingSSL.in unexpectedly pushed in Connecting state")
+          case Connected ⇒
+            push(bytesOut, grab(sslIn))
+        }
+      }
+
+      override def onUpstreamFinish(): Unit = {
```
note: I've added custom `onDownstreamFinish` and `onUpstreamFinish` to all inlets/outlets to be fully transparent for errors. Thanks to that in case Proxy is down we're getting `akka.stream.StreamTcpException: Tcp command [Connect(localhost:8888,None,List(),Some(10 seconds),true)] failed` instead of cryptic `akka.http.impl.engine.client.OutgoingConnectionBlueprint$UnexpectedConnectionClosureException: The http server closed the connection unexpectedly before delivering responses for 1 outstanding requests`. 

@jrudolph  Here I just want to ensure that it's ok and we will not "leak" this GraphStage because of not calling `completeStage`. I debugged it and seemed like `postStop` is always called (no matter if upstream and downstream just completed or failed) but I lack knowledge to be completely sure.

note:
Hello @jrudolph,

I have two more questions apart from https://github.com/akka/akka-http/pull/921#discussion_r105948312 and https://github.com/akka/akka-http/pull/921#issuecomment-285743616:

1. `ProxySettings` - where to put them? Current version is probably just temporary - probably we should try to structure them as e.g. `ParserSettings`. So they should be taken from config and we should provide appropiate Java API and so on. There's already an entry in `reference.conf` that can be used - this is `akka.http.client.proxy.https`. I know you mentioned in some comment that you don't want any configuration sugar but I think treating `ProxySettings` as any other settings would result in more consistent code and doesn't require much work to achieve it.

2. `ProxySettings.nonProxyHosts` are not currently taken into account. It should be easily doable in `ClientTransport.connectTo` but I am not sure what behavior we want to have for subdomains. We may want to stick to `https.nonProxyHosts` Java System Property - then `*.domain.com` would mean `domain.com with subdomains` while `domain.com` whould mean just `domain.com`. We may also change `ProxySettings.nonProxyHosts` from `Seq[String]` to `Seq[(String, Boolean)]` with Boolean set to true having meaning "with subdomains". Of course instead of tuple we may extract some case class. BTW probably it should be `Set` opposed to `Seq`.

When we have common view on how to approach those issues I will ensure if functionality also works with Java. Will also add some documentation.

Would be grateful for quite quick response. Don't want to push but now I have some spare time I can use for this and don't know how it will be in some time :) Thanks in advance for understanding :)

telser:
Is there any further news on this? Proxy support is pretty important to where I'm working.

jrudolph:
I just didn't get to do another round of reviews (we are just having a planning meeting). I will try to get to it next week. We are also interested to get this merged before the next release because it is a often requested feature.

akka-http-core/src/main/scala/akka/http/impl/engine/client/ProxyGraphStage.scala
jrudolph: > That may result in overall faster response but wanted to start with simple approach. Is it ok?

Yes, I guess that makes sense for now.

> I see little chance of server sending us anything more than 2xx response and even if that happens I am not sure how to treat this.

You are right for HTTP / HTTPS it doesn't make sense. So, I guess it's ok for now (it would need to be fixed if someone is using the proxy for other kinds of traffic or when we do the optimize you mention in 1)

> Actually we are not getting MessageEnd at all here.

I see, we probably would have to change the HttpResponseParser to special case for `CONNECT`. You could try changing [this line](https://github.com/akka/akka-http/blob/master/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpResponseParser.scala#L145) to ignore any content also for `CONNECT` requests. Otherwise, I think the current behavior is fine. We should open up a follow-up ticket to see if we want to improve the behavior.

akka-http-core/src/main/scala/akka/http/impl/engine/client/ProxyGraphStage.scala
jrudolph: Please add `akka.stream.testkit.Utils.assertAllStagesStopped` around all tests to make sure no stages are left over after the test finished.

Transparent handling of completion signals sounds fine.

jrudolph:
I will think about the question of the settings. This is looking good!

jrudolph:
Regarding the settings, I think we should try to separate tasks here. The first and most important thing is to provide the actual transport in this PR. 

I created tickets for the remaining features: 
 * #995 to allow configuration of the transport in the configuration file
 * #996 to allow any kind of per-hosts setting overrides

akka-http-core/src/main/scala/akka/http/impl/engine/client/ProxyGraphStage.scala
note: @jrudolph added

akka-http-core/src/main/scala/akka/http/impl/engine/client/ProxyGraphStage.scala
note: > You could try changing this line to ignore any content also for CONNECT requests

Did that in here: https://github.com/note/akka-http/pull/1/files. As you see it's a little bit hacky. Let me know which version you would prefer to proceed with.

note:
@jrudolph I adjusted PR to your latest remarks. Let me know if there's something more to work on. I assume that #995 and #996 are to be done after this PR is merged

note:
@jrudolph @2beaucoup @ktoso any suggestions how to move forward with merging this? 

If I remember correctly the only open discussion is on https://github.com/akka/akka-http/pull/921#discussion_r104925836 - I prepared a PR to this PR with proposition of change: https://github.com/note/akka-http/pull/1/files. Let me know which version you prefer.

jrudolph:
@note yes, sorry for keeping it hanging a bit.

I'll look at your extra change but I think the functionality itself seems complete now. Only things needed would be a few cleanups. I started doing this last week but was side-tracked. I try to pick it up and maybe send you another PR or a few notes how to proceed later.

ktoso:
I'll review now and let's try to get it in soon.

jrudolph:
@note I think this is fine to go with the additional parser fix you did. I added a small clean in https://github.com/note/akka-http/pull/2 so if you merge that one we have all the changes we need to go. Can you squash all the changes together into a single commit (we cannot squash using Github this time because I opened this PR)?

I have a few improvements that I will propose with a follow up PR once this one is merged.

ktoso:
Code is LGTM from my side as well, the only comments I have are nitpicks which @jrudolph can address in his cleanup followup in which he has accumulated a few of those already he said :)

akka-http-core/src/main/scala/akka/http/impl/engine/client/ProxyGraphStage.scala
```diff
@@ -0,0 +1,133 @@
...
+  @scala.throws[Exception](classOf[Exception])
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new GraphStageLogic(shape) {
+    private var state: State = Starting
+
+    val parser = new HttpResponseParser(settings.parserSettings, HttpHeaderParser(settings.parserSettings, log)()) {
+      override def handleInformationalResponses = false
+    }
+    parser.setContextForNextResponse(HttpResponseParser.ResponseContext(HttpMethods.CONNECT, None))
+
+    setHandler(sslIn, new InHandler {
```
ktoso: I'm not sure if this is cleaner than installing state specific handlers... one would have to try it out I guess.
Technically avoiding all the state matching could be good for perf, though not sure it actually matters - likely other things dominate here

akka-http-core/src/main/scala/akka/http/scaladsl/ClientTransport.scala
```diff
@@ -37,4 +38,19 @@ object ClientTransport {
         settings.socketOptions, halfClose = true, settings.connectingTimeout, settings.idleTimeout)
         .mapMaterializedValue(_.map(tcpConn ⇒ OutgoingConnection(tcpConn.localAddress, tcpConn.remoteAddress))(system.dispatcher))
   }
+
+  def proxy(localAddress: Option[InetSocketAddress], proxySettings: ProxySettings, settings: ClientConnectionSettings): ClientTransport =
```
ktoso: needs docs explaining what it enables, even if simple notes

akka-http-core/src/test/scala/akka/http/impl/engine/client/ProxyGraphStageSpec.scala
```diff
@@ -0,0 +1,150 @@
...
+        val (source, sink) = TestSource.probe[ByteString]
+          .via(flowUnderTest)
+          .toMat(TestSink.probe)(Keep.both)
+          .run()
+
+        fn(source, flowInProbe, flowOutProbe, sink)
+      }
+    }
+  }
+}
```
ktoso: nice tests 👍

akka-http-core/src/main/scala/akka/http/impl/engine/client/ProxyGraphStage.scala
note: When it comes to code clarity from my point of view both options are similar. Regarding performance - you're right. Will give it a try and maybe will submit PR to this PR to have comparison.

note:
@ktoso thanks for comment, resolved all of them (except of https://github.com/akka/akka-http/pull/921#discussion_r115500962 which I want to give a try in a spare moment)

@jrudolph I applied your changes and squashed all changes.

I created a small repo with examples how to use Proxy `ClientTransport`: https://github.com/note/akka-http-client. When creating this I noticed one problem - in javadsl there's no way to pass `ConnectionPoolSettings` (and consequently `ClientTransport`) to `outgoingConnection`. In scaladsl there's no such problem as we have `outgoingConnectionUsingTransport`. For Host- and Request-Level code both Java and Scala works.

akka-http-core/src/main/scala/akka/http/impl/engine/client/ProxyGraphStage.scala
note: Spend a few moments on investigating alternative with installing handlers per state. Here is a PR with rough version of it: https://github.com/note/akka-http/pull/3. It can be refactored to be nicer, but still from my perspective current version is easier to read and comprehend. Therefore I would rather stick to current version. @ktoso WDYT?

note:
@jrudolph regarding problems with javadsl - I added `outgoingConnectionUsingTransport` to javadsl in https://github.com/note/akka-http/pull/4

note:
@jrudolph I think we should add a subsection to all three types of Client-Side API docs. I added such sample for Scala Host-Level Connection in `host-level.md`. Please let me know if the approach is ok and if exact wording makes sense to you. If it's ok I will propagate analogous to other types of Client-Side API and to Java docs too.

Also, should I merge https://github.com/note/akka-http/pull/4 to this PR?

Scalahansolo:
Hey Akka crew. Just wanted to bump this guy and ask if there was any work that could be done by the community to help get this in?

jrudolph:
Superseded by #1200.

jrudolph:
@note thanks for carrying through, I opened up #1200 with a few final touches. I think https://github.com/note/akka-http/pull/4 is worthwhile to do but let's open another PR for that.

jrudolph:
@Scalahansolo we try to get a first version of this out with the next release. Then, testing it in the wild and adding documentation where it is missing would help a lot.

Scalahansolo:
The second Im able to use this I'll be giving feedback for sure!

note:
@Scalahansolo you may take a look at https://github.com/note/akka-http-client - it has code samples ready to run. You just need to build code from branch and replace akka-http version in `build.sbt`

PR #924, opened 2017-03-04
Update versions in FIXMEs and TODOs

jonas:
Update FIXMEs referring to Akka 3.x and Akka HTTP 10.x to refer to Akka
HTTP 11.x.

---

I suggest we also create a 11.x milestone ticket to ensure we remember to remove and rename related methods.

### DIFF
--- a/akka-http-core/src/main/java/akka/http/javadsl/model/HttpEntities.java
+++ b/akka-http-core/src/main/java/akka/http/javadsl/model/HttpEntities.java
@@ -43,7 +43,7 @@ public final class HttpEntities {
     }
 
     /**
-     * @deprecated Will be removed in Akka 3.x, use {@link #create(ContentType, Path)} instead.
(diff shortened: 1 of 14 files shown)


ktoso:
We have a milestone, add issues to it :) https://github.com/akka/akka-http/milestone/15

akka-http-core/src/main/java/akka/http/javadsl/model/HttpEntities.java
```diff
@@ -43,7 +43,7 @@ private HttpEntities() {}
     }
 
     /**
-     * @deprecated Will be removed in Akka 3.x, use {@link #create(ContentType, Path)} instead.
+     * @deprecated Will be removed in Akka HTTP 11.x, use {@link #create(ContentType, Path)} instead.
```
ktoso: yeah good point!

ktoso:
Great catch. I don't think these need a specific ticket btw, going over deprecated and TODO is crucial part of a new Major anyway :)

Issue #574, opened 2016-11-22
Add OSGi manifest headers to bundle


michael-smith:
Both akka and spray support OSGi so it would be nice if the akka-http builds also included the bundle headers.

ktoso:
Would you offer validating once we add the configurations back?
We don't actively use OSGi so help in validating it's set up correctly would be welcome.

michael-smith:
No problem, thank you.

michael-smith:
Is there any update on when this might be available?

collymy:
I would be happy to validate an OSGI version of HTTP as I can't upgrade to Akka Http 10 until I get an OSGI version which means my project is currently stuck at Akka 2.4.11.

ktoso:
Would you be able to contribute some of the work needed to do?
It's currently not a priority for us (we need to focus on http/2 and play support).

collymy:
Hi,
I don't know a great deal about the build system for Akka but I do know quite a bit about the OSGI side of things. If somebody could add the basics to the Akka Http build system I would be happy to clone it and try to produce OSGI versions of the of the jars and submit back a pull request. I would need some hand holding as I have never contributed to a project before but I don't imagine it would take much as I am an experienced Scala and Java developer.
Colly

jrudolph:
The problem is that there currently is no working pipeline that would allow someone to maintain osgi metadata without much hassle. In the past, support for osgi has broken several releases because the sbt plugins deleted files from jar files etc. Unfortunately, so far, trying to support osgi was a net negative because it harmed us more than it helped for the small number of people who cared.

I guess as with every other packaging solution (deb, rpm, ...) there is also a way for osgi to create package for upstream modules without having to include packaging metadata into the upstream source code. This package could then be maintained outside of this module by people who actually care for osgi support and know what needs to be done to make it work. Could that be a solution?

collymy:
I don't know. If it is a solution then why can't Akka Http provide that solution as a separate task under SBT to produce the OSGI bundle.

mpilquist:
In my experience, the cost of maintaining OSGi versions of JARs separately from the main project is prohibitive and error prone. Additionally, it leads to classpath / dependency management problems, requiring exclusions and other nastiness. 

[...]

Issue #230, opened 2016-09-12
Reenable genjavadoc for Access$minusControl$minusAllow$minusOrigin$ in akka-http-core


ktoso:
<a href="https://github.com/jrudolph"><img src="https://avatars.githubusercontent.com/u/9868?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [jrudolph](https://github.com/jrudolph)**
_Thursday Jul 23, 2015 at 07:14 GMT_
_Originally opened as https://github.com/akka/akka/issues/18056_

----

While fixing #15452, there remained a single file with errors for which doc generation will be disabled for now.

It currently fails with these kind of errors:

```
[error] /home/johannes/git/opensource/akka-streams/per-branch-targets/w_15452-fix-gen-scala-doc/akka-http-core-experimental/java/akka/http/scaladsl/model/headers/Access$minusControl$minusAllow$minusOrigin$.java:10: error: <identifier> expected
[error]   public  akka.http.scaladsl.model.headers.Access$minusControl$minusAllow$minusOrigin null () { throw new RuntimeException(); }
[error]                                                                                      ^
```

Here are the generated files and logs:

https://gist.github.com/jrudolph/6147585cf713444aee82

ktoso:
<a href="https://github.com/jrudolph"><img src="https://avatars.githubusercontent.com/u/9868?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [jrudolph](https://github.com/jrudolph)**
_Thursday Jul 23, 2015 at 07:22 GMT_

----

The problem is that there's a symbol with the name "null" in backticks:

```scala
object `Access-Control-Allow-Origin` extends ModeledCompanion[`Access-Control-Allow-Origin`] {
  val `null` = forRange(HttpOriginRange())
  // ...
}
```

How does genjavadoc handle similar issues with backticked symbol names that are keywords in Java?

ktoso:
<a href="https://github.com/jrudolph"><img src="https://avatars.githubusercontent.com/u/9868?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [jrudolph](https://github.com/jrudolph)**
_Thursday Jul 23, 2015 at 07:23 GMT_

----

Btw. this file is part of the scaladsl, so there's not much gained by fixing this.

ktoso:
[...]

PR #925, opened 2017-03-04
Various sbt project fixes and updates

jonas:
- Removes unused plugins
- Fix MiMa error logging
- Bump Paradox and sbt-unidoc to latest version (see generated [unidoc](https://jonas.github.io/akka-http/unidoc) and [javaunidoc](https://jonas.github.io/akka-http/javaunidoc))
- Ignore org.specs2 from scaladoc


jonas:
PLS BUILD ALL

project/Doc.scala
```diff
@@ -139,17 +141,15 @@ object UnidocRoot extends AutoPlugin {
...
 
   override lazy val projectSettings = UnidocRoot.CliOptions.genjavadocEnabled.ifTrue(
-    genjavadocExtraSettings ++ Seq(
+    Seq(
       javacOptions in compile += "-Xdoclint:none",
       javacOptions in test += "-Xdoclint:none",
       javacOptions in doc += "-Xdoclint:none",
       scalacOptions in Compile += "-P:genjavadoc:fabricateParams=true",
-      unidocGenjavadocVersion in Global := "0.10",
-      // FIXME: see https://github.com/akka/akka-http/issues/230
```
2beaucoup: Scould we reference #230 in commit msg or PR?

project/Doc.scala
```diff
@@ -47,7 +49,7 @@ object Scaladoc extends AutoPlugin {
       "-groups",
       "-doc-source-url", urlString,
       "-sourcepath", base.getAbsolutePath,
-      "-skip-packages", "akka.pattern"
+      "-skip-packages", "akka.pattern:org.specs2"
```
2beaucoup: Why is this needed?

project/plugins.sbt
```diff
@@ -12,14 +12,13 @@ resolvers += Resolver.url("2m-sbt-plugin-releases", url("https://dl.bintray.com/
 addSbtPlugin("com.typesafe.sbt" % "sbt-multi-jvm" % "0.3.8")
 addSbtPlugin("com.typesafe" % "sbt-mima-plugin" % "0.1.14")
 addSbtPlugin("org.scalariform" % "sbt-scalariform" % "1.6.0")
-addSbtPlugin("com.lightbend.paradox" % "sbt-paradox" % "0.2.8")
+addSbtPlugin("com.lightbend.paradox" % "sbt-paradox" % "0.2.9")
 addSbtPlugin("com.dwijnand" % "sbt-dynver" % "1.1.1")
-addSbtPlugin("com.typesafe.sbt" % "sbt-osgi" % "0.9.1")
-addSbtPlugin("com.eed3si9n" % "sbt-unidoc" % "0.3.3")
+//addSbtPlugin("com.typesafe.sbt" % "sbt-osgi" % "0.9.1")
```
2beaucoup: If we agree to not support OSGI then let's just remove it.

project/Doc.scala
2beaucoup: Ouch! Could you please add a comment that includes https://issues.scala-lang.org/browse/SI-10028? Thanks!

project/Doc.scala
jonas: This doesn't fix that issue, it merely allows to remove one of the two filters that was added as a work around/

See https://github.com/akka/akka-http/blob/b825f3e9f2495d6d1562c8b504a3a07ab375782a/project/Doc.scala#L125

project/plugins.sbt
jonas: I would be happy to, but #574 seems to still be alive. My motivation behind commenting this out is that we don't pay the cost of loading that plugin, however still leave `git grep -i osgi` helpful in case someone wants to add OGSI support.

ktoso: I don't think we can just ignore questions for osgi, even though I'd love to.
Don't remove the plugin yet please

project/Doc.scala
jonas: Good idea, done.

project/Doc.scala
2beaucoup: I see.

project/plugins.sbt
jonas: OK, I've removed any surgery related to OSGI.

project/MiMa.scala
```diff
@@ -92,9 +92,7 @@ object MiMa extends AutoPlugin {
 
     if (failures.isEmpty) mappings.map(_.right.get).toMap
     else {
-      // TODO: actually report errors using a logger, can only be done once keys have been converted to
-      // taskKeys, see https://github.com/typesafehub/migration-manager/pull/157
-      failures.flatMap(_.left.get).foreach(ex => println(ex.getMessage))
+      failures.flatMap(_.left.get).foreach(ex => logger.error(ex.getMessage))
```
jrudolph: Thanks, good one.

jrudolph:
PLS BUILD

project/Doc.scala
```diff
@@ -47,7 +49,8 @@ object Scaladoc extends AutoPlugin {
       "-groups",
       "-doc-source-url", urlString,
       "-sourcepath", base.getAbsolutePath,
-      "-skip-packages", "akka.pattern"
+      // Workaround https://issues.scala-lang.org/browse/SI-10028
+      "-skip-packages", "akka.pattern:org.specs2"
```
jrudolph: Thanks, I guess next time it will be a completely different one but one can you do...

jrudolph:
PLS BUILD

PR #928, opened 2017-03-05
Add more patterns to .gitignore

jlprat:
Add patterns to .gitignore to ignore:
* ensime config files
* emacs backup files
* vim backup files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,10 @@
 target
 
 *.orig
+
+*#
+*~
+.#*
+.*.swp
+*.vim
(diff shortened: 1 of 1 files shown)


ktoso:
Ok sure, i dont mind

jrudolph:
I don't mind about that one. Btw. there are other ways to ignore things either per user or per checkout, e.g. by putting ignorings into `.git/info/exclude` or `~/.gitignore`.

jlprat:
Good to know about those other alternatives :smile:

PR #929, opened 2017-03-06
Minor janitorial fixes and improvements

jonas:


akka-http-core/src/test/scala/akka/http/javadsl/JavaInitializationSpec.scala
```diff
@@ -12,7 +12,7 @@ class JavaInitializationSpec extends WordSpec with Matchers {
 
     "initializes the right field" in {
       akka.http.scaladsl.model.headers.LanguageRange.`*` // first we touch the scala one, it should force init the Java one
-      akka.http.javadsl.model.headers.LanguageRange.ALL // touching this one should not fail
+      val _ = akka.http.javadsl.model.headers.LanguageRange.ALL // touching this one should not fail
```
jlprat: What is the purpose of this "discarded" val?

jonas: From my understanding it's a regression test for some tricky code which ensures correct initialization of static fields in the Java API.

Blame points to this ticket: https://github.com/akka/akka/issues/20403

akka-http-core/src/main/scala/akka/http/impl/util/LogByteStringTools.scala
```diff
@@ -47,7 +47,7 @@ private[akka] object LogByteStringTools {
 
   def logTlsInbound(name: String, maxBytes: Int = MaxBytesPrinted): Flow[SslTlsInbound, SslTlsInbound, NotUsed] =
     Flow[SslTlsInbound].log(name, {
-      case s @ SessionTruncated         ⇒ s
+      case s: SessionTruncated          ⇒ s
```
ktoso: ok

akka-http-core/src/test/scala/akka/http/javadsl/JavaInitializationSpec.scala
ktoso: I guess scalac would panic that there's a pure statement here and it takes no effect?
Wouldn't we need to do the same for the other values?

jonas: I'll validate tomorrow, but I agree that this is surprising.

jonas: OK, I've looked into this and managed to reproduce a NPE with this patch -https://gist.github.com/jonas/6ccec6a28a105c89e3793b9f410b71ef
based on the test case provided here https://github.com/akka/akka/issues/20403#issuecomment-215356663.

I will create a ticket with these findings.

jonas: Created #936

akka-parsing/src/main/scala/akka/macros/LogHelper.scala
```diff
@@ -22,9 +22,9 @@ private[akka] trait LogHelper {
   /** Override to prefix every log message with a user-defined context string */
   protected def prefixString: String = ""
 
-  def debug(msg: String) = macro LogHelper.debugMacro
-  def info(msg: String) = macro LogHelper.infoMacro
-  def warning(msg: String) = macro LogHelper.warningMacro
+  def debug(msg: String): Unit = macro LogHelper.debugMacro
```
jrudolph: Thanks for cleaning up after me.

Issue #930, opened 2017-03-06
Remove usage of IteratorInterpreter in MultipartUnmarshaller


jrudolph:
This will allow us to finally remove the IteratorInterpreter in akka-stream.

It should be easy to remove it by parsing the strict data asynchronously.

/cc @johanandren

ktoso:
This was solved, unlocks using new materializer

PR #931, opened 2017-03-06
=htp #930 replace IteratorInterpreter usage by asynchronous folding

jrudolph:
Refs #930.

### DIFF
--- a/akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/MultipartUnmarshallers.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/MultipartUnmarshallers.scala
@@ -4,22 +4,20 @@
 
 package akka.http.scaladsl.unmarshalling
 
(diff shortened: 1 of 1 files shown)


viktorklang:
@jrudolph Cool. Do we have any results for the performance impact of the change? :)

ktoso:
@viktorklang this is to unlock using akka-http with the new materializer, perofrmance impact we'll consider separately. This week we need it to work on the new one and look at performance then.

I'd vote to merge to unlock progress

viktorklang:
Absolutely

-- 
Cheers,
√

jrudolph:
Let's not merge this before getting at least some rough numbers about how much slower it will be. I didn't know before that `IteratorInterpreter` would be easy enough to maintain so maybe we can keep it.

johanandren:
We should make it public api if we keep it though, depending on internals is not nice.

jrudolph:
This has been included in #958.

PR #940, opened 2017-03-08
Revert "Upgrade sbt-unidoc to version 0.4.0"

jrudolph:
This change prevents us from using our integrated build with akka/akka. We need to make sure to make this change on both `master` and `release-2.4` in akka at the same time.

So let's revert for now.

This reverts commit 6a1257c58addf86c97e012dd2e45e24821669c37.

/cc @jonas


jrudolph:
PLS BUILD

jonas:
Ah, OK. I thought the nightly integration build was using snapshots.

project/ValidatePullRequest.scala
```diff
@@ -158,7 +158,7 @@ object ValidatePullRequest extends AutoPlugin {
           .filter(l =>
             l.startsWith("akka-") ||
               l.startsWith("docs") ||
-              (l.startsWith("project") && l != "project/MiMa.scala") ||
+              (l.startsWith("project") && l != "project/MiMa.scala") ||
```
jlprat: I think we can get rid of this trailing space here

jrudolph:
Fixed the trailing whitespace.

PR #947, opened 2017-03-08
Change `Unmarshal#to` to use a default context #78

athieriot:
Attempt to fix issue #78

As per the corresponding Gitter discussion, this PR will make ```Unmarshall#to``` use the execution context provided by the Materializer if none is provided implicitly.

### DIFF
--- a/akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/Unmarshal.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/Unmarshal.scala
@@ -16,5 +16,9 @@ class Unmarshal[A](val value: A) {
   /**
    * Unmarshals the value to the given Type using the in-scope Unmarshaller.
    */
-  def to[B](implicit um: Unmarshaller[A, B], ec: ExecutionContext, mat: Materializer): Future[B] = um(value)
+  def to[B](implicit um: Unmarshaller[A, B], ec: ExecutionContext = null, mat: Materializer): Future[B] = {
+    val context: ExecutionContext = if (ec == null) mat.executionContext else ec
+
+    um(value)(context, mat)
+  }
 }
--- a/akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/Unmarshal.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/Unmarshal.scala
@@ -15,6 +15,9 @@ object Unmarshal {
 class Unmarshal[A](val value: A) {
   /**
    * Unmarshals the value to the given Type using the in-scope Unmarshaller.
+   
+   * Uses the default materializer [[ExecutionContext]] if no implicit execution context is provided. 
+   * If you expect the marshalling to be heavy, it is suggested to provide a specialized context for those operations.
    */
   def to[B](implicit um: Unmarshaller[A, B], ec: ExecutionContext = null, mat: Materializer): Future[B] = {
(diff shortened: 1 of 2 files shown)


ktoso:
OK TO TEST

docs/src/test/scala/docs/http/scaladsl/UnmarshalSpec.scala
```diff
@@ -28,4 +28,21 @@ class UnmarshalSpec extends AkkaSpec {
     //#use-unmarshal
   }
 
+  "use unmarshal without execution context" in {
+    //#use-unmarshal
```
2beaucoup: should this also be included in the docs?

athieriot: Ah yeah I wasn't quite sure what those comments were used for. Do they mean that the test is added automatically to the documentation?

I would tend to think a full example is not necessary.
However, would like me to add a line in the Unmarshall documentation where we can suggest the executionContext  is optional? Or should the Scala doc be enough?

2beaucoup: > Ah yeah I wasn't quite sure what those comments were used for. Do they mean that the test is added automatically to the documentation?

Yes. The result is [here](http://doc.akka.io/docs/akka-http/current/scala/http/common/unmarshalling.html#using-unmarshallers) ( [source](https://github.com/akka/akka-http/blob/master/docs/src/main/paradox/scala/http/common/unmarshalling.md#using-unmarshallers)).

> I would tend to think a full example is not necessary.
> However, would like me to add a line in the Unmarshall documentation where we can suggest the executionContext is optional? Or should the Scala doc be enough?

A note in the ScalaDoc would be appreciated.

athieriot:
@2beaucoup Updated the PR with a small addition to the documentation

akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/Unmarshal.scala
```diff
@@ -16,5 +16,9 @@ class Unmarshal[A](val value: A) {
   /**
    * Unmarshals the value to the given Type using the in-scope Unmarshaller.
    */
-  def to[B](implicit um: Unmarshaller[A, B], ec: ExecutionContext, mat: Materializer): Future[B] = um(value)
+  def to[B](implicit um: Unmarshaller[A, B], ec: ExecutionContext = null, mat: Materializer): Future[B] = {
```
ktoso: Please add to the scaladoc:

```
Uses the default materializer [[ExecutionContext]] if no implicit execution context is provided. 
If you expect the marshalling to be heavy, it is suggested to provide a specialized context for those operations.
```

ktoso:
Excellent, thanks!

athieriot:
Updated. Thank you for giving me the text directly ;)

I can't squash the commits right now but could do tonight if need to

ktoso:
We can squash, thanks a lot! :-)

jlprat:
Just throwing it here, WDYT about overloading the Java's `Unmarshaller#unmarshal` with an alternative without an `ExecutionContext`?

ktoso:
Yes, good idea I think, using Java marshallers is a pain already we can help a bit with that.
Would you PR that @jlprat ?

jlprat:
Yup! I'll create an issue and assign it to me to not forget it :)

jlprat:
For reference https://github.com/akka/akka-http/issues/967

PR #948, opened 2017-03-09
=htc #945 in StreamUtils.delayCancellation cancel timeout if stage was stopped otherwise

### LINKED ISSUE #945: DelayCancellationStage doesn't cancel timeout when it stops
This will lead to HTTP graph structures being kept referenced in memory longer than necessary (i.e. until the `lingerTimeout` has fired).

jrudolph:
Before, the timeout kept referencing the stage which kept the whole
GraphInterpreter and indirectly the ActorGraphInterpreter referenced. The memory
was only freed after the timeout had triggered.

Fixes #945.

### DIFF
--- a/akka-http-core/src/main/mima-filters/10.0.4.backwards.excludes
+++ b/akka-http-core/src/main/mima-filters/10.0.4.backwards.excludes
@@ -12,3 +12,7 @@ ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.scaladsl.setting
 ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.impl.settings.ConnectionPoolSettingsImpl.apply")
 ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.impl.settings.ConnectionPoolSettingsImpl.copy")
 ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.impl.util.JavaMapping.flowMapping")
+
+# changes to internal classes, #945
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.impl.util.StreamUtils#ScheduleSupport.scheduleOnce")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.impl.util.StreamUtils#ScheduleSupport.scheduleOnce")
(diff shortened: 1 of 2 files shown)


akka-http-core/src/main/mima-filters/10.0.4.backwards.excludes
```diff
@@ -12,3 +12,7 @@ ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.scaladsl.setting
 ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.impl.settings.ConnectionPoolSettingsImpl.apply")
 ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.impl.settings.ConnectionPoolSettingsImpl.copy")
 ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.impl.util.JavaMapping.flowMapping")
+
+# changes to internal classes, #945
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.impl.util.StreamUtils#ScheduleSupport.scheduleOnce")
```
jonas: Would it be possible to exclude the whole `akka.http.impl` instead?

ktoso: AFAIR we could, I don't mind knowing what we touch in there though

akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
```diff
@@ -249,12 +250,16 @@ private[http] object StreamUtils {
...
+
       override def onDownstreamFinish(): Unit = {
         cancelAfter match {
           case finite: FiniteDuration ⇒
-            scheduleOnce(finite) {
-              log.debug(s"Stage was canceled after delay of $cancelAfter")
-              completeStage()
+            timeout = OptionVal.Some {
+              scheduleOnce(finite) {
+                log.debug(s"Stage was canceled after delay of $cancelAfter")
```
ktoso: Prefer the {} syntax

```
log.debug(s"Stage was canceled after delay of {}", cancelAfter)
```

Issue #946, opened 2017-03-08
Wrong formatting in Docs


jlprat:
Bot Scala and Java docs for the `entity` directive have a wrong formatted table.
Link to pages:
* http://doc.akka.io/docs/akka-http/10.0.4/scala/http/routing-dsl/directives/marshalling-directives/entity.html
* http://doc.akka.io/docs/akka-http/10.0.4/java/http/routing-dsl/directives/marshalling-directives/entity.html

PR #949, opened 2017-03-09
fix entity directive doc format

zhxiaogg:
ref #946


jlprat:
OK TO TEST

jlprat:
Thanks @zhxiaogg !!

Issue #957, opened 2017-03-13
Make akka-http compile against both 2.4 and master


jrudolph:
We used several internal stream APIs that are gone in 2.5.

 * `SubSource.kill` (see #956)
 * `IteratorInterpreter`
 * `Fusing.aggressive`
 * `copyFromPorts`

jrudolph:
Fixed by #958.

Issue #956, opened 2017-03-13
Replace usages of `SubSource.kill`


jrudolph:
In 2.5 it is gone and it fails now even for Akka 2.4:

https://jenkins.akka.io:8498/job/akka-http-nightly-against-akka-release-2.4/100//consoleFull

E.g. this stack trace:

```
[info]   - with one element and name value in quotes *** FAILED ***
[info]     java.lang.UnsupportedOperationException: cannot drop Source of type akka.stream.impl.fusing.GraphStageModule
[info]     at akka.stream.impl.fusing.SubSource$.kill(StreamOfStreams.scala:652)
[info]     at akka.http.impl.engine.parsing.BodyPartParser$$anon$1$$anonfun$parseEntity$default$4$1$$anonfun$apply$3.apply(BodyPartParser.scala:217)
[info]     at akka.http.impl.engine.parsing.BodyPartParser$$anon$1$$anonfun$parseEntity$default$4$1$$anonfun$apply$3.apply(BodyPartParser.scala:216)
[info]     at akka.http.scaladsl.unmarshalling.MultipartUnmarshallers$$anonfun$multipartUnmarshaller$1$$anonfun$apply$1$$anonfun$apply$2$$anonfun$apply$3.apply(MultipartUnmarshallers.scala:89)
[info]     at akka.http.scaladsl.unmarshalling.MultipartUnmarshallers$$anonfun$multipartUnmarshaller$1$$anonfun$apply$1$$anonfun$apply$2$$anonfun$apply$3.apply(MultipartUnmarshallers.scala:87)
[info]     at scala.collection.Iterator$class.foreach(Iterator.scala:893)
[info]     at akka.stream.impl.fusing.IteratorInterpreter$IteratorDownstream.foreach(IteratorInterpreter.scala:43)
[info]     at akka.http.scaladsl.unmarshalling.MultipartUnmarshallers$$anonfun$multipartUnmarshaller$1$$anonfun$apply$1$$anonfun$apply$2.apply(MultipartUnmarshallers.scala:87)
[info]     at akka.http.scaladsl.unmarshalling.MultipartUnmarshallers$$anonfun$multipartUnmarshaller$1$$anonfun$apply$1$$anonfun$apply$2.apply(MultipartUnmarshallers.scala:71)
[info]     at akka.http.scaladsl.unmarshalling.Unmarshaller$$anon$1.apply(Unmarshaller.scala:58)
[info]     ...
```

PR #958, opened 2017-03-13
Changes need to make akka-http compile with both 2.4 and 2.5

jrudolph:
Refs #956 and #957.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolConductor.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolConductor.scala
@@ -35,7 +35,10 @@ private object PoolConductor {
         slotEventIn.carbonCopy(),
         slotOuts.map(_.carbonCopy()))
 
-    override def copyFromPorts(inlets: immutable.Seq[Inlet[_]], outlets: immutable.Seq[Outlet[_]]): Shape =
+    /**
+     * Needs to be implemented for 2.4. Not needed in 2.5 any more.
+     */
+    /* override */ def copyFromPorts(inlets: immutable.Seq[Inlet[_]], outlets: immutable.Seq[Outlet[_]]): Shape =
       Ports(
         inlets.head.asInstanceOf[Inlet[RequestContext]],
         inlets.last.asInstanceOf[Inlet[RawSlotEvent]],
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/BodyPartParser.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/BodyPartParser.scala
@@ -214,7 +214,7 @@ private[http] final class BodyPartParser(
                       emitFinalPartChunk: (List[HttpHeader], ContentType, ByteString) ⇒ Unit = {
                         (headers, ct, bytes) ⇒
                           emit(BodyPartStart(headers, { rest ⇒
-                            SubSource.kill(rest)
+                            StreamUtils.cancelSource(rest)(materializer)
                             HttpEntity.Strict(ct, bytes)
                           }))
                       })(input: ByteString, offset: Int): StateResult =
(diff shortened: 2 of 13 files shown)


jrudolph:
Removing `Fusing.aggressive` calls lead to a 15% performance hit using `ab` without persistent connections (3500 connections per second down to 3000 connections per second).

I therefore removed that change and replaced it with a change that uses reflection to call `Fusing.aggressive` in 2.4 to keep the old performance.

jrudolph:
FWIW with the same code on 2.5 it's 7000 connections per second.

ktoso:
> FWIW with the same code on 2.5 it's 7000 connections per second.

awww yiiis

jrudolph:
PLS BUILD

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolConductor.scala
```diff
@@ -35,7 +35,10 @@ private object PoolConductor {
         slotEventIn.carbonCopy(),
         slotOuts.map(_.carbonCopy()))
 
-    override def copyFromPorts(inlets: immutable.Seq[Inlet[_]], outlets: immutable.Seq[Outlet[_]]): Shape =
+    /**
+     * Needs to be implemented for 2.4. Not needed in 2.5 any more.
+     */
+    /* override */ def copyFromPorts(inlets: immutable.Seq[Inlet[_]], outlets: immutable.Seq[Outlet[_]]): Shape =
```
ktoso: good, thanks for the comment

akka-http-core/src/main/scala/akka/http/impl/engine/parsing/BodyPartParser.scala
```diff
@@ -214,7 +214,7 @@ private[http] final class BodyPartParser(
                       emitFinalPartChunk: (List[HttpHeader], ContentType, ByteString) ⇒ Unit = {
                         (headers, ct, bytes) ⇒
                           emit(BodyPartStart(headers, { rest ⇒
-                            SubSource.kill(rest)
+                            StreamUtils.cancelSource(rest)(materializer)
```
ktoso: ok good

akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
```diff
@@ -303,6 +224,52 @@ private[http] object StreamUtils {
...
+    case EmptySource ⇒ // nothing to do with empty source
+    case x ⇒
+      val mat =
+        GraphInterpreter.currentInterpreterOrNull match {
+          case null if materializer ne null ⇒ materializer
+          case null                         ⇒ throw new IllegalStateException("Need to pass materializer to cancelSource if not run from GraphInterpreter context.")
+          case x                            ⇒ x.subFusingMaterializer // try to use fuse if already running in interpreter context
+        }
+      x.runWith(Sink.ignore)(mat)
+  }
```
ktoso: good

ktoso: I wonder if my weird materializer could be used here actually once it's out there...

akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
```diff
@@ -303,6 +224,52 @@ private[http] object StreamUtils {
...
+
+      new Fuser {
+        def aggressive[S <: Shape, M](g: Graph[S, M]): Graph[S, M] = instance.aggressive(g)
+      }
+    } catch {
+      case ex: ClassNotFoundException ⇒
+        new Fuser {
+          def aggressive[S <: Shape, M](g: Graph[S, M]): Graph[S, M] = g // no fusing
+        }
+    }
```
ktoso: ok good, allows us to work with both nicely for now

akka-http-core/src/test/scala/akka/stream/testkit/Utils.scala
```diff
@@ -16,6 +16,11 @@ object Utils {
 
   case class TE(message: String) extends RuntimeException(message) with NoStackTrace
 
+  type ActorMaterializerImpl = {
```
ktoso: add a comment that this is for 2.4 compat?

Issue #963, opened 2017-03-14
Content negotiation does not take into account media type parameters


raboof:
https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.1 describes how content negotiation should work when responding to requests with an `Accept` header. The current implementation doesn't appear to take into account media type parameters.

jrudolph:
Hey @raboof, good to have you here ;)

Thanks for tackling this. IIRC it was mentioned somewhere before but I cannot seem to find the place.

ktoso:
Can we close this as solved? The linked PR had some minor questions but seems to fix this

jrudolph:
I think we just forget to close.

PR #964, opened 2017-03-14
Use media type parameters in content negotiation (#963)

raboof:
https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.1 describes how to interpret the Accept header during content negotiation.

Notably, it specifies "The media type quality factor associated with a given type is determined by finding the media range with the highest precedence which matches that type", and then gives some examples.

'matching' is implemented in akka-http in `MediaRange::matches`, but for `MediaRange.One` fails to take into account parameters.

'precedence' is implemented in akka-http when determining the sorting of `acceptedMediaRanges` in `MediaTypeNegotiator`, but for non-wildcard media types sorts by the qValue instead of by the number of parameters.

### DIFF
--- a/akka-http-tests/src/test/scala/akka/http/scaladsl/marshalling/ContentNegotiationSpec.scala
+++ b/akka-http-tests/src/test/scala/akka/http/scaladsl/marshalling/ContentNegotiationSpec.scala
@@ -147,9 +147,14 @@ class ContentNegotiationSpec extends FreeSpec with Matchers {
       accept(`text/html` withParams Map("level" → "1"), `text/html` withParams Map("format" → "flowed")) should select(`text/html` withParams Map("level" → "1") withCharset `UTF-8`)
     }
 
-    /** when both acceptable media types are equally specific, we must select the one with the highest qValue */
+    /** when both acceptable media types are available as separate marshallers and equally specific, we must select the one with the highest qValue */
     "Accept: text/html;level=1;q=0.4, text/html;format=flowed;q=0.7" test { accept ⇒
       accept(`text/html` withParams Map("level" → "1"), `text/html` withParams Map("format" → "flowed")) should select(`text/html` withParams Map("format" → "flowed") withCharset `UTF-8`)
+    }
+
+    /** when there's 2 equally specific matching media types for the available marshaller, we should choose the one with the highest qValue */
+    "Accept: text/html;level=1;q=0.4, text/html;format=flowed;q=1, text/html;q=0.7" test { accept ⇒
+      accept(`text/html` withParams Map("level" → "1", "format" → "flowed"), `text/html`) should select(`text/html` withParams Map("level" → "1", "format" → "flowed") withCharset `UTF-8`)
     }
 
     // https://tools.ietf.org/html/rfc7231#section-5.3.2
--- a/akka-http-tests/src/test/scala/akka/http/scaladsl/marshalling/ContentNegotiationSpec.scala
+++ b/akka-http-tests/src/test/scala/akka/http/scaladsl/marshalling/ContentNegotiationSpec.scala
@@ -6,6 +6,7 @@ package akka.http.scaladsl.marshalling
 
 import scala.concurrent.Await
 import scala.concurrent.duration._
+import akka.http.scaladsl.server.MediaTypeNegotiator
 import akka.http.scaladsl.server.ContentNegotiator.Alternative
 import akka.util.ByteString
 import org.scalatest.{ Matchers, FreeSpec }
@@ -136,6 +137,20 @@ class ContentNegotiationSpec extends FreeSpec with Matchers {
       accept(`text/plain`, `text/xml`) should select(`text/plain` withCharset `UTF-8`)
       accept(`text/xml`, `text/plain`) should select(`text/xml` withCharset `UTF-8`)
     }
+
+    // https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.1
+    "Correctly calculate qValues" in {
+      val negotiator = new MediaTypeNegotiator(parseHeaders(
+        "Accept: text/*;q=0.3, text/html;q=0.7, text/html;level=1, text/html;level=2;q=0.4, */*;q=0.5"
+      ))
+      negotiator.qValueFor(`text/html` withParams Map("level" → "1")) should be(1.0f)
+      negotiator.qValueFor(`text/html`) should be(0.7f)
+      negotiator.qValueFor(`text/plain`) should be(0.3f)
+
+      negotiator.qValueFor(`image/jpeg`) should be(0.5f)
+      negotiator.qValueFor(`text/html` withParams Map("level" → "2")) should be(0.4f)
+      negotiator.qValueFor(`text/html` withParams Map("level" → "3")) should be(0.7f)
+    }
   }
 
   def testHeaders[U](headers: HttpHeader*)(body: ((Alternative*) ⇒ Option[ContentType]) ⇒ U): U = {
@@ -161,21 +176,22 @@ class ContentNegotiationSpec extends FreeSpec with Matchers {
   def reject = equal(None)
   def select(contentType: ContentType) = equal(Some(contentType))
 
+  def parseHeaders(example: String): List[HttpHeader] = example.stripMarginWithNewline("\n").split('\n').toList map { rawHeader ⇒
+    val Array(name, value) = rawHeader.split(':')
+    HttpHeader.parse(name.trim, value) match {
+      case HttpHeader.ParsingResult.Ok(header, Nil) ⇒ header
+      case result                                   ⇒ fail(result.errors.head.formatPretty)
+    }
+  }
+
   implicit class AddStringToIn(example: String) {
     def test(body: ((Alternative*) ⇒ Option[ContentType]) ⇒ Unit): Unit = example in {
       val headers =
         if (example != "(without headers)") {
-          example.stripMarginWithNewline("\n").split('\n').toList map { rawHeader ⇒
-            val Array(name, value) = rawHeader.split(':')
-            HttpHeader.parse(name.trim, value) match {
-              case HttpHeader.ParsingResult.Ok(header, Nil) ⇒ header
-              case result                                   ⇒ fail(result.errors.head.formatPretty)
-            }
-          }
+          parseHeaders(example)
         } else Nil
 
       testHeaders(headers: _*)(accept ⇒ body(accept))
     }
   }
 }
-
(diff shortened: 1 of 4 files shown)


ktoso:
Refs #963

ktoso:
PLS WHITELIST

ktoso:
OK TO TEST

akka-http/src/main/scala/akka/http/scaladsl/server/ContentNegotation.scala
```diff
@@ -22,9 +22,9 @@ final class MediaTypeNegotiator(requestHeaders: Seq[HttpHeader]) {
       Accept(mediaRanges) ← requestHeaders
       range ← mediaRanges
     } yield range).sortBy { // `sortBy` is stable, i.e. upholds the original order on identical keys
```
2beaucoup: Could you please remove the comment (which is a remnant of the previous implementation logic) and update the ScalaDoc?

akka-http-tests/src/test/scala/akka/http/scaladsl/marshalling/ContentNegotiationSpec.scala
```diff
@@ -136,6 +137,20 @@ class ContentNegotiationSpec extends FreeSpec with Matchers {
       accept(`text/plain`, `text/xml`) should select(`text/plain` withCharset `UTF-8`)
       accept(`text/xml`, `text/plain`) should select(`text/xml` withCharset `UTF-8`)
     }
+
+    // https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.1
```
jrudolph: The currently valid RFC for HTTP/1.1 is this one: https://tools.ietf.org/html/rfc7231#section-5.3.2. Could you update the link and cross-check that everything is still valid?

akka-http/src/main/scala/akka/http/scaladsl/server/ContentNegotation.scala
```diff
@@ -22,9 +22,9 @@ final class MediaTypeNegotiator(requestHeaders: Seq[HttpHeader]) {
...
-      case x if x.isWildcard     ⇒ 2f // most general, needs to come last
-      case MediaRange.One(_, qv) ⇒ -qv // most specific, needs to come first
-      case _                     ⇒ 1f // simple range like `image/*`
+      case x if x.isWildcard   ⇒ 2f // most general, needs to come last
+      case one: MediaRange.One ⇒ -one.params.size // most specific, needs to come first
```
2beaucoup: 👍  `ContentNegotiator.pickContentType` [explicitly sorts by qVal](https://github.com/akka/akka-http/pull/964/files#diff-3be11dbf4c6d44ec7f6610db0bd4d589L120).

akka-http/src/main/scala/akka/http/scaladsl/server/ContentNegotation.scala
raboof: Actually mulling this over a bit I think we should still take the `qValue` into account here, so the comments are actually correct and my change was wrong :).

This is a bit of a cornercase of a cornercase, but getting this right makes the solution easier to understand in the end IMO :). Added a testcase and fix in 7b3c4a4d22ad4e310805f74caf62b1d4e62f5624 but could use another pair of eyes :).

akka-http-tests/src/test/scala/akka/http/scaladsl/marshalling/ContentNegotiationSpec.scala
```diff
@@ -136,6 +136,40 @@ class ContentNegotiationSpec extends FreeSpec with Matchers {
       accept(`text/plain`, `text/xml`) should select(`text/plain` withCharset `UTF-8`)
       accept(`text/xml`, `text/plain`) should select(`text/xml` withCharset `UTF-8`)
     }
+
+    /** when both acceptable media types are equally specific and have the same qValue, we choose to send the first */
+    "Accept: text/html;level=1, text/html;format=flowed" test { accept ⇒
+      accept(`text/html` withParams Map("level" → "1"), `text/html` withParams Map("format" → "flowed")) should select(`text/html` withParams Map("level" → "1") withCharset `UTF-8`)
+    }
+
+    /** when an acceptable media type is more specific, it is matched even when it has a lower qValue */
```
jrudolph: Is that behavior specified somewhere? In particular, why is `text/html;format=flowed` less specific than `text/html`?

akka-http-tests/src/test/scala/akka/http/scaladsl/marshalling/ContentNegotiationSpec.scala
```diff
@@ -136,6 +136,40 @@ class ContentNegotiationSpec extends FreeSpec with Matchers {
...
+    "Accept: text/html;format=flowed;q=0.5, text/html;q=1" test { accept ⇒
+      accept(`text/html` withParams Map("level" → "1"), `text/html` withParams Map("format" → "flowed")) should select(`text/html` withParams Map("level" → "1") withCharset `UTF-8`)
+    }
+
+    /** when both acceptable media types are available as separate marshallers and equally specific, we must select the one with the highest qValue */
```
jrudolph: Hmm, here as well, why would the exact matching content type be equally specific than one which is missing a parameter?

akka-http-tests/src/test/scala/akka/http/scaladsl/marshalling/ContentNegotiationSpec.scala
```diff
@@ -136,6 +136,40 @@ class ContentNegotiationSpec extends FreeSpec with Matchers {
...
+    "Accept: text/html;level=1;q=0.4, text/html;format=flowed;q=0.7" test { accept ⇒
+      accept(`text/html` withParams Map("level" → "1"), `text/html` withParams Map("format" → "flowed")) should select(`text/html` withParams Map("format" → "flowed") withCharset `UTF-8`)
+    }
+
+    /** when there's 2 equally specific matching media types for the available marshaller, we should choose the one with the highest qValue */
```
jrudolph: Here on the other hand, it seems somewhat clear that all of the candidates are equally unspecific.

akka-http-tests/src/test/scala/akka/http/scaladsl/marshalling/ContentNegotiationSpec.scala
raboof: `text/html;format=flowed` is more specific than `text/html` because it specifies (restricts) more parameters.

In this example, we have 2 marshallers available, one with media type `text/html;level=1` and one with media type `text/html;format=flowed`. For each one we calculate the qValue based on the preferences in the `Accept` header. 

The `text/html;level=1` marshalling does not match `text/html;format=flowed` because it is missing the `format` parameter, but does match `text/html`, so it gets the associated qValue of 1. 

The `text/html;format=flowed` marshalling matches both `text/html;format=flowed` and `text/html`. Because `text/html;format=flowed` is more specific, this one is selected, and a qValue of 0.5 is assigned to this marshalling.

Now that we have assigned the qValues, we select the marshalling with the largest qValue, which is `text/html;level=1`.

This is my interpretation of https://tools.ietf.org/html/rfc7231#section-5.3.2 , though I agree it's not too easy to read.

akka-http-tests/src/test/scala/akka/http/scaladsl/marshalling/ContentNegotiationSpec.scala
raboof: Ah I agree this comment is confusing, 'equally specific' doesn't have anything to do with the choice here because each marshaller only matches 1 of the types from the Accept header, so a better comment would be simply "when both acceptable media types are available as separate marshallers, we must select the one with the highest qValue"

akka-http-tests/src/test/scala/akka/http/scaladsl/marshalling/ContentNegotiationSpec.scala
jrudolph: Or maybe just the description is wrong?

Reading through the description in https://tools.ietf.org/html/rfc7231#section-5.3.2, precedence for the `Accept` header would be this (most specific first):

 * `text/html;format=flowed;q=0.5`
 * `text/html;q=1`

Then for the available content-type for this resource the table would look like this:

| Content Type | Matched Range | qValue |
| - | - | - |
| `text/html;level=1` | `text/html;q=1` | `1` |
| `text/html;format=flowed` |  `text/html;format=flowed;q=0.5` | `0.5` |

So, the result makes sense here (but not the comment because in this case the less specific range was chosen).

akka-http-tests/src/test/scala/akka/http/scaladsl/marshalling/ContentNegotiationSpec.scala
jrudolph: Ok, so this example is correct as well (I misread the example). Both ranges in the accept clause are equally specific.

akka-http-tests/src/test/scala/akka/http/scaladsl/marshalling/ContentNegotiationSpec.scala
```diff
@@ -136,6 +136,40 @@ class ContentNegotiationSpec extends FreeSpec with Matchers {
...
+    "Accept: text/html;level=1;q=0.4, text/html;format=flowed;q=1, text/html;q=0.7" test { accept ⇒
+      accept(`text/html` withParams Map("level" → "1", "format" → "flowed"), `text/html`) should select(`text/html` withParams Map("level" → "1", "format" → "flowed") withCharset `UTF-8`)
+    }
+
+    // https://tools.ietf.org/html/rfc7231#section-5.3.2
```
jrudolph: Cool. Good to have the example from the spec here.

akka-http-tests/src/test/scala/akka/http/scaladsl/marshalling/ContentNegotiationSpec.scala
raboof: Perhaps indeed the comment is confusing. What I meant there was: because `text/html;format=flowed;q=0.5` is more specific, it is selected for `text/html;format=flowed`, causing this marshaller *not* to be selected for the end-result since `text/html;level=1` matches a higher qValue.

I really like that table, perhaps we should see whether we can find a way to write these tests so they look like that :).

akka-http/src/main/scala/akka/http/scaladsl/server/ContentNegotation.scala
```diff
@@ -21,10 +21,10 @@ final class MediaTypeNegotiator(requestHeaders: Seq[HttpHeader]) {
     (for {
       Accept(mediaRanges) ← requestHeaders
       range ← mediaRanges
-    } yield range).sortBy { // `sortBy` is stable, i.e. upholds the original order on identical keys
-      case x if x.isWildcard     ⇒ 2f // most general, needs to come last
-      case MediaRange.One(_, qv) ⇒ -qv // most specific, needs to come first
-      case _                     ⇒ 1f // simple range like `image/*`
+    } yield range).sortBy { // `sortBy` is stable, i.e. upholds the original order on identical weights
+      case x if x.isWildcard   ⇒ 2f // most general, needs to come last
```
jrudolph: What if that has params as well? Should that go into the specificity calculation?

raboof:
@jrudolph actually extending this to the wildcard types too was not too bad - ready to squash and merge?

jrudolph:
Yes, please go ahead.

Issue #440, opened 2016-10-26
X-Forwarded-For includes the port


derekwyatt:
`X-Forwarded-For` contains a sequence of `RemoteAddress` objects, but they contain the port and the render doesn't seem to take this into account.  If I create this on the client side:

``` scala
HttpRequest(headers = List(`X-Forwarded-For`(addr)))
```

Then I see this when the same request is seen by an Akka server:

```
Illegal request header: Illegal 'x-forwarded-for' header: Invalid input ':', expected HEXDIG, h8, listSep or 'EOI' (line 1, column 16): 0:0:0:0:0:0:0:1:60052
```

jrudolph:
Thanks, @derekwyatt. Does that only happen with IPv6 addresses?

derekwyatt:
@jrudolph It's both.  I added a couple of tests to verify it, but fixing it is a bit of problem.  I can easily change the `RemoteAddress` instances just-in-time for rendering but the test fails with:
```
Ok(X-Forwarded-For: 1:2:3:4:5:6:7:8,List()) did not equal Ok(X-Forwarded-For: 1:2:3:4:5:6:7:8,List()) (HttpHeaderSpec.scala:693)
```
Changing them prior to construction would be better, but that's tricky due to the duplicate signatures.  e.g.
``` scala
def apply(addrs: immutable.Seq[RemoteAddress]): `X-Forwarded-For` = // transform here
...
final case class `X-Forwarded-For`(addresses: immutable.Seq[RemoteAddress]) 
```
I could create `RemoteIPAddress` to make a truly type-safe way of doing it, but that might be overkill.

Thoughts?

jonas:
I played around with this and the bug also affects `X-Real-Ip`.

@jrudolph What is the reason for failing tests if the round-trip fails? (i.e. `parsedHeader != origHeader`)?

If this constraint cannot be lifted we need to make the main constructor private for those headers so the port can be removed from `RemoteAddress.IP` before passed to `X-Forwarded-For` or `X-Real-Ip`. Not sure what options are available if we want to ensure bincompat.

jrudolph:
> @jrudolph What is the reason for failing tests if the round-trip fails? (i.e. parsedHeader != origHeader)?

WDYM? The round-trip is required so that we test both the parsing and the rendering. You can turn off checking the rendering if that's not what you want.

jonas:
The easiest way to fix X-Forwarded-For and X-Real-Ip is to change their
[...]

PR #974, opened 2017-03-19
=htc #440 Exclude port when rendering X-Forwarded-For and X-Real-Ip

jonas:
Refs #440.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/RemoteAddress.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/RemoteAddress.scala
@@ -56,6 +56,13 @@ object RemoteAddress {
     try IP(InetAddress.getByAddress(bytes)) catch { case _: UnknownHostException ⇒ Unknown }
   }
 
+  private[akka] val renderWithoutPort = new Renderer[RemoteAddress] {
+    def render[R <: Rendering](r: R, address: RemoteAddress): r.type = address match {
+      case IP(ip, _) ⇒ r ~~ ip.getHostAddress
+      case _         ⇒ r ~~ address
+    }
+  }
+
   JavaInitialization.initializeStaticFieldWith(
     Unknown, classOf[jm.RemoteAddress].getField("UNKNOWN"))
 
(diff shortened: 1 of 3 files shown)


akka-http-core/src/main/scala/akka/http/scaladsl/model/RemoteAddress.scala
```diff
@@ -56,6 +56,13 @@ object RemoteAddress {
     try IP(InetAddress.getByAddress(bytes)) catch { case _: UnknownHostException ⇒ Unknown }
   }
 
+  private[akka] val renderWithoutPort = new Renderer[RemoteAddress] {
+    def render[R <: Rendering](r: R, address: RemoteAddress): r.type = address match {
+      case IP(ip, _) ⇒ r ~~ ip.getHostAddress
+      case _         ⇒ r ~~ address
```
jlprat: Nice alignment!

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
```diff
@@ -953,6 +953,7 @@ final case class `WWW-Authenticate`(challenges: immutable.Seq[HttpChallenge]) ex
 // http://en.wikipedia.org/wiki/X-Forwarded-For
 object `X-Forwarded-For` extends ModeledCompanion[`X-Forwarded-For`] {
   def apply(first: RemoteAddress, more: RemoteAddress*): `X-Forwarded-For` = apply(immutable.Seq(first +: more: _*))
+  implicit val addressRenderer = RemoteAddress.renderWithoutPort
```
jlprat: Naming is complicated, but maybe some other name not so easy to be confused with the other implicit?

jonas: What about `singleAddressRenderer` or `remoteAddressRenderer`?

jlprat: I think I prefer `singleAddressRenderer` :smiley:

jonas: OK, done

jonas:
PLS BUILD

Issue #936, opened 2017-03-07
NPE when acessing scaladsl.model.headers.HttpOriginRange.`*`


jonas:
While testing #929 I looked into akka/akka@2aa3d7bb1dc9099d00243047b6cdf4597fd570a0 and found that it is still possible to reproduce the java initialization issue using the test case given in https://github.com/akka/akka/issues/20403#issuecomment-215356663

Here's a test version:
```scala
class JavaInitializationSpec extends WordSpec with Matchers {

  implicit class HeaderCheck[T](self: T) {
    def =!=(expected: String) = self.toString shouldBe expected
  }

  "LanguageRange" should {

    "initializes the right field" in {
      // Uncommenting the following line fixes the NPE
      // akka.http.javadsl.model.headers.HttpOriginRange.ALL =!= "*"
      akka.http.scaladsl.model.headers.HttpOriginRange.`*` =!= "*"
    }
  }
}
```

Based on https://gist.github.com/jonas/6ccec6a28a105c89e3793b9f410b71ef

jrudolph:
Thanks, great reproducer.

PR #978, opened 2017-03-19
=htc #936 Fix NPE when access static Java constant fields

jonas:
The workaround added in 34bd3d9f1b50ac133896b9e73b47d9408a2bd033 to
ensure static fields in Java model classes inherited from Scala models
are initialized properly put the runtime patching on the enclosing Scala
object, however, it also changed the Java classes to redirect static
fields to the new API, which meant that the initialization hook was
never called.

To fix this problem move the initialization to the Scala object
referenced from the Java static field.

---

Note, it is possible to further clean up use of `JavaInitialization` and for example remove it from `HttpEntity.Empty` since it is defined as a `val` instead of `object`.

### DIFF
--- a/akka-http-core/src/main/java/akka/http/javadsl/model/RemoteAddress.java
+++ b/akka-http-core/src/main/java/akka/http/javadsl/model/RemoteAddress.java
@@ -37,5 +37,5 @@ public abstract class RemoteAddress {
      *             Will be removed in Akka HTTP 11.x, use {@link RemoteAddresses#UNKNOWN} instead.
      */
     @Deprecated
-    public static final RemoteAddress UNKNOWN = RemoteAddresses.UNKNOWN;
+    public static final RemoteAddress UNKNOWN = akka.http.scaladsl.model.RemoteAddress.Unknown$.MODULE$;
 }
--- a/akka-http-core/src/main/java/akka/http/javadsl/model/headers/EntityTagRange.java
+++ b/akka-http-core/src/main/java/akka/http/javadsl/model/headers/EntityTagRange.java
@@ -20,5 +20,5 @@ public abstract class EntityTagRange {
      *             Will be removed in Akka HTTP 11.x, use {@link EntityTagRanges#ALL} instead.
      */
     @Deprecated
-    public static final EntityTagRange ALL = EntityTagRanges.ALL;
+    public static final EntityTagRange ALL = akka.http.scaladsl.model.headers.EntityTagRange.$times$.MODULE$;
 }
--- a/akka-http-core/src/main/java/akka/http/javadsl/model/headers/HttpEncodingRange.java
+++ b/akka-http-core/src/main/java/akka/http/javadsl/model/headers/HttpEncodingRange.java
@@ -25,5 +25,5 @@ public abstract class HttpEncodingRange {
      *             Will be removed in Akka HTTP 11.x, use {@link HttpEncodingRanges#ALL} instead.
      */
     @Deprecated
-    public static final HttpEncodingRange ALL = HttpEncodingRanges.ALL;
+    public static final HttpEncodingRange ALL = akka.http.scaladsl.model.headers.HttpEncodingRange.$times$.MODULE$;
 }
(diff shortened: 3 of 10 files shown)


jonas:
Hmm, sbt.ResolceError https://jenkins.akka.io:8498/job/pr-validator-akka-http/884/console

jonas:
PLS BUILD

jonas:
PLS BUILD

jonas:
PLS BUILD

jrudolph:
@jonas have you completely understood what the issue is? It seems previous attempts to fix this have never completely succeeded. This [comment](https://github.com/akka/akka/issues/19162#issuecomment-195228014) seemed to give the best explanation. Do you agree with it? Was it just an oversight that it wasn't fixed completely the last two times? Would be nice to document your findings somewhere.

jonas:
The referenced comment seem correct and it seems like having the deprecated Java static field reference the new Java static field didn't cause the initialization code to be called in some cases and in other cases the initialization code was put in an enclosing Scala object which didn't get initialized when accessing a child member.

There were two cases working (that is, causing the JavaInitializer code to be called): `HttpEntity.Empty` and `LanguageRange.*`, the first is a `val` (and doesn't actually need the JavaInitializer code as hinted above) the other is an `object` extending a `case class`. `LanguageRange.*` is similar to `HttpEncodingRange.*` in that both are `object`s extending a `case class` however `HttpEncodingRange` is defined as a `sealed abstract class` whereas `LanguageRange` is defined as a `sealed trait`. This might explain why `HttpEncodingRange.*` failed.

The remaining 3 cases are all `case object` which might explain why the JavaInitializer was not called when accessing the enclosing object.

The main part of the fix here is to make the Java static field reference the Scala static field (rather than the non-deprecated version of the Java static field) to trigger the initialization of the scala object and call to the JavaInitializer code. That is why I moved it inside the defining scala object since initialization of the enclosing object is not guaranteed when accessing an enclosed scala object.

I can muse a bit more about it in the commit message if you'd like.

jrudolph:
Thanks for the explanation. One problem is that we cannot even properly test the behavior in a simple fashion because the classes will only be loaded and initialized once, so we cannot check that all variations of accessing one first and then the other work. So, we always run the risk of just flipping the buggy behavior from one side to the other. Also, when running the test suite, we don't know if the fields have been initialized even before the test, so it's hard to say if the tests even do anything. Do we have an idea how to deal with that?

jrudolph:
But I understand how the fix works. 

And maybe I was wrong about the tests because it seems it can only go wrong in one direction: if the Scala one is accessed first and we always delegate to the scala one, then only the scala side `MODULE$` can still be `null` when the Java side static initializer is running.

Crazy stuff, also that particular minor changes in the exact class / trait / object setup will change the behavior so much. Thanks a lot for fixing that, @jonas.

jrudolph:
> The referenced comment seems correct and it seems that having the deprecated Java static field reference the new Java static field didn't cause the initialization code to be called in some cases and in other cases the initialization code was put in an enclosing Scala object which didn't get initialized when accessing a child member.

Btw, the explanation was wrong I think. Here's what I think happens:

When e.g. `scaladsl.model.headers.EntityTagRange.*` is accessed:

 * The access to `scaladsl.model.headers.EntityTagRange.*` is compiled to reading `scaladsl.model.headers.EntityTagRange$$times$.MODULE$` which is a static field of `scaladsl.model.headers.EntityTagRange$$times$`.
 * Therefore, this class is loaded and its static initializer should be run.
 * However, before the static initializer of `scaladsl.model.headers.EntityTagRange$$times$` is run, static initializers for all the superclasses are run.
 * Therefore the static initializer for `javadsl.model.header.EntityTagRange` is run.
 * This static initializer is also responsible for setting `ALL = scaladsl.model.headers.EntityTagRange$$times$.MODULE$`. This is where the issue was: at that point the class has been loaded but the static initializer of `scaladsl.model.headers.EntityTagRange$$times$` just hasn't been run so the static `MODULE$` field is still `null`. (This is basically a cyclic initialization of static fields. If the values would have been modelled with `lazy val` instead it would actually result in a stack overflow.)
 * The static initializer for scaladsl.mdoel.header.EntityTagRange` is run
 * Finally, the static initializer for `scaladsl.model.headers.EntityTagRange$$times$`  is run
 * This calls just `new scaladsl.model.headers.EntityTagRange$$times$()`, i.e. the instance constructor
 * The instance constructor calls the super constructors and then sets `MODULE$` to `this`. 
 * The instance constructor runs the remaining initialization as given in the body of the `object`.
 * (with your fix): part of the object initialization is `JavaInitialization.initializeStaticFieldWith` is called which accesses again `MODULE$` which already contains the instance and puts it into the static field of the super class.

jrudolph:
Refs #936.

jonas:
Looking forward to removing all those hacks for version 11

Issue #768, opened 2017-01-14
UnexpectedConnectionClosureException cannot be catched since it's in private class


sjoerdmulder:
Akka HTTP version 10.0.1

Somehow we have received a 
```
Caused by: akka.http.impl.engine.client.OutgoingConnectionBlueprint$UnexpectedConnectionClosureException: The http server closed the connection unexpectedly before delivering responses for 1 outstanding requests
	at akka.http.impl.engine.client.OutgoingConnectionBlueprint$$anonfun$apply$2.apply(OutgoingConnectionBlueprint.scala:127)
	at akka.http.impl.engine.client.OutgoingConnectionBlueprint$$anonfun$apply$2.apply(OutgoingConnectionBlueprint.scala:127)
	at akka.http.impl.util.One2OneBidiFlow$One2OneBidi$$anon$1$$anon$4.onUpstreamFinish(One2OneBidiFlow.scala:95)
	at akka.stream.impl.fusing.GraphInterpreter.processEvent(GraphInterpreter.scala:732)
	at akka.stream.impl.fusing.GraphInterpreter.execute(GraphInterpreter.scala:616)
```

But we should beable to `.recover` the future right? So i thought so let's retry the request if that happens:

```
.recoverWith {
      // The http server closed the connection unexpectedly before delivering responses
      case e: UnexpectedConnectionClosureException if retries > 0 =>
        logger.warn(s"Unexpected connection closure, retries left: $retries", e)
        //Retry the request with 1 less retry
        executeRequest(request, retries - 1)
```

But i'm not allowed to get to the exception since it's inside a private class. 

**Request**: make sure that all exception that can be thrown outside to the system are part of the public API.

```
Error:(5, 37) object OutgoingConnectionBlueprint in package client cannot be accessed in package akka.http.impl.engine.client
import akka.http.impl.engine.client.OutgoingConnectionBlueprint.UnexpectedConnectionClosureException
```

ktoso:
Would you submit a PR with the change? Thanks a lot.

jrudolph:
We do not expose all internal exceptions. That doesn't mean that you cannot get at the exception message, it just means that you cannot rely on a particular exception type being thrown in a certain situation. This is standard practice for exception, i.e. the JDK doesn't offer all kinds of different `IOException` for every single potential failure condition. It's hard enough to keep the API compatible in the happy cases, we don't want to be bound to be exception-compatible as well.

[...]

PR #979, opened 2017-03-19
Make UnexpectedConnectionClosureException public #768

jlprat:
Issue: #768
Make `UnexpectedConnectionClosureException` accessible for final users

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/ErrorInfo.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/ErrorInfo.scala
@@ -91,3 +91,6 @@ final case class EntityStreamSizeException(limit: Long, actualSize: Option[Long]
 }
 
 case class RequestTimeoutException(request: HttpRequest, message: String) extends RuntimeException(message)
+
+case class UnexpectedConnectionClosureException(outstandingResponses: Int)
+  extends RuntimeException(s"The http server closed the connection unexpectedly before delivering responses for $outstandingResponses outstanding requests")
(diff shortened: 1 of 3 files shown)


jlprat:
I tried to add a test to check this case, but with no luck.
What I did was:
* Start a server with own system
** Server terminates its actor system (and waits for termination) when serving its route
* Client requests route
* Expected to get the exception

I'm open to ideas on how to reproduce it.

akka-http-core/src/main/scala/akka/http/scaladsl/model/ErrorInfo.scala
```diff
@@ -91,3 +91,6 @@ final case class EntityStreamSizeException(limit: Long, actualSize: Option[Long]
 }
 
 case class RequestTimeoutException(request: HttpRequest, message: String) extends RuntimeException(message)
+
+case class UnexpectedConnectionClosureException(outstandingResponses: Int)
```
jlprat: Can make this exception extend `ExceptionWithErrorInfo` as well

jlprat:
Failed because of `sbt.ResolveException: unresolved dependency: com.typesafe.akka#akka-http_2.11;10.0.5: not found`

jlprat:
PLS BUILD

jlprat:
PLS BUILD

jlprat:
PLS BUILD

jrudolph:
Hmm, I'm actually not so sure about the underlying issue. You can handle exceptions without needing access to them. Where would be start and where would we end to expose exceptions?

jlprat:
IMO if the exception reaches the client unwrapped, it should be public. Users might want to  recover differently according to the exception received.

Internal exceptions could still be thrown as long as they are wrapped inside a public one.

jrudolph:
I don't think I agree. In Scala code we don't use exceptions to transport semantics so exceptions shouldn't be matched on. We could think about creating a public superclass for all of our exceptions but why would it help?

jlprat:
Maybe we can head over the issue and discuss there if we want something else instead.

Right now the Http client `Future` is failed with that private exception, one can always recover from any exception, without caring about the type obviously. Question if that's the way we want the users to write their code.

I agree, usually exceptions are not thrown in Scala, however, this is one of the few cases where one is thrown.

jlprat:
I guess this PR makes not much sense now (see discussion on issue: #768

jrudolph:
Closing this one. See the discussion in #768,

Issue #641, opened 2016-12-11
Java `complete` directive(s) should take arguments of type `ResponseEntity` instead of `RequestEntity`


hseeberger:
This extends to `Marshaller[T, RequestEntity]`.

Notice that a `RequestEntity` is a `ResponseEntity` and not the other way round!

With the current implementation, it's not possible to write a marshaller which marshals to `CloseDelimited`, because that's nole a `ResponseEntity` and not a `RequestEntity`.

jrudolph:
Looks like an oversight to me, left over from the Java port.

btomala:
If no one working on it I will take a look on it. Looks pretty simple.

jlprat:
Just wondering, isn't this change a potential breaker for users of those classes? Basically, is it binary and source compatible?

btomala:
@jlprat It is source compatible but isn't binary. And it is very simple change. Do you want this change even though?

jlprat:
Yes, the ticket is up to be picked. It would be cool if you could provide a PR! :smiley:

jrudolph:
Fixed in #982.

Issue #642, opened 2016-12-11
Construct `ToResponseMarshaller` from `Marshaller[A, ResponseEntity]`


hseeberger:
Currently only `Marshaller[A, MessageEntity]` is supported, which is `Marshaller[A, RequestEntity]`, which won't for work for all `ResponseEntity`s (namely `CloseDelimited`).

Related 641.

jrudolph:
Which call do you mean exactly? One of those in `javadsl.marshalling.Marshaller`?

hseeberger:
Sorry for the terse description. I'm talking about the Scala API and the implicit construction of marshallers. Currently, as long as you have an implicit `Marshaller[A, MessageEntity]`, an implicit `ToResponseMarshaller` is constructed and the request can be completed. But that's to restrictive, because `type MessageEntity = RequestEntity`, i.e. only works for response entities which can also be request entities, i.e. not for `CloseDelimited`.

jrudolph:
Aha, so it seems an even lower priority `EvenLowerPriorityToResponseMarshallerImplicits.liftToResponseEntityMarshaller` is needed for that case.

PR #982, opened 2017-03-20
Complete directives take `ResponseEntity` as a parameter instead of `RequestEntity`

btomala:
Issue: 641
`complete` directives take `ResponseEntity` as a parameter instead of `RequestEntity`

I think in next steps `Marshaller[T, RequestEntity]` should be replaced by `Marshaller[T, ResponseEntity]`. But it requires #642 first.

### DIFF
--- a/akka-http/src/main/scala/akka/http/javadsl/server/directives/RouteDirectives.scala
+++ b/akka-http/src/main/scala/akka/http/javadsl/server/directives/RouteDirectives.scala
@@ -17,6 +17,7 @@ import akka.http.impl.model.JavaUri
 import akka.http.javadsl.model.HttpHeader
 import akka.http.javadsl.model.HttpResponse
 import akka.http.javadsl.model.RequestEntity
+import akka.http.javadsl.model.ResponseEntity
 import akka.http.javadsl.model.StatusCode
 import akka.http.javadsl.model.Uri
 import akka.http.javadsl.server.{ RoutingJavaMapping, Rejection, Route }
@@ -112,7 +113,7 @@ abstract class RouteDirectives extends RespondWithDirectives {
   /**
    * Completes the request using the given status code, headers, and response entity.
    */
-  def complete(status: StatusCode, headers: java.lang.Iterable[HttpHeader], entity: RequestEntity) = RouteAdapter {
+  def complete(status: StatusCode, headers: java.lang.Iterable[HttpHeader], entity: ResponseEntity) = RouteAdapter {
     D.complete(scaladsl.model.HttpResponse(status = status.asScala, entity = entity.asScala, headers = Util.immutableSeq(headers).map(_.asScala))) // TODO avoid the map()
   }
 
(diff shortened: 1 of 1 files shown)


jlprat:
OK TO TEST

jrudolph:
I guess this change isn't binary compatible. We might have to introduce new overloads and deprecate the old ones for now.

jlprat:
Could we maybe add a test using the new overloaded `complete` methods?

btomala:
I could but I think it doesn't make sense. Because there are test for old API and old API use new one, so it is already tested. Even more when depreciated functions will be remove then test will be use new API. Because we have source compatibility.

jlprat:
Oh yeah sorry, I forgot that `RequestEntity` extends `ResponseEntity`!

btomala:
Thanks :)

akka-http/src/main/scala/akka/http/javadsl/server/directives/RouteDirectives.scala
```diff
@@ -112,11 +113,18 @@ abstract class RouteDirectives extends RespondWithDirectives {
...
    */
-  def complete(status: StatusCode, headers: java.lang.Iterable[HttpHeader], entity: RequestEntity) = RouteAdapter {
+  def complete(status: StatusCode, headers: java.lang.Iterable[HttpHeader], entity: ResponseEntity) = RouteAdapter {
     D.complete(scaladsl.model.HttpResponse(status = status.asScala, entity = entity.asScala, headers = Util.immutableSeq(headers).map(_.asScala))) // TODO avoid the map()
   }
 
   /**
+   * Completes the request using the given status code, headers, and response entity.
+   */
+  @deprecated("This directive is for binary compatibility only", "10.0.6")
```
jrudolph: I think we should try out in which cases this overload is going to be used. If the most specific overload is picked, then it might be always this one (apart from `CloseDelimited` which isn't a `RequestEntity`). What we don't want is that every legitimate usage of `complete` is flagged with a deprecation warning.

akka-http/src/main/scala/akka/http/javadsl/server/directives/RouteDirectives.scala
```diff
@@ -112,11 +113,18 @@ abstract class RouteDirectives extends RespondWithDirectives {
...
+  def complete(status: StatusCode, headers: java.lang.Iterable[HttpHeader], entity: RequestEntity): RouteAdapter =
+    complete(status, headers, entity.asInstanceOf[ResponseEntity])
```
jrudolph: `entity: ResponseEntity` should work as well.

akka-http/src/main/scala/akka/http/javadsl/server/directives/RouteDirectives.scala
btomala: Sorry for long time without response..  Yes you have right, the most specific one is picked. I removed annotation and leave overloaded functions.

akka-http/src/main/scala/akka/http/javadsl/server/directives/RouteDirectives.scala
btomala: 👍

btomala:
How you want to test this? It is already tested in `CodingDirectivesTest`. In test it is called complete for `RequestEntity` and this call complete for `ResponseEntity`. What should I additionally test here?

jrudolph:
Ah, you are right, all of those go through the new and old methods. Sorry for the noise.

jrudolph:
Thanks a lot, @btomala, for fixing.

jrudolph:
Fixed #641.

Issue #213, opened 2016-09-09
Offer migration path from Spray Caching


ktoso:
<a href="https://github.com/ktoso"><img src="https://avatars.githubusercontent.com/u/120979?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [ktoso](https://github.com/ktoso)**
_Wednesday Jun 01, 2016 at 09:15 GMT_
_Originally opened as https://github.com/akka/akka/issues/20679_

---

The question keeps coming up, and it's only one simple file https://github.com/spray/spray/tree/master/spray-caching/src/main/scala/spray/caching

We could provide it as separate mini project, declare it community lead.
Please ping me if you'd like to contribute this port so I can create a repo for it.

ktoso:
<a href="https://github.com/zhxiaogg"><img src="https://avatars.githubusercontent.com/u/6303886?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [zhxiaogg](https://github.com/zhxiaogg)**
_Wednesday Jun 01, 2016 at 12:57 GMT_

----

@ktoso It's about migrating the spray cache directive to Akka HTTP server DSL, right? If so, I would like to give it a try.

ktoso:
<a href="https://github.com/ktoso"><img src="https://avatars.githubusercontent.com/u/120979?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [ktoso](https://github.com/ktoso)**
_Wednesday Jun 01, 2016 at 12:59 GMT_

----

Yes exactly - shouldn't be hard I think, and I think it should live outside of the akka/akka repo, but we can host it under the `akka/` organisation perhaps as `akka/akka-http-caching`. I would like to see it be a community by-demand driven project, I don't think we (core team) will spend time on it, and at the same time we do want to allow the HTTP ecosystem to flourish a bit more.

ktoso:
<a href="https://github.com/zhxiaogg"><img src="https://avatars.githubusercontent.com/u/6303886?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [zhxiaogg](https://github.com/zhxiaogg)**
_Wednesday Jun 01, 2016 at 12:59 GMT_

----

BTW, why a new project? Is it a planned repo to hold all extra directives?

ktoso:
[...]

PR #986, opened 2017-03-21
Migrate Spray caching as akka-http-caching #213

tomrf1:
Guided by the discussion [here](https://github.com/akka/akka-http/issues/213) (@ktoso, @ben-manes and @ianclegg) this change migrates the [Spray cache](https://github.com/spray/spray/tree/master/spray-caching/src/main/scala/spray/caching) and [cache directive](https://github.com/spray/spray/blob/b473d9e8ce503bafc72825914f46ae6be1588ce7/spray-routing/src/main/scala/spray/routing/directives/CachingDirectives.scala) across to a separate project, akka-http-caching.

This is an almost like-for-like replacement of the spray implementation (including the tests). I've generally left the code the same where possible, but there are some changes:

1. Whereas spray used `ConcurrentLinkedHashMap` for `LruCache`, it now uses caffeine's `AsyncLoadingCache`.
2. Removed `ascendingKeys` from the `Cache` trait because `AsyncLoadingCache` doesn't appear to have an equivalent method. I'm also not sure when you'd use it.
3. Removed implicit `ExecutionContext` from `apply` methods because it was only being used for completing the futures, whereas now it simply calls `get` on the caffeine cache. Does anyone think this will be a problem?
4. Because `AsyncLoadingCache` requires a loader on creation, a 'default' loader function must be passed in when creating a `LruCache`. Calling code can still pass in a separate loading function to `apply`.
5. **Edit:** Changed signature of `def remove(key: Any): Option[Future[V]]` to `def remove(key: Any): Unit`, as the value is unlikely to be needed


typesafehub-validator:
Hi @tomrf1,

ktoso:
OK TO TEST

ktoso:
NICE! Looking forward to reviewing :-)
We're doing team meeting and planning these few days so sorry for the delay in reviewing that may happen

ben-manes:
This looks really nice!

`ascendingKeys(n)` is a snapshot from coldest to hottest, which mirrors `descendingKeySet` that goes hottest to coldest. These were added for Cassandra which wanted to save the cache on shutdown for a warm restart. I'm not sure what they were used for by Spray. The equivalent in Caffeine is in `Policy.Eviction`, `hottest(n)` and `coldest(n)`. This can be obtained using `cache.synchronous().policy().eviction()`. 

Note that Caffeine uses [TinyLFU](https://arxiv.org/abs/1512.00727) instead of LRU for the eviction algorithm. Your class names with `Lru` may give the wrong expectations.

The `ExecutionContext` could be honored if you use `get(key, (key, executor) => future)`. Instead of using the supplied executor, you could use caller's. Alternatively, you could supply the `ExecutionContext` at creation time as the cache's executor by setting it in the builder.

It hasn't been clear to me whether introducing an `AsyncCache` would be useful. I'm glad you found the workaround. If there is a general consensus that it would be a good addition then the API was structures to make it seamless. My concern is that users will be inclined to use `getIfPresent` and `put` methods instead of the preferred loading methods, e.g. `get(key, func)`.

As a side note, if there is a desire for variable expiration, then that is a feature that we could collaborate on. The design that I've settled on is a hierarchical timer wheel, as shown in this [slide deck](https://docs.google.com/presentation/d/1NlDxyXsUG1qlVHMl4vsUUBQfAJ2c2NsFPNPr2qymIBs) (slide 10). This is an O(1) algorithm that would allow entries to have different expiration times, e.g. if caching external http resources based on their expiration header.

akka-http-caching/src/main/scala/akka/http/caching/LruCache.scala
```diff
@@ -0,0 +1,92 @@
...
+
+trait LruCache[V] extends Cache[V] {
+
+  private[caching] val store: AsyncLoadingCache[Any, V]
+
+  def get(key: Any): Option[Future[V]] = Option(store.getIfPresent(key)).map(_.toScala)
+
+  def apply(key: Any, genValue: () ⇒ Future[V]): Future[V] = store.get(key, toJavaMappingFunction(genValue)).toScala
+
+  def remove(key: Any): Option[Future[V]] = Option(store.synchronous().asMap().remove(key)).map(Future.successful)
```
ben-manes: When would a caller require the value on a removal? Its rare for a cache, so the penalty of waiting until the underlying future returns might be avoided.

If this is useful, we could look into supplying an `asMap()` view on the `AsyncLoadingCache`.

tomrf1: The spray implementation does not use `remove`, and I can't think why it would be useful to return the value. I had tried to keep the definitions consistent, but it's tempting to change this to:
`def remove(key: Any): Unit`

ben-manes: A scan of [usages on Github](https://github.com/search?l=Scala&p=4&q=%22spray.caching%22+remove&ref=searchresults&type=Code&utf8=%E2%9C%93) indicate the value is often ignored.

You might argue that this is a simple cache for the common cases, but users with more advanced needs should use Caffeine (or alternatives) directly. That way feature requests are kept at a minimum, as you aren't trying to cater to every edge case.

akka-http-caching/src/main/scala/akka/http/routing/directives/CachingDirectives.scala
```diff
@@ -0,0 +1,92 @@
...
+
+import scala.concurrent.duration.Duration
+import scala.concurrent.{ ExecutionContext, Future }
+import akka.http.scaladsl.server._
+import akka.http.scaladsl.model.HttpMethods.GET
+import akka.http.scaladsl.model.headers._
+import akka.http.scaladsl.model.headers.CacheDirectives._
+import akka.http.scaladsl.server.RouteResult.Rejected
+
+trait CachingDirectives {
```
jonas: Each exposed directives should be documented in a separate page in the `docs` project and listed in the [alphabetically](https://github.com/akka/akka-http/blob/0106fc93a05debd9276eafcf2dde94faee4cba15/docs/src/main/paradox/scala/http/routing-dsl/directives/alphabetically.md) and [by-trait](https://github.com/akka/akka-http/blob/0106fc93a05debd9276eafcf2dde94faee4cba15/docs/src/main/paradox/scala/http/routing-dsl/directives/by-trait.md) pages. The [spray cache directives docs](https://github.com/spray/spray/tree/master/docs/documentation/spray-routing/caching-directives) look like a good start.

Some minimal documentation introducing caching based on
https://github.com/spray/spray/blob/master/docs/documentation/spray-caching/index.rst would also be great. This would also be a good place to make it clear that using the caching directives requires adding an additional dependency.

akka-http-caching/src/main/scala/akka/http/routing/directives/CachingDirectives.scala
```diff
@@ -0,0 +1,92 @@
+package akka.http.routing.directives
```
jonas: The package for the Scala directives should be `akka.http.scaladsl.server.directives`

tomrf1:
@jonas thanks for the feedback, I've pushed some changes plus docs/examples for the caching directives.
I'm not familiar with the experimental annotation, is this specific to akka?

jonas:
The annotation I was thinking of is this one https://github.com/akka/akka/blob/v2.4.17/akka-actor/src/main/java/akka/annotation/ApiMayChange.java

akka-http-caching/src/test/scala/akka/http/caching/ExpiringLfuCacheSpec.scala
```diff
@@ -0,0 +1,135 @@
...
+      Await.result(cache(3)("C"), 3.seconds) should be("C")
+      cache(4)("D")
+      Thread.sleep(50)
+      cache.size should be(3)
+    }
+    "expire old entries" in {
+      val cache = lfuCache[String](timeToLive = 75 millis span)
+      Await.result(cache(1)("A"), 3.seconds) should be("A")
+      Await.result(cache(2)("B"), 3.seconds) should be("B")
+      Thread.sleep(50)
```
ben-manes: You could control time by using a fake Ticker instead of relying on sleeping.

tomrf1: thanks @ben-manes.
I've now tried using a `Ticker`. In order to get this test ("expire old entries") to reliably evict old items, I have to:
1. write another item to the cache _after_ their ttl has expired,
2. wrap the whole test in a `Future { }` and await.

Otherwise it still occassionally fails to evict before the checks - is this expected?

ben-manes: Expiration is performed as part of a maintenance cycle, which is asynchronous by default. You could call cleanUp() instead. Otherwise the entry remains but is masked as a miss if queried. The [testing](https://github.com/ben-manes/caffeine/wiki/Testing) and [cleanup](https://github.com/ben-manes/caffeine/wiki/Cleanup) wiki pages offers some hints.

tomrf1: After repeated runs it still occassionally fails to evict an item - even with calls to `.synchronous().cleanUp()` and after initialisation with an `Executor`.
It would be nice to get this working reliably, though it's tempting to remove these tests since they're now effectively testing a thin wrapper around caffeine, which I'm confident is well tested :)

Here's how I'm currently testing it - this often passes many times out of 50 before failing:
```
for (a ← 0 to 50) {
  val cache = lfuCache[String](timeToLive = 75 millis span)

  Await.result(cache(1)("A"), 3.seconds) should be("A")
  ticker.advance(50 millis span)

  Await.result(cache(2)("B"), 3.seconds) should be("B")
  cache.size should be(2)  //always passes

  ticker.advance(50 millis span)

  eventually {  //repeatedly check if the item has been evicted
    cache.store.synchronous().cleanUp()
    Thread.sleep(50)

    //If this fails once then it never passes, no matter how many times it runs
    cache.get(1) should be(None)
    cache.size should be(1)
  }
}
```

ben-manes: I guess somehow item B was preferred before A in the linked list. Then since it peeks at the head, A doesn't get expired since B is valid. That could happen with races and is an okay limitation in practice, so it's used to make expiration an O(1) operation. I don't see why it would race here, but sounds like the problem.

rgueldem:
Would it be possible to include a variant that is backed by a non-loading cache? When migrating  from Spray Caching, adding a default loader to the constructor requires some significant changes on the consumer side.

ben-manes:
I think that is what the [routeCache](https://github.com/akka/akka-http/pull/986/files#diff-f62e83b45b34ac0701814bad9052ce0bR59) factory method is for? It uses a defaultLoader that returns null, to workaround the lack of a non-loading `AsyncCache` type.

rgueldem:
@ben-manes we use spray caching in a different context without akka-http (or spray for that matter). Thank you for the suggestion, I'm now using a default loader that always returns a failed future. My comment was mainly about simplifying the migration path, since adding a default loader introduces breaking changes. But not a big deal, excited to see this move forward.

ben-manes:
It does sound like using a default parameter might be desirable.

tomrf1:
@rgueldem I've just removed the need for a loader argument, since the loader is never actually used by `LfuCache` and it's not really being used as a loading cache (`AsyncLoadingCache` just has some useful properties). Hopefully this removes any breaking changes for you

rgueldem:
Is there an estimate on when this might get merged and release? Spray Caching is the last dependency for us that isn't available on Scala 2.12. Would be great to migrate to akka-http-caching soon.

jrudolph:
Thanks for the reminder, @rgueldem, to bring this PR over the finish line.

Without having yet looked deeply into this PR, I want to say that I agree a lot with the idea of resurrecting spray-caching. Great collaboration so far, @tomrf1, @ben-manes, @jonas, and @rgueldem. Thanks also for adding background information, @ben-manes, it's great to have you here. After reading up a bit on caffeine I fully agree with the move to caffeine. Good that it doesn't have any additional dependencies.

I'll try to have a deeper look over the details this week.

akka-http-caching/src/main/scala/akka/http/caching/Cache.scala
```diff
@@ -0,0 +1,76 @@
...
+ * General interface implemented by all akka-http cache implementations.
+ */
+trait Cache[V] { cache ⇒
+
+  /**
+   * Selects the (potentially non-existing) cache entry with the given key.
+   */
+  def apply(key: Any) = new Keyed(key)
+
+  class Keyed(key: Any) {
```
jrudolph: Do we still remember why this (somewhat uncommon) code structure was used? In which way is it better then simply providing a few more overloads for `Cache.apply`?

akka-http-caching/src/main/scala/akka/http/caching/Cache.scala
```diff
@@ -0,0 +1,76 @@
...
+    /**
+     * Returns either the cached Future for the key or evaluates the given call-by-name argument
+     * which produces either a value instance of type `V` or a `Future[V]`.
+     */
+    def apply(magnet: ⇒ ValueMagnet[V]): Future[V] =
```
jrudolph: (At least) my personal opinion about magnet have shifted a bit. In my opinion, we should use overloads or just different methods where possible instead of magnets (which come with the extra burden of implicit resolution errors on wrong usage).

akka-http-caching/src/main/scala/akka/http/caching/LfuCache.scala
```diff
@@ -0,0 +1,91 @@
...
+
+object LfuCache {
+
+  /**
+   * Creates a new [[akka.http.caching.ExpiringLfuCache]] or
+   * [[akka.http.caching.SimpleLfuCache]] instance depending on whether
+   * a non-zero and finite timeToLive and/or timeToIdle is set or not.
+   */
+  def apply[V](
+    maxCapacity:     Int      = 500,
```
jrudolph: Let's move sets of configuration parameters to settings classes for easier (= binary compatible) evolution of the API later on.

akka-http-caching/src/main/scala/akka/http/caching/LfuCache.scala
```diff
@@ -0,0 +1,91 @@
...
+  def dummyLoader[V] = new AsyncCacheLoader[Any, V] {
+    def asyncLoad(k: Any, e: Executor) =
+      Future.failed[V](new RuntimeException("Dummy loader should not be used by LfuCache")).toJava.toCompletableFuture
+  }
+
+  def toJavaMappingFunction[V](genValue: () ⇒ Future[V]) =
+    asJavaBiFunction[Any, Executor, CompletableFuture[V]]((k, e) ⇒ genValue().toJava.toCompletableFuture)
+}
+
+trait LfuCache[V] extends Cache[V] {
```
jrudolph: I have a few ideas about how to improve that actual implementation code structure (but we can still do this afterwards in another PR). I'll add them in extra comments.

jrudolph: Couldn't we make this a concrete class with a `store` parameter? We could then remove the two subclasses and move the initialization to `LfuCache.apply` (or extra helper methods).

jrudolph: Can also be `private[caching]`.

akka-http-caching/src/main/scala/akka/http/scaladsl/server/directives/CachingDirectives.scala
```diff
@@ -0,0 +1,91 @@
...
+  def routeCache(maxCapacity: Int = 500, initialCapacity: Int = 16, timeToLive: Duration = Duration.Inf,
+                 timeToIdle: Duration = Duration.Inf): Cache[RouteResult] = {
+    LfuCache(maxCapacity, initialCapacity, timeToLive, timeToIdle)
+  }
+  //#
+}
+
+object CachingDirectives extends CachingDirectives
+
+trait CacheSpecMagnet {
```
jrudolph: I don't think we need a magnet here any more. Let's go with either of those possibilities:

 * Make `cache` take both the cache and the keyer:

```scala
def cache(cache: Cache[RouteResult], keyer: PartialFunction[RequestContext, Any] = Default): Direction0
```
 * Or make `cache` and/or `keyer` methods of `CachingDirectives` with or without default implementations. Users could then customize behavior by implementing / overriding these methods when extending from `CachingDirectives`.

akka-http-caching/src/main/scala/akka/http/scaladsl/server/directives/CachingDirectives.scala
```diff
@@ -0,0 +1,91 @@
...
+      def liftedKeyer = keyer.lift
+      implicit def executionContext = factory.dispatcher
+    }
+}
+
+trait CacheKeyer extends (PartialFunction[RequestContext, Any])
+
+object CacheKeyer {
+  implicit val Default: CacheKeyer = CacheKeyer {
+    case r: akka.http.scaladsl.server.RequestContext if r.request.method == GET ⇒ r.request.uri
```
jrudolph: I know that's same as in spray but I think it's somewhat dangerous to only regard the URI as the key. If you place the cache at the wrong level of the routing hierarchy it might remove access controls, use cookie values from other users etc.

So we should at least try to be a bit more diligent to figure out what kind of headers to take into account as well (or use as a blacklist). E.g. we could say also never to cache response that contain `Authorization` headers. It's harder for cookies because we don't know which cookies are relevant for routing and result generation.

And then we need to figure out how to provide reasonable default behavior. E.g. we could also change authentication directives to add `no-cache` headers.

Or we decide to bail out and add a fat warning that users need to know what they are using when the `Default` keyer is used. Does someone has any information about how other software, e.g. transparent caches, deal with this problem?

Any other ideas?

docs/src/main/paradox/scala/http/routing-dsl/directives/alphabetically.md
```diff
@@ -3,6 +3,7 @@
 
 |Directive                                                                                                                               | Description                                                                                                                                                                                               |
 |----------------------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+|@ref[alwaysCache](caching-directives/alwaysCache.md#alwaysCache)                                                                        | Wraps its inner Route with caching support using the given `akka.http.caching.Cache` implementation and the in-scope keyer function.                                                                      |
```
jrudolph: Thanks for porting documentation as well.

project/MiMa.scala
```diff
@@ -35,7 +35,7 @@ object MiMa extends AutoPlugin {
           "10.0.4",
           "10.0.5"
       )
-        .map((version: String) => organization.value %% name.value % version)
+        .collect { case version if name.value != "akka-http-caching" => organization.value %% name.value % version }
```
jrudolph: Let's put that module into a list of `ignoredModules` with a comment so it's clear what we are doing here.

akka-http-caching/src/main/scala/akka/http/caching/Cache.scala
tomrf1: I'm very much in favour of removing this use of magnets here.
And thinking about it now, this was used by the original spray implementation of `CachingDirectives.alwaysCache`, but that now directly calls `Cache.apply`. Unless anyone can think of a reason for keeping it, I'll remove the `Keyed` class.

akka-http-caching/src/main/scala/akka/http/scaladsl/server/directives/CachingDirectives.scala
tomrf1: Good point, it could result in some seriously undesirable behaviour (for our specific use-case it's not a problem).
It's tempting to remove the default keyer entirely, to force developers to think carefully about when to cache responses.

ben-manes:
fyi, I released 2.5.0 last night. This provides support for variable expiration, which could be used to model cases like caching an external HTTP response based on its `expires` header.

akka-http-caching/src/main/scala/akka/http/caching/LfuCache.scala
tomrf1: Can you clarify what you mean by settings classes please? Just a normal class?

jrudolph: In the simplest case it would just be a case class that carries all the arguments. However, because case classes are likely to break binary compatibility when changes are necessary we now usually use a public abstract class / private case class implementation combo. Basically like this:

```scala
abstract class LfuCacheSettings private[http]() { self: LfuCacheSettingsImpl =>
 def maxCapacity:     Int
 def initialCapacity: Int
 def timeToLive:      Duration
 def timeToIdle:      Duration 

  def withMaxCapacity(newMaxCapacity: Int): LfuCacheSettings = copy(maxCapacity = newMaxCapacity)
 // similar for other settings
}
object LfuCacheSettings {
  def apply(): LfuCacheSettings = LfuCacheSettingsImpl()
  
  private[http] case class LfuCacheSettingsImpl(
      maxCapacity:     Int      = 500,
      initialCapacity: Int      = 16,
      timeToLive:      Duration = Duration.Inf,
      timeToIdle:      Duration = Duration.Inf
  ) extends LfuCacheSettings
}
```

then the signature could be

```scala
def apply[V](settings: LfuCacheSettings = LfuCacheSettings()): Cache[V]
```

It's a bit of boilerplate but it is helpful later when changes might become necessary.

tomrf1:
@jrudolph To recap, here are the changes in response to your main comments:
1. _replace Cache.Keyed with more overloads if possible_
Removed `Cache.Keyed` and the use of magnet here - https://github.com/akka/akka-http/pull/986/commits/95e5de800f62b219bec61e562db3772e49fa93cc

2. _remove magnet_
Removed `CacheSpecMagnet` - https://github.com/akka/akka-http/pull/986/commits/f737f66e9b80f09fef360f064bf360ef7769bfcc

3. _think about securing the default keyer_
Removed default keyer parameter - https://github.com/akka/akka-http/pull/986/commits/538ae8d70fa6ca36227506cba5c291acb268a5e4. It may be worth devising more sophisticated keyers, though this is perhaps beyond the scope of this PR?

jonas:
Build failure is #492

PLS BUILD

build.sbt
```diff
@@ -122,6 +123,11 @@ lazy val httpMarshallersJava = project("akka-http-marshallers-java")
 lazy val httpJackson =
   httpMarshallersJavaSubproject("jackson")
 
+lazy val httpCaching = project("akka-http-caching")
```
jonas: I am wondering if it would be better to put the core caching API in the `akka-http` project and then have the Caffeine related code in `akka-http-caching/akka-http-caffeine` similar to how we manage marshallers with external dependencies. This could make caching more self-documenting.

```scala
import akka.http.caching.caffeine.CaffeineCachingSupport.routeCache

val route =
  cache(routeCache(), simpleKeyer) {
    // ...
  }
```

henders:
Is this being merged soon since it has at least 1 approval?

build.sbt
tomrf1: The decision to put it all under akka-http-caching was based on @ktoso 's [original comment](https://github.com/akka/akka-http/issues/213#issuecomment-271832937). I'm open to changing this though

jonas: Ignore my comment. I've lost track of the conversation.

jonas:
I will try to find some time to migrate the documentation from Spray.

@henders If you have time and want to help remove this blocker, let me know.

daandi:
Is there anything one could do to support the merging of this ticket ? Would be so nice to have it.

akka-http-caching/src/main/scala/akka/http/caching/Cache.scala
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * General interface implemented by all akka-http cache implementations.
+ */
+trait Cache[V] { cache ⇒
+
+  /**
+   * Returns either the cached Future for the given key or evaluates the given value generating
+   * function producing a `Future[V]`.
+   */
+  def apply(key: Any, genValue: () ⇒ Future[V]): Future[V]
```
raboof: That would make the cache easier to use indeed

akka-http-caching/src/main/scala/akka/http/scaladsl/server/directives/CachingDirectives.scala
```diff
@@ -0,0 +1,72 @@
...
+    case r: RequestContext if r.request.method == GET ⇒ r.request.uri
+  }
+
+  /**
+   * Passes only requests to the inner route that explicitly forbid caching with a `Cache-Control` header with either
+   * a `no-cache` or `max-age=0` setting.
+   */
+  def cachingProhibited: Directive0 =
+    extract(_.request.headers.exists {
+      case x: `Cache-Control` ⇒ x.directives.exists {
```
raboof: Huh, didn't know this was supported on requests. Neat.

akka-http-caching/src/main/scala/akka/http/caching/Cache.scala
```diff
@@ -0,0 +1,63 @@
...
+
+  /**
+   * Returns either the cached Future for the key or evaluates the given function which
+   * should lead to eventual completion of the promise.
+   */
+  def apply[U](key: Any, f: Promise[V] ⇒ U): Future[V] =
```
raboof: `U` is unused here - would it make sense to make it specifically `Unit`? Or would that make it too hard to use?

tomrf1:
@raboof @jonas I've just pushed a commit to make the key type generic instead of `Any`, what do you think?

raboof:
PLS BUILD

akka-http-caching/src/main/scala/akka/http/caching/Cache.scala
```diff
@@ -0,0 +1,63 @@
...
+  def get(key: K, block: ⇒ V): Future[V] = apply(key, () ⇒ Future.successful(block))
+
+  /**
+   * Retrieves the future instance that is currently in the cache for the given key.
+   * Returns None if the key has no corresponding cache entry.
+   */
+  def get(key: K): Option[Future[V]]
+
+  /**
+   * Removes the cache item for the given key. Returns the removed item if it was found (and removed).
```
yannick-cw: The comment is not valid anymore with returning `Unit`

raboof:
(that last failure looks like #595)

raboof:
PLS BUILD

raboof:
PLS BUILD

raboof:
(latest looks like a new failure, added #1220)

daddykotex:
To merge? 😄

Sebruck:
Merge please? We are waiting for this :-)

akka-http-caching/src/main/scala/akka/http/caching/Cache.scala
```diff
@@ -0,0 +1,63 @@
...
+
+  /**
+   * Removes the cache item for the given key.
+   */
+  def remove(key: K): Unit
+
+  /**
+   * Clears the cache by removing all entries.
+   */
+  def clear()
```
ktoso: explicit `: Unit`

ktoso:
I'll propose some changes for it to get java API today or tomorrow :)
Should be the quickest way to get it merged I believe (there's some mystical dancing around 💃  involved in getting JavaDSL right)

akka-http-caching/src/main/scala/akka/http/caching/Cache.scala
```diff
@@ -0,0 +1,63 @@
...
+
+import akka.annotation.ApiMayChange
+
+import scala.concurrent.{ Future, Promise }
+
+/**
+ * General interface implemented by all akka-http cache implementations.
+ */
+@ApiMayChange
+trait Cache[K, V] { cache ⇒
```
jonas: Do you mean it should be in the `akka.http.caching.scaladsl` package?

jonas:
My two cents: @tomrf1 has done an amazing job answering and updating this PR over the last three months. Merging will make it easier for others to contribute docs and the Java DSL. We already marked the API as work in progress so I suggest we meet sooner rather than later.

ktoso:
> My two cents: @tomrf1 has done an amazing job answering and updating this PR over the last three months. 

Yes, absolutely!

> Merging will make it easier for others to contribute docs and the Java DSL. 

I'm not sure I agree. That's only the upsides. The downside is that we can't release until someone does. I do not want to have to release ever again with a half baked feature. We did a few times, where a feature was in but had no docs, or a feature was in but without javadsl. This is bad for our reputation which is: that we always deliver completed features, including javadsl and docs.

> We already marked the API as work in progress so I suggest we meet sooner rather than later.

Again, I disagree here and I have personally stepped up contributing the missing bits and will do so over my "free time" over the weekend contributing the missing buts of the javadsl.

I think this PR is great, and *because* I want it in, I will spend personal time contributing to pushing it over the finish line. I don't mean do sound dismissing, I hope you understand my rationale and commitment for this specific feature. (Recently we did a half-baked release, with akka-sse, and I would not want another one like this)

ktoso:
Thanks for the push @tomrf1 :) Give me a day to propose the javadsl for these - heading to bed now in my timezone though :-)

jlprat:
Failure is about formatting in `docs/src/test/scala/docs/http/scaladsl/server/directives/CachingDirectivesExamplesSpec.scala`

ktoso:
I didn't manage to fully complete the missing bits, but here's a start and point for discussion: https://github.com/ktoso/akka-http/commit/b19c874f86ced992862eabf1dd5c741ceb046be9

Please see comments in-line.

In one word, it's "just" missing javadsl, but that has some implications on where to put things. Sure, one could say - "well just move things around later!", but that right I'd reserve for `@ApiMayChange` APIs when we really can't avoid breaking things, and here, we just need to add the API before releasing. The biggest question is:

- I'm worried about exporting things into already taken namespace. Will this work under jigsaw, doesn't this break osgi already? Specifically, putting the CachingDirectives under `akka.http.server.directives
- Making the Cache actually implement the java/scala dsl's as I've started outlining but did not have time to finish over the weekend.

If someone wants to take over from that PR please let me know so we don't duplicate efforts -- cc @raboof @2m @jrudolph @tomrf1

ktoso:
FYI on behalf of the Akka team, we're very very focused on shipping a different feature this week so we're really sorry if seeminly a bit unresponsive this week. I can promise though that in the weeks after that we'll give http and friends much attention (including this lovely PR 😍 ).

rgueldem:
Sorry for nagging again, but is there any chance this gets released soon? We'd love to use this to migrate from spray caching. If it is not part of the akka roadmap, we will have to look for alternatives. It's a blocker for us since we can't run on Scala 2.12 with spray caching.

ktoso:
Presence of PR and work on it means it *is* in the roadmap, just that the last weeks team was 
slimmed down. As you'll notice, there's active discussion and work on it.

Best way to get it in is to help out on the remaining bits of course, which would be very nice.

tomrf1:
@ktoso I've pushed my WIP attempt to continue your javadsl work.

akka-http-caching/src/main/scala/akka/http/caching/scaladsl/Cache.scala
```diff
@@ -39,14 +40,16 @@ abstract class Cache[K, V] extends akka.http.caching.javadsl.Cache[K, V] {
   /**
    * Returns either the cached Future for the given key or the given value as a Future
    */
-  def apply(key: K, block: () ⇒ V): Future[V] =
+  def get(key: K, block: () ⇒ V): Future[V] =
```
tomrf1: this cannot be `apply` because it clashes with `def apply(key: K, genValue: () ⇒ Future[V]): Future[V]` after type erasure

akka-http-caching/src/main/scala/akka/http/javadsl/server/directives/CachingDirectives.scala
```diff
@@ -27,49 +25,64 @@ class CachingDirectives {
    *
    * Use [[akka.japi.JavaPartialFunction]] to build the `keyer`.
    */
-  def cache[K](cache: Cache[K, RouteResult], keyer: PartialFunction[RequestContext, K]) = RouteAdapter {
-    D.cache(cache, keyer)
+  def cache[K](cache: Cache[K, RouteResult], keyer: PartialFunction[RequestContext, K], inner: Supplier[Route]) = RouteAdapter {
+    D.cache(
+      cache.asInstanceOf[akka.http.caching.scaladsl.Cache[K, akka.http.scaladsl.server.RouteResult]],
```
tomrf1: as discussed, needs to be implemented using `JavaMapping` style

TimothyKlim:
Any news?

markusthoemmes:
Anything left to push this over the line? Maybe I can help?

tomrf1:
@ktoso when you have time it would be great to have your thoughts on the java dsl commits. Thanks!

ktoso:
Cool, looking at the javadsl then!

akka-http-caching/src/main/scala/akka/http/caching/CacheJavaMapping.scala
```diff
@@ -0,0 +1,23 @@
+package akka.http.caching
```
ktoso: Copyright ;)

akka-http-caching/src/main/scala/akka/http/caching/CacheJavaMapping.scala
```diff
@@ -0,0 +1,23 @@
...
+
+object CacheJavaMapping {
+
+  /**
+   * Impossible to use `Inherited` without covariance in `Cache`?
+   */
+  //  import akka.http.impl.util.JavaMapping.Inherited
+  //  def cacheMapping[JK, JV, SK <: JK, SV <: JV]: JavaMapping[Cache[JK, JV], akka.http.caching.scaladsl.Cache[SK, SV]] =
+  //    new Inherited[Cache[JK, JV], akka.http.caching.scaladsl.Cache[SK, SV]]
+
```
ktoso: Hmm, perhaps... not super important, we can look into it separately

akka-http-caching/src/main/java/akka/http/caching/javadsl/Cache.java
```diff
@@ -0,0 +1,63 @@
...
+package akka.http.caching.javadsl;
+
+import akka.japi.Creator;
+import akka.japi.Procedure;
+
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+
+public interface Cache<K, V> {
```
ktoso: `@ApiMayChange` as well as `@DoNotInherit` I think

akka-http-caching/src/main/java/akka/http/caching/javadsl/Cache.java
```diff
@@ -0,0 +1,63 @@
...
+  java.util.Set<K> getKeys();
+
+  /**
+   * Returns the upper bound for the number of currently cached entries.
+   * Note that this number might not reflect the exact number of active, unexpired
+   * cache entries, since expired entries are only evicted upon next access
+   * (or by being thrown out by a capacity constraint).
+   */
+  int size();
+}
```
ktoso: API looks good!

akka-http-caching/src/main/scala/akka/http/caching/scaladsl/Cache.scala
```diff
@@ -0,0 +1,97 @@
...
+import scala.concurrent.{ Future, Promise }
+
+/**
+ * API MAY CHANGE
+ *
+ * General interface implemented by all akka-http cache implementations.
+ */
+@ApiMayChange
+@DoNotInherit
+abstract class Cache[K, V] extends akka.http.caching.javadsl.Cache[K, V] {
```
ktoso: good

akka-http-caching/src/main/scala/akka/http/caching/scaladsl/Cache.scala
```diff
@@ -0,0 +1,97 @@
...
+  override def getStrict(key: K, block: Creator[V]): CompletionStage[V] =
+    futureToJava(get(key, () ⇒ block.create))
+
+  /**
+   * Returns the upper bound for the number of currently cached entries.
+   * Note that this number might not reflect the exact number of active, unexpired
+   * cache entries, since expired entries are only evicted upon next access
+   * (or by being thrown out by a capacity constraint).
+   */
+  override def size(): Int
```
ktoso: no need to override here I think

akka-http-caching/src/main/scala/akka/http/javadsl/server/directives/CachingDirectives.scala
```diff
@@ -0,0 +1,94 @@
...
+          keyer(javaRequestContext)
+        }
+    }
+  }
+
+  /**
+   * A simple keyer function that will cache responses to *all* GET requests, with the URI as key.
+   * WARNING - consider whether you need special handling for e.g. authorised requests.
+   */
+  val simpleKeyer: PartialFunction[RequestContext, Uri] = {
```
ktoso: should this be private? or is this intended to be used by people?

akka-http-caching/src/main/scala/akka/http/javadsl/server/directives/CachingDirectives.scala
```diff
@@ -0,0 +1,94 @@
...
+   * keyer function. Note that routes producing streaming responses cannot be wrapped with this directive.
+   */
+  def alwaysCache[K](cache: Cache[K, RouteResult], keyer: PartialFunction[RequestContext, K], inner: Supplier[Route]) = RouteAdapter {
+    D.alwaysCache(
+      cache.asInstanceOf[akka.http.caching.scaladsl.Cache[K, akka.http.scaladsl.server.RouteResult]],
+      toScalaKeyer(keyer)
+    ) { inner.get.delegate }
+  }
+
+  def routeCache[K](): Cache[K, RouteResult] =
```
ktoso: scaladoc on these?

akka-http-caching/src/main/scala/akka/http/javadsl/server/directives/CachingDirectives.scala
tomrf1: It's only used by the tests/example code. It could be moved out, but I'm ok with it remaining public as an example of a basic keyer (as it was in spray)

jonas:
Instance of #492
PLS BUILD

jonas:
Instance of #1376
PLS BUILD

tomrf1:
@ktoso any plans to get this merged/released soon?

ktoso:
Thanks for the ping, seems javadsl was added? It's on our shortlist to have a last look here and merge as soon as possible. Thanks for your patience :)

abv-prakash:
@ktoso this is the most awaited and also a dependent feature for us to release our application on akka-http. Any tentative dates by which we can expect to get this merged and have new release version available .

akka-http-caching/src/main/scala/akka/http/caching/CacheJavaMapping.scala
```diff
@@ -0,0 +1,20 @@
+/**
+ * Copyright (C) 2017 Lightbend Inc. <http://www.lightbend.com/>
+ */
+
+package akka.http.caching
+
+import akka.http.caching.javadsl.Cache
+import akka.http.impl.util.JavaMapping
+
+object CacheJavaMapping {
```
jrudolph: Should be internal API at least for now.

akka-http-caching/src/main/scala/akka/http/caching/LfuCache.scala
```diff
@@ -0,0 +1,129 @@
...
+    require(settings.maxCapacity >= 0, "maxCapacity must not be negative")
+    require(settings.initialCapacity <= settings.maxCapacity, "initialCapacity must be <= maxCapacity")
+
+    if (settings.timeToLive.isFinite() || settings.timeToIdle.isFinite()) expiringLfuCache(settings.maxCapacity, settings.initialCapacity, settings.timeToLive, settings.timeToIdle)
+    else simpleLfuCache(settings.maxCapacity, settings.initialCapacity)
+  }
+
+  /**
+   * Creates a new [[akka.http.caching.LfuCache]], with optional expiration depending
+   * on whether a non-zero and finite timeToLive and/or timeToIdle is set or not.
```
jrudolph: Add "Java API" marker to Scaladoc.

akka-http-caching/src/main/scala/akka/http/caching/LfuCache.scala
```diff
@@ -0,0 +1,129 @@
...
+   */
+  def create[K, V](): akka.http.caching.javadsl.Cache[K, V] =
+    apply(LfuCacheSettings())
+
+  /**
+   * Creates a new [[akka.http.caching.LfuCache]], with optional expiration depending
+   * on whether a non-zero and finite timeToLive and/or timeToIdle is set or not.
+   */
```
jrudolph: "Java API"

akka-http-caching/src/main/scala/akka/http/caching/LfuCache.scala
```diff
@@ -0,0 +1,129 @@
...
+  def remove(key: K): Unit = store.synchronous().invalidate(key)
+
+  def clear(): Unit = store.synchronous().invalidateAll()
+
+  def keys: Set[K] = store.synchronous().asMap().keySet().asScala.toSet
+
+  override def size: Int = store.synchronous().asMap().size()
+}
+
+abstract class LfuCacheSettings private[http] () { self: LfuCacheSettingsImpl ⇒
```
jrudolph: Good to put them into this class :+1:

akka-http-caching/src/main/scala/akka/http/caching/scaladsl/Cache.scala
```diff
@@ -0,0 +1,97 @@
...
+    futureToJava(apply(key, promise ⇒ {
+      val completableFuture = new CompletableFuture[V]
+      f(completableFuture)
+      promise.completeWith(futureToScala(completableFuture))
+    }))
+
+  /**
+   * Returns either the cached CompletionStage for the given key or the given value as a CompletionStage
+   */
+  override def getStrict(key: K, block: Creator[V]): CompletionStage[V] =
```
jrudolph: Why is it called `getStrict`?

jrudolph: I see, it's in opposition to `getFuture`. It sounds a bit weird because it still returns a `CompletionStage` which isn't strict. The reason it has to, is because there might already be a computation running for that key. Maybe we could call it `getOrCreateStrict`?

akka-http-caching/src/main/scala/akka/http/javadsl/server/directives/CachingDirectives.scala
jrudolph: I'd favor moving it out. It has all kinds of problems because it ignores important parts of a request, most importantly the existence of authentication information.

akka-http-caching/src/main/scala/akka/http/javadsl/server/directives/CachingDirectives.scala
```diff
@@ -0,0 +1,108 @@
...
+  /**
+   * Creates an [[LfuCache]] with default settings
+   */
+  def routeCache[K](): Cache[K, RouteResult] =
+    JavaMapping.toJava(D.routeCache[K](500, 16, Duration.Inf, Duration.Inf))
+
+  /**
+   * Creates an [[LfuCache]]
+   */
+  def routeCache[K](maxCapacity: Int): Cache[K, RouteResult] =
```
jrudolph: Let's use `LfuCacheSettings` instead of all these overloads. It gets only worse from there ;)

akka-http-caching/src/main/scala/akka/http/scaladsl/server/directives/CachingDirectives.scala
```diff
@@ -0,0 +1,81 @@
...
+        case Some(key) ⇒ cache.apply(key, () ⇒ route(ctx))
+        case None      ⇒ route(ctx)
+      }
+    }
+  }
+
+  /**
+   * Creates an [[LfuCache]]
+   */
+  def routeCache[K](maxCapacity: Int = 500, initialCapacity: Int = 16, timeToLive: Duration = Duration.Inf,
```
jrudolph: Let's use `LfuCacheSettings` here as well.

build.sbt
jrudolph: I think it's good for now. We can think about moving it in later on.

tomrf1:
@jrudolph thanks for the comments, see latest commits.
Regarding `keyer` documentation, the standard defined here is useful: https://www.greenbytes.de/tech/webdav/rfc7234.html#response.cacheability.
`keyer` definitions could vary quite a lot depending on the use case. I could add to the docs a simple example which does not cache where the Authorization header is present?

build.sbt
```diff
@@ -132,7 +132,7 @@ lazy val httpJackson =
 
 lazy val httpCaching = project("akka-http-caching")
   .settings(Dependencies.httpCaching)
-  .settings(Version.versionSettings)
+  .settings(VersionGenerator.versionSettings)
```
jonas: Why is the caching generating the same class and config file as http-core?

ktoso:
Attempting to get this merged today, thanks for patiently working on this for such a long time!

ktoso:
PLS BUILD

ktoso:
Awesome work @tomrf1!

I believe this is ready to merge and I have some slight follow up changes I'll PR separately. I checked the comments that @jrudolph had and those are addressed as well or I'll do so in the PR I have prepped.

I'll finish figuring out why the build fails with `[error] com.lightbend.paradox.markdown.Index$LinkException: Unknown page [low-level-api.md] linked from [java/http/server-side/index.md]` etc and merge this PR and do my follow up (prepared already) shortly then.

:shipit:

jonas:
It could be #1371

jonas:
I started on porting the doc here: https://github.com/jonas/akka-http/tree/caching-docs

ktoso:
Yeah it's the symlink that did not get created and blows up the PR now https://github.com/akka/akka-http/pull/1371/files#r147919705
It's not really failing the new docs though, but all of the normal dos files (not the caching ones)... very weird.

jonas:
PLS BUILD

raboof:
PLS BUILD

ktoso:
@raboof please notice that I continued work on this over in https://github.com/akka/akka-http/pull/1503 and that should be merged (very carefully to keep commits), rather than this PR I think

Issue #981, opened 2017-03-20
'Status code misses trailing space' Error && Handling Status lines without phrasing


sheinbergon:
Hello.

I've recently came across an API that didn't provide fully HTTP 1.1 complaint  status lines in case of errors, meaning no phrasing  ('HTTP/1.1 400' when it should be 'HTTP/1.1 400 Bad Request', for example). 

Though there's no doubt it's an improper implementation, other client libraries/command line tools seem to be able to cope with it, while akka-http breaks with an error emitted from [HttpResponseParser](https://github.com/akka/akka-http/blob/master/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpResponseParser.scala) (line 99-100). 

Is there any chance this behavior is considered to strict and/or could/should be opted out?

jrudolph:
Yes, I think that it might make sense to be loosened up.

PR #988, opened 2017-03-23
Make Response parser more relaxed on reason message #981

athieriot:
Proposal for issue #981 that will allow, in ParsingMode.Relax only, to
parse a non-standard Http Response which does not contain a status
message.

Example: 'HTTP/1.1 400' instead of 'HTTP/1.1 400 Bad Request'


athieriot:
Ah sorry I must be tired
I can't just use uriParsing mode like that!

I'll be back with another proposal

PR #1000, opened 2017-03-29
Extract headerValue with default value Java #639

jlprat:
Issue: #639
Java example on how to provide a default value for a header extractor using directive composition
Reformatted the code

### DIFF
--- a/docs/src/test/java/docs/http/javadsl/server/directives/HeaderDirectivesExamplesTest.java
+++ b/docs/src/test/java/docs/http/javadsl/server/directives/HeaderDirectivesExamplesTest.java
@@ -22,6 +22,9 @@ import akka.japi.JavaPartialFunction;
 import akka.http.javadsl.testkit.TestRoute;
 import scala.PartialFunction;
 
+import static akka.http.javadsl.server.Directives.*;
+import static akka.http.javadsl.common.PartialApplication.*;
+
 public class HeaderDirectivesExamplesTest extends JUnitRouteTest {
 
   @Test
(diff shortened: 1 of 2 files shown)


docs/src/test/java/docs/http/javadsl/server/directives/HeaderDirectivesExamplesTest.java
```diff
@@ -22,6 +22,9 @@
 import akka.http.javadsl.testkit.TestRoute;
 import scala.PartialFunction;
 
+import static akka.http.javadsl.server.Directives.*;
+import static akka.http.javadsl.common.PartialApplication.*;
```
jonas: Are these needed by the new example?

docs/src/test/java/docs/http/javadsl/server/directives/HeaderDirectivesExamplesTest.java
```diff
@@ -36,87 +39,113 @@ public void testHeaderValue() {
...
+      .assertStatusCode(StatusCodes.NOT_FOUND)
+      .assertEntity("The requested resource could not be found.");
     //#headerValue
   }
 
   @Test
+  public void testHeaderValueWithDefault() {
+    //#headerValue-with-default
+    final Function<HttpHeader, Optional<String>> extractExampleHeader = header -> {
+      if (header.name().equalsIgnoreCase("exampleHeader")) {
```
jrudolph: You could also use `header.is("exampleheader")` (`is` requires a lower case name to compare against). Maybe it would make sense to not use camel case for a header name but something more "realistic" like `X-Example-Header` and then `header.is("x-example-header")`.

docs/src/test/java/docs/http/javadsl/server/directives/HeaderDirectivesExamplesTest.java
jlprat: Yes, I need them to write something like
```scala
anyOf(
      bindParameter(this::headerValue, extractExampleHeader),
      bindParameter(this::provide, "newValue"),
     (String value) -> complete("header is " + value));
```

The alternative is more verbose coding myself the half applied functions.
Alternatively, I could just import the classes and reference the static methods explicitly.

docs/src/test/java/docs/http/javadsl/server/directives/HeaderDirectivesExamplesTest.java
jlprat: Sure thing. I will apply those changes later today

jrudolph:
Thanks, @jlprat!

PR #1008, opened 2017-03-31
Add OSGi headers to manifests (Issue #574)

michael-smith:


project/plugins.sbt
```diff
@@ -23,5 +23,6 @@ addSbtPlugin("net.virtual-void" % "sbt-dependency-graph" % "0.8.0") // for advan
 addSbtPlugin("io.spray" % "sbt-boilerplate" % "0.6.0")
 addSbtPlugin("me.lessis" % "bintray-sbt" % "0.3.0-8-g6d0c3f8")
 addSbtPlugin("com.lightbend.sbt" % "sbt-javaagent" % "0.1.2")
+addSbtPlugin("com.typesafe.sbt" % "sbt-osgi" % "0.9.1")
```
jonas: The plugin is already added in line 17

jonas:
OK TO TEST

project/OSGi.scala
```diff
@@ -0,0 +1,87 @@
...
+    OsgiKeys.bundleVersion := version.value.replaceFirst("\\+", ".").replaceAll("\\+", "-")
+  )
+
+  // Need to specify scala.reflect.macros here because the bnd macro will not import this with the proper version
+  val parsing = osgiPkgHeaders(
+    exports = Seq("akka.parboiled2.*", "akka.shapeless.*", "akka.macros.*"),
+    imports = Seq("scala.reflect.macros", optionalResolution("scala.quasiquotes")))
+
+  val httpCore = osgiPkgHeaders(
+    exports = Seq("akka.http.*"),
```
jonas: Similar to http2Support, should `akka.http.impl` not be in `OsgiKeys.privatePackage`?

project/plugins.sbt
jonas: Yes, we can squash when merging so I wouldn't worry about that.

project/OSGi.scala
michael-smith: Generally I'd agree but unfortunately there are other bundles (akka-http itself, for instance) that reference the impl package from core.

project/OSGi.scala
```diff
@@ -0,0 +1,87 @@
...
+    imports = Seq(scalaJava8CompatImport, akkaParboiledImport, akkaShapelessImport, akkaMacrosImport))
+
+  val http = osgiPkgHeaders(
+    exports = Seq(
+      "akka.http.scaladsl.client.*",
+      "akka.http.$DSL$.server.*",
```
ktoso: Can we pick a different char than `$` here? This will cause scala compiler warnings that *"hey, did you forget to say `s"$DSL"`?"* Let's go with `#` maybe

michael-smith: Yes, replaced with `#DSL#`

project/OSGi.scala
```diff
@@ -0,0 +1,87 @@
...
+
+  private lazy val akkaParboiledImport = projectVersioned("akka.parboiled2.*")
+  private lazy val akkaShapelessImport = projectVersioned("akka.shapeless.*")
+  private lazy val akkaMacrosImport = projectVersioned("akka.macros.*")
+  private lazy val akkaImport = "akka," + patchVersioned("akka.*")
+  private lazy val configImport = patchVersioned("com.typesafe.config.*")
+  private lazy val scalaImport = minorVersioned("scala.*")
+  private lazy val scalaJava8CompatImport = """scala.compat.java8.*;version="$<range;[==,1)>""""
+
+  private def projectVersioned(pkg: String) = pkg + """;version="$<range;[===,=+);$<Bundle-Version>>""""
```
ktoso: hm I don't really understand those... very weird syntax hm

michael-smith: It's the bnd tool range syntax. This allows us to specify ranges without the need to manually enter what range versions we want for a specific library. Without this, it is maddening to keep up with the version changes on every release.

michael-smith:
Absolutely I will help confirm issues if need-be.

ktoso:
Great, thanks.

Ok, let's just do it then :shipit:

ktoso:
Resolves #574

Issue #1012, opened 2017-04-05
idle-timeout became broken between 10.0.4 and 10.0.5?


ktoso:
It seems we may have broken idle timeouts in 10.0.5, I will look into this immediately,

Notes: 

> It appears that this might indeed be a bug introduced during a change to make the transport API more pluggable. You can see that in v10.0.4 Tcp().outgoingConnection explicitly sets the idleTimeout, whereas in the current master it relies on that to be set by the passed in, pluggable transport. This would be ok as @ClientTransport.TCPTransport utilizes the passed in idleTimeout which the transport is utilized by the PoolInterfaceActor after having been created by the ConnectionSettings here. So, all that said, at a fair review it looks like the timeout should still be handled appropriately as everything seems to be instantiated at the correct times. So we will have to dig a bit deeper to determine if and how this pluggable change affected items such as timeout.

ktoso:
Working on it

jrudolph:
Ok, I see what the problem is here. It has nothing to do with passing the settings around but it's this particular usage of setting the idle-timeout which doesn't work any more:

```scala
val poolSettings = ConnectionPoolSettings(system).withConnectionSettings( 
  ClientConnectionSettings(system) 
    .withConnectingTimeout(6.seconds) 
    .withIdleTimeout(6.seconds) 
)
```

Setting the idle-timeout from `application.conf` should still work.

The reason that it doesn't work any more is that `ConnectionPoolSettings(system)` already configures the transport with the `idle-timeout` as given in the config file and `withConnectionSettings` doesn't make sure to apply a changed idle-timeout also to the transport.

ktoso:
I would still call that "passing settings around", but yeah - it's about that case.

> The reason that it doesn't work any more is that ConnectionPoolSettings(system) already configures the transport with the idle-timeout as given in the config file and withConnectionSettings doesn't make sure to apply a changed idle-timeout also to the transport.

Yes, which I think I was addressing in my PR, though it's been a week so lemme try to review it myself again and see if really fixing the root issue.

Issue #1009, opened 2017-04-01
Configuration options could use clarification on dependencies between them.


ShaneDelmore:
Config options around sizing are a bit confusing right now.  

Consider setting up 2 Akka-Http services, A and B, where A makes requests to B for streaming data, say a 20MB file.

Which options would you set?

Akka-Http Client on service A defaults to 4 connections, Akka-Http Server on service B defaults to serving 1024 connections.

You will run into a runtime error if testing with a 5MB file you will be surprised at runtime to find out that a 10MB file will not work due to Max-Content-Length.

What settings will you need to make sure this does not fail with chunk size incorrect errors?  What if someone sets the chunk size different on each service?

max-content-length?
decode-max-bytes-per-chunk?
max-chunk-size?

Which of these are dependent upon which and will break if they are set to incompatible settings?

Do they need to be set in:
akka.http.client?
akka.http.server?
akka.http.pool?
akka.http.pool.client?
akka.http.parsing?
akka.http.client.parsing?
akka.http.host-connection-pool.client.parsing?

If you set one and miss another will the service fail at runtime due to incompatible sizes or will it adjust/re-chunk dynamically?

Is there a best practice advice to verify all of these settings are correct and compatible?  For example always test with a file at least 2x the size of akka.http.foo.setting?

ktoso:
Thanks for the ticket, definitely a candidate for docs improvements - triaged.

ShaneDelmore:
After more exploration I believe the configuration options were not the core problem in the particular case I was dealing with, but that they were actually ignored.  I was able to find no configuration setting that actually limited the size of the chunks being created by Akka-Http server and eventually found the Chunker implementation in the Akka-Streams cookbook.  

This implementation is useful enough to include in the cookbook, it may be useful enough to include in akka-streams directly, or even better build in support for re-chunking according to maximum chunk size in Akka-Http directly.

jrudolph:
I think there's no general rule here. A server will usually have a number of clients so the constraints may be different ones than you face on the client. We might be able to improve chunking. If the user is not interested in the actual chunk boundaries (which they shouldn't be as intermediates are free to rechunk), then we can just deliver partial chunks without having to buffer. (Indefinite buffering is what the check is trying to prevent.)

PR #1014, opened 2017-04-06
Resolve timeout handling regression in 10.0.5 in new Transports style

ktoso:
Resolves https://github.com/akka/akka-http/issues/1012

Though we may want to discuss the entire settings design in depth rather soon.
See also: https://github.com/akka/akka-http/issues/1009

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/javadsl/ClientTransport.scala
+++ b/akka-http-core/src/main/scala/akka/http/javadsl/ClientTransport.scala
@@ -25,7 +25,7 @@ abstract class ClientTransport { outer ⇒
 
 object ClientTransport {
   def TCP(localAddress: Optional[InetSocketAddress], settings: ClientConnectionSettings): ClientTransport =
-    scaladsl.ClientTransport.TCP(localAddress.asScala, settings.asScala).asJava
+    ??? // scaladsl.ClientTransport.TCP(localAddress.asScala).asJava FIXME
 
   def fromScala(scalaTransport: scaladsl.ClientTransport): ClientTransport =
     scalaTransport match {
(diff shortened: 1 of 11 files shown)


akka-http-core/src/main/scala/akka/http/javadsl/ClientTransport.scala
```diff
@@ -39,15 +40,15 @@ object ClientTransport {
...
-      JavaMapping.toJava(delegate.connectTo(host, port)(system))
+      JavaMapping.toJava(delegate.connectTo(host, port, settings)(system))
     }
   }
   private class JavaWrapper(val delegate: ClientTransport) extends scaladsl.ClientTransport {
-    def connectTo(host: String, port: Int)(implicit system: ActorSystem): akka.stream.scaladsl.Flow[ByteString, ByteString, Future[scaladsl.Http.OutgoingConnection]] = {
+    def connectTo(host: String, port: Int, settings: ClientConnectionSettings)(implicit system: ActorSystem): akka.stream.scaladsl.Flow[ByteString, ByteString, Future[scaladsl.Http.OutgoingConnection]] = {
       import system.dispatcher
-      JavaMapping.toScala(delegate.connectTo(host, port, system))
+      ??? // JavaMapping.toScala(delegate.connectTo(host, port, system))
```
ktoso: fixme

jrudolph:
It's not clear to me (as it didn't seem to be to the original reporter) that something is broken here and what that should be. Can you explain?

This PR reverts the ClientTransport feature which I guess was unintended? Or was that the resolution?

It seems to contain some nice refactorings that could make sense in another PR.

ktoso:
Let's it around and let's discuss it next week - AFAIR the situation is that the setting was taken from the "initial", place and whatever people would be setting when calling outgoingConnection (by providing the settings in there) ends up being ignored with the current transports version.

Could be that my impl here is still not completely correct - there's a lot of dancing around the settings / passing and what it should actually mean in general..

ktoso:
@jrudolph could I ask you for a review picking the things/refactorings you'd like from this PR?

jrudolph:
I think I'll follow up on this one. It's me who broke it with the `ClientTransport`, so I'll have another look about how to fix it and I'll take the ideas from this one if possible.

ktoso:
Yeah go ahead and take it if you want :) Shall we close this one?

ktoso:
Just FYI: This is tagged as pick next, but in no shape to get into today's release.
Let me know if I should close it (afair you wanted to take it over @jrudolph)

jrudolph:
Superseded by #1195.

ktoso:
Great, thanks. Will review :)

Issue #1017, opened 2017-04-06
Problem with "super-marshaller" for various content types


Susak:
For example I have two marshallers for different content-types, which provide marshalling for MyClass to String.
I have run the following code. 
```
import akka.http.scaladsl.marshalling.Marshaller
import akka.http.scaladsl.model.{HttpRequest, MediaTypes}
import akka.http.scaladsl.model.headers.Accept

object MarshallerTest {
  def main(args: Array[String]): Unit = {
    import scala.concurrent.ExecutionContext.Implicits.global

    case class MyClass(value: String)

    implicit val superMarshaller = {
      val jsonMarshaller =
        Marshaller.stringMarshaller(MediaTypes.`application/json`)
          .compose[MyClass] { mc =>
          println(s"jsonMarshaller marshall $mc")
          mc.value
        }
      val textMarshaller = Marshaller.stringMarshaller(MediaTypes.`text/html`)
        .compose[MyClass] { mc =>
        println(s"textMarshaller marshall $mc")
        throw new IllegalArgumentException(s"Unexpected value $mc")
      }

      Marshaller.oneOf(jsonMarshaller, textMarshaller)
    }
    val request =
      HttpRequest(uri = "/test")
        .withHeaders(Accept(MediaTypes.`application/json`))
    akka.http.scaladsl.marshalling.Marshal(MyClass("test"))
      .toResponseFor(request) onComplete println
  }
}
```
Marshaller `jsonMarshaller` marshall for `application/json` MediaType and marshaller `textMarshaller` do same for  `text/html` MediaType.
I marshall `MyClass("test")` to response for `request` with `Accept: application/json` header, in my opinion here should choose `jsonMarshaller` and marshalling should be successful.
But before you find the right implementation we call apply method here https://github.com/akka/akka-http/blob/master/akka-http/src/main/scala/akka/http/scaladsl/marshalling/Marshal.scala#L42
and then apply value for both implementations https://github.com/akka/akka-http/blob/master/akka-http/src/main/scala/akka/http/scaladsl/marshalling/Marshaller.scala#L92
Can you please explain is it feature or bug?

ktoso:
It is how the current infra works, yes - it invokes both indeed.

There is a ticket to change it, if we can: https://github.com/akka/akka-http/issues/243

I've added some more docs about this in: https://github.com/akka/akka-http/pull/1018

ktoso:
I'll close this as docs issue, but generally speaking it's duplicate of 243.

[...]

PR #1018, opened 2017-04-06
=htp #1017 document marshalling behaviour of oneOf

ktoso:
=htp #1017 document marshalling behaviour of oneOf

### DIFF
--- a/akka-http/src/main/scala/akka/http/scaladsl/marshalling/Marshaller.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/marshalling/Marshaller.scala
@@ -102,6 +102,11 @@ object Marshaller
   /**
    * Helper for creating a "super-marshaller" from a number of "sub-marshallers".
    * Content-negotiation determines, which "sub-marshaller" eventually gets to do the job.
+   *
+   * Please note that all marshallers will actualy be invoked in order to get the Marshalling object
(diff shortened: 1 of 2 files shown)


akka-http/src/main/scala/akka/http/scaladsl/marshalling/Marshaller.scala
```diff
@@ -102,13 +102,23 @@ object Marshaller
   /**
    * Helper for creating a "super-marshaller" from a number of "sub-marshallers".
    * Content-negotiation determines, which "sub-marshaller" eventually gets to do the job.
+   * 
+   * Please note that both marshallers will actualy be invoked in order to get the Marshalling object
```
jlprat: Shouldn't be "all passed" instead of "both"?

Susak: It would be more correct.

jlprat:
I think the failure is just formatting

akka-http/src/main/scala/akka/http/scaladsl/marshalling/Marshaller.scala
ktoso: Thanks guys 👍

ktoso:
Amended

Issue #243, opened 2016-09-12
Trigger content negotiation failures before response generation


ktoso:
<a href="https://github.com/sirthias"><img src="https://avatars.githubusercontent.com/u/135347?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [sirthias](https://github.com/sirthias)**
_Monday Mar 02, 2015 at 10:22 GMT_
_Originally opened as https://github.com/akka/akka/issues/16974_

----

Currently the marshalling infrastructure only gets invoked once the actual response has been produced. This means that content negotiation failures are triggered late, after the backend has already processed the response. For safe methods (e.g. GET) this only results in the wasted work for retrieving the resource (which might already be significant), but for unsafe methods this could actually result in a state change that doesn't get signaled properly to the client.

Currently a marshaller is defined as

```scala
sealed abstract class Marshaller[-A, +B] extends (A ⇒ Future[List[Marshalling[B]]])
```

which means that switching to a "fail early" model does require modification to the very core of the marshaling infrastructure. However, I currently don't see why it shouldn't be possible.

ktoso:
<a href="https://github.com/sirthias"><img src="https://avatars.githubusercontent.com/u/135347?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [sirthias](https://github.com/sirthias)**
_Monday Mar 02, 2015 at 10:23 GMT_

----

/c @jrudolph

ktoso:
<a href="https://github.com/rkuhn"><img src="https://avatars.githubusercontent.com/u/470469?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [rkuhn](https://github.com/rkuhn)**
_Tuesday Jun 16, 2015 at 07:24 GMT_

----

not sure what to make of this: should this be regarded as “triaged”, and if yes on which milestone?

ktoso:
<a href="https://github.com/sirthias"><img src="https://avatars.githubusercontent.com/u/135347?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [sirthias](https://github.com/sirthias)**
_Tuesday Jun 16, 2015 at 10:11 GMT_

----

It's something that I'd keep as a general improvement point for the marshalling/unmarshalling infrastructure and as something that users can hang their hat onto, if they experience the problem.
My take is that it's not a high prio issue that we really need to schedule onto any particular milestone at this point.

ktoso:
[...]

PR #1019, opened 2017-04-07
[WIP] #1017 make marshaller composition more lazy

raboof:
When using `Marshalling.oneOf`, in some situations each marshaller is actually executed before picking the most suitable result.

While indeed we can't avoid allocating a `Marshalling` for each `Marshaller` (that would need a wider overhaul as described in #243), the actual marshalling can still be lazy at that point (https://github.com/akka/akka-http/blob/master/akka-http/src/main/scala/akka/http/scaladsl/marshalling/Marshaller.scala#L168).

It seems `Marshaller.compose` is forcing the early evaluation. This PR appears to confirm that, though it isn't obvious we can work that through without performing the ugly hacks in here.

### DIFF
--- a/akka-http/src/main/scala/akka/http/scaladsl/marshalling/Marshaller.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/marshalling/Marshaller.scala
@@ -126,14 +126,34 @@ object Marshaller
   /**
    * Helper for creating a synchronous [[Marshaller]] to content with a fixed charset from the given function.
    */
-  def withFixedContentType[A, B](contentType: ContentType)(marshal: A ⇒ B): Marshaller[A, B] =
-    strict { value ⇒ Marshalling.WithFixedContentType(contentType, () ⇒ marshal(value)) }
+  def withFixedContentType[A, B](contentType: ContentType)(marshal: A ⇒ B): Marshaller[A, B] = {
+    val f1 = (value: A) ⇒ Marshalling.WithFixedContentType(contentType, () ⇒ marshal(value))
+    val f2 = (_: ExecutionContext) ⇒ (a: A) ⇒ FastFuture.successful(f1(a) :: Nil)
+    new Marshaller[A, B] {
+      def apply(value: A)(implicit ec: ExecutionContext) =
+        try f2(ec)(value)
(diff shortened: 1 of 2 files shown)


akka-http/src/main/scala/akka/http/scaladsl/marshalling/Marshaller.scala
```diff
@@ -126,14 +126,34 @@ object Marshaller
...
    * Helper for creating a synchronous [[Marshaller]] to content with a fixed charset from the given function.
    */
-  def withFixedContentType[A, B](contentType: ContentType)(marshal: A ⇒ B): Marshaller[A, B] =
-    strict { value ⇒ Marshalling.WithFixedContentType(contentType, () ⇒ marshal(value)) }
+  def withFixedContentType[A, B](contentType: ContentType)(marshal: A ⇒ B): Marshaller[A, B] = {
+    val f1 = (value: A) ⇒ Marshalling.WithFixedContentType(contentType, () ⇒ marshal(value))
+    val f2 = (_: ExecutionContext) ⇒ (a: A) ⇒ FastFuture.successful(f1(a) :: Nil)
+    new Marshaller[A, B] {
+      def apply(value: A)(implicit ec: ExecutionContext) =
+        try f2(ec)(value)
```
jrudolph: Could you inline those functions? The whole execution is strict so preallocation those lambdas is not beneficial.

ktoso:
> While indeed we can't avoid allocating a Marshalling for each Marshaller (that would need a wider overhaul as described in #243), the actual marshalling can still be lazy at that point (https://github.com/akka/akka-http/blob/master/akka-http/src/main/scala/akka/http/scaladsl/marshalling/Marshaller.scala#L168).

Good analysis, thanks.

akka-http/src/main/scala/akka/http/scaladsl/marshalling/Marshaller.scala
ktoso: +1 plz :)

jrudolph:
@raboof can you inline those lambdas and merge after validation? Would be nice to have in the next release?

ktoso:
I'll merge and cleanup

ktoso:
Resolves #1017

PR #1029, opened 2017-04-13
[WIP] Potential marshaller improvements

jrudolph:
Based on @raboof's #1019, I went a step further to clean up marshaller constructors and apply the optimizations from #1019 also to other code paths. So far it's only a (binary incompatible) experiment.

In the medium term, it might make sense to make an ADT out of the classes introduced here (`DynamicMarshaller`, `FixedContentTypeMarshaller`, `OpenCharsetMarshaller`, and `OpaqueMarshaller`). That would allow us to further optimize the common simple combinations.


jrudolph:
Would be nice to include changes like that at a later point, closing for now.

ktoso:
I agree, good that we have the `stashed PR` thingy :)

PR #1031, opened 2017-04-13
Fix #377: HEAD requests for client: return an empty entity but with a Content-Length

### LINKED ISSUE #377: Allow access to Content-Length header in response to HEAD request
According to the [HTTP headers documentation](http://doc.akka.io/docs/akka/2.4.11/scala/http/common/http-model.html#HTTP_Headers), the `Content-Length` header is not added to the response headers collection. It is part of the HTTP entity. 

It seems to be a correct behaviour in case of a GET request. Unfortunately, it does not work at all in the case of a HEAD request. Some API (eg AWS S3) uses the HEAD request to obtain information about the GET request. In particular, the S3 API return the object `Content-Length` in response to a HEAD request.

According to the HTTP specification:

> The HEAD method is identical to GET except that the server MUST NOT return a message-body in the response. **The metainformation contained in the HTTP headers in response to a HEAD request SHOULD be identical to the information sent in response to a GET request**. This method can be used for obtaining metainformation about the entity implied by the request without transferring the entity-body itself. This method is often used for testing hypertext links for validity, accessibility, and recent modification.

I think the default behaviour should be to return all the response headers or at least be configurable.

EDIT:
I would like to describe the problem I am facing better. I would like to implement S3 client able to do a multipart download. It will be implemented by executing a set of GET request with the `Range` header to fetch a part of the object to download. 

To implement this mechanism, I need to obtain the full object length as a first step. In S3 it is doable by making a HEAD request to the same object, that would return all the metadata for the object itself by mean of HTTP response headers, included the `Content-Length`.

Akka HTTP client does strip the `Content-Length` header from the HTTP response in  [HttpMessageParser](https://github.com/akka/akka-http/blob/d0b7829dc6d7091cf106dd9b51c4980c82729e70/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpMessageParser.scala#L139) so it is impossible to access this information in any way.

Hope it will help.

erwan:
See https://github.com/akka/akka-http/issues/377 for more details: expose the `Content-Length` header for HEAD requests.

I implemented it the way @jrudolph proposed: https://github.com/akka/akka-http/issues/377#issuecomment-257572126

See https://github.com/akka/akka-http/pull/1031/files?w=1 for a shorter diff (without whitespaces).


jrudolph:
OK TO TEST

erwan:
I just added a commit to fix the test. Not sure what the policy is about rebase+push force during the review process so I left it as is, but it would probably be better to squash it in the previous one.

jlprat:
We can squash when merging 😊

jrudolph:
I think it looks quite good. Can you add a test for that fix to `LowLevelOutgoingConnectionSpec`?

erwan:
@jrudolph I looked at this file, but it doesn't seem the headers are parsed in that case - not just HEAD requests but GET or POST all have `List()` as headers. Am I missing something?

erwan:
@jrudolph the test is here

thunderkid:
Hi, I'm trying to test this fix by getting object metadata from aws S3. For this I'm following the S3 docs and sending a `HEAD` request, but I'm still not getting the `Content-Length` header in the response. Everything else is there (`Last-Modified` etc), but still not `Content-Length`.  My request is like this:  

```
HttpRequest(HttpMethod(HEAD),https://qcar-quakes.s3-us-west-2.amazonaws.com/quakes/1908/08,List(Host: qcar-quakes.s3-us-west-2.amazonaws.com, x-amz-date: 20170420T175911Z, x-amz-content-sha256: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855, Authorization: AWS4-HMAC-SHA256 Credential=AKIOURSENGNEQOZQQ/20170420/us-west-2/s3/aws4_request, SignedHeaders=host;x-amz-content-sha256;x-amz-date, Signature=a622bb25efb20970614c8d0cb33zdff9a320a87ad5d3e4d8bf9a11e374bcf),HttpEntity.Strict(none/none,ByteString()),HttpProtocol(HTTP/1.1))
```

But the response is 

```
HttpResponse(200 OK,List(x-amz-id-2: gu6S6PkyW7loIanTcdX4YJb0fbV6CmVagxVAdoXF+IDxaNjyY7NDkdX7GbOGi8/HrY67ia90Tso=, x-amz-request-id: CA86C292D97A70C7, Date: Thu, 20 Apr 2017 17:59:14 GMT, Last-Modified: Wed, 08 Feb 2017 01:40:21 GMT, ETag: "b5b7639ffbe6775f974a0f4dff76f87e", Accept-Ranges: bytes, Server: AmazonS3),HttpEntity.Default(text/html; charset=UTF-8,966 bytes total),HttpProtocol(HTTP/1.1))
```

No `Content-Length`. Should this be working with your changes or am I missing something?

erwan:
@thunderkid this looks correct, the Content-Length information is in the entity not with the rest of the headers (that's how the Akka HTTP model works)

jrudolph:
@thunderkid as @erwan said to get the length, you can use `response.entity.contentLengthOption`.

thereisnospoon:
Guys, 
As far as I understand this PR will fix handling of Content-Length for HEAD requests on client side.
But what about HEAD requests in akka-http on server side? Currently, I believe, there is no way to respond on HEAD request with empty body and non-zero Content-Length.

jrudolph:
@thereisnospoon it should be possible with `transparent-head-requests = off` and using `HttpEntity.Default(contentType, length, Source.empty)`.

thereisnospoon:
Thanks, I will try

akka-http-core/src/test/scala/akka/http/impl/engine/client/LowLevelOutgoingConnectionSpec.scala
```diff
@@ -869,6 +869,28 @@ class LowLevelOutgoingConnectionSpec extends AkkaSpec("akka.loggers = []\n akka.
...
+        """HEAD / HTTP/1.1
+          |Host: example.com
+          |User-Agent: akka-http/test
+          |
+          |""")
+      sendWireData(
+        """HTTP/1.1 200 OK
+          |Content-Length: 6
+          |
+          |abcdef
```
jrudolph: The response shouldn't have a body for a HEAD request but maybe add a content type instead.

akka-http-core/src/test/scala/akka/http/impl/engine/client/LowLevelOutgoingConnectionSpec.scala
```diff
@@ -869,6 +869,28 @@ class LowLevelOutgoingConnectionSpec extends AkkaSpec("akka.loggers = []\n akka.
...
+          |""")
+
+      val HttpResponse(StatusCodes.OK, _, HttpEntity.Default(_, contentLength, _), _) = expectResponse()
```
jrudolph: Let's also test that the returned entity is empty.

akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpResponseParser.scala
```diff
@@ -154,41 +154,51 @@ private[http] class HttpResponseParser(protected val settings: ParserSettings, p
...
-              StreamedEntityCreator { entityParts ⇒
-                val data = entityParts.collect { case EntityPart(bytes) ⇒ bytes }
-                HttpEntity.CloseDelimited(contentType(cth), HttpEntity.limitableByteSource(data))
-              }
-            }
+    if (statusCode.allowsEntity) {
+      contextForCurrentResponse.get.requestMethod match {
+        case HttpMethods.HEAD ⇒ clh match {
+          case Some(`Content-Length`(contentLength)) if contentLength > 0 ⇒
+            emitResponseStart(defaultEntity(cth, contentLength))
```
jrudolph: I think we need something extra here to create a `StrictEntityCreator` with a Default entity that has a `Source.empty`. That's because `StrictEntityCreator` will later correctly cancel the empty stream it gets created with.

thunderkid:
@jrudolph , @erwan , 
Sorry, I'm not understanding - `response.entity.contentLengthOption` is returning `Some(0)` from a `HEAD` call to S3, rather than the size of the S3 object, which is what AWS is actually returning in Content-Length.

akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpResponseParser.scala
erwan: I'm nut sure I understand how we should do it. It appears that `StrictEntity` calculates its `contentLength` from its body, so how can we create one with an empty body but non-null `contentLength`?

jrudolph: Yes, we need to change `StrictEntityCreator` to work with `Default` entities as well (as paradox as it sounds). What we want to have is a `StrictEntityCreator(HttpDefault(cth, contentLength, Source.empty)`.

erwan:
@jrudolph that's done, is it good for you?

jrudolph:
Looking good from first glance, will have deeper look soon.

erwan:
@jrudolph should be good now, what do you think about my last commit?

erwan:
thank you!

jrudolph:
Code itself LGTM now. I cleaned up a few things around the data structures and made sure that it cleanly compiles with current master. Will merge after validation.

Thanks a lot, @erwan.

jrudolph:
(Unrelated test failures, will fix)

jrudolph:
Fixed conflicts with master.

jrudolph:
Great one, @erwan, thanks for fixing this!

PR #1032, opened 2017-04-13
=htc use ByteArrayRendering for response rendering

jrudolph:
Small scale buffer copies are faster than fine-grained chained ByteStrings.


jrudolph:
PLS BUILD

akka-http-core/src/main/scala/akka/http/impl/engine/rendering/HttpResponseRendererFactory.scala
```diff
@@ -229,12 +229,18 @@ private[http] class HttpResponseRendererFactory(
                 renderEntityContentType(r, entity)
                 renderContentLengthHeader(data.length) ~~ CrLf
 
-                if (!noEntity) r ~~ data
+                val finalBytes = {
+                  if (!noEntity)
+                    if (data.size < r.remainingCapacity) (r ~~ data).asByteString
```
ktoso: nice

PR #1033, opened 2017-04-14
Add multiple file upload directive

adlawson:
Adds support for multiple file uploads with the same field name.
I've followed the same implementation and testing approach as `fileUpload`. I did think about changing the `fileUpload` implementation to take the head of `fileUploadAll` but it would remove any benefit `Sink.headOption` gives you over `Sink.seq`.

```scala
// scaladsl
def fileUploadAll(fieldName: String): Directive1[immutable.Seq[(FileInfo, Source[ByteString, Any])]]

// javadsl
def fileUploadAll(fieldName: String, inner: JFunction[JList[JMap.Entry[FileInfo, Source[ByteString, Any]]], Route]): Route
```

---

### Tasks
Following review at https://github.com/akka/akka-http/pull/1033#pullrequestreview-50296030 there are a few outstanding pieces of work that need to be done before merge
 - [x] Rename directives
 - [x] Separate destination file creation
 - [x] Refactor `Source(files).map` to just `files.map`
 - [x] Java Documentation
 - [x] Scala Documentation
 - [x] ~Deprecate `uploadedFile` directive(?)~ Todo in a following PR


lightbend-cla-validator:
Hi @adlawson,

akka-http/src/main/scala/akka/http/javadsl/server/directives/FileUploadDirectives.scala
```diff
@@ -34,6 +37,19 @@ abstract class FileUploadDirectives extends FileAndResourceDirectives {
   def fileUpload(fieldName: String, inner: BiFunction[FileInfo, Source[ByteString, Any], Route]): Route = RouteAdapter {
     D.fileUpload(fieldName) { case (info, src) ⇒ inner.apply(info, src.asJava).delegate }
   }
+
+  /**
+   * Collects each body part that is a multipart file as a tuple containing metadata and a `Source`
+   * for streaming the file contents somewhere. If there is no such field the request will be rejected.
+   */
+  def fileMultiUpload(fieldName: String, inner: JFunction[JList[JMap.Entry[FileInfo, Source[ByteString, Any]]], Route]): Route = RouteAdapter {
+    def toJava[F <: FileInfo](info: F): FileInfo = info
```
adlawson: I'm not super happy about this bit, but without it I'd need to `asInstanceOf` to make it work with `Map.Entry`.

jlprat:
OK TO TEST

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileUploadDirectives.scala
```diff
@@ -85,6 +86,32 @@ trait FileUploadDirectives {
...
+    entity(as[Multipart.FormData]).flatMap { formData ⇒
+      extractRequestContext.flatMap { ctx ⇒
+        implicit val mat = ctx.materializer
+        implicit val ec = ctx.executionContext
+
+        val partSource: Source[(FileInfo, Source[ByteString, Any]), Any] = formData.parts
+          .filter(part ⇒ part.filename.isDefined && part.name == fieldName)
+          .map(part ⇒ (FileInfo(part.name, part.filename.get, part.entity.contentType), part.entity.dataBytes))
+
+        val partF = partSource.runWith(Sink.seq[(FileInfo, Source[ByteString, Any])])
```
jrudolph: Unfortunately, this approach will not work if the file uploads are big enough. You try to collect ongoing sources for the parts into a seq. However, without reading the first source you will not get the second source, so that the `seq` will never complete and the stream will deadlock. This is an unfortunate consequence of the squential `multipart` format. It will work with smaller payloads that fit into all the buffers but if the whole request doesn't fit into the buffers it will deadlock.

I think we need to do the same that other web servers / frameworks are doing and buffer the incoming files on the disk so that we can drain the whole request body before giving out any data.

adlawson: Yeah that's a real shame. I've been trying to get this to work with larger files today and hitting timeouts, but couldn't trace the root of the problem. Super obvious when you think about it.

If you'd like I can continue working on it and submit updated changes, since it's something I need anyway?

jrudolph: Yes, of course. We'll be glad if you keep working on it. I think buffering to disk isn't completely trivial to do but would definitely be a worthwhile addition.

adlawson:
I've updated the implementation to buffer to disk and squashed the commits into one. The changes now include 2 directives, where the first depends on the second.

```scala
def fileMultiUpload(fieldName: String): Directive1[immutable.Seq[(FileInfo, Source[ByteString, Any])]]

def uploadedFiles(fieldName: String): Directive1[immutable.Seq[(FileInfo, File)]]
```

I had started to go down the route of a custom GraphStage to buffer to disk, just like Alpakka's [akka.stream.alpakka.s3.impl](https://github.com/akka/alpakka/blob/master/s3/src/main/scala/akka/stream/alpakka/s3/impl/DiskBuffer.scala) but it proved harder to implement and more complex than necessary.

Another solution I tried was `Flow.fromSinkAndSource` but it appears the Source doesn't wait for the Sink to fully complete, so this wouldn't work.

The final solution looks too simple to me, but I've tested it with multiple large files and they stream over perfectly fine. I suppose the only downside is that we effectively have to wait for all ByteString sinks to complete before we can begin sourcing again.

patrox:
hi @adlawson - I'm quite interested in the functionality from this PR, so I would like to ask if you are going to finish it ?

@jrudolph Can you please give me some overview of what has to be done in order to get this PR to a mergable state ?

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileUploadDirectives.scala
```diff
@@ -58,6 +59,37 @@ trait FileUploadDirectives {
     }
 
   /**
+   * Streams the bytes of the file submitted using multipart with the given field name into temporary files on disk.
+   * If there is an error writing to disk the request will be failed with the thrown exception, if there is no such
+   * field the request will be rejected. Stored files are cleaned up on exit but not on failure.
+   *
+   * @group fileupload
+   */
+  def uploadedFiles(fieldName: String): Directive1[immutable.Seq[(FileInfo, File)]] =
```
jrudolph: This needs a name which better describes what it does (maybe `storeUploadedFilesToTempFiles`? better ideas?). It should have a parameter to specify the directory to put the temporary files into.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileUploadDirectives.scala
```diff
@@ -58,6 +59,37 @@ trait FileUploadDirectives {
...
+    entity(as[Multipart.FormData]).flatMap { formData ⇒
+      extractRequestContext.flatMap { ctx ⇒
+        implicit val mat = ctx.materializer
+        implicit val ec = ctx.executionContext
+
+        val uploaded: Source[(FileInfo, File), Any] = formData.parts
+          .filter(part ⇒ part.filename.isDefined && part.name == fieldName)
+          .mapAsync(1) { part ⇒
+            val fileInfo = FileInfo(part.name, part.filename.get, part.entity.contentType)
+            val dest = File.createTempFile("akka-http-upload", ".tmp")
```
jrudolph: Maybe this part should be completely abstracted away, so that the user can choose the destination freely based on the file info.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileUploadDirectives.scala
```diff
@@ -85,6 +117,35 @@ trait FileUploadDirectives {
...
+   *
+   * @group fileupload
+   */
+  def fileMultiUpload(fieldName: String): Directive1[immutable.Seq[(FileInfo, Source[ByteString, Any])]] =
+    extractRequestContext.flatMap { ctx ⇒
+      implicit val mat = ctx.materializer
+      implicit val ec = ctx.executionContext
+
+      uploadedFiles(fieldName).flatMap { files ⇒
+        val partsSource: Source[(FileInfo, Source[ByteString, Any]), Any] = Source(files).map {
```
jrudolph: This can be even simpler: no need for `Source(files)` when source and target datastructures are a `Seq`. You can just map the `files` datastructure.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileUploadDirectives.scala
```diff
@@ -85,6 +117,35 @@ trait FileUploadDirectives {
...
     }
+
+  /**
+   * Collects each body part that is a multipart file as a tuple containing metadata and a `Source`
+   * for streaming the file contents somewhere. If there is no such field the request will be rejected.
+   * Files are buffered into temporary files on disk so in-memory buffers don't overflow.
+   *
+   * @group fileupload
+   */
+  def fileMultiUpload(fieldName: String): Directive1[immutable.Seq[(FileInfo, Source[ByteString, Any])]] =
```
jrudolph: Should be marked as `@ApiMayChange`.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileUploadDirectives.scala
jrudolph: Should be marked as `@ApiMayChange`.

jrudolph:
(And sorry for keeping it around for so long. It's not because the PR wasn't good, we were just busy.)

adlawson:
Thanks for the feedback. I stuck with similar names to the single file uploads, just for consistency, but I'm happy to change them to be more descriptive. I'll also get onto your other points like separating temp file construction etc. too.

Don't worry about the delay - I've been buried under a mountain of work too. I'll try to get this updated within the next week.

adlawson:
Hey @jrudolph, I've made some updates since your last review
 - Renamed `uploadedFiles` to `storeUploadedFiles` to better describe what it does
 - Separated out the temporary file creation
 - Added `@ApiMayChange` where necessary, and removed the mima-filter exclusions

I've also taken a couple of further steps
 - Renamed `fileMultiUpload` to `fileUploadAll` so "multi" isn't confused with "multipart"
 - Added a `storeUploadedFile` (singular) to have a consistent API with the new directive. This should replace `uploadedFile` (which still exists and behaves the same as before, but could be deprecated in favour of this new function)

Let me know what you think, and whether you agree with `storeUploadedFile` or not.

adlawson:
Oh, the build failed on
```
[error] (akka-http/*:mimaReportBinaryIssues) akka-http: Binary compatibility check failed!
```

I thought the purpose of the `@ApiMayChange` annotations was to remove the need for the mima-filter exclusions. Am I mistaken or have I missed something else?

ktoso:
> I thought the purpose of the @ApiMayChange annotations was to remove the need for the mima-filter exclusions. Am I mistaken or have I missed something else?

Yes, though that's the long term plan but we have not yet implemented that feature in MiMa - we're only getting ready for it in the Akka sources so far. So the exclusions still have to be done manually. The issue for it is here in case you'd want to help on it: https://github.com/typesafehub/migration-manager/issues/160

adlawson:
I can see one test failed, but it doesn't seem to be related to anything I've done

>[info] - should send CONNECT message with ProxyAuth header and then forward incoming messages *** FAILED ***
>[info]   "CONNECT akka.io:443 HTTP/1.1
>[info]   Host: akka.io
>[info]   Proxy-Authorization: Basic dGVzdFVzZXJuYW1lOnRlc3RQYXNzd29yZA==
>[info]   " did not start with substring "CONNECT akka.io:443 HTTP/1.1
>[info]   Host: akka.io
>[info]   Proxy-Authorization: Basic dGVzdFVzZXJuYW1lOnRlc3RQYXNzd29yZA==
>[info]   
>[info]   " (HttpsProxyGraphStageSpec.scala:55)

Does this often fail intermittently or expected to fail in certain circumstances and would be worth rerunning?

jlprat:
I think you just stumbled upon a new regression: https://github.com/akka/akka-http/issues/1319

PLS BUILD

jlprat:
PLS BUILD

adlawson:
@jlprat it looks like the failing test has been fixed in #1320. Is it safe to rerun the build?

ktoso:
Yes, re running

ktoso:
PLS BUILD

adlawson:
Awesome, I'll get on to the documentation.

adlawson:
Hey @jrudolph, I think I've finished the work and I've written documentation for these new directives. I've rebased my work with latest master as the build kept failing with doc formatting conflicts.

I've left the old `uploadedFile` directive as it is, but I think it should be marked for deprecation, either in this PR or a subsequent one. I'm not so sure about your deprecation procedures.

Also if you need me to, I can squash commits into nicer chunks for merge.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileUploadDirectives.scala
```diff
@@ -33,31 +35,74 @@ trait FileUploadDirectives {
...
+   */
+  @ApiMayChange
+  def storeUploadedFiles(fieldName: String, destFn: FileInfo ⇒ File): Directive1[immutable.Seq[(FileInfo, File)]] =
+    entity(as[Multipart.FormData]).flatMap { formData ⇒
+      extractRequestContext.flatMap { ctx ⇒
+        implicit val mat = ctx.materializer
+        implicit val ec = ctx.executionContext
+
+        val uploaded: Source[(FileInfo, File), Any] = formData.parts
+          .filter(part ⇒ part.filename.isDefined && part.name == fieldName)
```
jrudolph: You need to make sure to discard entities for parts that are ignored. Otherwise, big parts that are not files will block the whole stream. Would be nice to add a testcase for that one as well.

Maybe use

```scala
.mapConcat { part =>
  case x if filter(x) => x :: Nil
  case x => 
     x.entity.discardBytes()
     Nil
}
```

jrudolph:
And, yes, we should do the deprecation. But we can do it in a follow-up PR.

Would be nice to squash the commits into logically self-sustained pieced. If that's not possible we can also squash everything into one big change.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileUploadDirectives.scala
adlawson: Ah yep, I see what you mean 👍

adlawson:
I've added the `discardBytes` call you asked for, but I haven't been able to come up with a working test for it. The closest I had was observing completing promises then the streams are consumed, just like the tests in [akka/http/scaladsl/model/EntityDiscardingSpec.scala#L44-L53](https://github.com/akka/akka-http/blob/f3e3505/akka-http-core/src/test/scala/akka/http/scaladsl/model/EntityDiscardingSpec.scala#L44-L53), but the futures never complete. I also tried simply consuming streams again to trigger and catch an `IllegalStateException` but I was able to consume them all again fine, even the ones consumed to files. If you have any pointers I'll keep trying, but I'm at a loss right now.

In addition to the change, I've squashed the commits into more reasonable, safe, chunks of work.

jrudolph:
@adlawson looking very good now. I added a few commits with smallish cleanups and the deprecation of `uploadedFile`. Could you review and merge https://github.com/adlawson/akka-http/pull/1 into this PR (couldn't push to your branch)?

adlawson:
Changes look good, @jrudolph. Thanks for the updates.
All merged and ready to go 👍

ktoso:
Great, will give it a final read then :)

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileUploadDirectives.scala
```diff
@@ -32,28 +35,76 @@ trait FileUploadDirectives {
...
+  /**
+   * Streams the bytes of the file submitted using multipart with the given file name into a designated file on disk.
+   * If there is an error writing to disk the request will be failed with the thrown exception, if there is no such
+   * field the request will be rejected, if there are multiple file parts with the same name, the first one will be
+   * used and the subsequent ones ignored.
+   *
+   * @group fileupload
+   */
+  @ApiMayChange
+  def storeUploadedFile(fieldName: String, destFn: FileInfo ⇒ File): Directive[(FileInfo, File)] =
```
ktoso: nice name btw 👍

ktoso:
Resolved https://github.com/akka/akka-http/issues/1273 Thanks!

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileUploadDirectives.scala
```diff
@@ -32,28 +35,76 @@ trait FileUploadDirectives {
...
+          onSuccess(uploadedF)
+      }
+    }
 
-            case Success(uploaded) ⇒
-              provide(uploaded)
+  /**
+   * Streams the bytes of the file submitted using multipart with the given field name into designated files on disk.
+   * If there is an error writing to disk the request will be failed with the thrown exception, if there is no such
+   * field the request will be rejected. Stored files are cleaned up on exit but not on failure.
```
mary-element: Does storeUploadedFiles Directive reject request if field is empty? 
It seems there is no implementation of it

Issue #1034, opened 2017-04-17
Exception thrown when using HttpApp on SBT exit


jlprat:
HttpApp default's completion promise can be completed in two different ways, via IO and via JVM shutdown. The latter doesn't check if the promise is already completed before trying to complete it itself. This causes an exception to be thrown when the JVM is terminated:
```
[error] (shutdownHook1) java.lang.IllegalStateException: Promise already completed.
java.lang.IllegalStateException: Promise already completed.
	at scala.concurrent.Promise.complete(Promise.scala:49)
	at scala.concurrent.Promise.complete$(Promise.scala:48)
	at scala.concurrent.impl.Promise$DefaultPromise.complete(Promise.scala:183)
	at scala.concurrent.Promise.success(Promise.scala:82)
	at scala.concurrent.Promise.success$(Promise.scala:82)
	at scala.concurrent.impl.Promise$DefaultPromise.success(Promise.scala:183)
	at akka.http.scaladsl.server.HttpApp.$anonfun$waitForShutdownSignal$1(HttpApp.scala:137)
	at scala.sys.ShutdownHookThread$$anon$1.run(ShutdownHookThread.scala:34)
[trace] Stack trace suppressed: run last compile:run for the full output.
```

This only affects the Scala version, the Java API for completing promises doesn't throw in this case.

Issue #864, opened 2017-02-14
FAILED: ResponseRendererSpec -> The response preparation logic should properly render The 'Connection' header should be rendered correctly


jlprat:
https://jenkins.akka.io:8498/job/pr-validator-akka-http/743/testReport/junit/akka.http.impl.engine.rendering/ResponseRendererSpec/The_response_preparation_logic_should_properly_render_The__Connection__header_should_be_rendered_correctly/

jrudolph:
Probably a follow-up to https://github.com/akka/akka-http/issues/332

jrudolph:
https://jenkins.akka.io:8498/job/pr-validator-akka-http/799/testReport/junit/akka.http.impl.engine.rendering/ResponseRendererSpec/The_response_preparation_logic_should_properly_render_The__Connection__header_should_be_rendered_correctly/

note:
Happened again here: https://jenkins.akka.io:8498/job/pr-validator-akka-http/1092/

jrudolph:
https://jenkins.akka.io:8498/job/pr-validator-akka-http/1136/testReport/junit/akka.http.impl.engine.rendering/ResponseRendererSpec/The_response_preparation_logic_should_properly_render_The__Connection__header_should_be_rendered_correctly/

jlprat:
https://jenkins.akka.io:8498/job/pr-validator-akka-http/1350/testReport/junit/akka.http.impl.engine.rendering/ResponseRendererSpec/The_response_preparation_logic_should_properly_render_The__Connection__header_should_be_rendered_correctly/

jrudolph:
https://jenkins.akka.io:8498/job/pr-validator-akka-http/1448/testReport/junit/akka.http.impl.engine.rendering/ResponseRendererSpec/The_response_preparation_logic_should_properly_render_The__Connection__header_should_be_rendered_correctly/

PR #1035, opened 2017-04-17
Do not throw if promise is completed #1034

jlprat:
Issue: #1034
Do not throw if promise is already completed when called by the JVM shutdown hook

### DIFF
--- a/akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
@@ -17,7 +17,7 @@ import akka.http.scaladsl.settings.ServerSettings
 import com.typesafe.config.ConfigFactory
 
 import scala.concurrent.duration.Duration
-import scala.concurrent.{ Await, ExecutionContext, Future, Promise }
+import scala.concurrent.{ Await, ExecutionContext, Future, Promise, blocking }
 import scala.io.StdIn
 import scala.util.{ Failure, Success, Try }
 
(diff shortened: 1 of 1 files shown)


akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
```diff
@@ -134,7 +134,7 @@ abstract class HttpApp extends Directives {
   protected def waitForShutdownSignal(system: ActorSystem)(implicit ec: ExecutionContext): Future[Done] = {
     val promise = Promise[Done]()
     sys.addShutdownHook {
-      promise.success(Done)
+      promise.trySuccess(Done)
```
viktorklang: what about Line 141?
Should Line 140 be in a `blocking {}`?

jlprat: Line 141 could be also replaced with `trySuccess`.
I guess you are right about the `blocking { }` part, the whole purpose of this line is to block.

I'll apply the feedback

jlprat:
Welcome back!

jlprat:
PLS BUILD

jlprat:
Failure was: #864

PLS BUILD

ktoso:
LGTM again. Not much to debate here, merging. 
Thanks!

jrudolph:
@jlprat small note (while I'm sifting through commits to write release notes): it helps a lot if the commit line contains enough context to understand what was the subject of the change, like `HttpApp` in this case. That helps both with glancing through git logs when trying to find related changes and when writing release notes. Thanks ;)

jlprat:
Will provide more context on the commits from now on. No problem!

Issue #821, opened 2017-01-25
Prepare for Akka 2.5


jrudolph:
Akka 2.5 milestone is due soon. To support Akka 2.5 we'll have to figure out a few things:

 * Is Akka 2.5 binary compatible enough, that we can either build just with Akka 2.4 or just with Akka 2.5 and hope that it will work with the other version unchanged?
 * Otherwise, can we cross-compile from one source branch to both versions? 
   * How do we set this up for releasing?
   * How do we distinguish versions? By artifact name suffix (`akka-http_akka-2.5_2.11`) ? By version suffix?
 * Otherwise, do we have to maintain several branches? We really to avoid that.
 * As already shown in #820 we already cannot cross-compile documentation examples between 2.4 and 2.5 any more because of source-incompatible changes in `AbstractActor`. Do we have to maintain examples for multiple akka versions as well?

jrudolph:
As 2.5 should be compatible enough, it should be possible to stay on Akka 2.4 for the time being. We should run our tests against 2.5 binary, though. I created #826 to track this.

jrudolph:
For documentation examples, we should start to build these against Akka 2.5 as soon as Akka 2.5 is released. We'll then only show Java examples building against Akka 2.5 (but may want to add comments to the examples using `AbstractActor` to point people to Akka 2.5).

jlprat:
Should we create a specific issue for the java examples using actors?

jrudolph:
Yes, could you create one?

jlprat:
Done, https://github.com/akka/akka-http/issues/992

ktoso:
Thanks :)

lespea:
Just curious if there is a timeline for getting a release of akka-http out that is compatible with 2.5?  Tried running with 2.5 and it threw this exception (probably expected but figured I'd post an update just to be sure...)

```
Detected java.lang.NoSuchMethodError error, which MAY be caused by incompatible Akka versions on the classpath. Please note that a given Akka version MUST be the same across all modules of Akka that you are using, e.g. if you use akka-actor [2.5.0 (resolved from current classpath)] all other core Akka modules MUST be of the same version. External projects like Alpakka, Persistence plugins or Akka HTTP etc. have their own version numbers - please make sure you're using a compatible set of libraries.
Uncaught error from thread [default-akka.actor.default-dispatcher-4] shutting down JVM since 'akka.jvm-exit-on-fatal-error' is enabled for ActorSystem[default]
[...]

PR #1036, opened 2017-04-18
Update to Akka 2.5

ktoso:
Checked locally if there's any issues while digging through emails and PRs.
Seems passes locally.

Do we want to do this in a 10.1.0 bump?
Might be good to just signal that "minor akka version changed" even though no other changes required.

Only reason to get this out soon seems to be that someone reported an incompatibility when mixing 10.0.5 with 2.5 hm...

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -7,7 +7,7 @@ import scala.language.implicitConversions
 object Dependencies {
   import DependencyHelpers._
 
-  val akkaVersion = "2.4.17"
+  val akkaVersion = "2.5.0"
   val junitVersion = "4.12"
   val h2specVersion = "1.5.0"
   val h2specName = s"h2spec_${DependencyHelpers.osName}_amd64"


ktoso:
Refs https://github.com/akka/akka-http/issues/821

jrudolph:
But if we update, we also have the same problem when mixing in the other direction (as indicated in https://github.com/akka/akka-http/issues/821#issuecomment-293934198).

jlprat:
@ktoso If by reported incompatibility you mean what was mentioned on the user list, it's like @jrudolph said in another ticket, it was because of the wrong akka-streams. So, if we don't upgrade to akka 2.5 we might want to add a note on the migration page.

jlprat:
Obviously, the change itself LGTM (if we want to proceed)

ktoso:
Hm, I see - both instances of the issue were indeed that someone did not upgrade Akka Streams properly along with the rest.

You're right we should not change the dependency yet I think.
Thanks for coping with me, getting back into Akking mode this week ;-)

Issue #1037, opened 2017-04-18
FileInputStream and FileOutputStream are not GC friendly


patriknw:
https://www.cloudbees.com/blog/fileinputstream-fileoutputstream-considered-harmful

jrudolph:
Fixed in #1038.

PR #1038, opened 2017-04-18
Replace FileInputStream and FileOutputStream, #1037

patriknw:
* because they use finalize that is not gc friendly
* they were only used in tests, but good to stay away from them
  anyway

Refs #1037

### DIFF
--- a/akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/FileUploadDirectivesSpec.scala
+++ b/akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/FileUploadDirectivesSpec.scala
@@ -4,7 +4,8 @@
 
 package akka.http.scaladsl.server.directives
 
-import java.io.{ File, FileInputStream }
+import java.io.File
+import java.nio.file.Files
 
 import akka.http.scaladsl.model._
 import akka.http.scaladsl.server.{ MissingFormFieldRejection, RoutingSpec }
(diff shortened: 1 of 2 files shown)


akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/FileUploadDirectivesSpec.scala
```diff
@@ -147,7 +148,7 @@ class FileUploadDirectivesSpec extends RoutingSpec {
   }
 
   private def read(file: File): String = {
-    val in = new FileInputStream(file)
+    val in = Files.newInputStream(file.toPath)
```
viktorklang: Is there a reason why we open the file for every read?

viktorklang: Note that [this line](https://github.com/akka/akka-http/pull/1038/files#diff-e1d2ed324b8dda05ffd842932a01ce76L154) could corrupt the input if the last byte(s) are an unclosed UTF-8 sequence. (Also, I think it would be faster to pass in the [UTF-8](https://docs.oracle.com/javase/7/docs/api/java/nio/charset/StandardCharsets.html#UTF_8) Charset rather than the "UTF-8"-String there.

jrudolph: I think this whole method works only accidentally (if at all). It's only in tests, so maybe no need to fix it right now.

Issue #1043, opened 2017-04-20
URI query parse errors during parameter extraction throw an unhandled IllegalUriException


hgiddens:
If parsing the URI query fails during parameter extraction, the resulting `IllegalUriException` is unhandled, leading to (by default) a 500 being sent back to the client, rather than a 400 as might be expected.

For example, if we `curl -i http://localhost:8080/?a==` the following route (appropriately bound):

```scala
def route: Route =
  parameter('a) { _ =>
    complete(StatusCodes.OK)
  }
```

We see in the logs:

```text
/?a==
akka.http.scaladsl.model.IllegalUriException: Illegal query: Invalid input '=', expected part, '&' or 'EOI' (line 1, column 3): a==
  ^
```

The client receives:

```text
± curl -i 'http://localhost:8080/?a=='
HTTP/1.1 500 Internal Server Error
Server: akka-http/10.0.5
Date: Thu, 20 Apr 2017 14:58:37 GMT
Content-Type: text/plain; charset=UTF-8
Content-Length: 35

There was an internal server error.
```

PR #1044, opened 2017-04-20
Handle failure while parsing the URI in parameter extraction #1043

hgiddens:
This changes the behaviour so that rather than throwing an exception leading to an Internal Server Error, instead we reject with a `MalformedRequestContentRejection`. This is used instead of `MalformedQueryParameterRejection` as the parse failure doesn't give us information about the query parameter that led to the parse failure.

I hope the tests are sufficient, but I'm more than happy to add more if desired.

### DIFF
--- a/akka-http-tests/src/test/java/akka/http/javadsl/server/directives/ParameterDirectivesTest.java
+++ b/akka-http-tests/src/test/java/akka/http/javadsl/server/directives/ParameterDirectivesTest.java
@@ -35,7 +35,7 @@ public class ParameterDirectivesTest extends JUnitRouteTest {
     route
       .run(HttpRequest.create().withUri("/abc?stringParam=a=b"))
       .assertStatusCode(400)
-      .assertEntity("The request content was malformed:\nThe request's query string is invalid");
+      .assertEntity("The request content was malformed:\nThe request's query string is invalid: Illegal query: Invalid input '=', expected '+', query-char, 'EOI', '&' or pct-encoded (line 1, column 14): stringParam=a=b\n             ^");
   }
 
   @Test
--- a/akka-http-tests/src/test/java/akka/http/javadsl/server/directives/ParameterDirectivesTest.java
+++ b/akka-http-tests/src/test/java/akka/http/javadsl/server/directives/ParameterDirectivesTest.java
@@ -35,7 +35,7 @@ public class ParameterDirectivesTest extends JUnitRouteTest {
     route
       .run(HttpRequest.create().withUri("/abc?stringParam=a=b"))
       .assertStatusCode(400)
-      .assertEntity("The request content was malformed:\nThe request's query string is invalid: Illegal query: Invalid input '=', expected '+', query-char, 'EOI', '&' or pct-encoded (line 1, column 14): stringParam=a=b\n             ^");
+      .assertEntity("The request content was malformed:\nThe request's query string is invalid: stringParam=a=b");
   }
 
   @Test
--- a/akka-http-tests/src/test/java/akka/http/javadsl/server/directives/ParameterDirectivesTest.java
+++ b/akka-http-tests/src/test/java/akka/http/javadsl/server/directives/ParameterDirectivesTest.java
@@ -31,6 +31,11 @@ public class ParameterDirectivesTest extends JUnitRouteTest {
       .run(HttpRequest.create().withUri("/abc?stringParam=john"))
       .assertStatusCode(200)
       .assertEntity("john");
+
+    route
+      .run(HttpRequest.create().withUri("/abc?stringParam=a=b"))
+      .assertStatusCode(400)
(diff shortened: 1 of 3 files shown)


raboof:
OK TO TEST

hgiddens:
I've signed the CLA.

ktoso:
PLS BUILD

ktoso:
Refs https://github.com/akka/akka-http/issues/1043

ktoso:
CLA indeed signed, I checked. 

One more LGTM as a sanity check please @akka/akka-http-team

akka-http/src/main/scala/akka/http/scaladsl/server/directives/ParameterDirectives.scala
```diff
@@ -125,7 +125,10 @@ object ParameterDirectives extends ParameterDirectives {
       extractRequestContext flatMap { ctx ⇒
         import ctx.executionContext
         import ctx.materializer
-        handleParamResult(paramName, fsou(ctx.request.uri.query().get(paramName)))
+        Try(ctx.request.uri.query()) match {
+          case Success(query) ⇒ handleParamResult(paramName, fsou(query.get(paramName)))
+          case Failure(t)     ⇒ reject(MalformedRequestContentRejection("The request's query string is invalid", t))
```
jrudolph: Should we add the message to the string as well, at least if we know that it is a parser error, which we surface to the client at other places?

ktoso: I was thinking about the same thing, so looked at usages of the rejection and we don't really add the malformed thing in the message... Though now I realise that in other cases that would have been the entity payload, so likely too big to include in log. 

In this case though, so it's shorter at least should be... sounds good to include it :)

hgiddens: Done :)

akka-http/src/main/scala/akka/http/scaladsl/server/directives/ParameterDirectives.scala
```diff
@@ -125,7 +125,10 @@ object ParameterDirectives extends ParameterDirectives {
       extractRequestContext flatMap { ctx ⇒
         import ctx.executionContext
         import ctx.materializer
-        handleParamResult(paramName, fsou(ctx.request.uri.query().get(paramName)))
+        Try(ctx.request.uri.query()) match {
+          case Success(query) ⇒ handleParamResult(paramName, fsou(query.get(paramName)))
+          case Failure(t)     ⇒ reject(MalformedRequestContentRejection(s"The request's query string is invalid: ${t.getMessage}", t))
```
2beaucoup: "The request's query string is invalid:" looks a bit redundant.

akka-http-tests/src/test/java/akka/http/javadsl/server/directives/ParameterDirectivesTest.java
```diff
@@ -31,6 +31,11 @@ public void testStringParameterExtraction() {
       .run(HttpRequest.create().withUri("/abc?stringParam=john"))
       .assertStatusCode(200)
       .assertEntity("john");
+
+    route
+      .run(HttpRequest.create().withUri("/abc?stringParam=a=b"))
+      .assertStatusCode(400)
+      .assertEntity("The request content was malformed:\nThe request's query string is invalid: Illegal query: Invalid input '=', expected '+', query-char, 'EOI', '&' or pct-encoded (line 1, column 14): stringParam=a=b\n             ^");
```
2beaucoup: would be nice if we can combine these three phrases "The request content was malformed:\nThe request's query string is invalid: Illegal query"

also there should be a newline before `stringParam=a=b\n` so that the error marker makes sense (or just remove it?)

akka-http/src/main/scala/akka/http/scaladsl/server/directives/ParameterDirectives.scala
ktoso: Not `t.getMessage`, put `ctx.request.uri.query()` in the message.
Keep the text IMO, nicer than a raw error with no text.

ktoso: Meh, wait... mistake in my reotply

ktoso: Text should be:

`s"""The request's query string is invalid: ${x.request.uri.rawQueryString.getOrElse("")""", t`

hgiddens: Oops, that makes much more sense. Updated accordingly.

ktoso:
Good to merge @jrudolph @2beaucoup ?

jrudolph:
Thanks, @hgiddens.

Issue #983, opened 2017-03-20
[NEEDS REPRODUCER] cachedHostConnectionPool can yield BufferOverflowException while still draining cancelled requests.


jamesmulcahy:
@drexin and I have been looking into an issue we experienced where we ran into a BufferOverFlow exception when using a cachedHostConnectionPool.  Our software has its own concurrency limiter, which was enforcing that no more than 50 requests would be active at once, and despite having the max-connections and max-open-requests values configured to accommodate this (50, and 64 respectively), we still saw this exception. Here's one specific example, for reference:

```
akka.stream.BufferOverflowException: Exceeded configured max-open-requests value of [64]. This means that the request queue of this pool (HostConnectionPoolSetup(REDACTED_HOSTNAME,443,ConnectionPoolSetup(ConnectionPoolSettings(50,0,5,64,1,30 seconds,ClientConnectionSettings(Some(User-Agent: akka-http/10.0.4),10 seconds,1 minute,512,None,<function0>,List(),ParserSettings(2048,16,64,64,8192,64,8388608,256,10485760,Strict,RFC6265,true,Full,Error,Map(If-Range -> 0, If-Modified-Since -> 0, If-Unmodified-Since -> 0, default -> 12, Content-MD5 -> 0, Date -> 0, If-Match -> 0, If-None-Match -> 0, User-Agent -> 32),false,akka.stream.impl.ConstantFun$$$Lambda$711/1719727892@7bd99559,akka.stream.impl.ConstantFun$$$Lambda$711/1719727892@7bd99559,akka.stream.impl.ConstantFun$$$Lambda$712/1404150776@2e95f241))),akka.http.scaladsl.HttpsConnectionContext@6fcfc89e,akka.event.MarkerLoggingAdapter@2c4f232d))) has completely filled up because the pool currently does not process requests fast enough to handle the incoming request load. Please retry the request later. See http://doc.akka.io/docs/akka-http/current/scala/http/client-side/pool-overflow.html for more information.
```

We've tracked this problem down to a behavior in akka-http when handling stream cancellation.  In our use case, the entity streams were very large, and/or slow, and could take many hours to fetch.  In some cases, we'd cancel these streams mid way through, and expected that the HTTP request was also cancelled and that the request "slot" was immediately available for re-use.

It turns out that akka-http continues to stream the dataBytes on a cancelled connection, effectively using up request "slots" which the user might otherwise reasonably expect to be available to initiate a new request.

[...]

PR #1058, opened 2017-04-26
=htc #983 canceling connection, kills entity stream on that conn

ktoso:
Resolves https://github.com/akka/akka-http/issues/983

Key changes are making one of the merges eagerCancel - otherwise the code in onDownstreamFinished in the Prepare stage was actually never called anyway, and I changed the handling a bit - no more "delaying" and I think that the entity should be killed with an error if we cancelled the connection "under its feet" - please review the semantics, if these seem to make sense.

I'm unsure about the half closed things... If we just do a cancel, that does not imply closing the connection yet right? AFAICS the server continues to write now after the cancellation is propagated, is this fine or am I not done yet with the cancelation propagation?

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
@@ -15,7 +15,7 @@ import scala.concurrent.Promise
 import scala.collection.mutable.ListBuffer
 import akka.stream.TLSProtocol._
 import akka.util.ByteString
-import akka.event.LoggingAdapter
+import akka.event.{ Logging, LoggingAdapter }
 import akka.stream._
 import akka.stream.scaladsl._
 import akka.http.scaladsl.Http
@@ -24,9 +24,7 @@ import akka.http.scaladsl.model.{ HttpRequest, HttpResponse, IllegalResponseExce
 import akka.http.impl.engine.rendering.{ HttpRequestRendererFactory, RequestRenderingContext }
 import akka.http.impl.engine.parsing._
 import akka.http.impl.util._
-import akka.stream.stage.GraphStage
-import akka.stream.stage.GraphStageLogic
-import akka.stream.stage.{ InHandler, OutHandler }
+import akka.stream.stage._
 import akka.http.impl.util.LogByteStringTools._
 
 import scala.util.control.NoStackTrace
@@ -100,7 +98,7 @@ private[http] object OutgoingConnectionBlueprint {
         .mapConcat(ConstantFun.scalaIdentityFunction)
         .via(new PrepareResponse(parserSettings))
 
-      val terminationFanout = b.add(Broadcast[HttpResponse](2))
+      val terminationFanout = b.add(Broadcast[HttpResponse](2, eagerCancel = true))
 
       val logger = b.add(Flow[ByteString].mapError { case t ⇒ log.error(t, "Outgoing request stream error"); t }.named("errorLogger"))
       val wrapTls = b.add(Flow[ByteString].map(SendBytes))
@@ -176,16 +174,17 @@ private[http] object OutgoingConnectionBlueprint {
 
     val shape = new FlowShape(responseOutputIn, httpResponseOut)
 
-    override def createLogic(effectiveAttributes: Attributes) = new GraphStageLogic(shape) with InHandler with OutHandler {
+    override def createLogic(effectiveAttributes: Attributes) = new GraphStageLogic(shape) with InHandler with OutHandler with StageLogging {
+
       private var entitySource: SubSourceOutlet[ResponseOutput] = _
       private def entitySubstreamStarted = entitySource ne null
       private def idle = this
-      private var completionDeferred = false
       private var completeOnMessageEnd = false
 
-      def setIdleHandlers(): Unit =
-        if (completeOnMessageEnd || completionDeferred) completeStage()
+      def setIdleHandlers(): Unit = {
+        if (completeOnMessageEnd) completeStage()
         else setHandlers(responseOutputIn, httpResponseOut, idle)
+      }
 
       def onPush(): Unit = grab(responseOutputIn) match {
         case ResponseStart(statusCode, protocol, headers, entityCreator, closeRequested) ⇒
@@ -205,14 +204,11 @@ private[http] object OutgoingConnectionBlueprint {
       }
 
       override def onDownstreamFinish(): Unit = {
-        // if downstream cancels while streaming entity,
-        // make sure we also cancel the entity source, but
-        // after being done with streaming the entity
         if (entitySubstreamStarted) {
-          completionDeferred = true
-        } else {
-          completeStage()
+          val ex = new RuntimeException("Connection stream cancelled while still reading incoming entity!")
+          entitySource.fail(ex)
         }
+        completeStage()
       }
 
       setIdleHandlers()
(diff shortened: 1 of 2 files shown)


ktoso:
Failure was weird, on `johannes` box hm...

```

+ h2spec_junit_file=./akka-http2-support/target/test-reports/h2spec-junit.xml
+ sed -i 's/<failure /<skipped /g' ./akka-http2-support/target/test-reports/h2spec-junit.xml
sed: can't read ./akka-http2-support/target/test-reports/h2spec-junit.xml: No such file or directory
+ true
+ sed -i 's|</failure>|</skipped>|g' ./akka-http2-support/target/test-reports/h2spec-junit.xml
sed: can't read ./akka-http2-support/target/test-reports/h2spec-junit.xml: No such file or directory
+ true
+ sed -i 's|skipped=|SKIPPPPPED=|g' ./akka-http2-support/target/test-reports/h2spec-junit.xml
sed: can't read ./akka-http2-support/target/test-reports/h2spec-junit.xml: No such file or directory
```

ktoso:
PLS BUILD

ktoso:
Ah I'm blind, failure was `[error] Unformatted code found. Please run 'test:compile' and commit the reformatted code.`

akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
```diff
@@ -205,14 +204,8 @@ private[http] object OutgoingConnectionBlueprint {
...
       override def onDownstreamFinish(): Unit = {
-        // if downstream cancels while streaming entity,
-        // make sure we also cancel the entity source, but
-        // after being done with streaming the entity
-        if (entitySubstreamStarted) {
-          completionDeferred = true
-        } else {
-          completeStage()
-        }
+        if (entitySubstreamStarted) entitySource.complete()
```
ktoso: question: complete, as we do nowadays, or is this actually a fail... see the previous commit in this PR (completing is the current semantics expected in other tests)

akka-http-tests/src/test/scala/akka/http/scaladsl/OutgoingConnectionCancellationShouldKillEntityStreamSpec.scala
```diff
@@ -0,0 +1,122 @@
...
+import org.scalatest.concurrent.ScalaFutures
+
+import scala.concurrent.duration._
+import scala.concurrent.{ Await, Future, Promise }
+import scala.util.{ Success, Try }
+
+/**
+ * Spec for https://github.com/akka/akka-http/issues/983
+ */
+class OutgoingConnectionCancellationShouldKillEntityStreamSpec extends AkkaSpec with ScalaFutures
```
jrudolph: Please do not add one-off test setups. We have `LowLevelOutgoingConnectionSpec` for the low-level tests and `ConnectionPoolSpec` if you want to add a test that examines the pool behavior. These test files have all the necessary setup.

If you look through the tests in `LowLevelOutgoingConnectionSpec` there seem to be tests that test related behavior. It would be nice to understand what was going on when those tests were added.

jrudolph:
@ktoso have you looked at https://github.com/akka/akka/pull/19528, https://github.com/akka/akka/issues/18010, https://github.com/akka/akka/issues/18203, https://github.com/akka/akka/issues/19741, and https://github.com/akka/akka/pull/19742?

We should make sure we understood the rationale behind the changes and decisions taken back then (e.g. why did #19742 revert the functionality introduced in #18203 after two days even if it claims superior test coverage that apparently didn't prevent this bug but rather seemed to have introduced it).

jrudolph:
Especially, that we now might have four sets of tests (the original ones, the ones added in #19528, the ones added in #19742 and yours) for that same behavior that all succeeded all the way through even if the behavior might only now be the correct one.

ktoso:
Thanks for digging up those old issues - no, I have not investigated them this time.
I'll put this on halt and re-visit in depth with investigating all those on the next bug fixing day then.

akka-http-tests/src/test/scala/akka/http/scaladsl/OutgoingConnectionCancellationShouldKillEntityStreamSpec.scala
ktoso: Hm indeed it seems `close the connection if response entity stream has been cancelled` is about this behavior, I'll dig deeper to see what the difference really is between the test now and that one.

I do like starting out with a high level reproducer and working back from that. Yeah, I can remove the this test once I have made a lower level one that shows off the issue properly.

jrudolph:
It seems the bug was introduced here: https://github.com/akka/akka/pull/19742/files#diff-9d641a9b26bb92cd487f4e6c81832424L200 where the `onDownstreamFinished` callback was removed. Wouldn't the default implementation of `onDownstreamFinished` (I find the behavior of those SubSourceOutlet handlers really confusing) just shutdown the stage and therefore lead to cancellation of the connection?

akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
```diff
@@ -205,14 +204,8 @@ private[http] object OutgoingConnectionBlueprint {
       }
 
       override def onDownstreamFinish(): Unit = {
-        // if downstream cancels while streaming entity,
```
jrudolph: Do we actually need to change that? The previous behavior meant that cancelling the *response* stream would keep open the connection until the entity stream was done with. That seems sensible.

akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
```diff
@@ -100,7 +98,7 @@ private[http] object OutgoingConnectionBlueprint {
         .mapConcat(ConstantFun.scalaIdentityFunction)
         .via(new PrepareResponse(parserSettings))
 
-      val terminationFanout = b.add(Broadcast[HttpResponse](2))
+      val terminationFanout = b.add(Broadcast[HttpResponse](2, eagerCancel = true))
```
jrudolph: Is that change alone enough to fix the problem?

ktoso:
Back on this issue then, thanks for comments!

akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
ktoso: I just checked, it's not enough - the changes below and this change make the specific test pass.
I will look deeper into the semantics to see what to do with the changes below.

jrudolph:
I played a bit around with the existing tests and added another one for the pool. The existing test seems fine even when hardened. I then added a connection pool test which also works. There's one peculiarity that the server-side response stream doesn't seem to be cancelled when it is on the client side (but that could be another bug).

Here are the tests:

https://github.com/akka/akka-http/compare/master...jrudolph:pull/1058?expand=1

So far, from my side the status of the issue is "cannot reproduce".

From my perspective, your test is complex enough that I cannot tell from first glance if the problem your test shows is actually showing the issue or if it is a problem with your test. Taking my succeeding tests as another data point it would be nice to understand what your test does differently. Let's not spent more than another half day on this before asking for a reproducer on the ticket.

ktoso:
Thanks for digging more into it! I'll try to see what my test does differently, or if it's wrong indeed. Won't spend a full day on it, agreed.

Let's reconcile findings during daily standup about the status of that (that's my work-day-mid-point anyway) and we'll see what we do there then.

ktoso:
We conclude that things actually work as they should and we're unable to reproduce the issue.
We'll add a test covering the "reuse slot" as well.

Issue #463, opened 2016-11-02
Integrate with akka-http-core under a flag


jrudolph:
Advantages of merging into akka-http-core:
 * no code changes necessary to use HTTP/2 (when already using HTTPS)
 * this showcases our claim that HTTP/2 can be supported transparently
 * so, everyone can try HTTP/2 just by flipping a switch
 * "feature switches" are a best practice and seem to be successful in other large projects like the Linux kernel

Disadvantages:
 * people might get confused about what it means that stable akka-http-core contains an experimental implementation
 * for production-level HTTP/2 support users might want to add third-party components anyway using netty-openssl as an alternative, faster TLS backend

ktoso:
It's also a timing question. We can always "merge it in" once we're somewhat confident in it.
For now let's continue as separate experimental module, and once happy with it we can merge it in.

@patriknw once mentioned that releasing with `-experimental` was in general not that useful. Do you think we should not have the marker in the artifact name?

viktorklang:
If the feature is completely transparent to the end user, I see no issue having an experimental feature flag.

ktoso:
There's two things:

- flag in code to enable this (possibly not needed)
- marking the feature as experimental (and how to best go about this)

patriknw:
It's not about useful, it's about that we can't promote an experimental feature without releasing a new major version. The typical example is akka-distributed-data-experimental that has been ready for promotion for ages, but we can't do it until Akka 2.5.

The reason it can't/shouldn't be done in a patch release is that it's too easy to end up with multiple jar files (transitive dependencies) with the same classes and accidentally use the wrong version.

patriknw:
I think experimental should be in documentation and in opt-in config feature flag if that makes sense. The annotation we talked about elsewhere is also useful.

jrudolph:
> flag in code to enable this (possibly not needed)

The idea of directly merging it into akka-http-core is not to provide any extra entrypoints. The flag would be the only thing needed to enable Http/2 on HTTPS server bindings. It would be switched off by default and would come with documentation (e.g. in `reference.conf`) that the feature is still experimental.

Apart from that I think there's another dimension that we need to decide about when we actually maintain that code inside of the master branch. Maintaining an extra branch has some cost as well as maintaining code in the master branch (where PR validation runs all the time and it will be included in releases).

But maybe let's start with making it useful at all and then come back here ;)

ktoso:
> > flag in code to enable this (possibly not needed)
[...]

PR #1062, opened 2017-04-27
+htc #463 allow binding as HTTP/2 via configuration flag via Http()

ktoso:
Resolves https://github.com/akka/akka-http/issues/463
No docs or real integration test in this PR yet, I'll make tickets for that - verified locally though.

The name of the option is ok but I'm not in love with it, in case someone has a better name.


akka-http-core/src/main/resources/reference.conf
```diff
@@ -18,6 +18,15 @@ akka.http {
     # the request, no `Server` header will be rendered at all.
     server-header = akka-http/${akka.http.version}
 
+    # EXPERIMENTAL FEATURE: Bind using HTTP/2
```
patriknw: thought we got rid of the term experimental? This is experimental in the sense that we are uncertain of its production readiness? Could we find another word for that so that we don't make things even more confusing.

akka-http-core/src/main/scala/akka/http/scaladsl/Http2Shadow.scala
```diff
@@ -0,0 +1,75 @@
...
+        fm)
+        .asInstanceOf[Future[ServerBinding]]
+    } catch {
+      case ex: Throwable ⇒ throw Http2SupportNotPresentException(ex)
+    }
+  }
+
+  final case class Http2SupportNotPresentException(cause: Throwable)
+    extends RuntimeException("Unable to invoke HTTP2 binding logic (as enabled setting `akka.http.server.use-http2`). " +
+      """Please make sure that `com.typesafe.akka" %% "akka-http2-support" % AkkaHttpVersion` is on the classpath.""", cause)
```
patriknw: s/`com.typesafe.akka/"com.typesafe.akka/

ktoso:
Will fix MiMa

akka-http-core/src/main/resources/reference.conf
ktoso: Yes, it's not quite ready for production prime time, so let's use "preview" I think. Also not to revive the old experimental wording anymore - will change

akka-http-core/src/main/scala/akka/http/scaladsl/Http2Shadow.scala
ktoso: Oh thanks

ktoso:
Failure was https://github.com/akka/akka-http/issues/1063 and is unrelated

ktoso:
PLS BUILD

akka-http-core/src/main/resources/reference.conf
jrudolph: I'd suggest we have a whole section of `preview` configuration settings and put it there. I'd like people to see that they are enabling preview stuff. We can later deprecate those settings in favor of the real setting in a compatible way.

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
```diff
@@ -242,7 +242,8 @@ class HttpExt(private val config: Config)(implicit val system: ActorSystem) exte
     settings:          ServerSettings    = ServerSettings(system),
     parallelism:       Int               = 1,
     log:               LoggingAdapter    = system.log)(implicit fm: Materializer): Future[ServerBinding] =
-    bindAndHandle(Flow[HttpRequest].mapAsync(parallelism)(handler), interface, port, connectionContext, settings, log)
+    if (settings.useHttp2) Http2Shadow.bindAndHandleAsync(handler, interface, port, connectionContext, settings, parallelism, log)(fm)
```
jrudolph: Check `connectionContext` if https is enabled. If not, don't use Http/2.

ktoso: Oh, good idea - current impl is a bit aggresive and blows up if it's not, falling back is nicer indeed :)

akka-http-core/src/main/scala/akka/http/scaladsl/Http2Shadow.scala
```diff
@@ -0,0 +1,75 @@
...
+  def bindAndHandleAsync(
+    handler:   HttpRequest ⇒ Future[HttpResponse],
+    interface: String, port: Int,
+    httpsContext: ConnectionContext,
+    settings:     ServerSettings,
+    parallelism:  Int,
+    log:          LoggingAdapter)(implicit fm: Materializer): Future[ServerBinding] = {
+
+    val mat = ActorMaterializerHelper.downcast(fm)
+    require(httpsContext.isSecure, "In order to use HTTP/2 you MUST provide an HttpsConnectionContext.")
```
jrudolph: Ok, we could force it like that as well, but that would mean if you want to have both an HTTP and an HTTPS/Http2 server you need to have two sets of settings.

akka-http-core/src/main/scala/akka/http/scaladsl/Http2Shadow.scala
```diff
@@ -0,0 +1,75 @@
...
+import akka.http.scaladsl.settings.ServerSettings
+import akka.stream.{ ActorMaterializerHelper, Materializer }
+
+import scala.concurrent.Future
+
+/** INTERNAL API: Uses reflection to reach for Http2 support if available or fails with an exception */
+@InternalApi
+private[akka] object Http2Shadow {
+
+  type ShadowHttp2 = AnyRef
```
jrudolph: Just make these structural types. Then we can remove most of the reflection stuff, right?

ktoso: Interesting idea, I never used them like that - will try :)

jrudolph: In HTTP2 land we might be able to add a compilation test that does nothing more than

`Http2: Http2Shadow.ShadowHttp2`

so that we will catch if the signatures would get out of sync.

akka-http-core/src/main/scala/akka/http/scaladsl/Http2Shadow.scala
ktoso: I'm with with the "if https then http2" makes sense

ktoso:
PLS BUILD

akka-http-core/src/main/scala/akka/http/scaladsl/Http2Shadow.scala
ktoso: Added such test, sneaky and fun!

ktoso: (yes it works)

ktoso:
Uff, finally

akka-http-core/src/main/resources/reference.conf
```diff
@@ -18,6 +18,17 @@ akka.http {
...
+    # "PREVIEW" features that are not yet fully production ready.
+    # These flags can can change or be removed between patch releases.
+    preview {
+      # ONLY WORKS WITH `bindAndHandleAsync` (currently)
+      # 
+      # If this setting is enabled AND tha akka-http2-support is found 
+      # on the classpath the usual Http().bind... method calls will bind 
+      # using HTTP/2. Please note that you must configure HTTPS while doing so.
+      use-http2 = off
+    }
```
ktoso: sounds good @patriknw @akka/akka-http-team ?

akka-http-core/src/main/scala/akka/http/impl/settings/PreviewServerSettings.scala
```diff
@@ -0,0 +1,25 @@
...
+  /**
+   * Configures the Http extension to bind using HTTP/2 if given an
+   * [[akka.http.scaladsl.HttpsConnectionContext]]. Otherwise binds as plain HTTP.
+   *
+   * Please note that when using this mode of binding you MUST include
+   * `"com.typesafe.akka" %% "akka-http2-support" % AkkaHttpVersion` in your
+   * dependencies / classpath.
+   */
+  useHttp2: Boolean
+)
```
ktoso: probably fine to keep it as case class right?

akka-http-core/src/main/resources/reference.conf
jlprat: Typo `AND tha akka-http2-support`

ktoso: Fixed, thanks @jlprat

ktoso:
Should be ready to go.

I made the preview settings into an evolution friendly class as well. Please check naming and let's ask Play team to verify a bit during tomorrow (with which I'll assist).

ktoso:
Meh formatting error..

ktoso:
Formatting fixed and adjusted my code style in intellij to do the `import com.{ A, B }` style

akka-http-core/src/main/resources/reference.conf
```diff
@@ -18,6 +18,17 @@ akka.http {
...
 
+    # "PREVIEW" features that are not yet fully production ready.
+    # These flags can can change or be removed between patch releases.
+    preview {
+      # ONLY WORKS WITH `bindAndHandleAsync` (currently)
+      # 
+      # If this setting is enabled AND the akka-http2-support is found 
+      # on the classpath the usual Http().bind... method calls will bind 
+      # using HTTP/2. Please note that you must configure HTTPS while doing so.
+      use-http2 = off
```
jrudolph: How about `enable-http2`? Otherwise, it sounds a bit as if Http/2 would be used in any case (but it is only used when the client supports it). No strong feelings about it, though.

ktoso: enable sounds good, will change - let me know if other changes too or LGTM and I'll push the change :)

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
```diff
@@ -241,8 +241,19 @@ class HttpExt(private val config: Config)(implicit val system: ActorSystem) exte
...
-    log:               LoggingAdapter    = system.log)(implicit fm: Materializer): Future[ServerBinding] =
-    bindAndHandle(Flow[HttpRequest].mapAsync(parallelism)(handler), interface, port, connectionContext, settings, log)
+    log:               LoggingAdapter    = system.log)(implicit fm: Materializer): Future[ServerBinding] = {
+    val useHttp2 = settings.previewServerSettings.useHttp2
+    if (useHttp2) {
+      if (connectionContext.isSecure) {
+        log.debug("Binding server using HTTP/2...")
+        Http2Shadow.bindAndHandleAsync(handler, interface, port, connectionContext, settings, parallelism, log)(fm)
+      } else {
+        log.debug("The preview.use-http2 flag was set, but a plain HttpConnectionContext (not Https) was given, binding using plain HTTP...")
```
jrudolph: Good logging :+1:

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
```diff
@@ -241,8 +241,19 @@ class HttpExt(private val config: Config)(implicit val system: ActorSystem) exte
...
+        bindAndHandle(Flow[HttpRequest].mapAsync(parallelism)(handler), interface, port, connectionContext, settings, log)
```
jrudolph: Can we DRY this up with below?

akka-http-core/src/main/scala/akka/http/scaladsl/Http2Shadow.scala
```diff
@@ -0,0 +1,70 @@
...
+  type ShadowHttp2Ext = {
```
jrudolph: :+1:

akka-http-core/src/main/mima-filters/10.0.5.backwards.excludes
```diff
@@ -1,7 +1,27 @@
...
 ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.javadsl.model.HttpResponse.encoding")
 ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.scaladsl.model.HttpMessage.transformEntityDataBytes")
 
 # Internal API changes
-ProblemFilters.exclude[IncompatibleMethTypeProblem]("akka.http.impl.engine.rendering.RenderSupport.renderByteStrings")
\ No newline at end of file
+ProblemFilters.exclude[IncompatibleMethTypeProblem]("akka.http.impl.engine.rendering.RenderSupport.renderByteStrings")
+
+# HTTP/2 mode flag
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.javadsl.settings.ServerSettings.useHttp2")
```
jrudolph: Left-over from before?

akka-http-core/src/main/scala/akka/http/scaladsl/Http2Shadow.scala
```diff
@@ -0,0 +1,70 @@
+/*
+ * Copyright (C) 2009-2017 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.http.scaladsl
```
jrudolph: Can we move this internal class into `impl`?

akka-http2-support/src/test/scala/akka/http/scaladsl/Http2BindingViaConfigSpec.scala
```diff
@@ -0,0 +1,82 @@
...
+
+        // TODO we currently don't verify it really bound as h2, since we don't have a client lib
+        fishForDebugMessage(p, "Binding server using HTTP/2")
+      } finally if (binding ne null) binding.map(_.unbind())
+    }
+  }
+
+  "Http2Shadow" should {
+    "have ShadowHttp2Ext type aligned with real Http2Ext" in {
+      val ext: Http2Ext = Http2()
```
jrudolph: :+1:

akka-http2-support/src/test/scala/akka/http/scaladsl/Http2ServerTest.scala
```diff
@@ -67,7 +67,8 @@ object Http2ServerTest extends App {
       } yield (binding1, binding2)
 
     Await.result(bindings, 1.second) // throws if binding fails
-    println("Server online at http://localhost:9001")
+    println("Server (HTTP/1.1) online at http://localhost:9000")
+    println(Console.BOLD + "Server (HTTP/2) online at http://localhost:9001" + Console.RESET)
```
jrudolph: hehe, why not ;)

jrudolph: We can try to change the example above to use the common `Http()` api and set the flag in code.

akka-http-core/src/main/scala/akka/http/scaladsl/Http2Shadow.scala
ktoso: Yes, will move - sorry for merge before LGTM

PR #1069, opened 2017-04-28
Some more lambda inlining (#1019)

raboof:
I have this nagging feeling this could be made more elegant by doing something
other than making these anonymous `Marshaller` subclasses and overriding
'compose', but for now this'll do :).


project/plugins.sbt
```diff
@@ -19,7 +19,7 @@ addSbtPlugin("com.eed3si9n" % "sbt-unidoc" % "0.3.3")
 addSbtPlugin("com.thoughtworks.sbt-api-mappings" % "sbt-api-mappings" % "1.0.0")
 addSbtPlugin("pl.project13.scala" % "sbt-jmh" % "0.2.16")
 addSbtPlugin("pl.project13.sbt" % "sbt-jol" % "0.1.1")
-addSbtPlugin("net.virtual-void" % "sbt-dependency-graph" % "0.8.0") // for advanced PR validation features
+addSbtPlugin("net.virtual-void" % "sbt-dependency-graph" % "0.8.2") // for advanced PR validation features
```
raboof: Oops this slipped into this PR... but shouldn't hurt either :)

ktoso: Yeah, that's fine :)

Issue #953, opened 2017-03-09
HttpApp.route perhaps should be called routes?


ktoso:
I somehow feel `routes` would be more natural.

As one would say "and here we define the routes".

WDYT?

ktoso:
https://github.com/akka/akka-http/blob/master/akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala#L149

jlprat:
I have no strong opinion, I agree `routes` is more natural and `route` maybe is closer to the type. But I tend to agree with you.

We could maybe just deprecate the `route` method, instead of removing it, just to keep some source compatibility.

viktorklang:
routesTo?

-- 
Cheers,
√

jlprat:
Any further opinions? just rename it to `routes`?

ktoso:
Yes, just add a method `routes` and make the `route` deprecated I guess

PR #1070, opened 2017-04-28
Provide routes method in HttpApp #953

jlprat:
Issue: #953
Provide routes method in HttpApp
Deprecate route mothod in HttpApp
Use the new method in tests and docs
Update docs
HttpApp still uses the route method internally to ease the update

### DIFF
--- a/akka-http/src/main/java/akka/http/javadsl/server/HttpApp.java
+++ b/akka-http/src/main/java/akka/http/javadsl/server/HttpApp.java
@@ -73,7 +73,7 @@ public abstract class HttpApp extends AllDirectives {
 
     CompletionStage<ServerBinding> bindingFuture = Http
       .get(theSystem)
-      .bindAndHandle(routes().flow(theSystem, materializer), // uses the deprecated method to have an easy migration.
+      .bindAndHandle(routes().flow(theSystem, materializer),
         ConnectHttp.toHost(host, port),
         settings,
         theSystem.log(),
--- a/akka-http/src/main/java/akka/http/javadsl/server/HttpApp.java
+++ b/akka-http/src/main/java/akka/http/javadsl/server/HttpApp.java
@@ -73,7 +73,7 @@ public abstract class HttpApp extends AllDirectives {
 
     CompletionStage<ServerBinding> bindingFuture = Http
       .get(theSystem)
-      .bindAndHandle(route().flow(theSystem, materializer), // uses the deprecated method to have an easy migration.
+      .bindAndHandle(routes().flow(theSystem, materializer), // uses the deprecated method to have an easy migration.
         ConnectHttp.toHost(host, port),
         settings,
         theSystem.log(),
@@ -161,19 +161,7 @@ public abstract class HttpApp extends AllDirectives {
 
   /**
    * Override to implement the route that will be served by this http server.
-   * @deprecated This method is deprecated please use `routes` instead.
    */
-  @Deprecated
-  protected Route route() {
-    return routes();
-  }
-
-
-  /**
-   * Override to implement the route that will be served by this http server.
-   */
-  protected Route routes() {
-    throw new RuntimeException("Please override `routes` to provide the route that will be served by this http server");
-  }
+  protected abstract Route routes();
 
 }
--- a/akka-http/src/main/java/akka/http/javadsl/server/HttpApp.java
+++ b/akka-http/src/main/java/akka/http/javadsl/server/HttpApp.java
@@ -73,7 +73,7 @@ public abstract class HttpApp extends AllDirectives {
 
     CompletionStage<ServerBinding> bindingFuture = Http
       .get(theSystem)
-      .bindAndHandle(route().flow(theSystem, materializer),
+      .bindAndHandle(route().flow(theSystem, materializer), // uses the deprecated method to have an easy migration.
         ConnectHttp.toHost(host, port),
         settings,
         theSystem.log(),
@@ -161,8 +161,19 @@ public abstract class HttpApp extends AllDirectives {
 
   /**
    * Override to implement the route that will be served by this http server.
+   * @deprecated This method is deprecated please use `routes` instead.
    */
-  protected abstract Route route();
+  @Deprecated
+  protected Route route() {
+    return routes();
+  }
 
+
+  /**
+   * Override to implement the route that will be served by this http server.
+   */
+  protected Route routes() {
+    throw new RuntimeException("Please override `routes` to provide the route that will be served by this http server");
+  }
 
 }
--- a/akka-http/src/main/mima-filters/10.0.6.backwards.excludes
+++ b/akka-http/src/main/mima-filters/10.0.6.backwards.excludes
@@ -1,0 +1,4 @@
+# Provide routes method in HttpApp #953
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.javadsl.server.HttpApp.routes")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.scaladsl.server.HttpApp.routes")
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.scaladsl.server.HttpApp.route")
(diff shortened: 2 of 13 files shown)


jlprat:
General note about this solution:
The new `routes` method is implemented by default by throwing an exception. This is to make old code compile when updating akka-http version. Otherwise, any old class extending `HttpApp` would fail to compile because of the missing `routes` method.
Once `route` is removed, we could go back to make `routes` abstract.

jlprat:
Failure was: #1071

PLS BUILD

ktoso:
I'm not sure we should be so caucious with this refactor hm... (the throwing etc), it is an `@ApiMayChange` class for that reason - to change it

jlprat:
If we go that way, then probably makes no sense deprecate the old method, but just rename it. As it won't compile anyway...

jlprat:
Deleted the old method `route` and made `routes` abstract. If you prefer to still have the old `route` as deprecated, let me know. But I'm not sure it would be beneficial, as the code will compile but not work at all.

ktoso:
Yes this looks good, I think we need to start writing a migration guide for such things?
Perhaps we should just add a section for 10.0.6 in release notes about `Changes in @ApiMayChange  APIs` and add that rename there, sounds good?

jlprat:
Sounds good!
I would add it in both places, so it's easy to find.

I'll add a note on the migration guide, for this.

jlprat:
Failure was #492

PLS BUILD

jlprat:
New page for `ApiMayChange` under migration section

docs/src/main/paradox/java/http/migration-guide/index.md
```diff
@@ -7,5 +7,6 @@
 
 * [migration-from-old-http-javadsl](migration-from-old-http-javadsl.md)
 * [migration-guide-2.4.x-10.0.x](migration-guide-2.4.x-10.0.x.md)
+* [migration-within-10.0.x-Api-May-Change](migration-within-10.0.x-Api-May-Change.md)
```
ktoso: exactly, thanks!

docs/src/main/paradox/java/http/migration-guide/migration-within-10.0.x-Api-May-Change.md
```diff
@@ -0,0 +1,14 @@
+# Migration Guide within Akka HTTP 10.0.x
+
+## General Notes
+Akka HTTP is binary backwards compatible within for all version within the 10.0.x range. However, there are a set of APIs
+that are marked with the special annotation `@ApiMayChange` where this binary backwards compatibility is not guaranteed.
```
ktoso: are marked with the special annotation `@ApiMayChange` where this binary backwards compatibility is not guaranteed.

=>

are marked with the special annotation `@ApiMayChange` which are exempt from this rule, in order to allow them to be evolved more freely until stabilising them, by removing this annotation.

?

docs/src/main/paradox/java/http/migration-guide/migration-within-10.0.x-Api-May-Change.md
```diff
@@ -0,0 +1,14 @@
...
+See @extref:[The @DoNotInherit and @ApiMayChange markers](akka-docs:common/binary-compatibility-rules.html#The_@DoNotInherit_and_@ApiMayChange_markers) in the Akka documentation for further information.
```
ktoso: ~~in the Akka documentation~~

We are in the docs already so no need to explain that it'll be in the docs

docs/src/main/paradox/java/http/migration-guide/migration-within-10.0.x-Api-May-Change.md
```diff
@@ -0,0 +1,14 @@
...
+
+## Akka HTTP 10.0.x -> 10.0.6
```
ktoso: `10.0.5 -> 10.0.6` instead?

ktoso: i.e. the exact place where the change was made

docs/src/main/paradox/java/http/migration-guide/migration-within-10.0.x-Api-May-Change.md
```diff
@@ -0,0 +1,14 @@
+# Migration Guide within Akka HTTP 10.0.x
+
+## General Notes
+Akka HTTP is binary backwards compatible within for all version within the 10.0.x range. However, there are a set of APIs
+that are marked with the special annotation `@ApiMayChange` where this binary backwards compatibility is not guaranteed.
+See @extref:[The @DoNotInherit and @ApiMayChange markers](akka-docs:common/binary-compatibility-rules.html#The_@DoNotInherit_and_@ApiMayChange_markers) in the Akka documentation for further information.
+
```
ktoso: Needs one additional paragraph:

```
This migration guide aims to help developers who use these bleeding-edge APIs to migrate between their 
evolving versions within patch releases.
```

ktoso: Or propose a rewording, something like that however seems to be needed

docs/src/main/paradox/java/http/migration-guide/migration-within-10.0.x-Api-May-Change.md
```diff
@@ -0,0 +1,14 @@
...
+## Akka HTTP 10.0.x -> 10.0.6
+
+### `AkkaHttp#route` has been renamed to `AkkaHttp#routes`
+
+In order to provide a more descriptive name, `AkkaHttp#route` has been renamed to `AkkaHttp#routes`. The previous name
+might have led to some confusion by wrongly implying that only one route could be returned in that method.
+To migrate to 10.0.6, you must rename the old `route` method to `routes`.
```
ktoso: good

jlprat:
Sure, not so nitpicky, it's the first instance of this, and we should define the standard level of communication :)

docs/src/main/paradox/java/http/migration-guide/migration-within-10.0.x-Api-May-Change.md
jlprat: Actually it's 10.0.6 -> 10.0.7!

ktoso:
LGTM from my side, please have a look @jrudolph since that's the first instance we document / mini-migration-guide a change to ApiMayChange things

akka-http/src/main/mima-filters/10.0.6.backwards.excludes
```diff
@@ -0,0 +1,4 @@
+# Provide routes method in HttpApp #953
```
jrudolph: I think with the comments in the Mima file we should mostly describe why the exclusions are valid, i.e. in that case because the API in question is marked as `@ApiMayChange`.

akka-http/src/main/java/akka/http/javadsl/server/HttpApp.java
```diff
@@ -73,7 +73,7 @@ public void startServer(String host, int port, ServerSettings settings, Optional
 
     CompletionStage<ServerBinding> bindingFuture = Http
       .get(theSystem)
-      .bindAndHandle(route().flow(theSystem, materializer),
+      .bindAndHandle(routes().flow(theSystem, materializer), // uses the deprecated method to have an easy migration.
```
jrudolph: I don't understand the comment.

jlprat: Needs to be gone! I'll kill it

akka-http/src/main/mima-filters/10.0.6.backwards.excludes
jlprat: :+1:

jlprat:
Failure was unrelated. I'd say a similar incarnation of https://github.com/akka/akka-http/issues/299

PLS BUILD

jrudolph:
Great, thanks, @jlprat.

Issue #886, opened 2017-02-21
HTTP/2: MemoryLeak in AlpnSwitcher


jrudolph:
AlpnSwitcher currently doesn't propagate connection closure to inner substreams. This leads to the implementation staying alive after the connection has been closed.

ktoso:
Taking a stab at it

PR #1074, opened 2017-05-01
=h2 #886 fix leaks in ALPN switcher

ktoso:
This technically is very similar to CoupledTerminationFlow,
however since it's "sub" sources/sinks we could not directly use that.

Resolves https://github.com/akka/akka-http/issues/886

I may have overdone the binding of completions/cancels, however I'd rather do it this way now than omit a case. The important one is when downstream cancels net AFAIR - that is enough to make curl testing not leak.

### DIFF
--- a/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/AlpnSwitch.scala
+++ b/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/AlpnSwitch.scala
@@ -8,9 +8,9 @@ import javax.net.ssl.SSLException
 import akka.NotUsed
 import akka.http.scaladsl.model.{ HttpRequest, HttpResponse }
 import akka.stream.TLSProtocol.{ SessionBytes, SessionTruncated, SslTlsInbound, SslTlsOutbound }
-import akka.stream.scaladsl.{ BidiFlow, Flow }
+import akka.stream.scaladsl.{ BidiFlow, Flow, GraphDSL, Keep, Sink, Source }
 import akka.stream.stage.{ GraphStage, GraphStageLogic, InHandler, OutHandler }
-import akka.stream.{ Attributes, BidiShape, Inlet, Outlet }
+import akka.stream._
 
 object AlpnSwitch {
   type HttpServerBidiFlow = BidiFlow[HttpResponse, SslTlsOutbound, SslTlsInbound, HttpRequest, NotUsed]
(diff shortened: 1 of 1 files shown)


akka-http2-support/src/main/scala/akka/http/impl/engine/http2/AlpnSwitch.scala
```diff
@@ -75,11 +82,15 @@ object AlpnSwitch {
...
           }
 
-          // helpers to connect inlets and outlets
+          // helpers to connect inlets and outlets also binding completion signals of given ports
           def connect[T](in: Inlet[T], out: SubSourceOutlet[T], initialElement: Option[T]): Unit = {
             val propagatePull =
               new OutHandler {
-                def onPull(): Unit = pull(in)
+                override def onPull(): Unit = pull(in)
+                override def onDownstreamFinish(): Unit = {
```
jrudolph: I don't think these are necessary. No need to complete an `OutHandler`, that was just cancelled. Or am I missing something?

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/AlpnSwitch.scala
```diff
@@ -89,28 +100,59 @@ object AlpnSwitch {
...
                 case None ⇒ propagatePull
               }
 
             out.setHandler(firstHandler)
             setHandler(in, new InHandler {
-              def onPush(): Unit = out.push(grab(in))
+              override def onPush(): Unit = out.push(grab(in))
+
+              override def onUpstreamFinish(): Unit = {
+                out.complete()
```
jrudolph: I think that was the leaky one.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/AlpnSwitch.scala
```diff
@@ -89,28 +100,59 @@ object AlpnSwitch {
...
-            in.setHandler(new InHandler {
-              def onPush(): Unit = push(out, in.grab())
-            })
-            setHandler(out, new OutHandler {
-              def onPull(): Unit = in.pull()
-            })
+            val handler = new InHandler {
+              override def onPush(): Unit = push(out, in.grab())
+
+              override def onUpstreamFinish(): Unit = {
```
jrudolph: Not necessary to cancel the port that was just completed.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/AlpnSwitch.scala
ktoso: hmm, AFAIR this or the cancel below - was checking manually and pinpointed exactly which one. 

I thought I commented on github which one it was (likely have it noted some where if we want to include just that single fix)

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/AlpnSwitch.scala
```diff
@@ -89,28 +100,59 @@ object AlpnSwitch {
...
+              override def onUpstreamFailure(ex: Throwable): Unit = {
+                in.cancel()
+                super.onUpstreamFailure(ex)
+              }
+            }
+
+            val outHandler = new OutHandler {
+              override def onPull(): Unit = in.pull()
+              override def onDownstreamFinish(): Unit = {
+                in.cancel()
```
jrudolph: This would be the other leak suspect.

ktoso: Yeah, it was the complete above or this one, don't remember now - we want both IMO anyway

jrudolph: Yes, we need both, otherwise, we might still keep the `SubSource` / `SubSink` alive depending on how the race between completion / cancellation ends up.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/AlpnSwitch.scala
ktoso: Right, did not notice it's cancelling itself basically

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/AlpnSwitch.scala
```diff
@@ -89,28 +96,50 @@ object AlpnSwitch {
                   propagatePull
                 case Some(ele) ⇒
                   new OutHandler {
-                    def onPull(): Unit = {
+                    override def onPull(): Unit = {
                       out.push(initialElement.get)
                       out.setHandler(propagatePull)
                     }
+
+                    override def onDownstreamFinish(): Unit = {
```
jrudolph: This one isn't necessary as well.

jrudolph:
I think we still need to add tests to prevent leaks?

And can we go back to our former standard of requiring at least one LGTM before merging ;) I don't think we save any time by expediting merges and then having to follow up most PRs.

ktoso:
I kind of assumed those were ready to go since they had so many rounds of review.
Yes, I agree about being explicit with LGTMs, sorry about those.

Issue #1072, opened 2017-04-30
StackOverflowError on invalid media range in accept header.


mrumkovskis:
Route produces java.lang.StackOverflowError (shuts the server down) if http request contains Accept header with invalid media range like: '*/xml'

For example: Accept: */xml

Can be reproduced on almost any running akka-http server using routing DSL including sample one provided in the documentation: http://doc.akka.io/docs/akka-http/current/scala/http/introduction.html#using-akka-http
Was not able to reproduce error using ScalatestRouteTest

This seems due to the error in header parsing i.e.
`akka.http.scaladsl.model.headers.Accept.parseFromValueString("*/xml").toOption.get.acceptsAll == true`

akka-http version: 10.0.5

ktoso:
Hi and thanks for the reproducer!
You're right, it should not fail so not-nicely, I'll have a look into it right away.

jrudolph:
We seem to have multiple problems here leading to the explained outcome:
 * `*/xml` is allowed as a valid media range. It seems to be allowed by [the grammar in RFC 7231](https://tools.ietf.org/html/rfc7231#section-5.3.2) but it isn't clearly defined what the meaning would be. We also didn't decide and didn't foresee that there will be other media ranges with mainType == "*" other than "*/*" and returned `acceptsAll == true`.
 * `RequestContextImpl.complete` allows unbounded recursion through `attemptRecoveryFromUnacceptableResponseContentTypeException` (in this particular case, what happens is that `*/xml` qualifies as `acceptsAll == true` but still doesn't accept anything).

History:

In https://github.com/akka/akka/issues/16494, support for `withAcceptAll` was added but it didn't foresee the `*/xyz` media-range which seems to be valid by grammar (but what should it mean?). In the model it returns `true` for `acceptsAll` but the actual range checking doesn't accept all media types. 

In https://github.com/akka/akka/issues/19397 and https://github.com/akka/akka/issues/19842 (fixed in https://github.com/akka/akka/issues/19849), it was attempted to make non-200 responses ignore the accept header. The solution was to rerun the marshalling with `withAcceptAll` in the case of non-2xx responses. This handling was added in a prominent place in `RequestContextImpl`. This is problematic for various reasons:

 * The fix was added at a code path that basically every request runs through, risking unintended side-effects (like here).
 * Marshallers might be run twice for one `complete` call. That might be unexpected.
[...]

PR #1075, opened 2017-05-02
Fix weird MediaType parsing (Don't panic its only a weird media range)

ktoso:
Resolves in 2 parts https://github.com/akka/akka-http/issues/1072

One is to avoid infinite retrying by refining what a real wildcard actually is.
Two is to ignore weird media types in the form of `*/wat`

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/model/parser/AcceptHeader.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/model/parser/AcceptHeader.scala
@@ -32,8 +32,11 @@ private[parser] trait AcceptHeader { this: Parser with CommonRules with CommonAc
     }
   }
 
+  // this specific ordering PREVENTS that next rule is allowed to parse `*/xyz` as a valid media range
   def `media-range-def` = rule {
-    "*/*" ~ push("*") ~ push("*") | `type` ~ '/' ~ ('*' ~ !tchar ~ push("*") | subtype) | '*' ~ push("*") ~ push("*")
+    "*/*" ~ push("*") ~ push("*") |
+      '*' ~ push("*") ~ push("*") |
+      `type` ~ '/' ~ ('*' ~ !tchar ~ push("*") | subtype)
   }
 }
 
--- a/akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/IllegalHeadersIntegrationSpec.scala
+++ b/akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/IllegalHeadersIntegrationSpec.scala
@@ -1,0 +1,37 @@
+/*
+ * Copyright (C) 2009-2017 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.http.scaladsl.server.directives
+
+import akka.http.scaladsl.model.StatusCodes
+import akka.http.scaladsl.model.headers.{ Accept, RawHeader }
+import akka.http.scaladsl.server.IntegrationRoutingSpec
+
+import scala.concurrent.Await
+import scala.concurrent.duration._
+
+/**
+ * Has to excercise the entire stack, tgus an IntegrationRoutingSpec (not reproducable using just RouteTest).
+ */
+class IllegalHeadersIntegrationSpec extends IntegrationRoutingSpec {
+
+  "Illegal content type in request" should {
+    val route = extractRequest { req ⇒
+      complete(s"Accept:${req.header[Accept]}, byName:${req.headers.find(_.is("accept"))}")
+    }
+
+    // see: https://github.com/akka/akka-http/issues/1072
+    "not StackOverflow but be rejected properly" in {
+      val theIllegalHeader = RawHeader("Accept", "*/xml")
+      Get().addHeader(theIllegalHeader) ~!> route ~!> { response ⇒
+        import response._
+
+        status should ===(StatusCodes.OK)
+        val responseString = Await.result(response.entity.toStrict(1.second), 2.seconds).data.utf8String
+        responseString should ===("Accept:None, byName:Some(accept: */xml)")
+      }
+    }
+  }
+
+}
--- a/akka-http/src/main/scala/akka/http/scaladsl/server/RequestContextImpl.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/server/RequestContextImpl.scala
@@ -122,4 +122,7 @@ private[http] class RequestContextImpl(
     routingSettings:  RoutingSettings          = settings,
     parserSettings:   ParserSettings           = parserSettings) =
     new RequestContextImpl(request, unmatchedPath, executionContext, materializer, log, routingSettings, parserSettings)
+
+  override def toString: String =
+    s"""RequestContext($request, $unmatchedPath, [more settings])"""
 }
--- a/akka-http/src/main/scala/akka/http/scaladsl/server/RequestContext.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/server/RequestContext.scala
@@ -4,19 +4,23 @@
 
 package akka.http.scaladsl.server
 
-import scala.concurrent.{ Future, ExecutionContextExecutor }
+import akka.annotation.DoNotInherit
+
+import scala.concurrent.{ ExecutionContextExecutor, Future }
 import akka.stream.Materializer
 import akka.event.LoggingAdapter
 import akka.http.scaladsl.marshalling.ToResponseMarshallable
-
 import akka.http.scaladsl.model._
 import akka.http.scaladsl.model.StatusCodes._
-import akka.http.scaladsl.settings.{ RoutingSettings, ParserSettings }
+import akka.http.scaladsl.settings.{ ParserSettings, RoutingSettings }
 
 /**
+ * This class is not meant to be extended by user code.
+ *
  * Immutable object encapsulating the context of an [[akka.http.scaladsl.model.HttpRequest]]
  * as it flows through a akka-http Route structure.
  */
+@DoNotInherit
 trait RequestContext {
 
   /** The request this context represents. Modelled as a `val` so as to enable an `import ctx.request._`. */


ktoso:
Formatting check did not like my manually aligned thingy.

akka-http/src/main/scala/akka/http/scaladsl/server/RequestContextImpl.scala
```diff
@@ -109,7 +113,8 @@ private[http] class RequestContextImpl(
   /** Attempts recovering from the special case when non-2xx response is sent, yet content negotiation was unable to find a match. */
   private def attemptRecoveryFromUnacceptableResponseContentTypeException(trm: ToResponseMarshallable, supported: Set[ContentNegotiator.Alternative]): Future[RouteResult] =
     trm.value match {
-      case (status: StatusCode, value) if !status.isSuccess ⇒ this.withAcceptAll.complete(trm) // retry giving up content negotiation
```
jrudolph: I think that method doesn't belong into `RequestContextImpl` at all. Let's not touch it here and I'll follow up with a better fix for https://github.com/akka/akka/issues/19397. I think we need to be really careful not to introduce any custom logic like this into central pieces of akka-http because - like here - it can be hard to reason about all the code paths flowing through this code and if it has a bug it will always affect all akka-http applications.

akka-http/src/main/scala/akka/http/scaladsl/server/RequestContextImpl.scala
```diff
@@ -109,7 +113,8 @@ private[http] class RequestContextImpl(
...
+      case (status: StatusCode, value) if !status.isSuccess && !this.isAcceptAll /** don't retry once we have tried with accepting all already */ ⇒
```
jrudolph: Have you checked if that would have prevented the bug? I don't think so.

akka-http/src/main/scala/akka/http/scaladsl/server/RequestContextImpl.scala
ktoso: Yeah, the retry with accept all is pretty "tagged on" here, looking forward to your cleanup of that then - thanks in advance 👍

akka-http/src/main/scala/akka/http/scaladsl/server/RequestContextImpl.scala
```diff
@@ -122,4 +127,7 @@ private[http] class RequestContextImpl(
     routingSettings:  RoutingSettings          = settings,
     parserSettings:   ParserSettings           = parserSettings) =
     new RequestContextImpl(request, unmatchedPath, executionContext, materializer, log, routingSettings, parserSettings)
+
+  override def toString: String =
+    s"""RequestContext($request, $unmatchedPath, [more settings])"""
```
ktoso: please sanity check, should be fine right?

akka-http/src/main/scala/akka/http/scaladsl/server/RequestContextImpl.scala
ktoso: Yes I have checked, before implementing the throwing in the media types

akka-http-core/src/main/scala/akka/http/scaladsl/model/MediaRange.scala
```diff
@@ -92,6 +96,7 @@ object MediaRange {
     override def isMultipart = mediaType.isMultipart
     override def isText = mediaType.isText
     override def isVideo = mediaType.isVideo
+    override def isWildcard = mediaType.mainType == "*" && mediaType.subType == "*"
```
jrudolph: I don't think that makes sense. A `MediaType` should never have "*" somewhere, so this can be always `false`.

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/IllegalHeadersIntegrationSpec.scala
```diff
@@ -0,0 +1,37 @@
...
+class IllegalHeadersIntegrationSpec extends IntegrationRoutingSpec {
+
+  "Illegal content type in request" should {
+    val route = extractRequest { req ⇒
+      complete(s"Accept:${req.header[Accept]}, byName:${req.headers.find(_.is("accept"))}")
+    }
+
+    // see: https://github.com/akka/akka-http/issues/1072
+    "not StackOverflow but be rejected properly" in {
+      val theIllegalHeader = RawHeader("Accept", "*/xml")
```
ktoso: Yes it did

ktoso: killing the entire jvm with it

akka-http-core/src/main/scala/akka/http/scaladsl/model/MediaRange.scala
```diff
@@ -57,12 +57,15 @@ object MediaRange {
   private final case class Custom(mainType: String, params: Map[String, String], qValue: Float)
```
jrudolph: Let's not allow a custom type `*/*` in the first place.

akka-http-core/src/main/scala/akka/http/scaladsl/model/MediaRange.scala
```diff
@@ -57,12 +57,15 @@ object MediaRange {
...
+
+    def subType = "*"
+    private def subTypeChar = '*'
     def matches(mediaType: MediaType) = (mainType == "*" || mediaType.mainType == mainType) &&
       this.params.forall { case (key, value) ⇒ mediaType.params.get(key).contains(value) }
     def withParams(params: Map[String, String]) = custom(mainType, params, qValue)
     def withQValue(qValue: Float) = if (qValue != this.qValue) custom(mainType, params, qValue) else this
     def render[R <: Rendering](r: R): r.type = {
-      r ~~ mainType ~~ '/' ~~ '*'
+      r ~~ mainType ~~ '/' ~~ subTypeChar
```
jrudolph: How about 

```scala
r ~~ mainType ~~ "/*"
```

akka-http-core/src/main/scala/akka/http/scaladsl/model/MediaRange.scala
ktoso: gladly :) Since it's a predefined one.

akka-http-core/src/main/scala/akka/http/scaladsl/model/MediaRange.scala
```diff
@@ -15,14 +15,14 @@ sealed abstract class MediaRange extends jm.MediaRange with Renderable with With
   def params: Map[String, String]
   def qValue: Float
   def matches(mediaType: MediaType): Boolean
+  def isWildcard: Boolean
```
jrudolph: Let's just define it as 

```scala
this == `*/*`
```

ktoso: good one  👍

akka-http/src/main/scala/akka/http/scaladsl/server/RequestContextImpl.scala
jrudolph: Still fails for me if I just take the change in this file and your test case.

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/IllegalHeadersIntegrationSpec.scala
jrudolph: Ah, didn't see that this was an IntegrationRoutingSpec (or that we actually had something like that), so that the headers are run through the whole infrastructure.

akka-http-core/src/main/scala/akka/http/impl/model/parser/CommonActions.scala
```diff
@@ -32,6 +32,9 @@ private[parser] trait CommonActions {
         case "encrypted"   ⇒ multipart.encrypted(params)
         case custom        ⇒ MediaType.customMultipart(custom, params)
       }
+      case "*" if subType != "*" ⇒
+        throw new IllegalArgumentException(s"Unsupported */$subType MediaType detected, " +
```
jrudolph: I like the error message but we should fix it directly in the model, otherwise we might have missed other problematic code paths.

jrudolph:
Let's keep this patch to a minimum:

 * test looks fine
 * move the exception to the model to prevent that a media range `*/xyz` can be created

mrumkovskis:
May be Accept header parser should be fixed (end with parsing failure) instead:
https://github.com/akka/akka-http/blob/3c966af2255b58b02fbdacd6abec89c47f78ce48/akka-http-core/src/main/scala/akka/http/impl/model/parser/AcceptHeader.scala

since media type (*/<mime-subtype>) seems to be invalid:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept

jrudolph:
@mrumkovskis that's something I also thought, we might want to fix parsing as well as the model, so the invalid value cannot be produced both from code or incoming bytes.

jrudolph:
Thanks for that Mozilla docs. I don't think it is clearer about what to allow for `<MIME_type>` AFAICS. The most technical specification is the one in [RFC 7231](https://tools.ietf.org/html/rfc7231#section-3.1.1.1) (which we implemented but which probably isn't strict enough):

```
     media-type = type "/" subtype *( OWS ";" OWS parameter )
     type       = token
     subtype    = token
```

raboof:
[The section on the `Accept` header](https://tools.ietf.org/html/rfc7231#section-5.3.2) (which describes media ranges rather than media types) is a bit more explicit:

```
Accept = #( media-range [ accept-params ] )

     media-range    = ( "*/*"
                      / ( type "/" "*" )
                      / ( type "/" subtype )
                      ) *( OWS ";" OWS parameter )
     accept-params  = weight *( accept-ext )
     accept-ext = OWS ";" OWS token [ "=" ( token / quoted-string ) ]
```

.. which *suggests* `*` would not be valid for the first part, but strictly speaking indeed `type` seems to be just a `token` and [`token` allows `*`](https://tools.ietf.org/html/rfc7230#section-3.2.6)...

jrudolph:
@ktoso can you also change the definition of `type` in the parser to 

```
def `type`: Rule1[String] = rule {
    '*' ~ failX[HNil, String :: HNil]("'*' is not allowed as the main type of a media type or a media range (diverging from the grammar specified in RFC 7231 section-3.1.1.1) to prevent ambiguities with media ranges like '*/*'") |
      token
  }
```

ktoso:
Thanks for the parser snippet, fixing that, as well as adding the check to the model itself then.

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/IllegalHeadersIntegrationSpec.scala
ktoso: Hehe yeah, I implemented that some time ago since we needed to write some more "integration" tests :) Easy to miss if you have not had the need for it previously

akka-http-core/src/main/scala/akka/http/scaladsl/model/MediaRange.scala
ktoso: Hah, sadly it's too strict since we need the (old code, works):

```scala
  val `*/*` = new PredefinedMediaRange("*/*") {
    def matches(mediaType: MediaType) = true
    override def isWildcard: Boolean = true
  }
  val `*/*;q=MIN` = `*/*`.withQValue(Float.MinPositiveValue)
```

to work like that in order to have the proper ordering when picking qvalues it seems...
We can live without the fancy `this ==` impl IMHO

ktoso:
Uff, tricky stuff here - not all of the review comments were applied, some would lead to incorrect behaviour in qvalue picking etc. Parser impl changed and is correct now (a bit more than above comment hinted, but we figured it out with @jrudolph). I have removed the naive dont-go-crazy-recursive check as it did not help in practice, and we'll find a better place for it in any case.

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/IllegalHeadersIntegrationSpec.scala
```diff
@@ -0,0 +1,37 @@
...
+      Get().addHeader(theIllegalHeader) ~!> route ~!> { response ⇒
+        import response._
+
+        status should ===(StatusCodes.OK)
+        val responseString = Await.result(response.entity.toStrict(1.second), 2.seconds).data.utf8String
+        responseString should ===("Accept:None, byName:Some(accept: */xml)")
```
ktoso: After @jrudolph's fix this will fail more aggresively with a 406 which is good.

jrudolph:
I see that happens now because of a pecularity of PEG grammars that in alternatives (chains of `|`), left alternatives must not be prefixes of later alternatives. So, putting `*` first will prevent that the last clause may parse something that starts with `*`.

ktoso:
Correct, it's the parser rule itself that now protects from the bad input.

akka-http-core/src/main/scala/akka/http/impl/model/parser/AcceptHeader.scala
```diff
@@ -33,7 +33,9 @@ private[parser] trait AcceptHeader { this: Parser with CommonRules with CommonAc
   }
 
   def `media-range-def` = rule {
-    "*/*" ~ push("*") ~ push("*") | `type` ~ '/' ~ ('*' ~ !tchar ~ push("*") | subtype) | '*' ~ push("*") ~ push("*")
+    "*/*" ~ push("*") ~ push("*") |
+      '*' ~ push("*") ~ push("*") |
```
jrudolph: Can we add a comment here:

```
// ordering prevents that next rule is allowed to parse `*/xyz` as a valid media range
```

ktoso: yes wil do

jrudolph:
It's good that this is now the minimal change needed to fix the issue at hand. We should still strive to harden the other places (`type` parser and model classes). I created https://github.com/akka/akka-http/issues/1082 to track this as a follow-up.

ktoso:
Yes, there's more to be done around these. Merging this one to resolve the issue at hand, and we'll follow up on it

PR #1076, opened 2017-05-02
=htp #1072 move special non-2xx handling from RequestContextImpl to fromStatusCodeAndHeadersAndValue marshaller

### LINKED ISSUE #1072: StackOverflowError on invalid media range in accept header.

jrudolph:
This is an alternative solution for akka/akka#19397 that replaces the one
introduced in akka/akka#19849.

The former solution placed the special handling inside the main processing
path of requests in Akka HTTP. This solution is simpler and fixes the issue
in a narrower way.

The code comment explains the solution:

For non-2xx status, add an opaque fallback marshalling
using the first returned marshalling that will be used if the
result wouldn't otherwise be accepted. The reasoning is that users
often use routes like  `complete(400 -> "Illegal request in this context")`
to fail a route (instead of using rejection handling). If the client uses
narrow Accept headers like `Accept: application/json` the user
supplied error message will not be rendered because it will only
accept `text/plain` but not `application/json` responses and fail
the whole request with a "406 Not Acceptable" response.

Adding the opaque fallback rendering will give an escape hatch for
those situations.

(This would also fix #1072 but would make the test case fail with a 406 instead of ignoring the
strange Accept header. Being more strict there (as proposed in #1075 definitely makes sense as well.))


ktoso:
> (This would also fix #1072 but would make the test case fail with a 406 instead of ignoring the
strange Accept header. Being more strict there (as proposed in #1075 definitely makes sense as well.))

I'm in favour of failing with a 406 in that case, so that's all good.

akka-http/src/main/scala/akka/http/scaladsl/marshalling/Marshaller.scala
```diff
@@ -176,6 +176,13 @@ object Marshaller
  */
 sealed trait Marshalling[+A] {
```
ktoso: mark it as not meant for extension while here?

akka-http/src/main/scala/akka/http/scaladsl/marshalling/PredefinedToResponseMarshallers.scala
```diff
@@ -53,14 +53,34 @@ trait PredefinedToResponseMarshallers extends LowPriorityToResponseMarshallerImp
...
+          val mappedMarshallings = marshallings map (_ map (statusCodeAndEntityResponse(status, headers, _)))
+          if (status.isSuccess)
+            // for 2xx status codes delegate content-type negotiation to the value marshaller
+            mappedMarshallings
+          else
+            // For non-2xx status, add an opaque fallback marshalling using the first returned marshalling
+            // that will be used if the result wouldn't otherwise be accepted.
+            //
+            // The reasoning is that users often use routes like  `complete(400 -> "Illegal request in this context")`
+            // to fail a route (instead of using rejection handling). If the client uses narrow Accept headers like
```
ktoso: "fail" a route

I'm nitpicking, sorry :)

akka-http/src/main/scala/akka/http/scaladsl/marshalling/PredefinedToResponseMarshallers.scala
```diff
@@ -53,14 +53,34 @@ trait PredefinedToResponseMarshallers extends LowPriorityToResponseMarshallerImp
...
+            // `Accept: application/json` the user supplied error message will not be rendered because it will only
+            // accept `text/plain` but not `application/json` responses and fail the whole request with a "406 Not Acceptable"
+            // response.
+            //
+            // Adding the opaque fallback rendering will give an escape hatch for those situations.
+            // See akka/akka#19397, akka/akka#19842, and #1072.
+            mappedMarshallings match {
+              case Nil                   ⇒ Nil
+              case firstMarshalling :: _ ⇒ mappedMarshallings :+ firstMarshalling.toOpaque(HttpCharsets.`UTF-8`)
```
ktoso: good fix, much cleaner than the retry 👍

akka-http/src/main/scala/akka/http/scaladsl/server/RequestContextImpl.scala
```diff
@@ -41,11 +41,11 @@ private[http] class RequestContextImpl(
   override def complete(trm: ToResponseMarshallable): Future[RouteResult] =
     trm(request)(executionContext)
       .fast.map(res ⇒ RouteResult.Complete(res))(executionContext)
-      .fast.recoverWith {
+      .fast.recover {
         case Marshal.UnacceptableResponseContentTypeException(supported) ⇒
-          attemptRecoveryFromUnacceptableResponseContentTypeException(trm, supported)
+          RouteResult.Rejected(UnacceptedResponseContentTypeRejection(supported) :: Nil)
```
ktoso: 👍

akka-http/src/main/scala/akka/http/scaladsl/marshalling/Marshaller.scala
jrudolph: It's a `sealed trait` so it cannot be extended.

akka-http/src/main/scala/akka/http/scaladsl/marshalling/PredefinedToResponseMarshallers.scala
jrudolph: Good nitpick :) Improved the wording.

PR #1081, opened 2017-05-03
Added an example of route.concat usage, fixes #1020

### LINKED ISSUE #1020: Add examples of `Route.concat`
**EDIT**
As discussed below, the original issue as it was is not needed, but we could still add at least 1 example using `concat` under the [Directives page](http://doc.akka.io/docs/akka-http/current/scala/http/routing-dsl/directives/index.html), so people can see the alternative.

---------------------------------

**Original Issue**
Title: Promote `Route.concat` instead of `~`
All examples for the Scala DSL about Directive concatenation are based on the `~` operator. Even though using `~` comes very handy because of it's length and the fact that can be used as an operator, it's really error prone as it's very easy to forget one `~`. We already have a note on the [Directives page](http://doc.akka.io/docs/akka-http/current/scala/http/routing-dsl/directives/index.html#composing-directives) warning users about the risk of using `~` and forgetting one.
What do you think about rewriting some of the existing examples under the [Directives page](http://doc.akka.io/docs/akka-http/current/scala/http/routing-dsl/directives/index.html) so they use `concat` instead?

Some other comments about `~`:
- It's weirdly aligned when using formatting tools
- It's easy to forget one, and most of the times there is no compiler error to help
- When reading some existing code, some newcomers to Akka HTTP had a hard time to see that there was a `~` at the end of the line.

domitian:
Added one example of `concat` combinator usage.

### DIFF
--- a/docs/src/main/paradox/scala/http/routing-dsl/directives/index.md
+++ b/docs/src/main/paradox/scala/http/routing-dsl/directives/index.md
@@ -202,6 +202,12 @@ is in fact the most readable one.
 Still, the purpose of the exercise presented here is to show you how flexible directives can be and how you can
 use their power to define your web service behavior at the level of abstraction that is right for **your** application.
 
+### Composing Directives with `concat` Combinator
+
+Alternatively we can combine directives using `concat` combinator where we pass each directive as an argument to the combinator function instead of chaining them with `~` . Let's take a look at the usage of this combinator:
+
+@@snip [DirectiveExamplesSpec.scala](../../../../../../test/scala/docs/http/scaladsl/server/DirectiveExamplesSpec.scala) { #example-8 }
+
 ## Type Safety of Directives
 
 When you combine directives with the `|` and `&` operators the routing DSL makes sure that all extractions work as
--- a/docs/src/test/scala/docs/http/scaladsl/server/DirectiveExamplesSpec.scala
+++ b/docs/src/test/scala/docs/http/scaladsl/server/DirectiveExamplesSpec.scala
@@ -112,6 +112,25 @@ class DirectiveExamplesSpec extends RoutingSpec {
     //#example-7
   }
 
+  "example-8" in {
+    //#example-8
+    def innerRoute(id: Int): Route =
+      concat(get {
+        complete {
+          "Received GET request for order " + id
+        }
+      },
+      put {
+        complete {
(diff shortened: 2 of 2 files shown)


jlprat:
OK TO TEST

docs/src/main/paradox/scala/http/routing-dsl/directives/index.md
```diff
@@ -163,6 +163,10 @@ thing like this:
 
 @@snip [DirectiveExamplesSpec.scala](../../../../../../test/scala/docs/http/scaladsl/server/DirectiveExamplesSpec.scala) { #example-2 }
 
+Or better yet using `concat` combinator where we pass each directive as a parameter to the combinator function instead of chaining them with `~` . Let's take a look at the usage of this combinator:
```
ktoso: let's not judge 😉  "alternatively" instead of "better yet"

ktoso: It can get it's own sub section so that it's easy to link to though - please add a subsection for it :)

ktoso:
Refs #1020

docs/src/main/paradox/scala/http/routing-dsl/directives/index.md
domitian: you mean as a sub-section to 'composing directives' right?

jlprat: :+1:  for the "alternatively"

docs/src/main/paradox/scala/http/routing-dsl/directives/index.md
```diff
@@ -163,6 +163,12 @@ thing like this:
 
 @@snip [DirectiveExamplesSpec.scala](../../../../../../test/scala/docs/http/scaladsl/server/DirectiveExamplesSpec.scala) { #example-2 }
 
+### Composing Directives with `concat` Combinator
```
jlprat: Move it further down maybe? Now it looks the other combinators like `|` or `&` are under this new section

domitian: done

domitian:
Added a sub-section at the end of 'Composing Directives' section. And squashed the commits.

PR #1087, opened 2017-05-04
=htc move Http2Shadow to impl package

ktoso:
Review follow up from https://github.com/akka/akka-http/pull/1062


jrudolph:
Let's put it under `akka.http.impl.engine` instead. Otherwise, it's so alone in its package.

ktoso:
Ok, we don't want lonely classes :)

ktoso:
Moved it, will merge after validation, kk?

Issue #1088, opened 2017-05-04
Make sure remote-address-header flag is honored for Http/2


jrudolph:
It needs to work both for 

 * Http/2
 * Http/1 when Http/2 is enabled (it that case the attribute needs to be passed in to the graph correctly in `Http2.bindAndHandleAsync`)

ktoso:
I'll tackle this right away, since we expect to block Play from using h2

gmethvin:
I was just setting up http2 on play and got this error:

```
[ERROR] [05/10/2017 20:35:16.567] [play-dev-mode-akka.actor.default-dispatcher-6] [akka.actor.ActorSystemImpl(play-dev-mode)] Internal server error, sending 500 response
java.lang.IllegalStateException: `Remote-Address` header was missing
	at play.core.server.AkkaHttpServer.remoteAddressOfRequest(AkkaHttpServer.scala:176)
	at play.core.server.AkkaHttpServer.handleRequest(AkkaHttpServer.scala:153)
	at play.core.server.AkkaHttpServer.$anonfun$createServerBinding$4(AkkaHttpServer.scala:89)
	at akka.http.impl.engine.http2.Http2Blueprint$.$anonfun$handleWithStreamIdHeader$1(Http2Blueprint.scala:98)
	at akka.stream.impl.fusing.MapAsyncUnordered$$anon$24.onPush(Ops.scala:1250)
	at akka.stream.impl.fusing.GraphInterpreter.processPush(GraphInterpreter.scala:499)
	at akka.stream.impl.fusing.GraphInterpreter.execute(GraphInterpreter.scala:401)
	at akka.stream.impl.fusing.GraphInterpreterShell.runBatch(ActorGraphInterpreter.scala:571)
	at akka.stream.impl.fusing.GraphInterpreterShell$AsyncInput.execute(ActorGraphInterpreter.scala:457)
	at akka.stream.impl.fusing.GraphInterpreterShell.processEvent(ActorGraphInterpreter.scala:546)
	at akka.stream.impl.fusing.ActorGraphInterpreter.akka$stream$impl$fusing$ActorGraphInterpreter$$processEvent(ActorGraphInterpreter.scala:725)
	at akka.stream.impl.fusing.ActorGraphInterpreter.akka$stream$impl$fusing$ActorGraphInterpreter$$shortCircuitBatch(ActorGraphInterpreter.scala:715)
[...]

PR #1090, opened 2017-05-04
+h2 #1088 include RemoteAddress in ALPN switched to server

ktoso:
Quick impl of https://github.com/akka/akka-http/issues/1088

We forgot to propagate headers to the ALPN "switched to" server graph.
This is not cleaned up, impled quickly before leaving home today, if you have suggestions where to "pass around" the headers let me know, I marked some TODOs. The Parsing is a method now, now a GraphStage so it has to get it passed in (it can't just reach for the attributes), I don't mind this style I guess, but wonder if there would be a cleaner way

Resolves https://github.com/akka/akka-http/issues/1088

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpAttributes.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpAttributes.scala
@@ -18,9 +18,9 @@ import akka.stream.Attributes
 private[akka] object HttpAttributes {
   import Attributes._
 
-  private[akka] final case class RemoteAddress(address: Option[InetSocketAddress]) extends Attribute
+  private[akka] final case class RemoteAddress(address: InetSocketAddress) extends Attribute
 
-  private[akka] def remoteAddress(address: Option[InetSocketAddress]) =
+  private[akka] def remoteAddress(address: InetSocketAddress) =
     Attributes(RemoteAddress(address))
 
   private[akka] val empty =
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
@@ -97,7 +97,8 @@ private[http] object HttpServerBluePrint {
     override val shape: FlowShape[RequestOutput, HttpRequest] = FlowShape.of(in, out)
 
     override def createLogic(inheritedAttributes: Attributes) = new GraphStageLogic(shape) with InHandler with OutHandler {
-      val remoteAddress = inheritedAttributes.get[HttpAttributes.RemoteAddress].flatMap(_.address)
+      val remoteAddress = inheritedAttributes.get[HttpAttributes.RemoteAddress].map(_.address)
+
       var downstreamPullWaiting = false
       var completionDeferred = false
       var entitySource: SubSourceOutlet[RequestOutput] = _
@@ -120,6 +121,7 @@ private[http] object HttpServerBluePrint {
       override def onPush(): Unit = grab(in) match {
         case RequestStart(method, uri, protocol, hdrs, entityCreator, _, _) ⇒
           val effectiveMethod = if (method == HttpMethods.HEAD && settings.transparentHeadRequests) HttpMethods.GET else method
+
           val effectiveHeaders =
             if (settings.remoteAddressHeader && remoteAddress.isDefined)
               headers.`Remote-Address`(RemoteAddress(remoteAddress.get)) +: hdrs
--- a/akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
@@ -6,6 +6,8 @@ package akka.http.impl.util
 
 import akka.NotUsed
 import akka.actor.Cancellable
+import akka.annotation.InternalApi
+import akka.event.Logging
 import akka.stream._
 import akka.stream.impl.fusing.GraphInterpreter
 import akka.stream.impl.fusing.GraphStages.SimpleLinearGraphStage
@@ -214,6 +216,15 @@ private[http] object StreamUtils {
       i ⇒ f(i) :: Nil
     }
 
+  /**
+   * Lifts the streams attributes into an element and passes them to the function for each passed through element.
+   * Similar idea than [[FlowOps.statefulMapConcat]] but for a simple map.
+   *
+   * The result of `Attributes => (T => U)` is cached, and only the `T => U` function will be invoked afterwards for each element.
+   */
+  def statefulAttrsMap[T, U](functionConstructor: Attributes ⇒ T ⇒ U): Flow[T, U, NotUsed] =
+    Flow[T].via(ExposeAttributes[T, U](functionConstructor))
+
   trait ScheduleSupport { self: GraphStageLogic ⇒
     /**
      * Schedule a block to be run once after the given duration in the context of this graph stage.
@@ -279,4 +290,21 @@ private[http] class EnhancedByteStringSource[Mat](val byteStringStream: Source[B
     byteStringStream.runFold(ByteString.empty)(_ ++ _)
   def utf8String(implicit materializer: Materializer, ec: ExecutionContext): Future[String] =
     join.map(_.utf8String)
+}
+
+/** INTERNAL API */
+@InternalApi private[http] case class ExposeAttributes[T, U](functionConstructor: Attributes ⇒ T ⇒ U)
+  extends GraphStage[FlowShape[T, U]] {
+
+  val in = Inlet[T]("ExposeAttributes.in")
+  val out = Outlet[U]("ExposeAttributes.out")
+  override val shape = FlowShape(in, out)
+
+  override def createLogic(inheritedAttributes: Attributes) = new GraphStageLogic(shape) with InHandler with OutHandler {
+    val f = functionConstructor(inheritedAttributes)
+    override def onPush(): Unit = push(out, f(grab(in)))
+    override def onPull(): Unit = pull(in)
+
+    setHandlers(in, out, this)
+  }
 }
--- a/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
+++ b/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
@@ -9,8 +9,9 @@ import akka.event.LoggingAdapter
 import akka.http.impl.engine.http2.framing.{ Http2FrameParsing, Http2FrameRendering }
 import akka.http.impl.engine.http2.hpack.{ HeaderCompression, HeaderDecompression }
 import akka.http.impl.engine.parsing.HttpHeaderParser
+import akka.http.impl.engine.server.HttpAttributes
 import akka.http.impl.util.StreamUtils
-import akka.http.scaladsl.model.{ ErrorInfo, HttpRequest, HttpResponse }
+import akka.http.scaladsl.model._
 import akka.http.scaladsl.model.http2.Http2StreamIdHeader
 import akka.http.scaladsl.settings.{ ParserSettings, ServerSettings }
 import akka.stream.scaladsl.{ BidiFlow, Flow, Source }
@@ -21,11 +22,14 @@ import scala.concurrent.{ ExecutionContext, Future }
 /**
  * Represents one direction of an Http2 substream.
  */
-private[http2] final case class Http2SubStream(initialHeaders: ParsedHeadersFrame, data: Source[ByteString, Any]) {
+private[http2] final case class Http2SubStream(
+  initialHeaders: ParsedHeadersFrame,
+  data:           Source[ByteString, Any]) {
   def streamId: Int = initialHeaders.streamId
 }
 
 object Http2Blueprint {
+  
   // format: OFF
   def serverStack(settings: ServerSettings, log: LoggingAdapter): BidiFlow[HttpResponse, ByteString, ByteString, HttpRequest, NotUsed] =
     httpLayer(settings, log) atop
@@ -34,7 +38,6 @@ object Http2Blueprint {
       hpackCoding() atop
       // LogByteStringTools.logToStringBidi("framing") atop // enable for debugging
       framing()
-
   // format: ON
 
   def framing(): BidiFlow[FrameEvent, ByteString, ByteString, FrameEvent, NotUsed] =
@@ -82,9 +85,9 @@ object Http2Blueprint {
     }
     BidiFlow.fromFlows(
       Flow[HttpResponse].map(ResponseRendering.renderResponse(settings, log)),
-      Flow[Http2SubStream].via(StreamUtils.statefulMap { () ⇒
+      Flow[Http2SubStream].via(StreamUtils.statefulAttrsMap { attrs ⇒
         val headerParser = masterHttpHeaderParser.createShallowCopy()
-        RequestParsing.parseRequest(headerParser)
+        RequestParsing.parseRequest(headerParser, settings, attrs)
       }))
   }
 
--- a/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/RequestParsing.scala
+++ b/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/RequestParsing.scala
@@ -6,10 +6,15 @@ package akka.http.impl.engine.http2
 
 import akka.annotation.InternalApi
 import akka.http.impl.engine.parsing.HttpHeaderParser
+import akka.http.impl.engine.server.HttpAttributes
+import akka.http.scaladsl.model
 import akka.http.scaladsl.model._
+import akka.http.scaladsl.model.headers.`Remote-Address`
 import akka.http.scaladsl.model.http2.Http2StreamIdHeader
+import akka.http.scaladsl.settings.ServerSettings
+import akka.stream.Attributes
 import akka.stream.scaladsl.Source
-import akka.util.ByteString
+import akka.util.{ ByteString, OptionVal }
 
 import scala.annotation.tailrec
 import scala.collection.immutable.VectorBuilder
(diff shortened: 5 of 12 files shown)


akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -81,9 +86,12 @@ class Http2ServerDemux extends GraphStage[BidiShape[Http2SubStream, FrameEvent,
     new GraphStageLogic(shape) with Http2MultiplexerSupport with Http2StreamHandling with GenericOutletSupport with StageLogging {
       logic ⇒
 
+      private val remoteAddressHeader = inheritedAttributes.get[HttpAttributes.RemoteAddress].flatMap(_.address)
```
jrudolph: I agree that's one way to do it. It would be nicer if we could create the header in `RequestParsing` directly. Maybe we could add a variant of `def mapStateful(f: Attributes => T => U): Flow[T, U]` that would allow us to extract the attributes directly in `RequestParsing`?

very-big-data
```diff
@@ -0,0 +1,1756 @@
+this data is so big, wow
```
jrudolph: Wow, I agree it is big, indeed.

Does it need to be checked in, though? ;)

ktoso: lol, need to remove that :)

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
ktoso: Yeah, sounds ok - will try that soon

ktoso:
Reimplemented to pass the attrs into the parser as well as added a test that checks the header is present and valid.

Please review @jrudolph 

// FYI we'll get you this as soon as possible so you're unblocked @gmethvin

ktoso:
Please review today @jrudolph

akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
```diff
@@ -215,6 +217,13 @@ private[http] object StreamUtils {
       i ⇒ f(i) :: Nil
     }
 
+  /**
+   * Lifts the streams attributes into an element and passes them to the function for each passed through element.
+   * Similar idea than [[FlowOps.statefulMapConcat]] but for a simple map.
+   */
+  def statefulAttrsMap[T, U](functionConstructor: Attributes ⇒ T ⇒ U): Flow[T, U, NotUsed] =
```
jrudolph: Actually, the idea was to pass the attributes only once as they don't change while the stream is running.

akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
```diff
@@ -275,9 +284,25 @@ private[http] object StreamUtils {
...
+/** INTERNAL API */
+@InternalApi private[http] case class ExposeAttributes[T, U](functionConstructor: Attributes ⇒ T ⇒ U)
+  extends GraphStage[FlowShape[T, U]] {
+
+  val in = Inlet[T]("ExposeAttributes.in")
+  val out = Outlet[U]("ExposeAttributes.out")
+  override val shape = FlowShape(in, out)
+
+  override def createLogic(inheritedAttributes: Attributes) = new GraphStageLogic(shape) with InHandler with OutHandler {
+    override def onPush(): Unit = push(out, functionConstructor(inheritedAttributes)(grab(in)))
```
jrudolph: I'd suggest to run the `functionConstructor` only once. I.e.

```scala
val f  = functionConstructor(inheritedAttributes)
override def onPush(): Unit = push(out, f(grab(in)))
```

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
```diff
@@ -82,9 +85,9 @@ object Http2Blueprint {
     }
     BidiFlow.fromFlows(
       Flow[HttpResponse].map(ResponseRendering.renderResponse(settings, log)),
-      Flow[Http2SubStream].via(StreamUtils.statefulMap { () ⇒
+      Flow[Http2SubStream].via(StreamUtils.statefulAttrsMap { attrs ⇒
```
jrudolph: It's even needed to run this function only once, as explained above, otherwise, the parser will be recreated every time and we will use the internal caching of the parser.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/RequestParsing.scala
```diff
@@ -39,6 +46,15 @@ private[http2] object RequestParsing {
 
         headers += Http2StreamIdHeader(subStream.streamId)
 
+        if (serverSettings.remoteAddressHeader) {
```
jrudolph: If you change the signature of the method to

```scala
def parseRequest(httpHeaderParser: HttpHeaderParser): Http2SubStream => HttpRequest
```

this calculation can be done only once per connection and then be reused for every request afterwards like this:

```scala
val remoteAddressHeader = 
   if (serverSettings.remoteAddressHeader) 
     OptionVal.Some { ... }
   else OptionVal.None

subStream => 

if (remoteAddressHeader.isDefined) headers += remoteAddressHeader.get
```

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -678,6 +685,46 @@ class Http2ServerSpec extends AkkaSpec("" + "akka.loglevel = debug")
...
+
+        lazy val theAddress = "127.0.0.1"
+        lazy val thePort = 1337
+        override def modifyServer(server: BidiFlow[HttpResponse, ByteString, ByteString, HttpRequest, NotUsed]) =
+          BidiFlow.fromGraph(StreamUtils.fuseAggressive(server).withAttributes(
+            HttpAttributes.remoteAddress(Some(new InetSocketAddress(theAddress, thePort)))
+          ))
+
+        val probe = TestProbe()
+        override def handlerFlow: Flow[HttpRequest, HttpResponse, NotUsed] =
```
jrudolph: If you use `RequestResponseProbes` you already have probes to use. Or is this just dead code?

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -678,6 +685,46 @@ class Http2ServerSpec extends AkkaSpec("" + "akka.loglevel = debug")
...
+          Flow.fromSinkAndSource(
+            Flow[HttpRequest].map({ request ⇒
+              probe.ref ! request.header[headers.`Remote-Address`]
+              request
+            }).to(Sink.fromSubscriber(requestIn)),
+            Source.fromPublisher(responseOut)
+          )
+
+        val headerBlock = HPackSpecExamples.C41FirstRequestWithHuffman
```
jrudolph: Could make sense to use `AutomaticHpackWireSupport` to send a request directly.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
ktoso: Good catch, thanks - rewriting the functions

akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
ktoso: Right, good catch

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/RequestParsing.scala
ktoso: Ah much better, less looking and constructing

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
ktoso: Oh, indeed just dead code - the probe is not used - was from a previous version of the test I forgot to remove it seems; srry

akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
```diff
@@ -275,9 +286,26 @@ private[http] object StreamUtils {
 /**
  * INTERNAL API
  */
-private[http] class EnhancedByteStringSource[Mat](val byteStringStream: Source[ByteString, Mat]) extends AnyVal {
+@InternalApi private[http] class EnhancedByteStringSource[Mat](val byteStringStream: Source[ByteString, Mat]) extends AnyVal {
```
jrudolph: Let's just make these `private`. No need to add those annotations everywhere if the owning object has it already.

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
```diff
@@ -371,6 +371,7 @@ object HttpEntity {
     contentLength: Long,
     data:          Source[ByteString, Any])
     extends jm.HttpEntity.Default with UniversalEntity {
+    println(s"contentLength = ${contentLength}")
```
jrudolph: Leftover.

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -120,10 +119,13 @@ private[http] object HttpServerBluePrint {
       override def onPush(): Unit = grab(in) match {
         case RequestStart(method, uri, protocol, hdrs, entityCreator, _, _) ⇒
           val effectiveMethod = if (method == HttpMethods.HEAD && settings.transparentHeadRequests) HttpMethods.GET else method
-          val effectiveHeaders =
-            if (settings.remoteAddressHeader && remoteAddress.isDefined)
-              headers.`Remote-Address`(RemoteAddress(remoteAddress.get)) +: hdrs
-            else hdrs
```
jrudolph: Let's keep it as it was before, it was simpler to read and faster for the case where the header was actually used. The extra once-per-connection slowdown of looking up a single attribute should be negligible.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -83,7 +88,7 @@ class Http2ServerDemux extends GraphStage[BidiShape[Http2SubStream, FrameEvent,
 
       override protected def logSource: Class[_] = classOf[Http2ServerDemux]
 
-      val multiplexer = createMultiplexer(frameOut, StreamPrioritizer.first())
+      override val multiplexer = createMultiplexer(frameOut, StreamPrioritizer.first())
```
jrudolph: Let's not add `override`s in arbitrary places. `override` was supposed to be used as a clue to the compiler and the reader's eye to show that a method is actually overridden (in contrast to just being implemented). Putting it in arbitrary places means that we actually lose a bit of safety where an implementation is added to the super-class later on.

I know the akka codebase had the rule somewhere that to add `override` everywhere possible. IMO it doesn't help and was never enforced strictly so it is not really useful. If in doubt, let's go with the more concise version.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -186,15 +191,17 @@ class Http2ServerDemux extends GraphStage[BidiShape[Http2SubStream, FrameEvent,
       })
 
       val bufferedSubStreamOutput = new BufferedOutlet[Http2SubStream](substreamOut)
-      def dispatchSubstream(sub: Http2SubStream): Unit = bufferedSubStreamOutput.push(sub)
```
jrudolph: No change needed here.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/RequestParsing.scala
```diff
@@ -19,69 +23,84 @@ import scala.collection.immutable.VectorBuilder
...
-          rec(remainingHeaders.tail, method, scheme, authority, path, contentType, contentLength, headers += httpHeaderParser.resultHeader)
-      }
-
-    rec(subStream.initialHeaders.keyValuePairs)
+
+  def parseRequest(httpHeaderParser: HttpHeaderParser, serverSettings: ServerSettings, attributes: Attributes): Http2SubStream ⇒ HttpRequest = {
+    val remoteAddressHeader =
+      if (serverSettings.remoteAddressHeader) {
+        // in order to avoid searching all the time for the attribute, we need to guard it with the setting condition
+        attributes.get[HttpAttributes.RemoteAddress] match {
```
jrudolph: While you are at it, you can fix `HttpAttributes.RemoteAddress.address` to be an `InetSocketAddress` instead of an `Option[InetSocketAddress]`. The option does not seem to serve any purpose. This code can then be just

```scala
attributes.get[HttpAttributes.RemoteAddress].map(remote => model.headers.`Remote-Address`(RemoteAddress(remote.address)))
```

(Actually, no need for `OptionVal` here, sorry, did only notice it while reading it another time)

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -678,6 +685,32 @@ class Http2ServerSpec extends AkkaSpec("" + "akka.loglevel = debug")
...
+        val target = Uri("http://www.example.com/")
+        sendRequest(1, HttpRequest(uri = target))
+        requestIn.ensureSubscription()
+
+        val request = expectRequestRaw()
+        val remoteAddressHeader = request.header[headers.`Remote-Address`].get
+        remoteAddressHeader.address.getAddress.get().toString shouldBe ("/" + theAddress)
+        remoteAddressHeader.address.getPort shouldBe thePort
+
+        val streamIdHeader = request.header[Http2StreamIdHeader].getOrElse(Http2Compliance.missingHttpIdHeaderException)
```
jrudolph: I think it's ok to leave response handling out of the test code.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -689,8 +722,14 @@ class Http2ServerSpec extends AkkaSpec("" + "akka.loglevel = debug")
 
     def handlerFlow: Flow[HttpRequest, HttpResponse, NotUsed]
 
+    // hook to modify server, for example add attributes
+    def modifyServer(server: BidiFlow[HttpResponse, ByteString, ByteString, HttpRequest, NotUsed]) = server
```
jrudolph: :+1:

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -953,10 +992,11 @@ class Http2ServerSpec extends AkkaSpec("" + "akka.loglevel = debug")
     }
     def decodeHeadersToResponse(bytes: ByteString): HttpResponse =
       decodeHeaders(bytes).foldLeft(HttpResponse())((old, header) ⇒ header match {
-        case (":status", value)        ⇒ old.copy(status = value.toInt)
-        case ("content-length", value) ⇒ old.copy(entity = HttpEntity.Default(old.entity.contentType, value.toLong, Source.empty))
-        case ("content-type", value)   ⇒ old.copy(entity = old.entity.withContentType(ContentType.parse(value).right.get))
-        case (name, value)             ⇒ old.addHeader(RawHeader(name, value)) // FIXME: decode to modeled headers
+        case (":status", value)                             ⇒ old.copy(status = value.toInt)
+        case ("content-length", value) if value.toLong == 0 ⇒ old.copy(entity = HttpEntity.Empty)
```
jrudolph: :+1:

akka-http2-support/src/test/scala/akka/http/scaladsl/Http2ServerTest.scala
```diff
@@ -24,6 +24,9 @@ import scala.util.Random
 object Http2ServerTest extends App {
   val testConf: Config = ConfigFactory.parseString("""
     akka.loglevel = INFO
+
```
jrudolph: Left over from testing?

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
ktoso: Whoops, thanks

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
ktoso: True, reverting

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
ktoso: I was still operating under the old rule :-)
I don't mind if we explicitly want to use the other rule - should we add it to the contributing guide too then?

akka-http2-support/src/test/scala/akka/http/scaladsl/Http2ServerTest.scala
ktoso: yes, cleaned up

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
ktoso: yup, ok; it does not really check things in relation to the synthetic header after all

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpAttributes.scala
```diff
@@ -18,9 +18,9 @@ import akka.stream.Attributes
 private[akka] object HttpAttributes {
   import Attributes._
 
-  private[akka] final case class RemoteAddress(address: Option[InetSocketAddress]) extends Attribute
+  private[akka] final case class RemoteAddress(address: InetSocketAddress) extends Attribute
```
jrudolph: :+1:

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -119,7 +118,9 @@ private[http] object HttpServerBluePrint {
 
       override def onPush(): Unit = grab(in) match {
         case RequestStart(method, uri, protocol, hdrs, entityCreator, _, _) ⇒
+          val remoteAddress = inheritedAttributes.get[HttpAttributes.RemoteAddress].map(_.address)
```
jrudolph: Let's keep it where it was and where it only is executed once per connection.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/RequestParsing.scala
```diff
@@ -19,69 +24,79 @@ import scala.collection.immutable.VectorBuilder
...
-
-          httpHeaderParser.parseHeaderLine(ByteString(concHeaderLine))()
-          rec(remainingHeaders.tail, method, scheme, authority, path, contentType, contentLength, headers += httpHeaderParser.resultHeader)
-      }
-
-    rec(subStream.initialHeaders.keyValuePairs)
+
+  def parseRequest(httpHeaderParser: HttpHeaderParser, serverSettings: ServerSettings, attributes: Attributes): Http2SubStream ⇒ HttpRequest = {
+    val remoteAddressHeader: Option[`Remote-Address`] =
+      if (serverSettings.remoteAddressHeader) {
```
jrudolph: :+1:

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
ktoso: Ah good catch thanks

ktoso:
PLS BUILD

ktoso:
What the hell... `Caused by: org.scalatest.exceptions.TestFailedException: X-Real-Ip: 1.2.3.4 did not equal X-Real-Ip: 1.2.3.4` and passes locally... wat :|

ktoso:
Ok, passed and added failures to tickets.

PR #1092, opened 2017-05-04
=doc add info that spray is not affected by recent Accept vulnerability

jrudolph:


ktoso:
Ah we missed that indeed. LGTM

PR #1094, opened 2017-05-04
Fix several small http2 tests

jrudolph:


ktoso:
Let's put it on some milestone, how about the one for play?

ktoso:
Or just 10.0.7 actually... if something does not have an issue for tracking it makes sense to put the PR on a milestone IMHO, WDYT?

jrudolph:
I agree let's keep it on 10.0.7.

PR #1098, opened 2017-05-05
Modify OSGi header range of akka imports to upper limit of next major version (fix #1097)

### LINKED ISSUE #1097: Manifest header for OSGi Import-Package of akka package should allow range up to next minor version
The built akka-http libraries are incompatible with any akka library that is 2.5.0 and above because the generation of the Import-Package for akka defines a range that goes from the current dependent patch version up to (but not including) the next minor version.  

This should be changed so that is allows a range from the current lowest dependent version up to, but not including, the next major version.

For example, instead of the range being `[2.4.17, 2.5.0)`, it should be `[2.4.17, 3.0.0)`

michael-smith:
Addresses issue #1097 so that the Import-Package for `akka` packages are generated from the current dependent version up to, but not including, the next major version.
For example, `[2.4.17, 3.0.0)`

### DIFF
--- a/project/OSGi.scala
+++ b/project/OSGi.scala
@@ -50,7 +50,7 @@ object OSGi {
 
   val httpTestkit = osgiPkgHeaders(
     exports = Seq("akka.http.scaladsl.testkit.*", "akka.http.javadsl.testkit.*"),
-    imports = Seq("akka.testkit.*", "akka.stream.testkit.*", "org.junit") map { pkg => optionalResolution(minorVersioned(pkg)) }
+    imports = Seq("akka.testkit.*", "akka.stream.testkit.*", "org.junit") map { pkg => optionalResolution(minorVersionToNextMinorVersion(pkg)) }
   )
 
   val httpSprayJson = osgiPkgHeaders(exports = Seq("akka.http.scaladsl.marshallers.sprayjson"))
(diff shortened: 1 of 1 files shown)


jlprat:
OK TO TEST

project/OSGi.scala
```diff
@@ -75,13 +75,14 @@ object OSGi {
   private lazy val akkaParboiledImport = projectVersioned("akka.parboiled2.*")
   private lazy val akkaShapelessImport = projectVersioned("akka.shapeless.*")
   private lazy val akkaMacrosImport = projectVersioned("akka.macros.*")
-  private lazy val akkaImport = "akka," + patchVersioned("akka.*")
-  private lazy val configImport = patchVersioned("com.typesafe.config.*")
-  private lazy val scalaImport = minorVersioned("scala.*")
+  private lazy val akkaImport = "akka," + patchVersionToNextMajorVersion("akka.*")
+  private lazy val configImport = patchVersionToNextMinorVersion("com.typesafe.config.*")
+  private lazy val scalaImport = minorVersionToNextMinorVersion("scala.*")
```
jonas: Shouldn't Scala be patch version to next patch version? For example, 2.12.0 => 2.12.1.

ktoso: I think this means "until 2.13 but excluding that one" which means all patches right @michael-smith ?

ktoso:
fixes #1097

project/OSGi.scala
michael-smith: Yes, @ktoso . The scala import would turn into a range of `[2.12.0, 2.13.0)` where `[` is inclusive and `)` is exclusive.

ktoso: Thanks for sanity checking, seems this is good to merge IMHO.

ktoso:
LGTM, will merge tomorrow - in case someone osgi knowing still has a second to look.

ktoso:
Has had enough wait time, merging.

Issue #1102, opened 2017-05-08
Should response/request.header[T] be ModeledCustomHeader aware?


ktoso:
This requires a bunch of hacks though, as the impl that makes it work in akka-http is the `HeaderMagnet` infra which we do not have immediately available in akka-http-core, and I'm not sure if we want to have it there.

https://github.com/akka/akka-http/blob/master/akka-http/src/main/scala/akka/http/javadsl/server/directives/HeaderDirectives.scala#L84

We can make it work using some smarts however making it possible for Java should likely use the partial function style as we suggest anyway for the routing DSL for it.

czare1:
At least the documentation should mention that it won't work this way.

jroper:
Agreed, the documentation does need to be updated, there's no mention [here](https://doc.akka.io/docs/akka-http/10.0/common/http-model.html) that if you create a custom header, and try to use it with `request.header[MyCustomHeader]`, it won't work.

jroper:
I think we need to do more than just document this. Even we are [tripped up](https://github.com/akka/akka-grpc/issues/1897) by this, I know I've spent multiple hours staring at my code not understanding why it doesn't work because of this. If that's our experience, what of a regular akka-http user?

johanandren:
If we implement this, it still means users must remember to register modeled header types, if they forget, we'd still have the same problem as now where it seems correct but always returns `None`.

PR #1103, opened 2017-05-08
=htc #1102 allow r.header[ModeledCustomHeader] to work

ktoso:
This allows #1102 to work in Scala, we use a similar trick elsewhere already (when javadsl uses directives to find a header by type, but it turns out it is a scala custom modeled one).

Though this will not work in JavaDSL, I think there we should do something else I guess.
The JavaDSL style also recommends `BlaBla / BlaBlaFactory` nowadays so it's hard to find that factory - we should ask for it explicitly (bad API), or instead provide `getHeader(partial function)`  I think (slightly better and consistent with how routing looks for java), WDYT?

### DIFF
--- a/akka-http-core/src/main/java/akka/http/javadsl/model/headers/ModeledCustomHeader.java
+++ b/akka-http-core/src/main/java/akka/http/javadsl/model/headers/ModeledCustomHeader.java
@@ -10,7 +10,7 @@ import akka.util.Helpers;
 /**
  * Support class for building user-defined custom headers defined by implementing `name` and `value`.
  * By implementing a {@link ModeledCustomHeader} along with {@link ModeledCustomHeaderFactory} instead of {@link CustomHeader} directly,
- * convenience methods that allow parsing this user-defined header from {@link akka.http.javadsl.model.HttpHeader}  are
+ * convenience methods that allow parsing this user-defined header from {@link akka.http.javadsl.model.HttpHeader} are
  * available to use.
  */
 public abstract class ModeledCustomHeader extends CustomHeader {
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
@@ -9,7 +9,7 @@ import akka.stream.{ FlowShape, Graph }
 import java.io.File
 import java.nio.file.Path
 import java.lang.{ Iterable ⇒ JIterable }
-import java.util.Optional
+import java.util.{ Locale, Optional }
 import java.util.concurrent.CompletionStage
 
 import scala.compat.java8.FutureConverters
@@ -19,6 +19,7 @@ import scala.collection.immutable
 import scala.compat.java8.OptionConverters._
 import scala.reflect.{ ClassTag, classTag }
 import akka.Done
+import akka.actor.ReflectiveDynamicAccess
 import akka.parboiled2.CharUtils
 import akka.stream.Materializer
 import akka.util.{ ByteString, HashCode, OptionVal }
@@ -26,6 +27,8 @@ import akka.http.impl.util._
 import akka.http.javadsl.{ model ⇒ jm }
 import akka.http.scaladsl.util.FastFuture._
 import headers._
+
+import scala.util.Success
 
 /**
  * Common base class of HttpRequest and HttpResponse.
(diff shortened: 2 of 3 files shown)


ktoso:
PLS BUILD

jlprat:
The only way I can image we could get the factory for a given `ModeledCustomHeader` is using annotations. But I'm not so sure we want to go that way. 

Basically, factories should have this annotation that would link back to the `ModeledCustomHeader`, for the directive then, we "just" need to scan all classes with given annotation and take the one we want.

akka-http-core/src/main/java/akka/http/javadsl/model/headers/ModeledCustomHeader.java
```diff
@@ -10,7 +10,7 @@
 /**
  * Support class for building user-defined custom headers defined by implementing `name` and `value`.
  * By implementing a {@link ModeledCustomHeader} along with {@link ModeledCustomHeaderFactory} instead of {@link CustomHeader} directly,
- * convenience methods that allow parsing this user-defined header from {@link akka.http.javadsl.model.HttpHeader}  are
+ * convenience methods that allow parsing this user-defined header from {@link akka.http.javadsl.model.HttpHeader} are
```
jlprat: nice catch!

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
```diff
@@ -106,13 +109,30 @@ sealed trait HttpMessage extends jm.HttpMessage {
...
+    // allows locating "companion" without implicits, same trick is applied in javadsl
+    val refl = new ReflectiveDynamicAccess(getClass.getClassLoader)
+    refl.getObjectFor[ModeledCustomHeaderCompanion[_]](clazz.getName) match {
+      case Success(companion) ⇒
+        headers.collectFirst {
+          case RawHeader(name, value) if name.toLowerCase(Locale.ROOT) == companion.lowercaseName ⇒ companion(value)
+        }
+      case _ ⇒
+        throw new IllegalStateException("Attempted to find ModeledCustomHeader via `header[...]` method, " +
+          "however unable to locate the custom headers companion, which would be used to parse the header.") // TODO should we try to find the companion by name?
```
jlprat: `the custom header's companion` if I'm not mistaken. Anyway this sentence sounds strange as it doesn't have a subject.
Maybe `However, the custom header's companion class, which would be used to parse the header, could not be located.`

akka-http-tests/src/test/scala/akka/http/scaladsl/server/ModeledCustomHeaderSpec.scala
```diff
@@ -131,11 +140,11 @@ class ModeledCustomHeaderSpec extends RoutingSpec {
 
     "fail with useful message when unable to parse" in {
       val ex = intercept[Exception] {
-        DifferentHeader("Hello world") // illegal " "
+        DifferentHeader("Hello xxx world") // illegal " "
```
jlprat: change comment to `// illegal, it contains "x"`

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
jlprat: Regarding trying to locate by name, which are the chances to find it this way if it couldn't be located with the previous method?

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
```diff
@@ -106,13 +109,30 @@ sealed trait HttpMessage extends jm.HttpMessage {
...
     }
   }
 
+  private def findHeaderViaModeledCompanion(clazz: Class[_]) = {
+    // allows locating "companion" without implicits, same trick is applied in javadsl
+    val refl = new ReflectiveDynamicAccess(getClass.getClassLoader)
+    refl.getObjectFor[ModeledCustomHeaderCompanion[_]](clazz.getName) match {
+      case Success(companion) ⇒
+        headers.collectFirst {
+          case RawHeader(name, value) if name.toLowerCase(Locale.ROOT) == companion.lowercaseName ⇒ companion(value)
```
jrudolph: Parsing here introduces a new failure condition for the method which I'd rather not like to introduce. Also it's inconsistent with what we do for regular modeled headers which won't be found if they only exist as a `RawHeader`.

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
```diff
@@ -106,13 +109,30 @@ sealed trait HttpMessage extends jm.HttpMessage {
...
+      findHeaderViaModeledCompanion(clazz).asInstanceOf[Option[T]]
+    } else HttpHeader.fastFind[T](clazz, headers) match {
       case OptionVal.Some(h)                     ⇒ Some(h)
       case _ if clazz == classOf[`Content-Type`] ⇒ Some(`Content-Type`(entity.contentType)).asInstanceOf[Option[T]]
       case _                                     ⇒ None
     }
   }
 
+  private def findHeaderViaModeledCompanion(clazz: Class[_]) = {
+    // allows locating "companion" without implicits, same trick is applied in javadsl
```
jrudolph: I wonder if it wouldn't make more sense to try to find better ways in the long-term which use implicits for the same effect.

ktoso:
> How about trying to create an new method in HttpMessage that transparently parses headers as found? For that method we can even make it work with implicits for all cases.

That is an interesting idea, I can try it out later
// Slipping from 10.0.7 though

jrudolph:
Let's close this one for now and experiment with alternative solutions.

ktoso:
Ok, unmarked in progress and added tags then

Issue #1071, opened 2017-04-28
FAILED: HttpExtensionApiSpec.The Java HTTP extension should properly bind and handle a server with a flow (with seven parameters)


jlprat:
https://jenkins.akka.io:8498/job/pr-validator-akka-http/1016/consoleFull
```
Error Message

akka.stream.StreamTcpException: Tcp command [Connect(localhost:41269,None,List(),Some(10 seconds),true)] failed
Stacktrace

java.util.concurrent.ExecutionException: akka.stream.StreamTcpException: Tcp command [Connect(localhost:41269,None,List(),Some(10 seconds),true)] failed
	at java.util.concurrent.CompletableFuture.reportGet(CompletableFuture.java:357)
	at java.util.concurrent.CompletableFuture.get(CompletableFuture.java:1915)
	at scala.concurrent.java8.FuturesConvertersImpl$CF.scala$concurrent$java8$FuturesConvertersImpl$CF$$super$get(FutureConvertersImpl.scala:84)
	at scala.concurrent.java8.FuturesConvertersImpl$CF$$anonfun$get$2.apply(FutureConvertersImpl.scala:84)
	at scala.concurrent.BlockContext$DefaultBlockContext$.blockOn(BlockContext.scala:53)
	at scala.concurrent.package$.blocking(package.scala:123)
	at scala.concurrent.java8.FuturesConvertersImpl$CF.get(FutureConvertersImpl.scala:84)
	at akka.http.javadsl.HttpExtensionApiSpec.akka$http$javadsl$HttpExtensionApiSpec$$waitFor(HttpExtensionApiSpec.scala:473)
	at akka.http.javadsl.HttpExtensionApiSpec$$anonfun$1$$anonfun$apply$mcV$sp$7.apply(HttpExtensionApiSpec.scala:154)
	at akka.http.javadsl.HttpExtensionApiSpec$$anonfun$1$$anonfun$apply$mcV$sp$7.apply(HttpExtensionApiSpec.scala:144)
	at org.scalatest.OutcomeOf$class.outcomeOf(OutcomeOf.scala:85)
	at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
	at org.scalatest.Transformer.apply(Transformer.scala:22)
	at org.scalatest.Transformer.apply(Transformer.scala:20)
	at org.scalatest.WordSpecLike$$anon$1.apply(WordSpecLike.scala:1078)
[...]

PR #1105, opened 2017-05-08
=htc #1071 increase timeout and wait for unbind

ktoso:
so we dont have too many servers running at the same time, which may have been a reason for
tiem timeouts - just too many servers active at the same time perhaps

The timeouts max is 10 seconds, but usually it won't take that long, but is better than failing once in a while randomly

Resolves https://github.com/akka/akka-http/issues/1071


ktoso:
Made a silly typo

PR #1107, opened 2017-05-08
=pro add 10.0.6 to mima versions

jrudolph:


jrudolph:
That's why we have the checklist at least ;)

jrudolph:
Hmm, validation missed to run the mima check on this change and so missed incompatibilities we already introduced after 10.0.6. Seems the PR validation filter is too lenient.

PR #1108, opened 2017-05-08
Add missing 10.0.6 mima rules

jrudolph:
and also trigger PR validation when Mima.scala changes.


jrudolph:
Will merge right away after validation to unblock PR validation for other PRs.

PR #1122, opened 2017-05-13
Fix akka-docs extref to expand the akka version sbt setting

jonas:
The regression was introduced in ffd69f87ac1a18a393fff28be285ef7989cd2c32.


jonas:
Merging since this one is trivial

jrudolph:
Good catch, thanks.

PR #1123, opened 2017-05-13
Reduce number of warnings when compiling with Scala 2.13

jonas:


jonas:
Inspired by akka/akka#22581

jonas:
Test failure looks similar to #299 - https://jenkins.akka.io:8498/job/pr-validator-akka-http/1106/testReport/junit/akka.http.impl.engine.client/ConnectionPoolSpec/be_able_to_handle_500__Connection__close__requests_against_the_test_server/

jonas:
PLS BUILD

akka-http-core/src/test/scala/akka/http/impl/engine/parsing/HttpHeaderParserSpec.scala
```diff
@@ -284,8 +284,8 @@ abstract class HttpHeaderParserSpec(mode: String, newLine: String) extends WordS
       val c = nextRandomPrintableChar()
       if (CharacterClasses.ALPHANUM(c)) c else nextRandomAlphaNumChar()
     }
-    @tailrec final def nextRandomString(charGen: () ⇒ Char, len: Int, sb: JStringBuilder = new JStringBuilder): String =
-      if (sb.length < len) nextRandomString(charGen, len, sb.append(charGen())) else sb.toString
+    @tailrec final def nextRandomString(charGen: ⇒ Char, len: Int, sb: JStringBuilder = new JStringBuilder): String =
```
jrudolph: Not sure, if we want this change. Is this used somewhere, btw?

jonas: Yes, it is used. I will revert this for now. It might also be worth it to consider using scalacheck instead.

jrudolph:
Thanks, @jonas.

jrudolph:
Btw. how many warnings were removed and how much are left? ;)

jonas:
I didn't count, but there are quite a few remaining related with "automatic tuple adaptations", which of course would make the use of the magnet pattern more verbose.

Issue #952, opened 2017-03-09
ScalatestRouteTest ignoring withRequestTimeout in Route


miguel-leon:
Hi,

In this code for example, similar to the code in the akka-http documentation:

```scala
path("test") {
  withRequestTimeout(100 millis) {
    onSuccess(Future { Thread.sleep(300) }) {
      complete(OK)
    }
  }
}
```

If I then write a test like such:

```scala
Get("/test") ~> route ~> check {
  status shouldBe ServiceUnavailable
}
```

I get a log WARN saying "withRequestTimeout was used in route however no request-timeout is set!"
and the test fails with "200 OK was not equal to 503 Service Unavailable"

However if I test this manually, say with Postman, it works as expected.

miguel-leon:
Needless to say, I didn't find documentation for this or comments of this happening to someone else anywhere on the internet.

ktoso:
This is because a route test like this avoids starting the real http server, the ticket that would solve it is: https://github.com/akka/akka-http/issues/39

Basically, if you want to test timeouts or "the entire http stack" you should start the full http server in your test nowadays.

I'll keep the issue open but the solution is the linked Issue I believe.

jrudolph:
Will be solved in 10.2.0 with the new route integration testing merged in #3014. Use `~!> check {}` in tests instead of `~> check`.

l15k4:
The warning is logged even with `~!>` : 
```
    "add headers to a timed out response" in {
      Get("/api/valid").withHeaders(Origin("http://www.example.com")) ~!> route ~> check {
        status shouldBe StatusCodes.ServiceUnavailable
        headers should contain(`Access-Control-Allow-Origin`(HttpOrigin("http://www.example.com")))
      }
    }
```

jrudolph:
> I'm troubleshooting why `cors` does not set the response headers if timeout happens.

The timeout is generated outside of the route logic. You can use the [`withRequestTimeoutResponse` directive](https://doc.akka.io/docs/akka-http/current/routing-dsl/directives/timeout-directives/withRequestTimeoutResponse.html) to change the generated response.

l15k4:
Yeah, I just did that, I figured it is not a coincidence that the `handler` is there :-) thanks

PR #1125, opened 2017-05-14
[WIP] Route integration testing

jonas:
I am interested to know if this approach makes sense. By extending `RouteTest` we get most of the integration test code for free with the nice API for checking the response, however, there might also be "holes" where the existing route test API doesn't apply to integration tests.

Still missing:
- [ ] Improve documentation
- [ ] Java testkit API
- [x] Support for customizing the server via implicit `ServerSettings`?

References #39, #952


jonas:
Implicit `ServerSetting` is passed to `Http().bindAndHandle` and the `RoutingSetings` passed via `withSettings` directive wrapping the original route.

raboof:
This seems pretty nice!

jrudolph:
I mostly finished it in #3014.

Issue #1129, opened 2017-05-15
Update documentation to use the new Akka theme and paradox plugin


jonas:
See related bits from the Akka repo:

 - https://github.com/akka/akka/blob/wip-akka-docs-new/project/plugins.sbt
 - https://github.com/akka/akka/blob/wip-akka-docs-new/akka-docs/build.sbt

jrudolph:
Let's keep this open until we have a final release of sbt-akka-paradox.

PR #1130, opened 2017-05-15
=doc #1129 Use the new Akka theme and Paradox plugin

jonas:
Generated docs can be seen here: https://jonas.github.io/akka-http/paradox-theme-akka/index.html

/cc @patriknw @jlprat


jlprat:
I took a quick look, and they look very nice. I would change something if possible.
https://jonas.github.io/akka-http/paradox-theme-akka/scala.html should not be there, and point directly to https://jonas.github.io/akka-http/paradox-theme-akka/scala/http/index.html

jlprat:
I guess we need deeper TOC depth for these kin of pages now: https://jonas.github.io/akka-http/paradox-theme-akka/scala/http/routing-dsl/index.html

build.sbt
```diff
@@ -165,6 +166,8 @@ lazy val docs = project("docs")
         val v = if (isSnapshot.value) "current" else version.value
         s"http://doc.akka.io/japi/akka-http/$v"
       },
+      "algolia.docsearch.api_key" -> "0ccbb8bf5148554a406fbf07df0a93b9",
+      "algolia.docsearch.index_name" -> "akka-http",
```
patriknw: something for google analytics also https://github.com/akka/akka/pull/22942/files

jonas: Added the akka.io ones from that PR in b1434b66ed194ca8d2e77bcba8af6d8891dbd35f

jonas:
@jlprat Yes, good point. Removed the "middle-man".

Regarding ToC level, do we even need those with level one? The new layout is not very efficient for listing content deep in the structure.

jonas:
Failures are #299 and https://jenkins.akka.io:8498/job/pr-validator-akka-http/1120/testReport/junit/akka.http.impl.model.parser/HttpHeaderSpec/The_HTTP_header_model_must_correctly_parse_and_render_the_headers_X_Real_Ip/

jlprat:
This type of pages look not so nice, as they have big boxes instead of a ToC. So I would either remove them or show them with 2 levels.
Removing is probably the best option

jonas:
PLS BUILD

jonas:
Depends on akka/akka-paradox#1

docs/src/main/paradox/java.md
```diff
@@ -1,10 +0,0 @@
-<a id="java-api"></a>
```
jrudolph: So nice to see those go! :100:

project/plugins.sbt
```diff
@@ -23,5 +23,6 @@ addSbtPlugin("net.virtual-void" % "sbt-dependency-graph" % "0.8.2") // for advan
 addSbtPlugin("io.spray" % "sbt-boilerplate" % "0.6.0")
 addSbtPlugin("me.lessis" % "bintray-sbt" % "0.3.0-8-g6d0c3f8")
 addSbtPlugin("com.lightbend.sbt" % "sbt-javaagent" % "0.1.2")
+addSbtPlugin("com.lightbend.akka" % "sbt-paradox-akka" % "f811271f")
```
jrudolph: Using these for now is fine for me. We'll hopefully get a final release pretty soon.

jrudolph:
Let's take these changes as is because it's also needed to make the integrated build with akka work again.

jrudolph:
Refs #1129

PR #1135, opened 2017-05-16
Optimize route sealing

jrudolph:
Smallish optimization for base routing infrastructure.

### DIFF
--- a/akka-http/src/main/scala/akka/http/scaladsl/server/RejectionHandler.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/server/RejectionHandler.scala
@@ -12,7 +12,6 @@ import akka.http.scaladsl.server.directives.BasicDirectives
 
 import scala.annotation.tailrec
 import scala.collection.immutable
-import scala.concurrent.ExecutionContext
 import scala.reflect.ClassTag
 
 trait RejectionHandler extends (immutable.Seq[Rejection] ⇒ Option[Route]) { self ⇒
@@ -152,7 +151,7 @@ object RejectionHandler {
    * Default [[RejectionHandler]] instance.
    */
   final val default =
-    newBuilder()
+    new Builder(isDefault = true)
       .handleAll[SchemeRejection] { rejections ⇒
         val schemes = rejections.map(_.supported).mkString(", ")
         complete((BadRequest, "Uri scheme not allowed, supported schemes: " + schemes))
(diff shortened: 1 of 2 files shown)


akka-http/src/main/scala/akka/http/scaladsl/server/RejectionHandler.scala
```diff
@@ -261,6 +269,7 @@ object RejectionHandler {
       .handle { case ValidationRejection(msg, _) ⇒ complete((BadRequest, msg)) }
       .handle { case x ⇒ sys.error("Unhandled rejection: " + x) }
       .handleNotFound { complete((NotFound, "The requested resource could not be found.")) }
+      .asDefault()
```
jonas: Not sure if you consider to just change how the builder is instantiated.

```scala
   final val default =
     new Builder(isDefault = true)
     ...
```

jrudolph: Yeah, good point, that should suffice for that single use case of the flag.

jrudolph: Yes, the single line change looks much better ;) Thanks for the hint.

akka-http/src/main/scala/akka/http/scaladsl/server/Route.scala
```diff
@@ -37,11 +37,9 @@ object Route {
...
                          exceptionHandler: ExceptionHandler = null): Route = {
     import directives.ExecutionDirectives._
-    handleExceptions(ExceptionHandler.seal(exceptionHandler)) {
-      handleRejections(rejectionHandler.seal) {
-        route
-      }
-    }
+    // optimized as this is the root handler for all akka-http applications
+    (handleExceptions(ExceptionHandler.seal(exceptionHandler)) & handleRejections(rejectionHandler.seal))
+      .tapply(_ ⇒ route) // execute above directives eagerly, avoiding useless laziness of Directive.addByNameNullaryApply
```
ktoso: nice one, do you remember how much it showed up in profiling?

ktoso:
LGTM

akka-http/src/main/scala/akka/http/scaladsl/server/Route.scala
jrudolph: Quite a lot in the techempower example. But the bigger share is that marshalling is still somewhat slow.

PR #1138, opened 2017-05-16
Fix 10.0.x migration guide

jonas:
Change `AkkaHttp` to `HttpApp` in the migration guide. I also took the liberty to rename the file to better match the others, the title remains the same.

### DIFF
--- a/docs/src/main/paradox/scala/http/migration-guide/index.md
+++ b/docs/src/main/paradox/scala/http/migration-guide/index.md
@@ -7,6 +7,6 @@
 
 * [migration-from-spray](migration-from-spray.md)
 * [migration-guide-2.4.x-10.0.x](migration-guide-2.4.x-10.0.x.md)
-* [migration-within-10.0.x-Api-May-Change](migration-within-10.0.x-Api-May-Change.md)
+* [migration-guide-10.0.x](migration-guide-10.0.x.md)
 
 @@@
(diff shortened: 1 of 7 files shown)


docs/src/main/paradox/scala/http/migration-guide/index.md
```diff
@@ -7,6 +7,6 @@
 
 * [migration-from-spray](migration-from-spray.md)
 * [migration-guide-2.4.x-10.0.x](migration-guide-2.4.x-10.0.x.md)
-* [migration-within-10.0.x-Api-May-Change](migration-within-10.0.x-Api-May-Change.md)
+* [migration-guide-10.0.x](migration-guide-10.0.x.md)
```
jlprat: No strong feelings, now it matches the pattern. I used this name to show clearly that it's only about changes on `ApiMayChange`.

jonas: My main worry is that at some point we will switch to Akka 2.5 and we'll need to document some hoops to jump through when moving from Akka 2.4, like migrating to use AbstractActor.

jlprat: Fair point.

jrudolph: Yes, let's keep it simple and collect all those information in one place.

In the middle-term, we'll try not to lose compatibility with 2.4 so it should at least be possible to still use Akka 2.4 for a while regardless of whether we compile our binaries against 2.5 or 2.4.

Issue #1139, opened 2017-05-17
Akka rejects requests when UTF-8 charset is explicitly specified in Accept header along with `application/json`


Leammas:
The problem occurred after updating to 10.0.6 from 10.0.5. I've made failing unit test here: 
https://github.com/Leammas/akka-http/commit/194378e5927afd1e735cc5061a4826e3ad6aa185

According to [iana](https://www.iana.org/assignments/media-types/application/json) the "charset" parameter should not affect compatibility.

The problem was introduced [here](https://github.com/akka/akka-http/commit/4758a44219b5e8df686b69717f192d3ffa21edb5#diff-95adf484e1280a1b0d76142e2901b916R98) but I'm not sure how to fix it without breaking new Accept-handling behavior

jrudolph:
The behavior you describe is implemented when the `charset` parameter is offered using the `Accept-Charset` header. However, [RFC 7231, section 5.3.2](https://tools.ietf.org/html/rfc7231#section-5.3.2) explicitly mentions that `charset` is an allowed parameter to put into the `Accept`-header as well so arguably putting it there instead of using `Accept-Charset` should make no difference.

That means, we might need to treat the charset parameter in the `Accept` header specially. @raboof wdyt?

jrudolph:
Maybe we could improve the line at https://github.com/akka/akka-http/commit/4758a44219b5e8df686b69717f192d3ffa21edb5#diff-95adf484e1280a1b0d76142e2901b916R98, to check a mediarange charset parameter against the charset given in the fixed charset mediatype?

Leammas:
Is the solution suggested in the PR #1140 acceptable?

2m:
Fixed in #1140

PR #1140, opened 2017-05-17
Add specific handling for `charset` param when comparing MediaTypes #1139

Leammas:
Refs #1139.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/MediaRange.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/MediaRange.scala
@@ -8,6 +8,7 @@ import language.implicitConversions
 import java.util
 import akka.http.impl.util._
 import akka.http.javadsl.{ model ⇒ jm }
+import akka.http.scaladsl.model.MediaType.WithFixedCharset
 
 sealed abstract class MediaRange extends jm.MediaRange with Renderable with WithQValue[MediaRange] {
   def value: String
@@ -48,6 +49,12 @@ sealed abstract class MediaRange extends jm.MediaRange with Renderable with With
 }
 
 object MediaRange {
+
+  private[http] def extractCharset(mediaType: MediaType): Option[HttpCharset] = mediaType match {
+    case mt: WithFixedCharset ⇒ Some(mt.charset)
+    case _: MediaType         ⇒ None
+  }
+
   private[http] def splitOffQValue(params: Map[String, String], defaultQ: Float = 1.0f): (Map[String, String], Float) =
     params.get("q") match {
       case Some(x) ⇒ (params - "q") → (try x.toFloat catch { case _: NumberFormatException ⇒ 1.0f })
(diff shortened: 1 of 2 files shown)


jlprat:
OK TO TEST

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/MarshallingDirectivesSpec.scala
```diff
@@ -224,5 +224,11 @@ class MarshallingDirectivesSpec extends RoutingSpec with Inside {
         rejection shouldEqual UnacceptedResponseContentTypeRejection(Set(ContentType(`application/json`)))
       }
     }
+    "render JSON response when `Accept` header is present with charset UTF-8 explicitly specified" in {
```
jrudolph: I think we need more tests for more combination of `Accept` / `Accept-Charset` parameters to be sure it works with all combinations.

Leammas: Well it's a bit harder because the first obvious case is when `charset` in Accept header is non-UTF. And here we come to the point that Accept header in the akka-http model has no typed method to distinguish charsets (like `HttpCharsetRange` in the `Accept-Charset`). 
Moreover, I've found another failing case: when Accept-Charset is Unicode but not UTF-8 (UTF-16 for example). It should be served as well according to [RFC](https://tools.ietf.org/html/rfc4627#section-3).
Should I try to introduce optional `HttpCharsetRange` param in the `Accept` header class?

Leammas:
I reviwed RFCs and IANA and came to the next point: 

[RFC2616](https://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html) (section 3.7) passes parameter definitions to IANA. And IANA states for `application/json` that adding `charset` parameter should has no effect. So the `charset` parameter could be ignored at least for `json`, not processed as the `Accept-Charset`. 
And the usage of UTF-16 is probably out of this issue.

More tests added.

akka-http-core/src/main/scala/akka/http/scaladsl/model/MediaRange.scala
```diff
@@ -95,7 +102,10 @@ object MediaRange {
     def matches(mediaType: MediaType) =
       this.mediaType.mainType == mediaType.mainType &&
         this.mediaType.subType == mediaType.subType &&
-        this.mediaType.params.forall { case (key, value) ⇒ mediaType.params.get(key).contains(value) }
+        this.mediaType.params
+        .filterNot { case (key, value) ⇒ key == "charset" }
+        .forall { case (key, value) ⇒ mediaType.params.get(key).contains(value) } &&
+        extractCharset(this.mediaType) == extractCharset(mediaType)
```
jrudolph: This line has no effect so far because `mt.charset` above will always return the same fixed charset. In `this.mediaType` the custom charset is instead part of the params.

jrudolph:
I removed the extra test which is unlikely to ever make the match fail (and which wasn't tested to fail anyway). I think in the current form, just ignoring the `charset` parameter in `Accept` headers is a reasonable choice for now.

jrudolph:
Thanks for the contribution, @Leammas!

PR #1141, opened 2017-05-17
Fix #1134: Create representation of Content-Type with missing charset

### LINKED ISSUE #1134: Akka incorrectly assumes UTF-8 for incoming text/xml media type without charset
According to [iana](https://www.iana.org/assignments/media-types/text/xml), the `text/xml` media type's `charset` property is optional.

When creating a `ContentType` in akka itself, as the http client, we currently require providing a charset for `text/xml`, which is a great recommendation.

However, as a server, we can't control what clients send. But an incoming request which just has `Content-Type: text/xml` is now represented as `ContentType(text/xml, charset=UTF-8)`, e.g.

```
      path("echo") {
        extractRequest { rq => 
          complete("" + rq.header[`Content-Type`])
        }
      }
```
and the client:
```
curl -v -X POST -H 'Content-Type: text/xml' -d '<?xml><hello/>' http://localhost:8889/echo
Some(Content-Type: text/xml; charset=UTF-8)
```

where I would have hoped to see a `ContentType` with `mediaType` `text/xml` and `charsetOption` of `None`.

jypma:
The text/xml content-type (and possibly others) is best used with a
charset parameter, and akka inforces this on the type level. That's
great practice, but _incoming_ requests can certainly have a
Content-Type header of "text/xml" without a charset. Akka currently
just blindly adds "UTF-8", which is incorrect and a potential problem.

This commit adds a specific type to the ContentType hierarchy to represent
"a media type that should have had a charset but didn't".

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/javadsl/model/ContentType.scala
+++ b/akka-http-core/src/main/scala/akka/http/javadsl/model/ContentType.scala
@@ -5,30 +5,35 @@ package akka.http.javadsl.model
 
 import java.util.Optional
 
+// Has to be defined in Scala even though it's JavaDSL because of:
+// https://issues.scala-lang.org/browse/SI-9621
+object ContentType {
+  /** Represents a content-type which we know not to contain text (will never have a charset) */
+  trait Binary extends ContentType
+
+  /** Represents a content-type which we know to contain text, and has a specified charset. */
+  trait NonBinary extends ContentType {
+    def charset: HttpCharset
+  }
+
+  /**
+   * Represents a content-type which we know to contain text, and would be better off having a charset,
+   * but the client hasn't provided that. For example, "text/xml" without a charset parameter.
+   */
+  trait WithMissingCharset extends ContentType
+
+  /** Represents a content-type which we know to contain text, where the charset always has the same predefined value. */
+  trait WithFixedCharset extends NonBinary
+
+  /** Represents a content-type which we know to contain text, and the charset is known at runtime. */
+  trait WithCharset extends NonBinary
+}
+
 /**
  * Represents an Http content-type. A content-type consists of a media-type and an optional charset.
  *
  * See [[ContentTypes]] for convenience access to often used values.
  */
-// Has to be defined in Scala even though it's JavaDSL because of:
-// https://issues.scala-lang.org/browse/SI-9621
-object ContentType {
-
-  trait Binary extends ContentType {
-  }
-
-  trait NonBinary extends ContentType {
-    def charset: HttpCharset
-  }
-
-  trait WithFixedCharset extends NonBinary {
-  }
-
-  trait WithCharset extends NonBinary {
-  }
-
-}
-
 trait ContentType {
   /**
    * The media-type of this content-type.
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/ContentType.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/ContentType.scala
@@ -30,9 +30,10 @@ object ContentTypeRange {
   implicit def apply(mediaRange: MediaRange): ContentTypeRange = apply(mediaRange, HttpCharsetRange.`*`)
   implicit def apply(contentType: ContentType): ContentTypeRange =
     contentType match {
-      case ContentType.Binary(mt)           ⇒ ContentTypeRange(mt)
-      case ContentType.WithFixedCharset(mt) ⇒ ContentTypeRange(mt)
-      case ContentType.WithCharset(mt, cs)  ⇒ ContentTypeRange(mt, cs)
+      case ContentType.Binary(mt)             ⇒ ContentTypeRange(mt)
+      case ContentType.WithFixedCharset(mt)   ⇒ ContentTypeRange(mt)
+      case ContentType.WithCharset(mt, cs)    ⇒ ContentTypeRange(mt, cs)
+      case ContentType.WithMissingCharset(mt) ⇒ ContentTypeRange(mt)
     }
 }
 
@@ -53,27 +54,41 @@ sealed trait ContentType extends jm.ContentType with ValueRenderable {
 }
 
 object ContentType {
+  /** Represents a content-type which we know not to contain text (will never have have a charset) */
   final case class Binary(mediaType: MediaType.Binary) extends jm.ContentType.Binary with ContentType {
     def binary = true
     def charsetOption = None
   }
 
+  /** Represents a content-type which we know to contain text, and has a specified charset. */
   sealed trait NonBinary extends jm.ContentType.NonBinary with ContentType {
     def binary = false
     def charset: HttpCharset
     def charsetOption = Some(charset)
   }
 
+  /** Represents a content-type which we know to contain text, where the charset always has the same predefined value. */
   final case class WithFixedCharset(val mediaType: MediaType.WithFixedCharset)
     extends jm.ContentType.WithFixedCharset with NonBinary {
     def charset = mediaType.charset
   }
 
+  /** Represents a content-type which we know to contain text, and the charset is known at runtime. */
   final case class WithCharset(val mediaType: MediaType.WithOpenCharset, val charset: HttpCharset)
     extends jm.ContentType.WithCharset with NonBinary {
 
     private[http] override def render[R <: Rendering](r: R): r.type =
       super.render(r) ~~ ContentType.`; charset=` ~~ charset
+  }
+
+  /**
+   * Represents a content-type which we know to contain text, and would be better off having a charset,
+   * but the client or server hasn't provided that. For example, "text/xml" without a charset parameter.
+   */
+  final case class WithMissingCharset(mediaType: MediaType.WithOpenCharset)
+    extends jm.ContentType.WithMissingCharset with ContentType {
+    def binary = false
+    def charsetOption = None
   }
 
   implicit def apply(mediaType: MediaType.Binary): Binary = Binary(mediaType)
--- a/akka-http-core/src/test/scala/akka/http/impl/model/parser/HttpHeaderSpec.scala
+++ b/akka-http-core/src/test/scala/akka/http/impl/model/parser/HttpHeaderSpec.scala
@@ -208,6 +208,8 @@ class HttpHeaderSpec extends FreeSpec with Matchers {
         `Content-Type`(`application/json`)
       "Content-Type: text/plain; charset=utf8" =!=
         `Content-Type`(ContentType(`text/plain`, `UTF-8`)).renderedTo("text/plain; charset=UTF-8")
+      "Content-Type: text/plain" =!=
+        `Content-Type`(ContentType.WithMissingCharset(MediaTypes.`text/plain`)).renderedTo("text/plain")
       "Content-Type: text/xml2; version=3; charset=windows-1252" =!=
         `Content-Type`(MediaType.customWithOpenCharset("text", "xml2", params = Map("version" → "3"))
           withCharset HttpCharsets.getForKey("windows-1252").get)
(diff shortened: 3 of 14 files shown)


jypma:
@jrudolph first shot at `WithMissingContentType`. Let's hope I made some tests fail :-)

jypma:
Fluke failure in [ResponseRendererSpec](https://jenkins.akka.io:8498/job/pr-validator-akka-http/1136/testReport/junit/akka.http.impl.engine.rendering/ResponseRendererSpec/The_response_preparation_logic_should_properly_render_The__Connection__header_should_be_rendered_correctly/), is green locally.

ktoso:
Please ref issues in comments - otherwise it's not "linked" #1134 :)

jypma:
@ktoso Isn't it enough to ref it in the subject of the commit message? Github showed that as link for me

jlprat:
@jypma I usually write the commit with the issue id twice in the body:
```
Apply some changes #xxx

Issue: #xxx
some further description and so on
```

This links the commit to the issue, and creates a PR with the issue on the title (pretty useless for github) but also the issue will be in the description of the PR (Github likes this, and then links the PR back to the issue).

jypma:
Ah thanks.

jypma:
PLEASE TEST

jypma:
PLS BUILD

jypma:
Green! @akka/akka-http-team anyone up for review?

akka-http-core/src/main/scala/akka/http/scaladsl/model/ContentType.scala
```diff
@@ -53,29 +54,43 @@ sealed trait ContentType extends jm.ContentType with ValueRenderable {
...
   final case class WithCharset(val mediaType: MediaType.WithOpenCharset, val charset: HttpCharset)
     extends jm.ContentType.WithCharset with NonBinary {
 
     private[http] override def render[R <: Rendering](r: R): r.type =
       super.render(r) ~~ ContentType.`; charset=` ~~ charset
   }
 
+  /**
+   * Represents a content-type which we know to contain text, and would be better off having a charset,
+   * but the client hasn't provided that. For example, "text/xml" without a charset parameter.
```
jrudolph: "client or server" or "peer"

akka-http-core/src/main/scala/akka/http/javadsl/model/ContentType.scala
```diff
@@ -5,30 +5,40 @@ package akka.http.javadsl.model
...
- * Represents an Http content-type. A content-type consists of a media-type and an optional charset.
- *
- * See [[ContentTypes]] for convenience access to often used values.
- */
 // Has to be defined in Scala even though it's JavaDSL because of:
 // https://issues.scala-lang.org/browse/SI-9621
 object ContentType {
-
+  /** Represents a content-type which we know not to contain text (will never have have a charset) */
   trait Binary extends ContentType {
```
jrudolph: Could you also remove the extraneous braces for all those traits?

akka-http-core/src/main/scala/akka/http/scaladsl/model/ContentType.scala
jypma: OK

akka-http-core/src/main/scala/akka/http/javadsl/model/ContentType.scala
jypma: yupp

jypma:
@jrudolph where should I add a test? `HttpModelIntegrationSpec.scala`? or would it do just to have a unit test around the changed `ContentTypeHeader.scala`?

jypma:
@jrudolph added test to `HttpHeaderSpec`. And tests green.

I chose to not break binary compat and not introduce `MediaType.withMissingCharset`. This does mean that an actual instance of `ContentType.WithMissingCharset` can only be constructed in the scala DSL, not in Java. Is it worth breaking binary compatibility to create requests we wanted to discourage (i.e. explicit requests with missing charsets)?

manonthegithub:
I've got the service which can not accept Content-Type with charset, and it is unlikely it ever will be fixed. I think this pr will be very useful for usecases like this.

akka-http-core/src/main/scala/akka/http/scaladsl/model/ContentType.scala
```diff
@@ -53,29 +54,43 @@ sealed trait ContentType extends jm.ContentType with ValueRenderable {
...
   }
 
+  /** Represents a content-type which we know to contain text, and has a specified charset. */
   sealed trait NonBinary extends jm.ContentType.NonBinary with ContentType {
     def binary = false
     def charset: HttpCharset
     def charsetOption = Some(charset)
   }
 
+  /** Represents a content-type which we know to contain text, where the charset always has the same predefined value. */
```
jrudolph: :+1: for adding all those extra documentation

jypma:
@jrudolph so is the PR mergeable in its current state? i.e. when clients what to send `Content-Type: application-xml`, we continue to require them to hack stuff like `MediaType.applicationBinary("xml", MediaType.Compressible).toContentType` ? We could also simply add a java API onto the java `ContentType` object, since in scala you can call `ContentType.WithMissingCharset.apply` just fine. 

It would be a bit inconsistent, but wouldn't break binary compatibility.
```
package akka.http.javadsl.model
object ContentType {
  // ...
  def withMissingCharset(mediaType: MediaType.WithOpenCharset): WithMissingCharset
}
```

jrudolph:
I think the functionality itself looks fine. It would be nice have some tests that show how the effect on marshallers / unmarshallers.

jrudolph:
I think the Java constructor belongs into `akka.http.javadsl.model.ContentTypes`. In general, I'm not sure we should make it as simple to create a `ContentType.WithMissingCharset` as the proper thing which has the charset because people will too easily use the wrong one. Any ideas how to solve that?

jypma:
not really... besides the fact that the expression _would_ be (java) `MediaTypes.APPLICATION_XML.toContentTypeWithMissingCharset()` or (scala) ```MediaTypes.`application/xml`.withMissingCharset```, which does point out that you're "missing" something. I'd argue that's enough of an indication that it's not what you'd typically want. Plus the javadocs indicating that.

jrudolph:
Yes, that's right. We also have other constructors in `ContentTypes` for Java and just by case-class instantiation for Scala but I agree if we think of the ones that you show as the normal ones than this is fine.

jypma:
pls build

jypma:
PLS BUILD

jrudolph:
Seems Mima rules are missing.

akka-http-core/src/main/scala/akka/http/javadsl/model/ContentType.scala
```diff
@@ -5,30 +5,35 @@ package akka.http.javadsl.model
...
 
-/**
- * Represents an Http content-type. A content-type consists of a media-type and an optional charset.
- *
- * See [[ContentTypes]] for convenience access to often used values.
- */
 // Has to be defined in Scala even though it's JavaDSL because of:
 // https://issues.scala-lang.org/browse/SI-9621
 object ContentType {
+  /** Represents a content-type which we know not to contain text (will never have have a charset) */
```
lomigmegard: small typo in comment.

jypma:
Added `UnmarshallingSpec` test, and wrote release notes.

docs/src/main/paradox/release-notes.md
```diff
@@ -1,5 +1,28 @@
...
+
+Akka-http has a strongly typed media type / content type system, and knows at compile time about which media types
+are supposed to express a character set attribute, e.g. `text/plain; charset=UTF-8`. Before this release, akka would
+silently assume UTF-8 for `ContentType` instances of media types with a missing `charset` attribute.
+
+From now on, content types missing a charset can be both parsed and expressed directly, using the new 
+`ContentType.WithMissingCharset` trait/class. 
+
+- If you happen to parse an http request with just `text/plain` as its `Content-Type` header, it'll 
+  show up as an instance of this class.
```
ktoso: A bit too casual for release notes, how about:

> For incoming Content-Type headers with values missing a charset, such as `text/plain` the header 
`ContentType` will be represented as `WithMissingCharset`, rather than like previously as a content typed with an assumed UTF-8 charset (which could have been a wrong guess.

docs/src/main/paradox/release-notes.md
```diff
@@ -1,5 +1,28 @@
...
+- If you need to create such a content type programmatically, use e.g. ```MediaTypes.`text/plain`.withMissingCharset```
+  (scala) or `MediaTypes.TEXT_PLAIN.toContentTypeWithMissingCharset()` (java).
+
+Although this change affects binary compatibility because of the extra introduced method, typical usages should be safe,
+as `ContentType` is a sealed trait within akka-http.
```
ktoso: the additional method does not matter, don't mention it - no one should extend the http model, ever.

The actual potentially breaking issue is the "note to scala users", it is binary compatible though.

akka-http-core/src/test/scala/akka/http/impl/model/parser/HttpHeaderSpec.scala
```diff
@@ -208,6 +208,8 @@ class HttpHeaderSpec extends FreeSpec with Matchers {
         `Content-Type`(`application/json`)
       "Content-Type: text/plain; charset=utf8" =!=
         `Content-Type`(ContentType(`text/plain`, `UTF-8`)).renderedTo("text/plain; charset=UTF-8")
+      "Content-Type: text/plain" =!=
+        `Content-Type`(ContentType.WithMissingCharset(MediaTypes.`text/plain`)).renderedTo("text/plain")
```
ktoso: good 👍

jypma:
@ktoso Sure thanks, please do, I won't come back to it until tomorrow.

jypma:
@ktoso I've updated the release notes according to your comments.

jypma:
PLS BUILD

ktoso:
Failure seems unrelated

ktoso:
PLS BUILD

akka-http-tests/src/test/scala/akka/http/scaladsl/unmarshalling/UnmarshallingSpec.scala
```diff
@@ -24,6 +24,9 @@ class UnmarshallingSpec extends FreeSpec with Matchers with BeforeAndAfterAll wi
     "stringUnmarshaller should unmarshal `text/plain` content in UTF-8 to Strings" in {
       Unmarshal(HttpEntity("Hällö")).to[String] should evaluateTo("Hällö")
     }
+    "stringUnmarshaller should assume UTF-8 for textual content type with missing charset" in {
```
jrudolph: :+1:

jypma:
@jrudolph Looks like we may have missed a compiler warning? Somebody on gitter posted this on 10.0.8:
```
scala.MatchError: application/xml (of class akka.http.scaladsl.model.ContentType$WithMissingCharset)
        at akka.http.scaladsl.model.ContentTypeRange.matches(ContentType.scala:17)
        at akka.http.scaladsl.unmarshalling.Unmarshaller$EnhancedFromEntityUnmarshaller$.$anonfun$forContentTypes$4(Unmarshaller.scala:112)
        at akka.http.scaladsl.unmarshalling.Unmarshaller$EnhancedFromEntityUnmarshaller$.$anonfun$forContentTypes$4$adapted(Unmarshaller.scala:112)
...
```
and indeed, [ContentType.scala](https://github.com/akka/akka-http/blob/master/akka-http-core/src/main/scala/akka/http/scaladsl/model/ContentType.scala#L17) wouldn't match `.WithMissingCharset`...

PR #1143, opened 2017-05-17
=pro Remove unused dependencies copied from Akka import

jonas:
When akka-http was migrated from the akka repo the original
project/Dependencies.scala file was copied more or less verbatim. Remove
dependencies that are not used by the akka-http components to reduce
clutter.


jrudolph:
Thanks a lot for cleaning up, @jonas.

PR #1145, opened 2017-05-18
=pro update to latest sbt and sbt-paradox-akka

jrudolph:
To fix the incompatibility with current akka master.


jrudolph:
Failed on travis with artifact retrieval problems but works on our jenkins.

Issue #1136, opened 2017-05-16
Replace confusing === with shouldEqual or similar in docs


manonthegithub:
There is sample code:
```scala
 "return a MethodNotAllowed error for PUT requests to the root path" in {
      // tests:
      Put() ~> Route.seal(smallRoute) ~> check {
        status === StatusCodes.MethodNotAllowed
        responseAs[String] shouldEqual "HTTP method not allowed, supported methods: GET"
      }
    }
```
the line `status === StatusCodes.MethodNotAllowed` actually does no check, but only returns `boolean` value, it is confusing and can lead to errors in tests and code.

jrudolph:
Great, catch. I think those are left overs from previous switches between specs2 and scalatest which differ in behavior in that regard.

Could you create a small PR to fix it?

manonthegithub:
yeah, with pleasure).
I'll try to do it in a week)

manonthegithub:
I was looking on version 10.0.4 of docs, and in current it was fixed. But nevertheless I found another class `FullSpecs2TestKitExampleSpec` where it was.

jonas:
As discussed [here](https://github.com/akka/akka-http/pull/1146#discussion_r117324096), `===` works as expected, but the assertions are different from the other examples, so this ticket is about using the same way of checking in the examples.

nikom1337:
After pr #1750 is merged there will be only one triple equals assertions to fix in `akka.http.impl.engine.ws.Utf8CodingSpecs` test class.

jlprat:
I'd say this one is solved, am I right?
If not please re-open

PR #1146, opened 2017-05-18
Replaced wrong equals assertion in docs #1136

manonthegithub:
Refs #1136

### DIFF
--- a/docs/src/test/scala/docs/http/scaladsl/server/FullSpecs2TestKitExampleSpec.scala
+++ b/docs/src/test/scala/docs/http/scaladsl/server/FullSpecs2TestKitExampleSpec.scala
@@ -32,14 +32,14 @@ class FullSpecs2TestKitExampleSpec extends Specification with Specs2RouteTest {
     "return a greeting for GET requests to the root path" in {
       // tests:
       Get() ~> smallRoute ~> check {
-        responseAs[String] === "Captain on the bridge!"
+        responseAs[String] shouldEqual "Captain on the bridge!"
       }
     }
 
     "return a 'PONG!' response for GET requests to /ping" in {
       // tests:
       Get("/ping") ~> smallRoute ~> check {
-        responseAs[String] === "PONG!"
+        responseAs[String] shouldEqual "PONG!"
       }
     }
 
@@ -53,8 +53,8 @@ class FullSpecs2TestKitExampleSpec extends Specification with Specs2RouteTest {
     "return a MethodNotAllowed error for PUT requests to the root path" in {
       // tests:
       Put() ~> Route.seal(smallRoute) ~> check {
-        status === StatusCodes.MethodNotAllowed
-        responseAs[String] === "HTTP method not allowed, supported methods: GET"
+        status shouldEqual StatusCodes.MethodNotAllowed
+        responseAs[String] shouldEqual "HTTP method not allowed, supported methods: GET"
       }
     }
   }


jonas:
OK TO TEST

docs/src/test/scala/docs/http/scaladsl/server/FullSpecs2TestKitExampleSpec.scala
```diff
@@ -53,8 +53,8 @@ class FullSpecs2TestKitExampleSpec extends Specification with Specs2RouteTest {
     "return a MethodNotAllowed error for PUT requests to the root path" in {
       // tests:
       Put() ~> Route.seal(smallRoute) ~> check {
-        status === StatusCodes.MethodNotAllowed
-        responseAs[String] === "HTTP method not allowed, supported methods: GET"
+        status shouldEqual StatusCodes.MethodNotAllowed
```
jonas: My understanding is that `shouldEqual` compares against `Any` and `===` compares against a value of the same type. So my question is whether it is not better to keep `===` so that the test fails to compile if you compare for example status values with a string?

manonthegithub: yeah, you are right in this case... there was a problem in `class FullTestKitExampleSpec extends WordSpec with Matchers with ScalatestRouteTest`(in 10.0.4) where, it took `===` from `rg.scalactic.TripleEqualsSupport.Equalizer` which just returns `boolean` and doesn't fail the test. 
I was not attentive enough so in case of `FullSpecs2TestKitExampleSpec` it is taken from `org.specs2.matcher.TypedEqual.TypedEqualExpectation` which is right and works fine. 

Nevetheless, I would suggest to use `shouldEqual`, that is not implemented so often as `===` in different libs and won't confuse people who can accidentally use wrong expression in tests. For example, if they would just copy this check in their existing code that implements `WordSpec with Matchers with ScalatestRouteTest` it will actually check nothing.

ktoso:
(CLA is signed)

docs/src/test/scala/docs/http/scaladsl/server/FullSpecs2TestKitExampleSpec.scala
jrudolph: How about using `should_===` everywhere?

jrudolph: Hmm in scalatest that would be `should ===()`

manonthegithub: I used shouldEqual because it is in other samples, for example in FullTestKitExampleSpec, so it should be replaced there and in other places too. What do you think is better?

jrudolph: Maybe let's keep it consistent by keeping `shouldEqual` even if we lose a bit of typesafety here right now. Could make sense to follow up on *all* example using the typesafe variant in the future. WDYT, @jonas?

jonas: Yes, that works. I don't have a strong opinion and mainly wanted to question the choice of wording in #1136 and the commit message in that the current assertions are not wrong and actually works as expected.

ktoso:
Sounds like a plan, thanks! :)

PR #1148, opened 2017-05-19
+htt Support separation of route execution from checking in the Java DSL

jonas:
Port the ScalatestRouteTestSpec to JUnitRouteTestTest and move await of
the route result from RouteTest to TestRouteResult similar to how the
Scala RouteTest works. This breaks binary compatibility of the testkit,
however, no guarantees are given for that part of the Akka HTTP API.

Preliminary work needed to implement the Java API for #39 (integration
route testing).

---

References #1125 (WIP fix for #39)


akka-http-testkit/src/main/scala/akka/http/javadsl/testkit/RouteTest.scala
```diff
@@ -76,5 +77,7 @@ abstract class RouteTest extends AllDirectives with WSTestRequestBuilding {
       def run(request: HttpRequest): TestRouteResult = runRoute(underlying, request)
     }
 
-  protected def createTestRouteResult(request: HttpRequest, result: RouteResult): TestRouteResult
+  protected def createTestRouteResult(request: HttpRequest, result: RouteResult): TestRouteResult =
```
jonas: Keeping `createTestRouteResult` allows to stay more compatible, but this is a low-level API so might not make sense.

akka-http-testkit/src/test/java/akka/http/javadsl/testkit/JUnitRouteTestTest.java
```diff
@@ -0,0 +1,97 @@
...
+          )
+        )
+      );
+
+    route.run(HttpRequest.POST("/abc").withEntity("content"))
+      /* XXX: Expected?
+      .assertRejections(
+        Rejections.method(HttpMethods.GET),
+        Rejections.method(HttpMethods.PUT)
+      )
```
jonas: For some reason the Scala `RouteTest` execution triggers the rejection but the Java `RouteTest` does not. Not sure if this is a bug or a known limitation.

raboof: Indeed the javadsl `RouteTest` does `runRoute` rather than `runRouteUnSealed`. It seems a lot more consistent to use `runRouteUnSealed` there, but that'd mean:
* updating a lot of tests that currently rely on tests being sealed
* providing a nice way to seal a route in the javadsl

jonas: Ah, thanks for taking a look @raboof and for clearing that up. If this is something we want to fix I can address it in this PR. As for API, adding a `runUnsealed` method to `TestRoute` seems more flexible. Let me know what you think.

jrudolph: I think both kinds of testing are reasonable. Testing unsealed makes sense for testing reusable route components, while testing sealed makes sense for testing complete routes from the outside.

Hmm, I think we should try to keep the styles consistent between Scala and Java or make it explicit at least what you want to do. So, how about deprecating `run` in favor of `runSealed` and introducing another `runUnsealed` to introduce the Scala-like variant?

raboof: That does seem a bit verbose - for test cases where the request isn't rejected it doesn't matter if you're running sealed or unsealed, and it'd be kind of a shame to make that so prominent there.

Ideally I think `run` would run unsealed (consistent with scaladsl) and we'd add a `runSealed` to run sealed. I'm not sure how much of our users already use this DSL though, so perhaps that ship has sailed - might be OK to just add `runUnsealed` next to `run` then though?

jonas:
Rebased and updated.

To follow up on https://github.com/akka/akka-http/pull/1148#discussion_r125524384, the sealed/unsealed notion is a bit moot in this context since test routes are always "semi-sealed" in that a default exception handler is provided. I've added a `runWithRejections` method, which I think more clearly communicates what can be expected.

akka-http-testkit/src/test/java/akka/http/javadsl/testkit/JUnitRouteTestTest.java
```diff
@@ -0,0 +1,92 @@
...
+            complete("naah")
+          )
+        )
+      );
+
+    route.runWithRejections(HttpRequest.POST("/abc").withEntity("content"))
+      .assertRejections(
+        Rejections.method(HttpMethods.GET),
+        Rejections.method(HttpMethods.PUT)
+      );
```
raboof: That doesn't looks too bad indeed!

akka-http-testkit/src/main/scala/akka/http/javadsl/testkit/TestRoute.scala
```diff
@@ -17,5 +17,16 @@ import akka.http.javadsl.server.Route
  */
 trait TestRoute {
   def underlying: Route
+
+  /**
+   * Run the request against the sealed route, meaning use a default handler
```
jrudolph: Here's a slightly more comprehensive description:

```
Run the request against the sealed route, meaning that exceptions and rejections will be handled by the default exception and rejection handlers. The default handlers will convert exceptions and rejections into HTTP responses with corresponding status codes (like 404 when no route matches the path or 500 in cases of exceptions).

If you want to assert on the original rejections instead, use `runWithRejections`.
```

akka-http-testkit/src/main/scala/akka/http/javadsl/testkit/TestRoute.scala
```diff
@@ -17,5 +17,16 @@ import akka.http.javadsl.server.Route
...
+   * as a fallback for all cases the route doesn’t handle itself.
+   */
   def run(request: HttpRequest): TestRouteResult
-}
\ No newline at end of file
+
+  /**
+   * Run the request against the semi-sealed route, meaning that a default
+   * handler is used for exceptions but not for rejections.
```
jrudolph: A longer form of that documentation:

```
Run the request against the "semi-sealed" route, meaning that exceptions will be handled by the default exception handler but rejections will not be handled. This means that the test will not see HTTP responses with error status codes for routes that rejected a request. Instead, the `TestRouteResult` allows access to the original rejection containing all the rejection details in structured form. Use `assertRejections` to check that a route rejected a request with expected rejections.

Otherwise, to assert on the actual error HTTP response generated by the default rejection handler, use the `run` method, instead.
```

jonas:
Thanks @jrudolph for the improved and detailed documentation. Integrated to also link in the Javadoc version and rebased the branch.

screen shot 2017-08-30 at 3 59 10 pm

jonas:
PLS BUILD

jrudolph:
Thanks a lot, @jonas!

jonas:
Thanks, will try to find some time to update the docs now that #1404 is merged.

Issue #1160, opened 2017-05-25
Http2: RemoteAddress setting not working for HTTP/1 behind ALPNSwitch


jrudolph:
The reason is that the settings are not passed to the HTTP1 `serverLayer` here: https://github.com/akka/akka-http/blob/2e0d2b5b219c21ce3acd1ced0cb2a2e7cd6aa1ee/akka-http2-support/src/main/scala/akka/http/scaladsl/Http2.scala#L79-L79

ktoso:
Hmm... Have you reproduced the issue or just looked at it?
I mean: 

```
class Http2Ext(private val config: Config)(implicit val system: ActorSystem)
```

so:

```
      AlpnSwitch(getChosenProtocol, Http().serverLayer(), http2Layer()) atop
```

is actually:

```
      AlpnSwitch(getChosenProtocol, Http()(system).serverLayer(), http2Layer()) atop
```

so it uses the right actor system, since the Http1 extension got the system, and the only place it could get it from is that implicit. So I don't think that was the issue.

Issue #1159, opened 2017-05-25
Configuring HTTP2 support is hard because of missing `withXXX` methods and missing overloads for scaladsl


jrudolph:
This should compile:

```scala
import akka.http.scaladsl.settings.ServerSettings
  import akka.http.scaladsl.settings.PreviewServerSettings
  val serverSettings: ServerSettings =
    ServerSettings(system)
      .withPreviewServerSettings(PreviewServerSettings(system).withEnableHttp2(true))
      .withRemoteAddressHeader(true)
```

Also following code fails at runtime because companion object is created with wrong config path:

```scala
PreviewServerSettings(system)
```

ktoso:
Oh, thanks! This may have been what greg bumped into perhaps hm...
I'll take a stab at it then.

PR #1161, opened 2017-05-25
+doc security page about recent vunerability

ktoso:
Resolves https://github.com/akka/akka-http/issues/1159
Resolves https://github.com/akka/akka-http/issues/1160 (though I don't think it was the bug)


ktoso:
Wrong branch..

Issue #1061, opened 2017-04-26
FAILED: akka.http.javadsl.server.HttpAppSpec


jrudolph:
https://jenkins.akka.io:8498/job/pr-validator-akka-http/995/testReport/junit/akka.http.javadsl.server/HttpAppSpec/HttpApp_Java_should_provide_binding_if_available/

"akka.http.javadsl.server.HttpAppSpec.HttpApp Java should provide binding if available"

`The future returned an exception of type: akka.stream.StreamTcpException, with message: Tcp command [Connect(localhost:36181,None,List(),Some(10 seconds),true)] failed.`

jrudolph:
@jlprat do you have an idea what this could be?

jlprat:
I'll take a look, but it smells like race condition trying to bind to whatever port it was meant to bind

jlprat:
So, it fails on the last line of these:
```
      minimal.bindingPromise.get(5, TimeUnit.SECONDS)
      minimal.binding().localAddress.getPort should ===(port)
      minimal.binding().localAddress.getHostName should ===(host)
      // Requesting the server to shutdown
      callAndVerify(host, port, "shutdown")
```
which is:
```
  private def callAndVerify(host: String, port: Int, path: String) = {
    implicit val mat = ActorMaterializer()
    val request = HttpRequest(uri = s"http://$host:$port/$path")
    val response = Http().singleRequest(request)
    response.futureValue.status should ===(StatusCodes.OK)
  }
```
So the binding completed correctly, but for whatever reason this same server wasn't listening. Something is not adding up, as if the promise is completed, the server should be up and running. (promise is completed when the `Future` of `bindAndHandle` is completed.

Will try to find some time to investigate further

jrudolph:
I added https://github.com/akka/akka/pull/22954 which should give us better error messages in the future.

jrudolph:
I also worked a bit on improving `TestUtils` to use all possible loopback interfaces to make random port conflicts less likely. See https://github.com/jrudolph/akka-http/tree/jr/w/choose-random-loopback-interface

Though, after going through all those "Tcp command Connect failed" tickets I'm actually not convinced that port conflicts are actually the reason. Also it doesn't seem to match up with what you reported about how it works here, @jlprat. So, maybe let's wait until the error messages have been improved to see what the cause actually is.

jlprat:
Let's wait till next failure and hopefully we get some more info :+1:

jrudolph:
[...]

PR #1162, opened 2017-05-25
=ht2,htc #1160 #1159 fix preview settings

### LINKED ISSUE #1159: Configuring HTTP2 support is hard because of missing `withXXX` methods and missing overloads for scaladsl

### LINKED ISSUE #1160: Http2: RemoteAddress setting not working for HTTP/1 behind ALPNSwitch

ktoso:
Resolves #1159
Resolves #1160 ~(though I don't think it was the bug)~


akka-http2-support/src/main/scala/akka/http/scaladsl/Http2.scala
```diff
@@ -76,7 +76,7 @@ class Http2Ext(private val config: Config)(implicit val system: ActorSystem)
       }
       val tls = TLS(createEngine, _ ⇒ Success(()), IgnoreComplete)
 
-      AlpnSwitch(getChosenProtocol, Http().serverLayer(), http2Layer()) atop
+      AlpnSwitch(getChosenProtocol, http.serverLayer(), http2Layer()) atop
```
2m: Wasn't the @jrudolph comment here https://github.com/akka/akka-http/issues/1160#issue-231293957 about passing settings to `serverLayer` like so:

```scala
serverLayer(settings = settings)
```

ktoso:
Failure was unrelated and known: https://github.com/akka/akka-http/issues/1061

akka-http2-support/src/main/scala/akka/http/scaladsl/Http2.scala
jrudolph: Yep, thanks for the extra eyes @2m.

PR #1165, opened 2017-05-25
Link to javadoc and scaladoc instead of files on GitHub

jonas:
Also fixes a broken snippet: http://doc.akka.io/docs/akka-http/10.0.7/scala/http/routing-dsl/directives/coding-directives/decodeRequest.html#decoderequest

### DIFF
--- a/docs/src/main/paradox/scala/http/routing-dsl/rejections.md
+++ b/docs/src/main/paradox/scala/http/routing-dsl/rejections.md
@@ -23,7 +23,7 @@ and handle any rejection.
 ## Predefined Rejections
 
 A rejection encapsulates a specific reason why a route was not able to handle a request. It is modeled as an object of
-type `Rejection`. Akka HTTP comes with a set of @github[predefined rejections](/akka-http/src/main/scala/akka/http/scaladsl/server/Rejection.scala), which are used by the many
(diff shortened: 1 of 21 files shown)


docs/src/main/paradox/scala/http/routing-dsl/rejections.md
```diff
@@ -33,7 +33,7 @@ the @ref[handleRejections](directives/execution-directives/handleRejections.md#h
 ## The RejectionHandler
 
 The @ref[handleRejections](directives/execution-directives/handleRejections.md#handlerejections) directive delegates the actual job of converting a list of rejections to its argument, a
-@github[RejectionHandler](/akka-http/src/main/scala/akka/http/scaladsl/server/RejectionHandler.scala), which is defined like this:
```
ktoso: Hmm... Links to sources have been kind of historically spray style and perhaps not the best idea for us in furniture esp with java docs etc. So I think I'm in favour of this change of linking to docs instead, thanks :)

jonas: Linking to javadocs/scaladocs is more friendly than dropping readers into the source code and also allows us to link to a specific method instead of a line number which may change.

ktoso: Yeah, I agree - LGTM from my side 👍

Issue #1142, opened 2017-05-17
FAILED: MessageSpec


ktoso:
Timeout in `akka.http.impl.engine.ws.MessageSpec.The WebSocket implementation should collect messages from frames apply backpressure to the network if a message isn't read by the user` 

https://jenkins.akka.io:8498/job/pr-validator-akka-http/1137/testReport/junit/akka.http.impl.engine.ws/MessageSpec/The_WebSocket_implementation_should_collect_messages_from_frames_apply_backpressure_to_the_network_if_a_message_isn_t_read_by_the_user/

jrudolph:
https://jenkins.akka.io:8498/job/pr-validator-akka-http/1433/testReport/junit/akka.http.impl.engine.ws/MessageSpec/The_WebSocket_implementation_should_collect_messages_from_frames_apply_backpressure_to_the_network_if_a_message_isn_t_read_by_the_user/

jrudolph:
https://jenkins.akka.io:8498/job/pr-validator-akka-http/1180/testReport/junit/akka.http.impl.engine.ws/MessageSpec/The_WebSocket_implementation_should_collect_messages_from_frames_apply_backpressure_to_the_network_if_a_message_isn_t_read_by_the_user/

raboof:
"Attempted 16 times over 4.927106658 seconds" is a bit weird given the check is:
```
        eventually(PatienceConfiguration.Timeout(500.millis.dilated), PatienceConfiguration.Interval(1.milli.dilated)) {
          the[AssertionError] thrownBy pushInput(ByteString("a"))
        }
```

jrudolph:
https://jenkins.akka.io:8498/job/pr-validator-akka-http/1543/testReport/junit/akka.http.impl.engine.ws/MessageSpec/The_WebSocket_implementation_should_collect_messages_from_frames_apply_backpressure_to_the_network_if_a_message_isn_t_read_by_the_user/

johanandren:
The last one is also weird, `Attempted 1 times over 783.3385999999999 milliseconds.`,  I can see that 783 is roughly 500 dilated with 1.5, but just trying 1 time when interval is set to 1.milli?? Looked in the ScalaTest bug tracker to see if there is something with `eventually` but nothing. Also looked a bit for GC pauses that could explain only trying once within the timeout but nothing there.

raboof:
Not seen anymore since #1777, let's close for now

PR #1175, opened 2017-06-02
Websocket illegal headers #1166 

disblader:
Attempted fix to issue #1166. 

Changed HttpHeaderParser such that by default illegal header value handling depends on the configured IllegalResponseHeaderValueProcessingMode.

A couple of problems:

First, I'm not sure how I should handle an unexpected value [here](https://github.com/disblader/akka-http/blob/websocket-illegal-headers/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpHeaderParser.scala#L452). Is there an exception that implementors generally use for these kinds of situations?

Secondly, I'm pretty bad with scalatest and I ended up in [this situation](https://github.com/disblader/akka-http/blob/websocket-illegal-headers/akka-http-core/src/test/scala/akka/http/impl/engine/parsing/HttpHeaderParserSpec.scala#L194-L201) where I have identical tests with different setups and I'm not sure if that's fine or not?

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpHeaderParser.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpHeaderParser.scala
@@ -446,7 +446,7 @@ private[http] object HttpHeaderParser {
 
   def defaultIllegalHeaderHandler(settings: HttpHeaderParser.Settings, log: LoggingAdapter): ErrorInfo ⇒ Unit =
     if (settings.illegalHeaderWarnings)
-      info ⇒ logParsingError(info withSummaryPrepended "Illegal response header", log, settings.errorLoggingVerbosity)
+      info ⇒ logParsingError(info withSummaryPrepended "Illegal header", log, settings.errorLoggingVerbosity)
     else
       (_: ErrorInfo) ⇒ _ // Does exactly what the label says - nothing
 
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpHeaderParser.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpHeaderParser.scala
@@ -441,8 +441,8 @@ private[http] object HttpHeaderParser {
     "Cache-Control: no-cache",
     "Expect: 100-continue")
 
-  def apply(settings: HttpHeaderParser.Settings, log: LoggingAdapter)(onIllegalHeader: ErrorInfo ⇒ Unit = defaultIllegalHeaderHandler(settings, log)) =
-    prime(unprimed(settings, log, onIllegalHeader))
+  def apply(settings: HttpHeaderParser.Settings, log: LoggingAdapter) =
+    prime(unprimed(settings, log, defaultIllegalHeaderHandler(settings, log)))
 
   def defaultIllegalHeaderHandler(settings: HttpHeaderParser.Settings, log: LoggingAdapter): ErrorInfo ⇒ Unit =
     if (settings.illegalHeaderWarnings)
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpHeaderParser.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpHeaderParser.scala
@@ -449,7 +449,6 @@ private[http] object HttpHeaderParser {
         case IllegalResponseHeaderValueProcessingMode.Error ⇒ info ⇒ throw IllegalHeaderException(info)
         case IllegalResponseHeaderValueProcessingMode.Warn ⇒ info ⇒ logParsingError(info withSummaryPrepended "Illegal response header", log, settings.errorLoggingVerbosity)
         case IllegalResponseHeaderValueProcessingMode.Ignore ⇒ _ ⇒ // Does exactly what the label says - nothing
-        case _ ⇒ ??? //TODO what would be the appropriate way to handle an unexpected value here?
       }
     }
 
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpHeaderParser.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpHeaderParser.scala
@@ -425,6 +425,7 @@ private[http] object HttpHeaderParser {
     def maxHeaderValueLength: Int
     def headerValueCacheLimit(headerName: String): Int
     def customMediaTypes: MediaTypes.FindCustom
+    def illegalHeaderWarnings: Boolean
     def illegalResponseHeaderValueProcessingMode: IllegalResponseHeaderValueProcessingMode
     def errorLoggingVerbosity: ErrorLoggingVerbosity
   }
@@ -444,13 +445,10 @@ private[http] object HttpHeaderParser {
     prime(unprimed(settings, log, onIllegalHeader))
 
   def defaultIllegalHeaderHandler(settings: HttpHeaderParser.Settings, log: LoggingAdapter): ErrorInfo ⇒ Unit =
-    {
-      settings.illegalResponseHeaderValueProcessingMode match {
-        case IllegalResponseHeaderValueProcessingMode.Error ⇒ info ⇒ throw IllegalHeaderException(info)
-        case IllegalResponseHeaderValueProcessingMode.Warn ⇒ info ⇒ logParsingError(info withSummaryPrepended "Illegal response header", log, settings.errorLoggingVerbosity)
-        case IllegalResponseHeaderValueProcessingMode.Ignore ⇒ _ ⇒ // Does exactly what the label says - nothing
-      }
-    }
+    if (settings.illegalHeaderWarnings)
+      info ⇒ logParsingError(info withSummaryPrepended "Illegal response header", log, settings.errorLoggingVerbosity)
+    else
+      (_: ErrorInfo) ⇒ _ // Does exactly what the label says - nothing
 
   def unprimed(settings: HttpHeaderParser.Settings, log: LoggingAdapter, warnOnIllegalHeader: ErrorInfo ⇒ Unit) =
     new HttpHeaderParser(settings, log, warnOnIllegalHeader)
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpHeaderParser.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpHeaderParser.scala
@@ -11,6 +11,7 @@ import java.lang.{ StringBuilder ⇒ JStringBuilder }
 import akka.annotation.InternalApi
 import akka.event.LoggingAdapter
 import akka.http.scaladsl.settings.ParserSettings.IllegalResponseHeaderValueProcessingMode
+import akka.http.scaladsl.settings.ParserSettings.ErrorLoggingVerbosity
 import akka.http.scaladsl.settings.ParserSettings
 
 import scala.annotation.tailrec
@@ -425,6 +426,7 @@ private[http] object HttpHeaderParser {
     def headerValueCacheLimit(headerName: String): Int
     def customMediaTypes: MediaTypes.FindCustom
     def illegalResponseHeaderValueProcessingMode: IllegalResponseHeaderValueProcessingMode
+    def errorLoggingVerbosity: ErrorLoggingVerbosity
   }
 
   private def predefinedHeaders = Seq(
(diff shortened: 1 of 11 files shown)


raboof:
OK TO TEST

jlprat:
The test is failing because the code didn't satisfy the code formatting. To fix it, just try to compile your code locally, in your case just run:
```
sbt test:compile
```

disblader:
I was wondering why the styling would change out of nowhere, but I kept on changing it back to what I wanted haha.

jlprat:
`HttpHeaderParserSpec` still fails ;)
You can see it here https://jenkins.akka.io:8498/job/pr-validator-akka-http/1179/console

disblader:
The jenkins PR validation seems to be taking a while, is that normal?

jlprat:
No it is not normal. Something went wrong I guess.

jlprat:
PLS BUILD

jrudolph:
Seems our Jenkins box is down right now, we'll try to fix.

disblader:
Doing `test:compile` and then a `git diff` does not yield any differences. Any idea why this is not working?

jlprat:
Now this thing is fine, apparently. Your current problem is shown here: https://jenkins.akka.io:8498/job/pr-validator-akka-http/1180/testReport/junit/akka.http.impl.engine.ws/MessageSpec/The_WebSocket_implementation_should_collect_messages_from_frames_apply_backpressure_to_the_network_if_a_message_isn_t_read_by_the_user/

You can click on the `Details` link next to the "Jenkins PR Validation" task.

jlprat:
In this case is an already known bug that happens occasionally: #1142

PLS BUILD

jlprat:
Failed again, now because of #299

PLS BUILD

joan38:
Should we rerun this again?

disblader:
I'm a little confused as to the whole binary compatibility thing. Why would my code have broken that?

jlprat:
Binary incompatibility is because of the added method `errorLoggingVerbosity`.
In this particular case it is fine as the class is annotated with `InternalApi`.
You can check in CONTRIBUTING how to add such exception.

jlprat:
Could you rebase against master? Current failure is because of this.

Sorry you are having a not so smooth process. 😥

disblader:
No worries at all haha. I find it all a very valuable learning experience.

I've done the merge already, so we'll see how that'll go.

akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpHeaderParser.scala
```diff
@@ -438,9 +440,19 @@ private[http] object HttpHeaderParser {
...
+  def apply(settings: HttpHeaderParser.Settings, log: LoggingAdapter)(onIllegalHeader: ErrorInfo ⇒ Unit = defaultIllegalHeaderHandler(settings, log)) =
     prime(unprimed(settings, log, onIllegalHeader))
 
+  def defaultIllegalHeaderHandler(settings: HttpHeaderParser.Settings, log: LoggingAdapter): ErrorInfo ⇒ Unit =
+    {
+      settings.illegalResponseHeaderValueProcessingMode match {
+        case IllegalResponseHeaderValueProcessingMode.Error ⇒ info ⇒ throw IllegalHeaderException(info)
+        case IllegalResponseHeaderValueProcessingMode.Warn ⇒ info ⇒ logParsingError(info withSummaryPrepended "Illegal response header", log, settings.errorLoggingVerbosity)
+        case IllegalResponseHeaderValueProcessingMode.Ignore ⇒ _ ⇒ // Does exactly what the label says - nothing
+        case _ ⇒ ??? //TODO what would be the appropriate way to handle an unexpected value here?
```
raboof: As the `apply` for IllegalResponseHeaderValueProcessingMode already throw `IllegalArgumentException` for anything else, perhaps it'd be fine to leave out this case (and deal with the `MatchException` if something else is found after all)?

jlprat: Maybe I'm missreading it (I'm on my mobile phone) but isn't `IllegalResponseHeaderValueProcessingMode` sealed and only with `Warn`, `Error` and `Ignore`?

disblader: It is, however I am wondering if there is anything that would normally be done to make sure that something like an introduction of a new mode would not break this kind of logic?

jlprat: Then in my opinion, the last case can be removed.
If introducing a new mode ever happens, that PR should be in charge of modifying this function if needed.

disblader: Done and committed.

akka-http-core/src/test/scala/akka/http/impl/engine/parsing/HttpHeaderParserSpec.scala
```diff
@@ -186,6 +187,19 @@ abstract class HttpHeaderParserSpec(mode: String, newLine: String) extends WordS
...
+
+    "continue parsing even if a header contains an invalid value when illegalResponseHeaderValueProcessingMode is set to Warn" in new TestSetup(testSetupMode = TestSetupMode.Default, createParserSettings(system, illegalResponseHeaderValueProcessingMode = IllegalResponseHeaderValueProcessingMode.Warn)) {
+      noException should be thrownBy parseLine(s"Server: something; something${newLine}x")
+    }
+
+    // TODO this test is almost identical to the test above, but has a different setup - what do?
+    "continue parsing even if a header contains an invalid value when illegalResponseHeaderValueProcessingMode is set to Ignore" in new TestSetup(testSetupMode = TestSetupMode.Default, createParserSettings(system, illegalResponseHeaderValueProcessingMode = IllegalResponseHeaderValueProcessingMode.Ignore)) {
+      noException should be thrownBy parseLine(s"Server: something; something${newLine}x")
+    }
+
```
disblader: Does this need to be changed or is it good enough as is?

jlprat: Instead of writing something twice, you can write something like `s"header with illegal chars $newline !"`
However, I think I'm nitpicking here.

disblader: What about the tests where they're identical other than the test setup? Is there a way to perform the same test with two different setups?

jlprat: you can extract the test code itself to a function and call it in both tests. But we are talking about 1 line here, IMO, in this specific situation is fine. If we would have the same kind of test several times, I would say it would be worth extracting it.

jlprat:
PLS BUILD

Jenkins timed out

disblader:
Oh no, I just wanted to remove a todo I had but now I'm breaking jenkins again 😢

akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpHeaderParser.scala
```diff
@@ -438,9 +440,18 @@ private[http] object HttpHeaderParser {
     "Cache-Control: no-cache",
     "Expect: 100-continue")
 
-  def apply(settings: HttpHeaderParser.Settings, log: LoggingAdapter)(onIllegalHeader: ErrorInfo ⇒ Unit = info ⇒ throw IllegalHeaderException(info)) =
+  def apply(settings: HttpHeaderParser.Settings, log: LoggingAdapter)(onIllegalHeader: ErrorInfo ⇒ Unit = defaultIllegalHeaderHandler(settings, log)) =
     prime(unprimed(settings, log, onIllegalHeader))
 
+  def defaultIllegalHeaderHandler(settings: HttpHeaderParser.Settings, log: LoggingAdapter): ErrorInfo ⇒ Unit =
+    {
```
jrudolph: Braces probably not needed

akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpHeaderParser.scala
```diff
@@ -438,9 +440,18 @@ private[http] object HttpHeaderParser {
     "Cache-Control: no-cache",
     "Expect: 100-continue")
 
-  def apply(settings: HttpHeaderParser.Settings, log: LoggingAdapter)(onIllegalHeader: ErrorInfo ⇒ Unit = info ⇒ throw IllegalHeaderException(info)) =
+  def apply(settings: HttpHeaderParser.Settings, log: LoggingAdapter)(onIllegalHeader: ErrorInfo ⇒ Unit = defaultIllegalHeaderHandler(settings, log)) =
```
jrudolph: What about the usage in `HttpServerBluePrint`? Shouldn't we also use the default one there?

disblader: I went over the usages of `HttpHeaderParser.apply()`  and there's one of two things that happens every time with regards to the illegal header handling:
- the call defers to the default value ([`RequestParserSpec 305`](https://github.com/akka/akka-http/blob/master/akka-http-core/src/test/scala/akka/http/impl/engine/parsing/RequestParserSpec.scala#L305), [`RequestParserSpec 587`](https://github.com/akka/akka-http/blob/master/akka-http-core/src/test/scala/akka/http/impl/engine/parsing/RequestParserSpec.scala#L587), [`WebSocketClientBlueprint`](https://github.com/akka/akka-http/blob/master/akka-http-core/src/main/scala/akka/http/impl/engine/ws/WebSocketClientBlueprint.scala#L65), [`ResponseParserSpec`](https://github.com/akka/akka-http/blob/master/akka-http-core/src/test/scala/akka/http/impl/engine/parsing/ResponseParserSpec.scala#L326), and `HttpHeaderParserSpec` where I just added a usage)
- a variation of the same function is passed, which looks at the parser settings and either warns about the illegal header or not and then moves on ([`OutgoingConnectionBlueprint`](https://github.com/akka/akka-http/blob/master/akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala#L89-L92), [`BodyPartParser`](https://github.com/akka/akka-http/blob/master/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/BodyPartParser.scala#L62-L64), [`HttpServerBluePrint`](https://github.com/akka/akka-http/blob/master/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala#L212-L215), [`Http2Blueprint`](https://github.com/akka/akka-http/blob/master/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala#L82-L85))

Which kinda begs the question of why the option to pass that function in exists at all. In my opinion either all of the usages currently passing in a function should defer to the default, or the option to pass a custom function for illegal header handling should be removed altogether. Thoughts?

jrudolph: I agree that we can remove the custom function if everything is covered by the settings. Would you like to try if that is possible?

disblader: Sure. Hopefully it will all just work haha.

disblader: This actually seems to have been a small can of worms. 

The entire fix that I applied was based on my misunderstanding of `illegal-response-header-value-processing-mode` parameter. I made the fix thinking that it is supposed to determine how an already raised error is to be handled, but looking more closely it is used to determine whether an error is to be raised at all. And so rather than what I did, the correct default should have been to either log the raised error and continue, or simply continue. This is what all of the usages of the parser do already individually, except `WebSocketClientBlueprint`, which is why the initial fault happened.

Anyway, I hope all of that makes sense. I'll go apply the changes now.

jrudolph: I see, that sounds even simpler, so I'm looking forward to your solution.

disblader:
It seems to have failed due to a timeout. Could someone issue a rebuild?

jlprat:
I think it failed due to MiMa:
`[error] (akka-http-core/*:mimaReportBinaryIssues) akka-http-core: Binary compatibility check failed!`
You can run the task locally.

disblader:
Heh, that's what I get for looking and console output on my mobile. Fixed that now.

Also, thanks to all of you for sticking with me 👍 . Turned out to be rather laborious for sucha a small fix haha.

jlprat:
You hit some regressions, that's unfortunate, but in general dealing with MiMa is not easy in the beginning. Thanks for persisting :+1:

disblader:
It's been a few days since this pull request progressed. Is there anything I can do to move this forward or do I just need to wait for someone from the akka team to look at this?

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
```diff
@@ -79,10 +79,7 @@ object Http2Blueprint {
     // This is master header parser, every other usage should do .createShallowCopy()
     // HttpHeaderParser is not thread safe and should not be called concurrently,
     // the internal trie, however, has built-in protection and will do copy-on-write
-    val masterHttpHeaderParser = HttpHeaderParser(parserSettings, log) { info ⇒
-      if (parserSettings.illegalHeaderWarnings)
-        logParsingError(info withSummaryPrepended "Illegal request header", log, parserSettings.errorLoggingVerbosity)
-    }
+    val masterHttpHeaderParser = HttpHeaderParser(parserSettings, log)
```
2m: After the change the possibly logged parsing error will be `Illegal request header` and it was `Illegal response header`. If it is not possible to differentiate in the `defaultIllegalHeaderHandler` then we should just leave `illegal header` here.

2m:
@disblader I was just going through this one last time and noticed the changed log messages. After those are addressed this will be good to go in.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
disblader: I could've sworn I changed that "illegal response header" message to "illegal header", but there you go. Will fix in a second. 

As far as differentiating between response and request header warnings, it's possible to do so by providing a custom message as an argument to log in case of illegal header, but I figured that introducing a parameter just for that was a little overkill as this would mostly be obvious from the origin of the warning.

2m:
Rebase needed as https://github.com/akka/akka-http/pull/1200 has been merged in the meantime.

disblader:
All fixed now 👍

PR #1182, opened 2017-06-08
Add sbt-whitesource

raboof:


ktoso:
Seems like this should work now - the second travis hook worked now.
LGTM

raboof:
Hmm this means it's triggering too early though: we'd want to upload to whitesource only after merging to master, not before. Will try to fix.

raboof:
PLS BUILD

.travis.yml
```diff
@@ -24,9 +24,8 @@ cache:
 
 script:
   # need to override as the default is to test
-  - sbt 'set credentials += Credentials("whitesource", "whitesourcesoftware.com", "", System.getenv("WHITESOURCE_KEY"))'
   - sbt -J-XX:ReservedCodeCacheSize=256m +update
-  - if [[ "$TRAVIS_BRANCH" == "master" ]]; then sbt whitesourceCheckPolicies; fi
+  - if [[ "$TRAVIS_BRANCH" == "master" && "$TRAVIS_EVENT_TYPE" == "push" ]]; then sbt 'set credentials += Credentials("whitesource", "whitesourcesoftware.com", "", System.getenv("WHITESOURCE_KEY"))' whitesourceCheckPolicies whitesourceUpdate; else sbt 'set credentials += Credentials("whitesource", "whitesourcesoftware.com", "", System.getenv("WHITESOURCE_KEY"))' whitesourceCheckPolicies; fi
```
ktoso: I guess I'm missing something, can we go over the process at some point? Only on push to master sounds a bit weird - does merging to master in github count then? If yes, then I get it - so we only update with things that we actually merged, and not random things that people PR right? Or did I misinterpret this?

raboof: That's the way I interpret it, too: valid values for `TRAVIS_EVENT_TYPE` are "push, pull_request, api, cron" (https://docs.travis-ci.com/user/environment-variables/), so I assumed this would be 'push' when building after merging. We'll see for sure after merging :).

ktoso: Ok :) Please check if it actually worked then 👍

raboof:
:tada: 
```
[info] Updated Projects:
[info] * akka-http-master
```

ktoso:
Awesome, thanks :)

build.sbt
```diff
@@ -176,3 +180,9 @@ lazy val docs = project("docs")
     additionalTasks in ValidatePR += paradox in Compile,
     deployRsyncArtifact := List((paradox in Compile).value -> s"www/docs/akka-http/${version.value}")
   )
+
+def hasCommitsAfterTag(description: Option[GitDescribeOutput]): Boolean = {
+  val result = description.get.commitSuffix.distance > 0
+  println(s"returning $result")
```
jonas: @raboof Can this be removed?

raboof: oops

Issue #541, opened 2016-11-17
formField fails if custom unmarshaller is in scope


marekzebrowski:
`formField` directive fails if there is too generic unmarshaller in scope, that prevents `FieldMagnet` machinery to work.
It can happen quite a lot - for example when using akka-http-json4s or other generic unmarshaller. 

Reproducer https://github.com/marekzebrowski/akka-http-form-field-bug

If it is not solvable, maybe a word of warning in documentation, or a guide how to write own unmarshaller would be helpful

ktoso:
Yeah sounds like a docs thing, thanks for reporting.
Up for grabs if someone has the time! :)

marekzebrowski:
I wonder if it is a bug. Defined marshaller should work only for `application/json` content-type. Request in test-case is `application/x-www-form-urlencoded`. If it is not a bug, it is at least confusing

silvaren:
Not sure sure if this is the same issue, but it looks very similar. When using circe + akka-http-json support for JSON (which I believe provides generic unmarshallers to the scope) as in:

```
  private def route(implicit mat: Materializer) = {
    import Directives._
    import de.heikoseeberger.akkahttpcirce.FailFastCirceSupport._
    import io.circe.generic.auto._

    pathSingleSlash {
      post {
        entity(as[Foo]) { foo =>
          complete {
            foo
          }
        }
      }
    } ~
      path("formurlencoded") {
        post {
          formFields('coolKey) { coolKey =>
            complete(s"The cool value is '${coolKey}'")
          }
        }
      }
  }
```
If I run and test the second route:
`curl -d "coolKey=coolValue" -X POST http://localhost:8000/formurlencoded`

I get the following error:
`The request's Content-Type is not supported. Expected: application/json or multipart/form-data`

But if I remove the JSON support it works correctly. [Full issue on akka-http-json](https://github.com/hseeberger/akka-http-json/issues/148), but it might be actually an Akka HTTP issue.

jrudolph:
Yes, it's an anti-pattern to bring generic implicits into scope that are able to marshal/unmarshal everything. When you think about it, it is somewhat obvious that this cannot work. So, the recommendation is to never provide something like

```
def unmarshaller[T]: FromEntityUnmarshaller[T]
```

where `T` is basically unconstrained.

jrudolph:
[...]

PR #1188, opened 2017-06-12
Simplify formField implicits a bit #541

jrudolph:
We now don't allow customizing what and how a form is interpreted any more (which no one should have customized so far anyway).

On the other hand, implicit resolution is now simpler and less error-prone and probably also faster for most `formField` usages.

Old implicit definitions are kept for now for binary compatibility.

### DIFF
--- a/akka-http/src/main/scala/akka/http/scaladsl/server/directives/FormFieldDirectives.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/server/directives/FormFieldDirectives.scala
@@ -5,6 +5,7 @@
 package akka.http.scaladsl.server
 package directives
 
+import akka.annotation.DoNotInherit
 import akka.http.impl.util._
 import akka.http.scaladsl.common._
 import akka.http.scaladsl.model.EntityStreamSizeException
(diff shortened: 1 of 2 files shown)


akka-http/src/main/scala/akka/http/scaladsl/server/directives/FormFieldDirectives.scala
```diff
@@ -135,17 +136,101 @@ object FormFieldDirectives extends FormFieldDirectives {
...
     import RouteDirectives._
     import akka.http.scaladsl.unmarshalling.{ FromStrictFormFieldUnmarshaller ⇒ FSFFU, _ }
+
     type SFU = FromEntityUnmarshaller[StrictForm]
     type FSFFOU[T] = Unmarshaller[Option[StrictForm.Field], T]
 
-    private def extractField[A, B](f: A ⇒ Directive1[B]): FieldDefAux[A, Directive1[B]] = fieldDef(f)
-    private def handleFieldResult[T](fieldName: String, result: Future[T]): Directive1[T] = onComplete(result).flatMap {
+    protected def extractField[A, B](f: A ⇒ Directive1[B]): FieldDefAux[A, Directive1[B]] = fieldDef(f)
+    protected def handleFieldResult[T](fieldName: String, result: Future[T]): Directive1[T] = onComplete(result).flatMap {
```
raboof: Maybe make them `protected final` to make it clear they can be used but aren't overridden?

jrudolph: Yep, good idea.

jrudolph: This is an `object`. so it's not possible to override them anyway.

PR #1190, opened 2017-06-12
Allow '=' in query param values in relaxed mode (Fix #1120)

### LINKED ISSUE #1120: "=" in query parameter values
When a query parameter value contains a "=", akka-http throws the following exception:
> akka.http.scaladsl.model.IllegalUriException: Illegal query: Invalid input '=', expected part, '&' or 'EOI' ...

It doesn't matter whether you're in [strict or relaxed mode](http://doc.akka.io/docs/akka-http/10.0.4/scala/http/common/uri-model.html#strict-and-relaxed-mode), the equals sign is not allowed.

When looking at [RFC3986 page 23](https://tools.ietf.org/html/rfc3986#page-23) however, query is defined as:

> The query component contains non-hierarchical data that, along with
data in the path component (Section 3.3), serves to identify a
resource within the scope of the URI's scheme and naming authority
(if any).  The query component is indicated by the first question
mark ("?") character and terminated by a number sign ("#") character
or by the end of the URI.
> 
> query       = *( pchar / "/" / "?" )
> 
> The characters slash ("/") and question mark ("?") may represent data
within the query component.  Beware that some older, erroneous
implementations may not handle such data correctly when it is used as
the base URI for relative references (Section 5.1), apparently
because they fail to distinguish query data from path data when
looking for hierarchical separators.  However, as query components
are often used to carry identifying information in the form of
"key=value" pairs and one frequently used value is a reference to
another URI, it is sometimes better for usability to avoid percent-
encoding those characters.

Where `pchar` is defined as:
> pchar = unreserved / pct-encoded / sub-delims / ":" / "@"

And its types as:
> unreserved = ALPHA / DIGIT / "-" / "." / "_" / "~"
> pct-encoded = "%" HEXDIG HEXDIG
> sub-delims = "!" / "$" / "&" / "'" / "(" / ")" / "*" / "+" / "," / ";" / "="

According to this definition the equals sign should be allowed. Am I missing something, or is this a bug in akka-http?

True, I wouldn't use "=" in values myself, but Angular2 for example doesn't escape it as well, hence requiring custom work otherwise.

gmethvin:
* Create new character class for query value chars and use this in relaxed mode
* Update relevant tests and examples to reflect this change

Fixes #1120

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/model/parser/CharacterClasses.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/model/parser/CharacterClasses.scala
@@ -43,12 +43,14 @@ private[http] object CharacterClasses {
   val `pchar-base-nc` = unreserved ++ `sub-delims` ++ '@'
   val `pchar-base` = `pchar-base-nc` ++ ':' // pchar without percent
   val `query-fragment-char` = `pchar-base` ++ "/?"
-  val `strict-query-char` = `query-fragment-char` -- "&=;"
-  val `strict-query-char-np` = `strict-query-char` -- '+'
+  val `strict-query-key-char` = `query-fragment-char` -- "&=;"
+  val `strict-query-value-char` = `query-fragment-char` -- "&=;"
+  val `strict-query-char-np` = `strict-query-value-char` -- '+'
 
   val `relaxed-fragment-char` = VCHAR -- '%'
   val `relaxed-path-segment-char` = VCHAR -- "%/?#"
-  val `relaxed-query-char` = VCHAR -- "%&=#"
+  val `relaxed-query-key-char` = VCHAR -- "%&=#"
+  val `relaxed-query-value-char` = VCHAR -- "%&#"
   val `raw-query-char` = VCHAR -- '#'
   val `scheme-char` = ALPHA ++ DIGIT ++ '+' ++ '-' ++ '.'
 
--- a/akka-http-core/src/main/scala/akka/http/impl/model/parser/UriParser.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/model/parser/UriParser.scala
@@ -84,9 +84,13 @@ private[http] final class UriParser(val input: ParserInput,
     case Uri.ParsingMode.Strict ⇒ `pchar-base`
     case _                      ⇒ `relaxed-path-segment-char`
   }
-  private[this] val `query-char` = uriParsingMode match {
-    case Uri.ParsingMode.Strict              ⇒ `strict-query-char`
-    case Uri.ParsingMode.Relaxed             ⇒ `relaxed-query-char`
+  private[this] val `query-key-char` = uriParsingMode match {
+    case Uri.ParsingMode.Strict              ⇒ `strict-query-key-char`
+    case Uri.ParsingMode.Relaxed             ⇒ `relaxed-query-key-char`
+  }
+  private[this] val `query-value-char` = uriParsingMode match {
+    case Uri.ParsingMode.Strict              ⇒ `strict-query-value-char`
+    case Uri.ParsingMode.Relaxed             ⇒ `relaxed-query-value-char`
   }
   private[this] val `fragment-char` = uriParsingMode match {
     case Uri.ParsingMode.Strict ⇒ `query-fragment-char`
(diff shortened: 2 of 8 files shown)


jlprat:
OK TO TEST

akka-http-core/src/main/scala/akka/http/impl/model/parser/CharacterClasses.scala
```diff
@@ -49,6 +49,7 @@ private[http] object CharacterClasses {
   val `relaxed-fragment-char` = VCHAR -- '%'
   val `relaxed-path-segment-char` = VCHAR -- "%/?#"
   val `relaxed-query-char` = VCHAR -- "%&=#"
+  val `relaxed-query-value-char` = VCHAR -- "%&#"
```
2m: To keep it consistent with the [rename below](https://github.com/akka/akka-http/pull/1190/files#diff-52e496fada3f67a2626c662c33c0d392R87) rename `relaxed-query-char` to `relaxed-query-key-char` and do the same for `strict-query-char`.

akka-http-core/src/main/scala/akka/http/impl/model/parser/UriParser.scala
```diff
@@ -184,12 +188,16 @@ private[http] final class UriParser(val input: ParserInput,
 
   // http://www.w3.org/TR/html401/interact/forms.html#h-17.13.4.1
   def query: Rule1[Query] = {
-    def part = rule(
-      clearSBForDecoding() ~ oneOrMore('+' ~ appendSB(' ') | `query-char` ~ appendSB() | `pct-encoded`) ~ push(getDecodedString())
+    def keyPart = rule(clearSBForDecoding() ~
+      oneOrMore('+' ~ appendSB(' ') | `query-key-char` ~ appendSB() | `pct-encoded`) ~ push(getDecodedString())
+        | push(""))
+
+    def valuePart = rule(clearSBForDecoding() ~
```
jrudolph: Can be simplified to

```scala
    def part(`query-char`: CharPredicate) =
      rule(clearSBForDecoding() ~
        oneOrMore('+' ~ appendSB(' ') | `query-char` ~ appendSB() | `pct-encoded`) ~ push(getDecodedString())
        | push(""))

    def keyValuePair: Rule2[String, String] = rule {
      part(`query-key-char`) ~ ('=' ~ part(`query-value-char`) | push(Query.EmptyValue))
    }
```

(then the error message expectations need to be adapted as well again).

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/ParameterDirectivesSpec.scala
```diff
@@ -49,15 +49,6 @@ class ParameterDirectivesSpec extends FreeSpec with GenericRoutingSpec with Insi
           }
         }
       }
-      "cause a MalformedRequestContentRejection on invalid query strings" in {
```
jrudolph: The test should be adapted and kept (or is there another reason for deleting it?)

akka-http-tests/src/test/java/akka/http/javadsl/server/directives/ParameterDirectivesTest.java
```diff
@@ -34,8 +34,8 @@ public void testStringParameterExtraction() {
 
     route
       .run(HttpRequest.create().withUri("/abc?stringParam=a=b"))
-      .assertStatusCode(400)
```
jrudolph: Keep the old assertions but use a different faulty query, we want to test the faulty case here.

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/ParameterDirectivesSpec.scala
gmethvin: Mainly because this type of error doesn't exist anymore. I can adapt to another failing case.

akka-http-core/src/main/scala/akka/http/impl/model/parser/CharacterClasses.scala
gmethvin: Sounds good. I had meant to come back and change these names.

gmethvin:
@jrudolph updated

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/ParameterDirectivesSpec.scala
jrudolph: Yep, thanks, that's what I meant.

akka-http-tests/src/test/java/akka/http/javadsl/server/directives/ParameterDirectivesTest.java
```diff
@@ -33,9 +33,14 @@ public void testStringParameterExtraction() {
       .assertEntity("john");
 
     route
+        .run(HttpRequest.create().withUri("/abc?stringParam=a%b"))
+        .assertStatusCode(400)
+        .assertEntity("The request content was malformed:\nThe request's query string is invalid: stringParam=a%b");
+
+    route
       .run(HttpRequest.create().withUri("/abc?stringParam=a=b"))
```
jrudolph: Ah I see we've now just got an extra case.

jrudolph:
PLS BUILD

jrudolph:
PLS BUILD

PR #1192, opened 2017-06-13
Add extra test to narrow down failure #443

raboof:
I agree with Johannes that the only code path to this failure is when the
parboiled dispatcher would throw `RuleNotFoundException`, but don't see how
this could be nondeterministic. To further narrow this down I added a test that
fails specifically in that case, should be interesting to see if that one also
fails (and whether they fail together).

Refs #443


jrudolph:
Maybe let's merge it and see what happens?

ktoso:
+1 for merging detective work helpers such as this ;)

PR #1193, opened 2017-06-13
+htk #1191 introduce internal ScalaTestJunitRouteTest to run JUnit tests with Scalatest

### LINKED ISSUE #1191: Reenable JUnit test for PR validation
Since #1189 our own Java JUnit tests are not run any more.

Unfortunately, just enabling the JUnit test interface for sbt is not sufficient because of https://github.com/sbt/junit-interface/issues/35.

So for now, we might need to introduce another intermediate class for our own tests that also extends from `JUnitSpecLike` to run JUnit tests with Scalatest as before.

jrudolph:
Fixes #1191.


jrudolph:
It seems that isn't needed after all.

PR #1194, opened 2017-06-13
Remove debugging output

raboof:


ktoso:
ah "debugging output". Was worried this removes test run output from title :)

jonas:
As they say, no news is good news!

PR #1195, opened 2017-06-13
Revamp SPI / API of ClientTransport infrastructure to fix #1012 and to simplify API

### LINKED ISSUE #1012: idle-timeout became broken between 10.0.4 and 10.0.5?

jrudolph:
This was inspired by part of @ktoso's fix in #1014 but goes a bit further to simplify the transport API.

A transport is now a static constant function value that gets passed
all parameters at connection time.

That fixes #1012 where client connection settings were both kept redundantly
inside the transport and inside the settings.

It's now also possible to specify the localAddress inside the settings. Methods
in the Http entrypoint that allow specifying the localAddress explicitly will
override whatever was given in the settings. In the future we'd like to
get rid of all those parameters to simplify entrypoint signatures in Http.

There is a basic disadvantage of passing `ClientConnectionSettings` to the `connectTo` method which is that whatever settings future `ClientTransport` come up with, they have to be put into those `ClientConnectionSettings`, so this might become an even bigger grab bag than it is right now.

Fixes #1012.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/ClientTransport.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/ClientTransport.scala
@@ -17,23 +17,28 @@ import scala.concurrent.Future
 
 /**
  * Abstraction to allow the creation of alternative transports to run HTTP on.
+ *
+ * (Still unstable) SPI for implementors of custom client transports.
  */
 @ApiMayChange
 trait ClientTransport { outer ⇒
-  def connectTo(host: String, port: Int)(implicit system: ActorSystem): Flow[ByteString, ByteString, Future[OutgoingConnection]]
+  def connectTo(host: String, port: Int, settings: ClientConnectionSettings)(implicit system: ActorSystem): Flow[ByteString, ByteString, Future[OutgoingConnection]]
 }
 
+/**
+ * (Still unstable) entry point to create or access predefined client transports.
+ */
+@ApiMayChange
 object ClientTransport {
-  def TCP(localAddress: Option[InetSocketAddress], settings: ClientConnectionSettings): ClientTransport =
-    new TCPTransport(localAddress, settings)
+  val TCP: ClientTransport = TCPTransport
 
-  private case class TCPTransport(localAddress: Option[InetSocketAddress], settings: ClientConnectionSettings) extends ClientTransport {
-    def connectTo(host: String, port: Int)(implicit system: ActorSystem): Flow[ByteString, ByteString, Future[OutgoingConnection]] =
+  private case object TCPTransport extends ClientTransport {
+    def connectTo(host: String, port: Int, settings: ClientConnectionSettings)(implicit system: ActorSystem): Flow[ByteString, ByteString, Future[OutgoingConnection]] =
       // The InetSocketAddress representing the remote address must be created unresolved because akka.io.TcpOutgoingConnection will
       // not attempt DNS resolution if the InetSocketAddress is already resolved. That behavior is problematic when it comes to
       // connection pools since it means that new connections opened by the pool in the future can end up using a stale IP address.
       // By passing an unresolved InetSocketAddress instead, we ensure that DNS resolution is performed for every new connection.
-      Tcp().outgoingConnection(InetSocketAddress.createUnresolved(host, port), localAddress,
+      Tcp().outgoingConnection(InetSocketAddress.createUnresolved(host, port), settings.localAddress,
         settings.socketOptions, halfClose = true, settings.connectingTimeout, settings.idleTimeout)
         .mapMaterializedValue(_.map(tcpConn ⇒ OutgoingConnection(tcpConn.localAddress, tcpConn.remoteAddress))(system.dispatcher))
   }
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
@@ -297,7 +297,7 @@ class HttpExt(private val config: Config)(implicit val system: ActorSystem) exte
                          localAddress: Option[InetSocketAddress] = None,
                          settings:     ClientConnectionSettings  = ClientConnectionSettings(system),
                          log:          LoggingAdapter            = system.log): Flow[HttpRequest, HttpResponse, Future[OutgoingConnection]] =
-    _outgoingConnection(host, port, settings, ConnectionContext.noEncryption(), ClientTransport.TCP(localAddress, settings), log)
+    _outgoingConnection(host, port, settings.withLocalAddressOverride(localAddress), ConnectionContext.noEncryption(), ClientTransport.TCP, log)
 
   /**
    * Same as [[#outgoingConnection]] but for encrypted (HTTPS) connections.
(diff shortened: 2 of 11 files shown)


jrudolph:
PLS BUILD

akka-http-core/src/main/scala/akka/http/scaladsl/ClientTransport.scala
```diff
@@ -17,23 +17,28 @@ import scala.concurrent.Future
...
+@ApiMayChange
 object ClientTransport {
-  def TCP(localAddress: Option[InetSocketAddress], settings: ClientConnectionSettings): ClientTransport =
-    new TCPTransport(localAddress, settings)
+  val TCP: ClientTransport = TCPTransport
 
-  private case class TCPTransport(localAddress: Option[InetSocketAddress], settings: ClientConnectionSettings) extends ClientTransport {
-    def connectTo(host: String, port: Int)(implicit system: ActorSystem): Flow[ByteString, ByteString, Future[OutgoingConnection]] =
+  private case object TCPTransport extends ClientTransport {
+    def connectTo(host: String, port: Int, settings: ClientConnectionSettings)(implicit system: ActorSystem): Flow[ByteString, ByteString, Future[OutgoingConnection]] =
```
ktoso: Yeah, this solves the problem of it being passed in "too early and only once", good.

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
```diff
@@ -297,7 +297,7 @@ class HttpExt(private val config: Config)(implicit val system: ActorSystem) exte
                          localAddress: Option[InetSocketAddress] = None,
                          settings:     ClientConnectionSettings  = ClientConnectionSettings(system),
                          log:          LoggingAdapter            = system.log): Flow[HttpRequest, HttpResponse, Future[OutgoingConnection]] =
-    _outgoingConnection(host, port, settings, ConnectionContext.noEncryption(), ClientTransport.TCP(localAddress, settings), log)
+    _outgoingConnection(host, port, settings.withLocalAddressOverride(localAddress), ConnectionContext.noEncryption(), ClientTransport.TCP, log)
```
ktoso: good, makes sense; this is where the passing in of the settings previously did not do what was expected AFAIR

akka-http-core/src/main/scala/akka/http/scaladsl/settings/ClientConnectionSettings.scala
```diff
@@ -61,6 +44,15 @@ abstract class ClientConnectionSettings private[akka] () extends akka.http.javad
...
+  def withLocalAddress(newValue: Option[InetSocketAddress]): ClientConnectionSettings = self.copy(localAddress = newValue)
+
+  /**
+   * Returns a new instance with the given local address set if the given override is `Some(address)`, otherwise
+   * return this instance unchanged.
+   */
+  def withLocalAddressOverride(overrideLocalAddressOption: Option[InetSocketAddress]): ClientConnectionSettings =
+    if (overrideLocalAddressOption.isDefined) withLocalAddress(overrideLocalAddressOption)
+    else this
 }
```
ktoso: good 👍

akka-http-core/src/main/scala/akka/http/scaladsl/settings/ClientConnectionSettings.scala
```diff
@@ -34,19 +29,7 @@ abstract class ClientConnectionSettings private[akka] () extends akka.http.javad
...
-  final override def getConnectingTimeout: FiniteDuration = connectingTimeout
-  final override def getParserSettings: js.ParserSettings = parserSettings
-  final override def getIdleTimeout: Duration = idleTimeout
-  final override def getSocketOptions: Iterable[SocketOption] = socketOptions.asJava
-  final override def getUserAgentHeader: Optional[UserAgent] = OptionConverters.toJava(userAgentHeader)
-  final override def getLogUnencryptedNetworkBytes: Optional[Int] = OptionConverters.toJava(logUnencryptedNetworkBytes)
-  final override def getRequestHeaderSizeHint: Int = requestHeaderSizeHint
-  final override def getWebsocketRandomFactory: Supplier[Random] = new Supplier[Random] {
-    override def get(): Random = websocketRandomFactory()
-  }
```
ktoso: moving them up one level looks good 👍

akka-http-core/src/test/scala/akka/http/impl/engine/client/ConnectionPoolSpec.scala
```diff
@@ -566,4 +591,45 @@ class ConnectionPoolSpec extends AkkaSpec("""
...
+      }
+    }
+
+    import system.dispatcher
+
+    val transport = new CustomTransport
+    val poolSettings =
+      ConnectionPoolSettings(system)
+        .withTransport(transport)
+        .withConnectionSettings(ClientConnectionSettings(system).withIdleTimeout(23.hours))
```
ktoso: extract the `23.hours` as value? a bit easier to see where it's set / expected then?

akka-http-core/src/test/scala/akka/http/impl/engine/client/ConnectionPoolSpec.scala
```diff
@@ -566,4 +591,45 @@ class ConnectionPoolSpec extends AkkaSpec("""
...
+    transport.in.expectUtf8EncodedString("GET /test HTTP/1.1\r\n")
+    transport.out.sendNext(ByteString("HTTP/1.1 200 OK\r\nConnection: close\r\n\r\n"))
+
+    val response = responseFuture.awaitResult(10.seconds)
+    response.status should ===(StatusCodes.OK)
+    transport.out.sendNext(ByteString("Hello World!"))
+    transport.out.sendComplete()
+
+    response.entity.dataBytes.utf8String.awaitResult(10.seconds) should ===("Hello World!")
+  }
```
ktoso: Great test setup!

ktoso:
> There is a basic disadvantage of passing ClientConnectionSettings to the connectTo method which is that whatever settings future ClientTransport come up with, they have to be put into those ClientConnectionSettings, so this might become an even bigger grab bag than it is right now.

Yeah... though perhaps won't been much more of a problem than it is already in practice; For now we only have the Proxy coming up there AFAICS; Technically we could pull off some trick like transportSettings: Map[TransportSettingKey[V], V] or something if we care about avoiding more fields in there etc (would allow keeping external projects to contribute transports - though I'm not sure that will happen or should be a goal - what was your plan about that?)

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
jrudolph: No, the problem was that `ConnectionPoolSettings.withConnectionSettings(...)` had no effect because the settings had already been passed to the transport.

PR #1201, opened 2017-06-14
=doc #1199 reference security announcements and release notes in both java and scala toc

### LINKED ISSUE #1199: Security announcements and release notes are very hard to see and access in new docs scheme
They are not shown in the main TOC (which defaults to file:///home/johannes/git/opensource/akka-http/docs/target/paradox/site/main/scala/http/index.html) and there's no link that would go to the main overview.

jrudolph:
Fixes #1199.


jrudolph:
Ah I see. Interesting idea. That would also solve that clicking on this link from the Java side of things switches over to the Scala side. If you like, you could follow up on this next week.

raboof:
(for the migration guides I'm including all guides for both Scala and Java, that should be okay in this case. If we ever 'really' have a section that only applies to one group we might have to add support for that to the ToC component, see https://github.com/lightbend/paradox/issues/128 )

PR #1202, opened 2017-06-14
[do-not-merge / preview] =par use hashmap instead of binary tree of parsers

jrudolph:
This has several advantages:
  * less code generation complexity
  * less code
  * less allocations as Function0 instances are all pre-allocated

I wonder if that would solve #443? Anyway, as this change might be performance sensitive we should run some benchmarks to see if the Scala immutable hashmap is fast enough or if we should take another hashmap for this fixed set of entries. Another optimization we could then do is to provide a way to query the map directly and cache the parser entry point functions directly in the highly optimized HttpHeaderParser tree.

/cc @raboof as you were looking at #443 as well.
fyi @sirthias

### DIFF
--- a/akka-parsing/src/main/scala/akka/parboiled2/DynamicRuleDispatch.scala
+++ b/akka-parsing/src/main/scala/akka/parboiled2/DynamicRuleDispatch.scala
@@ -54,32 +54,26 @@ object DynamicRuleDispatch {
 
   def __create[P <: Parser, L <: HList](c: whitebox.Context)(ruleNames: c.Expr[String]*)(implicit P: c.WeakTypeTag[P], L: c.WeakTypeTag[L]): c.Expr[(DynamicRuleDispatch[P, L], immutable.Seq[String])] = {
     import c.universe._
-    val names: Array[String] = ruleNames.map {
+    val names = ruleNames.map {
       _.tree match {
         case Literal(Constant(s: String)) ⇒ s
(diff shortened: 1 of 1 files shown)


akka-parsing/src/main/scala/akka/parboiled2/DynamicRuleDispatch.scala
```diff
@@ -54,32 +54,26 @@ object DynamicRuleDispatch {
...
 
     c.Expr[(DynamicRuleDispatch[P, L], immutable.Seq[String])] {
-      q"""val drd =
+      q"""val map: Map[String, $P => () => RuleN[$L]] = Map(..$ruleEntries)
+          val drd =
             new akka.parboiled2.DynamicRuleDispatch[$P, $L] {
               def apply(handler: akka.parboiled2.DynamicRuleHandler[$P, $L], ruleName: String): handler.Result =
-                 ${rec(0, names.length - 1)}
+                if (map.contains(ruleName)) {
+                  val rule = map(ruleName).apply(handler.parser)
```
raboof: first `.contains` and then `.apply` will have to do the lookup twice, right? Can this be rewritten to only do it once (without introducing other overhead)?

raboof:
Removed merge conflict in #1534

jrudolph:
Merged as #1534. Thanks, @raboof for porting to the latest master.

Issue #1121, opened 2017-05-13
FAILED: akka.http.scaladsl.server.HttpAppSpec.HttpApp should start without ActorSystem


jlprat:
https://jenkins.akka.io:8498/job/pr-validator-akka-http/1102/testReport/junit/akka.http.scaladsl.server/HttpAppSpec/HttpApp_should_start_without_ActorSystem/

jlprat:
Probably related to https://github.com/akka/akka-http/issues/1061

2m:
Fixed in #1207

PR #1207, opened 2017-06-18
=httst Improve test reliability for HttpApp #1121 #1061

jlprat:
Issues: #1121 #1061
Refactor test to avoid race condition between serving a request and shutting down the server itself.
Now, server is shutdown independently from serving a request.


raboof:
Good observation.

Would be cool if we had a nice way to "drain" ("stop listening for new requests but finish in-flight ones and shut down", with a timeout). Can we do that?

ktoso:
> Would be cool if we had a nice way to "drain" ("stop listening for new requests but finish in-flight ones and shut down", with a timeout). Can we do that?

Hm good point, that is what coordinated shutdown does so we should make it simple to use here

jlprat:
Cool, I'll read further about it.

PR #1211, opened 2017-06-19
adds Cache-Control 'immutable' value

matsluni:
Hi,

after reading [this](https://hacks.mozilla.org/2017/01/using-immutable-caching-to-speed-up-the-web/ ) a while ago I thought that the `immutable` Cache-Control value might be a good addition and so I created this PR.

Let me know what you think.

Best,
Matthias

See also
- https://bitsup.blogspot.de/2016/05/cache-control-immutable.html
- https://tools.ietf.org/wg/httpbis/draft-ietf-httpbis-immutable/

### DIFF
--- a/akka-http-core/src/main/java/akka/http/javadsl/model/headers/CacheDirectives.java
+++ b/akka-http-core/src/main/java/akka/http/javadsl/model/headers/CacheDirectives.java
@@ -42,5 +42,5 @@ public final class CacheDirectives {
     public static CacheDirective S_MAXAGE(long deltaSeconds) {
         return new akka.http.scaladsl.model.headers.CacheDirectives.s$minusmaxage(deltaSeconds);
     }
-    public static final CacheDirective IMMUTABLE = akka.http.scaladsl.model.headers.CacheDirectives.immutable$.MODULE$;
+    public static final CacheDirective IMMUTABLE = akka.http.scaladsl.model.headers.CacheDirectives.getImmutable();
 }
--- a/akka-http-core/src/main/java/akka/http/javadsl/model/headers/CacheDirectives.java
+++ b/akka-http-core/src/main/java/akka/http/javadsl/model/headers/CacheDirectives.java
@@ -42,4 +42,5 @@ public final class CacheDirectives {
     public static CacheDirective S_MAXAGE(long deltaSeconds) {
         return new akka.http.scaladsl.model.headers.CacheDirectives.s$minusmaxage(deltaSeconds);
     }
+    public static final CacheDirective IMMUTABLE = akka.http.scaladsl.model.headers.CacheDirectives.immutable$.MODULE$;
 }
--- a/akka-http-core/src/main/scala/akka/http/impl/model/parser/CacheControlHeader.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/model/parser/CacheControlHeader.scala
@@ -28,6 +28,7 @@ private[parser] trait CacheControlHeader { this: Parser with CommonRules with Co
       | "must-revalidate" ~ push(`must-revalidate`)
       | "proxy-revalidate" ~ push(`proxy-revalidate`)
       | "s-maxage=" ~ `delta-seconds` ~> (`s-maxage`(_))
+      | "immutable" ~ push(`immutable`)
       | token ~ optional(ws('=') ~ word) ~> (CacheDirective.custom(_, _)))
 
   def `field-names` = rule { `quoted-tokens` | token ~> (Seq(_)) }
--- a/akka-http-core/src/main/scala/akka/http/impl/model/parser/CacheControlHeader.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/model/parser/CacheControlHeader.scala
@@ -7,6 +7,7 @@ package akka.http.impl.model.parser
 import akka.parboiled2.Parser
 import akka.http.scaladsl.model.headers._
 import CacheDirectives._
+import CacheDirectives.ImmutableCacheValue._
 
 private[parser] trait CacheControlHeader { this: Parser with CommonRules with CommonActions with StringBuilding ⇒
 
--- a/akka-http-core/src/main/scala/akka/http/impl/model/parser/CacheControlHeader.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/model/parser/CacheControlHeader.scala
@@ -7,7 +7,6 @@ package akka.http.impl.model.parser
 import akka.parboiled2.Parser
 import akka.http.scaladsl.model.headers._
 import CacheDirectives._
-import CacheDirectives.ImmutableCacheValue._
 
 private[parser] trait CacheControlHeader { this: Parser with CommonRules with CommonActions with StringBuilding ⇒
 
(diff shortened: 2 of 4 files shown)


jlprat:
OK TO TEST

jlprat:
Hi @matsluni, could you create an issue for this as well?
As this is not just a typo fix, it is better to have the corresponding issue to keep track of it.
Thanks!

ktoso:
Created issue to track this https://github.com/akka/akka-http/issues/1212

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/CacheDirective.scala
```diff
@@ -86,22 +86,25 @@ object CacheDirectives {
...
 
+  // https://tools.ietf.org/wg/httpbis/draft-ietf-httpbis-immutable/
+  case object `immutable` extends SingletonValueRenderable with ResponseDirective
+
   /** Java API */
   def getPublic: ResponseDirective = `public`
 
   // http://tools.ietf.org/html/rfc7234#section-5.2.2.6
-  final case class `private`(fieldNames: immutable.Seq[String]) extends FieldNamesDirective with ResponseDirective
+  final case class `private`(fieldNames: colImmutable.Seq[String]) extends FieldNamesDirective with ResponseDirective
```
ktoso: Hm, we must be careful with this change. 
Think about users of this class, now when they upgrade and they used the popular style of writing `immutable.Seq` etc all their code is broken (this change is not source compatible in a popular case for users).

I think we perhaps have to namespace the object somehow else...?

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/CacheDirective.scala
```diff
@@ -86,22 +86,25 @@ object CacheDirectives {
...
   object `private` {
-    def apply(fieldNames: String*): `private` = new `private`(immutable.Seq(fieldNames: _*))
+    def apply(fieldNames: String*): `private` = new `private`(colImmutable.Seq(fieldNames: _*))
```
ktoso: (that's the breakage we should not cause for our users)

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/CacheDirective.scala
```diff
@@ -86,22 +86,25 @@ object CacheDirectives {
   case object `must-revalidate` extends SingletonValueRenderable with ResponseDirective
 
   // http://tools.ietf.org/html/rfc7234#section-5.2.2.2
-  final case class `no-cache`(fieldNames: immutable.Seq[String]) extends FieldNamesDirective with ResponseDirective
+  final case class `no-cache`(fieldNames: colImmutable.Seq[String]) extends FieldNamesDirective with ResponseDirective
 
   // http://tools.ietf.org/html/rfc7234#section-5.2.2.5
   case object `public` extends SingletonValueRenderable with ResponseDirective
 
+  // https://tools.ietf.org/wg/httpbis/draft-ietf-httpbis-immutable/
```
ktoso: Since it's a draft we perhaps have to mark it as `@ApiMayChange`

akka-http-core/src/main/scala/akka/http/impl/model/parser/CacheControlHeader.scala
```diff
@@ -28,6 +28,7 @@ private[parser] trait CacheControlHeader { this: Parser with CommonRules with Co
       | "must-revalidate" ~ push(`must-revalidate`)
       | "proxy-revalidate" ~ push(`proxy-revalidate`)
       | "s-maxage=" ~ `delta-seconds` ~> (`s-maxage`(_))
+      | "immutable" ~ push(`immutable`)
```
ktoso: ok

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/CacheDirective.scala
ktoso: Perhaps also make the comment an actual scaladoc, so people can read up on it (without having to get sources)

akka-http-core/src/main/java/akka/http/javadsl/model/headers/CacheDirectives.java
```diff
@@ -42,4 +42,5 @@ public static CacheDirective PRIVATE(String... fieldNames) {
     public static CacheDirective S_MAXAGE(long deltaSeconds) {
         return new akka.http.scaladsl.model.headers.CacheDirectives.s$minusmaxage(deltaSeconds);
     }
+    public static final CacheDirective IMMUTABLE = akka.http.scaladsl.model.headers.CacheDirectives.immutable$.MODULE$;
```
ktoso: same comments apply as to the scaladsl one

matsluni:
Thanks @ktoso for creating the issue and the comprehensive feedback. I will think about how this can be introduced in a not so breaking way.

matsluni:
@ktoso I updated the PR. Please have a look again.

I put the `immutable`-object in an additional `ImmutableCacheValue`object. This needs an additional import for the user to be used (see [HttpHeaderSpec](https://github.com/matsluni/akka-http/blob/3a9fa7129aef29fc5fdb33a447cd791b3444c28e/akka-http-core/src/test/scala/akka/http/impl/model/parser/HttpHeaderSpec.scala#L15))

I don't really like this additional import, but couldn't come up with a better solution. Do you have another idea, which maybe makes this additional import not necessary?

jrudolph:
I agree about the compatibility issue. Under the assumption that a user has done `import CacheDirectives._` and also `import scala.collection.immutable` the original version of that change won't be source compatible. Compilation will then fail with an error like this:

```
error: reference to immutable is ambiguous;
it is imported twice in the same scope by
import CacheDirectives._
and import scala.collection.immutable
       immutable
```

On the other hand, personally, I'm less concerned about source compatibility than about binary compatibility. It will be a simple fix for everyone to make their sources work with the addition. Also, any change that introduces symbols has the same problem in theory and we usually don't care about it. Symbols are lying in namespaces for exactly that reason. Also, some organization generally ban wildcard imports to prevent these kinds of issues.

But I also see, that `import collection.immutable` may be widespread enough to not introduce source incompatibilities in a fix release. We should maybe also add a clause to our compatibility notes, that we don't support that strict kind of source compatibility but that we are allowed to add any symbols in any version in the future even if that clashes with existing imports (at least in reasonable scopes like the somewhat uncommon `CacheDirectives` scope here).

So, maybe let's keep it simple:

 * For now just call the instance `CacheDirectives.immutableDirective`. In IDEs that should be as detectable as just `immutable`.
 * Decide later if, for 11.x, we add a forwarder which is called plain `immutable` (and deprecate the old one)

WDYT?

matsluni:
Thanks @jrudolph for your feedback.

Do I understand you correct that you suggest this:
```
/** https://tools.ietf.org/wg/httpbis/draft-ietf-httpbis-immutable */
@ApiMayChange
case object immutableDirective extends SingletonValueRenderable with ResponseDirective
```

jrudolph:
Yes, exactly (and sorry for keeping this question around for so long...)

matsluni:
I updated the PR to address review comments. Please have a look again.

ktoso:
Resolves https://github.com/akka/akka-http/issues/1212

PR #1214, opened 2017-06-20
=doc fix link to companion object docs in http model

ktoso:
It must be a link to the companion, previous link was 404


ktoso:
Please include in releasse @2m @raboof

PR #1221, opened 2017-06-21
Manually advance time in tests (#595)

raboof:
Makes tests a bit less realistic, but more controlled, faster and more predictable.


akka-http-core/src/test/scala/akka/testkit/ExplicitlyTriggeredScheduler.scala
```diff
@@ -0,0 +1,75 @@
...
+import scala.concurrent.ExecutionContext
+import scala.concurrent.duration.{ Duration, FiniteDuration }
+
+/**
+ * For testing: scheduler that does not look at the clock, but must be progressed manually by calling `timePasses`.
+ *
+ * This is not entirely realistic: jobs will be executed on the test thread instead of using the `ExecutionContext`, but does
+ * allow for faster and less timing-sensitive specs..
+ */
+class ExplicitlyTriggeredScheduler(config: Config, log: LoggingAdapter, tf: ThreadFactory) extends Scheduler {
```
jrudolph: Wouldn't all of the methods still need to be thread-safe?

akka-http-core/src/test/scala/akka/testkit/ExplicitlyTriggeredScheduler.scala
```diff
@@ -0,0 +1,75 @@
...
+  var scheduled: List[Item] = Nil
+
+  override def schedule(initialDelay: FiniteDuration, interval: FiniteDuration, runnable: Runnable)(implicit executor: ExecutionContext): Cancellable =
+    schedule(initialDelay, Some(interval), runnable)
+
+  override def scheduleOnce(delay: FiniteDuration, runnable: Runnable)(implicit executor: ExecutionContext): Cancellable =
+    schedule(delay, None, runnable)
+
+  def timePasses(amount: FiniteDuration)(implicit system: ActorSystem) = {
+    executeTasks(currentTime + amount.dilated.toMillis)
```
jrudolph: If the `currentTime` isn't advanced first, won't newly scheduled tasks be scheduled to an earlier time than expected? Couldn't that mean infinite loops in the worst case, e.g. if `amount = 50.ms` and one of the running tasks always schedules things to a time `< 50ms`?

akka-http-core/src/test/scala/akka/testkit/ExplicitlyTriggeredScheduler.scala
raboof: You're right - improved this.

akka-http-core/src/test/scala/akka/testkit/ExplicitlyTriggeredScheduler.scala
raboof: Indeed, fixed

raboof:
PLS BUILD

akka-http-core/src/test/scala/akka/http/impl/engine/server/HttpServerSpec.scala
```diff
@@ -1010,8 +1014,14 @@ class HttpServerSpec extends AkkaSpec(
 
       "are programmatically increased (expiring)" in assertAllStagesStopped(new RequestTimeoutTestSetup(50.millis) {
         send("GET / HTTP/1.1\r\nHost: example.com\r\n\r\n")
+
+        scheduler.timePasses(25.millis)
         expectRequest().header[`Timeout-Access`].foreach(_.timeoutAccess.updateTimeout(250.millis.dilated))
-        netOut.expectNoBytes(150.millis.dilated)
+
+        scheduler.timePasses(150.millis)
+        netOut.expectNoBytes(Duration.Zero)
```
jrudolph: We should try to add support for that scheduler right into those probes.

raboof:
Refs #595

Issue #1167, opened 2017-05-29
Access to websocket server materialized value


mathbruyen:
Websocket server methods do not allow accessing the materialized value of the flow it is given. They take a flow as argument and return the HTTP response to send back:

```java
HttpResponse response = WebSocket.handleWebSocketRequestWith(request, greeterFlow);
```

The server takes care of materializing the flow and ignores both values. This prevents users from getting them, which can be problematic. For example if the flow source is a queue, one needs to get its materialized value in order to push to the queue. A workaround is to setup a callback on `mapMaterializedValue` to send the queue to whoever is interested in it:

```java
Source.<Message>queue(5, OverflowStrategy.backpressure())
    .mapMaterializedValue(sourceQueue -> { /* send queue to whoever is interested */; return sourceQueue; });
```

But this is a workaround rather than a clean solution, as `mapMaterializedValue` is meant to transform the materialized value rather than broadcast it. The client has methods that return this materialized value:

```java
Pair<CompletionStage<WebSocketUpgradeResponse>, SourceQueueWithComplete<Message>> pair =
        http.singleWebSocketRequest(WebSocketRequest.create(connect), flow, materializer);
```

It would be great if the server had methods for doing something similar. This might be tricky since contrary to the client, the server takes care of materializing the value after the response it sent.

(mailing list pointer: [Graph materialized value of websocket upgrade](https://groups.google.com/forum/#!topic/akka-user/p4HDrgmu6fo))

jrudolph:
I agree, @mathbruyen, that it would be nice to get access to the materialized value. Right now the recommended way is indeed using `mapMaterializedValue`. Do you have an idea how the API would look like? E.g. we could add another method `handleWebSocketRequestWithMat` that also returns a `Future[Mat]`. Would that work for you?

ktoso:
Yeah, I think a method like `handleWebSocketRequestWithMat: (HttpResponse, Future[Mat])` would solve this and be pretty consistent with other APIs we have

mathbruyen:
It would work for me yes, supposedly returning `Pair[HttpResponse, Future[Mat]]` in order to make response to send available as well.

sportanova:
[...]

PR #1229, opened 2017-06-24
Provide access to materialized value from websocket server https://gi…

sportanova:
https://github.com/akka/akka-http/issues/1167

* Modify the Websocket object, as well as the UpgradeToWebSocket traits (Scala + Java DSLs) to
  return a future of the materialized view, along with the HttpResponse

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/javadsl/model/ws/WebSocket.scala
+++ b/akka-http-core/src/main/scala/akka/http/javadsl/model/ws/WebSocket.scala
@@ -7,6 +7,9 @@ package akka.http.javadsl.model.ws
 import akka.stream.javadsl.Flow
 import akka.http.javadsl.model._
 import akka.http.impl.util.JavaMapping.Implicits._
+import akka.japi.Pair
+import akka.stream.{ FlowShape, Graph }
+import scala.concurrent.Future
 
 object WebSocket {
   /**
@@ -18,5 +21,22 @@ object WebSocket {
     request.asScala.header[UpgradeToWebSocket] match {
       case Some(header) ⇒ header.handleMessagesWith(handler)
       case None         ⇒ HttpResponse.create().withStatus(StatusCodes.BAD_REQUEST).withEntity("Expected WebSocket request")
+    }
+
+  /**
+   * If a given request is a WebSocket request a response accepting the request is returned using the given handler to
+   * handle the WebSocket message stream, as well as a future materialized value. If the request wasn't a WebSocket
+   * request a response with status code 400 is returned, as well as a failed future.
+   */
+  def handleWebSocketRequestWithMat[Mat](
+    request:     HttpRequest,
+    handler:     Graph[FlowShape[Message, Message], Mat],
+    subprotocol: String): Pair[HttpResponse, Future[Mat]] =
+    request.asScala.header[UpgradeToWebSocket] match {
+      case Some(header) ⇒ header.handleMessagesWithMat(handler, subprotocol)
+      case None ⇒ {
+        val httpResponse = HttpResponse.create().withStatus(StatusCodes.BAD_REQUEST).withEntity("Expected WebSocket request")
+        Pair(httpResponse, Future.failed(new IllegalStateException("Expected WebSocket request")))
+      }
     }
 }
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/ws/UpgradeToWebSocket.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/ws/UpgradeToWebSocket.scala
@@ -69,7 +69,7 @@ trait UpgradeToWebSocket extends jm.ws.UpgradeToWebSocket {
    */
   def handleMessagesWithMat[Mat](
     handlerFlow: Graph[FlowShape[Message, Message], Mat],
-    subprotocol: Option[String] = None): (HttpResponse, Future[Mat]) = {
+    subprotocol: Option[String]                          = None): (HttpResponse, Future[Mat]) = {
     val promise = Promise[Mat]
     val response = handleMessages(Flow.fromGraph(handlerFlow).mapMaterializedValue(mat ⇒ promise.success(mat)), subprotocol)
     (response, promise.future)
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/ws/UpgradeToWebSocket.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/ws/UpgradeToWebSocket.scala
@@ -5,14 +5,17 @@
 package akka.http.scaladsl.model.ws
 
 import java.lang.Iterable
+
 import scala.collection.immutable
 import akka.NotUsed
+import akka.annotation.DoNotInherit
 import akka.stream._
 import akka.http.impl.util.JavaMapping
 import akka.http.javadsl.{ model ⇒ jm }
 import akka.http.scaladsl.model.HttpResponse
 import akka.japi.Pair
 import akka.stream.scaladsl.Flow
+
 import scala.concurrent.{ Future, Promise }
 
 /**
@@ -20,6 +23,7 @@ import scala.concurrent.{ Future, Promise }
  * enables a request handler to upgrade this connection to a WebSocket connection and
  * registers a WebSocket handler.
  */
+@DoNotInherit
 trait UpgradeToWebSocket extends jm.ws.UpgradeToWebSocket {
   /**
    * A sequence of protocols the client accepts.
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/ws/UpgradeToWebSocket.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/ws/UpgradeToWebSocket.scala
@@ -11,6 +11,9 @@ import akka.stream._
 import akka.http.impl.util.JavaMapping
 import akka.http.javadsl.{ model ⇒ jm }
 import akka.http.scaladsl.model.HttpResponse
+import akka.japi.Pair
+import akka.stream.scaladsl.Flow
+import scala.concurrent.{ Future, Promise }
 
 /**
  * A custom header that will be added to an WebSocket upgrade HttpRequest that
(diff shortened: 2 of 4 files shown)


sportanova:
I haven't added tests yet - would the best place to add them be in WebSocketServerSpec? And the main test I can think of is to test the HttpResponse, like the other specs in their, but also test that the Future[Mat] materializes

ktoso:
OK TO TEST

akka-http-core/src/main/scala/akka/http/scaladsl/model/ws/UpgradeToWebSocket.scala
```diff
@@ -88,6 +108,18 @@ trait UpgradeToWebSocket extends jm.ws.UpgradeToWebSocket {
...
+  /**
+   * Java API
+   */
+  def handleMessagesWithMat[Mat](handlerFlow: Graph[FlowShape[jm.ws.Message, jm.ws.Message], Mat]): (HttpResponse, Future[Mat]) =
+    handleMessagesWithMat(JavaMapping.toScala(handlerFlow))
+
+  /**
+   * Java API
+   */
+  def handleMessagesWithMat[Mat](handlerFlow: Graph[FlowShape[jm.ws.Message, jm.ws.Message], Mat], subprotocol: String): (HttpResponse, Future[Mat]) =
```
ktoso: this must return `akka.japi.Pair` instead, a Scala tuple is not usable from Java

akka-http-core/src/main/scala/akka/http/scaladsl/model/ws/UpgradeToWebSocket.scala
```diff
@@ -88,6 +108,18 @@ trait UpgradeToWebSocket extends jm.ws.UpgradeToWebSocket {
     subprotocol: String): HttpResponse =
     handleMessages(createScalaFlow(inSink, outSource), subprotocol = Some(subprotocol))
 
+  /**
+   * Java API
+   */
+  def handleMessagesWithMat[Mat](handlerFlow: Graph[FlowShape[jm.ws.Message, jm.ws.Message], Mat]): (HttpResponse, Future[Mat]) =
```
ktoso: must return Pair since Java API

sportanova: ah ok - so should I change it in the java dsl as well? Basically everywhere should return the ```Pair``` case class

akka-http-core/src/main/scala/akka/http/scaladsl/model/ws/UpgradeToWebSocket.scala
```diff
@@ -88,6 +109,24 @@ trait UpgradeToWebSocket extends jm.ws.UpgradeToWebSocket {
...
+  }
+
+  /**
+   * Java API
+   */
+  def handleMessagesWithMat[Mat](
+    handlerFlow: Graph[FlowShape[jm.ws.Message, jm.ws.Message], Mat],
+    subprotocol: String): Pair[jm.HttpResponse, Future[Mat]] = {
+    val tuple: (HttpResponse, Future[Mat]) = handleMessagesWithMat(JavaMapping.toScala(handlerFlow), subprotocol = Some(subprotocol))
+    Pair(tuple._1, tuple._2)
```
sportanova: @ktoso returning ```Pair``` for these now

sportanova:
@ktoso I'm not sure what to add in HttpExtensionApiSpec, what do you think would be a good direction for tests?

raboof:
(jenkins error was something on the node, `Error: Unable to access jarfile /usr/share/sbt-launcher-packaging/bin/sbt-launch.jar`)

raboof:
PLS BUILD

jrudolph:
Formatting problem:

```diff
diff --git a/akka-http-core/src/main/scala/akka/http/scaladsl/model/ws/UpgradeToWebSocket.scala b/akka-http-core/src/main/scala/akka/http/scaladsl/model/ws/UpgradeToWebSocket.scala
index f572cad..d1f1d5e 100644
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/ws/UpgradeToWebSocket.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/ws/UpgradeToWebSocket.scala
@@ -69,7 +69,7 @@ trait UpgradeToWebSocket extends jm.ws.UpgradeToWebSocket {
    */
   def handleMessagesWithMat[Mat](
     handlerFlow: Graph[FlowShape[Message, Message], Mat],
-    subprotocol: Option[String] = None): (HttpResponse, Future[Mat]) = {
+    subprotocol: Option[String]                          = None): (HttpResponse, Future[Mat]) = {
     val promise = Promise[Mat]
```

akka-http-core/src/main/scala/akka/http/javadsl/model/ws/WebSocket.scala
```diff
@@ -19,4 +22,21 @@ object WebSocket {
...
+   * If a given request is a WebSocket request a response accepting the request is returned using the given handler to
+   * handle the WebSocket message stream, as well as a future materialized value. If the request wasn't a WebSocket
+   * request a response with status code 400 is returned, as well as a failed future.
+   */
+  def handleWebSocketRequestWithMat[Mat](
+    request:     HttpRequest,
+    handler:     Graph[FlowShape[Message, Message], Mat],
+    subprotocol: String): Pair[HttpResponse, Future[Mat]] =
+    request.asScala.header[UpgradeToWebSocket] match {
+      case Some(header) ⇒ header.handleMessagesWithMat(handler, subprotocol)
```
jrudolph: We'd like to prevent duplication of functional code, so either `handleWebSocketRequest` should use the new method or an abstraction should be created to handle both cases.

raboof:
Jenkins fails due to the binary compatibility checks:
```
[info] akka-http-core: found 5 potential binary incompatibilities while checking against com.typesafe.akka:akka-http-core_2.11:10.0.2  (filtered 171)
[error]  * abstract method handleMessagesWithMat(akka.stream.Graph,java.lang.String)akka.japi.Pair in interface akka.http.javadsl.model.ws.UpgradeToWebSocket is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.javadsl.model.ws.UpgradeToWebSocket.handleMessagesWithMat")
[error]  * method handleMessagesWithMat(akka.stream.Graph,scala.Option)scala.Tuple2 in trait akka.http.scaladsl.model.ws.UpgradeToWebSocket is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.scaladsl.model.ws.UpgradeToWebSocket.handleMessagesWithMat")
[error]  * method handleMessagesWithMat(akka.stream.Graph)akka.japi.Pair in trait akka.http.scaladsl.model.ws.UpgradeToWebSocket is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.scaladsl.model.ws.UpgradeToWebSocket.handleMessagesWithMat")
[error]  * method handleMessagesWithMat(akka.stream.Graph,java.lang.String)akka.japi.Pair in trait akka.http.scaladsl.model.ws.UpgradeToWebSocket is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.scaladsl.model.ws.UpgradeToWebSocket.handleMessagesWithMat")
[error]  * synthetic method handleMessagesWithMat$default$2()scala.Option in trait akka.http.scaladsl.model.ws.UpgradeToWebSocket is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.scaladsl.model.ws.UpgradeToWebSocket.handleMessagesWithMat$default$2")
```
This basically means you added some methods to public types, which means if anybody had made their own subclasses of those types they would now be missing those methods. In this case this is likely OK since these classes weren't actually intended to be inherited by end-users anyway.

@sportanova Could you make this explicit by adding a `@DoNoInherit` annotation to those classes, and adding these exceptions as documented in `CONTRIBUTING.md`?

akka-http-core/src/main/scala/akka/http/javadsl/model/ws/WebSocket.scala
sportanova: @jrudolph i'm not sure i understand, are you referring to `handleWebSocketRequestWith`? and what is being duplicated?

akka-http-core/src/main/scala/akka/http/scaladsl/model/ws/UpgradeToWebSocket.scala
```diff
@@ -5,18 +5,25 @@
...
+import akka.stream.scaladsl.Flow
+
+import scala.concurrent.{ Future, Promise }
 
 /**
  * A custom header that will be added to an WebSocket upgrade HttpRequest that
  * enables a request handler to upgrade this connection to a WebSocket connection and
  * registers a WebSocket handler.
  */
+@DoNotInherit
```
sportanova: @raboof added the `@DoNotInherit` here and to the javadsl, as well as in the .excludes file, but still getting the binary compatibility error. Any ideas?

johanandren:
As this is against master the exclude file should be the latest released version, which is 10.1.1

raboof:
This feature makes sense to me but there's a couple of things to do in this PR before it can go in - @sportanova would you be interested in taking this one further?

akka-http-core/src/main/scala/akka/http/scaladsl/model/ws/UpgradeToWebSocket.scala
johanandren: That is just a documentation flag for users to not inherit, the bincomp checking tool does not know about it. You will need to add a filter entry to the right mima.excludes file (given that we are sure that this binary compatibility breakage is safe). You can see how to do that in the contributing guide: https://github.com/akka/akka-http/blob/master/CONTRIBUTING.md#binary-compatibility

raboof:

PR #1223, opened 2017-06-22
#1222 Correctly handle ContentType.withMissingCharset in ContentTypeRange

jypma:
This scenario was overlooked in #1141, because jm.ContentType isn't a
sealed trait (and can't be, because it is implemented in the scala DSL).

Fixes #1222.

Issue #1222, opened 2017-06-21
"scala.MatchError: application/xml" when using alpakka-s3


vvlevykin:
Versions: 
* Akka: 2.5.3
* Akka-HTTP: 10.0.8
* Alpakka-s3: 0.9

I'm trying to upload a stream using alpakka-s3 and getting an error:
```
scala.MatchError: application/xml (of class akka.http.scaladsl.model.ContentType$WithMissingCharset)
        at akka.http.scaladsl.model.ContentTypeRange.matches(ContentType.scala:17)
        at akka.http.scaladsl.unmarshalling.Unmarshaller$EnhancedFromEntityUnmarshaller$.$anonfun$forContentTypes$4(Unmarshaller.scala:112)
        at akka.http.scaladsl.unmarshalling.Unmarshaller$EnhancedFromEntityUnmarshaller$.$anonfun$forContentTypes$4$adapted(Unmarshaller.scala:112)
        at scala.collection.IndexedSeqOptimized.prefixLengthImpl(IndexedSeqOptimized.scala:37)
        at scala.collection.IndexedSeqOptimized.exists(IndexedSeqOptimized.scala:45)
        at scala.collection.IndexedSeqOptimized.exists$(IndexedSeqOptimized.scala:45)
        at scala.collection.mutable.WrappedArray.exists(WrappedArray.scala:38)
        at akka.http.scaladsl.unmarshalling.Unmarshaller$EnhancedFromEntityUnmarshaller$.$anonfun$forContentTypes$3(Unmarshaller.scala:112)
        at akka.http.scaladsl.unmarshalling.Unmarshaller$$anon$1.apply(Unmarshaller.scala:58)
        at akka.http.scaladsl.unmarshalling.Unmarshaller$EnhancedUnmarshaller$.$anonfun$mapWithInput$3(Unmarshaller.scala:91)
        at akka.http.scaladsl.unmarshalling.Unmarshaller$$anon$1.apply(Unmarshaller.scala:58)
        at akka.http.scaladsl.unmarshalling.Unmarshaller.$anonfun$transform$3(Unmarshaller.scala:23)
        at akka.http.scaladsl.unmarshalling.Unmarshaller$$anon$1.apply(Unmarshaller.scala:58)
        at akka.http.scaladsl.unmarshalling.Unmarshal.to(Unmarshal.scala:25)
        at akka.stream.alpakka.s3.impl.S3Stream.$anonfun$signAndGetAs$2(S3Stream.scala:242)
[...]

PR #1232, opened 2017-06-26
=htt #1222 improve test case to test fixed behavior

jrudolph:
Follow up for #1223 to actually test the behavior that the unmarshaller will match a `WithMissingCharset` content-type and assume UTF-8.

/cc @jypma

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/ContentType.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/ContentType.scala
@@ -11,13 +11,12 @@ import akka.http.javadsl.{ model ⇒ jm }
 import akka.http.impl.util.JavaMapping.Implicits._
 
 final case class ContentTypeRange(mediaRange: MediaRange, charsetRange: HttpCharsetRange) extends jm.ContentTypeRange with ValueRenderable {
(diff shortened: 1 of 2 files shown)


akka-http-core/src/main/scala/akka/http/scaladsl/model/ContentType.scala
```diff
@@ -11,13 +11,12 @@ import akka.http.javadsl.{ model ⇒ jm }
 import akka.http.impl.util.JavaMapping.Implicits._
 
 final case class ContentTypeRange(mediaRange: MediaRange, charsetRange: HttpCharsetRange) extends jm.ContentTypeRange with ValueRenderable {
-  def matches(contentType: jm.ContentType) = {
```
jypma: don't want to start wars here, but I've grown to _really_ like braces around anything that's multi-line, since otherwise a stray empty line would suddenly cause the rest to become constructor statements...

jrudolph: Yeah, we could reconsider how we do things. But for now I guess we should keep it consistent with the rest of the code.

ktoso:
Assigning this PR to 10.0.10 then i guess, just for tracking, no need for separate issue

Issue #1152, opened 2017-05-23
[Doc] Document how to increase default timeout of 1 second in TestKit


jlprat:
The default timeout of the Route TestKit is 1 second. On the current documentation page there is no information, nor hints on how to increase it.

Please provide documentation (and maybe an example?) on how to change the default timeout.

(AFAIK for Scala one should provide an implicit of type `RouteTestTimeout`)

jypma:
hmm, does it come from [RouteTestTimeout](https://github.com/akka/akka-http/blob/master/akka-http-testkit/src/main/scala/akka/http/scaladsl/testkit/RouteTestTimeout.scala) ?

jypma:
yeah, see [RouteTest](https://github.com/akka/akka-http/blob/master/akka-http-testkit/src/main/scala/akka/http/scaladsl/testkit/RouteTest.scala#L143). If you have an implicit `RouteTestTimeout` instance in scope when you say `~>` in your test, that's the timout that'll be used. In Scala.

jlprat:
Yeah, I'd say it comes from [RouteTestTimeout](https://github.com/akka/akka-http/blob/master/akka-http-testkit/src/main/scala/akka/http/scaladsl/testkit/RouteTestTimeout.scala)

After adding an `implicit val timeout = RouteTestTimeout(3 seconds)` the test apparently waits up to 3 seconds for the response.

PR #1238, opened 2017-06-26
=doc Explain how to increase default timeout on tests #1152

jlprat:
Issue: #1152
* Provide Snippet for Java
* Provide Snippet for Scala
* Add docs for both.

### DIFF
--- a/docs/src/main/paradox/scala/http/routing-dsl/testkit.md
+++ b/docs/src/main/paradox/scala/http/routing-dsl/testkit.md
@@ -110,6 +110,14 @@ This is what the full working test looks like:
 @@snip [TestKitFragmentSpec.scala](../../../../../test/scala/docs/http/scaladsl/server/TestKitFragmentSpec.scala) { #source-quote }
 
 
+## Accounting for Slow Test Systems
+
+Sometimes, the timeout defined for a route is finely tuned for our production environment, but our Continuous
+Integration server fails consistently as it is usually under high load or it's simply not that powerful. To account for
+these situations, timeout durations can be scaled by a given factor.
+Check the @extref[Akka Docs](akka-docs:scala/testing.html#accounting-for-slow-test-systems) for further information.
+
+
 ## Increase Timeout
 
 The default timeout when testing your routes using the testkit is 1 second. Sometimes, though, this might not be enough.
@@ -117,6 +125,7 @@ In order to extend this default timeout, to say 5 seconds, just add the followin
 
 @@snip [TestKitFragmentSpec.scala](../../../../../test/scala/docs/http/scaladsl/server/TestKitFragmentSpec.scala) { #timeout-setting }
 
+Remember to configure the timeout using `dilated` if you want to account for slow test systems.
 
 ## Examples
 
--- a/docs/src/main/paradox/scala/http/routing-dsl/testkit.md
+++ b/docs/src/main/paradox/scala/http/routing-dsl/testkit.md
@@ -110,6 +110,14 @@ This is what the full working test looks like:
 @@snip [TestKitFragmentSpec.scala](../../../../../test/scala/docs/http/scaladsl/server/TestKitFragmentSpec.scala) { #source-quote }
 
 
+## Increase Timeout
+
+The default timeout when testing your routes using the testkit is 1 second. Sometimes, though, this might not be enough.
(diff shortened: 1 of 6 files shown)


docs/src/main/paradox/scala/http/routing-dsl/testkit.md
```diff
@@ -110,6 +110,14 @@ This is what the full working test looks like:
 @@snip [TestKitFragmentSpec.scala](../../../../../test/scala/docs/http/scaladsl/server/TestKitFragmentSpec.scala) { #source-quote }
 
 
+## Increase Timeout
+
+The default timeout when testing your routes using the testkit is 1 second. Sometimes, though, this might not be enough.
+In order to extend this default timeout, to say 5 seconds, just add the following implicit in scope:
```
raboof: does akka-http also take into account the [dilation/timefactor](http://doc.akka.io/docs/akka/snapshot/scala/testing.html#accounting-for-slow-test-systems)? If so it'd be good to mention it here

jlprat: I'll take a look

jlprat: So, by default, the Scala testkit is 1 second (with dilation), when extending the timeout, one can decide to use the dilated variant or not.
However, the Java testkit, is not using any dilation and on the current implementation, it can't have it as `RouteTest` does not extend `JavaTestKit`, that is where the `dilated` method is defined.

Shall I document that the Scala testkit is using dilated by default, and that one can get advantage of it when overriding it? And then, create another issue to have `RouteTest` extend `JavaTestKit` and use `dilated` on the default timeout?

raboof: Yes, I think it would be good to add a bit of Scala-specific documentation about the possibility of dilation here, and indeed create an issue to introduce dilation for Java as well.

docs/src/main/paradox/scala/http/routing-dsl/testkit.md
```diff
@@ -110,6 +110,23 @@ This is what the full working test looks like:
...
+
+
+## Increase Timeout
+
+The default timeout when testing your routes using the testkit is 1 second. Sometimes, though, this might not be enough.
+In order to extend this default timeout, to say 5 seconds, just add the following implicit in scope:
+
+@@snip [TestKitFragmentSpec.scala](../../../../../test/scala/docs/http/scaladsl/server/TestKitFragmentSpec.scala) { #timeout-setting }
+
+Remember to configure the timeout using `dilated` if you want to account for slow test systems.
```
raboof: Perhaps add an example ('such as CI servers') and a link to http://doc.akka.io/docs/akka/current/scala/testing.html#accounting-for-slow-test-systems ?

jlprat: This link is on line 118 already :) Should I add it again? I think it might be too much. WDYT

raboof: Hmm, coffee underrun - this is perfect ;)

docs/src/main/paradox/scala/http/routing-dsl/testkit.md
```diff
@@ -110,6 +110,23 @@ This is what the full working test looks like:
 @@snip [TestKitFragmentSpec.scala](../../../../../test/scala/docs/http/scaladsl/server/TestKitFragmentSpec.scala) { #source-quote }
 
 
+## Accounting for Slow Test Systems
+
+Sometimes, the timeout defined for a route is finely tuned for our production environment, but our Continuous
```
jrudolph: Hmm why would you run tests on the production environment? And who is "us"? ;) Wouldn't it make sense to add that section to the Java side as well?

jlprat: Hmm, I'll rephrase it.
I meant, one might have its application correctly configured with the right timeouts for each route, but tests on a slow machine run slower than such timeouts...

jlprat: Regarding Java side, the default timeout is not yet using dilated, this is why I didn't add it there.

jrudolph:
Thanks a lot, @jlprat.

jlprat:
You're welcome!

PR #1244, opened 2017-06-27
Correct command for symlinks, add api/japi ones

raboof:


jrudolph:
Oops, yes, thanks for checking.

raboof:
we should really automate more of this though :)

jrudolph:
Yep, definitely, still too many steps.

PR #1246, opened 2017-06-28
Add release step to update reactive-platform-docs

raboof:


jrudolph:
Is that to add the right version number to this page: https://developer.lightbend.com/docs/reactive-platform/2.0/supported-modules/index.html

It seems it might make sense to deploy a service somewhere that knows about the latest version numbers so we don't have to touch an ever increasing amount of files all around?

ktoso:
> It seems it might make sense to deploy a service somewhere that knows about the latest version numbers so we don't have to touch an ever increasing amount of files all around?

Yeah it is slowly time to bite the bullet and just do that. Can be simple static page that just serves JSON?

raboof:
@jrudolph yes, and yes.

would be kind of neat if we could somehow use https://index.scala-lang.org/akka/akka-http for that?

jrudolph:
The best I've could come up with to get the latest version was this: http://search.maven.org/solrsearch/select?q=g:%22com.typesafe.akka%22+AND+a:%22akka-http-core_2.12%22&rows=20&wt=json which returns a json that has latest version in it (though, it is fixed to the 2.12 variant, so it will need to be changed in a few years)

jrudolph:
Created https://github.com/akka/akka-http/issues/1268 to track this feature.

PR #1249, opened 2017-06-28
Removed proxy from conf per #1227

### LINKED ISSUE #1227: Remove client proxy config from reference.conf
Until we actually implement #115 / #583 it's confusing we have some configuration for client-side HTTP proxy'ing while that feature is not actually implemented.

We should probably remove it from reference.conf until the feature is in fact there.

Scalahansolo:
Fixes #1227

### DIFF
--- a/akka-http-core/src/main/resources/reference.conf
+++ b/akka-http-core/src/main/resources/reference.conf
@@ -291,38 +291,6 @@ akka.http {
       # doesn't have to be fiddled with in most applications.
       request-header-size-hint = 512
 
-      # The proxy configurations to be used for requests with the specified
-      # scheme.
-      proxy {
-        # Proxy settings for unencrypted HTTP requests
-        # Set to 'none' to always connect directly, 'default' to use the system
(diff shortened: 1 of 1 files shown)


jlprat:
PLS BUILD

jlprat:
OK TO TEST

akka-http-core/src/main/resources/reference.conf
```diff
@@ -291,38 +291,6 @@ akka.http {
...
-        #   port = 8080
-        #   non-proxy-hosts = ["*.direct-access.net"]
-        # }
-        http = default
-
-        # Proxy settings for HTTPS requests (currently unsupported)
-        https = default
-      }
-
-      # Socket options to set for the listening socket. If a setting is left
```
jrudolph: I think these are still needed.

raboof:
PLS BUILD

Test failure might be the instability (instabilities?) tracked in #299

jrudolph:
Thanks, @Scalahansolo!

Issue #1228, opened 2017-06-23
Clarify whether https HTTP CONNECT can be used for http endpoints


raboof:
If not, it might be good to mention it in http://doc.akka.io/docs/akka-http/current/scala/http/client-side/client-transport.html#https-proxy .

jrudolph:
Technologically, it should support any kind of protocol as an HTTPS proxy shouldn't look into the payload (and you wouldn't expect to see anything with HTTPS). Proxies might still be administratively restricted by not allowing connections everywhere.

ktoso:
I think it "should" work, but I tried some proxies and none of them (free ones) allowed getting to a random http site hm (my homesite). The good news though is that we fail nicely then, saying that `failStage(new ProxyConnectionFailedException(s"The HTTPS proxy rejected to open a connection to $targetHostName:$targetPort with status code: $statusCode"))`

But yeah, in theory should work...

jrudolph:
Fixed in #1250.

PR #1250, opened 2017-06-29
Document HTTP CONNECT could support HTTP endpoints

raboof:
The HTTPS proxy client implementation that uses HTTP CONNECT could in
theory also be used for proxy'ing to HTTP endpoints, but we haven't
found an implementation that in fact allows this yet.

Refs #1228

### DIFF
--- a/docs/src/main/paradox/scala/http/client-side/client-transport.md
+++ b/docs/src/main/paradox/scala/http/client-side/client-transport.md
@@ -17,13 +17,13 @@ transport layer itself.
 ## Configuring Client Transports
 
 A `ClientTransport` is configured slightly differently for the various layers of the HTTP client.
-Right now, configuration is only possible with code (and not through config files). There's currently no 
+Right now, configuration is only possible with code (and not through config files). There's currently no
 predefined way that would allow you to select different transports per target host (but you can easily define any kind
-of strategy by implementing `ClientTransport` yourself). 
+of strategy by implementing `ClientTransport` yourself).
 
 ### Connection Pool Usage
 
-The `ConnectionPoolSettings` class allows setting a custom transport for any of the pool methods. Use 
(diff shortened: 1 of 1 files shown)


docs/src/main/paradox/scala/http/client-side/client-transport.md
```diff
@@ -37,14 +37,17 @@ method.
...
 
-A transport that connects to target servers via an HTTPS proxy. An HTTPS proxy uses the HTTP `CONNECT` method (as 
-specified in [RFC 7231 Section 4.3.6](https://tools.ietf.org/html/rfc7231#section-4.3.6)) to create tunnels to target 
+A transport that connects to target servers via an HTTPS proxy. An HTTPS proxy uses the HTTP `CONNECT` method (as
+specified in [RFC 7231 Section 4.3.6](https://tools.ietf.org/html/rfc7231#section-4.3.6)) to create tunnels to target
 servers. The proxy itself should transparently forward data to the target servers so that end-to-end encryption should
 still work (if TLS breaks, then the proxy might be fussing with your data).
 
+This approach is commonly used to securely proxy requests to HTTPS endpoints. In theory it could also be used to proxy
+requests targeting HTTP endpoints, but we have not yet found a proxy that in fact allows this.
```
jrudolph: By which means has the server prevented it? Does it also mean that most proxies don't allow connecting to arbitrary ports?

raboof: I'm not sure, @ktoso you tried a couple, right?

raboof:
We can thank Atom for the whitespace changes :)

jrudolph:
Ok, then thanks Atom ;) I think I have configured IntelliJ similarly but it does only seem to apply to Scala source files. (Also markdown has significant trailing whitespace to force newlines so I wonder if Atom's handling would change that as well...)

PR #1252, opened 2017-06-29
HttpResponseParser - add parsing for status codes without reason (#1251)

### LINKED ISSUE #1251: HttpResponseParser returns invalid status code for statuses without reason phrase
When server responds with HTTP status without reason phase then it is not parsed as status code, but only skipped, so always default statusCode value (200 OK) is returned.

mateuszgruszczynski:
Fixed #1251 and updated tests to use non-default status code

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpResponseParser.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpResponseParser.scala
@@ -108,6 +108,7 @@ private[http] class HttpResponseParser(protected val settings: ParserSettings, p
           maxResponseReasonLength + " characters")
       skipReason(startIdx)
     } else if (isNewLine(cursor + 3)) {
+      parseStatusCode()
       // Status format with no reason phrase and no trailing space accepted, diverging from the spec
       // See https://github.com/akka/akka-http/pull/989
       skipNewLine(cursor + 3)
--- a/akka-http-core/src/test/scala/akka/http/impl/engine/parsing/ResponseParserSpec.scala
+++ b/akka-http-core/src/test/scala/akka/http/impl/engine/parsing/ResponseParserSpec.scala
@@ -85,12 +85,12 @@ abstract class ResponseParserSpec(mode: String, newLine: String) extends FreeSpe
       }
 
       "a response with a missing reason phrase" in new Test {
-        s"HTTP/1.1 200 ${newLine}Content-Length: 0${newLine}${newLine}" should parseTo(HttpResponse(OK))
+        s"HTTP/1.1 404 ${newLine}Content-Length: 0${newLine}${newLine}" should parseTo(HttpResponse(NotFound))
         closeAfterResponseCompletion shouldEqual Seq(false)
       }
 
       "a response with no reason phrase and no trailing space" in new Test {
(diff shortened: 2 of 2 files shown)


jlprat:
OK TO TEST

akka-http-core/src/test/scala/akka/http/impl/engine/parsing/ResponseParserSpec.scala
```diff
@@ -85,12 +85,12 @@ abstract class ResponseParserSpec(mode: String, newLine: String) extends FreeSpe
...
 
       "a response with a missing reason phrase" in new Test {
-        s"HTTP/1.1 200 ${newLine}Content-Length: 0${newLine}${newLine}" should parseTo(HttpResponse(OK))
+        s"HTTP/1.1 404 ${newLine}Content-Length: 0${newLine}${newLine}" should parseTo(HttpResponse(NotFound))
         closeAfterResponseCompletion shouldEqual Seq(false)
       }
 
       "a response with no reason phrase and no trailing space" in new Test {
-        s"""HTTP/1.1 200${newLine}Content-Length: 0${newLine}${newLine}""".stripMargin should parseTo(HEAD, HttpResponse())
+        s"""HTTP/1.1 404${newLine}Content-Length: 0${newLine}${newLine}""".stripMargin should parseTo(HEAD, HttpResponse(NotFound))
```
ktoso: Why these changes? Seems unrelated to what the commit wants to fix.

No test for the case the PR claims to fix?

akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpResponseParser.scala
```diff
@@ -108,6 +108,7 @@ private[http] class HttpResponseParser(protected val settings: ParserSettings, p
           maxResponseReasonLength + " characters")
       skipReason(startIdx)
     } else if (isNewLine(cursor + 3)) {
+      parseStatusCode()
```
ktoso: I'm not confident this does what you think, see below comment which explains that we actually DO accept such responses (and I remember the PR actually too..)

akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpResponseParser.scala
```diff
@@ -108,6 +108,7 @@ private[http] class HttpResponseParser(protected val settings: ParserSettings, p
...
       // Status format with no reason phrase and no trailing space accepted, diverging from the spec
       // See https://github.com/akka/akka-http/pull/989
```
ktoso: AFAICS what you want is implemented and commented/explained here, what's not working about it?

mateuszgruszczynski:
Well, I could chose better name for that PR. Responses were accepted, but since `parseStatusCode()` was not executed in `else if` block, what was done as "parsing" was only skipping chars that contain status code, instead of: 
- reading that code int value
- matching it with predefined StatusCodes 
- updating statusCode variable, which is by default set to `StatusCode.OK`

so all tests that was testing `HTTP 200` status code was always false positive. If it it still not clear what was the problem please just try to run those two tests on master branch, with any code other than 200.

ktoso:
I see, that explains things, would be nice to include such context in initial PR comment :-)
Checked the flow control and can confirm this fixes the issue, thanks a ton! 👍

PR #1253, opened 2017-06-29
Add reference docs for SSE (closes #1198)

### LINKED ISSUE #1198: Add documentation for SSE API and (un)marshallers
Follow up to #1116.

hseeberger:


docs/src/main/paradox/java/http/sse-support.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Client-side usage: unmarshalling
+
+In order to unmarshal an event stream as `Source<ServerSentEvent, NotUsed>`, one has to use the
+`EventStreamUnmarshalling.fromEventStream` unmarshaller:
+
+@@snip [EventStreamMarshallingTest.java](../../../../../../akka-http-tests/src/test/java/akka/http/javadsl/unmarshalling/sse/EventStreamUnmarshallingTest.java) { #event-stream-unmarshalling-example }
+
+Notice that if you are looking for a resilient way to permanently subscribe to an event stream,
+Alpakka provides the [EventSource](http://developer.lightbend.com/docs/alpakka/current/sse.html)
```
ktoso: Very nice to connect to alpakka here :)

docs/src/main/paradox/java/http/sse-support.md
```diff
@@ -0,0 +1,55 @@
...
+connector which reconnects automatically with the id of the last seen event.
```
ktoso: good doc in general!

docs/src/test/scala/docs/http/scaladsl/ServerSentEventsExampleSpec.scala
```diff
@@ -0,0 +1,44 @@
+package docs.http.scaladsl
```
ktoso: needs copy header :)
yeah we should use your plugin for it, but had some trouble with it back then ;)

ktoso:
closes #1198

hseeberger:
Yeah ;-)

BTW, please give sbt-header 2.0 a try. It's much better than back then and should work almost out of the box.

docs/src/main/paradox/scala/http/sse-support.md
```diff
@@ -0,0 +1,57 @@
+<a id="sse-support-scala"></a>
```
raboof: I don't think id's need to be unique across java and scala versions anymore - actually might be better to have them the same because we want to eventually merge them.

We can avoid some duplication by making the markdown in the 'scala' folder support both java and scala (using paradox groups, http://developer.lightbend.com/docs/paradox/latest/features/groups.html), then the file in the 'java' folder can simply be a symlink. Not necessary for this PR though IMO, looks good!

jonas: I have a WIP branch somewhere to remove the IDs since as you point out they are not needed at the "page level".

docs/src/test/scala/docs/http/scaladsl/ServerSentEventsExampleSpec.scala
```diff
@@ -0,0 +1,44 @@
...
+
+import akka.NotUsed
+import akka.http.scaladsl.Http
+import akka.http.scaladsl.model.sse.ServerSentEvent
+import akka.http.scaladsl.server.{ Directives, Route, RoutingSpec }
+import akka.http.scaladsl.unmarshalling.Unmarshal
+import akka.stream.scaladsl.Source
+import docs.CompileOnlySpec
+import java.time.LocalTime
+import java.time.format.DateTimeFormatter.ISO_LOCAL_TIME
```
2beaucoup: should these two be moved into the included snippet?

docs/src/main/paradox/java/http/sse-support.md
```diff
@@ -0,0 +1,55 @@
+<a id="sse-support-scala"></a>
+# Server-Sent Events Support
+
+Server-Sent Events (SSE) is a lightweight and [standardized](http://www.w3.org/TR/eventsource)
+protocol for pushing notifications from a HTTP server to a HTTP client. In contrast to WebSocket,
```
hseeberger: OK, I'll fix that.

docs/src/main/paradox/java/http/sse-support.md
```diff
@@ -0,0 +1,55 @@
...
+which offers bi-directional communication, SSE only allows for one-way communication from the server
+to the client. If that's all you need, SSE has the advantages to be much simpler and to rely on HTTP
+only.
```
hseeberger: Good idea!

docs/src/test/scala/docs/http/scaladsl/ServerSentEventsExampleSpec.scala
hseeberger: Which two?

2beaucoup: the `java.time` ones

hseeberger: No, I don't think so. The important import for this use case is shown, the other imports (not just for Java Time) are "boiler plate".

2beaucoup: The copy-n-paste-ability would be higher with a few more imports moved into the snippets IMHO.

docs/src/test/scala/docs/http/scaladsl/ServerSentEventsExampleSpec.scala
```diff
@@ -0,0 +1,44 @@
...
+import scala.concurrent.duration.DurationInt
+
+final class ServerSentEventsExampleSpec extends RoutingSpec with Directives with CompileOnlySpec {
```
jrudolph: IIRC, not adding a test case will lead to leaking ActorSystems with `RoutingSpec` because of ScalaTest. So, maybe just don't extends from `RoutingSpec` and add dummy implicit definitions where needed, instead.

docs/src/test/scala/docs/http/scaladsl/ServerSentEventsExampleSpec.scala
jrudolph: I agree that's what we do in doc snippets. But then add all the imports that are not related to the test setup.

docs/src/test/scala/docs/http/scaladsl/ServerSentEventsExampleSpec.scala
ktoso: The trick is to do `"stream example" in compileOnlySpec { `, but yet, current one would lead to leaks

ktoso: Added that, should pass

docs/src/test/scala/docs/http/scaladsl/ServerSentEventsExampleSpec.scala
ktoso: I'll address this comment and merge

ktoso:
This looks good, I moved the imports into the snippet and will follow up with a PR.
Attribution will remain of course.

Thanks a ton Heiko and sorry for the delay, we're back on focusing on the usual projects and maintenance now :)

ktoso:
Will be merged as https://github.com/akka/akka-http/pull/1286

Issue #1063, opened 2017-04-27
FAILED javadsl FutureDirectivesExamplesTest.testOnCompleteWithBreaker onCompleteWithBreaker


ktoso:
```
Error Message

Unexpected status code! expected:<503 Service Unavailable> but was:<500 Internal Server Error>   Request was:      HttpRequest(HttpMethod(GET),/divide/10/0,List(),HttpEntity.Strict(none/none,ByteString()),HttpProtocol(HTTP/1.1))   Route result was: Complete(HttpResponse(500 Internal Server Error,List(),HttpEntity.Strict(text/plain; charset=UTF-8,An error occurred: / by zero),HttpProtocol(HTTP/1.1))) 
Stacktrace

java.lang.AssertionError: Unexpected status code! expected:<503 Service Unavailable> but was:<500 Internal Server Error>
  Request was:      HttpRequest(HttpMethod(GET),/divide/10/0,List(),HttpEntity.Strict(none/none,ByteString()),HttpProtocol(HTTP/1.1))
  Route result was: Complete(HttpResponse(500 Internal Server Error,List(),HttpEntity.Strict(text/plain; charset=UTF-8,An error occurred: / by zero),HttpProtocol(HTTP/1.1)))

	at akka.http.javadsl.testkit.JUnitRouteTestBase$$anon$1.reportDetails(JUnitRouteTest.scala:47)
	at akka.http.javadsl.testkit.JUnitRouteTestBase$$anon$1.assertEquals(JUnitRouteTest.scala:32)
	at akka.http.javadsl.testkit.TestRouteResult.assertEqualsKind(TestRouteResult.scala:227)
	at akka.http.javadsl.testkit.TestRouteResult.assertStatusCode(TestRouteResult.scala:136)
	at docs.http.javadsl.server.directives.FutureDirectivesExamplesTest.testOnCompleteWithBreaker(FutureDirectivesExamplesTest.java:138)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
[...]

PR #1259, opened 2017-06-30
The scaladsl route seal example was leaking an actor system

johanandren:
Which in turn would hang the entire test suite

### DIFF
--- a/docs/src/test/scala/docs/http/scaladsl/RouteSealExampleSpec.scala
+++ b/docs/src/test/scala/docs/http/scaladsl/RouteSealExampleSpec.scala
@@ -10,17 +10,20 @@ import docs.CompileOnlySpec
 
 class RouteSealExampleSpec extends RoutingSpec with Directives with CompileOnlySpec {
 
-  compileOnlySpec {
+  "seal route example" in compileOnlySpec {
     //#route-seal-example
     val route = respondWithHeader(RawHeader("special-header", "you always have this even in 404")) {
       Route.seal(
         get {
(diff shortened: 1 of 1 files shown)


johanandren:
Not sure, I didn't investigate further than looking at the other `compileOnly` style tests and saw they were different.

ktoso:
Yep, that's the thing then - very weird but I remember that used to be the thing (and why we invented the compile only thing)

ktoso:
PLS BUILD

ktoso:
Previous failure is known and unrelated https://github.com/akka/akka-http/issues/1063

ktoso:
Merging as it was causing trouble to people and this certainly is what it was supposed to be to begin with 👍

jrudolph:
Do we understand why it was leaking and hanging the test suite only now? Thanks anyway for fixing it quickly. :+1:

johanandren:
I'd expect `compileOnlySpec` to not trigger creation of the actor system even if called from the constructor like that, but I didn't dig deeper.

jrudolph:
Maybe it was that a file without any tests still runs the constructor (which creates the ActorSystem) but doesn't run the `afterAll`?

ktoso:
> Maybe it was that a file without any tests still runs the constructor (which creates the ActorSystem) but doesn't run the afterAll?

AFAIR that's exactly the case - it would not run beforeAll / afterAll hooks when no tests are detected. But we create the system not in a hook, thus the leak. 

I think we had this some time in the past somewhere else too

Issue #1257, opened 2017-06-30
Missing final boundary for empty multipart entities


arnohaase:
When rendering a multipart entity with no parts, BodyPartRenderer does not write the final boundary. This results in an entity that Akka HTTP can not parse:

```
val entity = Multipart.General
  .Strict(MediaTypes.`multipart/mixed`, Vector.empty)
  .toEntity("asdf")

val um = MultipartUnmarshallers.defaultMultipartGeneralUnmarshaller
val unmarshalled = Await.result(um(entity), 10.seconds) // triggers an exception

println (unmarshalled)
```

The suggested fix is to move renderFinalBoundary out of the if block in class BodyPartRenderer.

```
  def strict(parts: immutable.Seq[Multipart.BodyPart.Strict], boundary: String,
             partHeadersSizeHint: Int, log: LoggingAdapter): ByteString = {
    val r = new ByteStringRendering(partHeadersSizeHint)
    if (parts.nonEmpty) {
      for (part ← parts) {
        renderBoundary(r, boundary, suppressInitialCrLf = part eq parts.head)
        renderEntityContentType(r, part.entity)
        renderHeaders(r, part.headers, log)
        r ~~ part.entity.data
      }
      // <--- renderFinalBoundary removed here
    }
    renderFinalBoundary(r, boundary)   // <--- ... and moved here
    r.get
  }
```

arnohaase:
This being my first bug report - I am unsure as to the best procedure. Should I create a pull request?

raboof:
Hi Arno, Welcome!

Yes, a pull request would be great! You can read more about our contribution process in CONTRIBUTING.md - it's quite a bit of text, but don't worry, most of it is pretty light once you get the hang of it. It'd be great if you could reproduce the problem in a unit test. If you get stuck anywhere feel free to drop by on gitter at https://gitter.im/akka/dev - we're always happy to help.

arnohaase:
Thanks - I'll do that over the weekend.

jrudolph:
Fixed by #1260.

PR #1260, opened 2017-06-30
Add final boundary for multipart entities with no parts #1257

arnohaase:


ktoso:
OK TO TEST

arnohaase:
I signed the Lightbend CLA,

jrudolph:
Hi, @arnohaase, thanks for the contribution.

The change seems reasonable, though it is unclear to me if multipart entities without any part would (or should) be allowed at all? Do have any reference for that?

arnohaase:
The spec does not specifically forbid it, at least as far as I can see. But it does specify that the final boundary must always be present. https://www.w3.org/Protocols/rfc1341/7_2_Multipart.html

The problem appeared in a real-world problem where a multipart response was used to represent documents matching a search, with no parts being a valid response document.

MultipartUnmarshallers.defaultMultipartGeneralUnmarshaller can handle Multipart entities without entries  already, provided the final boundary is present.

jrudolph:
@arnohaase 

> The spec does not specifically forbid it, at least as far as I can see.

It says:

> In the case of multiple part messages, in which one or more different sets of data are combined in a single body

I understand it that "one or more" wouldn't include zero. But I guess that's only a nitpick.

I agree we should do it consistently for marshallers and unmarshallers and also for strict and streamed. Would be nice to have tests being run for strict and streamed variants. Is that something you would like to work on, @arnohaase or @raboof?

raboof:
@jrudolph I think that'd be interesting but am not planning to do it right now - perhaps we should merge this and keep that idea for later? Created https://github.com/akka/akka-http/issues/1309 to track it.

jrudolph:
Yep, sounds good.

PR #1264, opened 2017-07-03
fix small spelling errors

Silv3rcircl3:


akka-http-core/src/main/java/akka/http/javadsl/model/HttpHeader.java
```diff
@@ -26,7 +26,7 @@
     public abstract String lowercaseName();
 
     /**
-     * Returns true iff nameInLowerCase.equals(lowercaseName()).
+     * Returns true if nameInLowerCase.equals(lowercaseName()).
```
raboof: I'm afraid this was not entirely a typo: "iff" is sometimes short for "if any only if" (https://en.wikipedia.org/wiki/If_and_only_if).

Might still be nice to replace this with something less obscure - perhaps "Returns true if nameInLowerCase.equals(lowercaseName()), false otherwise"? A bit verbose though.

raboof:
OK TO TEST

raboof:
Thanks for taking the time to prepare this patch! As I noted above 'iff' actually means something, but it might be good to not use such obscure language in our API documentation.

akka-http-core/src/main/java/akka/http/javadsl/model/HttpHeader.java
Silv3rcircl3: Interesting, didn't know that 😃 
What about `Returns only true if...` ? 
I mean `Boolean` kind of implies that it returns false otherwise.

raboof: yes, or `Returns true exactly when`?

Silv3rcircl3: sounds good

ktoso:
I like the `iff` a lot :-(

Silv3rcircl3:
I have no strong opinions against `iff`, just thought it's a typo, so you could also keep it as it is and close this one ;-)

raboof:
I kind of like 'iff', though it's hard to say how clear it is (esp to non-native speakers perhaps?)... I'm OK with either merging or closing :)

jrudolph:
In my opinion we should replace it with "if and only if" which is the accurate expansion of "iff".

ktoso:
That also sounds like a good idea (to spell out `if and only if`) 👍

Silv3rcircl3:
Done ;-)

jrudolph:
Thanks ;)

PR #1270, opened 2017-07-05
Preserve the order of repeated parameters when retrieving query string as a multimap

marcospereira:
While I couldn't find any saying in RFC3986, I think this is better than using the reverse order.

See playframework/playframework#7567

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/Uri.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/Uri.scala
@@ -574,6 +574,22 @@ object Uri {
         if (q.isEmpty) map else append(map.updated(q.key, q.value), q.tail)
       append(Map.empty, this)
     }
+
+    /**
+     * Returns this query as a map where keys can have multiple values. The parameter order is
+     * preserved, so that the following query:
+     *
+     * {{{
+     *   a=1&a=2&a=3&a=4&b=1
+     * }}}
+     *
+     * Will return a map like:
+     *
+     * {{{
+     *   "a" -> List(1, 2, 3, 4),
+     *   "b" -> List(1)
+     * }}}
+     */
     def toMultiMap: Map[String, List[String]] = {
       @tailrec def append(map: Map[String, List[String]], q: Query): Map[String, List[String]] =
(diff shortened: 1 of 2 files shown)


jlprat:
OK TO TEST

akka-http-core/src/main/scala/akka/http/scaladsl/model/Uri.scala
```diff
@@ -576,7 +576,7 @@ object Uri {
     }
     def toMultiMap: Map[String, List[String]] = {
       @tailrec def append(map: Map[String, List[String]], q: Query): Map[String, List[String]] =
-        if (q.isEmpty) map else append(map.updated(q.key, q.value :: map.getOrElse(q.key, Nil)), q.tail)
+        if (q.isEmpty) map else append(map.updated(q.key, map.getOrElse(q.key, Nil) :+ q.value), q.tail)
```
jlprat: I know prepending is way more optimal than appending on `List`. I don't know if we need this performance or not.

From the Scaladoc:
> Performance
>
>Time: List has O(1) prepend and head/tail access. Most other operations are O(n) on the number of elements in the list. This includes the index-based lookup of elements, length, append and reverse.

marcospereira: Yep, I was worried about the performance cost too.

Decide to submit the PR anyway because I guess the additional cost will be relevant only if we have a large number of parameters with the same key, which is not usual.

jlprat: I agree it's not the most conventional way, but it is possible to submit an arbitrarily big number of parameters with the same key.
Let's see what the others think about it :)

Thanks for the PR, BTW!

jrudolph: The only other solution (under the constraint that we cannot change the signature of the method) would be to collect in reverse order as before and then change all values afterwards. So, I agree that it is quite unlikely that people will use it in a way where it starts to matter. If that's the case, they could still provide their own faster implementation.

marcospereira:
@jrudolph scaladoc added. Rebased against master branch too.

jrudolph:
Unrelated failure

jrudolph:
PLS BUILD

Issue #1271, opened 2017-07-05
Provide Dilated Timeouts for Java Testkit


jlprat:
As discovered in #1238, Java Testkit can't benefit of the dilated tests.
The current timeout when using Scala Testkit is 1 second (dilated) while the Java one is just 3 seconds.

In order to let Java Testkit use `dilated`, the `RouteTest` should extend `JavaTestKit` (that is where the `dilated` method is defined).

jiminhsieh:
I found `JavaTestKit` prepare to deprecate since Akka 2.5.0. It would be replace by `akka.testkit.javadsl.TestKit`. Do we still need to fix this? I also found `akka.http.javadsl.testkit.RouteTest` could not just simply extend JavaTestKit. It already extended abstract class - `akka.http.javadsl.server.AllDirectives`. Maybe I just misunderstand what your idea. Thanks!

jlprat:
That might well be (deprecation)! I guess I was checking the 2.4.x series (which Akka HTTP depends on).
Basically, what I meant is that https://github.com/akka/akka-http/blob/master/akka-http-testkit/src/main/scala/akka/http/javadsl/testkit/RouteTest.scala#L36 this should have a `dilated` timeout.

raboof:
(just to clarify, in case someone stumbles upon this ticket: while Akka HTTP indeed depends on Akka 2.4.x, it also works with Akka 2.5.x as those are [binary compatible](http://doc.akka.io/docs/akka/current/scala/common/binary-compatibility-rules.html). We plan to update the dependency as part of #1151)

jrudolph:
Fixed in #1272.

Issue #1151, opened 2017-05-22
Depend on Akka 2.5 directly, and call it 10.1.x


ktoso:
Please note that Akka HTTP 10.0.x *does* work with Akka 2.5, due to all the compatibility rules we have in place.

We do however depend on 2.4 nowadays since we want to still be compatible with it, thus do not want to leak in 2.5 only API usage.

What we should do to avoid confusion like in https://github.com/akka/akka-http/issues/1101 : 

- Depend on 2.5 directly
- *Always* test with 2.4 and 2.5 in PR validation and other jobs
- Announce it as 10.1.0 as if someone only depended on the "akka http version" then it would upgrade their Akka (actor etc), so we should bump the minor to signify this

ktoso:
Ah, the super ticket is here: https://github.com/akka/akka-http/issues/821

ktoso:
Should we decide on a specific plan here some time soon?

ktoso:
The 1.1.10 milestone does not have issues assigned; if we have any issues that we want to address before doing this jump we should assign them to this milestone and figure out a timeline for it :)

raboof:
We have discussed this further, and as Akka 2.4 is nearing EOL (this month), it makes sense to require Akka 2.5 for akka-http 10.1.x, and not to test against 2.4 anymore at all.

If we encounter any really serious (e.g. security) issues we should back-port them to 10.0.x to also make the fix available for users who are still on 10.0.x, but this should be rare.

jrudolph:
:+1: 

We'll track the next steps for 10.1.x in #1645. Closing this ticket.

PR #1272, opened 2017-07-06
Issue #1271 - Provide Dilated Timeouts for Java Testkit.

jiminhsieh:
Fix for #1271.

### DIFF
--- a/akka-http-testkit/src/main/scala/akka/http/javadsl/testkit/RouteTest.scala
+++ b/akka-http-testkit/src/main/scala/akka/http/javadsl/testkit/RouteTest.scala
@@ -22,6 +22,7 @@ import akka.http.scaladsl.server.{ ExceptionHandler, Route ⇒ ScalaRoute }
 import akka.http.scaladsl.settings.RoutingSettings
 import akka.http.scaladsl.util.FastFuture
 import akka.stream.Materializer
+import akka.testkit.TestDuration
 
 /**
  * A base class to create route tests for testing libraries. An implementation needs to provide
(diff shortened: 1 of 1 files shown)


jlprat:
OK TO TEST

akka-http-testkit/src/main/scala/akka/http/javadsl/testkit/RouteTest.scala
```diff
@@ -33,7 +34,7 @@ abstract class RouteTest extends AllDirectives with WSTestRequestBuilding {
   implicit def materializer: Materializer
   implicit def executionContext: ExecutionContextExecutor = system.dispatcher
 
-  protected def awaitDuration: FiniteDuration = 3.seconds
+  protected def awaitDuration: FiniteDuration = new JavaTestKit(system).dilated(1.seconds).asInstanceOf[FiniteDuration];
```
jlprat: I would leave the default timeout to 3 seconds.

I'm not so sure about using this class, as it is deprecated

jlprat: I guess `TestKit` is the direct substitution for it

jiminhsieh: Sure, I would keep original setting. 
ktoso mentioned
> We do however depend on 2.4 nowadays since we want to still be compatible with it, thus do not want to leak in 2.5 only API usage.
Original post - #1151

akka-http-testkit/src/main/scala/akka/http/javadsl/testkit/RouteTest.scala
```diff
@@ -33,7 +34,7 @@ abstract class RouteTest extends AllDirectives with WSTestRequestBuilding {
   implicit def materializer: Materializer
   implicit def executionContext: ExecutionContextExecutor = system.dispatcher
 
-  protected def awaitDuration: FiniteDuration = 3.seconds
+  protected def awaitDuration: FiniteDuration = new JavaTestKit(system).dilated(3.seconds).asInstanceOf[FiniteDuration];
```
jrudolph: I think we should use the default Scala testkit definition of dilated here like this:

```scala
import akka.testkit.TestDuration

protected def awaitDuration: FiniteDuration = 3.seconds.dilated
```

jiminhsieh: Thanks for your suggestion! This looks much more elegant. : )

PR #1281, opened 2017-07-16
Added SSE client timeout

### LINKED ISSUE #1279: SSE unmarshaller should allow an idle timeout
One problem with consuming data from long lived network connections that you never send data on, such as is the case when consuming SSE event streams, is that there's no way to detect if the remote end suddenly disappeared (eg, network partition), since the only way to detect that is to send data upstream and not get a TCP ack back.

Well behaving SSE servers send heartbeats to ensure proxies don't cut them off, this can be used by the client to detect disconnects, and this can be done in Akka streams by inserting an idle timeout into the flow. To do this with the Akka SSE client unmarshaller, the timeout needs to be inserted before the decoder, since the decoder ignores the heartbeats and doesn't emit anything in response to them.  So you can't just attach an idle timeout to the Source produced by the unmarshaller, it has to be the unmarshaller itself that inserts the timeout, before the decoder.  This timeout could perhaps be passed explicitly, or via an implicit timeout parameter.

jroper:
Fixes #1279

Note I made the timeout compulsory, even if a server doesn't send heartbeats, it is better to close the connection every so often and reconnect than to leak a connection when network partitions happen.

Binary and source compatability have been maintained.

### DIFF
--- a/akka-http-tests/src/test/scala/akka/http/scaladsl/unmarshalling/sse/EventStreamUnmarshallingSpec.scala
+++ b/akka-http-tests/src/test/scala/akka/http/scaladsl/unmarshalling/sse/EventStreamUnmarshallingSpec.scala
@@ -7,15 +7,21 @@ package scaladsl
 package unmarshalling
 package sse
 
-import akka.NotUsed
+import java.util.concurrent.TimeoutException
+
+import akka.{ Done, NotUsed }
 import akka.http.scaladsl.model.HttpEntity
 import akka.http.scaladsl.model.MediaTypes.`text/event-stream`
 import akka.http.scaladsl.model.sse.ServerSentEvent
 import akka.stream.scaladsl.{ Sink, Source }
 import java.util.{ List ⇒ JList }
+
+import akka.util.{ ByteString, Timeout }
 import org.scalatest.{ AsyncWordSpec, Matchers }
+
 import scala.collection.JavaConverters
 import scala.collection.immutable.Seq
+import scala.concurrent.duration._
 
 object EventStreamUnmarshallingSpec {
 
@@ -44,5 +50,28 @@ final class EventStreamUnmarshallingSpec extends AsyncWordSpec with Matchers wit
         .flatMap(_.runWith(Sink.seq))
         .map(_ shouldBe events)
     }
+
+    "timeout if the server doesn't send any heartbeats" in {
+      implicit val timeout = Timeout(500.millis)
+      Unmarshal(HttpEntity(`text/event-stream`, Source.maybe))
+        .to[Source[ServerSentEvent, NotUsed]]
+        .flatMap(_.runWith(Sink.ignore))
+        .map(_ ⇒ "failed")
+        .recover {
+          case _: TimeoutException ⇒ "succeeded"
+        }.map(_ shouldBe "succeeded")
+    }
+
+    "not timeout if the server does send heartbeats" in {
+      implicit val timeout = Timeout(500.millis)
+      Unmarshal(HttpEntity(
+        `text/event-stream`,
+        Source.tick(200.millis, 200.millis, ByteString("\r\n"))
+          .take(4)
+      )).to[Source[ServerSentEvent, NotUsed]]
+        .flatMap(_.runWith(Sink.ignore))
+        .map(_ shouldBe Done)
+    }
+
   }
 }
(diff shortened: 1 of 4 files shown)


jonas:
OK TO TEST

akka-http-tests/src/test/scala/akka/http/scaladsl/unmarshalling/sse/EventStreamUnmarshallingSpec.scala
```diff
@@ -44,5 +50,28 @@ final class EventStreamUnmarshallingSpec extends AsyncWordSpec with Matchers wit
...
+      Unmarshal(HttpEntity(`text/event-stream`, Source.maybe))
+        .to[Source[ServerSentEvent, NotUsed]]
+        .flatMap(_.runWith(Sink.ignore))
+        .map(_ ⇒ "failed")
+        .recover {
+          case _: TimeoutException ⇒ "succeeded"
+        }.map(_ shouldBe "succeeded")
+    }
+
+    "not timeout if the server doesn't does send heartbeats" in {
```
jonas: Remove `doesn't`?

jonas:
Looks like the `mimaReportBinaryIssues` check failed.

akka-http-tests/src/test/scala/akka/http/scaladsl/unmarshalling/sse/EventStreamUnmarshallingSpec.scala
jroper: :white_check_mark:

jroper:
Binary compatibility check failed because I added a new method to the unmarshaller trait. I guess it's not expected that it will be implemented outside of the project, it is marked as `@ApiMayChange` anyway, and in Scala 2.12 this isn't even a problem. I added it to the exclusions.

akka-http/src/main/scala/akka/http/javadsl/unmarshalling/sse/EventStreamUnmarshalling.scala
```diff
@@ -21,8 +22,31 @@ object EventStreamUnmarshalling {
   /**
    * Lets a `HttpEntity` with a `text/event-stream` media type be unmarshalled to a source of `ServerSentEvent`s.
    */
+  @deprecated("Use fromEventStreamWithTimeout", "10.0.10")
```
ktoso: AFAIR we also need to `@Deprecated` for Java people to know, since this is javadsl

akka-http/src/main/scala/akka/http/javadsl/unmarshalling/sse/EventStreamUnmarshalling.scala
```diff
@@ -21,8 +22,31 @@ object EventStreamUnmarshalling {
...
+      .map(_.map(_.asInstanceOf[ServerSentEvent]).asJava)
+      .asInstanceOf[Unmarshaller[HttpEntity, Source[ServerSentEvent, NotUsed]]]
+  }
+
+  /**
+   * Lets a `HttpEntity` with a `text/event-stream` media type be unmarshalled to a source of `ServerSentEvent`s.
+   *
+   * Will shutdown the connection if the server doesn't send any events or heartbeats for 60 seconds.
+   */
+  def fromEventStreamWithTimeout() = {
```
ktoso: Hmmm, we could just add the timeout to the normal method, it'd change semantics a bit but in a helpful way I guess. And leave the methods be called fromEventStream, just adding the 1 param overload

jrudolph:
PLS BUILD

jrudolph:
I think the mima filters need to be moved to the 10.0.10 excludes file. Do you want to address the comments from Konrad, @jroper? If you could do so soon, we could include this change in the new release this week.

akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/sse/EventStreamUnmarshalling.scala
```diff
@@ -48,13 +50,35 @@ trait EventStreamUnmarshalling {
...
+    Unmarshaller.strict(unmarshal).forContentTypes(`text/event-stream`)
+  }
+
+  /**
+   * Lets a `HttpEntity` with a `text/event-stream` media type be unmarshalled to a source of `ServerSentEvent`s.
+   *
+   * @param idleTimeout How long the connection should be allowed to be idle for before the connection is failed.
+   *                    Defaults to 1 minute.
+   */
+  implicit final def fromEventStreamWithTimeout(implicit idleTimeout: Timeout = Timeout(1.minute)): FromEntityUnmarshaller[Source[ServerSentEvent, NotUsed]] = {
```
jrudolph: Can we make the default timeout a config setting?

jrudolph:
It's unclear for me if or why the connection-level timeout wouldn't trigger in the given case, also the PR has been stalled for a long time so closing for now.

Issue #1290, opened 2017-07-19
Consolidate java/scala content in akka-http


richardimaoka:
Like we are doing on akka/akka#23052  "consolidate java/scala content", let's consolidate the java and scala contents in akka-http, too.

Below is the procedure described by @raboof.

The process is as follows:

* Mention in this issue which pages you're planning to merge 
  * Either the Akka maintainers or @richard-imaoka can update the list below
* Compare a file in `docs/src/main/paradox/java/...` with the same one under `docs/src/main/paradox/scala/....` They should cover the same contents, but one is for Java and the other is for Scala. The 'compare files' feature of IntelliJ IDEA is a nice tool for this.
* Update the file in the `scala` folder to contain both the Scala and Java content. You may use:
`@scala` and `@java` directives for inline text
`@@@ div { .group-scala }` for blocks
[code blocks](http://developer.lightbend.com/docs/paradox/latest/features/groups.html) become tabbed
* Replace the 'java' version of the page with a symlink:
  * Remove the old version (`git rm akka-docs/src/main/paradox/java/...`)
  * Create a symlink (`cd akka-docs/src/main/paradox/java; ln -s ../scala/... ...; cd -`)
  * Add the symlink to git (`git add akka-docs/src/main/paradox/java/...`)
Test your changes (`sbt akka-docs/paradox`, see CONTRIBUTING.md for details, check both the Java and Scala versions)

You could look at https://github.com/akka/akka/pull/23039/files as an example - it's probably best to contribute PR's on a page-per-page basis at least at first. Feel free to let us know if you have any questions/problems!

richard-imaoka:
Let me know which one you want to work on, and I'll mark your name for the file!

Under **docs/src/main/paradox/java/http/**
* [ ] client-side/client-https-support.md
* [ ] client-side/connection-level.md
* [ ] client-side/host-level.md
* [ ] client-side/index.md
* [ ] client-side/pool-overflow.md
* [ ] client-side/request-level.md
* [ ] client-side/websocket-support.md
* [ ] common/de-coding.md
* [ ] common/http-model.md
* [ ] common/index.md
* [ ] common/json-support.md
* [ ] common/marshalling.md
* [ ] common/timeouts.md
* [ ] common/unmarshalling.md
* [ ] common/uri-model.md
* [ ] configuration.md
* [ ] handling-blocking-operations-in-akka-http-routes.md
* [ ] implications-of-streaming-http-entity.md
* [ ] index.md
[...]

PR #1298, opened 2017-07-21
Merge common/timeouts.md and a few client-side pages

jonas:
Part of #1290 

- client-side/client-https-support.md
- client-side/client-transport.md
- client-side/connection-level.md
- common/timeouts.md


docs/src/main/paradox/scala/http/client-side/client-https-support.md
```diff
@@ -1,15 +1,19 @@
...
 
 The central vehicle for configuring encryption is the `HttpsConnectionContext`, which can be created using
 the static method `ConnectionContext.https` which is defined like this:
 
-@@snip [ConnectionContext.scala](../../../../../../../akka-http-core/src/main/scala/akka/http/scaladsl/ConnectionContext.scala) { #https-context-creation }
+Scala
+:  @@snip[ConnectionContext.scala](../../../../../../../akka-http-core/src/main/scala/akka/http/scaladsl/ConnectionContext.scala) { #https-context-creation }
+
+Java
+:  @@snip [ConnectionContext.scala](../../../../../../../akka-http-core/src/main/scala/akka/http/javadsl/ConnectionContext.scala) { #https-context-creation }
```
jlprat: I see that the previous version also pointed to a Scala file, but shall we do something here to not show Scala code on the Java documentation?

docs/src/main/paradox/scala/http/client-side/client-transport.md
```diff
@@ -1,9 +1,14 @@
...
-transport is represented by an instance of @scaladoc[akka.http.scaladsl.ClientTransport](akka.http.scaladsl.ClientTransport):
+transport is represented by an instance of
+@scala[@scaladoc[akka.http.scaladsl.ClientTransport](akka.http.scaladsl.ClientTransport)]@java[@javadoc[akka.http.javadsl.ClientTransport](akka.http.javadsl.ClientTransport)]:
 
-@@snip [ClientTransport.scala](../../../../../../../akka-http-core/src/main/scala/akka/http/scaladsl/ClientTransport.scala) { #client-transport-definition }
+Scala
+:  @@snip [ClientTransport.scala](../../../../../../../akka-http-core/src/main/scala/akka/http/scaladsl/ClientTransport.scala) { #client-transport-definition }
+
+Java
+:  @@snip [ClientTransport.scala](../../../../../../../akka-http-core/src/main/scala/akka/http/javadsl/ClientTransport.scala) { #client-transport-definition }
```
jlprat: I see that the previous version also pointed to a Scala file, but shall we do something here to not show Scala code on the Java documentation?

jonas: Here I don't think quoting the API is necessary since the Scaladoc and Javadoc links provides a way to dig deeper.

jonas:
Yes I noticed the inclusion of Scala code in the Java documentation. Given the size of the PR due to clean ups of cross references, I didn't want to sneak in any unrelated change.

I agree that it would be good to come up with recommendations on how to improve this. My suggestion is to avoid quoting API code and instead link to the API docs when possible. The two cases in this PR could benefit from an example instead.

jlprat:
I agree, linking to API docs seems the best solution 😊

docs/src/main/paradox/scala/http/client-side/client-transport.md
```diff
@@ -61,6 +68,12 @@ Here are some ideas for custom (or future predefined) transports:
 
 ## Usage Examples
 
-### Use HTTPS Proxy with `Http().singleRequest`
+### Use HTTPS Proxy with `Http.get(...).singleRequest`
```
jrudolph: This heading can be removed.

jrudolph:
Btw. I also think those "Java" snippets should be improved but we don't need to do that here. Let's create a ticket to follow up later.

docs/src/main/paradox/scala/http/client-side/client-transport.md
jonas: Thanks for the thorough review!

jonas:
Create #1307 to improve the Java documentation to not include Scala code.

Issue #1294, opened 2017-07-21
Add HttpApp.startServer(host, port, system)


ktoso:
Currently we require passing in the config and system if one wants to pass in system.
I just want to pass in the system, and have the default config.

Encountered this painfully in real code where I needed to start an HttpApp from a Java codebase, but the server was Scala. Creating the Scala settings object is a huge pain in Java due to the type system not managing well the apply return type of Akka Extensions..

Adding:

```
def startServer(host: String, port: Int, system: ActorSystem): Unit = {
    super.startServer(host, port, ServerSettings(system.settings.config), system)
  }
```

Will enable this use case, and is also just a missing overload I think

ktoso:
FYI @jlprat

jlprat:
I completely agree. I will take this one :) This overloaded crossed my mind as well and I thought I added an issue for it.

ktoso:
Great, thanks!

jlprat:
Closed via #1299

PR #1299, opened 2017-07-23
+htp Provide overloaded HttpApp.startServer method with system #1294

jlprat:
Issue: #1294
Add overloaded HttpApp.startServer method
Add docs for this new method
Add tests for this new method

### DIFF
--- a/akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
@@ -46,8 +46,8 @@ abstract class HttpApp extends Directives {
   /**
    * Start a server on the specified host and port, using the provided [[ActorSystem]].
    * Note that this method is blocking
-   * 
-   * @param system ActorSystem to use for starting the app, if `null` is passed in a new default ActorSystem will be created instead   
+   *
+   * @param system ActorSystem to use for starting the app, if `null` is passed in a new default ActorSystem will be created instead
    */
   def startServer(host: String, port: Int, system: ActorSystem): Unit = {
     startServer(host, port, ServerSettings(system), Option(system))
@@ -64,7 +64,7 @@ abstract class HttpApp extends Directives {
   /**
    * Start a server on the specified host and port, using the provided settings and [[ActorSystem]].
    * Note that this method is blocking.
-   * 
+   *
    * @param system ActorSystem to use for starting the app, if `null` is passed in a new default ActorSystem will be created instead
    */
   def startServer(host: String, port: Int, settings: ServerSettings, system: ActorSystem): Unit = {
@@ -74,7 +74,7 @@ abstract class HttpApp extends Directives {
   /**
    * Start a server on the specified host and port, using the provided settings and [[ActorSystem]] if present.
    * Note that this method is blocking.
-   * 
+   *
    * @param system ActorSystem to use for starting the app, if `None` is passed in a new default ActorSystem will be created instead
    */
   def startServer(host: String, port: Int, settings: ServerSettings, system: Option[ActorSystem]): Unit = {
--- a/akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
@@ -44,7 +44,15 @@ abstract class HttpApp extends Directives {
   }
 
   /**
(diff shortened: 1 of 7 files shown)


jlprat:
I was thinking to trim some of the current `startServer` methods, and just leave 1 or 2. Probably the one with `host`, `port` and maybe one with the `ActorSystem`.
The builder pattern, makes more and more sense now. And we can also offer `with` methods for the `ExceptionHandler` and `RejectionHandler` and similar.
I'll create an issue

jrudolph:
Yep, let's solve this as another issue.

jlprat:
Issue: https://github.com/akka/akka-http/issues/1300

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
```diff
@@ -44,23 +44,31 @@ abstract class HttpApp extends Directives {
   }
 
   /**
-   * Start a server on the specified host and port, using provided settings.
+   * Start a server on the specified host and port, using the provided [[ActorSystem]].
+   * Note that this method is blocking
+   */
+  def startServer(host: String, port: Int, system: ActorSystem): Unit = {
+    startServer(host, port, ServerSettings(system), Some(system))
```
ktoso: to be consistent with JavaDSL and also do what we want here, this should be `Option(system)` or we should explicitly validate it's not null

docs/src/test/scala/docs/http/scaladsl/HttpAppExampleSpec.scala
```diff
@@ -48,11 +48,63 @@ class HttpAppExampleSpec extends WordSpec with Matchers
...
+        }
+    }
+
+    // Starting the server
+    val system = ActorSystem("ownActorSystem")
+    val settings = ServerSettings(ConfigFactory.load).withVerboseErrorMessages(true)
+    WebServer.startServer("localhost", 8080, settings, system)
+    system.terminate()
+    //#with-actor-system-settings
+  }
```
ktoso: this was just moving the test I think?

ktoso:
Yeah, it's still not too crazy but I like the idea to explore the builder style here, could be good... :)

docs/src/test/scala/docs/http/scaladsl/HttpAppExampleSpec.scala
jlprat: Yes, placing it next to the other ones

akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala
jlprat: To be fair we don't validate many fields for non-nullity, but sure, I can check if it's null and call further with `None`. This should do the trick

ktoso: Yup :) that works, thanks!

ktoso:
Hmm I skimmed the other methods and they are a bit mixed in bevavoir. Javadsl will throw if null is passed in since optional.of does so, in Scala we have some Some and some Option which the first will lead to NPE later in an ugly way, and the later to silently falling back to creating a new system. Sorry to bother you with those but we should make them consistent - seems like throwing when we get a null explicitly Would be best IMHO so just like the javadsl does now

jlprat:
We could use `Optional.ofNullable` in Java, this should behave like Scala's `Option.apply`.
I will go through other usages of Some, and convert them to the right version that supports null :)

jlprat:
Amended the previous commit. Check again. Now all `Option` or `Optional` use the right version that supports null.s

jlprat:
PLS BUILD

failure was https://github.com/akka/akka-http/issues/299

jlprat:
PLS BUILD

Failure was https://github.com/akka/akka-http/issues/864

ktoso:
I documented how the parameter is handled when null is passed in so people are not surprised, LGTM thanks!

ktoso:
Grrrr... those trailing whitespace issues are driving me mad ;-)

jlprat:
Do you want me to fix those trailing whitespaces?
Thanks for the extra doc in case of Null :)

ktoso:
Probably easiest for you :) please pull, compile, push :)
Thanks and sorry for the hassle with this seemingly simple change :)

ktoso:
(ok to merge once pushed and validated)

Issue #737, opened 2017-01-09
Http2: Inbound flow control


jrudolph:
The implementation currently never sends `WINDOW_UPDATE` frames for incoming data. This will stall incoming data entity streams (uploads) after a while. Instead, we should sent back `WINDOW_UPDATE` when the user has consumed part or all of the data received for a stream so far.

We should support multiple strategies of when to update the window. It seems that in the other direction browsers seem to update after about half the window is available again. We also need to determine what a good default for the initial value would be. This is also the size of the incoming data buffer per stream.

Additionally, the per-connection window also needs to be managed in some way. A simple solution could be to define some window size and then update it together with every per-stream update.

ktoso:
I'll start on this

ktoso:
Putting it back, worked on Play things

jrudolph:
Fixed in #1301.

PR #1301, opened 2017-07-24
Http2 inbound flow control

jrudolph:
The tests pass but the code is not yet in its final shape. A few constants need to be converted to configuration values. There's currently only a simplistics strategy of sending out WINDOW_UPDATE frames. It is configured with maximum values for connection-level and stream-level windows. When outstanding + buffered data falls below half the maximum configured number it sends out a WINDOW_UPDATE frame to the peer.

I think there are a few hard flow-control problems that cannot be easily solved (but putting it in strategies should help experimenting with other ones):
 * If `max-stream-level-window * num_connections > max-connection-level-window`, there can be head-of-line blocking between streams. Since this is only for uploads, it doesn't seem to be a serious concern for now, the client can still prioritize streams in any way it likes.
 * If the [BDP](https://en.wikipedia.org/wiki/Bandwidth-delay_product) is lower than the configured connection-level window, then there will be head-of-line blocking also for control frames (stuck in TCP buffers at the peer). If the BDP is higher than the configured connection-level window, then the full bandwidth will not be used. Any better solution would at least need some way of estimating the real, current but dynamic BDP. This problem is probably as hard as any kind of TCP congestion control. One potential building block could be trying to read the current BDP as estimated by the OS (`tcp_info` structure in Linux).

Refs #737

### DIFF
--- a/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
+++ b/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
@@ -81,6 +81,11 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with GenericOut
     def handle(event: StreamFrameEvent): IncomingStreamState = event match {
       case d: DataFrame ⇒
         outlet.push(d.payload)
+
+        // FIXME: this completely removes backpressure, data will accumulate in the BufferedOutlet
+        multiplexer.pushControlFrame(WindowUpdateFrame(0, d.payload.size))
+        multiplexer.pushControlFrame(WindowUpdateFrame(d.streamId, d.payload.size))
+
         maybeFinishStream(d.endStream)
       case r: RstStreamFrame ⇒
         outlet.fail(new PeerClosedStreamException(r.streamId, r.errorCode))
--- a/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
+++ b/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
@@ -8,22 +8,29 @@ import akka.NotUsed
 import akka.annotation.InternalApi
 import akka.http.impl.engine.http2.Http2Protocol.ErrorCode
 import akka.http.scaladsl.model.http2.PeerClosedStreamException
+import akka.http.scaladsl.settings.Http2ServerSettings
 import akka.stream.scaladsl.Source
-import akka.stream.stage.{ GraphStageLogic, StageLogging }
+import akka.stream.stage.{ GraphStageLogic, OutHandler, StageLogging }
 import akka.util.ByteString
 
 import scala.collection.immutable
 
 /** INTERNAL API */
 @InternalApi
-private[http2] trait Http2StreamHandling { self: GraphStageLogic with GenericOutletSupport with StageLogging ⇒
+private[http2] trait Http2StreamHandling { self: GraphStageLogic with StageLogging ⇒
   // required API from demux
   def multiplexer: Http2Multiplexer
+  def settings: Http2ServerSettings
   def pushGOAWAY(errorCode: ErrorCode, debug: String): Unit
   def dispatchSubstream(sub: Http2SubStream): Unit
 
+  def flowController: IncomingFlowController = IncomingFlowController.default(settings)
+
   private var incomingStreams = new immutable.TreeMap[Int, IncomingStreamState]
   private var largestIncomingStreamId = 0
+  private var outstandingConnectionLevelWindow = Http2Protocol.InitialWindowSize
+  private var totalBufferedData = 0
+
   private def streamFor(streamId: Int): IncomingStreamState =
     incomingStreams.get(streamId) match {
       case Some(state) ⇒ state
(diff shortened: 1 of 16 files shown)


akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -77,32 +120,107 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with GenericOut
...
+        if (windowSizeIncrement > 0) {
+          multiplexer.pushControlFrame(WindowUpdateFrame(Http2Protocol.NoStreamId, windowSizeIncrement))
+          outstandingConnectionLevelWindow += windowSizeIncrement
+        }
+
+        buffer.onDataFrame(d)
         maybeFinishStream(d.endStream)
       case r: RstStreamFrame ⇒
-        outlet.fail(new PeerClosedStreamException(r.streamId, r.errorCode))
+        buffer.onRstStreamFrame(r)
```
ktoso: btw; love the general move to these named on* callbacks (everywhere in the infra) 👍

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -77,32 +120,107 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with GenericOut
...
+      if (streamLevel > 0 && !wasClosed /* No reason to update window on closed streams */ ) {
+        multiplexer.pushControlFrame(WindowUpdateFrame(streamId, streamLevel))
+        currentWindowSpace += streamLevel
+      }
+
+      log.debug(
+        s"adjusting con-level window by $connectionLevel, stream-level window by $streamLevel," +
+          s"remaining window space now $currentWindowSpace, buffered: ${buffer.size}," +
+          s"remaining connection window space now $outstandingConnectionLevelWindow, total buffered: $totalBufferedData")
+    }
```
ktoso: very clean, like it a lot

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -62,11 +109,7 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with GenericOut
           if (endStream) (Source.empty, HalfClosedRemote)
           else {
             val subSource = new SubSourceOutlet[ByteString](s"substream-out-$streamId")
-            (Source.fromGraph(subSource.source), Open(new BufferedOutlet[ByteString](subSource) {
-              override def onDownstreamFinish(): Unit =
-                multiplexer.pushControlFrame(RstStreamFrame(streamId, ErrorCode.CANCEL))
-              incomingStreams -= streamId
```
2m: Ooh. There was a bug here, right? Missing parenths.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -77,32 +120,107 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with GenericOut
       case x ⇒ receivedUnexpectedFrame(x)
     }
   }
-  sealed abstract class ReceivingData(outlet: BufferedOutlet[ByteString], afterEndStreamReceived: IncomingStreamState) extends IncomingStreamState { _: Product ⇒
+  class Buffer(streamId: Int, outlet: SubSourceOutlet[ByteString]) extends OutHandler {
```
jrudolph: WDYM?

ktoso: just naming of the class, "Buffer" is a bit bland, though I guess it's only used here hm... still :)

jrudolph: You mean to implement auto-to-strict in HTTP/2?

jrudolph: Ah, I see, yeah that would make sense.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
jrudolph: Yes, without that statement in the `onDownstreamFinish` we would have leaked memory for already closed streams.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
jrudolph: Renamed it and moved it around a bit.

jrudolph:
Updated. The biggest change is the introduction of a settings structure. I chose to implement the settings directly in akka-http-core and put them below `ServerSettings`. That allows us to keep `akka-http2-support` as basically implementation only. I added a stub trait `Http2InternalServerSettings` that can later be used to allow customizing internal behavior like changing those strategies.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -115,6 +132,75 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with GenericOut
...
+      }
+      if (streamLevel > 0 && !wasClosed /* No reason to update window on closed streams */ ) {
+        multiplexer.pushControlFrame(WindowUpdateFrame(streamId, streamLevel))
+        outstandingStreamWindow += streamLevel
+      }
+
+      log.debug(
+        s"adjusting con-level window by $connectionLevel, stream-level window by $streamLevel," +
+          s"remaining window space now $outstandingStreamWindow, buffered: ${buffer.size}," +
+          s"remaining connection window space now $outstandingConnectionLevelWindow, total buffered: $totalBufferedData")
```
jonas: Bikeshedding (from looking at review comments): missing space after commas when the string is constructed like this.

jrudolph:
PLS BUILD

jrudolph:
PLS BUILD

jrudolph:
Thanks for the reviews!

PR #1306, opened 2017-07-25
Fixes rendering of HttpChallenge if no realm is set

### LINKED ISSUE #1295: HttpChallenge rendering is broken if realm parameter is None
A challenge like this

```scala
HttpChallenge("Bearer", None, Map("error" ->"invalid_token", "error_description" -> "The access token expired"))
```

is rendered as 
`Bearer,error=invalid_token,error_description="The access token expired"`

Expected behaviour: The header value should be
`Bearer error=invalid_token,error_description="The access token expired"`

Realm is not required by RF7235 and even in the constructor the realm is an Option. I'm using akka-http 10.0.9 but HttpChallenge seems to be unchanged in master.

jostick:
Fixes #1295

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/HttpChallenge.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/HttpChallenge.scala
@@ -14,8 +14,15 @@ final case class HttpChallenge(scheme: String, realm: String,
 
   def render[R <: Rendering](r: R): r.type = {
     r ~~ scheme
-    if (realm != null) r ~~ " realm=" ~~#! realm
-    if (params.nonEmpty) params.foreach { case (k, v) ⇒ r ~~ ',' ~~ k ~~ '=' ~~# v }
+
+    if (realm != null) {
+      r ~~ " realm=" ~~#! realm
+      params.foreach { case (k, v) ⇒ r ~~ ',' ~~ k ~~ '=' ~~# v }
+    } else if (params.nonEmpty) {
+      r ~~ ' ' ~~ params.head._1 ~~ '=' ~~# params.head._2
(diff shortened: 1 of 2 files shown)


lightbend-cla-validator:
Hi @jostick,

jrudolph:
OK TO TEST

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/HttpChallenge.scala
```diff
@@ -14,8 +14,15 @@ final case class HttpChallenge(scheme: String, realm: String,
 
   def render[R <: Rendering](r: R): r.type = {
     r ~~ scheme
-    if (realm != null) r ~~ " realm=" ~~#! realm
-    if (params.nonEmpty) params.foreach { case (k, v) ⇒ r ~~ ',' ~~ k ~~ '=' ~~# v }
+
+    if (realm != null) {
+      r ~~ " realm=" ~~#! realm
+      params.foreach { case (k, v) ⇒ r ~~ ',' ~~ k ~~ '=' ~~# v }
```
jrudolph: A simpler solution could be to use `if (params.nonEmpty) r ~~ ","` instead of this line and then make the rest of the method unconditional.

jostick: I'm not 100% sure what you mean, but I don't think it will pass. You also have to get the whitespace between scheme and attributes right. I tried several variants, but this seems to be the easiest way to do it. Maybe I missed something...

jostick: I made a more compact fix which also seems to be easier to read. Is it better now?

Issue #1245, opened 2017-06-27
requests hang if pool is shutdown before receiving response


brharrington:
I'm seeing requests hang after a while when using a connection pool. In this case the code is using `Http().singleRequest(request)` and `onComplete` is never called for the returned future. This might be the same issue as reported in #908, but if I understand correctly they have a high request rate all the time. For the case I'm seeing, I think the problem is a race condition between shutting down an idle connection pool and a new request being made using that pool. For the connections I have seen get in this state where I have detailed logging, I see it push a request to the connection just before the pool is shutdown.

The idle-timeout is set to 30s. Normally I see a pattern of pushing, Received, Finished for a given connection. In the case where it hangs I see pushing and then "Slot was stopped". Snippet of log shown below:

```
2017-06-24T18:29:08.204 DEBUG [default-dispatcher-578] akka.http.impl.engine.client.PoolGateway: [0] </IP:35922->HOST:7001> pushing request to connection: GET /PATH Empty
2017-06-24T18:29:08.235 DEBUG [default-dispatcher-250] akka.http.impl.engine.client.PoolGateway: [0] </IP:35922->HOST:7001> Received response: GET /PATH Empty -> 200 OK Chunked
2017-06-24T18:29:08.319 DEBUG [default-dispatcher-399] akka.http.impl.engine.client.PoolGateway: [0] </IP:35922->HOST:7001> Finished reading response entity for GET /PATH Empty -> 200 OK Chunked
2017-06-24T18:29:08.403 DEBUG [default-dispatcher-506] akka.http.impl.engine.client.PoolGateway: [0] </IP:35922->HOST:7001> pushing request to connection: GET /PATH Empty
2017-06-24T18:29:08.442 DEBUG [default-dispatcher-321] akka.http.impl.engine.client.PoolGateway: [0] </IP:35922->HOST:7001> Received response: GET /PATH Empty -> 200 OK Chunked
2017-06-24T18:29:08.504 DEBUG [default-dispatcher-45] akka.http.impl.engine.client.PoolGateway: [0] </IP:35922->HOST:7001> Finished reading response entity for GET /PATH Empty -> 200 OK Chunked

... A little over 30s later ...

2017-06-24T18:29:38.691 DEBUG [default-dispatcher-397] akka.http.impl.engine.client.PoolGateway: [0] </IP:35922->HOST:7001> pushing request to connection: GET /PATH Empty
[...]

PR #1311, opened 2017-07-26
=htc #1245 fix pool idle-timeout race condition

jrudolph:
Previously, a shutdown request didn't check if there are outstanding
requests so that outstanding responses were just lost.

This is still missing tests, but I haven't been able to reproduce it reliably. The reason is that there's only a short time frame where you can hit the race, i.e. while the scheduled `gateway.shutdown` is dispatched through `PoolMasterActor` to `PoolInterfaceActor` a new request must come in.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolGateway.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolGateway.scala
@@ -53,7 +53,9 @@ private[http] final class PoolGateway(gatewayRef: ActorRef, val hcps: HostConnec
 
   /**
    * Shutdown the corresponding pool and signal its termination. If the pool is not running or is
-   * being shutting down, this does nothing,
+   * being shutting down, this does nothing.
+   *
+   * The shutdown will wait for all ongoing requests to be completed.
    *
    * @return a Future completed when the pool has been shutdown.
    */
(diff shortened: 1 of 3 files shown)


jrudolph:
> I see why adding a good reproducer will be tricky... Starting a server that's very slot in delivering the responses perhaps could do it?

No, actually that's not where the race condition is. It's between `PoolMasterActor` and `PoolInterfaceActor`. PIA schedules the idle-timeout and when it triggers it sends a `Shutdown` message to the PMA. This will then send a `Shutdown` message to the PIA. The other process is an incoming request that will be dispatched from the PMA to the PIA and which will reset the idle-timeout. If the request comes in right after the idle-timeout triggered the problem happened.

jrudolph:
Added another comment as you suggested.

jrudolph:
I have an idea for a test, let's try that.

ktoso:
Thanks for explaining the race in detail, that's indeed very tricky to reproduce :O Looking forward to see how you get around to it :)

jrudolph:
Haha, the test revealed another race condition :roll_eyes:

ktoso:
The pool is a present that keeps on giving... 🙄

jrudolph:
The test turned out to end up quite simple. It failed for me reliably with the timings given but that might be different on another machine.

jrudolph:
Fixes #1245

PR #1316, opened 2017-07-28
Reject CONNECT requests gracefully

### LINKED ISSUE #1315: (server) Better error reporting when failing to handle CONNECT requests
akka-http doesn't support `CONNECT` requests (see #101), and that's ok, but it should fail more gracefully when it receives a `CONNECT` request than it currently does, with a relevant error message. This is what I currently see in my logs when bots (presumably looking to exploit open proxy servers) attempt to make a `CONNECT` request on my server:

```
Request is missing required `Host` header: 'Host' header value of request to `www.linode.com:443` doesn't match request target authority: Host header: Some(Host: www.linode.com:443)
request target authority:
```

This is a very confusing error message - especially considering that the request is actually a valid `CONNECT` request, this is what it most likely looks like:

```
CONNECT www.linode.com:443 HTTP/1.1
Host: www.linode.com:443
```

So the problem is that Akka HTTP is failing with the wrong error message, and this could be confusing and/or alarming to users of internet facing web servers, especially given that these requests actually do have malicious intent, I think it's important to give users the right error message so that they can then determine whether they can safely ignore it.

jroper:
Fixes #1315

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
@@ -211,6 +211,8 @@ private[http] object HttpServerBluePrint {
     val rootParser = new HttpRequestParser(parserSettings, rawRequestUriHeader, HttpHeaderParser(parserSettings, log))
 
     def establishAbsoluteUri(requestOutput: RequestOutput): RequestOutput = requestOutput match {
+      case connect: RequestStart if connect.method == HttpMethods.CONNECT ⇒
+        MessageStartError(StatusCodes.BadRequest, ErrorInfo(s"CONNECT requests are not supported", s"Rejecting CONNECT request to '${connect.uri}'"))
       case start: RequestStart ⇒
         try {
           val effectiveUri = HttpRequest.effectiveUri(start.uri, start.headers, isSecureConnection, defaultHostHeader)
--- a/akka-http-core/src/test/scala/akka/http/impl/engine/server/HttpServerSpec.scala
+++ b/akka-http-core/src/test/scala/akka/http/impl/engine/server/HttpServerSpec.scala
@@ -1267,6 +1267,29 @@ class HttpServerSpec extends AkkaSpec(
         expectRequest().mapEntity(nameDataSource("foo")).expectDefaultEntityWithSizeError(limit = 10, actualSize = 11)
       }
     }
+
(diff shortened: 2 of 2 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -211,6 +211,9 @@ private[http] object HttpServerBluePrint {
     val rootParser = new HttpRequestParser(parserSettings, rawRequestUriHeader, HttpHeaderParser(parserSettings, log))
 
     def establishAbsoluteUri(requestOutput: RequestOutput): RequestOutput = requestOutput match {
+      case connect: RequestStart if connect.method == HttpMethods.CONNECT ⇒
+        // Well formed CONNECT requests typically fail in the code below anyway, so reject them gracefully
```
ktoso: no need for the comment, error message explains what's going on

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -211,6 +211,9 @@ private[http] object HttpServerBluePrint {
...
+        MessageStartError(StatusCodes.BadRequest, ErrorInfo(s"CONNECT requests are not supported", s"Rejecting CONNECT request to '${connect.uri}'"))
```
ktoso: good 👍

jroper:
By the way, it may seem better to reply with a `405 Method Not Allowed` status, but a valid response according to the HTTP spec:

> The response MUST include an Allow header containing a list of valid methods for the requested resource.

In this context, there are no allowable methods for the resource, the error code is meant to semantically tell the client how to fix the request but there's no semantic way to fix it, so it doesn't make sense to use that error code.

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
jroper: Removed.

ktoso:
Right, so BadRequest is good here 👍  :)

jlprat:
OK TO TEST

ktoso:
Oh weird, I would have thought we whitelisted James hm... Will fix

ktoso:
PLS WHITELIST

Issue #1319, opened 2017-07-31
[Failed] HttpsProxyGraphStageSpec .A ProxyGraphStage should send CONNECT message with ProxyAuth header and then forward incoming messages


jlprat:
Jenkins output:
https://jenkins.akka.io:8498/job/pr-validator-akka-http/1384/testReport/junit/akka.http.impl.engine.client/HttpsProxyGraphStageSpec/A_ProxyGraphStage_should_send_CONNECT_message_with_ProxyAuth_header_and_then_forward_incoming_messages/

ktoso:
Seems like a bug in test where I merged something today.
Thanks! Tests were good so I wonder what happened (I've run all tests hm). 
Will fix this

PR #1320, opened 2017-08-01
=htc #1319 fix missing \r\n (to get double) in HttpProxy stage CONNECT

ktoso:
This was my mistake while amending polishing up the Auth support in the stage.
Double \r\n is ofc needed at end of http message.

Resolves https://github.com/akka/akka-http/issues/1319


ktoso:
(Tests do catch this, though I missed they were in core and I executed only http ones)

ktoso:
Merging as this breaks validation

PR #1321, opened 2017-08-03
Update marshalling.md

mb720:
Grammar, wording, typo.

### DIFF
--- a/docs/src/main/paradox/scala/http/common/marshalling.md
+++ b/docs/src/main/paradox/scala/http/common/marshalling.md
@@ -1,9 +1,9 @@
 # Marshalling
 
-"Marshalling" is the process of converting a higher-level (object) structure into some kind of lower-level
-representation, often a "wire format". Other popular names for it are "Serialization" or "Pickling".
+Marshalling is the process of converting a higher-level (object) structure into some kind of lower-level
+representation, often a "wire format". Other popular names for marshalling are "serialization" or "pickling".
 
-In Akka HTTP "Marshalling" means the conversion of an object of type `T` into a lower-level target type,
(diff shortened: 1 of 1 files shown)


jlprat:
OK TO TEST

jlprat:
Thanks for the PR!!

docs/src/main/paradox/scala/http/common/marshalling.md
```diff
@@ -1,9 +1,9 @@
 # Marshalling
 
-"Marshalling" is the process of converting a higher-level (object) structure into some kind of lower-level
-representation, often a "wire format". Other popular names for it are "Serialization" or "Pickling".
+Marshalling is the process of converting a higher-level (object) structure into some kind of lower-level
+representation, often a "wire format". Other popular names for marshalling are "serialization" or "pickling".
```
jlprat: I'm not so sure about not capitalizing the quoted serialization and pickling. But maybe is the German influence that biases me.

mb720: I have to fight German bias too :-)
But I'd argue that according to [the rules of capitalization](https://en.wikipedia.org/wiki/Capitalization_in_English) those words should be lowercase.

jlprat: Yes, I think your version is right.

mb720:
You're welcome!
:-)

mb720:
Whoops. Closed prematurely.

jlprat:
Usually, we need two approvals to merge, but now being summertime, things are little slower. I'll wait a day or two, just in case somebody else wants to review it, then I'll merge.

PR #1339, opened 2017-08-08
=doc Consolidate client host level docs #1290

jlprat:
Issue: #1290
Update Java signature to `Pair`


docs/src/main/paradox/scala/http/client-side/host-level.md
```diff
@@ -1,6 +1,6 @@
 # Host-Level Client-Side API
 
-As opposed to the @ref[Connection-Level Client-Side API](connection-level.md#connection-level-api) the host-level API relieves you from manually managing individual HTTP
+As opposed to the @ref[Connection-Level Client-Side API](connection-level.md) the host-level API relieves you from manually managing individual HTTP
```
jonas: 👍, 5bcbd7552128881d4e665f045eea2987bf11c28d didn't remove all use of the `connection-level-api{,-java}` anchor.

docs/src/main/paradox/scala/http/client-side/host-level.md
```diff
@@ -56,13 +57,20 @@ the same pooled connections.
...
 ```
+@@@
+@@@ div { .group-java }
+```java
+Flow<Pair<HttpRequest, T>, Pair<Try<HttpResponse>, T>, HostConnectionPool>
+```
+@@@
 
-This means it consumes tuples of type `(HttpRequest, T)` and produces tuples of type `(Try[HttpResponse], T)`
+This means it consumes tuples of type @scala[`(HttpRequest, T)`]@java[`Pair<HttpRequest, T>`] and produces tuples of type @scala[`(Try[HttpResponse], T)`]@java[`Pair<Try<HttpResponse>, T>`]
```
jonas: IMO, "pairs" sounds better here than "tuples", but might be general enough.

johanandren: I agree, pairs is better, but I think tuples is good enough as well, so do as you please.

jlprat: I'll replace with pairs, as 2 people think the same way, and the same thought crossed my mind

PR #1344, opened 2017-08-11
Fix nightly build against Akka master

jlprat:
It failed because of the Actor for this example could not be compiled anymore. Now it is inlined again


jlprat:
My bad, I removed the wrong file, I don't know how it happened. Fixing this

jlprat:
I deleted the wrong file by mistake. Now should be good to go

jrudolph:
Thanks @jlprat.

jrudolph:
I reran nightlies manually and they are fixed again.

jlprat:
I moved back to a file the Akka Actor example you inlined previously and that caused the nightlies fail :(

PR #1347, opened 2017-08-14
=doc #1290 Merge server-side-https-support.md

jonas:
References #1290


docs/src/main/paradox/java/http/server-side/server-https-support.md
```diff
@@ -0,0 +1 @@
+../../../scala/http/server-side/server-https-support.md
```
jonas: II would even suggest to drop the `server-` and `client-` prefixes from the name on both the server-side and client-side pages.

docs/src/main/paradox/scala/http/server-side/server-https-support.md
```diff
@@ -1,20 +1,24 @@
 # Server-Side HTTPS Support
```
jlprat: Github detects this page as a moved from `java/http/server-side-https-support.md` to the `scala/http/server-side-https-support.md` and not as a symlink. But looking at the files themselves seems good.

docs/src/main/paradox/scala/http/server-side/server-https-support.md
```diff
@@ -47,37 +55,63 @@ keystores using the JDK keytool utility can be found [here](https://docs.oracle.
...
-or via [SSL-Config](#ssl-config-java) (not explained here though).
+Java
+:  @@snip [SimpleServerApp.java](../../../../../../../akka-http-tests/src/main/java/akka/http/javadsl/server/examples/simple/SimpleServerApp.java) { #https-http-config }
+
+or via [SSL-Config](#ssl-config) (not explained here though).
 
 Then, call `bind...` methods twice like below.
-The blow `SimpleServerApp.useHttps(system)` is calling the above defined HTTP `public static HttpsConnectionContext useHttps(ActorSystem system)` method.
+@scala[The passed `https` context is from the above code snippet.]
+@java[The blow `SimpleServerApp.useHttps(system)` is calling the above defined HTTP `public static HttpsConnectionContext useHttps(ActorSystem system)` method.]
```
jlprat: Sorry for the question, but what does blow mean? Is it just misspelled "below"?

docs/src/main/paradox/scala/http/server-side/server-https-support.md
jonas: I think it is because of the path or my preference for taking the Java wordings.

docs/src/main/paradox/scala/http/server-side/server-https-support.md
jonas: Oh, yes, I forgot to add a note about this one. "Below" sounds right, but we can reword to: 

> `SimpleServerApp.useHttps(system)` is calling the above defined ...

jonas: Done in bf4dbb

docs/src/main/paradox/java/http/server-side/server-https-support.md
raboof: While I like this change (and agree dropping `server-` makes sense), this will break external links to these pages in the documentation. One way to fix that might be to maintain a `.htaccess` file with redirects in our documentation codebase (https://github.com/lightbend/paradox/issues/131)

docs/src/main/paradox/scala/http/server-side/server-https-support.md
```diff
@@ -1,22 +1,22 @@
...
 
-Akka HTTP supports TLS encryption on the server-side as well as on the @ref[client-side](client-side/client-https-support.md).
+Akka HTTP supports TLS encryption on the server-side as well as on the @ref[client-side](../client-side/client-https-support.md).
 
 The central vehicle for configuring encryption is the `HttpsConnectionContext`, which can be created using
 the static method `ConnectionContext.https` which is defined like this:
 
 Scala
-:  @@snip [ConnectionContext.scala](../../../../../../akka-http-core/src/main/scala/akka/http/scaladsl/ConnectionContext.scala) { #https-context-creation }
+:  @@snip [ConnectionContext.scala](../../../../../../../akka-http-core/src/main/scala/akka/http/scaladsl/ConnectionContext.scala) { #https-context-creation }
```
raboof: We could introduce a variable here like we did for akka at https://github.com/akka/akka/blob/master/akka-docs/build.sbt#L45

docs/src/main/paradox/java/http/server-side/server-https-support.md
raboof: Looks like this is actually simple: we can just drop a `.htaccess` in `src/main/paradox` with something like:

```
RedirectMatch 301 ^(.*)/server-side-https-support.html$ $1/server-side/https-support.html
```

and that appears to work

jonas: Good idea, I've added the `.htaccess` file.

docs/src/main/paradox/scala/http/server-side/server-https-support.md
jonas: Yes, I agree that a lot of paths are getting rather insane. I suggest we do that in a separate PR and preferable wait until the current pending doc consolidation PRs have been merged.

raboof: Doing that in a separate PR makes a lot of sense :+1:

jonas:
Rebased, will merge if Jenkins agrees.

PR #1334, opened 2017-08-07
=doc Consolidate configuration doc #1290

jlprat:
Issue: #1290

PR #1350, opened 2017-08-14
=doc #1290 Link client-side docs at the directory level

jonas:
After @jlprat 's hard work to unify the client-side docs let's reduce the link clutter.
References #1290


jonas:
Pending merge of #1334 ...

Issue #1358, opened 2017-08-15
WebSocket server never send back close connection 


buster84:
Hi,
I found a problem that WebSocket server never send back WebSocket Close Connection when server output is infinite. Some WS libraries can't close connection because of this problem(At least [SocketRocket](https://github.com/facebook/SocketRocket) and [jetfire](https://github.com/acmacalister/jetfire) can't close connection). 

I checked [InfinitOutputServer.scala](https://gist.github.com/buster84/11f0e86bef7e1306394523eb81c2671e#file-infinitoutputserver-scala) never send back Close connection and [InfinitOutputServer.js](https://gist.github.com/buster84/11f0e86bef7e1306394523eb81c2671e#file-infinitoutputserver-js) do send back close connection using wireshark. 

InfinitOutputServer.scala result is:
![screen shot 2017-08-15 at 18 45 23](https://user-images.githubusercontent.com/1047397/29340907-7adb088e-825c-11e7-9506-661f8dda04fc.png)

And InfinitOutputServer.js result is:
![screen shot 2017-08-15 at 18 46 17](https://user-images.githubusercontent.com/1047397/29340928-8ff6f930-825c-11e7-9f83-4d9fdf87f132.png)

[rfc6455](https://tools.ietf.org/html/rfc6455#section-5.5.1) says:

> If an endpoint receives a Close frame and did not previously send a
   Close frame, the endpoint MUST send a Close frame in response.  (When
   sending a Close frame in response, the endpoint typically echos the
   status code it received.)  It SHOULD do so as soon as practical.  An
   endpoint MAY delay sending a Close frame until its current message is
   sent (for instance, if the majority of a fragmented message is
   already sent, an endpoint MAY send the remaining fragments before
   sending a Close frame).  However, there is no guarantee that the
   endpoint that has already sent a Close frame will continue to process
   data.

I guess current behavior breaks RFC because of `the endpoint MUST send a Close frame in response.`. 

Problematic code is [here](https://github.com/akka/akka-http/blob/master/akka-http-core/src/main/scala/akka/http/impl/engine/ws/FrameOutHandler.scala#L45-L48)

raboof:
[...]

PR #1360, opened 2017-08-16
Fix not sending back WebSocket Close Connection bug(#1358)

buster84:
This will fix this issue. https://github.com/akka/akka-http/issues/1358
This behavior will be sending back WebSocket Close Connection immediately either normal or abnormal cases instead of waiting for user handler closed.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/ws/FrameOutHandler.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/ws/FrameOutHandler.scala
@@ -42,10 +42,6 @@ private[http] class FrameOutHandler(serverSide: Boolean, _closeTimeout: FiniteDu
         grab(in) match {
           case start: FrameStart   ⇒ push(out, start)
           case DirectAnswer(frame) ⇒ push(out, frame)
-          case PeerClosed(code, reason) if !code.exists(Protocol.CloseCodes.isError) ⇒
-            // let user complete it, FIXME: maybe make configurable? immediately, or timeout
-            setHandler(in, new WaitingForUserHandlerClosed(FrameEvent.closeFrame(code.getOrElse(Protocol.CloseCodes.Regular), reason)))
-            pull(in)
           case PeerClosed(code, reason) ⇒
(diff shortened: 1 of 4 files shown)


jlprat:
OK TO TEST

jrudolph:
Thanks, @buster84 for the proposal. As I mentioned somewhere else, the current behavior was intended because the spec seemed to allow keeping open the connection for a while. We didn't want to make that call for the user so we allowed the user to keep sending in the still-open direction.

I agree that it probably is not the best default but I'm a bit worried about changing the behavior right now. Do we have an idea whether and how existing user code could already depend on the existing behavior??

buster84:
Thank @jrudolph for responding. In client perspective, this behavior is weird because I feel no client want to wait fore new message coming after sending close connection. But this is who-know thing. So I understand your concern. But we should change the default behavior eventually because of four reasons:
1. Most of web socket cases are infinite output( otherwise people just use standard http to sending back in one response )
2. Potentially client may make a lot of connection. This cause memory leak and making a lot of network traffic.
3. SocketRocket, which is very popular web socket client in iOS, can't close connection.
4. Default behavior may be against RFC. 

Again, I agree with that changing default behavior may affect existing user code so I guess we can change default behavior step by step like following way:
1) Allow user to change behavior but default behavior is same as current and deprecated.
2) Change default behavior in next next version. 

So somehow we need to allow user to change the behavior if we want to change default behavior step by step. I guess there are two options to achieve it:
1) Enable to write configuration file to change behavior. Maybe options are "waiting for output source finishing", "Timeout" or "immediately". 
2) Enable to write call back function to change behavior. 

I am okay to close this PR and send other PR if I can( I am not so much familiar with Akka stream actually ). Let me know your thought @jrudolph.

raboof:
I think we can close this one because in #1358 we gravitated towards recommending users to use `Flow.fromSinkAndSourceCoupled` to achieve this result when desired instead.

jrudolph:
After re-reading the spec a while ago I think I somewhat reversed my opinion. We should probably support the reasonable behavior to close automatically (after a while?) but for now under a config flag.

PR #1361, opened 2017-08-16
=doc #1290 Merge routing-dsl/rejections.md

jonas:
References #1290


docs/src/main/paradox/java/http/routing-dsl/rejections.md
```diff
@@ -1,137 +0,0 @@
-# Rejections
```
ktoso: But that will contain Scala tests/examples and not contain the Java ones?
We need to update the scala file to include the java examples before this can be merged

jonas: Somehow GitHub screwed up this PR. It doesn't show the Scala file diff.

Issue #1342, opened 2017-08-09
[Doc] Rename de-coding.md to encoding.md


jlprat:
After https://github.com/akka/akka-http/pull/1338 is merged, rename the doc page `de-coding.md` to `encoding.md` as the hyphenated name is not used at all on Akka HTTP code base.

jlprat:
Closed via https://github.com/akka/akka-http/commit/32565921738c433aec911b5c941737f204e804b3

PR #1369, opened 2017-08-19
=doc #1290,#1342 Rename pages

jonas:
References #1290, #1342


jonas:
Once this is merged I have a follow-up to create a server-side index page.

jlprat:
Thanks!!

PR #1372, opened 2017-08-21
=htc #864 fix race condition in ResponseRendererSpec

### LINKED ISSUE #864: FAILED: ResponseRendererSpec -> The response preparation logic should properly render The 'Connection' header should be rendered correctly

jrudolph:
The idea is to provide a stage / API that allows to find out the result
of running elements through a flow. The race condition previously was that
finding out whether the flow was completed could only be observed by waiting
for some amount of time.

This adds a new test utility stage that injects an event into the mailbox
of the ActorGraphInterpeter (via AsyncCallback) which is hoped to be only
processed after all internal events have been processed. This seems to work
reliably as long as the complete flow is fused and synchronous and the whole
processing needs less events than the event queue processes in one go.

Fixes #864

### DIFF
--- a/akka-http-core/src/test/scala/akka/http/impl/util/CollectionStage.scala
+++ b/akka-http-core/src/test/scala/akka/http/impl/util/CollectionStage.scala
@@ -1,0 +1,127 @@
+/*
+ * Copyright (C) 2009-2017 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.http.impl.util
+
+import java.util.concurrent.atomic.AtomicReference
+
+import akka.annotation.InternalApi
+import akka.stream.scaladsl.{ Flow, Source }
+import akka.stream.{ Attributes, Inlet, Materializer, SinkShape }
+import akka.stream.stage.{ GraphStageLogic, GraphStageWithMaterializedValue, InHandler }
+
+import scala.annotation.tailrec
+import scala.collection.immutable.VectorBuilder
+import scala.concurrent.{ Future, Promise }
+import scala.util.{ Failure, Success, Try }
+
+@InternalApi
+sealed private[akka] trait Collect[T] {
+  def collectAndCompleteNow(): Future[(Seq[T], Boolean)]
+}
+
+@InternalApi private[akka] object CollectorStage {
+  def resultAfterSourceElements[T, U](source: Source[T, Any], flow: Flow[T, U, Any])(implicit materializer: Materializer): Future[(Seq[U], Boolean)] = {
+    val collector =
+      (source ++ Source.maybe[T] /* Never complete*/ )
+        .via(flow)
+        .runWith(new CollectorStage[U])
+    collector.collectAndCompleteNow()
+  }
+}
+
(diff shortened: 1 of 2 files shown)


jrudolph:
PLS BUILD

akka-http-core/src/test/scala/akka/http/impl/util/CollectionStage.scala
```diff
@@ -0,0 +1,127 @@
...
+  sealed trait State
+  case object Uninitialized extends State
+  case class Scheduled(callback: Promise[(Seq[T], Boolean)]) extends State
+  case object Initialized extends State
+  case class Completed(elements: Try[(Seq[T], Boolean)]) extends State
+
+  def createLogicAndMaterializedValue(inheritedAttributes: Attributes): (GraphStageLogic, Collect[T]) = {
+    val logic = new GraphStageLogic(shape) with Collect[T] with InHandler {
+      var collectedElements = new VectorBuilder[T]()
+      @volatile var state = new AtomicReference[State](Uninitialized)
```
jonas: Does this need to be a `var` to trigger a special memory access? I didn't find any assignments to this reference.

jrudolph:
Thanks, good point, that var is left over from a previous version.

Am 21.08.2017 4:52 nachm. schrieb "Jonas Fonseca" <notifications@github.com

jrudolph:
PLS BUILD

jrudolph:
PLS BUILD

jrudolph:
Something is still broken but the failures don't reproduce on my machine.

raboof:
`ResponseRendererSpec` fails with timeouts fairly reliably for me when I run `test` in `akka-http-core`, but not when I run `testOnly *ResponseRendererSpec`

jrudolph:
Ah, thanks good suggestion. I'll try that.

jrudolph:
The new infrastructure had indeed a small race condition. Should be fixed now :crossed_fingers:.

jrudolph:
PLS BUILD

akka-http-core/src/test/scala/akka/http/impl/util/CollectionStage.scala
```diff
@@ -0,0 +1,127 @@
...
+      }
+
+      def collectAndCompleteNow(): Future[(Seq[T], Boolean)] = {
+        val p = Promise[(Seq[T], Boolean)]()
+
+        @tailrec def handleState(): Unit =
+          state.get() match {
+            case Initialized ⇒
+              if (!state.compareAndSet(Initialized, Scheduled(p))) handleState()
+              else collectNowCallback.invoke(p)
```
jrudolph: Had to move the callback invocation here so that it cannot overtake the state change (or be even invoked multiple times).

PR #1373, opened 2017-08-21
Only check whitesource when a token is available

raboof:


jrudolph:
PLS BUILD

jrudolph:
(I guess previous trigger / check was lost because of github outage/slowness yesterday)

raboof:
(logged the failed build as #1376)

PR #1374, opened 2017-08-21
Document mutual authentication for server-side HTTPS

raboof:


docs/src/main/paradox/scala/http/server-side/server-https-support.md
```diff
@@ -113,6 +113,16 @@ Scala
 Java
 :  @@snip [SimpleServerHttpHttpsApp.java](../../../../../../../akka-http-tests/src/main/java/akka/http/javadsl/server/examples/simple/SimpleServerHttpHttpsApp.java) { #both-https-and-http }
 
+## Mutual authentication
+
+To require clients to authenticate themselves when connecting, pass in `Some(Need)` as the `clientAuth` parameter of the
```
jlprat: `Some(need)` is too Scala for the Java API. Something like `Optional.of(need)` might be the appropriate one

PR #1383, opened 2017-08-24
Use 'absolute' paths for code snippets

raboof:
Unfortunately the same approach doesn't appear to work for `@@signature`?

For updating the docs I used:
```
find docs -type f -name "*.md" | xargs sed -i -e "s/\(@@snip[^\(]*(\)\(\.\.\/\)*akka-/\1\$akka-http\$\/akka-/"
find docs -type f -name "*.md" | xargs sed -i -e "s/\(@@snip[^\(]*(\)\(\.\.\/\)*test/\1\$code\$/"
```


build.sbt
```diff
@@ -175,7 +175,9 @@ lazy val docs = project("docs")
       "algolia.docsearch.index_name" -> "akka-http",
       "google.analytics.account" -> "UA-21117439-1",
       "google.analytics.domain.name" -> "akka.io",
-      "github.base_url" -> GitHub.url(version.value)
+      "github.base_url" -> GitHub.url(version.value),
+      "snip.code.base_dir" -> (sourceDirectory in Test).value.getAbsolutePath,
+      "snip.akka-http.base_dir" -> ((baseDirectory in Test).value / "..").getAbsolutePath
```
jonas: Maybe `$docs$` or `$test$` instead of `$code$`.

`baseDirectory in ThisBuild` should resolve to the project root, however, Paradox already defines `$root$`:

> Also, paradox always auto-defines the placeholder $root$ to denote the absolute path of the SBT project’s root directory.

raboof: Unfortunately, `$root$` points to the root of the subproject, not of the root project, so it's different from $akka-http$.

`$docs$` doesn't make too much sense to me as this is all related to the documentation, but `$test$` is also fine, I'll change it to that.

ktoso:
Actually, this seems to fail paradox:

`[error] (docs/compile:paradoxMarkdownToHtml) com.lightbend.paradox.markdown.Url$Error: property [snip.test.base_dir] is not defined`

ktoso:
I'm having a really slow morning or something... did not `reload`.
This works and seems fine.

Issue #1377, opened 2017-08-22
Provide support for Forwarded Header


jlprat:
Currently, Akka HTTP doesn't support the standard header `Forwarded`, but only one of the de-facto standard `X-Forwarded-For` header that is only part of the standard `Forwarded`.

It would be good to add support for `Forwarded` header as described in https://tools.ietf.org/html/rfc7239

Additionally, it might be also good to implement `X-Forwarded-Proto` and `X-Forwarded-Host`:
* https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host
* https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto

Headers added
- [x] X-Forwarded-Proto and X-Forwarded-Host
- [ ] Forwarded

jrudolph:
:+1

jlprat:
I will take this one, as we need it at work :)

idzivinskyi:
👍 I'm looking forward to this. I spent few days to understand that `akka-http` does not support `Forwarded` headers, especially `X-Forwarded-Proto` header.

jlprat:
PR to add support for `X-Forwarded-Proto` and `X-Forwarded-Host` is here: https://github.com/akka/akka-http/pull/1387

Adding support for `Forwarded` will come later.

jlprat:
Headers added
- [ ] X-Forwarded-Proto and X-Forwarded-Host
- [ ] Forwarded

PR #1387, opened 2017-08-28
+htc Add X-Forwarded-Proto and X-Forwarded-Host #1377

jlprat:
Refs: #1377

This PR adds support for the _de facto_ standard headers:
- `X-Forwarded-Host`
- `X-Forwarded-Proto`

### DIFF
--- a/akka-http-core/src/main/java/akka/http/javadsl/model/headers/XForwardedHost.java
+++ b/akka-http-core/src/main/java/akka/http/javadsl/model/headers/XForwardedHost.java
@@ -5,7 +5,7 @@
 package akka.http.javadsl.model.headers;
 
 /**
- *  Model for the `X-Forwarded-Proto` header.
+ *  Model for the `X-Forwarded-Host` header.
  *  Specification: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host
  */
 public abstract class XForwardedHost extends akka.http.scaladsl.model.HttpHeader {
--- a/akka-http-core/src/main/java/akka/http/javadsl/model/headers/XForwardedHost.java
+++ b/akka-http-core/src/main/java/akka/http/javadsl/model/headers/XForwardedHost.java
@@ -9,9 +9,9 @@ package akka.http.javadsl.model.headers;
  *  Specification: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host
  */
 public abstract class XForwardedHost extends akka.http.scaladsl.model.HttpHeader {
-    public abstract akka.http.javadsl.model.RemoteAddress getAddress();
+    public abstract akka.http.javadsl.model.Host getAddress();
 
-    public static XForwardedHost create(akka.http.javadsl.model.RemoteAddress address) {
-        return new akka.http.scaladsl.model.headers.X$minusForwarded$minusHost(((akka.http.scaladsl.model.RemoteAddress) address));
+    public static XForwardedHost create(akka.http.javadsl.model.Host host) {
+        return new akka.http.scaladsl.model.headers.X$minusForwarded$minusHost(((akka.http.scaladsl.model.Uri.Host) host));
     }
 }
--- a/akka-http-core/src/main/java/akka/http/javadsl/model/headers/XForwardedHost.java
+++ b/akka-http-core/src/main/java/akka/http/javadsl/model/headers/XForwardedHost.java
@@ -1,0 +1,17 @@
+/**
+ * Copyright (C) 2009-2017 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.http.javadsl.model.headers;
+
+/**
+ *  Model for the `X-Forwarded-Proto` header.
+ *  Specification: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host
+ */
+public abstract class XForwardedHost extends akka.http.scaladsl.model.HttpHeader {
+    public abstract akka.http.javadsl.model.RemoteAddress getAddress();
+
+    public static XForwardedHost create(akka.http.javadsl.model.RemoteAddress address) {
+        return new akka.http.scaladsl.model.headers.X$minusForwarded$minusHost(((akka.http.scaladsl.model.RemoteAddress) address));
+    }
+}
--- a/akka-http-core/src/main/java/akka/http/javadsl/model/headers/XForwardedHost.java
+++ b/akka-http-core/src/main/java/akka/http/javadsl/model/headers/XForwardedHost.java
@@ -9,7 +9,7 @@ package akka.http.javadsl.model.headers;
  *  Specification: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host
  */
 public abstract class XForwardedHost extends akka.http.scaladsl.model.HttpHeader {
-    public abstract akka.http.javadsl.model.Host getAddress();
+    public abstract akka.http.javadsl.model.Host getHost();
 
     public static XForwardedHost create(akka.http.javadsl.model.Host host) {
         return new akka.http.scaladsl.model.headers.X$minusForwarded$minusHost(((akka.http.scaladsl.model.Uri.Host) host));
--- a/akka-http-core/src/main/scala/akka/http/impl/model/parser/SimpleHeaders.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/model/parser/SimpleHeaders.scala
@@ -230,10 +230,12 @@ private[parser] trait SimpleHeaders { this: Parser with CommonRules with CommonA
     rule { oneOrMore(addr).separatedBy(listSep) ~ EOI ~> (`X-Forwarded-For`(_)) }
   }
 
+  // de-facto standard as per https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host
   def `x-forwarded-host` = rule {
-    runSubParser(newUriParser(_).`relaxedHost-pushed`) ~ EOI ~> (`X-Forwarded-Host`(_))
+    host ~> (hostHeader ⇒ `X-Forwarded-Host`(hostHeader.host))
   }
 
+  // de-facto standard as per https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto
   def `x-forwarded-proto` = rule {
     runSubParser(newUriParser(_).`scheme-pushed`) ~ EOI ~> (`X-Forwarded-Proto`(_))
   }
--- a/akka-http-core/src/main/scala/akka/http/impl/model/parser/SimpleHeaders.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/model/parser/SimpleHeaders.scala
@@ -4,14 +4,17 @@
 
 package akka.http.impl.model.parser
 
+import akka.annotation.InternalApi
 import akka.parboiled2.Parser
 import akka.http.scaladsl.model.RemoteAddress
 import akka.http.scaladsl.model.headers._
 
 /**
+ * INTERNAL API.
  * Parser rules for all headers that can be parsed with one single rule.
  * All header rules that require more than one single rule are modelled in their own trait.
  */
+@InternalApi
 private[parser] trait SimpleHeaders { this: Parser with CommonRules with CommonActions with IpAddressParsing ⇒
 
   // http://tools.ietf.org/html/rfc7233#section-2.3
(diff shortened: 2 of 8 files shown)


akka-http-core/src/main/java/akka/http/javadsl/model/headers/XForwardedHost.java
```diff
@@ -0,0 +1,17 @@
+/**
+ * Copyright (C) 2009-2017 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.http.javadsl.model.headers;
+
+/**
+ *  Model for the `X-Forwarded-Proto` header.
```
jonas: Copy-pasta: s/Proto/Host/

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
```diff
@@ -976,6 +976,33 @@ final case class `X-Forwarded-For`(addresses: immutable.Seq[RemoteAddress]) exte
   def getAddresses: Iterable[jm.RemoteAddress] = addresses.asJava
 }
 
+object `X-Forwarded-Host` extends ModeledCompanion[`X-Forwarded-Host`] {
+  implicit val addressRenderer = {
+    implicit val singleAddressRenderer = RemoteAddress.renderWithoutPort
+    Renderer.defaultSeqRenderer[RemoteAddress] // cache
```
jonas: `X-Forwarded-Host` only holds one address, and this creates a renderer for a `Seq[RemoteAddress]`. There's no test to validate, but it looks like you want this:

```
implicit val singleAddressRenderer = RemoteAddress.renderWithoutPort
```

and the import in the `case class` should be updated accordingly.

akka-http-core/src/test/scala/akka/http/scaladsl/model/headers/HeaderSpec.scala
```diff
@@ -121,6 +121,8 @@ class HeaderSpec extends FreeSpec with Matchers {
         Upgrade(Vector(UpgradeProtocol("HTTP", Some("2.0")))),
         `User-Agent`("Akka HTTP Client 2.4"),
         `X-Forwarded-For`(RemoteAddress(InetAddress.getByName("192.168.0.1"))),
+        `X-Forwarded-Host`(RemoteAddress(InetAddress.getByName("192.168.0.2"))),
+        `X-Forwarded-Proto`("https"),
```
jonas: Would be nice to also add additional test cases in akka-http-core/src/test/scala/akka/http/impl/model/parser/HttpHeaderSpec.scala that also validate rendering when the passed `RemoteAddress` has a port.

akka-http-core/src/main/java/akka/http/javadsl/model/headers/XForwardedHost.java
jlprat: Ups

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
jlprat: Yes, you are right. I will change this.

akka-http-core/src/test/scala/akka/http/scaladsl/model/headers/HeaderSpec.scala
jlprat: I'll add some tests for the new headers

akka-http-core/src/main/scala/akka/http/impl/model/parser/SimpleHeaders.scala
```diff
@@ -226,6 +229,14 @@ private[parser] trait SimpleHeaders { this: Parser with CommonRules with CommonA
     rule { oneOrMore(addr).separatedBy(listSep) ~ EOI ~> (`X-Forwarded-For`(_)) }
   }
 
+  def `x-forwarded-host` = rule {
+    (`ip-v4-address` | `ip-v6-address`) ~ EOI ~> (b ⇒ `X-Forwarded-Host`(RemoteAddress(b)))
+  }
+
+  def `x-forwarded-proto` = rule {
+    token ~ EOI ~> (`X-Forwarded-Proto`(_))
```
jonas: Instead of `token`, this could be more strict by using zeroOMore([`scheme-char`](https://github.com/akka/akka-http/blob/82ac13a71c7342ce5887fdc9bb7a313f14b1659f/akka-http-core/src/main/scala/akka/http/impl/model/parser/CharacterClasses.scala#L55))

akka-http-core/src/test/scala/akka/http/impl/model/parser/HttpHeaderSpec.scala
```diff
@@ -722,6 +764,8 @@ class HttpHeaderSpec extends FreeSpec with Matchers {
       try parsedHeader should equal(header)
       catch {
         case e: TestFailedException if parsedHeader.toString == header.toString ⇒
+          println(s"!!!!!!$parsedHeader")
+          println(s"!!!!!!$header")
```
jonas: No need to include these printlns

akka-http-core/src/test/scala/akka/http/impl/model/parser/HttpHeaderSpec.scala
```diff
@@ -631,6 +631,48 @@ class HttpHeaderSpec extends FreeSpec with Matchers {
...
+      "X-Forwarded-Host: 1:2::" =!=> "1:2:0:0:0:0:0:0"
+      "X-Forwarded-Host: 1::" =!=> "1:0:0:0:0:0:0:0"
+      "X-Forwarded-Host: 1::3:4:5:6:7:8" =!=> "1:0:3:4:5:6:7:8"
+      "X-Forwarded-Host: 1:2::4:5:6:7:8" =!=> "1:2:0:4:5:6:7:8"
+      "X-Forwarded-Host: 1:2:3::5:6:7:8" =!=> "1:2:3:0:5:6:7:8"
+      "X-Forwarded-Host: 1:2:3:4::6:7:8" =!=> "1:2:3:4:0:6:7:8"
+      "X-Forwarded-Host: 1:2:3:4:5::7:8" =!=> "1:2:3:4:5:0:7:8"
+      "X-Forwarded-Host: 1:2:3:4:5:6::8" =!=> "1:2:3:4:5:6:0:8"
+      "X-Forwarded-Host: ::" =!=> "0:0:0:0:0:0:0:0"
+      "X-Forwarded-Host: akka.io" =!=
```
jonas: The [spec](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host) suggests that domain names are allowed, it contains the example `id42.example-cdn.com`.

akka-http-core/src/test/scala/akka/http/impl/model/parser/HttpHeaderSpec.scala
jlprat: This was clearly a mistake!
Sorry for that

akka-http-core/src/test/scala/akka/http/impl/model/parser/HttpHeaderSpec.scala
jlprat: Right!

akka-http-core/src/main/scala/akka/http/impl/model/parser/SimpleHeaders.scala
jlprat: Shouldn't it be exactly once?
I don't remember seeing any rule for that, hence the plain `token`

jonas: True, I don't know if we want to accept any char or only valid schema chars. BTW, should obviously be ```oneOrMore(`schema-char`)```.

akka-http-core/src/test/scala/akka/http/impl/model/parser/HttpHeaderSpec.scala
```diff
@@ -631,6 +631,22 @@ class HttpHeaderSpec extends FreeSpec with Matchers {
...
     }
 
+    "X-Forwarded-Host" in {
+      "X-Forwarded-Host: 1.2.3.4" =!= `X-Forwarded-Host`(Uri.Host("1.2.3.4"))
+      "X-Forwarded-Host: [2001:db8:cafe:0:0:0:0:17]" =!= `X-Forwarded-Host`(Uri.Host("2001:db8:cafe:0:0:0:0:17"))
+      "X-Forwarded-Host: [1234:5678:9abc:def1:2345:6789:abcd:ef00]" =!= `X-Forwarded-Host`(Uri.Host("1234:5678:9abc:def1:2345:6789:abcd:ef00"))
+      "X-Forwarded-Host: [1234:567:9a:d:2:67:abc:ef00]" =!= `X-Forwarded-Host`(Uri.Host("1234:567:9a:d:2:67:abc:ef00"))
+      "X-Forwarded-Host: [1:2:3:4:5:6:7:8]" =!= `X-Forwarded-Host`(Uri.Host("1:2:3:4:5:6:7:8"))
+      "X-Forwarded-Host: akka.io" =!= `X-Forwarded-Host`(Uri.Host("akka.io"))
+      "X-Forwarded-Host: [1:2:3:4::6:7:8]" =!= `X-Forwarded-Host`(Uri.Host("1:2:3:4::6:7:8"))
```
jlprat: Please note that as it is using a `Host` class it is rendering IPV6 classes verbatim

akka-http-core/src/main/scala/akka/http/impl/model/parser/SimpleHeaders.scala
```diff
@@ -226,6 +230,14 @@ private[parser] trait SimpleHeaders { this: Parser with CommonRules with CommonA
     rule { oneOrMore(addr).separatedBy(listSep) ~ EOI ~> (`X-Forwarded-For`(_)) }
   }
 
+  def `x-forwarded-host` = rule {
```
jrudolph: Are there any specifications we could link to?

jlprat: More or less official, https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto and https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host

The links are on the Java code, I can copy them over the Scala ones as well (I was following the existing pattern).

jlprat: Are these sources good enough? We could also link to the official `Forwarded` RFC that defines each sub-part of it (host, proto, for) https://tools.ietf.org/html/rfc7239#section-5.3 and https://tools.ietf.org/html/rfc7239#section-5.4

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
```diff
@@ -976,6 +976,30 @@ final case class `X-Forwarded-For`(addresses: immutable.Seq[RemoteAddress]) exte
   def getAddresses: Iterable[jm.RemoteAddress] = addresses.asJava
 }
 
+object `X-Forwarded-Host` extends ModeledCompanion[`X-Forwarded-Host`] {
+  implicit val hostRenderer = UriRendering.HostRenderer // cache
+}
+final case class `X-Forwarded-Host`(host: Uri.Host) extends jm.headers.XForwardedHost
```
jrudolph: I wonder if we should mark them as `@ApiMayChange` if the specification is not yet completely done, so we reserve the right to change the model later on?

akka-http-core/src/main/scala/akka/http/impl/model/parser/SimpleHeaders.scala
```diff
@@ -226,6 +230,14 @@ private[parser] trait SimpleHeaders { this: Parser with CommonRules with CommonA
...
+    runSubParser(newUriParser(_).`relaxedHost-pushed`) ~ EOI ~> (`X-Forwarded-Host`(_))
```
jrudolph: Hmm, calling the UriParser is a bit weird. Is there precedence for doing it that way?

jlprat: I'll take a look at the `Host` rule and try to re-use it.

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
jlprat: Open to that, they are X headers after all...

akka-http-core/src/main/scala/akka/http/impl/model/parser/SimpleHeaders.scala
jlprat: Let's see if this is what you meant.

akka-http-core/src/main/java/akka/http/javadsl/model/headers/XForwardedHost.java
```diff
@@ -0,0 +1,17 @@
...
+ */
+
+package akka.http.javadsl.model.headers;
+
+/**
+ *  Model for the `X-Forwarded-Host` header.
+ *  Specification: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host
+ */
+public abstract class XForwardedHost extends akka.http.scaladsl.model.HttpHeader {
+    public abstract akka.http.javadsl.model.Host getAddress();
```
jonas: I suggest to rename this method to `getHost`. The type is `Host` and the Scala case class member is also called `host`.

Issue #486, opened 2016-11-05
Several of the directive examples in the documentation are underlined red in IntelliJ


jrudolph:
This is really confusing for new users which often come from an IDE-only background. We should try to prevent these kind of situations (or make it really explicit in the documentation that this can happen).

It is especially bad, since it also concern the basic examples from here:

http://doc.akka.io/docs/akka-http/current/scala/http/routing-dsl/index.html#minimal-example

PR #1397, opened 2017-08-31
Some of Response code are not correct

SHUB9914:
I found error at some places like at  line number 28 , 31 , 34 ,  38 , 40 , 69 . which was because of not proper format of  HttpResponse , There should be the use of StatusCodes before defining any error and entity before defining message in HttpResponse , so I have corrected it  please check and confirm.

### DIFF
--- a/docs/src/test/scala/docs/http/scaladsl/server/RejectionHandlerExamplesSpec.scala
+++ b/docs/src/test/scala/docs/http/scaladsl/server/RejectionHandlerExamplesSpec.scala
@@ -25,19 +25,19 @@ object MyRejectionHandler {
     implicit def myRejectionHandler =
       RejectionHandler.newBuilder()
         .handle { case MissingCookieRejection(cookieName) =>
-          complete(HttpResponse(BadRequest, entity = "No cookies, no service!!!"))
+          complete(HttpResponse(StatusCodes.BadRequest, entity = "No cookies, no service!!!"))
         }
         .handle { case AuthorizationFailedRejection =>
-          complete((Forbidden, "You're out of your depth!"))
+          complete(HttpResponse(StatusCodes.Forbidden, entity = "You're out of your depth!"))
         }
         .handle { case ValidationRejection(msg, _) =>
-          complete((InternalServerError, "That wasn't valid! " + msg))
+          complete(HttpResponse(StatusCodes.InternalServerError, entity = "That wasn't valid! " + msg))
         }
         .handleAll[MethodRejection] { methodRejections =>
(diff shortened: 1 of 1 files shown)


lightbend-cla-validator:
Hi @SHUB9914,

raboof:
OK TO TEST

raboof:
Hi, thanks for taking the time to get in touch about this. You modified the documentation and changed the 'shorthand' `complete()` notation to 'long form' notation in a couple of places.

I think it's not too bad to show both, did you have problems with the shorthand notation?

SHUB9914:
Hi @raboof 
yes It's giving me error if I use this shorthand notation . That's why I updated it to the proper format. 
if you are passing only message in complete method than it's working fine , but if you are using some error with it than in that case you have to use StatusCodes for defining the error and the entity for the message , other wise  it will give an error .

SHUB9914:
hi @lightbend-cla-validator 
I have successfully signed the CLA

raboof:
Can you share which error you're getting? The examples in our documentation are also compiled so we're pretty confident they should work, so it should be interesting to learn why they don't work for you. Are you sure you have all the imports?

SHUB9914:
Hi @raboof 
image


Here you can see the red line occurs when I did not use StatusCodes before error "NotFound" .

raboof:
When using the shorthand notation you should not have `entity =` in there. Also, do you have an `import StatusCodes._` like in the documentation to bring `NotFound` into scope?

SHUB9914:
hi @raboof 
as you told for using 
import StatusCodes._    and  you should not have entity = 
so I did it . but after that it's not working , you can see below  

image

SHUB9914:
and if I am using StatusCodes then it's working , you can see below 
image

jrudolph:
That's a bug in IntelliJ IDEA. Can make sure you run the latest version of IntelliJ IDEA and otherwise create an issue at their bug tracker at https://youtrack.jetbrains.com/issues/SCL

raboof:
It seems you accidentally imported `org.omg.CosNaming.NamingContextPackage.NotFound`, removing that might also help?

SHUB9914:
Hi @raboof 
Now it's working , it was not working because of  org.omg.CosNaming.NamingContextPackage.NotFound

raboof:
@SHUB9914 that's great! I suppose there's not a lot we can do in the documentation to make this easier, so perhaps we should just close this PR?

SHUB9914:
yes sure @raboof

jonas:
Related with #486

PR #1403, opened 2017-09-02
=doc Consolidate Routing DSL source streaming support docs #1290

atiqsayyed:
Ref #1290
Merged source streaming support docs


ktoso:
OK TO TEST

Issue #992, opened 2017-03-27
Adapt Actor Examples to Akka 2.5


jlprat:
Once Akka 2.5 is released, please convert all existing usages of Akka Actors to the new API introduced in Akka 2.5.

Please add some note to hint readers that this is the new Actor API and offer some links to `AbstractActor`

ilke-zilci:
Hey, is anyone working on this?

ktoso:
Hm, I was thinking we did that for actors.md while we released 2.5, what's missing @jlprat ?

ktoso:
Ah, I missed this ticket is in HTTP.
Let's do this however we'll want to publish versioning formalised guidelines (Johannes will post proposal this week) and bump the version - so then we actually depend on 2.5 and should update the docs - let's keep it parked until then - should be just a few weeks

jlprat:
Sure thing

jlprat:
This one is up for picking @ilke-zilci would you like to take this one?

ilke-zilci:
Yes sure. I started working on it.

jlprat:
Awesome!

ktoso:
Great -- thanks! -- marked as in progress :)

jrudolph:
Fixed in #1831.

PR #1413, opened 2017-09-07
Fix the "Blocking needs careful management" link

TimMoore:
This section has moved to the "Dispatchers" page in the latest version of the Akka documentation.

### DIFF
--- a/docs/src/main/paradox/scala/http/handling-blocking-operations-in-akka-http-routes.md
+++ b/docs/src/main/paradox/scala/http/handling-blocking-operations-in-akka-http-routes.md
@@ -132,5 +132,5 @@ This is the recommended way of dealing with any kind of blocking in reactive
 applications. It is referred to as "bulkheading" or "isolating" the bad behaving
 parts of an app. In this case, bad behaviour of blocking operations.
 
-There is good documentation available in Akka docs section, 
-[Blocking needs careful management](http://doc.akka.io/docs/akka/current/scala/dispatchers.html#blocking-needs-careful-management).
+There is good documentation available in Akka docs section,
+@extref[Blocking needs careful management](akka-docs:general/actor-systems.html#Blocking_Needs_Careful_Management).
--- a/docs/src/main/paradox/scala/http/handling-blocking-operations-in-akka-http-routes.md
+++ b/docs/src/main/paradox/scala/http/handling-blocking-operations-in-akka-http-routes.md
@@ -133,4 +133,4 @@ applications. It is referred to as "bulkheading" or "isolating" the bad behaving
 parts of an app. In this case, bad behaviour of blocking operations.
 
 There is good documentation available in Akka docs section, 
-[Blocking needs careful management](http://doc.akka.io/docs/akka/current/general/actor-systems.html#Blocking_Needs_Careful_Management).
+[Blocking needs careful management](http://doc.akka.io/docs/akka/current/scala/dispatchers.html#blocking-needs-careful-management).


jlprat:
OK TO TEST

docs/src/main/paradox/scala/http/handling-blocking-operations-in-akka-http-routes.md
```diff
@@ -133,4 +133,4 @@ applications. It is referred to as "bulkheading" or "isolating" the bad behaving
 parts of an app. In this case, bad behaviour of blocking operations.
 
 There is good documentation available in Akka docs section, 
-[Blocking needs careful management](http://doc.akka.io/docs/akka/current/general/actor-systems.html#Blocking_Needs_Careful_Management).
+[Blocking needs careful management](http://doc.akka.io/docs/akka/current/scala/dispatchers.html#blocking-needs-careful-management).
```
jlprat: It needs to point to the the Java version or the Scala one accordingly. Basically you would need to use the `@scala[]` and `@java[]` directives.
You can read further here http://developer.lightbend.com/docs/paradox/latest/features/groups.html#inline

jlprat:
Hi @TimMoore ! Thanks for your contribution!

docs/src/main/paradox/scala/http/handling-blocking-operations-in-akka-http-routes.md
TimMoore: Thanks for pointing that out. I'll fix it.

jlprat: Thank you for the quick response :+1:

TimMoore:
This has turned into a bit of a yak-shaving exercise 😄 

I noticed while working on the requested change that there is another link in the file with the form: `@extref[Dispatchers](akka-docs:scala/dispatchers.html)`

This made me think:

1. There are other links that need to be updated to switch between Java/Scala versions
2. The link that I'm changing here should probably be using the `akka-docs` `extref` directive

I started by updating this link to do both:

```
There is good documentation available in Akka docs section,
@scala[@extref[Blocking needs careful management](akka-docs:scala/dispatchers.html#blocking-needs-careful-management)]
@java[@extref[Blocking needs careful management](akka-docs:java/dispatchers.html#blocking-needs-careful-management)].
```

I found that the generated link goes to the 2.4 docs (http://doc.akka.io/docs/akka/2.4.19/java/dispatchers.html#blocking-needs-careful-management) which is wrong again.

The build defines the external reference as:

```scala
      "extref.akka-docs.base_url" -> s"http://doc.akka.io/docs/akka/${Dependencies.akkaVersion.value}/%s",
```

`Dependencies.akkaVersion` is an sbt setting that I guess is designed to allow for building against different Akka versions. It currently defaults to `"2.4.19"`.

So, there are a few ways I could proceed with this:

1.  Use `@extref` and keep the 2.4 URL structure. This would mean that the link will break again if the default changes to link to the 2.4 docs in the future (though I notice there are many other links that will also break when this happens).
2.  Keep the hardcoded link to the Akka docs site that is in there now, updated for the new structure in the 2.5 docs as in my previous commit, but using the `@scala`/`@java` group tags to switch between the two versions of the Akka 2.5 docs.
3.  Change the docs build to default to the 2.5 docs, by changing the `extref` definition to:
    ```scala
    "extref.akka-docs.base_url" -> s"http://doc.akka.io/docs/akka/${Dependencies.akka25Version.value}/%s"
    ```
    This is probably a good idea, since the Akka 2.5 docs are a big improvement from the previous ones, but will break a lot of links, so it should probably be a separate change.

I'll plan to go with the first option for now, just to get the link working again, but if the maintainers think it would be a good idea to update to 2.5 across the board, I can try to spend some time fixing things up in a separate pull request.

ktoso:
> Dependencies.akkaVersion is an sbt setting that I guess is designed to allow for building against different Akka versions. It currently defaults to "2.4.19".

Whoa, good catch!

In truth, the version there should be the same version as the "thing being published right now" and that means just `version` I believe:

```
> show version
[info] 2.5-SNAPSHOT
> show akka:version
[info] 2.5-SNAPSHOT
```

Since we change that version when we publish and it will be the right link then.
So it's option 3, but with `version` instead of adding something new.

TimMoore:
@ktoso but this is for Akka HTTP :smile:

I guess it is intentional that it still builds against Akka 2.4 by default (for backwards compatibility), so the question is whether the Akka HTTP documentation should continue to link to the Akka 2.4 documentation, or if the dependency and the documentation links should be decoupled to get the latest version of the docs even with an earlier version of the binary dependency.

My recommendation, now that I've looked into it a little more, is to keep linking to the 2.4 docs for now, because otherwise a lot of links will break. Then, as part of to move to depend directly on Akka 2.5 (#1151) the links can all be updated to the new structure in one pass. The content also needs updates (#992) so it sounds like it could be a big effort.

With the latest commit on this branch, I think it's best to merge as is to fix the link to point to Akka 2.4, like the rest do, and make the bigger changes later.

ktoso:
> @ktoso but this is for Akka HTTP 😄

Ah damn, missed that. Yeah you're right then...

In general perhaps pointing to "current" would be our long term goal here. If you think there will be much work indeed in the move then let's wait as you said until we change to 2.5 dep.

In general though I would not mind linking to /current/ right away, it's what we'd want people to use anyway - regardless of the fact that the dependency is made to 2.4 nowadays still...

cc @jrudolph 

This fix seems good though, so let's merge as is, thanks a ton!

Issue #235, opened 2016-09-12
Allow configuration of how TLS should handle possible truncation attacks


ktoso:
<a href="https://github.com/jrudolph"><img src="https://avatars.githubusercontent.com/u/9868?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [jrudolph](https://github.com/jrudolph)**
_Tuesday Feb 10, 2015 at 15:16 GMT_
_Originally opened as https://github.com/akka/akka/issues/16825_

----

spray's latest version of SslTlsSupport is very strict in its handling of truncated messages. It turns out that there are several situations in combination with HTTP where some kind of truncation is likely to happen and produces warnings regularly. For https://groups.google.com/d/topic/spray-user/8FUAZ_QAfs4/discussion I made a quick test how current browser deal with potential truncation attacks and it seems that browsers are usually much more lenient than spray.

We should review that situation with akka-streams SSL implementation and maybe provide a configuration option about how to deal with possible truncation attacks.

See spray/spray#756.

/cc @sirthias

jrudolph:
Right now the client-side just ignores truncations here:

https://github.com/akka/akka-http/blob/e56661d3e759bee94e6c2b966e9a1e5b75373e7d/akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala#L102-L102

We might want to add a paranoid mode that fails the stream if truncation occurs.

PR #1420, opened 2017-09-11
=htc #380 add test case for server->client missing close_notify scenario

### LINKED ISSUE #380: HttpEntity.CloseDelimited toStrict times out (on the client-side) when using SSL
HTTP Client fails to get all data in a CloseDelimited entity when requests are done over https.

The .toStrict method times out even though the remote end has closed the connection and all data has been sent.
This is only a problem with SSL connections, plaintext-connections work as expected.

I have confirmed this behaviour against akka http servers and other http servers.

The attachked file contains a scalatest test class to reproduce this problem using akka as client and server and also a keystore required for ssl to work.

[tostrict.zip](https://github.com/akka/akka-http/files/521655/tostrict.zip)

jrudolph:
That is we are simulating a TLS truncation attack here. Before
https://github.com/akka/akka/pull/21786 the TLS stage just got stuck
(I tested again an old Akka version).

Now, all streams are properly closed and also the response entity
stream is completed. A stricter version would actually fail the response
entity stream when TLS truncation is detected. If that is worthwhile is
actually questionable. See https://github.com/akka/akka-http/issues/235
for more information.

Finally fixes #380.

### DIFF
--- a/akka-http-core/src/test/scala/akka/http/scaladsl/ClientServerSpec.scala
+++ b/akka-http-core/src/test/scala/akka/http/scaladsl/ClientServerSpec.scala
@@ -14,6 +14,7 @@ import scala.concurrent.duration._
 import scala.concurrent.{ Await, Future, Promise }
 import scala.util.{ Success, Try }
 import akka.actor.ActorSystem
+import akka.http.impl.engine.ws.ByteStringSinkProbe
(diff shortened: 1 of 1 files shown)


jrudolph:
PLS BUILD

PR #1424, opened 2017-09-14
=htc replace usage of `RuleNotFoundException` by introducing explicit result ADT

jrudolph:
Before every attempt to parse an unknown (i.e. not modeled) header would
end up in throwing and catching an exception. This does not really make
sense as parboiled already provides for delivering these signals without
an exception.

Refs playframework/playframework#7822

### DIFF
--- a/akka-http-core/src/main/mima-filters/10.0.10.backwards.excludes
+++ b/akka-http-core/src/main/mima-filters/10.0.10.backwards.excludes
@@ -1,0 +1,2 @@
+# Don't monitor changes to internal API
+ProblemFilters.exclude[Problem]("akka.http.impl.*")
--- a/akka-http-core/src/main/scala/akka/http/impl/model/parser/HeaderParser.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/model/parser/HeaderParser.scala
@@ -62,18 +62,21 @@ private[http] class HeaderParser(
 
   ///////////////// DynamicRuleHandler //////////////
 
-  type Result = Either[ErrorInfo, HttpHeader]
+  override type Result = HeaderParser.Result
   def parser: HeaderParser = this
-  def success(result: HttpHeader :: HNil): Result = Right(result.head)
-  def parseError(error: ParseError): Result = {
+  def success(result: HttpHeader :: HNil): Result = HeaderParser.Success(result.head)
+  def parseError(error: ParseError): HeaderParser.Failure = {
     val formatter = new ErrorFormatter(showLine = false)
-    Left(ErrorInfo(formatter.format(error, input), formatter.formatErrorLine(error, input)))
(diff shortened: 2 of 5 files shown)


jrudolph:
Added missing mima filters.

akka-http-core/src/main/mima-filters/10.0.10.backwards.excludes
```diff
@@ -0,0 +1,7 @@
+# Changes to internal API
+ProblemFilters.exclude[MissingClassProblem]("akka.http.impl.model.parser.HeaderParser$RuleNotFoundException$")
```
raboof: Can we `ProblemFilters.exclude[Problem]("akka.http.impl.*")`?

akka-http-core/src/main/scala/akka/http/impl/model/parser/HeaderParser.scala
```diff
@@ -62,18 +62,21 @@ private[http] class HeaderParser(
 
   ///////////////// DynamicRuleHandler //////////////
 
-  type Result = Either[ErrorInfo, HttpHeader]
+  type Result = HeaderParser.Result
```
raboof: I was confused why not just use `HeaderParser.Result` directly, but this is specifying the `Result` from `DeliveryScheme`.

It looks like `type` also accepts an `override` modifier, might be a nice way to make that explicit?

akka-http-core/src/main/scala/akka/http/impl/model/parser/HeaderParser.scala
```diff
@@ -62,18 +62,21 @@ private[http] class HeaderParser(
...
-  }
-  def ruleNotFound(ruleName: String): Result = throw HeaderParser.RuleNotFoundException
+  def failure(error: Throwable): HeaderParser.Failure =
+    HeaderParser.Failure {
+      error match {
+        case IllegalUriException(info) ⇒ info
+        case NonFatal(e)               ⇒ ErrorInfo.fromCompoundString(e.getMessage)
+      }
+    }
+  def ruleNotFound(ruleName: String): Result = HeaderParser.RuleNotFound
```
raboof: :+1: nice!

akka-http-core/src/main/mima-filters/10.0.10.backwards.excludes
jrudolph: I guess we can. It's somewhat interesting to see what has changed in the API but it's mostly distracting so I'll try.

akka-http-core/src/main/scala/akka/http/impl/model/parser/HeaderParser.scala
jrudolph: :+1:

Issue #995, opened 2017-03-28
Allow to configure ClientTransport with application.conf


jrudolph:
Right now it can only be done in code. The facility should allow to specify parameters, e.g. the upcoming HTTPS proxy transport needs a host and port configuration.

jiminhsieh:
It seems have to add 2 properties to  `akka.http.client` in application.conf. Do you have any idea of how this format should be? It could be `akka.http.client.https-proxy.host` or `akka.http.client.https-proxy-host`. Thanks!

jiminhsieh:
@jrudolph It seems original client proxy had been removed at #1249. So the format should likes?
```
akka.http {
  client {
    proxy {
      https {
        host = myproxy.com
        port = 8080
      }
    }
  }
}
```
Thanks!

raboof:
@jiminhsieh Right, that's what it looked like in Spray - sounds good to keep that general approach!

jiminhsieh:
After I traced through Spary’s source code, I have some questions. I would post full of original format of spray proxy config first.

```
spray.can {
  client {
    proxy {
      http {
      	host = myproxy.com
      	port = 8080
      }
      https = {default, none}
    }
  }
}
```

There are my questions: 
1. Does it need to support system properties which was represent as `default` in Spray?
2. Does it need `none` in addition to `default`? If it has `none`, should we use `none` as default value instead of `default`?

There is my idea of how to implement it. Create public API, `httpsProxy` without parameter, for client. The rest of code would be similar to Spary with current design of Akka-HTTP which split to abstract class and case class of `-impl`.

If there is any hint or suggestion, I will be glad to listen. : )

PR #1425, opened 2017-09-14
Added ability to set https proxy host and port through application.conf

Scalahansolo:
Fixes: #995

Right now, this does not allow for setting auth parameters in the settings as the HttpCredentials class in akka has a few different implementation, including Basic/Oauth/ect. As it includes a few I was able to come up with a concise implementation to get that set. If it would be desired, I can add the ability to add Basic auth at the very least in the settings, or... it can be kept as it and auth must always be passed directly to the httpsProxy method.


ktoso:
OK TO TEST

akka-http-core/src/main/resources/reference.conf
```diff
@@ -233,6 +233,15 @@ akka.http {
       tcp-no-delay = undefined
     }
 
+    # Client https proxy options. If a setting is left undefined/-1, an error will be thrown
+    # indicating that the options must be set.
```
ktoso: Seems a bit weird - I mean, only if I actually want to use the proxy these should be required, reword the requirement somehow?

akka-http-core/src/main/resources/reference.conf
```diff
@@ -233,6 +233,16 @@ akka.http {
       tcp-no-delay = undefined
     }
 
+    # Client https proxy options. When using ClientTransport.httpsProxy() with or without credentials,
+    # host/port must be either passed explicitly or set here. When using the proxy from settings leaving these set as
+    # undefined/-1, an error will be thrown indicating that the options must be set.
+    proxy {
+      https {
+        host = undefined
+        port = -1
```
raboof: Hmm, we don't have a great way to encode 'by default unspecified' settings... could we find a default? For the port perhaps 443 might make sense as a default, maybe use the empty string for an unspecified host?

akka-http-core/src/main/scala/akka/http/javadsl/ClientTransport.scala
```diff
@@ -49,6 +47,15 @@ object ClientTransport {
   def httpsProxy(proxyAddress: InetSocketAddress): ClientTransport =
     scaladsl.ClientTransport.httpsProxy(proxyAddress).asJava
 
+  /**
+   * Returns a [[ClientTransport]] that runs all connection through the given HTTP(S) proxy using the
+   * HTTP CONNECT method.
+   *
+   * Pulls the host/port pair from the application.conf: akka.client.proxy.https.{host, port}
+   */
+  def httpsProxy()(implicit system: ActorSystem): ClientTransport =
```
raboof: As this is the `javadsl`, implicit parameters don't really make sense. You could expect the `ActorSystem` as an explicit parameter here, or perhaps expect a `HttpsProxySettings` and let the user take care of creating it... though that would make it part of the public API (and thus need converting it from a case class to a 'regular' class because of binary compatibility)

raboof:
PLS BUILD

raboof:
(previous build failure was a build server configuration thing)

raboof:
It looks like scalariform doesn't like the test formatting - perhaps this branch needs a rebase and `test:compile`?

raboof:
@Scalahansolo looks like this PR is almost there, would you be interested in taking it over the finish line?

Scalahansolo:
Hey Akka team / @raboof Its been a long time since Ive looked at this. I think that it would still be good to land this at some point. Ill try to take a crack at upgrading this PR and see how that goes.

Issue #1353, opened 2017-08-15
Parse and render token in HttpChallenge / GenericHttpCredentials correctly


jrudolph:
[RFC 7235](https://tools.ietf.org/html/rfc7235#appendix-C) defines the grammar for challenge and credentials to be the same:

```
challenge = auth-scheme [ 1*SP ( token68 / [ ( "," / auth-param ) *(
    OWS "," [ OWS auth-param ] ) ] ) ]
credentials = auth-scheme [ 1*SP ( token68 / [ ( "," / auth-param )
    *( OWS "," [ OWS auth-param ] ) ] ) ]
```

However, our modelling has several problems:

 * `HttpChallenge` has no token parameter (instead the token is parsed into a parameter with an empty name)
 * `GenericHttpCredentials` has a token parameter but it seems that the parser doesn't fill it but instead creates a parameter with an empty name

@dwickern provided a failing test case in https://github.com/akka/akka-http/pull/1329

jrudolph:
For a failing test see https://github.com/akka/akka-http/pull/1329

PR #1329, opened 2017-08-04
[failed test] WWW-Authenticate: NTLM header is rendered incorrectly

dwickern:
An NTLM header looks like this:
```
WWW-Authenticate: NTLM TlRMTVNTUAABAAAABzIAAAYABgArAAAACwALACAAAABXT1JLU1RBVElPTkRPTUFJTg==
```

But instead it's rendered as
```
WWW-Authenticate: NTLM ="TlRMTVNTUAABAAAABzIAAAYABgArAAAACwALACAAAABXT1JLU1RBVElPTkRPTUFJTg=="
```

PR #1427, opened 2017-09-16
Make HttpChallenge and GenericHttpCredentials parse and render token …

satansk:
…correctly. #1353

* Make the parser fill GenericHttpCredentials's token parameter correctly
* Add a token parameter to HttpChallenge, and make it parsed and rendered correctly
* Add tests for both


jonas:
Ok TO TEST

jonas:
```
[error] (akka-http-core/*:mimaReportBinaryIssues) akka-http-core: Binary compatibility check failed!
```

Likely due to 
```diff
-final case class HttpChallenge(scheme: String, realm: String,
+final case class HttpChallenge(scheme: String, token: String, realm: String,
```

Based on the description in #1353, can this be fixed by storing the token in the `params: Map[String, String]`?

satansk:
Hi @jonas , thanks for your advice!

#1353 said `HttpChallenge` has no token parameter, so I add it. If we store the token in `params: Map[String, String]`, this will be somewhat different from the grammar for challenge:

```
challenge = auth-scheme [ 1*SP ( token68 / [ ( "," / auth-param ) *(
    OWS "," [ OWS auth-param ] ) ] ) ]
```

BTW, why the `realm` parameter is treated differently, after all, it's just one of the auth-params? Maybe we can replace `realm` with `token` to fix the `Binary compatibility check failed!`, what do you think?

akka-http-core/src/main/scala/akka/http/impl/model/parser/CommonRules.scala
```diff
@@ -172,17 +172,22 @@ private[parser] trait CommonRules { this: Parser with StringBuilding ⇒
   def `token68` = rule { capture(oneOrMore(`token68-start`) ~ zeroOrMore('=')) ~ OWS }
```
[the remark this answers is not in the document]
satansk: Yes, no problem, I will change it :)

akka-http-core/src/test/scala/akka/http/impl/model/parser/HttpHeaderSpec.scala
```diff
@@ -591,6 +599,8 @@ class HttpHeaderSpec extends FreeSpec with Matchers {
         `WWW-Authenticate`(HttpChallenge("Basic", None, Map("attr" → "value")))
       """WWW-Authenticate: Fancy realm="Secure Area",nonce=42""" =!=
         `WWW-Authenticate`(HttpChallenge("Fancy", Some("Secure Area"), Map("nonce" → "42")))
+      """WWW-Authenticate: NTLM TlRMTVNTUAABAAAABzIAAAYABgArAAAACwALACAAAABXT1JLU1RBVElPTkRPTUFJTg==""" =!=
+        `WWW-Authenticate`(HttpChallenge("NTLM", "TlRMTVNTUAABAAAABzIAAAYABgArAAAACwALACAAAABXT1JLU1RBVElPTkRPTUFJTg=="))
```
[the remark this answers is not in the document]
satansk: Oh, I'm not quite sure about this. I will take a deeper look at it. 
And thanks a lot for your review :)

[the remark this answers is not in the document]
satansk: 1. For now, Akka HTTP will parse multiple challenges into one `WWW-Authenticate` header, as the tests show:

```scala
"WWW-Authenticate: Basic realm=\"WallyWorld\",attr=\"val>ue\", Fancy realm=\"yeah\"" =!=
  `WWW-Authenticate`(HttpChallenge("Basic", Some("WallyWorld"), Map("attr" → "val>ue")), HttpChallenge("Fancy", Some("yeah")))

"WWW-Authenticate: Basic realm=\"WallyWorld\",attr=\"val>ue\", Fancy TlRMTVNTUAABAAAABzIAAAYABgArAAAACwALACAAAABXT1JLU1RBVElPTkRPTUFJTg==" =!=
  `WWW-Authenticate`(HttpChallenge("Basic", Some("WallyWorld"), Map("attr" → "val>ue")),
    HttpChallenge("Fancy", "TlRMTVNTUAABAAAABzIAAAYABgArAAAACwALACAAAABXT1JLU1RBVElPTkRPTUFJTg=="))
```

2. As for repeated `WWW-Authenticate`, the `header[`WWW-Authenticate`].get.challenges` has only one element:
```scala
def myUserPassAuthenticator(credentials: Credentials): Option[String] =
    credentials match {
      case p @ Credentials.Provided(id) if p.verify("p4ssw0rd") => Some(id)
      case _ => None
    }

  val route =
    Route.seal {
      path("secured") {
        authenticateBasic(realm = "XXXX", myUserPassAuthenticator) { user1 => {
            authenticateBasic(realm = "YYYY", myUserPassAuthenticator) { user2 =>
              complete(s"User1 is '$user1', and user2 is '$user2'")
            }
          }
        }
      }
    }

  "www-authentication" should {
    "repeated headers" in {
      Get("/secured") ~> route ~> check {
        status shouldEqual StatusCodes.Unauthorized
        responseAs[String] shouldEqual "The resource requires authentication, which was not supplied with the request"
        header[`WWW-Authenticate`].get.challenges.head shouldEqual HttpChallenge("Basic", Some("XXXX"), Map("charset" → "UTF-8"))
//        header[`WWW-Authenticate`].get.challenges.tail.head shouldEqual HttpChallenge("Basic", Some("YYYY"), Map("charset" → "UTF-8"))
      }
    }
  }
```
I think, in the second snippet, the request is rejected by the first `authenticateBasic`, so there's only one `WWW-Authenticate`. Any way to construct repeated `WWW-Authenticate` headers? Thanks :)

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/HttpChallenge.scala
```diff
@@ -9,12 +9,13 @@ import akka.http.javadsl.{ model ⇒ jm }
 import akka.http.impl.util._
 import akka.http.impl.util.JavaMapping.Implicits._
 
-final case class HttpChallenge(scheme: String, realm: String,
+final case class HttpChallenge(scheme: String, token: String, realm: String,
```
jrudolph: The main problem here is that we need to find a binary backward compatible solution for adding the field.

satansk: The `realm` parameter is just one kind of `auth-param`, I think the `params` Map is a more suitable place for it, and it will be binary backward compatible.

raboof:
@satansk would you be interested in fixing that binary incompatibility? Seems like a useful improvement!

satansk:
@raboof I'd love to fix it. The thing is I don't quite know how. As my comment here:

> The `realm` parameter is just one kind of `auth-param`, I think the `params` Map is a more suitable place for it, and it will be binary backward compatible.

I will appreciate if you would provide some insight.

raboof:
The binary incompatibility there is you added `token` as a field in the existing `HttpChallenge` case class. Moving the `realm` into the `Map` and adding the `token` will probably still not be binary compatible, as the field name is part of the binary class as well. I see you have discussed this with @jonas above already, but would it be suitable to put the `token` into the Map instead of the `realm`?

satansk:
@jonas @raboof as we have discussed, it seems that we can only put the `token` into the `params` Map to avoid binary incompatibility. But this is some what inconsistent with the RFC.  Could we just do this?

raboof:
@satansk Fixing the behaviour in #1329 but still storing the token in the `params` map for binary compatibility seems fine to me.

The RFC doesn't dictate how we lay out the data in memory, as long as the parser/renderer does the right thing we should be fine. Of course it is confusing the mapping between the protocol and the case class is not so 1-to-1, but that's something we should fix when we can break binary compatibility in a later version.

satansk:
@raboof thanks, I will give it try :)

satansk:
Hi @raboof @jonas @jrudolph, now I store the token in the params map, and fix the binary compatibility, I'll appreciate it if you could give some other advice, thank you :)

akka-http-core/src/test/scala/akka/http/impl/model/parser/HttpHeaderSpec.scala
```diff
@@ -577,14 +585,14 @@ class HttpHeaderSpec extends FreeSpec with Matchers {
       "WWW-Authenticate: Basic realm=\"WallyWorld\"" =!=
         `WWW-Authenticate`(HttpChallenge("Basic", Some("WallyWorld")))
       "WWW-Authenticate: BaSiC rEaLm=WallyWorld" =!=
-        `WWW-Authenticate`(HttpChallenge("BaSiC", "WallyWorld")).renderedTo("BaSiC realm=\"WallyWorld\"")
+        `WWW-Authenticate`(HttpChallenge("BaSiC", Some("WallyWorld"))).renderedTo("BaSiC realm=\"WallyWorld\"")
```
raboof: (this change isn't strictly needed anymore, but can't harm I guess)

akka-http-core/src/test/scala/akka/http/impl/model/parser/HttpHeaderSpec.scala
```diff
@@ -596,6 +604,8 @@ class HttpHeaderSpec extends FreeSpec with Matchers {
         `WWW-Authenticate`(HttpChallenge("Basic", None, Map("attr" → "value")))
       """WWW-Authenticate: Fancy realm="Secure Area",nonce=42""" =!=
         `WWW-Authenticate`(HttpChallenge("Fancy", Some("Secure Area"), Map("nonce" → "42")))
+      """WWW-Authenticate: NTLM TlRMTVNTUAABAAAABzIAAAYABgArAAAACwALACAAAABXT1JLU1RBVElPTkRPTUFJTg==""" =!=
+        `WWW-Authenticate`(HttpChallenge("NTLM", None, Map("" → "TlRMTVNTUAABAAAABzIAAAYABgArAAAACwALACAAAABXT1JLU1RBVElPTkRPTUFJTg==")))
```
raboof: 👍

akka-http-core/src/test/scala/akka/http/impl/model/parser/HttpHeaderSpec.scala
satansk: Yes, we don't need the `Some` now, I'll remove it.

satansk:
@raboof A comment will be nice, and I will add it :)

PR #1428, opened 2017-09-18
=htc lookup predefined header parsers as early as possible

jrudolph:
Another small improvement as a follow up to #1424.

Previously, every header parsing attempt had to go through the
`DynamicRuleDispatch`, a binary search to find the parsing rule for an
HTTP header name.

However, the `HttpHeaderParser` already has a parsing tree that with a node
for every header name.

This change divides header parsing into two steps: looking up the rule
and invoking the parser. Looking up the rule now only needs to be once
per node in the `HttpHeaderParser` and afterwards it can just run
without going through the dispatch.


akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpHeaderParser.scala
```diff
@@ -494,17 +494,21 @@ private[http] object HttpHeaderParser {
 
   private[parsing] class ModeledHeaderValueParser(headerName: String, maxHeaderValueLength: Int, maxValueCount: Int, log: LoggingAdapter, settings: HeaderParser.Settings)
     extends HeaderValueParser(headerName, maxValueCount) {
+    val parser = HeaderParser.lookupParser(headerName, settings).getOrElse(
+      throw new IllegalStateException(s"Missing parser for modeled [$headerName].")
+    )
```
ktoso: ah nice cached here

PR #1431, opened 2017-09-19
Clarify discarding for strict entities #1430

### LINKED ISSUE #1430: HttpEntity.discardBytes Scaladoc
I am trying to figure something out and either I am misunderstanding the documentation or it is wrong.
Either way, I would appreciate some clarification :)

My question will be about `HttpEntity.discardBytes` and `HttpMessage.discardEntityBytes` (which is essentially the same thing)
https://github.com/akka/akka-http/blob/c3b670578547616caa081d55948b84dd0a60ae5d/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala#L96

The Scaladoc says this:
```
If the dataBytes source is materialized a second time, it will fail with an "stream can cannot be materialized more than once" exception.
```

I have this code:

```scala
implicit val system = ActorSystem(name = "Playground")
val materializer = ActorMaterializer() // Non-implicit by choice

val entity = HttpEntity(ContentTypes.`application/json`, "Hello world")
val request = HttpRequest(entity = entity)

val blubb1 = Await.result(request.discardEntityBytes()(materializer).future, 10.seconds)
val blubb2 = Await.result(request.discardEntityBytes()(materializer).future, 10.seconds)

println(blubb1)
println(blubb2)
```

Should I not see an exception as per Scaladoc? Or am I misunderstanding?
I do not, both `blubb{1,2}` are printed out...

I even found this test: https://github.com/akka/akka-http/blob/74302ef20101daa8aee784a1240134eab7fffb27/akka-http-core/src/test/scala/akka/http/scaladsl/model/EntityDiscardingSpec.scala#L58
Am I brain-farting here or is the test not doing essentially the same I am trying?!

NeQuissimus:
Fixes #1430 

Adds a small paragraph to `HttpEntity.discardBytes` and `HttpMessage.discardEntityBytes` that clarifies the different behaviour for `Strict` entities.

Tried to follow the tone and wording in other Scaladocs in the same files for consistency.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
@@ -91,6 +91,10 @@ sealed trait HttpEntity extends jm.HttpEntity {
    * of its streaming nature. If the dataBytes source is materialized a second time, it will fail with an
    * "stream can cannot be materialized more than once" exception.
    *
+   * When called on `Strict` entities or sources whose values can be buffered in memory,
+   * the above warnings can be ignored. Repeated materialization is not necessary in this case, avoiding
+   * the mentioned exceptions due to the data being held in memory.
+   *
    * In future versions, more automatic ways to warn or resolve these situations may be introduced, see issue #18716.
    */
   override def discardBytes(mat: Materializer): HttpMessage.DiscardedEntity =
--- a/akka-http-core/src/main/java/akka/http/javadsl/model/HttpEntity.java
+++ b/akka-http-core/src/main/java/akka/http/javadsl/model/HttpEntity.java
@@ -155,6 +155,10 @@ public interface HttpEntity {
      * of its streaming nature. If the dataBytes source is materialized a second time, it will fail with an
      * "stream can cannot be materialized more than once" exception.
      *
+     * When called on `Strict` entities or sources whose values can be buffered in memory,
+     * the above warnings can be ignored. Repeated materialization is not necessary in this case, avoiding
+     * the mentioned exceptions due to the data being held in memory.
+     *
      * In future versions, more automatic ways to warn or resolve these situations may be introduced, see issue #18716.
      */
     HttpMessage.DiscardedEntity discardBytes(Materializer materializer);
--- a/akka-http-core/src/main/java/akka/http/javadsl/model/HttpMessage.java
+++ b/akka-http-core/src/main/java/akka/http/javadsl/model/HttpMessage.java
@@ -77,6 +77,10 @@ public interface HttpMessage {
      * of its streaming nature. If the dataBytes source is materialized a second time, it will fail with an
      * "stream can cannot be materialized more than once" exception.
      *
+     * When called on `Strict` entities or sources whose values can be buffered in memory,
+     * the above warnings can be ignored. Repeated materialization is not necessary in this case, avoiding
+     * the mentioned exceptions due to the data being held in memory.
+     *
      * In future versions, more automatic ways to warn or resolve these situations may be introduced, see issue #18716.
      */
     DiscardedEntity discardEntityBytes(Materializer materializer);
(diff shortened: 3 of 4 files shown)


ktoso:
OK TO TEST

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
```diff
@@ -91,6 +91,10 @@ sealed trait HttpEntity extends jm.HttpEntity {
    * of its streaming nature. If the dataBytes source is materialized a second time, it will fail with an
    * "stream can cannot be materialized more than once" exception.
    *
+   * When called on `Strict` entities, the above warnings can be ignored. Discarding is not required because
+   * the underlying [[akka.stream.scaladsl.Source]] will have already been drained.
```
ktoso: I’d propose to reword to be mor precise, if something was already drained that is also the same situation as before. This is about that it is “already completely available in memory”, so there is no need to materialize the stream, it just has the strict data readily available

NeQuissimus: Wait, can you expand on that? So if the underlying source has already been read, discarding won't throw the exception, even if the entity is not `Strict`.
That makes sense. But now I am not so sure what exactly determines whether the exception is thrown.

`Source.empty` passed into an `HttpEntity.apply` won't do it.
`Source.single` would be your above example.
Even if I pass in a `Source` from an infinite `Stream`, I don't seem to see this exception, instead the discarding will just go on forever :)

Sorry but I am still missing a vital piece of understanding :(

ktoso: >  So if the underlying source has already been read, discarding won't throw the exception, even if the entity is not Strict. That makes sense. 

No, that's a case where it *would* throw. The thing is: if it's a real stream, you can materialize it once, because if you had to materialize it twice but expect "to get the data from the beginning of the stream" that means you're forced to buffer the entire thing in memory - otherwise you would not be able to send the first bytes, someone has already consumed them. This is the reason the double materialization is an issue – either you always have to buffer things, and be ready for multiple materializations or you don't buffer, and multiple ones would not get the same data so we prefer to just fail rather than send partial entity bytes. The same applies for any real streaming server, not just akka.

We have something in the works that will buffer "until a given point" (memory size), but again, it's the same trade off of real streaming vs. you wanting to materialize many times, which means it can't be completely fully streaming (since the data source could be a live tcp connection etc).

> Source.empty passed into an HttpEntity.apply won't do it.

This is special, "empty" in general is special most of the time ;)

> Source.single would be your above example.

Again, here the thing is in memory and is just a normal source, this would materialize multiple times just fine. A REAL source from a REAL tcp connection would not.

> Even if I pass in a Source from an infinite Stream, I don't seem to see this exception, instead the discarding will just go on forever :)

Again, you pass them in as "normal" Sources, they can be run as many times as you want. A "Source" that is connected to something "live" would not be able to.

ktoso: Hope this explains it somewhat :)

NeQuissimus: Ah, I think I get it now. Thank you very much. So, if I wanted to whip up a unit test against my own code to make sure a function properly discards entities, I would actually need to connect to some sort of "outside" `Source` rather than quickly whipping up one as a sort of "mock object".

I will update the wording and add to the Java DSL some time today.

ktoso: Yeah, exactly - it's the property of that "source of data" that introduces this issue here.

Thanks a ton! :)

NeQuissimus:
@ktoso Updated, added Javadocs and found another instance of the same Scaladoc :D

PR #1437, opened 2017-09-22
Make host-level-streamed-example more complete

raboof:


ktoso:
CLA Validator seems to be asleep

PR #1441, opened 2017-09-27
Needs to be changed some code

SHUB9914:
The existing code of the example is not working, I have changed something so please review and confirm it.
First thing Here is creating main by extending App within another object so when we run it, it will through an error which is mentioned below  - 
Error: Could not find or load main class docs.http.scaladsl.server.MyExplicitExceptionHandler.MyApp

and the second thing is no need to define the object within object if we do so then application terminates the execution

### DIFF
--- a/docs/src/test/scala/docs/http/scaladsl/server/ExceptionHandlerExamplesSpec.scala
+++ b/docs/src/test/scala/docs/http/scaladsl/server/ExceptionHandlerExamplesSpec.scala
@@ -24,16 +24,13 @@ object MyExplicitExceptionHandler extends App {
         complete(HttpResponse(InternalServerError, entity = "Bad numbers, bad result!!!"))
       }
   }
-
     implicit val system = ActorSystem()
     implicit val materializer = ActorMaterializer()
-
     val route: Route =
       handleExceptions(myExceptionHandler) {
         // ... some route structure
         null // hide
       }
-
     Http().bindAndHandle(route, "localhost", 8080)
   //#explicit-handler-example
 }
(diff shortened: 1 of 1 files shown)


jrudolph:
OK TO TEST

docs/src/test/scala/docs/http/scaladsl/server/ExceptionHandlerExamplesSpec.scala
```diff
@@ -25,7 +25,6 @@ object MyExplicitExceptionHandler {
       }
   }
 
-  object MyApp extends App {
     implicit val system = ActorSystem()
```
jrudolph: Can you also fix the indentation?

jrudolph:
@SHUB9914 this code was not supposed to be run verbatim but only existed there that the basic usage is compiled before being shown in the documentation. But if it is as easy to be completely copy-paste-able, then why not :)

Could you fix the indentation where you removed the extra wrapping object?

SHUB9914:
hi @jrudolph 
I have fixed the indentation

docs/src/test/scala/docs/http/scaladsl/server/ExceptionHandlerExamplesSpec.scala
jrudolph: Ah I meant removing that extra space characters in this block (and another one below) that formerly belonged to `MyApp`.

SHUB9914:
removed extra spaces

jrudolph:
If you look at the resulting source code, lines are still a bit misindented because you removed one layer of wrapping:

https://github.com/SHUB9914/akka-http/blob/ac965732e558a670ddfa0b3af30c5c87177c0c85/docs/src/test/scala/docs/http/scaladsl/server/ExceptionHandlerExamplesSpec.scala#L9-L36

raboof:
There's also a merge conflict now.. #1465 seems to have solved the same problem in a slightly different way, I'll merge that one in favor of this one. Nonetheless thanks for looking into this, much appreciated!

PR #1442, opened 2017-09-27
Change Typesafe to Lightbend in docs.

satansk:


raboof:
OK TO TEST

docs/src/main/paradox/scala/http/configuration.md
```diff
@@ -1,6 +1,6 @@
 # Configuration
 
-Just like any other Akka module Akka HTTP is configured via [Typesafe Config](https://github.com/typesafehub/config).
+Just like any other Akka module Akka HTTP is configured via [Lightbend Config](https://github.com/typesafehub/config).
```
jonas: I've never saw this library branded as "Lightbend Config". Akka [uses](https://doc.akka.io/docs/akka/snapshot/scala/general/configuration.html):

> Akka uses the Typesafe Config Library, which might also be a good choice for the configuration of your own application or library built with or without Akka. This library is implemented in Java with no external dependencies; ...

I suggest keeping as it or copy the wording of the Akka docs.

satansk: Oh, sorry for that, it's a mistake. 

In fact, I want to change "Typesafe SSL-Config" to "Lightbend SSL-Config", because they are mixed in [Client-Side HTTPS Support](http://doc.akka.io/docs/akka-http/10.0.10/scala/http/client-side/client-https-support.html), specifically where are two "Typesafe SSL-Config" and two "Lightbend SSL-Config". It might be better to be consistent. How do you think?

satansk:
Hi @jonas , maybe this PR should be closed, as It does not make sense?

jonas:
Yes, I think we can close this one.

PR #1444, opened 2017-09-28
+htp #1205 Time out requests while they are waiting on the queue

### LINKED ISSUE #1205: Cancelling requests from client connection pools with a timeout
Putting a timeout for each request can be very important in cases where the server we are calling responds very slowly occasionally (for example on its error paths because of some wrongly configured timeouts on that side - but let's assume we can't control the server we are talking to). 

I was trying to find a way to achieve this while using the *host connection pool* without much success. I uploaded the latest one of the failed attempts with a test server as well to https://github.com/vigoo/akka-http-client-cancellation.

Digging more into how the library works I realized I can implement this if I'm directly creating outgoing connections like in the following simplified example:

```scala
val connectionFlow = Http()
       .outgoingConnection(request.uri.authority.host.toString(), request.uri.effectivePort)

val responseFuture = Source
      .single(request)
      .via(connectionFlow)
      .idleTimeout(timeout)
      .runWith(Sink.head)
```
but to use it in production I'd have to build my custom connection pool around it which I'd like to avoid if possible.

So I checked the connection pool implementation and added a similar timeout to the connection flow in `PoolSlot`, tried it out and it works find in the use case I described. Here is this initial implementation:
https://github.com/vigoo/akka-http/commit/e2559cba62f46cabb78f8c7cbc7617ab197c4d8f

If this is something we could add to the library, I'm happy to work on it more and make it a proper PR. Otherwise please give me advice on how else could I solve this issue.

jypma:
Currently, in a situation of a high idle-timeout, a large
max-open-requests queue, and a troubled slowly-responding server, the
queue is going to fill up. Once the server recovers, potentially very
old requests are going to be sent (much more than idle-timeout).

This commit makes no request wait longer than idle-timeout on the queue,
since the request WOULD have timed out by then anyways, if a connection
had been available.

Fixes #1205.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolInterfaceActor.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolInterfaceActor.scala
@@ -21,9 +21,12 @@ import akka.stream.{ BufferOverflowException, Materializer }
 import scala.annotation.tailrec
 import scala.concurrent.Promise
 import scala.concurrent.duration.FiniteDuration
+import scala.concurrent.duration.Deadline
+import scala.concurrent.duration._
+import java.util.concurrent.TimeoutException
 
 private object PoolInterfaceActor {
-  final case class PoolRequest(request: HttpRequest, responsePromise: Promise[HttpResponse]) extends NoSerializationVerificationNeeded
+  final case class PoolRequest(request: HttpRequest, responsePromise: Promise[HttpResponse], deadline: Option[Deadline]) extends NoSerializationVerificationNeeded
 
   case object Shutdown extends DeadLetterSuppression
 
@@ -47,6 +50,8 @@ private object PoolInterfaceActor {
 
       override def getClazz(t: PoolGateway): Class[_] = classOf[PoolGateway]
     }
+
+  private case object Prune
 }
 
 /**
@@ -85,9 +90,17 @@ private class PoolInterfaceActor(gateway: PoolGateway)(implicit fm: Materializer
       "Please retry the request later. See http://doc.akka.io/docs/akka-http/current/scala/http/client-side/pool-overflow.html for " +
       "more information.")
 
+  private[this] val PoolTimeoutException = new TimeoutException(
+    s"Request timed out while being queued in pool ${gateway.hcps} after having waited " +
+      s"${gateway.hcps.setup.settings.idleTimeout} without being sent to a connection. "
+      + "Increase idle-timeout to avoid this.")
+
   log.debug("(Re-)starting host connection pool to {}:{}", hcps.host, hcps.port)
 
   initConnectionFlow()
+
+  import context.dispatcher
+  val pruneSubscription = context.system.scheduler.schedule(1.second, 1.second, self, Prune)
 
   /** Start the pool flow with this actor acting as source as well as sink */
   private def initConnectionFlow() = {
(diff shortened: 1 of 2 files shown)


jypma:
@akka/akka-http-team tests green, ready for review.

2m:
Seems clean enough. I wonder how hard would it be to write a test that tests this. We could increase all other timeouts to make sure that request is timeout out in the queue and not somewhere else.

jypma:
I decided to re-use the existing timeout mostly for simplicity's sake (understanding client timeouts already is hard enough). We could introduce something like `queue-timeout`, and update the timeout docs page. But really, [this](https://doc.akka.io/docs/akka-http/current/java/http/client-side/host-level.html#configuring-a-host-connection-pool) needs a rewrite as well, describing how `max-open-requests` is _actually_ configuring a queue.

Having a different timeout would of course make it easier to write @2m 's suggested test...

jypma:
I'm working on a test and introduced the separate `queueTimeout`.

I think I have a problem though: `PoolInterfaceActor` always gets an initial demand of 16 items from the downstream `PoolConductor`, irrespective of how many connections are actually available to dispatch requests on. Is there a way to configure that initial demand between the two  graph stages?

raboof:
PLS BUILD

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
```diff
@@ -57,6 +60,11 @@ private[http] final class PoolMasterActor extends Actor with ActorLogging {
     context.watch(ref)
   }
 
+  private def createDeadline(gateway: PoolGateway): Option[Deadline] = gateway.hcps.setup.settings.idleTimeout match {
+    case finite: FiniteDuration ⇒ Some(finite.fromNow)
```
raboof: Not sure if looking at the clock for each request might be expensive?

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolInterfaceActor.scala
```diff
@@ -174,6 +187,16 @@ private class PoolInterfaceActor(gateway: PoolGateway)(implicit fm: Materializer
         onCompleteThenStop()
       } else
         debug(s"Deferring shutting down host connection pool until all [$remainingRequested] responses have been dispatched")
+
+    case Prune ⇒ // Kick all expired requests out of the queue and fail them
+      while (inputBuffer.nonEmpty && inputBuffer.peek().deadline.filter(!_.hasTimeLeft()).isDefined) {
```
raboof: Perhaps `!inputBuffer.peek().deadline.forall(_.hasTimeLeft())`?

Nice that this can just look at the side of the buffer that has the oldest elements 👍

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
raboof: Probably OK since I just read `Deadline` uses `System.nanoTime`, that should be nice and fast IIRC

jypma:
Thanks for the extra review! I'm happy to address the remaining comments. However, @jrudolph, weren't you doing some re-write of the client pool stages that would make this irrelevant? Or am I totally making that up...

raboof:
@jrudolph I don't think we expect conflicts here, do we?

jypma:
I should note that this PR does not (fully) work: Because of the 16-item internal akka streams buffer, there will still be 16 items queued up that we can't purge. And though that buffer theoretically can be lowered to 1, I wasn't able to do that accurately without changing the global buffer size.

Plus, going from "we can't purge requests while on the queue" to "we can purge _some_ requests from the queue" may not help with the predictability of the http client... :)

PR #1457, opened 2017-10-06
Hide non-informative lines from snippets

2m:
With paradox code lines from snippets are being hidden by adding anything that starts with `#` to the line (see [here](https://github.com/lightbend/paradox/issues/157)). This fixes the issue where these non-informative lines where being included in code snippets like [here](https://doc.akka.io/docs/akka-http/10.0.10/scala/http/routing-dsl/case-class-extraction.html).


jrudolph:
Good to know. Thanks for the fix, @2m.

PR #1462, opened 2017-10-10
Include Akka's CoC

ktoso:
We've introduced it in Akka, to be on par with other Lightbend projects which have them already some time ago. Adding it here since Akka HTTP abides to the same guidelines.

The full rationale is:

In order to strengthen the welcoming nature of our community.

Please note that this addition is not in direct reaction to any specific incident; we've been blessed with a very nice and polite community, and have in the past handled a few situations without having to call out a rule set like this - thank you everyone :-) However, we do think stating it explicitly will make more people confident and comfortable in participating in our community - similarly as Scala and Play have done in their communities. It will also show up in the github community tab https://github.com/akka/akka/community and other such places.

This document does not change rules of engagement – please continue to be as nice and polite as you've always been in all online and real-life communication, thanks :-)

Refers to Akka's PR with it https://github.com/akka/akka/pull/23673

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,33 @@
+# Code of Conduct
+
+This Code of Conduct governs how we behave in any forum and whenever we will be judged by our actions. We expect it to be honored by everyone who represents the Akka community officially or informally, claims affiliation with the project or participates directly.
+
+We strive to:
+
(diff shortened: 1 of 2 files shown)


ktoso:
I think it's rather uncontroversial, as it basically spells out explicitly various rules that we've always followed in this community, and Akka, Play, Lagom and Scala all have adopted a similar wording to this. Please :+1: and LGTM as usual, and I'll want to get a nice bunch of those ACKs before merging, to make sure we're all good on these. 

Thanks 👍

ktoso:
Yes, good point, I'll do that tomorrow then before merging :)

ktoso:
Seems enough consensus and I added the references

PR #1464, opened 2017-10-10
In top-level API don't require materializer where it's just ignored afterwards

jrudolph:
It seems we had lots of these left over I guess from the early days where creating substreams required materializers to be passed. It seems no one cleaned that up so far. The changes were somewhat big to keep binary compatibility at least for a while. In Scala, it's binary and source compatible (if the materializer was passed implicitly). Java users need to remove the materializer argument.

Going forward, if we deprecate the old methods now and release another 10.0.x in October, we might think about cleaning those methods up for 10.1.0.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpAttributes.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpAttributes.scala
@@ -20,9 +20,15 @@ private[akka] object HttpAttributes {
 
   private[akka] final case class RemoteAddress(address: InetSocketAddress) extends Attribute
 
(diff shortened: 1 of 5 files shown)


jrudolph:
@raboof yes, that's something I also thought about. For all of the shared pool APIs it just doesn't make sense to pass a materializer because it would just be unclear what it would mean if different invocations targeting the same shared pool would get different materializers.

For `serverLayer`, we can hope that any change that would require a materializer can be implemented by using a `subFusingMaterializer`.

jrudolph:
Btw the reason I looked into this was that I saw code somewhere with multiple ActorMaterializers calling those shared pool methods and I wondered what would happen.

PR #1466, opened 2017-10-10
=htc slight logging tag improvement in wire logging stage

ktoso:
Small improvement in logging so it's more understandable.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/util/LogByteStringTools.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/util/LogByteStringTools.scala
@@ -24,11 +24,11 @@ private[akka] object LogByteStringTools {
       logByteString(s"$name UP  ", maxBytes)
     )
 
-  def logToStringBidi[A, B](name: String, maxBytes: Int = MaxBytesPrinted): BidiFlow[A, A, B, B, NotUsed] = {
-    def limitedName[T](tag: ClassTag[T]) = Logging.simpleName(tag.runtimeClass).take(10).padTo(10, " ")
+  def logToStringBidi[A: ClassTag, B: ClassTag](name: String, maxBytes: Int = MaxBytesPrinted): BidiFlow[A, A, B, B, NotUsed] = {
+    def limitedName[T](implicit tag: ClassTag[T]) = Logging.simpleName(tag.runtimeClass).take(10).padTo(10, " ")
     BidiFlow.fromFlows(
-      logToString(s"$name DOWN(${limitedName(implicitly[ClassTag[A]])}", maxBytes),
-      logToString(s"$name UP  ${limitedName(implicitly[ClassTag[A]])}", maxBytes)
+      logToString(s"$name [${limitedName[A]}]", maxBytes),
+      logToString(s"$name [${limitedName[B]}]", maxBytes)
     )
   }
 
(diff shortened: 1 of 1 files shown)


akka-http-core/src/main/scala/akka/http/impl/util/LogByteStringTools.scala
```diff
@@ -17,14 +17,14 @@ private[akka] object LogByteStringTools {
 
   def logByteStringBidi(name: String, maxBytes: Int = MaxBytesPrinted): BidiFlow[ByteString, ByteString, ByteString, ByteString, NotUsed] =
     BidiFlow.fromFlows(
-      logByteString(s"$name DOWN", maxBytes),
-      logByteString(s"$name UP  ", maxBytes)
+      logByteString(s"$name FromNet", maxBytes),
```
ktoso: things were inversed here; since the generic one just used up / down, but up/down "mean things" for http, so let's use more specific names here

jrudolph: I think it doesn't make sense to change it for `logByteStringBidi` because it might then just be wrong (where before it was arbitrary), depending on the direction in which the stage is used.

akka-http-core/src/main/scala/akka/http/impl/util/LogByteStringTools.scala
```diff
@@ -35,8 +35,8 @@ private[akka] object LogByteStringTools {
 
   def logTLSBidi(name: String, maxBytes: Int = MaxBytesPrinted): BidiFlow[SslTlsOutbound, SslTlsOutbound, SslTlsInbound, SslTlsInbound, NotUsed] =
     BidiFlow.fromFlows(
-      logTlsOutbound(s"$name DOWN", maxBytes),
-      logTlsInbound(s"$name UP  ", maxBytes))
+      logTlsOutbound(s"$name FromNet", maxBytes),
```
jrudolph: Here it needs to be swapped  around ;) Thanks to `TlsInbound` / `TlsOutbound` it is quite clear which direction is which.

akka-http-core/src/main/scala/akka/http/impl/util/LogByteStringTools.scala
```diff
@@ -17,14 +17,14 @@ private[akka] object LogByteStringTools {
...
     BidiFlow.fromFlows(
-      logByteString(s"$name DOWN", maxBytes),
-      logByteString(s"$name UP  ", maxBytes)
+      logByteString(s"$name FromNet", maxBytes),
+      logByteString(s"$name ToNet  ", maxBytes)
     )
 
   def logToStringBidi[A, B](name: String, maxBytes: Int = MaxBytesPrinted): BidiFlow[A, A, B, B, NotUsed] =
     BidiFlow.fromFlows(
-      logToString(s"$name DOWN", maxBytes),
```
jrudolph: Same as for `logByteStringBidi`, generic names were better here as we don't know how it is used.

jrudolph: Or, you could add a `ClassTag` for `A` and `B` and use some string from those here.

akka-http-core/src/main/scala/akka/http/impl/util/LogByteStringTools.scala
ktoso: Hah you're right, ctrl+r-ing this change was a bad idea, thanks for checking

akka-http-core/src/main/scala/akka/http/impl/util/LogByteStringTools.scala
ktoso: good idea, done, with limiting it to some chars

PR #1477, opened 2017-10-12
Add NoSuchElementException without stack trace being filled in

Abrasha:
Also this PR fixes one TODO with creating exception with suppressed stack trace:

    if (pred(r)) future
    else throw new NoSuchElementException("Future.filter predicate is not satisfied") // FIXME: avoid stack trace generation

NOTE: I am not sure how to properly name this exception.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/util/FastFuture.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/util/FastFuture.scala
@@ -28,7 +28,7 @@ class FastFuture[A](val future: Future[A]) extends AnyVal {
   def filter(pred: A ⇒ Boolean)(implicit executor: ExecutionContext): Future[A] =
     flatMap { r ⇒
       if (pred(r)) future
-      else throw new NoSuchElementException("Future.filter predicate is not satisfied") // FIXME: avoid stack trace generation
+      else throw new NoSuchElementException("Future.filter predicate is not satisfied")
     }
 
   def foreach(f: A ⇒ Unit)(implicit ec: ExecutionContext): Unit = map(f)


akka-http-core/src/main/scala/akka/http/scaladsl/util/FastFuture.scala
```diff
@@ -111,7 +111,7 @@ object FastFuture {
     else sequence(futures).fast.map(_.foldLeft(zero)(f))
 
   def reduce[T, R >: T](futures: TraversableOnce[Future[T]])(op: (R, T) ⇒ R)(implicit executor: ExecutionContext): Future[R] =
-    if (futures.isEmpty) failed(new NoSuchElementException("reduce attempted on empty collection"))
+    if (futures.isEmpty) failed(new scala.NoSuchElementException("reduce attempted on empty collection"))
```
Abrasha: Please advice me of we have to suppress stack trace here also.

akka-http-core/src/main/scala/akka/http/scaladsl/util/FastFuture.scala
```diff
@@ -28,7 +28,7 @@ class FastFuture[A](val future: Future[A]) extends AnyVal {
   def filter(pred: A ⇒ Boolean)(implicit executor: ExecutionContext): Future[A] =
     flatMap { r ⇒
       if (pred(r)) future
-      else throw new NoSuchElementException("Future.filter predicate is not satisfied") // FIXME: avoid stack trace generation
+      else throw new NoSuchElementException("Future.filter predicate is not satisfied")
```
2m: You can use the Scala's built-in [`NoStackTrace`](https://github.com/scala/scala/blob/v2.12.3/src/library/scala/util/control/NoStackTrace.scala#L22), so this becomes `new NoSuchElementException(...) with NoStackTrace`.

akka-http-core/src/main/scala/akka/http/scaladsl/util/FastFuture.scala
2m: Yes, sounds good.

Abrasha:
Hi @2m. Thank you for reviewing / suggestions!

I have updated this PR.

jrudolph:
PLS BUILD

jrudolph:
OK TO TEST

jrudolph:
Almost merged, but on second thought I'm not so sure any more. (Not your fault, @Abrasha, it's about whether the TODO was accurate or not).

As this is `FastFuture` the exceptions might contain a useful stack trace if they were generated in the fast path.

So, either we say, `FastFuture` is about "fast" and then we say it's fine that it reports the same crappy exception stack traces as regular futures. Or, we say that error cases don't have to be fast and instead keep providing the rich stack traces they offer. Does anyone else have an opinion on this?

viktorklang:
@jrudolph Or even better, how can we remove the need for FastFuture? See this for work-in-progress: https://github.com/viktorklang/scala-futures/tree/wip-optimizations-√

ktoso:
I'd lean to the side of:

> Or, we say that error cases don't have to be fast and instead keep providing the rich stack traces they offer. Does anyone else have an opinion on this?

that failure cases being more helpful rather than the-fastest is likely more important, making the TODO not really accurate. (Not your fault @Abrasha, thanks for the PR though!)

Abrasha:
Okay. I will just remove TODO

ktoso:
Sounds good :)

-- 
Konrad Malawski

Abrasha:
@ktoso @jrudolph Done. I have simply removed this FIXME as per this discussion.

I have one more question:

Are the any specific reasons to perform this casting?

screen shot 2017-10-30 at 09 59 07
screen shot 2017-10-30 at 09 59 02

ktoso:
If things indeed works fine and does not leak types without the casts then cool ;) Try it

Abrasha:
Well, seems that IDE suggests changes that are not valid. Code does not compile.
Let's close this one with removing redundant `FIXME`.

screen shot 2017-10-30 at 10 29 29

jrudolph:
PLS BUILD

jrudolph:
Thanks, @Abrasha. Every change counts even the really small ones ;)

PR #1480, opened 2017-10-14
Added withIdleTimeout and withoutIdleTimeout directives (#71)

catalin-ursachi:
I've written a prototype for a solution to https://github.com/akka/akka-http/issues/71; would anyone be able to check whether this is basically sound, to see if it's worth implementing this in full? 🙂


catalin-ursachi:
Okay, I've written what (I think) is a valid implementation of the feature. Note that:

* I've moved the idle timeout stage into the `HttpServerBluePrint`, just after the "parsing/rendering" stage; is that equivalent to having it next to the TCP layer? If not, is it fine anyway?
* I passed in the timeout override as a non-renderable header; would there be a better way?
* It all feels a bit messier than it was to start with, generally.

raboof:
OK TO TEST

catalin-ursachi:
Should I add the exclusions to `akka-http\src\main\mima-filters\10.0.10.backwards.excludes`, or create a new `10.0.11.backwards.excludes` file?

catalin-ursachi:
Ah, that's a shame; but I agree, this does seem like a bit of a heavy change to the stream for a relatively limited-use feature.

> An alternative approach could be for users to disable the idle timeout or set it to a high value and then add their own logic to the request and response stream processing to reenable it where needed. Wouldn't that work as well?
I don't this it's necessarily easy for users to do that, for the same reason it was tricky to do here: the idle timeout has to be managed next to the TCP layer, but the only place where you could decide to privilege some routes over others is the Request/Response handling layer.

In any case, do you think this PR can be closed, then? 🙂

catalin-ursachi:
Hi @raboof; not sure what action there is for me, here?

raboof:
Ah, sorry, misread that last comment. Tricky, might indeed best to keep things as they are and close this PR then... thanks for taking a stab at this though!

catalin-ursachi:
Well, it was a fun exercise anyway! 🙂

catalin-ursachi:
@raboof - So should I go ahead and close this PR and the associated ticket?

PR #1483, opened 2017-10-17
Note about same issue using the testkit libraries.

angelcervera:
Adding  the "com.typesafe.akka" %% "akka-http-testkit" in test scope in your project, automatically add the akka-stream-testkit transitional dependency, but the version 2.4
To avoid, it is necessary to add the right version for this library.

### DIFF
--- a/docs/src/main/paradox/scala/http/compatibility-guidelines.md
+++ b/docs/src/main/paradox/scala/http/compatibility-guidelines.md
@@ -38,6 +38,10 @@ sbt
     libraryDependencies += "com.typesafe.akka" %% "akka-http"   % akkaHttpVersion
     libraryDependencies += "com.typesafe.akka" %% "akka-actor"  % akkaVersion
     libraryDependencies += "com.typesafe.akka" %% "akka-stream" % akkaVersion
+    
+    // If testkit used, explicitly declare dependency on akka-streams-testkit in same version as akka-actor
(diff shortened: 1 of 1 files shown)


angelcervera:
I signed the CLA

jrudolph:
OK TO TEST

docs/src/main/paradox/scala/http/compatibility-guidelines.md
```diff
@@ -71,3 +71,39 @@ Maven
     </dependency>
     ```
     @@@
+    
+The same issue using `akka-http-testkit` library. So in this case, it is necessary to add:
```
jrudolph: Maybe we could just add this information to the above snippet and just add comments that you only need to add if you want to add the testkit. WDYT?

angelcervera: Agee. I will do it today.

angelcervera:
@jrudolph  @2m  Do you know why the paradox generator is now generating correctly the site? Tabs are broken after the change. Maybe the empty line before the test scope dependencies?

2m:
Thanks for checking! It seems that empty lines in the ``` block are not parsed correctly. I have created a ticket in the paradox issue tracker: https://github.com/lightbend/paradox/issues/161

In the meantime could you create a PR that removes those three empty lines?

angelcervera:
Done: https://github.com/akka/akka-http/pull/1487

PR #1487, opened 2017-10-23
Remove empty lines because Pradox bug

angelcervera:
Because a BUG in https://github.com/lightbend/paradox/issues/161 , the empty line is breaking the format.
Previous change was related with https://github.com/akka/akka-http/pull/1483


jrudolph:
OK TO TEST

PR #1435, opened 2017-09-20
Update build to use sbt 1.0

jonas:
Fixes #1434.

Includes the unidoc update changes from #925 which was reverted in #940 since it broke the  integrated build with akka/akka.

`sbt-jol` is removed based on https://github.com/akka/akka-http/issues/1434#issuecomment-331334611

PR #1489, opened 2017-10-24
Upgrade scalariform + use same formatting as akka

### LINKED ISSUE #1474: Building against akka/akka master fails
A bit weird:

https://jenkins.akka.io:8498/job/akka-http-nightly-against-akka-master-on-2.12/294//consoleFull

```
[error] /home/jenkins/.sbt/0.13/staging/8bd016127453b4b8fefa/akka/project/Formatting.scala:32: not found: value DoubleIndentConstructorArguments
[error]     .setPreference(DoubleIndentConstructorArguments, false)
[error]                    ^
[error] one error found
[error] (compile:compileIncremental) Compilation failed
```

chbatey:
Should fix #1474


project/Formatting.scala
```diff
@@ -9,35 +9,28 @@ import com.typesafe.sbt.SbtScalariform
...
-  def docFormattingPreferences = Def.setting {
-    import scalariform.formatter.preferences._
-    ScalariformKeys.preferences.value
-      .setPreference(RewriteArrowSymbols, false)
-      .setPreference(AlignParameters, true)
-      .setPreference(AlignSingleLineCaseStatements, true)
-      .setPreference(DanglingCloseParenthesis, Preserve)
-      .setPreference(DoubleIndentClassDeclaration, false)
-  }
+  def setPreferences(preferences: IFormattingPreferences, rewriteArrowSymbols: Boolean = true) = preferences
```
raboof: I'm leaning towards setting `rewriteArrowSymbols = true` for both the docs and the regular subprojects - but that'd have to be updated in akka also then ofc.

akka-http-core/src/main/scala/akka/http/impl/util/Rendering.scala
```diff
@@ -373,4 +373,4 @@ private[http] class CustomCharsetByteStringRendering(nioCharset: Charset, sizeHi
     }
     charBuffer.clear()
   }
-}
\ No newline at end of file
+}
```
raboof: Strange formatting in github - does this file have strange line endings? Not really related to this PR though

akka-http-core/src/main/scala/akka/http/impl/model/parser/IpAddressParsing.scala
```diff
@@ -24,10 +24,10 @@ private[parser] trait IpAddressParsing { this: Parser ⇒
   def `ip-v6-address`: Rule1[Array[Byte]] = {
     import CharUtils.{ hexValue ⇒ hv }
     var a: Array[Byte] = null
-    def zero(ix: Int) = rule { run(a(ix) = 0.toByte) }
+    def zero(ix: Int) = rule { run(a(ix)= 0.toByte) }
```
raboof: Strange... but not too bad I guess.

ktoso: hm indeed a bit weird... looks like bug? or weird setting?

project/Formatting.scala
ktoso: Would work for me, I like the utf symbols

akka-http-core/src/main/scala/akka/http/impl/model/parser/IpAddressParsing.scala
jrudolph: Yeah, looks like a bug. But there are more...

akka-http-core/src/main/scala/akka/http/scaladsl/model/Uri.scala
```diff
@@ -795,9 +795,9 @@ object Uri {
         case Slash(Segment("..", tail)) ⇒ process(
           input = if (tail.isEmpty) Path./ else tail,
           output =
-          if (output.startsWithSegment)
-            if (output.tail.startsWithSlash) output.tail.tail else tail
-          else output)
+            if (output.startsWithSegment)
```
jrudolph: Here a bug is fixed, good :+1:

akka-http-core/src/test/scala/akka/http/impl/engine/rendering/ResponseRendererSpec.scala
```diff
@@ -576,9 +576,9 @@ class ResponseRendererSpec extends FreeSpec with Matchers with BeforeAndAfterAll
         ResponseRenderingContext(
           response = HttpResponse(200, headers = resCH.toList,
             entity = if (resCD) HttpEntity.CloseDelimited(
-            ContentTypes.`text/plain(UTF-8)`,
-            Source.single(ByteString("ENTITY")))
-          else HttpEntity("ENTITY"), protocol = resProto),
+              ContentTypes.`text/plain(UTF-8)`,
+              Source.single(ByteString("ENTITY")))
+            else HttpEntity("ENTITY"), protocol = resProto),
```
jrudolph: New indentation seems (arguably) right.

akka-http-core/src/main/scala/akka/http/impl/model/parser/IpAddressParsing.scala
jrudolph: Aah, read the diff the wrong way round for the others... So, this seems to be the only problem I can see so far.

jrudolph:
(Obviously it's supposed to be only whitespace changes, but ones that don't move things from one line to another) ;)

project/Formatting.scala
jonas: Using the same set of preferences would fix one of the major head-aches I ran into with the [sbt-1.0 PR](https://github.com/akka/akka-http/pull/1435).

_Programming ligatures or editor plugins FTW!_

akka-http-core/src/main/scala/akka/http/impl/util/Rendering.scala
jonas: It has this line which is more than GitHub's Scala highlighter can manage:

```
val `\"` = CharPredicate('\\', '"')
```

jrudolph:
I reran the nightly builds after this hitting master and it now is green (i.e. yellow) again.

Issue #1492, opened 2017-10-24
Allow to override "User-Agent" header for client-side with application.conf


apehovski:
Custom value for key "akka.http.client.user-agent-header" currently does not override default value.
As I see, also "akka.http.parsing" values affected.

Minor pull-request:
https://github.com/akka/akka-http/pull/1493

jrudolph:
There's indeed a problem here but I'd say it's a bit different:

 * currently, for pool usages, `akka.http.host-connection-pool.client` settings override the general client settings but fall back when missing (which is how it should be)
 * however, `reference.conf` contains a copy of the client settings (which we introduced, I think, to make the reference.conf more readable). That means, however, that `akka.http.client` settings never apply to the pool because we provide this copy in the `reference.conf`.

In fact, iirc, this whole setup was only done, to be able to provide a self-contained version of `host-connection-pool` settings which needs to wrap a set of client settings. Having to create self-contained versions of these settings seems to be a rare case, though.

The solution seems to be, to remove all concrete settings from `akka.http.host-connection-pool.client`, so that in general you can set settings just in `akka.http.client` and only have to use `akka.http.host-connection-pool.client` in special cases.

`akka.http.{client,server}.parsing` already uses the same style.

PR #1493, opened 2017-10-24
Prefer custom config values instead default ones #1492

apehovski:


raboof:
OK TO TEST

raboof:
Refs #1492

jrudolph:
Hmm, I think we should consider a different solution. I guess we might want to remove the concrete values in the `akka.http.host-connection-pool` section of `reference.conf` so that in the default case the fallbacks are used.

apehovski:
About fix - maybe it is not correct way to handle it, since I'm newcomer in architecture of akka-http :) 
But for checking possible issue I've made simple project:
https://github.com/apehovski/akka-http-client-sample

To check, run:
`python server.py`
(simple server to see incoming headers)
`sbt run`

Output of running python server for me:
```
User-Agent: akka-http/10.0.10
127.0.0.1 - - [24/Oct/2017 19:56:02] "GET / HTTP/1.1" 200 -
```

Not overridden header. But application.conf contains lines from akka-http docs.
Is it looks like issue or it's my misunderstanding? Thanks in advance for your time!

jrudolph:
My suggestion was to remove everything under `akka.http.host-connection-pool.client` in the `reference.conf` and see if that would fix the issue as well. If yes, then that's the preferred resolution.

jrudolph:
I'm closing this PR for now as there's a better solution as outlined in the comments here and in #1492.

PR #1494, opened 2017-10-24
=doc Link Java server-side docs at the directory level

jonas:
Since all server-side documents are now shared between Java and Scala we can link the Java version at the directory level.


ktoso:
Thanks 👍

PR #1503, opened 2017-10-31
Migrate Spray caching as akka-http-caching #213 

### LINKED ISSUE #213: Offer migration path from Spray Caching

ktoso:
This is @tomrf1's PR #986 with some cleanup on top to make to pass nicely.

My changes are the last commits and basically make some more things internal, and align the settings style with the rest of akka-http, such that they canbe configured from the reference.conf. Though perhaps it's to be debated if that's really needed I thought the consistency with other settings objects at least was important. 

Will resolve #213

**We can change these APIs still since they are ApiMayChange.**


ktoso:
I'll rebase to cleanup some more still.

akka-http-caching/src/main/scala/akka/http/caching/LfuCache.scala
```diff
@@ -5,27 +5,32 @@ package akka.http.caching
...
+  def apply[K, V](implicit system: ActorSystem): akka.http.caching.scaladsl.Cache[K, V] =
+    apply(akka.http.caching.scaladsl.CachingSettings(system))
+
   /**
    * Creates a new [[akka.http.caching.LfuCache]], with optional expiration depending
    * on whether a non-zero and finite timeToLive and/or timeToIdle is set or not.
    */
-  def apply[K, V](settings: LfuCacheSettings = LfuCacheSettings()): akka.http.caching.scaladsl.Cache[K, V] = {
+  def apply[K, V](cachingSettings: CachingSettings): akka.http.caching.scaladsl.Cache[K, V] = {
+    val settings = cachingSettings.lfuCacheSettings.asInstanceOf[LfuCacheSettings]
```
raboof: is `asInstanceOf` needed here?

akka-http-caching/src/main/scala/akka/http/caching/javadsl/CachingSettings.scala
```diff
@@ -0,0 +1,42 @@
...
+import akka.http.caching.scaladsl.{ CachingSettingsImpl, LfuCacheSettingsImpl }
+
+import scala.concurrent.duration.Duration
+
+/**
+ * Public API but not intended for subclassing
+ */
+@DoNotInherit
+abstract class CachingSettings private[http] () { self: CachingSettingsImpl ⇒
+  def lfuCacheSettings: LfuCacheSettings
```
raboof: This seems a bit awkward since we only have `LfuCacheSettings` for now... it does seem the most future-proof thing to do though.

akka-http-caching/src/main/resources/reference.conf
```diff
@@ -0,0 +1,15 @@
...
+
+  # default configuration values for LfuCache
+  lfu-cache {
+
+    max-capacity = 512
+
+    initial-capacity = 16
+
+    time-to-live = infinite
+
```
raboof: I'd personally either use less newlines, or add descriptive comments here.

akka-http-caching/src/main/scala/akka/http/caching/javadsl/CachingSettings.scala
ktoso: Yeah, it's for future proofing... call me paranoid ;)

docs/src/test/scala/docs/http/scaladsl/server/directives/CachingDirectivesExamplesSpec.scala
```diff
@@ -0,0 +1,97 @@
...
+      val isGet: RequestContext ⇒ Boolean = _.request.method == GET
+      val isAuthorized: RequestContext ⇒ Boolean = _.request.headers.exists(_.is(Authorization.lowercaseName))
+      PartialFunction {
+        case r: RequestContext if isGet(r) && !isAuthorized(r) ⇒ r.request.uri
+      }
+    }
+
+    var i = 0
+    val route =
+      cache(routeCache(), simpleKeyer) {
```
tomrf1: This now needs the default `CachingSettings`

jonas:
The caching docs and Java directive tests are now ready for PR. Let me know if I should rebase on top of this PR or the original one.

Main doc is here: https://jonas.github.io/akka-http/caching-docs/java/http/common/caching.html

Some notes I took:

- LfuCache
  - `size` implementation: should it use `estimatedSize`?
  - Use `Long` for `maxCapacity` as is done for Caffeine?
- Cache
  - A Java implementation inheriting from akka.http.caching.javadsl.Cache
    will fail due to `asInstanceOf` usage in CacheJavaMapping (I think that has been fixed)
  - getOrFulfil -> getOrComplete ? (seems to better match the vocabulary surrounding futures in both Java/Scala land.
  - Use Long for `size()` to be more future proof.

docs/src/main/paradox/java/http/routing-dsl/index.md
```diff
@@ -1,6 +1,6 @@
 # High-level Server-Side API
 
-In addition to the @ref[Low-Level Server-Side API](../server-side/low-level-api.md) Akka HTTP provides a very flexible "Routing DSL" for elegantly
+In addition to the @ref[Low-Level Server-Side API](../../../scala/http/server-side/low-level-api.md) Akka HTTP provides a very flexible "Routing DSL" for elegantly
```
jonas: This will break the unified Java/Scala docs. We need to keep links as relative as possible.

docs/src/main/paradox/scala/http/server-side/index.md
```diff
@@ -36,9 +36,9 @@ from a background with non-"streaming first" HTTP Servers.
...
 @@@ index
 
-* [low-level-api](low-level-api.md)
-* [websocket-support](websocket-support.md)
-* [server-https-support](server-https-support.md)
-* [http2](http2.md)
+* [low-level-api](../../../scala/http/server-side/low-level-api.md)
+* [websocket-support](../../../scala/http/server-side/websocket-support.md)
+* [server-https-support](../../../scala/http/server-side/server-https-support.md)
+* [http2](../../../scala/http/server-side/http2.md)
```
jonas: What script created these "denormalized paths"?

docs/src/main/paradox/java/http/routing-dsl/index.md
jonas: This commit should be dropped completely it seems.

akka-http-caching/src/main/java/akka/http/caching/javadsl/Cache.java
```diff
@@ -0,0 +1,67 @@
...
+   * Returns either the cached CompletionStage for the given key or evaluates the given value generating
+   * function producing a `CompletionStage<V>`.
+   */
+  CompletionStage<V> getFuture(K key, Creator<CompletionStage<V>> genValue);
+
+  /**
+   * Returns either the cached {@code CompletionStage} for the key, or evaluates the given function which
+   * should lead to eventual completion of the completable future.
+   */
+  CompletionStage<V> getOrFulfil(K key, Procedure<CompletableFuture<V>> f);
```
jonas: Rename to `getOrComplete` to better match wording used regarding Java/Scala futures.

akka-http-caching/src/main/resources/reference.conf
```diff
@@ -0,0 +1,15 @@
+akka.http.caching {
+
+  # default configuration values for LfuCache
+  lfu-cache {
+
```
jonas:
```
# Maximum number of entries the cache may store.
# After the maximum capacity is reached new entries cause old ones
# to be evicted in a last-frequently-used manner, i.e. the entries that
# are accessed most frequently are kept in the cache.

File: akka-http-caching/src/main/resources/reference.conf [hunk 9]
```diff
@@ -0,0 +1,15 @@
...
+    max-capacity = 512
+
+    initial-capacity = 16
+
```
jonas:
```
# Upper limit to the time period an entry is allowed to remain in the cache.
# Set to 'infinite' to disable eviction based on time.
```

akka-http-caching/src/main/resources/reference.conf
jonas:
```
# Maximum time an entry is kept without having been accessed.
# Set to 'infinite' to disable eviction based on access time.
```

akka-http-caching/src/main/scala/akka/http/caching/CacheJavaMapping.scala
```diff
@@ -0,0 +1,23 @@
...
+@InternalApi
+private[akka] object CacheJavaMapping {
+
+  def cacheMapping[JK, JV, SK <: JK, SV <: JV] =
+    new JavaMapping[Cache[JK, JV], akka.http.caching.scaladsl.Cache[SK, SV]] {
+      def toScala(javaObject: Cache[JK, JV]): akka.http.caching.scaladsl.Cache[SK, SV] =
+        javaObject.asInstanceOf[akka.http.caching.scaladsl.Cache[SK, SV]]
+
+      def toJava(scalaObject: akka.http.caching.scaladsl.Cache[SK, SV]): Cache[JK, JV] =
+        scalaObject.asInstanceOf[Cache[JK, JV]]
```
jonas: Due to the casts here I put this note in the caching docs:

> Akka HTTP cache implementations must currently implement both the Java and Scala cache interface. As such, you are currently encouraged to implement new caches in Scala.

Is my understanding correct?

akka-http-caching/src/main/scala/akka/http/caching/LfuCache.scala
```diff
@@ -0,0 +1,115 @@
...
+   */
+  def create[K, V](system: ActorSystem): akka.http.caching.javadsl.Cache[K, V] =
+    apply(system)
+
+  /**
+   * Java API
+   * Creates a new [[akka.http.caching.LfuCache]], with optional expiration depending
+   * on whether a non-zero and finite timeToLive and/or timeToIdle is set or not.
+   */
+  def create[K, V](settings: CachingSettings): akka.http.caching.javadsl.Cache[K, V] =
```
jonas: 👍 this Java API looks better.

akka-http-caching/src/main/scala/akka/http/caching/LfuCache.scala
```diff
@@ -0,0 +1,115 @@
...
+    def tti: Caffeine[K, V] ⇒ Caffeine[K, V] = { builder ⇒
+      if (timeToIdle.isFinite) builder.expireAfterAccess(timeToIdle.toMillis, TimeUnit.MILLISECONDS)
+      else builder
+    }
+
+    val builder = Caffeine.newBuilder().asInstanceOf[Caffeine[K, V]]
+      .initialCapacity(initialCapacity)
+      .maximumSize(maxCapacity)
+
+    val store = (ttl andThen tti)(builder).buildAsync[K, V](dummyLoader[K, V])
```
jonas: It should be possible to get rid of `simpleLfuCache` altogether since `expiringLfuCache` "degrades" based on the durations.

akka-http-caching/src/main/scala/akka/http/caching/CacheJavaMapping.scala
jonas: Just now noticed the `@DoNotInherit` on the `Cache` API. Hmm.

akka-http-caching/src/main/scala/akka/http/caching/scaladsl/CachingSettings.scala
```diff
@@ -0,0 +1,70 @@
...
+@InternalApi
+private[http] final case class LfuCacheSettingsImpl(
+  maxCapacity:     Int,
+  initialCapacity: Int,
+  timeToLive:      Duration,
+  timeToIdle:      Duration
+) extends LfuCacheSettings
+
+object CachingSettings extends SettingsCompanion[CachingSettings]("akka.http.caching") {
+  def fromSubConfig(root: Config, c: Config) = {
```
jonas: How hard would it be to support something similar to the Akka dispatcher configuration were you can put load from any path. If I need multiple LFU caches I need to do something like this, right?

```
cache-one = ${akka.http.caching} {
  lfu-cache = { max-capacity = 1, time-to-live = 60 seconds }
}
# and load with: CacheSetings.fromSubConfig(..., config.getConfig("cache-one"))
```

akka-http-caching/src/main/scala/akka/http/javadsl/server/directives/CachingDirectives.scala
```diff
@@ -0,0 +1,74 @@
...
+  /**
+   * Creates an [[LfuCache]]
+   *
+   * Default settings are available via [[akka.http.caching.javadsl.CachingSettings.create]].
+   */
+  def routeCache[K](settings: CachingSettings): Cache[K, RouteResult] =
+    JavaMapping.toJava(D.routeCache[K](settings))
+}
+
+object CachingDirectives extends CachingDirectives
```
jonas: My understanding is we need to drop the `class` and only keep the `object` in the Java DSL since otherwise the static API needs to go though `CachingDirectives$.MODULE$.method`

I would like to raise the point again that if the cache API lives in `akka-http` and the LFU cache implementation lives in `akka-http-caffeine` or `akka-http-caching` then `CachingDirectives` could be mixed into `akka.http.javadsl.Directive` and work similar to the other directives.

akka-http-core/src/test/scala/akka/http/impl/engine/client/HttpConfigurationSpec.scala
```diff
@@ -137,10 +137,10 @@ class HttpConfigurationSpec extends AkkaSpec {
...
         """.stripMargin) { sys ⇒
 
-          val pool = ConnectionPoolSettings(sys)
-          pool.getMinConnections should ===(42)
-          pool.getMaxConnections should ===(43)
-        }
+        val pool = ConnectionPoolSettings(sys)
+        pool.getMinConnections should ===(42)
+        pool.getMaxConnections should ===(43)
+      }
```
jonas: This (and changes in the files following it) reverts the formatting fixes introduces as part of the sbt-scalariform update.

akka-http-caching/src/main/resources/reference.conf
ben-manes: Note that Caffeine is not an LFU, but rather frequency-biased (rather than recency-biased). In the JavaDoc we try to avoid being explicit as we want the flexibility to make changes later.

> the cache evicts entries that are less likely to be used again. For example, the cache may evict an entry because it hasn't been used recently or very often.

akka-http-caching/src/main/java/akka/http/caching/javadsl/Cache.java
```diff
@@ -0,0 +1,67 @@
...
+  /**
+   * Returns either the cached CompletionStage for the given key or the given value as a CompletionStage
+   */
+  CompletionStage<V> getOrCreateStrict(K key, Creator<V> block);
+
+  /**
+   * Retrieves the CompletionStage instance that is currently in the cache for the given key.
+   * Returns None if the key has no corresponding cache entry.
+   */
+  Optional<CompletionStage<V>> getOptional(K key);
```
ben-manes: Caffeine uses `getIfPresent`, where a `get` is computing (as that is the preferred idiom). That is to say the behavior of the Cache is active (get/getIfPresent), whereas a Map is passive (get/computeIfAbsent). Also, `Optional` uses `ifPresent` which implies it leans towards being populated. I think trying to be consistent with the Java Collections Framework is a good API idiom.

akka-http-caching/src/main/resources/reference.conf
ben-manes: Caffeine's `expireAfterWrite` is based on write time (create or update). An `Expiry` can be used to set more advanced policies, such as create only.

akka-http-caching/src/main/java/akka/http/caching/javadsl/Cache.java
```diff
@@ -0,0 +1,67 @@
...
+  void clear();
+
+  /**
+   * Returns the set of keys in the cache, in no particular order
+   * Should return in roughly constant time.
+   * Note that this number might not reflect the exact keys of active, unexpired
+   * cache entries, since expired entries are only evicted upon next access
+   * (or by being thrown out by a capacity constraint).
+   */
+  java.util.Set<K> getKeys();
```
ben-manes: Perhaps `keySet` like `Map`?

Note that some caches, like memcached, do not expose the keys. Generally it isn't good for applications to inspect such state. Guava/Caffeine uses the `asMap` view as an escape hatch, while keeping an opinionated API via Cache.

akka-http-caching/src/main/java/akka/http/caching/javadsl/Cache.java
ktoso: Yeah good point

akka-http-caching/src/main/java/akka/http/caching/javadsl/Cache.java
ktoso: The reason for "get" style here is that it's our style for when a class will end up mixing java and scala dsl, which is the case here:

```
def keys: immutable.Set[K]
override def getKeys: java.util.Set[K] = keys.asJava
```

or we could remove the method indeed

akka-http-caching/src/main/scala/akka/http/caching/CacheJavaMapping.scala
ktoso: I thought it's the same as the http model, in which we don't don't want to guarantee we won't add methods; and adding methods would be binary incompatible for users extending the cache type, thus this restriction

jonas: Yes makes sense. I will update the documentation.

akka-http-caching/src/main/resources/reference.conf
jonas: Thanks for the clarification. Using the LFU abbreviation in the name is misleading.

ktoso:
This needs @jonas's work to be applied from comment https://github.com/akka/akka-http/pull/1503#issuecomment-341571879 and then the small comments addressed and we can finally merge it then.

jonas:
@ktoso Should I rebase the doc and test updates on top of this PR?

ktoso:
That would be great, thanks!

ktoso:
You can PR into my branch

jonas:
Made a PR to revert formatting and doc changes as mentioned: ktoso/akka-http#4

The new `routeCache()` methods in the Scala DSL are currently causing a compile error in the examples (see below). It's really great that the cache can be configured in the config files, however, at the same time it would then be better to move the `CachingSettings` code to `akka-http-core` as we also do with the routing settings. I experimented with this locally and it also ended up catching a few issues such as `withLfuCacheSettings` being defined in `CachingSettingsImpl` as well as missing overrides of `productPrefix` in `CachingSettingsImpl` and `LfuCacheSettingsImpl`.

Moving the settings to core would allow us to (IMO) fix the awkward API of requiring an actor system to construct a route cache. In my experience, routes are often defined in a separate trait and thus seldom have access to such a thing. In addition if we expose the caching settings as part of the `RoutingSettings` we could then provide versions of the `cache`, `alwaysCache`, and `cacheProhibited` directives which does not take a cache argument but instead uses [`extractSettings`](https://doc.akka.io/docs/akka-http/current/scala/http/routing-dsl/directives/basic-directives/extractSettings.html).

```
[error] .../docs/src/test/scala/docs/http/scaladsl/server/directives/CachingDirectivesExamplesSpec.scala:30:13: overloaded method value routeCache with alternatives:
[error]   [K](settings: akka.http.caching.javadsl.CachingSettings)akka.http.caching.scaladsl.Cache[K,akka.http.scaladsl.server.RouteResult] <and>
[error]   [K](implicit s: akka.actor.ActorSystem)akka.http.caching.scaladsl.Cache[K,akka.http.scaladsl.server.RouteResult]
[error]  cannot be applied to ()
[error]       cache(routeCache(), simpleKeyer) {
[error]             ^
[error] .../docs/src/test/scala/docs/http/scaladsl/server/directives/CachingDirectivesExamplesSpec.scala:63:19: overloaded method value routeCache with alternatives:
[error]   [K](settings: akka.http.caching.javadsl.CachingSettings)akka.http.caching.scaladsl.Cache[K,akka.http.scaladsl.server.RouteResult] <and>
[error]   [K](implicit s: akka.actor.ActorSystem)akka.http.caching.scaladsl.Cache[K,akka.http.scaladsl.server.RouteResult]
[error]  cannot be applied to ()
[error]       alwaysCache(routeCache(), simpleKeyer) {
[error]                   ^
```

akka-http-caching/src/main/scala/akka/http/caching/scaladsl/CachingSettings.scala
```diff
@@ -0,0 +1,70 @@
...
+ * Public API but not intended for subclassing
+ */
+@DoNotInherit
+abstract class CachingSettings extends javadsl.CachingSettings { this: CachingSettingsImpl ⇒
+  override def lfuCacheSettings: LfuCacheSettings
+
+}
+/** INTERNAL API */
+@InternalApi
+private[http] final case class CachingSettingsImpl(lfuCacheSettings: LfuCacheSettings) extends CachingSettings {
```
jonas: This should `override def productPrefix = "CachingSettings"`.

akka-http-caching/src/main/scala/akka/http/caching/scaladsl/CachingSettings.scala
```diff
@@ -0,0 +1,70 @@
...
+  override def withLfuCacheSettings(newSettings: javadsl.LfuCacheSettings): CachingSettings = copy(lfuCacheSettings = newSettings.asInstanceOf[LfuCacheSettings])
```
jonas: `with...` methods should not exist in `Impl` classes and this one is already defined in the Scala DSL version.

akka-http-caching/src/main/scala/akka/http/caching/scaladsl/CachingSettings.scala
```diff
@@ -0,0 +1,70 @@
...
+
+  final def getMaxCapacity: Int = maxCapacity
+  final def getInitialCapacity: Int = initialCapacity
+  final def getTimeToLive: Duration = timeToLive
+  final def getTimeToIdle: Duration = timeToIdle
+
+  override def withMaxCapacity(newMaxCapacity: Int): LfuCacheSettings = copy(maxCapacity = newMaxCapacity)
+  override def withInitialCapacity(newInitialCapacity: Int): LfuCacheSettings = copy(initialCapacity = newInitialCapacity)
+  override def withTimeToLive(newTimeToLive: Duration): LfuCacheSettings = copy(timeToLive = newTimeToLive)
+  override def withTimeToIdle(newTimeToIdle: Duration): LfuCacheSettings = copy(timeToIdle = newTimeToIdle)
```
jonas: All the `with...` methods can be implemented directly in `javadsl.LfuCacheSettings`.

akka-http-caching/src/main/scala/akka/http/caching/scaladsl/CachingSettings.scala
```diff
@@ -0,0 +1,70 @@
...
+}
+/** INTERNAL API */
+@InternalApi
+private[http] final case class LfuCacheSettingsImpl(
+  maxCapacity:     Int,
+  initialCapacity: Int,
+  timeToLive:      Duration,
+  timeToIdle:      Duration
+) extends LfuCacheSettings
```
jonas: This should override def productPrefix = "LfuCacheSettings".

akka-http-caching/src/main/scala/akka/http/caching/javadsl/CachingSettings.scala
jonas: For extra paranoid-ness we should also add `require(...)` calls here or in `LfuCacheSettingsImpl` to ensure positive capacity values.

ktoso:
Thanks for the PR @jonas, some final compile problems now then...

```
error] /localhome/jenkinsakka/workspace/pr-validator-akka-http/docs/src/test/scala/docs/http/scaladsl/server/directives/CachingDirectivesExamplesSpec.scala:30:13: overloaded method value routeCache with alternatives:
[error]   [K](settings: akka.http.caching.javadsl.CachingSettings)akka.http.caching.scaladsl.Cache[K,akka.http.scaladsl.server.RouteResult] <and>
[error]   [K](implicit s: akka.actor.ActorSystem)akka.http.caching.scaladsl.Cache[K,akka.http.scaladsl.server.RouteResult]
[error]  cannot be applied to ()
[error]       cache(routeCache(), simpleKeyer) {
[error]             ^
[error] /localhome/jenkinsakka/workspace/pr-validator-akka-http/docs/src/test/scala/docs/http/scaladsl/server/directives/CachingDirectivesExamplesSpec.scala:63:19: overloaded method value routeCache with alternatives:
[error]   [K](settings: akka.http.caching.javadsl.CachingSettings)akka.http.caching.scaladsl.Cache[K,akka.http.scaladsl.server.RouteResult] <and>
[error]   [K](implicit s: akka.actor.ActorSystem)akka.http.caching.scaladsl.Cache[K,akka.http.scaladsl.server.RouteResult]
[error]  cannot be applied to ()
[error]       alwaysCache(routeCache(), simpleKeyer) {
[error]                   ^
```

jonas:
Yes, I also stumbled upon those and am wondering what is the best fix. The problem could be that the directive examples were not updated to provide an implicit actor system.

jonas:
What do you think about moving CachingSettings to akka-http-core and add an extractCachingSettings directive? More context given in the comment above.

ktoso:
This week I have to focus on something else, but perhaps one of the team can push it over the final finish line, we'll see.

ktoso:
I'm not so convienced about moving the settings, esp not not when we want to keep this "apimaychange"

akka-http-caching/src/main/scala/akka/http/caching/scaladsl/CachingSettings.scala
jonas: Addressed in ktoso/akka-http#5

akka-http-caching/src/main/scala/akka/http/caching/scaladsl/CachingSettings.scala
jonas: Addressed in ktoso/akka-http#5

akka-http-caching/src/main/scala/akka/http/scaladsl/server/directives/CachingDirectives.scala
```diff
@@ -0,0 +1,71 @@
...
+        case Some(key) ⇒ cache.apply(key, () ⇒ route(ctx))
+        case None      ⇒ route(ctx)
+      }
+    }
+  }
+
+  /**
+   * Creates an [[LfuCache]] with default settings obtained from the systems' configuration
+   */
+  @ApiMayChange // since perhaps we indeed have to hardcode the defaults in code rather than reference.conf?
```
raboof: I do agree it's a bit awkward to need an implicit `ActorSystem` here...

ktoso:
Whoop whoop! Thanks @jonas for the final fix. 
Please review team, feel free to PR into my branch whatever you see makes sense to address, I'll merge once into my branch I wake up -- let me know if this is LGTM or not too

ktoso:
Many things addressed and now even with docs by @jonas.
This is LGTM from my perspective, we can bikeshed names more but I think this has waited far too long already.

Will wait for europe to wake up but let's merge as soon as we can.

Resolved conflicts.

ben-manes:
A feature that you may want to consider adding someday is `refreshAfterWrite`. This was not provided by Spray, but is useful when using expiration. This way active entries will be asynchronously reloaded so that a periodic delay (expire-compute cycles) are observed, thereby reducing user-facing latencies.

Unfortunately this features requires that the `AsyncCacheLoader` be a real implementation, since the reload logic might differ from a full load (e.g. recheck external timestamps before an expensive computation). Since we want to promote users to memoize through the cache (via a `LoadingCache`) some features are only available in that form (e.g. bulk gets).

akka-http-caching/src/main/scala/akka/http/caching/scaladsl/Cache.scala
```diff
@@ -0,0 +1,97 @@
...
+   * Returns either the cached Future for the key or evaluates the given function which
+   * should lead to eventual completion of the promise.
+   */
+  def apply(key: K, f: Promise[V] ⇒ Unit): Future[V] =
+    apply(key, () ⇒ { val p = Promise[V](); f(p); p.future })
+
+  /**
+   * Returns either the cached Future for the given key or the given value as a Future
+   */
+  def get(key: K, block: () ⇒ V): Future[V] =
```
ben-manes: Note that `block` is not GC friendly, since it requires allocating a capturing lambda for the key. This is a small overhead, but worth being aware of. The alternative, e.g. in `Map#computeIfAbsent`, is to pass a function `k -> v` so that the caller can opt into the lamda being cached by the JVM.

ktoso: People can store the function though -- `val f: () => V = () => "hello"; get(key, f)`

ben-manes: Sure, but what about common usages like,

```
cache.get(userId, () -> repository.getById(userId))
```

The lambda has to be allocated to pass the key through.

ktoso: Adding the `get(K, V)` overload is not a problem, we can do it after merging this.
It's still possible to change APIs, and this PR has been running since March.

I agree and appropriate the feedback a lot, but also would want this *initial* (!) PR be finally merged

ktoso: How about we open tickets for these polishing-up tasks? Pretty sure people will happily pick them up?

ben-manes: Sure, any observations I make should not be considered blockers and just fyi.

ktoso: Agreed, will make them tickets 👍 
Only making sure we don't open the nitpicking for a week jar again without merging like has happened a few times (so team or anyone would not be sure if "to merge or not yet") :-) Please do keep the feedback coming, will ticketify and follow up!

raboof:
I squashed things a bit, let's merge this as #1503 if CI agrees :tada:

ktoso:
Moving this PR into @raboof's "final final rebase" of it: https://github.com/akka/akka-http/pull/1539

PR #1371, opened 2017-08-21
=doc Add server-side overview page

jonas:
Moves the general intro from the "Server-Side Low-Level API" page to a new index page.

If this is agreed to be a good idea, we should make sure to update this PR (or the PRs <del>#1347</del> and #1351 if merged before) to list the websocket and HTTPS support pages under this page.

PR #1504, opened 2017-10-31
Fix server-side doc link

jonas:
When #1371 was merged it broke the directory level symlink of the
server-side docs. Restore the symlink so the java doc group works
correctly.

CC @ktoso


ktoso:
Thanks for spotting the root cause

Issue #1026, opened 2017-04-12
Streams don't fail on WebSocket idle timeout on the server-side


MichaelZinsmaier:
Hi all,

when upgrading from Akka HTTP 10.0.0 to 10.0.5 I noticed that in case of a server side idle timeout the WebSocket server implementation does not fail the server side streams anymore.

Before 10.0.1 the server side stream terminated with
  - downstream finished
 -  upstream failure Timeout Exception

now it seems to be
 -  downstream finished
  - downstream finished

best take a look at the reproducer
https://github.com/MichaelZinsmaier/akka-http-timeout-bug

I think the 10.0.0 behavior is the better solution since server side code might want to react to a idle timeout in a specific way and propagating it with onError makes sense.

Best Michael

I also found this ticket but not sure if these issues are related
https://github.com/akka/akka-http/issues/1012

jrudolph:
Thanks for the report and the reproducer, @MichaelZinsmaier. 

If that worked before, it seems to have been the case accidentally as well:

```scala
  val greeterWebSocketService =
    Flow[Message]
      .via(new MyMonitorFlow("in"))
      .mapConcat {
        case tm: TextMessage.Strict => {
          val text = tm.text
          println(s"request from $text")
          TextMessage.Strict(s"Hello $text") :: Nil
        }
        case bm: BinaryMessage =>
          bm.dataStream.runWith(Sink.ignore)
          Nil
      }
      .via(new MyMonitorFlow("out"))
```

If you have seen a "out downstream finish", it will propagate through your own stream to the "in" side as well. It seems in the 10.0.0 case, this propagation was slower than the failure managed to be processed on the in side. This was a race condition and may still be one. (It's a pity that we cannot transport downstream errors btw)

If you need access to the error message you can try to prevent propagation of the cancellation for a while to get access to the error message. We had a similar problem in our infrastructure that cancellation would RST TCP connections to early. Recently we introduced a [`DelayCancellation` stage](https://github.com/akka/akka-http/blob/master/akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala#L166) that does implement exactly this functionality. Can you try copy this stage to your code and see you would profit from that as well?

MichaelZinsmaier:
Hi jrudolph,

I added a DelayCancellationFlow to the project, but the issue is still there.
Plus I added a new test based on a BidiPipe IdleTimeout.scala.
(remember that trick (-? something you helped me with still incredible useful)

[...]

PR #1507, opened 2017-11-01
[WIP] fail ws flow on idle timeout

johanandren:
Refs #1026

This is the first version I got to work, not sure about the consequences, needs both the custom idle stage that fails upstream rather than downstream _and_ the delay as close to the user flow as possible, not 100% sure but think that is because failing the upstream also leads to the connection failing and that in turn starts to cancel stages up the stream.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/HttpConnectionIdleTimeoutBidi.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/HttpConnectionIdleTimeoutBidi.scala
@@ -5,40 +5,89 @@
 package akka.http.impl.engine
 
 import java.net.InetSocketAddress
-import java.util.concurrent.TimeoutException
+import java.util.concurrent.{ TimeUnit, TimeoutException }
 
 import akka.NotUsed
 import akka.annotation.InternalApi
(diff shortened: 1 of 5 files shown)


johanandren:
Crap. Forgot that we made the blueprint reusable.

johanandren:
Grün, grün, grün sind alle meine Kleider! (actually they are all black, but the Jenkins build is green, woho!)

jrudolph:
That would also mean that the special IdleTimeout stage cannot help in all cases, because emitting a failure is enough to retrigger the behavior on a higher stage.

Another thing that needs to be tested is the behavior of still running message input and output streams.

ktoso:
Hmm yeah very tempting to take a big step back as we discussed on chat and here https://github.com/akka/akka/issues/17997 and think about bidirectional stages and cancellation in general before attacking this one hm

johanandren:
Should I close this PR then, given the "take a step back" and the fix just being duct-tape?

ktoso:
Hmm, this would relate to the ping/pong stuff as well actually. Do you want to revive this Johan or we close it for now?

johanandren:
Yeah I guess we can close it for now.

Issue #1508, opened 2017-11-01
SSE Performance with big events


kennedyoliveira:
Hi!

While consuming events from the [Marathon](https://mesosphere.github.io/marathon/) SSE via Alpakka SSE Connector, i noticed it wasn't consuming some events and instead closing the connection, investigating a bit, i found that the `akka.http.scaladsl.unmarshalling.sse.LineParser` used to feed lines in the `EventStreamParser` takes to long to parse a line who is big, i know usually the sse doesn't send big lines, but Marathon send some events that have 1 ~ 2 mb, and it takes too long to parse the line.

I mentioned the Marathon because is my use case, but any big event will do.

Is there a alternative for consuming big events?

ktoso:
What specifically is "long" and what is desired; how long do you think is achievable with a raw impl, and how much part of that is that the events are being pulled in slowly?

Please provide more details so this can be acted upon accordingly

kennedyoliveira:
By long i mean `1 ~ 2 MB`, the desired is to parse it fast, currently it can't be parsed in 1 minute and the TCP timeout, i don't know how much time is needed with the current implementation.

What do you mean by how long do you think is achievable with a raw impl?

viktorklang:
Kennedy,

do you have a JMH bench we can take a look at?

-- 
Cheers,
√

On Nov 1, 2017 18:45, "Kennedy Oliveira" <notifications@github.com> wrote:

By long i mean 1 ~ 2 MB, the desired is to parse it fast, currently it
can't be parsed in 1 minute and the TCP timeout, i don't know how much time
is needed with the current implementation.

What do you mean by how long do you think is achievable with a raw impl?

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/akka/akka-http/issues/1508#issuecomment-341183955>, or mute
the thread

kennedyoliveira:
@viktorklang i don't have, i will write one and post the link here

kennedyoliveira:
Hey @viktorklang i created a simple benchmark [akka-http-lineparser-bench](https://github.com/kennedyoliveira/akka-http-lineparser-bench) and was able to improve the performance of the parser via PR #1510. Here are some benchmark numbers i got from running the bench on my machine:

```
[info] Benchmark                     (chunkSize)  (lineSize)  Mode  Cnt     Score    Error  Units
[...]

PR #1510, opened 2017-11-02
Improve performance of LineParser for SSE unmarshalling #1508

kennedyoliveira:
Minor change improving the performance of the LineParser for SSE Unmarshaller

### DIFF
--- a/akka-http-tests/src/test/scala/akka/http/scaladsl/unmarshalling/sse/LineParserSpec.scala
+++ b/akka-http-tests/src/test/scala/akka/http/scaladsl/unmarshalling/sse/LineParserSpec.scala
@@ -15,18 +15,21 @@ final class LineParserSpec extends AsyncWordSpec with Matchers with BaseUnmarsha
 
   "A LineParser" should {
 
-    "parse lines terminated with either CR, LF or CRLF" in {
+    "parse lines terminated with either LF or CRLF" in {
       Source
         .single(ByteString("line1\nline2\rline3\r\nline4\nline5\rline6\r\n\n"))
         .via(new LineParser(1048576))
         .runWith(Sink.seq)
         .map(_ shouldBe Vector("line1", "line2", "line3", "line4", "line5", "line6", ""))
+    }
 
+    "ignore lines terminated with CR" in {
+      // if the line ends with a CR (\r) and there is no next character, this line can't be emitted
       Source
         .single(ByteString("line1\r"))
         .via(new LineParser(1048576))
         .runWith(Sink.seq)
-        .map(_ shouldBe Vector("line1"))
+        .map(_ shouldBe Vector.empty)
     }
 
     "ignore a trailing non-terminated line" in {
@@ -37,11 +40,25 @@ final class LineParserSpec extends AsyncWordSpec with Matchers with BaseUnmarsha
         .map(_ shouldBe Vector("line1", "line2", "line3", "line4", "line5", "line6", ""))
     }
 
+    "ignore a trailing non-terminated line between " in {
+      Source(ByteString("line1\nline2\rline3\r\nline4\nline5\rline6\r\n\nincomplete").map(ByteString(_)))
+        .via(new LineParser(1048576))
+        .runWith(Sink.seq)
+        .map(_ shouldBe Vector("line1", "line2", "line3", "line4", "line5", "line6", ""))
+    }
+
     "parse a line splitted into many chunks" in {
       Source(('a'.to('z') :+ '\n').map(ByteString(_)))
         .via(new LineParser(1048576))
         .runWith(Sink.seq)
         .map(_ shouldBe Vector('a'.to('z').mkString))
+    }
+
+    "parse lines with terminations between chunks" in {
+      Source(("after" :: "\n" :: "\n" :: "before" :: "\r" :: "middle" :: "\n" :: Nil).map(ByteString(_)))
+        .via(new LineParser(1048576))
+        .runWith(Sink.seq)
+        .map(_ shouldBe Vector("after", "", "before", "middle"))
     }
   }
 }
--- a/akka-http-tests/src/test/scala/akka/http/scaladsl/unmarshalling/sse/LineParserSpec.scala
+++ b/akka-http-tests/src/test/scala/akka/http/scaladsl/unmarshalling/sse/LineParserSpec.scala
@@ -15,21 +15,12 @@ final class LineParserSpec extends AsyncWordSpec with Matchers with BaseUnmarsha
 
   "A LineParser" should {
 
-    "parse lines terminated with either LF or CRLF" in {
+    "parse lines terminated with either CR, LF or CRLF" in {
       Source
         .single(ByteString("line1\nline2\rline3\r\nline4\nline5\rline6\r\n\n"))
         .via(new LineParser(1048576))
         .runWith(Sink.seq)
         .map(_ shouldBe Vector("line1", "line2", "line3", "line4", "line5", "line6", ""))
-    }
-
-    "ignore lines terminated with CR" in {
-      // if the line ends with a CR (\r) and there is no next character, this line can't be emitted
-      Source
-        .single(ByteString("line1\r"))
-        .via(new LineParser(1048576))
-        .runWith(Sink.seq)
-        .map(_ shouldBe Vector.empty)
     }
 
     "ignore a trailing non-terminated line" in {
@@ -40,7 +31,7 @@ final class LineParserSpec extends AsyncWordSpec with Matchers with BaseUnmarsha
         .map(_ shouldBe Vector("line1", "line2", "line3", "line4", "line5", "line6", ""))
     }
 
-    "ignore a trailing non-terminated line between " in {
+    "ignore a trailing non-terminated line when parsing byte by byte" in {
       Source(ByteString("line1\nline2\rline3\r\nline4\nline5\rline6\r\n\nincomplete").map(ByteString(_)))
         .via(new LineParser(1048576))
         .runWith(Sink.seq)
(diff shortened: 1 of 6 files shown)


ktoso:
OK TO TEST

ktoso:
Refs https://github.com/akka/akka-http/issues/1508#issuecomment-341283142

ktoso:
LGTM, though awaiting addition of benchmark (it exists, just needs to be inserted here)

akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/sse/LineParser.scala
```diff
@@ -58,7 +58,10 @@ private final class LineParser(maxLineSize: Int) extends GraphStage[FlowShape[By
               case _ ⇒
                 parseLines(bs, from, at + 1, parsedLines)
             }
-        buffer = parseLines(buffer ++ grab(in)) match {
+
+        // start the search where it ended, prevent iterating over all the buffer again
+        val currentBufferStart = if (buffer.isEmpty) 0 else buffer.length - 1
```
viktorklang: My suggestion would be `val currentBufferStart = math.max(0, buffer.length - 1)`

akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/sse/LineParser.scala
```diff
@@ -58,7 +58,10 @@ private final class LineParser(maxLineSize: Int) extends GraphStage[FlowShape[By
...
+        buffer = parseLines(buffer ++ grab(in), at = currentBufferStart) match {
```
viktorklang: :+1:

akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/sse/LineParser.scala
jrudolph: Do we have tests for all combinations of having the chunk boundary directly before and after `CR` and `LF` and when it happens to be between `CR` and `LF`?

kennedyoliveira: Updated with your suggestion!

kennedyoliveira: Could you elaborate a little more about it?

jrudolph: All of this logic is about matching CR LF pairs, so this change has the risk that we might now miss a pair or match it wrongly (e.g. adding an extra CR to the first element). We might already have those tests, if not we need to add them.

jrudolph:
This slightly adapted test fails before and after the change because extra empty lines are emitted if the chunk size is 1 byte (i.e. the chunk boundaries happens between `CR` and `LF`):

```scala
    "ignore a trailing non-terminated line" in {
      Source(ByteString("line1\nline2\rline3\r\nline4\nline5\rline6\r\n\nincomplete").map(ByteString(_)))
        .via(new LineParser(1048576))
        .runWith(Sink.seq)
        .map(_ shouldBe Vector("line1", "line2", "line3", "line4", "line5", "line6", ""))
    }
```

The [SSE spec](https://www.w3.org/TR/eventsource/#parsing-an-event-stream) says

> The stream must then be parsed by reading everything line by line, with a U+000D CARRIAGE RETURN U+000A LINE FEED (CRLF) character pair, a single U+000A LINE FEED (LF) character not preceded by a U+000D CARRIAGE RETURN (CR) character, and a single U+000D CARRIAGE RETURN (CR) character not followed by a U+000A LINE FEED (LF) character being the ways in which a line can end.

That means that after reading a `CR` you need to keep track that you've read a `CR` before and ignore the next character if it is a `LF`. In our streaming parser you need to keep track of this across chunk boundaries.

kennedyoliveira:
@jrudolph i understood, i'll work on that

kennedyoliveira:
@jrudolph i updated the logic according to the specification, could you review it please?

kennedyoliveira:
Can someone point the problem?

viktorklang:
@kennedyoliveira It would seem as you haven't run the build locally before committing (reformatting):

-              case `cr` if at >= bs.length - 1 =>
+              case `cr` if at >= bs.length - 1 ⇒

kennedyoliveira:
@viktorklang thanks! I really forgot about it, just fixed!

kennedyoliveira:
@jrudolph could you review please?

akka-http-tests/src/test/scala/akka/http/scaladsl/unmarshalling/sse/LineParserSpec.scala
```diff
@@ -15,18 +15,21 @@ final class LineParserSpec extends AsyncWordSpec with Matchers with BaseUnmarsha
...
-    "parse lines terminated with either CR, LF or CRLF" in {
+    "parse lines terminated with either LF or CRLF" in {
       Source
         .single(ByteString("line1\nline2\rline3\r\nline4\nline5\rline6\r\n\n"))
         .via(new LineParser(1048576))
         .runWith(Sink.seq)
         .map(_ shouldBe Vector("line1", "line2", "line3", "line4", "line5", "line6", ""))
+    }
 
+    "ignore lines terminated with CR" in {
```
raboof: Hmm, is this something we're willing to accept?

akka-http-tests/src/test/scala/akka/http/scaladsl/unmarshalling/sse/LineParserSpec.scala
```diff
@@ -37,11 +40,25 @@ final class LineParserSpec extends AsyncWordSpec with Matchers with BaseUnmarsha
         .map(_ shouldBe Vector("line1", "line2", "line3", "line4", "line5", "line6", ""))
     }
 
+    "ignore a trailing non-terminated line between " in {
```
raboof: Between what? It's good that this test makes this behavior explicit but I'm not sure it is obvious to me that this is the behavior we want..

akka-http-tests/src/test/scala/akka/http/scaladsl/unmarshalling/sse/LineParserSpec.scala
jrudolph: Looks like a bug to me.

akka-http-tests/src/test/scala/akka/http/scaladsl/unmarshalling/sse/LineParserSpec.scala
```diff
@@ -15,18 +15,21 @@ final class LineParserSpec extends AsyncWordSpec with Matchers with BaseUnmarsha
...
+    "parse lines terminated with either LF or CRLF" in {
       Source
         .single(ByteString("line1\nline2\rline3\r\nline4\nline5\rline6\r\n\n"))
         .via(new LineParser(1048576))
         .runWith(Sink.seq)
         .map(_ shouldBe Vector("line1", "line2", "line3", "line4", "line5", "line6", ""))
+    }
 
+    "ignore lines terminated with CR" in {
+      // if the line ends with a CR (\r) and there is no next character, this line can't be emitted
```
jrudolph: It can ;)

See my previous comment:

> That means that after reading a CR you need to keep track that you've read a CR before and ignore the next character if it is a LF. In our streaming parser you need to keep track of this across chunk boundaries.

akka-http-tests/src/test/scala/akka/http/scaladsl/unmarshalling/sse/LineParserSpec.scala
```diff
@@ -15,18 +15,21 @@ final class LineParserSpec extends AsyncWordSpec with Matchers with BaseUnmarsha
 
   "A LineParser" should {
 
-    "parse lines terminated with either CR, LF or CRLF" in {
+    "parse lines terminated with either LF or CRLF" in {
```
jrudolph: It still works for `"line2"` so it still works for `CR`.

akka-http-tests/src/test/scala/akka/http/scaladsl/unmarshalling/sse/LineParserSpec.scala
jrudolph: I think I forgot to finish that sentence in the test case I posted before. Let's call it `"ignore a trailing non-terminated line when parsing byte by byte"`

akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/sse/LineParser.scala
```diff
@@ -51,14 +51,20 @@ private final class LineParser(maxLineSize: Int) extends GraphStage[FlowShape[By
               // Lookahead for LF after CR
               case `cr` if at < bs.length - 1 && bs(at + 1) == lf ⇒
                 parseLines(bs, at + 2, at + 2, parsedLines :+ bs.slice(from, at).utf8String)
+              // if starts with a CR, but we don't know the next byte, we can't slice it yet, so simply advance
```
jrudolph: We can ;) But we also need to return a flag to remember that the last character read was a `CR`.

kennedyoliveira:
@jrudolph i fixed the behavior, but the code somehow doesn't look good to me, do you see any more improvement? Thank you

akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/sse/LineParser.scala
```diff
@@ -34,6 +34,7 @@ private final class LineParser(maxLineSize: Int) extends GraphStage[FlowShape[By
       import shape._
 
       private var buffer = ByteString.empty
+      private var slicedCr = false
```
jrudolph: Maybe we should call it `lastCharWasCr` everywhere then it's pretty self-explanatory.

kennedyoliveira: Agree, this sounds much better. I'll change it.

kennedyoliveira: Done!

PR #1516, opened 2017-11-06
Fix race condition in WebSocket switch

### LINKED ISSUE #1515: Probable Akka HTTP bug in server side WebSockets handling
Lagom is seeing a lot of these errors in its integration tests when using WebSockets:

```
2017-11-06 12:56:54,422 DEBUG akka.actor.RepointableActorRef - Aborting tcp connection to /127.0.0.1:42630 because of upstream failure: requirement failed: Cannot push port (SubSource.out(326501899)) twice
scala.Predef$.require(Predef.scala:224)
akka.stream.stage.GraphStageLogic.push(GraphStage.scala:548)
akka.stream.impl.fusing.SubSource$$anon$4$$anonfun$1.apply(StreamOfStreams.scala:724)
akka.stream.impl.fusing.SubSource$$anon$4$$anonfun$1.apply(StreamOfStreams.scala:721)
akka.stream.impl.fusing.GraphInterpreter.runAsyncInput(GraphInterpreter.scala:443)
akka.stream.impl.fusing.GraphInterpreterShell$AsyncInput.execute(ActorGraphInterpreter.scala:453)
akka.stream.impl.fusing.GraphInterpreterShell.processEvent(ActorGraphInterpreter.scala:546)
akka.stream.impl.fusing.ActorGraphInterpreter.akka$stream$impl$fusing$ActorGraphInterpreter$$processEvent(ActorGraphInterpreter.scala:728)
akka.stream.impl.fusing.ActorGraphInterpreter.akka$stream$impl$fusing$ActorGraphInterpreter$$shortCircuitBatch(ActorGraphInterpreter.scala:718)
akka.stream.impl.fusing.ActorGraphInterpreter$$anonfun$receive$1.applyOrElse(ActorGraphInterpreter.scala:744)
akka.actor.Actor$class.aroundReceive(Actor.scala:517)
akka.stream.impl.fusing.ActorGraphInterpreter.aroundReceive(ActorGraphInterpreter.scala:653)
akka.actor.ActorCell.receiveMessage(ActorCell.scala:527)
akka.actor.ActorCell.invoke(ActorCell.scala:496)
akka.dispatch.Mailbox.processMailbox(Mailbox.scala:257)
akka.dispatch.Mailbox.run(Mailbox.scala:224)
akka.dispatch.Mailbox.exec(Mailbox.scala:234)
akka.dispatch.forkjoin.ForkJoinTask.doExec(ForkJoinTask.java:260)
akka.dispatch.forkjoin.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1339)
akka.dispatch.forkjoin.ForkJoinPool.runWorker(ForkJoinPool.java:1979)
akka.dispatch.forkjoin.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:107)
```

The problem with this stack trace is that we don't know what invoked `SubSourceOutlet.push` due to the asynchronous boundary between this method and the `SubSource` handling. However, neither Play nor Lagom use any `SubSourceOutlet`'s (it's supposed to be an internal Akka streams API anyway), so the most likely cause of this is something wrong in Akka HTTP's WebSocket support which does create a `SubSourceOutlet`. I can't think of any other way this error could end up happening.

Also note that because `SubSource` just creates a generic name `SubSource.out` for its outlet, there's no way to map this error back to the original source. I'll raise a bug about that.

More details about the bug are reported in https://github.com/lagom/lagom/pull/1054.

These tests are doing a lot of starting up WebSockets and then terminating them immediately, so it could be a race condition on startup.

jroper:
Fixes #1515

Ensures that we only push to the WebSocket handling flow if it has first pulled.

I've tested this against the reproducer for this that we had in Lagom (https://github.com/lagom/lagom/pull/1054), and I haven't been able to reproduce after several test runs, whereas before I was reproducing on every test run.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
@@ -664,14 +664,25 @@ private[http] object HttpServerBluePrint {
           })
 
           setHandler(fromNet, new InHandler {
-            override def onPush(): Unit = sourceOut.push(grab(fromNet).bytes)
+            override def onPush(): Unit = {
+              if (sourceOut.isAvailable) {
+                sourceOut.push(grab(fromNet).bytes)
+              }
+            }
             override def onUpstreamFinish(): Unit = sourceOut.complete()
             override def onUpstreamFailure(ex: Throwable): Unit = sourceOut.fail(ex)
           })
           sourceOut.setHandler(new OutHandler {
             override def onPull(): Unit = {
-              if (!hasBeenPulled(fromNet)) pull(fromNet)
+              // This check only needed on the first pull due to potential element
+              // pushed in response to pull by previous source
+              if (isAvailable(fromNet)) {
+                sourceOut.push(grab(fromNet).bytes)
+              } else if (!hasBeenPulled(fromNet)) {
+                pull(fromNet)
+              }
               cancelTimeout(timeoutKey)
+
               sourceOut.setHandler(new OutHandler {
                 override def onPull(): Unit = if (!hasBeenPulled(fromNet)) pull(fromNet)
                 override def onDownstreamFinish(): Unit = cancel(fromNet)


jroper:
Not sure if there's a straight forward way to write a unit test for this, would need to be able to delay a pull from a flow, but Akka streams tends to be very eager about pulling, with buffers around the place.

jrudolph:
Thanks for the fix, @jroper. How does it relate to the issue that @ignasi35 described which rather seems to be a client issue?

jroper:
It wasn't a client issue at all, the client was receiving a connection reset by peer exception, ie, the server was reseting the connection.

jrudolph:
Interesting, the tcpdumps clearly also showed a client issue but maybe in the end it wasn't the problem that turned up in the tests.

jroper:
What was the client issue?

jrudolph:
What @ignasi35 showed in the slides:

screenshot from 2017-11-06 12 20 10

In the case shown the server was "lucky" that it had already sent out its FIN itself. With slightly different timing it could have seen a "Connection reset by peer" itself. As you found out it isn't relevant for the tests (but might show up in server logs).

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -664,14 +664,25 @@ private[http] object HttpServerBluePrint {
...
+              }
+            }
             override def onUpstreamFinish(): Unit = sourceOut.complete()
             override def onUpstreamFailure(ex: Throwable): Unit = sourceOut.fail(ex)
           })
           sourceOut.setHandler(new OutHandler {
             override def onPull(): Unit = {
-              if (!hasBeenPulled(fromNet)) pull(fromNet)
+              // This check only needed on the first pull due to potential element
+              // pushed in response to pull by previous source
```
jrudolph: Yep, exactly that's the reason. Good catch. Execution of subsource handlers is less deterministic than things running in a single interpreter because the exact interleaving of event processing depends on when exactly the event horizon of the participating interpreters are depleted.

jrudolph:
I created https://github.com/akka/akka-http/issues/1518 to track the client side issue.

jrudolph:
In https://github.com/akka/akka-http/blob/05cf0008a3889357d9816505960f9608f5b80dfd/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/AlpnSwitch.scala#L89 we have tried more general infrastructure to connect inlets/outlets. We should strive for making this general enough to include it in streams, so that these kinds of errors can be prevented in the first place. I created https://github.com/akka/akka/issues/23930 to track this.

Issue #1359, opened 2017-08-16
Decode failed using akka.http.scaladsl.coding.Deflate


DSRYhh:
When I try to use akka.http.scaladsl.coding.Deflate to decode HTTP response, there was an exception: `akka.stream.impl.io.ByteStringParser$ParsingException: Parsing failed in step akka.http.scaladsl.coding.DeflateDecompressor$$anon$1$$anon$2@1bb42e8
`. I believe I receive correct bytes, so I try to use `java.util.zip.InflaterInputStream` to decode the HTTP entity with deflate encode, and it works.

This is Akka version (will cause an exception):
```
     
      responseFuture
            .map(Deflate.decodeMessage(_))
            .map(_.entity)
            .flatMap(_.toStrict(1 seconds)(materializer))
            .map(_.data)
            .map(_.utf8String)//this will throw a parse error exception
```

and this is Java version:
```
responseFuture
            .map(_.entity)
            .flatMap(_.toStrict(1 seconds)(materializer))
            .map(_.dataBytes)
            .map(_.runWith(StreamConverters.asInputStream()))
            .map(new InflaterInputStream(_, new Inflater(true)))
            .map(Source.fromInputStream(_).mkString)//this should be fine
```

The entire source code can be seen in [Github gist](https://gist.github.com/DSRYhh/d4a1a8cd828508899b6dcc54693608c7)

jrudolph:
According to https://tools.ietf.org/html/rfc7230#section-4.2.2:

```
   The "deflate" coding is a "zlib" data format [RFC1950] containing a
   "deflate" compressed data stream [RFC1951] that uses a combination of
   the Lempel-Ziv (LZ77) compression algorithm and Huffman coding.

      Note: Some non-conformant implementations send the "deflate"
      compressed data without the zlib wrapper.
```

It seems the server you run against is missing the wrapping, so you need to use `Inflater(nowrap = true)` while akka-http's Deflate implementation uses `Inflater(nowrap = false)` according to the spec.

There might be a way to guess if the zlib wrapper exists or not, if that's the case, we would be open to a PR that adds auto detection.

DSRYhh:
@jrudolph can `Deflate.decodeMessage()` add a optional parameter like `decodeMessage(nowarp : Boolean = false)` to make it work in this case?

raboof:
That sounds neat to me (it'd probably have to be a second method for binary compatibility) - would you be up to creating a PR?

jrudolph:
[...]

PR #1522, opened 2017-11-08
Add nowrap to deflator

tom-walford:
Added support in Deflate for decoding a message overriding the noWrap setting on the Inflator

Not sure what needed to be done to maintain binary compatibility - I added an overloaded method for decodeMessage but added a default parameter to the DeflateDecompressor - hope that's ok?

### DIFF
--- a/akka-http-tests/src/test/scala/akka/http/scaladsl/coding/DeflateSpec.scala
+++ b/akka-http-tests/src/test/scala/akka/http/scaladsl/coding/DeflateSpec.scala
@@ -5,9 +5,18 @@
 package akka.http.scaladsl.coding
 
 import akka.util.ByteString
-
 import java.io.{ InputStream, OutputStream }
 import java.util.zip._
+
+import akka.http.scaladsl.model.HttpMethods.POST
+import akka.http.scaladsl.model.{ HttpEntity, HttpRequest }
+import akka.http.impl.util._
+import akka.http.scaladsl.model.headers.{ HttpEncodings, `Content-Encoding` }
+import akka.stream.scaladsl.Flow
+import akka.testkit._
+
+import scala.concurrent.duration._
+import scala.concurrent.ExecutionContext.Implicits.global
 
 class DeflateSpec extends CoderSpec {
   protected def Coder: Coder with StreamDecoder = Deflate
@@ -24,5 +33,20 @@ class DeflateSpec extends CoderSpec {
         ourDecode(ByteString(0, 1, 2, 3, 4))
       }).ultimateCause should be(a[DataFormatException])
     }
+    "properly round-trip encode/decode an HttpRequest using wrapping" in {
+      val request = HttpRequest(POST, entity = HttpEntity(largeText))
+      Deflate.decodeMessage(encodeNoWrappedMessage(request), noWrap = true).toStrict(3.seconds.dilated)
+        .awaitResult(3.seconds.dilated) should equal(request)
+    }
+  }
+
+  private def encodeNoWrappedMessage(request: HttpRequest): HttpRequest = {
+    val deflaterWithoutWrapping = new Deflate(Encoder.DefaultFilter) {
+      override def newCompressor = new DeflateCompressor {
+        override lazy val deflater = new Deflater(Deflater.BEST_COMPRESSION, true)
+      }
+    }
+    request.transformEntityDataBytes(deflaterWithoutWrapping.encoderFlow)
+      .withHeaders(`Content-Encoding`(HttpEncodings.deflate) +: request.headers)
   }
 }
--- a/akka-http-tests/src/test/scala/akka/http/scaladsl/coding/DeflateSpec.scala
+++ b/akka-http-tests/src/test/scala/akka/http/scaladsl/coding/DeflateSpec.scala
@@ -35,7 +35,7 @@ class DeflateSpec extends CoderSpec {
     }
     "properly round-trip encode/decode an HttpRequest using wrapping" in {
       val request = HttpRequest(POST, entity = HttpEntity(largeText))
-      Deflate.decodeMessage(encodeNoWrappedMessage(request), noWrap = true).toStrict(3.seconds.dilated)
+      Deflate.decodeMessage(encodeNoWrappedMessage(request)).toStrict(3.seconds.dilated)
         .awaitResult(3.seconds.dilated) should equal(request)
     }
   }
--- a/akka-http-tests/src/test/scala/akka/http/scaladsl/coding/DeflateSpec.scala
+++ b/akka-http-tests/src/test/scala/akka/http/scaladsl/coding/DeflateSpec.scala
@@ -33,17 +33,27 @@ class DeflateSpec extends CoderSpec {
         ourDecode(ByteString(0, 1, 2, 3, 4))
       }).ultimateCause should be(a[DataFormatException])
     }
-    "properly round-trip encode/decode an HttpRequest using wrapping" in {
+    "properly round-trip encode/decode an HttpRequest using no-wrap and best compression" in {
       val request = HttpRequest(POST, entity = HttpEntity(largeText))
-      Deflate.decodeMessage(encodeNoWrappedMessage(request)).toStrict(3.seconds.dilated)
+      Deflate.decodeMessage(encodeMessage(request, Deflater.BEST_COMPRESSION, noWrap = true)).toStrict(3.seconds.dilated)
+        .awaitResult(3.seconds.dilated) should equal(request)
+    }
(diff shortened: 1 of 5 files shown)


ktoso:
OK TO TEST

jrudolph:
Thanks for the contribution, @tom-walford.

The specification is pretty clear that deflate compressed streams need the wrapping. So, I'm wary of adding extra API for non-conforming behavior. On the other hand, it's also good practice to be liberal in what you expect. In the optimal case, the decoder could automatically detect that wrapping and remove it without needing extra API. Could you try if that's possible?

Thanks also the scaladoc changes. Could you split them off into an extra PR that would simplify reviewing and merging.

jrudolph:
Here are some pointers on the wrapping header: https://stackoverflow.com/questions/9050260/what-does-a-zlib-header-look-like. The idea would be to change `DecompressorParsingLogic` in a way that peeks if the first bytes contain the wrapping or not and only then create the `Inflator` with the correct flag.

tom-walford:
Urk, sorry, the docs shouldn't be in this PR!

tom-walford:
@jrudolph Am happy to rework to peek at the head bytes :)

tom-walford:
I've switched it to examining the first 4 bits to check the deflate header - is it possible that this check could pass if we're using no-wrap suitable data and we accidentally use the wrong inflater? I couldn't get much of a clear answer from the libraries themselves

jrudolph:
> I've switched it to examining the first 4 bits to check the deflate header - is it possible that this check could pass if we're using no-wrap suitable data and we accidentally use the wrong inflater?

Hmm, I wonder if that could be case. A deflate block has a 3-*bit* header ([RFC 1951](https://www.ietf.org/rfc/rfc1951.txt)):

```
      3.2.3. Details of block format

         Each block of compressed data begins with 3 header bits
         containing the following data:

            first bit       BFINAL
            next 2 bits     BTYPE

         Note that the header bits do not necessarily begin on a byte
         boundary, since a block does not necessarily occupy an integral
         number of bytes.

         BFINAL is set if and only if this is the last block of the data
         set.

         BTYPE specifies how the data are compressed, as follows:

            00 - no compression
            01 - compressed with fixed Huffman codes
            10 - compressed with dynamic Huffman codes
            11 - reserved (error)
```

If I'm not mistaken the first zlib header byte which is usually 0x78 would be 01111000 which would mean 000 for the 3-bit deflate block header which would be valid as well as a block with no compression. For that it says that:

```
3.2.4. Non-compressed blocks (BTYPE=00)

         Any bits of input up to the next byte boundary are ignored.
         The rest of the block consists of the following information:
```

So, I'd say it's pretty unlikely to see a deflate block starting with 0x78 because a block without compression would probably rather start with 0x00 in the first place. Would be nice to validate that though, as I might have flipped the bits or misread the specs.

tom-walford:
So i ran a few hundred thousand random strings though it (because i'm somewhat paranoid :) ) and they're definitely distinct - notably the CM header always seems to have the first bit set to 1.

If the CM header isn't 8 (the first four bits according to the SO post you linked) then the compression isn't defined, which would match my limited test

tom-walford:
Looking through the spec you're right, but there that risk that if you do have an uncompressed byte string that isn't the final block, then it will attempt to deflate it using nowrap=false.
Granted the code currently checks for 0x08 not 0x78, so I can change that to check for that header if preferred

akka-http/src/main/scala/akka/http/scaladsl/coding/Deflate.scala
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2009-2017 Lightbend Inc. <http://www.lightbend.com>
+ * Copyright (C) 2009-2016 Lightbend Inc. <http://www.lightbend.com>
```
jrudolph: Oops ;)

akka-http/src/main/scala/akka/http/scaladsl/coding/Deflate.scala
```diff
@@ -89,10 +89,19 @@ private[http] object DeflateCompressor {
 class DeflateDecompressor(maxBytesPerChunk: Int = Decoder.MaxBytesPerChunkDefault) extends DeflateDecompressorBase(maxBytesPerChunk) {
 
   override def createLogic(attr: Attributes) = new DecompressorParsingLogic {
-    override val inflater: Inflater = new Inflater()
+    var maybeNoWrappedInflater: Option[Inflater] = None
+    override def inflater: Inflater = maybeNoWrappedInflater.getOrElse(new Inflater())
```
jrudolph: Hmm, that construction is a bit weird. Maybe just like that:

```scala
var _inflater: Inflater = _
def inflater = if (_inflater ne null) _inflater else throw new IllegalStateException("Inflater used before probing for wrapping")
```

akka-http/src/main/scala/akka/http/scaladsl/coding/Deflate.scala
```diff
@@ -89,10 +89,19 @@ private[http] object DeflateCompressor {
...
 
     override val inflateState = new Inflate(true) {
```
jrudolph: Let's create a new state here while probing for the wrapping.

akka-http/src/main/scala/akka/http/scaladsl/coding/Deflate.scala
```diff
@@ -89,10 +89,19 @@ private[http] object DeflateCompressor {
...
       override def onTruncation(): Unit = completeStage()
+
+      override def parse(reader: ByteStringParser.ByteReader): ParseResult[ByteString] = {
+        val data = reader.remainingData
+        maybeNoWrappedInflater = maybeNoWrappedInflater
```
jrudolph: Here `_inflater` will be initialized for the first time.

akka-http/src/main/scala/akka/http/scaladsl/coding/Deflate.scala
```diff
@@ -89,10 +89,19 @@ private[http] object DeflateCompressor {
...
+          .orElse(data.headOption.filter(b ⇒ (b & 0x0F) != 8)
```
jrudolph: We should probably add a `peekByte` method to `ByteReader`. This then needs to be documented with all the information we researched here in this thread.

akka-http/src/main/scala/akka/http/scaladsl/coding/Deflate.scala
```diff
@@ -89,10 +89,19 @@ private[http] object DeflateCompressor {
...
+            .fold(Some(new Inflater()))(_ ⇒ Some(new Inflater(true))))
+        super.parse(new ByteStringParser.ByteReader(reader.takeAll()))
+      }
     }
 
     override def afterInflate = inflateState
```
jrudolph: Needs to be set to the new initial state.

akka-http-tests/src/test/scala/akka/http/scaladsl/coding/DeflateSpec.scala
```diff
@@ -24,5 +33,20 @@ class DeflateSpec extends CoderSpec {
         ourDecode(ByteString(0, 1, 2, 3, 4))
       }).ultimateCause should be(a[DataFormatException])
     }
+    "properly round-trip encode/decode an HttpRequest using wrapping" in {
+      val request = HttpRequest(POST, entity = HttpEntity(largeText))
```
jrudolph: Would be nice to add a few tests for the edge cases like the no compression deflate compression without wrapping case discussed in the PR.

akka-http/src/main/scala/akka/http/scaladsl/coding/Deflate.scala
tom-walford: Yeah wasn't sure if you're were ok with nulls - will change it round :)

akka-http/src/main/scala/akka/http/scaladsl/coding/Deflate.scala
tom-walford: yeah... oops :)

akka-http/src/main/scala/akka/http/scaladsl/coding/Deflate.scala
tom-walford: Possibly - ByteReader is from Akka Streams though - does that not need a seperate issue/PR?

jrudolph:
PLS BUILD

raboof:
Failure is #1560 - perhaps look at #1565

jrudolph:
PLS BUILD

jrudolph:
Refs #1359.

PR #1524, opened 2017-11-09
=doc #1505 be less drastic about whether akka http still uses actors

### LINKED ISSUE #1505: [Docs] Contradiction on Akka Streams and Akka Actors use on Akka HTTP
Hi dev,

Seems to be these two mentions are contradicting to each other and misguide the users.

1. [migration-from-spray.md#streams-everywhere](https://github.com/akka/akka-http/blob/v10.0.10/docs/src/main/paradox/scala/http/migration-guide/migration-from-spray.md#streams-everywhere)
> Akka HTTP is not implemented on Actors any more but is now completely based on akka-stream.

2. [AKKA HTTP introduction](https://github.com/akka/akka-http/blob/v10.0.10/docs/src/main/paradox/scala/http/introduction.md#introduction)
> The Akka HTTP modules implement a full server- and client-side HTTP stack on top of akka-actor and akka-stream. 

I'm new to AKKA as well as for Scala. First time I read 1st reference above, what I felt was, I can't use AKKA HTTP (which is currently based on Streams) with Actors. After I investigate, I found the 2nd reference and got what it exactly meant to be.

I think, it'll be better to update/elaborate more the 1st reference to avoid any confusion.

Thank you!

jrudolph:
Fixes #1505.

### DIFF
--- a/docs/src/main/paradox/scala/http/migration-guide/migration-from-spray.md
+++ b/docs/src/main/paradox/scala/http/migration-guide/migration-from-spray.md
@@ -11,8 +11,7 @@ little effort.
 
 ### Streams everywhere
 
-Akka HTTP is not implemented on Actors any more but is now completely based on akka-stream. This has some important
-consequences.
+Akka HTTP offers an API based on streams where spray offered an API based on actor messaging. This has important consequences.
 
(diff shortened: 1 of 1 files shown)


ktoso:
Failure unrelated

Issue #1312, opened 2017-07-27
Rewrite host connection pool as single GraphStage


jrudolph:
As far as I can see that should remove lots of the stream related bugs. The custom GraphStage should contain all logic from `PoolInterfaceActor`, `PoolConductor`, `PoolSlot`. The external interface should probably be a `FlowShape[HttpRequest, HttpResponse]`. The `PoolMasterActor` can then use that instead of sending messages to the current `PoolInterfaceActor`.

I guess in the best case, we would make this new implementation available under a feature flag depending on our confidence on getting it right.

tomrf1:
This sounds like an interesting one.
The `PoolInterfaceActor` currently has a request buffer with size `max-open-requests`. Is the intention to continue using the buffer, inside the new GraphStage? (Perhaps as a [rate decoupled graph stage](https://doc.akka.io/docs/akka/2.5.6/scala/stream/stream-customize.html#rate-decoupled-graph-stages)).
I'm not sure what the reason for the buffering here is. Why not just backpressure all the way and let the user decide whether or not to buffer?

jrudolph:
@tomrf1, good point. The reason for the buffer is that requests can always be pushed towards a pool by running them via `Http.singleRequest`. I think previously there was also another concern that new stream materializations created by `Http.cachedHostConnectionPool` or `Http.superPool` also need some buffer space because a pool cannot know from which of those materializations to request from, so it requests from all of them.

I think we might want to separate the effort into a part of just rewriting `PoolConductor`, `PoolFlow`, and `PoolSlot` leaving `PoolInterfaceActor` as it is.

Rewriting `PoolInterfaceActor` into something else could be another step in the effort.

jrudolph:
The new pool implementation will be available in 10.0.11 under a flag. I'm closing this one in favor of follow up tickets to flesh out details of behavior in the new implementation when it got some more testing.

benjamingeer:
Is the new pool implementation still available only under a flag in 10.1.1, and if so, what's the flag? And does it resolve #1459?

raboof:
The new host connection pool is the default since [10.1.0](https://doc.akka.io/docs/akka-http/current/release-notes/10.1.x.html#10-1-0). I'm not familiar enough with #1459 to say whether it is expected to resolve #1459.

PR #1533, opened 2017-11-16
#1312 Rewrite of PoolFlow / PoolSlot / PoolConductor as single GraphStage

jrudolph:
At this point all old tests are passing (though the new pool is not yet enabled yet. The new pool has the same interface as the old one. It can be swapped in by replacing

```scala
val poolFlow = PoolFlow(connectionFlow, settings, log).named("PoolFlow")
```

with 

```scala
val poolFlow = HostConnectionPool(connectionFlow, settings, log).named("PoolFlow")
```

in PoolInterfaceActor, here:

https://github.com/akka/akka-http/blob/50a9d327b96302d38aca4178f22e9bcd75931def/akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolInterfaceActor.scala#L98-L102

The idea is to make it configurable for a bit of time to let it mature.

I haven't implemented the new and interesting features like better timeout, reporting, etc yet.

I don't consider this PR ready yet. I'd like to fill in a few more new tests and clean up the implementation. If you are interested you may want to have a look but it's probably too early to review.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolFlow.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolFlow.scala
@@ -5,6 +5,7 @@
 package akka.http.impl.engine.client
 
 import akka.NotUsed
+import akka.annotation.InternalApi
 import akka.http.impl.engine.client.PoolConductor.PoolSlotsSetting
 import akka.http.scaladsl.settings.ConnectionPoolSettings
 
@@ -16,7 +17,9 @@ import akka.stream.scaladsl._
 import akka.http.scaladsl.model._
 import akka.http.scaladsl.Http
 
-private object PoolFlow {
+/** Internal API */
+@InternalApi
+private[client] object PoolFlow {
 
   case class RequestContext(request: HttpRequest, responsePromise: Promise[HttpResponse], retriesLeft: Int) {
     require(retriesLeft >= 0)
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/SlotState.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/SlotState.scala
@@ -1,0 +1,221 @@
+/*
+ * Copyright (C) 2009-2017 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.http.impl.engine.client.pool
+
+import akka.annotation.InternalApi
+import akka.http.impl.engine.client.PoolFlow.RequestContext
+import akka.http.impl.util._
+import akka.http.scaladsl.Http
+import akka.http.scaladsl.model.{ HttpRequest, HttpResponse }
+import akka.http.scaladsl.settings.ConnectionPoolSettings
+
+import scala.concurrent.Future
+import scala.concurrent.duration._
+
+/**
+ * Internal API
+ *
+ * Interface between slot states and the actual slot.
+ */
+@InternalApi
+private[pool] abstract class SlotContext {
+  def openConnection(): Future[Http.OutgoingConnection]
+  def pushRequestToConnectionAndThen(request: HttpRequest, nextState: SlotState): SlotState
+  def closeConnection(): Unit
+  def isConnectionClosed: Boolean
+
+  def dispatchFailure(req: RequestContext, cause: Throwable): Unit
+  def dispatchResponse(req: RequestContext, res: HttpResponse): Unit
+
+  def willCloseAfter(res: HttpResponse): Boolean
+
+  def debug(msg: String): Unit
+  def debug(msg: String, arg1: AnyRef): Unit
+
+  def warning(msg: String): Unit
+  def warning(msg: String, arg1: AnyRef): Unit
+
+  def settings: ConnectionPoolSettings
+}
+
+/* Internal API */
+@InternalApi
+private[pool] sealed abstract class SlotState extends Product {
+  def isIdle: Boolean
+  def isConnected: Boolean
+
+  def onPreConnect(ctx: SlotContext): SlotState = illegalState(ctx, "preConnect")
+  def onConnectedAttemptSucceeded(ctx: SlotContext, outgoingConnection: Http.OutgoingConnection): SlotState = illegalState(ctx, "connected attempt succeeded")
+  def onConnectionAttemptFailed(ctx: SlotContext, cause: Throwable): SlotState = illegalState(ctx, "connection attempt failed")
+
+  def onNewRequest(ctx: SlotContext, requestContext: RequestContext): SlotState = illegalState(ctx, "new request")
+
+  /** Will be called either immediately if the request entity is strict or otherwise later */
+  def onRequestEntityCompleted(ctx: SlotContext): SlotState = illegalState(ctx, "request entity completed")
+  def onRequestEntityFailed(ctx: SlotContext, cause: Throwable): SlotState = illegalState(ctx, "request entity failed")
+
+  def onResponseReceived(ctx: SlotContext, response: HttpResponse): SlotState = illegalState(ctx, "receive response")
+  def onResponseEntitySubscribed(ctx: SlotContext): SlotState = illegalState(ctx, "responseEntitySubscribed")
+
+  /** Will be called either immediately if the response entity is strict or otherwise later */
+  def onResponseEntityCompleted(ctx: SlotContext): SlotState = illegalState(ctx, "response entity completed")
+  def onResponseEntityFailed(ctx: SlotContext, cause: Throwable): SlotState = illegalState(ctx, "response entity failed")
+
+  def onConnectionCompleted(ctx: SlotContext): SlotState = illegalState(ctx, "connection completed")
+  def onConnectionFailed(ctx: SlotContext, cause: Throwable): SlotState = illegalState(ctx, "connection failed")
+
+  def onTimeout(ctx: SlotContext): SlotState = illegalState(ctx, "timeout")
+
+  def onShutdown(ctx: SlotContext): Unit = ()
+
+  /** A slot can define a timeout for that state after which onTimeout will be called. */
+  def stateTimeout: Duration = Duration.Inf
+
+  protected def illegalState(ctx: SlotContext, what: String): SlotState = {
+    ctx.debug(s"Got unexpected event [$what] in state [$name]]")
+    throw new IllegalStateException(s"Cannot [$what] when in state [$name]")
+  }
+
+  def name: String = productPrefix
+}
+
+/**
+ * Internal API
+ *
+ * Implementation of slot logic that is completed decoupled from the machinery bits which are implemented in the GraphStageLogic
+ * and exposed only through [[SlotContext]].
+ */
+@InternalApi
+private[pool] object SlotState {
+  sealed abstract class ConnectedState extends SlotState {
+    def isConnected: Boolean = true
+  }
+  sealed trait IdleState extends SlotState {
+    final override def isIdle = true
+  }
+  sealed trait BusyState extends SlotState {
+    final override def isIdle = false // no HTTP pipelining right now
+    def ongoingRequest: RequestContext
+  }
+
+  case object Unconnected extends SlotState with IdleState {
+    def isConnected: Boolean = false
+
+    override def onPreConnect(ctx: SlotContext): SlotState = {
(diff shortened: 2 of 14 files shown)


jrudolph:
Tests currently fail, because TLS tests only work if akka.example.org resolves to localhost.

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolFlow.scala
```diff
@@ -16,7 +17,9 @@ import akka.stream.scaladsl._
 import akka.http.scaladsl.model._
 import akka.http.scaladsl.Http
 
-private object PoolFlow {
+/** Internal Api */
```
ktoso: that's spelled in CAPS everywhere usually

akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/HostConnectionPool.scala
```diff
@@ -0,0 +1,335 @@
...
+package akka.http.impl.engine.client.pool
+
+import java.util
+
+import akka.NotUsed
+import akka.annotation.InternalApi
+import akka.dispatch.ExecutionContexts
+import akka.event.LoggingAdapter
+import akka.http.impl.engine.client.PoolFlow.{ RequestContext, ResponseContext }
+import akka.http.impl.util.StageLoggingWithOverride
```
ktoso: huh, I forgot we have that.

code info for others:

```
// TODO Try to reconcile with what Akka provides in StageLogging.
// We thought this could be removed when https://github.com/akka/akka/issues/18793 had been implemented
// but we need a few more changes to be able to override the default logger. So for now we keep it here.
```

akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/HostConnectionPool.scala
```diff
@@ -0,0 +1,335 @@
...
+          // FIXME: optimize by keeping track of idle connections
+          slots.exists(_.isIdle)
+
+        def dispatchResponse(req: RequestContext, res: HttpResponse): Unit =
+          dispatchResponseContext(ResponseContext(req, Success(res)))
+
+        def dispatchFailure(req: RequestContext, cause: Throwable): Unit = {
+          // FIXME: or retry
+          if (req.retriesLeft > 0) {
+            log.debug("Request has {} retries left, retrying...", req)
```
ktoso: `req.retriesLeft`

akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/HostConnectionPool.scala
```diff
@@ -0,0 +1,335 @@
...
+          def shutdown(): Unit = {
+            // TODO: should we offer errors to the connection?
+            closeConnection()
+
+            state.onShutdown(this)
+          }
+
+          def initialize(): Unit =
+            if (slotId < settings.minConnections)
+              updateState(_.onPreConnect(this))
```
ktoso: Is the above `+        // TODO: implement min-connections` not needed anymore? Looks like that's the opening connections eagerly already as expected

akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/HostConnectionPool.scala
```diff
@@ -0,0 +1,335 @@
...
+            updateState(_.onResponseReceived(this, response))
+
+          def onResponseEntityCompleted(): Unit =
+            updateState(_.onResponseEntityCompleted(this))
+          def onResponseEntityFailed(cause: Throwable): Unit =
+            updateState(_.onResponseEntityFailed(this, cause))
+
+          def onConnectionCompleted(): Unit = updateState(_.onConnectionCompleted(this))
+          def onConnectionFailed(cause: Throwable): Unit = updateState(_.onConnectionFailed(this, cause))
+
```
ktoso: this callback style reads excellent each time I see it in action ;)

akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/HostConnectionPool.scala
```diff
@@ -0,0 +1,335 @@
...
+            // put additional bookkeeping here (like keeping track of idle connections)
+          }
+          protected def setState(newState: SlotState): Unit =
+            updateState(_ ⇒ newState)
+
+          def debug(msg: String): Unit =
+            log.debug("[{} ({})] {}", slotId, state.productPrefix, msg)
+
+          def debug(msg: String, arg1: AnyRef): Unit =
+            log.debug(s"[{} ({})] $msg", slotId, state.productPrefix, arg1)
```
ktoso: this one seems wrong somehow?

akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/HostConnectionPool.scala
```diff
@@ -0,0 +1,335 @@
...
+            }
+            override def onUpstreamFailure(ex: Throwable): Unit = {
+              slot.debug("Connection failed")
+              if (slot.isConnected)
+                slot.onConnectionFailed(ex)
+            }
+          })
+          responseIn.pull()
+
+          slot.debug("Establishing connection")
```
ktoso: add info where to?

akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/HostConnectionPool.scala
```diff
@@ -0,0 +1,335 @@
...
+          val connection =
+            Source.fromGraph(requestOut.source)
+              .viaMat(connectionFlow)(Keep.right)
+              .toMat(responseIn.sink)(Keep.left)
+              .run()(subFusingMaterializer)
+
+          connection.onComplete(safely {
```
ktoso: really like this pattern btw

akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/SlotState.scala
```diff
@@ -0,0 +1,195 @@
...
+      Unconnected
+    }
+
+    // we ignore these signals here and expect that it will also be flagged on the entity stream
+    // FIXME: should we still add timeouts for these cases?
+    override def onConnectionFailed(ctx: SlotContext, cause: Throwable): SlotState = this
+    override def onConnectionCompleted(ctx: SlotContext): SlotState = this
+  }
+
+}
```
ktoso: great way to model these 👍  It's super clear what causes what now

akka-http-core/src/test/scala/akka/http/impl/engine/client/HostConnectionPoolSpec.scala
```diff
@@ -0,0 +1,588 @@
...
+  testSet(poolImplementation = NewPoolImplementation, clientServerImplementation = PassThrough)
+  testSet(poolImplementation = NewPoolImplementation, clientServerImplementation = AkkaHttpEngineNoNetwork)
+  testSet(poolImplementation = NewPoolImplementation, clientServerImplementation = AkkaHttpEngineTCP)
+  //testSet(poolImplementation = NewPoolImplementation, clientServerImplementation = AkkaHttpEngineTLS)
+
+  testSet(poolImplementation = OldPoolImplementation, clientServerImplementation = PassThrough)
+  testSet(poolImplementation = OldPoolImplementation, clientServerImplementation = AkkaHttpEngineNoNetwork)
+  testSet(poolImplementation = OldPoolImplementation, clientServerImplementation = AkkaHttpEngineTCP)
+  //testSet(poolImplementation = OldPoolImplementation, clientServerImplementation = AkkaHttpEngineTLS)
+
```
ktoso: 👍

akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/HostConnectionPool.scala
```diff
@@ -0,0 +1,335 @@
...
+/**
+ * Internal API
+ *
+ * New host connection pool implementation.
+ *
+ * Backpressure logic of the external interface:
+ *
+ *  * pool pulls if there's a free slot
+ *  * pool buffers responses until they are pulled. The buffer is unlimited in theory with the reasoning that
+ *    reasonable behavior can be expected from downstream consumers, i.e. at least one pull for every request sent in.
```
ktoso: The assumption sounds ok to me

akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/SlotState.scala
```diff
@@ -0,0 +1,195 @@
...
+
+  // FIXME: unify with Connecting
+  case object PreConnecting extends ConnectedState with IdleState with WithRequestDispatching {
+    override def onConnectedAttemptSucceeded(ctx: SlotContext, outgoingConnection: Http.OutgoingConnection): SlotState = {
+      ctx.debug("Slot connection was (pre-)established")
+      Idle
+    }
+    override def onConnectionAttemptFailed(ctx: SlotContext, cause: Throwable): SlotState = {
+      ctx.debug("Connection attempt failed.")
+      // FIXME: register failed connection attempt, schedule request for rerun, backoff new connection attempts
```
ktoso: so slot context needs a way to schedule a thing, and that'd call into openConnection I guess?
State should become something signifying that we're in this backoff period I guess?

jrudolph:
PLS BUILD

akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/HostConnectionPool.scala
jrudolph: Oops, thanks, good one.

akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/HostConnectionPool.scala
jrudolph: Yep, a bit verbose but it for reading the explicit style is quite good.

akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/HostConnectionPool.scala
jrudolph: I think it's right, we don't check but `msg` should contain another placeholder.

akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/HostConnectionPool.scala
```diff
@@ -0,0 +1,335 @@
...
+            requestOut.push(newRequest)
+          }
+          def close(): Unit = {
+            requestOut.complete()
+            responseIn.cancel()
+          }
+          def isClosed: Boolean = requestOut.isClosed || responseIn.isClosed
+        }
+        def openConnection(slot: Slot): SlotConnection = {
+          val requestOut = new SubSourceOutlet[HttpRequest]("PoolSlot.requestOut")
```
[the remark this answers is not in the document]
jrudolph: Yeah, why not.

akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/HostConnectionPool.scala
jrudolph: I think right now we don't have that information here (it's hidden in the `connectionFlow` instance) but the logger from `PoolInterfaceActor` will add that information.

akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/SlotState.scala
jrudolph: We might add a dedicated `SlotState` but the important thing here is that the backoff state should be shared between all slots.

jrudolph:
Squashed.

jrudolph:
I checked again that all changes should only be active by using the switch so I went ahead and merged the change. To be followed up when it got a bit more exposure.

akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/SlotState.scala
johanandren: I agree, god design this!

PR #1534, opened 2017-11-20
Use hashmap instead of binary tree of parsers

raboof:
Updated version of #1202 after #1428. Might not be as relevant anymore? But
replacing the macro-binary-search with a HashMap still makes the code a bit
simpler.


jrudolph:
PLS BUILD

jrudolph:
> Might not be as relevant anymore?  But
> replacing the macro-binary-search with a HashMap still makes the code a bit
> simpler.

I guess it's not as relevant any more but I think it still makes sense. Let's merge after another PR validation round.

raboof:
PLS BUILD

Issue #1545, opened 2017-11-22
Add support for proxying HTTPS server connections on the client side... for Web Sockets too


JeanFrancoisGuena:
As suggested by the title, this issue is related to #192 and its PRs #917, #921 and #1200.

We are trying to connect to an "external" server through an HTTP(S) Proxy that does not allow any "wss://..." request.

So, the `singleWebSocketRequest` method may be of some help as soon as its "settings" parameter handles a custom `ClientTransport`, as explained [here, in the doc](https://doc.akka.io/docs/akka-http/current/scala/http/client-side/client-transport.html#use-http-s-proxy-with-) for `singleRequest` method...

Unfortunately, `singleWebSocketRequest` method takes a `ClientConnectionSettings` "settings" parameter, while `singleRequest` method takes a `ConnectionPoolSettings`, and there is no way to tell a `ClientConnectionSettings` that its `ClientTransport` should not always be the default `ClientTransport.TCP`

So we tried...

- we made `ClientConnectionSettings` just look like `ConnectionPoolSettings` regarding its "transport" parameter
- we made a very little change in the way `singleWebSocketRequest` method calls for the private `_outgoingTlsConnectionLayer` method in Http.scala file (through `webSocketClientFlow` method actually)

and it works like a charm... **_for us_**.

So, I worked on a PR I should be able to submit very soon (the longest step from now on is: find the good person in my company to sign the Corporate CLA :-/)

The goal here is, apart from sharing a solution, to verify that "it works like a charm", but not only **_for us_** ;-)

PR #1548, opened 2017-11-23
Enable Web Socket connection through HTTP(S) Proxy on the client side #1545

JeanFrancoisGuena:
Here is the PR for issue #1545...

Be aware that, for now, I have a "Binary compatibility check failed!"

All tests passed locally and I have updated the documentation...

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/settings/ClientConnectionSettingsImpl.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/settings/ClientConnectionSettingsImpl.scala
@@ -10,6 +10,7 @@ import java.util.Random
 import akka.annotation.InternalApi
 import akka.http.impl.engine.ws.Randoms
 import akka.http.impl.util._
+import akka.http.scaladsl.ClientTransport
 import akka.http.scaladsl.model.headers.`User-Agent`
 import akka.http.scaladsl.settings.ClientConnectionSettings.LogUnencryptedNetworkBytes
 import akka.http.scaladsl.settings.ParserSettings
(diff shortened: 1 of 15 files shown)


lightbend-cla-validator:
Hi @JeanFrancoisGuena,

akka-http-core/src/main/scala/akka/http/impl/settings/ClientConnectionSettingsImpl.scala
```diff
@@ -30,7 +31,8 @@ private[akka] final case class ClientConnectionSettingsImpl(
   websocketRandomFactory:     () ⇒ Random,
   socketOptions:              immutable.Seq[SocketOption],
   parserSettings:             ParserSettings,
-  localAddress:               Option[InetSocketAddress])
+  localAddress:               Option[InetSocketAddress],
+  transport:                  ClientTransport)
```
jrudolph: We cannot add another setting at that level because we already have it on `ConnectionPoolSettings` which also wraps `ClientConnectionSettings`. Having it on two layers would be quite confusing. The pattern used for other non-pool client APIs is to add another overload that takes a `ClientTransport` parameter directly.

We might also want to reconsider the decision to put the transport into the pool settings in the first place, instead of putting it into `ClientConnectionSettings` where it could be used by all the APIs.

jrudolph: I created https://github.com/akka/akka-http/issues/1562 to track this issue. Maybe you would be interested in doing that change as a preliminary to this one, @JeanFrancoisGuena? Thanks anyway for making us aware of it.

raboof:
OK TO TEST

raboof:
(failure is #1560, looking into it)

akka-http-core/src/main/scala/akka/http/impl/settings/ClientConnectionSettingsImpl.scala
JeanFrancoisGuena: I made some change to move `ClientTransport` from `ConnectionPoolSettings` to `ClientConnectionSettings` and pushed.

Unfortunately, I still have a "Binary compatibility check failed!"...

JeanFrancoisGuena:
Hi @jrudolph , what about my last push ? Is it all right the way I did it or do you prefer that I submit another PR related to #1562 ?

Issue #1554, opened 2017-11-26
Add support for Retry-After Http header


ivanopagano:
As per [RFC7231](https://tools.ietf.org/html/rfc7231#section-7.1.3), add framework support to read and add the header from/to the response.

ivanopagano:
The WIP branch is on my [fork](https://github.com/ivanopagano/akka-http/tree/retry-after-header-http-core) but I can't yet open a PR because I need support for a failing test for which I can't fathom the cause.

The failing test is [here](https://github.com/ivanopagano/akka-http/commit/10616a860d6ffd0e9efb316b5b5d7ad84c1caeb2#diff-49a3dd98ddc9a8db7812dc5d874fa512) and instead of converting the `h.toString` to the expected value of `Retry-After: 120` it serialize to `Content-Type: 120` (!).

I looked into the `HeaderParser` and `HttpHeader` definitions a few times without figuring out what's missing: any help is appreaciated.

ivanopagano:
I would also appreciate to be directed to any other missteps

ktoso:
> The WIP branch is on my fork but I can't yet open a PR because I need support for a failing test for which I can't fathom the cause.

To clarify: It's absolutely fine to open a PR and ask in there about the remaining issue :)

I'll look into it

ivanopagano:
the test should be ok now, thanks to Rooney.
I'll run the `validatePullRequest` now to figure out if something else is missing.

ivanopagano:
All issues have been solved and the PR is ready on its way to be picked up

Thanks everyone

jrudolph:
Fixed in #1555.

PR #1555, opened 2017-11-26
Http header retry-after #1554

ivanopagano:
* Add the class and companion to the headers
 * Create the corresponding java class
 * Add case in the HeaderParser and SimpleHeaders
 * Create test cases

Note: the current commit won't pass the HttpHeaderSpec, for reasons still
under investigation.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/model/parser/SimpleHeaders.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/model/parser/SimpleHeaders.scala
@@ -6,7 +6,7 @@ package akka.http.impl.model.parser
 
 import akka.annotation.InternalApi
 import akka.parboiled2.Parser
-import akka.http.scaladsl.model.{ RemoteAddress, Uri, DateTime }
+import akka.http.scaladsl.model.{ RemoteAddress, Uri }
 import akka.http.scaladsl.model.headers._
 import CharacterClasses.`scheme-char`
 
--- a/akka-http-core/src/main/scala/akka/http/impl/model/parser/SimpleHeaders.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/model/parser/SimpleHeaders.scala
@@ -189,7 +189,7 @@ private[parser] trait SimpleHeaders { this: Parser with CommonRules with CommonA
 
   //https://tools.ietf.org/html/rfc7231#section-7.1.3
   def `retry-after` = rule {
-    (`HTTP-date` ~> (RetryAfterDateTime(_)) | `delta-seconds` ~> (RetryAfterDuration(_))) ~ EOI
+    (`HTTP-date` ~> (RetryAfterDateTime(_)) | `delta-seconds` ~> (RetryAfterDuration(_))) ~ EOI ~> (`Retry-After`(_))
   }
 
   // http://tools.ietf.org/html/rfc7231#section-7.4.2
--- a/akka-http-core/src/main/scala/akka/http/impl/model/parser/SimpleHeaders.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/model/parser/SimpleHeaders.scala
@@ -6,7 +6,7 @@ package akka.http.impl.model.parser
 
 import akka.annotation.InternalApi
 import akka.parboiled2.Parser
-import akka.http.scaladsl.model.{ RemoteAddress, Uri }
+import akka.http.scaladsl.model.{ RemoteAddress, Uri, DateTime }
 import akka.http.scaladsl.model.headers._
 import CharacterClasses.`scheme-char`
 
@@ -185,6 +185,11 @@ private[parser] trait SimpleHeaders { this: Parser with CommonRules with CommonA
     // we are bit more relaxed than the spec here by also parsing a potential fragment
     // but catch it in the `Referer` instance validation (with a `require` in the constructor)
     uriReference ~ EOI ~> (Referer(_))
+  }
+
+  //https://tools.ietf.org/html/rfc7231#section-7.1.3
+  def `retry-after` = rule {
+    (`HTTP-date` ~> (Right(_)) | `delta-seconds` ~> (Left(_))) ~ EOI ~> (`Retry-After`(_))
   }
 
   // http://tools.ietf.org/html/rfc7231#section-7.4.2
--- a/akka-http-core/src/main/scala/akka/http/impl/model/parser/SimpleHeaders.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/model/parser/SimpleHeaders.scala
@@ -189,7 +189,7 @@ private[parser] trait SimpleHeaders { this: Parser with CommonRules with CommonA
 
   //https://tools.ietf.org/html/rfc7231#section-7.1.3
   def `retry-after` = rule {
-    (`HTTP-date` ~> (Right(_)) | `delta-seconds` ~> (Left(_))) ~ EOI ~> (`Retry-After`(_))
+    (`HTTP-date` ~> (RetryAfterDateTime(_)) | `delta-seconds` ~> (RetryAfterDuration(_))) ~ EOI
   }
 
   // http://tools.ietf.org/html/rfc7231#section-7.4.2
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
@@ -704,30 +704,35 @@ final case class Referer(uri: Uri) extends jm.headers.Referer with RequestHeader
 }
 
 //https://tools.ietf.org/html/rfc7231#section-7.1.3
-sealed trait `Retry-After` extends jm.headers.RetryAfter
-final case class RetryAfterDuration(delay: Long) extends `Retry-After` with ResponseHeader {
-  def renderValue[R <: Rendering](r: R): r.type = r ~~ delay
-
-  protected def companion = `Retry-After`
-
-  /** Java API suppport */
-  override protected def delaySeconds(): Option[java.lang.Long] = Some(Long.box(delay))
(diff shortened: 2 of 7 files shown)


lightbend-cla-validator:
Hi @ivanopagano,

ivanopagano:
I have a failing test that I can't figure out.

Refer to issue #1554

ktoso:
OK TO TEST

ivanopagano:
The validator seems to point to a formatting error, but I can't see any... I retried running the formatter, but the code highlighted from jenkins wasn't changed

jonas:
The formatting diff printed in [the build](https://jenkins.akka.io:8498/job/pr-validator-akka-http/1746/) is:
```diff
diff --git a/akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala b/akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
index 6839146..47a5020 100644
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
@@ -711,8 +711,8 @@ object `Retry-After` extends ModeledCompanion[`Retry-After`] {
 final case class `Retry-After`(delaySecondsOrDateTime: Either[Long, DateTime]) extends jm.headers.RetryAfter with ResponseHeader {
   require(delaySecondsOrDateTime.left.toOption.forall(_ > 0), "Retry-after header must not contain a negative delay in seconds")
   def renderValue[R <: Rendering](r: R): r.type = delaySecondsOrDateTime match {
-    case Left(delay)     => r ~~ delay
-    case Right(dateTime) => dateTime.renderRfc1123DateTimeString(r)
+    case Left(delay)     ⇒ r ~~ delay
+    case Right(dateTime) ⇒ dateTime.renderRfc1123DateTimeString(r)
   }
   protected def companion = `Retry-After`
 
```

akka-http-core/src/main/scala/akka/http/impl/model/parser/SimpleHeaders.scala
```diff
@@ -6,7 +6,7 @@ package akka.http.impl.model.parser
 
 import akka.annotation.InternalApi
 import akka.parboiled2.Parser
-import akka.http.scaladsl.model.{ RemoteAddress, Uri }
+import akka.http.scaladsl.model.{ RemoteAddress, Uri, DateTime }
```
jonas: It doesn't look like this import is used after all.

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
```diff
@@ -703,6 +703,25 @@ final case class Referer(uri: Uri) extends jm.headers.Referer with RequestHeader
   def getUri: akka.http.javadsl.model.Uri = uri.asJava
 }
 
+//https://tools.ietf.org/html/rfc7231#section-7.1.3
+object `Retry-After` extends ModeledCompanion[`Retry-After`] {
+  def apply(delaySeconds: Long): `Retry-After` = apply(Left(delaySeconds))
+  def apply(timestamp: DateTime): `Retry-After` = apply(Right(timestamp))
+}
+final case class `Retry-After`(delaySecondsOrDateTime: Either[Long, DateTime]) extends jm.headers.RetryAfter with ResponseHeader {
+  require(delaySecondsOrDateTime.left.toOption.forall(_ > 0), "Retry-after header must not contain a negative delay in seconds")
```
jonas: Can delaySeconds be zero?
The check can be reduced to:

```scala
delaySecondsOrDateTime.left.forall(...)
```

akka-http-core/src/main/scala/akka/http/impl/model/parser/SimpleHeaders.scala
ivanopagano: I'll check and remove, in case

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
ivanopagano: nice, thanks

yes, I checked again and I made a mistake here with the validation check

will fix soon

ivanopagano:
@jonas I see the diff too in the jenkins output, but I didn't use the double arrow char on my source, and other than that I see no formatting error, could someone point me in the right direction here?

jonas:
Maybe the font or editor you are using shows `=>` as a combined symbol?

ivanopagano:
For the records: I have signed the LA already

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
```diff
@@ -703,6 +703,25 @@ final case class Referer(uri: Uri) extends jm.headers.Referer with RequestHeader
   def getUri: akka.http.javadsl.model.Uri = uri.asJava
 }
 
+//https://tools.ietf.org/html/rfc7231#section-7.1.3
+object `Retry-After` extends ModeledCompanion[`Retry-After`] {
+  def apply(delaySeconds: Long): `Retry-After` = apply(Left(delaySeconds))
+  def apply(timestamp: DateTime): `Retry-After` = apply(Right(timestamp))
+}
+final case class `Retry-After`(delaySecondsOrDateTime: Either[Long, DateTime]) extends jm.headers.RetryAfter with ResponseHeader {
```
jrudolph: We would usually prefer modelling the data type explicitly instead of using the more general `Either`. The benefit is that user code will read nicer. How about adding an ADT like this:

```scala
sealed abstract class RetryAfter
final case class RetryAfterDuration(duration: Duration) extends RetryAfter
final case class RetryAfterTimestamp(timestamp: DateTime) extends RetryAfter
```

WDYT?

jrudolph: In lack of a better option, the java version would still be fine.

ivanopagano: Moving to an ADT could work fine

Just to be clear, you expect the java supertype to still look the same or to do an equivalent operation there (splitting the java hierarchy too?).
I don't see any benefit there to be honest, but I may be missing some way to make it more usable in java too?

ivanopagano: @jrudolph I'm reading again your comment 'cause I'm not sure anymore that I correctly understood: is the ADT a substitute of the java class definition or just an implementation detail of the scala version?

I would keep the java version as is, because I'd avoid forcing users to do `instanceof` checks against the actual class, and would prefer the safer `Optional`-based api, in lack of a better alternative.
This is why I'm about to change only the scala implementation to encode it with the ADT in place of the internal `Either` member.

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
```diff
@@ -703,6 +703,33 @@ final case class Referer(uri: Uri) extends jm.headers.Referer with RequestHeader
   def getUri: akka.http.javadsl.model.Uri = uri.asJava
 }
 
+//https://tools.ietf.org/html/rfc7231#section-7.1.3
+sealed trait `Retry-After` extends jm.headers.RetryAfter
+final case class RetryAfterDuration(delay: Long) extends `Retry-After` with ResponseHeader {
+  def renderValue[R <: Rendering](r: R): r.type = r ~~ delay
+
+  protected def companion = `Retry-After`
```
ivanopagano: is this "shared" reference to the same ModeledCompanion at risk of creating some problem later on?

ivanopagano:
Is the test failure related to the specifics of this PR? Looks like an occasional timeout... locally the test pass (I ran the validatePR task too)

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
jrudolph: Yeah, sorry if I wasn't clear. The suggestion was only for the Scala side. I don't have a better suggestion for the Java side so we can keep it this way.

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
```diff
@@ -703,6 +703,33 @@ final case class Referer(uri: Uri) extends jm.headers.Referer with RequestHeader
   def getUri: akka.http.javadsl.model.Uri = uri.asJava
 }
 
+//https://tools.ietf.org/html/rfc7231#section-7.1.3
+sealed trait `Retry-After` extends jm.headers.RetryAfter
+final case class RetryAfterDuration(delay: Long) extends `Retry-After` with ResponseHeader {
```
jrudolph: Our current pattern for these cases is a bit different. We don't make the header class an ADT (sorry if that was unclear), but give it a parameter with the type of the ADT. So it would be more like

```scala
final case class `Retry-After`(retryAfter: RetryAfter) extends jm.headers.RetryAfter with ResponseHeader { ... }
```

and then in an extra file

```scala
sealed abstract class RetryAfter
final case class RetryAfterDuration(delay: Long) extends RetryAfter
...
```

It's a bit more verbose but it follows our best practices.

ivanopagano: that was what I understood in the beginning, but then something made me change my mind, it's ok, I'll fix that soon, thanks

ivanopagano: just to know: why `sealed abstract class` and not `sealed trait` for the hierarchy root?

and also... can I make the ADT leaves as _Value Classes_?

ivanopagano:
@jrudolph let me know if this works now as you intended

jrudolph:
Fixes #1554.

Issue #1558, opened 2017-11-28
Suprising (additional) percent escaping on Uri#authority rendering


ktoso:
It was reported that:

```scala
class UriTest extends WordSpecLike with MustMatchers {
val userInfo = "user:p%40ssword"
val hostAndPath = "host/direc%20tory/fi%20le.tgz?query=test%25#frag%20ment"
val uriString = s"https://$userInfo@$hostAndPath"

"A Java URI" must {
"convert into an Akka HTTP Uri" in {
val javaUri = JavaURI.create(uriString)
val akkaUri = Uri(javaUri.toString)
akkaUri.toString mustEqual javaUri.toString
}
}
```

The issue seems to be not in parsing but rendering, I'll submit a proposal but not 100% sure if it's the best way to address this.

jrudolph:
Regarding the general issue: we don't promise round trips for string representations. The minimum invariant we support is equivalence as described in [RFC 3986 Section 6.1](https://tools.ietf.org/html/rfc3986#section-6.1). I.e.

```
val uriString = 
val javaUri: java.net.URI = java.net.URI.create(uriString)
val akkaUri = Uri(uriString)

akkaUri == Uri(javaUri.toString)
```

A stricter behavior could be to render fully normalized URIs. But even then, strings would only be the same if java would also promise to render normalized URIs and that both implementations actually implement the same logic. I haven't checked to which degree we render completely normalized URIs.

For that reason, I think we should leave the behavior of java.net.URI out of the picture for now.

jrudolph:
For me it looks as if our parser wouldn't properly percent decode the user info before passing it into the model as it is doing everywhere else.

jrudolph:
I.e. for me the bug is this:

```scala
scala> Uri("http://test:p%40ssword@example.com").authority.userinfo
res3: String = test:p%40ssword
```

while the result should be "test:p@ssword".

jrudolph:
If you compare the parser logic e.g. to its processing of a host name it seems the percent decoding was just forgotten. Compare

https://github.com/akka/akka-http/blob/f9e8783db221ace458e51a2b6b8656fcaa62e2a3/akka-http-core/src/main/scala/akka/http/impl/model/parser/UriParser.scala#L170

with

https://github.com/akka/akka-http/blob/f9e8783db221ace458e51a2b6b8656fcaa62e2a3/akka-http-core/src/main/scala/akka/http/impl/model/parser/UriParser.scala#L143

[...]

PR #1559, opened 2017-11-28
=htc #1558 try to make authority rendering closer to round-triping to Java

ktoso:
Resolves https://github.com/akka/akka-http/issues/1558 though perhaps this should be solved in a different way...

These escaping things are always a bit tricky when not clearly defined when one is to escape and when not. The users expected to be able to pass in the Java URI, but that renders as escaped, so that would not work. I think we should in addition provide a method that takes Uri as a parameter?

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/Uri.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/Uri.scala
@@ -502,7 +502,9 @@ object Uri {
     def /(segment: String): Path = this ++ Path.Slash(segment :: Path.Empty)
     def startsWith(that: Path): Boolean
     def dropChars(count: Int): Path
-    override def toString = UriRendering.PathRenderer.render(new StringRendering, this).get
+    override def toString = {
+      UriRendering.PathRenderer.render(new StringRendering, this).get
+    }
   }
   object Path {
     val SingleSlash = Slash(Empty)
@@ -723,6 +725,8 @@ object Uri {
   private[http] def decode(string: String, charset: Charset, ix: Int)(sb: JStringBuilder = new JStringBuilder(string.length).append(string, 0, ix)): String =
     if (ix < string.length) string.charAt(ix) match {
       case '%' ⇒
+        println(s"  AT ix=${ix}, rest: ${string.substring(ix)}")
+
         def intValueOfHexWord(i: Int) = {
           def intValueOfHexChar(j: Int) = {
             val c = string.charAt(j)
@@ -880,7 +884,7 @@ object UriRendering {
   def renderAuthority[R <: Rendering](r: R, authority: Authority, path: Path, scheme: String, charset: Charset): r.type =
     if (authority.nonEmpty) {
       import authority._
-      if (!userinfo.isEmpty) encode(r, userinfo, charset, `userinfo-char`, detectAlreadyEscapedChars = true) ~~ '@'
+      if (!userinfo.isEmpty) encode(r, userinfo, charset, `userinfo-char`) ~~ '@'
       r ~~ host
       if (port > 0) r ~~ ':' ~~ port else r
     } else scheme match {
@@ -914,8 +918,7 @@ object UriRendering {
   }
 
   private[http] def encode(r: Rendering, string: String, charset: Charset, keep: CharPredicate,
-                           replaceSpaces:             Boolean = false,
-                           detectAlreadyEscapedChars: Boolean = false): r.type = {
+                           replaceSpaces: Boolean = false): r.type = {
     val asciiCompatible = isAsciiCompatible(charset)
     @tailrec def rec(ix: Int): r.type = {
       def appendEncoded(byte: Byte): Unit =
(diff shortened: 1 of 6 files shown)


akka-http-core/src/test/scala/akka/http/scaladsl/model/UriSpec.scala
```diff
@@ -721,11 +722,23 @@ class UriSpec extends WordSpec with Matchers {
       Uri("https://host:3030/").withPort(4450).effectivePort shouldEqual 4450
     }
```
ktoso: We don't have a test for java URI <=> Akka Uri roundtrip, If we want to support that roundtrip the test should be added and perhaps fuzzy tested with scalacheck?

akka-http-core/src/main/scala/akka/http/scaladsl/model/Uri.scala
```diff
@@ -909,14 +914,28 @@ object UriRendering {
...
+            var c1: Char = 0
+            var c2: Char = 0
+            val isEscapedChar = string.length > ix + 2 && {
+              c1 = string.charAt(ix + 1)
+              c2 = string.charAt(ix + 2)
+              c1 < 127 && c2 < 127
+            }
+            if (isEscapedChar) { r ~~ '%' ~~ c1 ~~ c2; 3 }
+            else { appendEncoded('%'.toByte); 1 }
+          }
```
ktoso: I'm very torn about this actually... This new logic is used only in the authority really.
The reason the authority is assumed to be escaped already is that otherwise we'd get double escapes

akka-http-core/src/main/scala/akka/http/scaladsl/model/Uri.scala
```diff
@@ -215,6 +215,12 @@ object Uri {
   def apply(input: ParserInput, charset: Charset, mode: Uri.ParsingMode): Uri =
     new UriParser(input, charset, mode).parseUriReference()
 
+  def apply(uri: java.net.URI): Uri = {
```
ktoso: TODO, should we provide this method that'd take the various URI parts from the uri and pass them in directly rather than going via strings and parsing? Sounds like a good idea I thought

akka-http-tests/src/test/scala/akka/http/scaladsl/FormDataSpec.scala
```diff
@@ -32,3 +37,48 @@ class FormDataSpec extends AkkaSpec {
   }
 
 }
+
+class UriSpec extends WordSpec with MustMatchers {
```
ktoso: FIXME remove this spec.

Output FYI 

```
[info] - must convert into an Akka HTTP Uri
[info]   +               string: https://user:p%40ssword@host/direc%20tory/fi%20le.tgz?query=test%25#frag%20ment
[info]   +                 java: https://user:p%40ssword@host/direc%20tory/fi%20le.tgz?query=test%25#frag%20ment
[info]   +   toASCIIString java: https://user:p%40ssword@host/direc%20tory/fi%20le.tgz?query=test%25#frag%20ment
[info]   +    from initial akka: https://user:p%40ssword@host/direc%20tory/fi%20le.tgz?query=test%25#frag%20ment
[info]   + from java ASCII akka: Success(https://user:p%40ssword@host/direc%20tory/fi%20le.tgz?query=test%25#frag%20ment)
[info]   +         manual akka: Success(https://user:p%40ssword@host/direc%20tory/fi%20le.tgz?query=test%25#frag%20ment)
```

akka-http-core/src/main/scala/akka/http/scaladsl/model/Uri.scala
jrudolph: We could do this but the question is if it is any better (better performing, correcter) than going through the well specified URI string representation.

akka-http-core/src/main/scala/akka/http/scaladsl/model/Uri.scala
```diff
@@ -875,9 +880,9 @@ object UriRendering {
   def renderAuthority[R <: Rendering](r: R, authority: Authority, path: Path, scheme: String, charset: Charset): r.type =
     if (authority.nonEmpty) {
       import authority._
-      if (!userinfo.isEmpty) encode(r, userinfo, charset, `userinfo-char`) ~~ '@'
+      if (!userinfo.isEmpty) encode(r, userinfo, charset, `userinfo-char`, detectAlreadyEscapedChars = true) ~~ '@'
```
jrudolph: This will not work as it introduces ambiguities. We cannot just guess if something accidentally introduced encodings or if the value was meant that way.

akka-http-core/src/test/scala/akka/http/scaladsl/model/UriSpec.scala
```diff
@@ -721,11 +722,23 @@ class UriSpec extends WordSpec with Matchers {
...
 
+    "parse authority" in {
+      Uri.Authority.parse("localhost").toString shouldEqual "localhost"
+      Uri.Authority.parse("example.com:80").toString shouldEqual "example.com:80"
+      Uri.Authority.parse("user@host").toString shouldEqual "user@host"
+
+      val a = Uri.Authority.parse("user:p%40ssword@host")
+      a.userinfo.length shouldEqual "user:p%40ssword".length
```
jrudolph: `a.userinfo shouldEqual "user:p@ssword"` would be correct.

akka-http-core/src/main/scala/akka/http/scaladsl/model/Uri.scala
ktoso: Yeah, that's the root of my worry here.

akka-http-core/src/test/scala/akka/http/scaladsl/model/UriSpec.scala
ktoso: Hmm, I see so we SHOULD interpret it during parsing... I'll give it a shot

ktoso:
Attempted to change things but ended up in a dead corner again.
We do treat percent encodings later on in parsing so  not sure why this wasn't fine -- need to debug more, though I'll be offline the next 2 days

jrudolph:
Followed up by #1576. Closing this one...

ktoso:
Great, thanks

PR #1561, opened 2017-11-28
Be explicit about expected logging

raboof:
.. and suppress it while we're at it


akka-http-tests/src/test/scala/akka/http/javadsl/server/HttpAppSpec.scala
```diff
@@ -190,8 +191,9 @@ class HttpAppSpec extends AkkaSpec with RequestBuilding with Eventually {
       }
 
       "after binding is unsuccessful" in withSneaky { (sneaky, host, _) ⇒
-
-        sneaky.startServer(host, 1, ServerSettings.create(ConfigFactory.load))
+        EventFilter[SocketException](message = "Permission denied", occurrences = 1) intercept {
```
jrudolph: Let's hope that this is platform independent enough.

Issue #1560, opened 2017-11-28
FAILED: FileAndResourceDirectivesSpec getFromResource_should_reject_requests_to_directory_resources_from_an_archive


jrudolph:
https://jenkins.akka.io:8498/view/Akka-http/job/akka-http-nightly-against-akka-master/362/testReport/akka.http.scaladsl.server.directives/FileAndResourceDirectivesSpec/getFromResource_should_reject_requests_to_directory_resources_from_an_archive/

```
Error Message

true did not equal false


Stacktrace

org.scalatest.exceptions.TestFailedException: true did not equal false
	at org.scalatest.MatchersHelper$.indicateFailure(MatchersHelper.scala:340)
	at org.scalatest.Matchers$AnyShouldWrapper.shouldEqual(Matchers.scala:6742)
	at akka.http.scaladsl.server.directives.FileAndResourceDirectivesSpec$$anonfun$3$$anonfun$apply$mcV$sp$22$$anonfun$apply$24.apply(FileAndResourceDirectivesSpec.scala:211)
	at akka.http.scaladsl.server.directives.FileAndResourceDirectivesSpec$$anonfun$3$$anonfun$apply$mcV$sp$22$$anonfun$apply$24.apply(FileAndResourceDirectivesSpec.scala:211)
	at scala.util.DynamicVariable.withValue(DynamicVariable.scala:58)
	at akka.http.scaladsl.testkit.RouteTest$$anonfun$check$1.apply(RouteTest.scala:57)
	at akka.http.scaladsl.testkit.RouteTest$$anonfun$check$1.apply(RouteTest.scala:57)
	at akka.http.scaladsl.testkit.RouteTestResultComponent$RouteTestResult.$tilde$greater(RouteTestResultComponent.scala:50)
	at akka.http.scaladsl.server.directives.FileAndResourceDirectivesSpec$$anonfun$3$$anonfun$apply$mcV$sp$22.apply(FileAndResourceDirectivesSpec.scala:211)
	at akka.http.scaladsl.server.directives.FileAndResourceDirectivesSpec$$anonfun$3$$anonfun$apply$mcV$sp$22.apply(FileAndResourceDirectivesSpec.scala:211)
```

jrudolph:
https://jenkins.akka.io:8498/job/pr-validator-akka-http/1763/testReport/akka.http.scaladsl.server.directives/FileAndResourceDirectivesSpec/getFromResource_should_reject_requests_to_directory_resources_from_an_archive/

[...]

PR #1565, opened 2017-11-28
Explicitly check if a resource is a dir

raboof:
Refs #1560

It appears in a previous JRE version(?) `jar.getInputStream` would
fail for directories, while now an explicit check is needed.

We could even argue opening the input stream is no longer needed
at all, why did we not simply check `isDirectory` before?

### DIFF
--- a/akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectives.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectives.scala
@@ -283,7 +283,8 @@ object FileAndResourceDirectives extends FileAndResourceDirectives {
         val jar = new java.util.zip.ZipFile(filePath)
         try {
           val entry = jar.getEntry(resourcePath)
-          Option(jar.getInputStream(entry)) map { is ⇒
+          if (entry.isDirectory) None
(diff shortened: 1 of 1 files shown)


jrudolph:
> We could even argue opening the input stream is no longer needed
> at all, why did we not simply check isDirectory before?

Hmm I wonder if there could be other reasons for `getInputStream` returning `null`? IIRC a directory entry is just an empty file with a name ending with `/`. So, maybe at some point other empty files also returned `null`? In any case, it's not documented to return null.

jrudolph:
PLS BUILD

raboof:
Heh indeed `isDirectory` is implemented as `name.endsWith("/")` ;).

Let's change this to `is (entry.isDirectory) None else Some(ResourceFile(url, entry.getSize, entry.getTime))` then?

raboof:
PLS BUILD

raboof:
Added failure #1566

raboof:
PLS BUILD

jrudolph:
> Let's change this to is (entry.isDirectory) None else Some(ResourceFile(url, entry.getSize, entry.getTime)) then?

I'd leave it like it is for now. It seems to depend on the return value of the native library if null is returned and we just don't know if it was added just for the directories.

PR #1570, opened 2017-11-28
=htc simplify decompressor/deflate infrastructure

jrudolph:
The idea is to replace the handlers in `DecompressorParsingLogic` with simple parameters of the `Inflate` step.

### DIFF
--- a/akka-http/src/main/scala/akka/http/scaladsl/coding/Deflate.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/coding/Deflate.scala
@@ -88,72 +88,62 @@ private[http] object DeflateCompressor {
 
 class DeflateDecompressor(maxBytesPerChunk: Int = Decoder.MaxBytesPerChunkDefault) extends DeflateDecompressorBase(maxBytesPerChunk) {
 
-  override def createLogic(attr: Attributes) = new DecompressorParsingLogic {
-    var _inflater: Inflater = _
-    override def inflater: Inflater = if (_inflater != null) {
-      _inflater
-    } else {
-      throw new IllegalStateException("Inflater used before the header was analysed for wrapping")
-    }
-
-    override val inflateState = new Inflate(true) {
+  override def createLogic(attr: Attributes) = new ParsingLogic {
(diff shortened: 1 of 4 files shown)


jrudolph:
These changes should be backported to the copy in akka-stream.

jrudolph:
The coding implementation classes were still public (accidentally, I'd say). So, I made them private to allow the binary incompatible changes. Strictly taken this is a violation of our binary compatibility guarantees. I'd say that it is unlikely that other code relies on that code but we cannot know of course. We could also postpone the actual changes to 10.1.0 but maybe that's overkill for these bits of infrastructure.

jrudolph:
PLS BUILD

akka-http/src/main/scala/akka/http/scaladsl/coding/Deflate.scala
```diff
@@ -86,74 +90,68 @@ private[http] object DeflateCompressor {
...
      * https://www.ietf.org/rfc/rfc1951.txt (no-wrap)
      */
-    private def examineAndBuildInflater(bytes: ByteString): Inflater =
-      bytes.headOption.filter(b ⇒ (b & 0x0F) != 8)
-        .fold(new Inflater())(_ ⇒ new Inflater(true))
-
-    override def afterInflate = inflateState
-    override def afterBytesRead(buffer: Array[Byte], offset: Int, length: Int): Unit = {}
+    private def examineAndBuildInflater(bytes: ByteString): Inflater = {
+      val wrapped = (bytes.head & 0x0F) == 0x08
```
johanandren: The old one took into account that the `ByteString` could be empty, but now we are sure it isn't?

jrudolph: Yes, exactly. It's an invariant of `ParseStep.parse` that the byte reader will contain at least a single byte to read.

jrudolph:
Added a section to upcoming release notes about the new incompatibility.

jrudolph:
PLS BUILD

tom-walford:
Thanks for getting this all sorted @jrudolph :)

PR #1571, opened 2017-11-28
=htp #1508 cosmetic changes: rename remaining var and move comments

jrudolph:
A small follow up to #1510 renaming other instances of the new flag and moving comments for better readability. Also changed the constants to upper letters according to our standards.

/cc @kennedyoliveira


jrudolph:
Thanks for the review, @johanandren.

PR #1572, opened 2017-11-28
=htc #1492 remove duplicate settings from akka.http.host-connection-pool.client

### LINKED ISSUE #1492: Allow to override "User-Agent" header for client-side with application.conf

jrudolph:
This way the defaults from akka.http.client automatically apply.

Fixes #1492.


jrudolph:
Thanks, @johanandren. Good point about the config checker. See https://github.com/lightbend/akka-commercial-addons/pull/274 for a related fix over there.

PR #1576, opened 2017-11-29
Percent decode userinfo while parsing URIs (#1558)

### LINKED ISSUE #1558: Suprising (additional) percent escaping on Uri#authority rendering

jrudolph:
Follow up to #1559.

The third commit just reenables code formatting for UriParser.scala. Can be ignored for reviewing.

Fixes #1558.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/model/parser/UriParser.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/model/parser/UriParser.scala
@@ -14,65 +14,66 @@ import Parser.DeliveryScheme.Either
 import Uri._
 import akka.annotation.InternalApi
 
-// format: OFF
 /**
  * INTERNAL API
  *
  * http://tools.ietf.org/html/rfc3986
  */
 @InternalApi
-private[http] final class UriParser(val input: ParserInput,
-                              val uriParsingCharset: Charset,
-                              val uriParsingMode: Uri.ParsingMode,
-                              val maxValueStackSize: Int) extends Parser(maxValueStackSize)
+private[http] final class UriParser(
+  val input:             ParserInput,
+  val uriParsingCharset: Charset,
+  val uriParsingMode:    Uri.ParsingMode,
+  val maxValueStackSize: Int) extends Parser(maxValueStackSize)
   with IpAddressParsing with StringBuilding {
   import CharacterClasses._
 
-  def this(input: ParserInput,
-           uriParsingCharset: Charset = UTF8,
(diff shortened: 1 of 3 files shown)


akka-http-core/src/main/scala/akka/http/impl/model/parser/UriParser.scala
```diff
@@ -127,23 +128,23 @@ private[http] final class UriParser(val input: ParserInput,
...
+      | clearSB() ~ ALPHA ~ appendLowered() ~ zeroOrMore(`scheme-char` ~ appendLowered()) ~ &(':') ~ run(_scheme = sb.toString))
 
-  def `scheme-pushed` = rule { oneOrMore(`scheme-char` ~ appendLowered()) ~ run(_scheme = sb.toString) ~ push(_scheme)}
+  def `scheme-pushed` = rule { oneOrMore(`scheme-char` ~ appendLowered()) ~ run(_scheme = sb.toString) ~ push(_scheme) }
 
   def authority = rule { optional(userinfo) ~ hostAndPort }
 
   def userinfo = rule {
-    clearSB() ~ zeroOrMore(`userinfo-char` ~ appendSB()| `pct-encoded`) ~ '@' ~ run(_userinfo = sb.toString)
+    clearSBForDecoding() ~ zeroOrMore(`userinfo-char` ~ appendSB() | `pct-encoded`) ~ '@' ~ run(_userinfo = getDecodedString())
```
ktoso: Oh man, so simple -- I totally missed that, even though that I was staring at getDecodedString quite a bit. Focused too much on it, and not that it wasnt actually called in `userinfo` it seems

jrudolph: Yeah, no worries, it's a bit weird how it is done right here next to parsing but I guess it was done so for performance reasons.

Issue #1550, opened 2017-11-24
Add setting to control parsing of headers


richdougherty:
In Play 2.6 we switched from Netty to Akka HTTP as the default HTTP backend. This has led to a number of bugs because of the different way that Akka HTTP handles headers compared to Netty. Netty headers are raw string whereas Akka HTTP uses type-safe objects for many headers. These type safe objects are parsed from the underlying header values according to the their specification.

Type-safe headers are a good idea in principle but it runs into problems when:

1. Play users need to send/receive non-spec compliant headers (https://github.com/playframework/playframework/issues/7549)
2. Akka HTTP isn't currently spec-compliant (https://github.com/playframework/playframework/pull/7515, https://github.com/playframework/playframework/issues/7997, https://github.com/playframework/playframework/issues/7501)
3. Akka HTTP is spec compliant, but there is more than one way to be spec compliant and Play users need to send/receive headers in a specific way (https://github.com/playframework/playframework/issues/7737, https://github.com/playframework/playframework/issues/7733, https://github.com/akka/akka-http/issues/966, https://github.com/playframework/playframework/pull/7498)

Akka HTTP does make it possible to use `RawHeader`s. Play uses `RawHeader`s for HTTP responses, but we can't use `RawHeader`s when receiving requests, since the header objects are created internally by Akka HTTP inside the header parsing code.

My suggestion is that we add a new setting to Akka HTTP to give Play (and other users) more control over how HTTP headers are parsed. Here are a few ideas for how a setting could work:

1. A `Boolean` setting that switches Akka HTTP into "raw" mode - all headers are parsed as `RawHeader`s.
2. A `String => Boolean` setting that maps header names to a `Boolean` value, giving more fine-grained control over which headers can be parsed into typed headers and which should be parsed as `RawHeader`s.
3. A `HeaderParser => HeaderParser` (e.g. where `type HeaderParser = (String, String) => Either[ErrorResult, HttpHeader]`) setting that allows the user to override the default `HeaderParser` logic and patch their own logic in. This trivially allows replacing all headers with `ReadHeader`s. Users would also be able to continue to use typed headers and even alter the parsing logic for those headers if they need.

jrudolph:
[...]

PR #1577, opened 2017-11-29
+htc #1550 allow disabling of parsing to modeled headers

jrudolph:
Do you think that would suffice for now, @richdougherty?

/cc @richdougherty


raboof:
MiMa:
```
[error]  * abstract method getModeledHeaderParsing()Boolean in class akka.http.javadsl.settings.ParserSettings is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.javadsl.settings.ParserSettings.getModeledHeaderParsing")
[error]  * abstract method modeledHeaderParsing()Boolean in class akka.http.scaladsl.settings.ParserSettings is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.scaladsl.settings.ParserSettings.modeledHeaderParsing")
```

jrudolph:
Yep, already fixed ;)

johanandren:
Refs #1550

PR #1579, opened 2017-11-29
paradox unidoc links

raboof:
For easier linking to javadoc/scaladoc. Rough prototype, might not be
performant and appears to leak memory, needs work to take into account
generics, but might make linking to API docs a lot nicer.


project/ParadoxSignatureDirective.scala
```diff
@@ -13,18 +13,54 @@ import org.pegdown.ast.{DirectiveNode, HtmlBlockNode, VerbatimNode, Visitor}
...
+      val directives = paradoxDirectives.value
+      Def.task { Seq(
         { context: Writer.Context ⇒
-          new SignatureDirective(context.location.tree.label, context.properties, msg ⇒ log.warn(msg))
-        }
-      }
+            new SignatureDirective(context.location.tree.label, context.properties, msg ⇒ log.warn(msg))
+        },
+        { context: Writer.Context ⇒ {
+            val scaladocDirective = directives.map(_.apply(context)).collect { case x: ScaladocDirective => x }.head
```
jrudolph: `collectFirst` ;)

project/ParadoxSignatureDirective.scala
```diff
@@ -13,18 +13,54 @@ import org.pegdown.ast.{DirectiveNode, HtmlBlockNode, VerbatimNode, Visitor}
...
+          }
+        },
+      )}
     }.value
   )
 
+  class UnidocDirective(scaladocDirective: ScaladocDirective, javadocDirective: JavadocDirective, scanner: ScanResult) extends InlineDirective("unidoc") {
+    def render(node: DirectiveNode, visitor: Visitor, printer: Printer): Unit = {
+      val matches = scanner.getNamesOfAllStandardClasses.asScala.filter(_.startsWith("akka.http")).filter(_.endsWith('.' + node.label))
+      if (matches.size > 2)
```
jrudolph: I guess that's because we expect one `scaladsl` and one `javadsl` instance?

raboof: Yes - so indeed this will break in the case where there's 2 matches that are not the `javadsl`/`scaladsl` instances, but that seemed like it might be rare.

jrudolph: I guess it should also check that it matches *something*.

raboof: The next step is matches.foreach, so that's implicit ;). When we figure out how to do this in a way that performs well and does not leak memory we'll also lay it out in an easier to read way :).

jrudolph: But `foreach` doesn't fail, right? I think we should at least warn if no match could be found.

raboof: Aah yes indeed it should!

raboof:
Removed the 'preview' marker: seems to perform acceptably now, no support for generics yet but that can come in a subsequent PR. Useful for #1290.

raboof:
PLS BUILD

build.sbt
```diff
@@ -251,6 +251,12 @@ lazy val docs = project("docs")
         val v = if (isSnapshot.value) "current" else version.value
         s"http://doc.akka.io/japi/akka-http/$v"
       },
+      "javadoc.akka.base_url" -> s"https://doc.akka.io/japi/akka/${Dependencies.akka25Version}",
+      "scaladoc.akka.http.base_url" -> {
+        val v = if (isSnapshot.value) "current" else version.value
+        s"http://doc.akka.io/api/akka-http/$v"
+      },
+      "scaladoc.akka.base_url" -> s"https://doc.akka.io/api/akka/${Dependencies.akka25Version}",
```
jrudolph: Nice, so the APIs are now cross-linked?

PR #1582, opened 2017-11-30
Make akka-actor / akka-stream a provided dependency

jrudolph:
Not sure it even builds like this currently, so lets see where we can get with this.

### DIFF
--- a/docs/src/main/paradox/scala/http/compatibility-guidelines.md
+++ b/docs/src/main/paradox/scala/http/compatibility-guidelines.md
@@ -20,15 +20,13 @@ of your libraries: `Detected java.lang.NoSuchMethodError error, which MAY be cau
 
 ### Akka HTTP 10.0.x with Akka 2.5.x
 
-Akka HTTP 10.0.x is (binary) compatible with *both* Akka `2.4.x` as well as Akka `2.5.x`, however in order to facilitate 
-this the build (and thus released artifacts) depend on the `2.4` series. Depending on how you structure your dependencies,
-you may encounter a situation where you depended on `akka-actor` of the `2.5` series, and you depend on `akka-http`
-from the `10.0` series, which in turn would transitively pull in the `akka-streams` dependency in version `2.4` which 
-breaks the binary compatibility requirement that all Akka modules must be of the same version, so the `akka-streams` 
-dependency MUST be the same version as `akka-actor` (so the exact version from the `2.5` series).
-
-In order to resolve this dependency issue, you must depend on akka-streams explicitly, and make it the same version as
-the rest of your Akka environment, for example like this:
+Akka HTTP 10.0.x is (binary) compatible with *both* Akka `2.4.x` as well as Akka `2.5.x`. However, using Akka HTTP with Akka 2.5 used to be
+a bit confusing, because Akka HTTP explicitly depended on Akka 2.4. Trying to use it together with Akka 2.5,
+running an Akka HTTP application could fail with class loading issues like the above if you forgot to add a dependency to
+both `akka-actor` *and* `akka-stream` of the same version. For that reason, we changed the policy not to depend on `akka-stream`
+explicitly any more but mark it as a `provided` dependency in our build. That means that you will *always* have to add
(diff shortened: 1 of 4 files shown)


raboof:
What would be the reason to do this? To avoid eviction messages when people include both akka 2.4 and 2.5? I guess we want to update the documentation in `introduction.md` to always include the akka dependencies.

jrudolph:
Yes, exactly. Things would likely fail sooner than now where it often only fails at runtime when a newer akka-actor is mixed with the transitively fetched older akka-stream.

jrudolph:
The downside is that you have to add an explicit dependency to akka-stream in any case which makes the simple case of just using akka-http a bit more verbose (but also hopefully trains people to do the right thing).

2m:
I think this makes sense. We already have notes in the docs about adding akka-streams dependency explicitly if you want to use akka-http with newer akka-streams. Therefore in quite a substantial number of projects the build would not change when switching akka-http to provided.

ktoso:
The idea is good. I wonder if it's good to introduce this in 10.0 or just in 10.1 though hm...
I guess it could even be fine for 10.0 series if we put a big section in the release notes about it actually; since in 10.0 this has the largest gain hm. Though definitely good to do in 10.1 anyway, so people realise the version detachments (and if ever 2.6 happens, it's easier to explain to people they can bump the dep).

jrudolph:
Yeah, it shouldn't have any real compatibility issues. In the end, you have to edit your build when you want to upgrade and this time you are required to add another line (which you might quite likely already use anyway).

ktoso:
> In the end, you have to edit your build when you want to upgrade and this time you are required to add another line (which you might quite likely already use anyway).

Yeah, this makes it reasonable to do in the 10.0 line... let's do this and remember to add a big note about it in the annoucement for the next 10.0.xrelease

jrudolph:
Added notes to describe the issue in more detail.

docs/src/main/paradox/scala/http/compatibility-guidelines.md
```diff
@@ -20,15 +20,13 @@ of your libraries: `Detected java.lang.NoSuchMethodError error, which MAY be cau
...
-from the `10.0` series, which in turn would transitively pull in the `akka-streams` dependency in version `2.4` which 
-breaks the binary compatibility requirement that all Akka modules must be of the same version, so the `akka-streams` 
-dependency MUST be the same version as `akka-actor` (so the exact version from the `2.5` series).
-
-In order to resolve this dependency issue, you must depend on akka-streams explicitly, and make it the same version as
-the rest of your Akka environment, for example like this:
+Akka HTTP 10.0.x is (binary) compatible with *both* Akka `2.4.x` as well as Akka `2.5.x`. However, using Akka HTTP with Akka 2.5 used to be
+a bit confusing, because Akka HTTP explicitly depended on Akka 2.4. Trying to use it together with Akka 2.5,
+running an Akka HTTP application could fail with class loading issues like the above if you forgot to add a dependency to
+both `akka-actor` *and* `akka-stream` of the same version. For that reason, we changed the policy not to depend on Akka HTTP
```
raboof: 'not to depend on `akka-stream` explicitly anymore' ?

docs/src/main/paradox/scala/http/release-notes.md
```diff
@@ -1,5 +1,17 @@
 # Release Notes
 
+## Next Version 10.0.12 (or 10.1.0)
+
+### Akka is not an explicit dependency any more
+
+Akka HTTP supports both running on top of Akka 2.4 and Akka 2.5. However, using Akka HTTP with Akka 2.5 used to be
+a bit confusing, because Akka HTTP explicitly depended on Akka 2.4. Trying to use it together with Akka 2.5,
+running an Akka HTTP application could fail with class loading issues if you forgot to add a dependency to
+both `akka-actor` *and* `akka-stream` of the same version. For that reason, we changed the policy not to depend on Akka HTTP
```
raboof: (same here)

docs/src/main/paradox/scala/http/compatibility-guidelines.md
jrudolph: Woops, nicely spotted.

PR #1589, opened 2017-12-02
Update h2pec 1.5.0 -> 2.1.0 (Fix #1578)

### LINKED ISSUE #1578: Update to recent version of h2spec
Our version of h2spec is a bit outdated. The newer version is published in a slightly different format which requires changes to our build.

In #1535, a first attempt was made to update to the new version. It didn't work because we couldn't easily unpack `tar.gz` archives. The new solution should probably just use command line tools like `"tar xfz ...".!!` to extract the binaries.

/cc @golem131

golem131:


build.sbt
```diff
@@ -121,12 +122,20 @@ lazy val http2Support = project("akka-http2-support")
       resourceGenerators in Test += Def.task {
         val log = streams.value.log
         val h2spec = h2specPath.value
-
         if (!h2spec.exists) {
           log.info("Extracting h2spec to " + h2spec)
 
-          for (zip <- (update in Test).value.select(artifact = artifactFilter(name = h2specName, extension = "zip")))
-            IO.unzip(zip, (target in Test).value)
+          for (archive <- (update in Test).value.select(artifact = artifactFilter(name = h2specName, extension = h2specExt))) {
```
jrudolph: Could you add the version number to the path, so that it is redownloaded when we change the version number? I guess we can do that by changing the path like this:

```scala
     lazy val h2specPath = Def.task {
       (target in Test).value / h2specName / Dependencies.h2specVersion / h2specExe
     }
```

Can you try if that works?

project/Dependencies.scala
```diff
@@ -71,7 +76,13 @@ object Dependencies {
 
       // HTTP/2
       val alpnAgent    = "org.mortbay.jetty.alpn"      % "jetty-alpn-agent"             % "2.0.6"            % "test" // ApacheV2
-      val h2spec       = "io.github.summerwind"        % h2specName                     % h2specVersion      % "test" from(h2specUrl) // MIT
+      val h2spec       = "io.github.summerwind"        % h2specName                     % h2specVersion      % "test" artifacts(Artifact(h2specName,
+                                                                                                                                         h2specExt,
```
jrudolph: Extra new lines.

jrudolph:
OK TO TEST

raboof:
Hmm this appears to cause an (empty?) `akka-http2-support/target/test-reports/h2spec-junit.xml` that jenkins appears to dislike, does that make any sense?

jrudolph:
I'd say, empty h2spec-junit.xml would be a bug in this case.

raboof:
PLS BUILD

raboof:
@golem131 would you be interested in rebasing and looing at that reporting issue?

jrudolph:
Seems a bit outdated by now. I'm closing this for now because it seems a bit of work is still left to make it work properly.

Thanks for your work so far, @golem131 and feel to reopen if you'd like to fix the remaining issues!

PR #1596, opened 2017-12-05
=doc Link to https for announcement link in release notes

jlprat:
HTTPS link instead of plain HTTP one.

### DIFF
--- a/docs/src/main/paradox/scala/http/release-notes.md
+++ b/docs/src/main/paradox/scala/http/release-notes.md
@@ -2,7 +2,7 @@
 
 ## 10.0.11
 
-See the [announcement](http://akka.io/blog/news/2017/12/01/akka-http-10.0.11-released.html) and
+See the [announcement](https://akka.io/blog/news/2017/12/01/akka-http-10.0.11-released.html) and
 closed tickets on the [10.0.11 milestone](https://github.com/akka/akka-http/milestone/31?closed=1).
 
 This release adds the long awaited akka-http-caching module inspired by spray-caching.


docs/src/main/paradox/scala/http/release-notes.md
```diff
@@ -2,7 +2,7 @@
 
 ## 10.0.11
 
-See the [announcement](http://akka.io/blog/news/2017/12/01/akka-http-10.0.11-released.html) and
+See the [announcement](http://akka.io/blog/news/2017/12/01/akka-http-10.0.11-released) and
```
raboof: Hmm this URL works for me both with and without `.html`? Would be nice to link to https btw

jlprat: I go to the page and click the link it fails: https://doc.akka.io/docs/akka-http/current/scala/http/release-notes.html

jlprat: Wait, I think I know what happens, @jonas already fixed it, but I didn't realize!

jlprat: I can upgrade the link to https though :)

jrudolph: This link has already been fixed to go to `/blog/news/` instead of `/news`. The `/news` path is old and gets redirected to the new location but that doesn't seem to work for the `.html` version.

jlprat:
So, sorry for the confusion, I updated the PR then. Now it is only about making it go to HTTPS instead.

docs/src/main/paradox/scala/http/release-notes.md
jlprat: I realised that now! sorry

jrudolph:
I also added a redirection at our blog site so that the broken link now works as well. https://github.com/akka/akka.github.com/commit/15eac2a51e84c8f98ff09fe7423e0bf069c97138

jrudolph:
Thanks anyway ;) Since we are often copying around these snippets from the last release to the next when writing the next release notes, it will definitely help to have the right url in there for the next time.

PR #1598, opened 2017-12-05
Add markdown heading.

jiminhsieh:
The original requirement is from https://github.com/akka/akka-http-quickstart-scala.g8/issues/13.
I think it could be useful for external sites to link this table of inspectors directly. Thanks! :)


jlprat:
OK TO TEST

PR #1600, opened 2017-12-06
Consolidate routing-dsl index

raboof:
Refs #1290

Needs a paradox with https://github.com/lightbend/paradox/pull/170 to remove scala-specific elements from
the indexes

### DIFF
--- a/docs/src/main/paradox/scala/http/routing-dsl/index.md
+++ b/docs/src/main/paradox/scala/http/routing-dsl/index.md
@@ -32,7 +32,11 @@ from a background with non-"streaming first" HTTP Servers.
 
 This is a complete, very basic Akka HTTP application relying on the Routing DSL:
 
-@@snip [HttpServerExampleSpec.scala]($test$/scala/docs/http/scaladsl/HttpServerExampleSpec.scala) { #minimal-routing-example }
+Scala
+:  @@snip [HttpServerExampleSpec.scala]($test$/scala/docs/http/scaladsl/HttpServerExampleSpec.scala) { #minimal-routing-example }
+
+Java
+:  @@snip [HttpServerMinimalExampleTest.java]($test$/java/docs/http/javadsl/HttpServerMinimalExampleTest.java) { #minimal-routing-example }
 
 It starts an HTTP Server on localhost and replies to GET requests to `/hello` with a simple response.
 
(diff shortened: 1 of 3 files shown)


raboof:
(CI fails because we haven't released a paradox with https://github.com/lightbend/paradox/pull/170 yet)

docs/src/main/paradox/scala/http/routing-dsl/index.md
```diff
@@ -45,11 +49,16 @@ in the Akka documentation.
...
+Scala
+:  @@snip [HttpAppExampleSpec.scala]($test$/scala/docs/http/scaladsl/HttpAppExampleSpec.scala) { #minimal-routing-example }
+
+Java
+:  @@snip [HttpAppExampleTest.java]($test$/java/docs/http/javadsl/server/HttpAppExampleTest.java) { #minimal-routing-example }
 
 See @ref[HttpApp Bootstrap](HttpApp.md) for more details about setting up a server using this approach.
 
-<a id="long-example"></a>
+@@@ div { .group-scala }
```
jlprat: Pity to have this only for Scala, but this can come in another PR

docs/src/main/paradox/scala/http/routing-dsl/path-matchers.md
```diff
@@ -1,3 +1,10 @@
+@@@ div { .group-java }
+
+This DSL is only relevant when using the Scala API.
```
ktoso: Should we then hide it completely maybe?

docs/src/main/paradox/scala/http/routing-dsl/index.md
raboof: Added #1639

docs/src/main/paradox/scala/http/routing-dsl/path-matchers.md
raboof: The question is what should happen when you are on the Scala version of this page and switch to the Java version.

We could go fancy and use page attributes to mark the whole page as scala-only, and then generate a page that does a redirect (though to where?) for when you hit the the java version (https://github.com/lightbend/paradox/pull/169 might be an inspiration)... but that seems kind of tricky, this should be nice and simple for now :).

jlprat:
@raboof could you resolve the conflicts for this? I'd love to have this merged, as I'm working on the `parameter` directives unification and I'd need to do the same with `case-class-extraction.md`

raboof:
@jlprat I'd be happy to resolve the conflicts, but for this to actually work we'd need a paradox version with lightbend/paradox#170 to be released.. I guess we could temporarily merge without that, but it would show the Scala-only sections in the Java version of the documentation.

jlprat:
I copied your changes for that page to https://github.com/akka/akka-http/pull/1658 so you can wait to merge this one if you want :)

jrudolph:
Great, thanks @raboof for making sure we improve paradox as we go :+1:

PR #1601, opened 2017-12-07
Add non-lambda-capturing getOrLoad method to Cache (#1536)

### LINKED ISSUE #1536: Add caching get(K, V) overload to avoid function allocation
We should add an overload that does not force the function here: 

`def get(key: K, block: () ⇒ V): Future[V] =`

Feedback via @ben-manes

Note that block is not GC friendly, since it requires allocating a capturing lambda for the key. This is a small overhead, but worth being aware of. The alternative, e.g. in Map#computeIfAbsent, is to pass a function k -> v so that the caller can opt into the lamda being cached by the JVM.

https://github.com/akka/akka-http/pull/1503#discussion_r152172802

catalin-ursachi:
(Tentatively) fixes #1536

(cc: @ktoso, @ben-manes)

### DIFF
--- a/akka-http-caching/src/main/mima-filters/10.0.11.backwards.excludes
+++ b/akka-http-caching/src/main/mima-filters/10.0.11.backwards.excludes
@@ -1,0 +1,4 @@
+# Mima filters needed to check newer versions against 10.0.11
+
+# Issue #1536 - Add a new method to Cache; which is annotated with @ApiMayChange.
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.caching.scaladsl.Cache.getOrLoad")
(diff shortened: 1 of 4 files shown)


ktoso:
OK TO TEST

catalin-ursachi:
Could I just check: if the latest released version is 10.0.11, does the `ReversedMissingMethodProblem` filter need to go in `10.0.11.backwards.excludes`, or `10.0.12.backwards.excludes`?

jrudolph:
Into 10.0.11.backwards.excludes. Just added a note about that at the bottom of the paragraph here: https://github.com/akka/akka-http/blob/master/CONTRIBUTING.md#binary-compatibility

PR #1604, opened 2017-12-07
Consolidate the first few directive docs

raboof:
Would like to get feedback on these before we apply the same pattern to the
other directive pages.


raboof:
(👍, I'll see if I can do some of these common cases in one go)

PR #1605, opened 2017-12-07
Further harden and document whitesource

raboof:


.travis.yml
```diff
@@ -34,6 +34,7 @@ deploy:
   provider: script
   script:
     - sbt -J-XX:ReservedCodeCacheSize=256m +publish
+    - sbt 'set credentials += Credentials("whitesource", "whitesourcesoftware.com", "", System.getenv("WHITESOURCE_KEY"))' whitesourceUpdate
```
jrudolph: Is there a reason not to add the credentials line to the build proper (making it a no-op when the key is not defined)?

raboof: I think the 'typical' use is to put the credential in your `~/.sbt/1.0/creds.sbt`, so this is an exceptional case. Might indeed be nice to add something to the build that turns that environment variable into a `Credential` if available, though, indeed

RELEASING.md
```diff
@@ -1,4 +1,5 @@
 ## Releasing
+1. If this is a new minor (not patch) release, rename the 'akka-http-x.x-stable' reporting project in [WhiteSource](http://saas.whitesourcesoftware.com/) accordingly
```
jonas: Use `https://`, at least it redirects to an HTTPS URL for me.

raboof: Oops!

PR #1606, opened 2017-12-07
=pro dependency updates (jackson, specs2, caffeine, gson, macro parad…

jrudolph:
…ise)

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -19,7 +19,8 @@ addSbtPlugin("net.virtual-void" % "sbt-dependency-graph" % "0.9.0") // for advan
 addSbtPlugin("io.spray" % "sbt-boilerplate" % "0.6.1")
 addSbtPlugin("org.foundweekends" % "sbt-bintray" % "0.5.1")
 addSbtPlugin("com.lightbend.sbt" % "sbt-javaagent" % "0.1.4")
-addSbtPlugin("com.lightbend.akka" % "sbt-paradox-akka" % "0.4")
-addSbtPlugin("com.lightbend" % "sbt-whitesource" % "0.1.6")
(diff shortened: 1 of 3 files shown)


jrudolph:
Seems ExpiringLfuCacheSpec fails probably because of the caffeine update.

project/plugins.sbt
```diff
@@ -19,7 +19,8 @@ addSbtPlugin("net.virtual-void" % "sbt-dependency-graph" % "0.9.0") // for advan
 addSbtPlugin("io.spray" % "sbt-boilerplate" % "0.6.1")
 addSbtPlugin("org.foundweekends" % "sbt-bintray" % "0.5.1")
 addSbtPlugin("com.lightbend.sbt" % "sbt-javaagent" % "0.1.4")
-addSbtPlugin("com.lightbend.akka" % "sbt-paradox-akka" % "0.4")
-addSbtPlugin("com.lightbend" % "sbt-whitesource" % "0.1.6")
+addSbtPlugin("com.lightbend.akka" % "sbt-paradox-akka" % "0.5")
```
jrudolph: Will that break Java / Scala switching for now, @raboof?

raboof: Oh, well-spotted - yes it will, 0.5 assumes switching can be done in all-javascript and we're not quite there yet. Probably best to stick to 0.4 until we have completed the consolidation effort

jrudolph: Reverted that particular change.

jlprat:
Now that you are at it, you can also update Akka25 to 2.5.8 and Scala to 2.11.12
Otherwise, LGTM

jrudolph:
Removed the update to Scala 2.11.12 for now since it needs a new genjavadoc release.

PR #1607, opened 2017-12-07
=doc #1603 add note to 10.0.11 release notes about top-level HTTP API changes

### LINKED ISSUE #1603: Binary incompatibility in 10.0.11 vs 10.0.10
Some bridge methods to the old API in HttpExt taking an implicit materializer are private in 10.0.11.
They used to be public API as far as I understand.

* `def cachedHostConnectionPoolHttps[T]` with materializer
* `def cachedHostConnectionPool[T]` with materializer

possibly even more

jrudolph:
Fixes #1603


jrudolph:
PLS BUILD

PR #1608, opened 2017-12-07
Low-hanging doc consolidation for directives

raboof:


jrudolph:
Wow, already great progress :+1:

docs/src/main/paradox/java/http/routing-dsl/directives/basic-directives/extract.md
```diff
@@ -1,5 +1,13 @@
 # extract
 
+@@@ div { .group-scala }
```
jonas: Why put Scala signatures in the Java version?

jonas:
Nice

docs/src/main/paradox/java/http/routing-dsl/directives/basic-directives/extract.md
raboof: Basically my strategy is to make the difference between the Java and Scala versions as small as possible until they are the same and the Java version can be replaced by a symlink

raboof:
(rebased & fixed conflicts - should be nice to merge if green :) )

PR #1616, opened 2017-12-11
Make 'sbt unidoc' compile again

raboof:
I cannot quite explain why 'sbt unidoc' would fail to compile without this
change, but this makes things a little nicer anyway and appears to fix the
problem


ktoso:
LGTM, replicated error and fix. 
Weird but "oh well"

jrudolph:
Seems to compile fine in 2.12 but I guess we can merge this to get rid of the error now or we could just postpone until it is a real problem for anyone (the release is done on 2.12 which seems to be fine right now).

raboof:
Ahh yes..

Issue #1614, opened 2017-12-11
scaladoc published under `japi`


raboof:
Since 10.0.11 it appears we publish the scaladoc under `japi` instead of javadoc - which is bad because it breaks links.

ktoso:
Whoop, that's broken then, needs to be fixed.

PR #1618, opened 2017-12-12
=htc #1614 fix genjavadoc generation

jrudolph:
It seems something prevented merging the "Java API" docs with the deprecation
docs leading to compilation errors during javadoc generation.


ktoso:
resolved #1614

Issue #1610, opened 2017-12-08
Failed: HostConnectionPoolSpec


chbatey:
https://jenkins.akka.io:8498/job/akka-http-nightly-against-akka-master/375/consoleFull

```
[info] HostConnectionPoolSpec:
[info] NewPoolImplementation on PassThrough
[info] - should complete a simple request/response cycle with a strict request and response *** FAILED ***
[info]   java.lang.AssertionError: expected OnNext, found OnComplete
[info]   at akka.stream.testkit.TestSubscriber$ManualProbe.expectNext(StreamTestKit.scala:327)
[info]   at akka.stream.testkit.TestSubscriber$ManualProbe.expectNext(StreamTestKit.scala:316)
[info]   at akka.stream.testkit.TestSubscriber$Probe.requestNext(StreamTestKit.scala:753)
[info]   at akka.http.impl.engine.client.HostConnectionPoolSpec$$anonfun$testSet$1$SetupWithServerProbes$2$ServerConnection.expectRequest(HostConnectionPoolSpec.scala:370)
[info]   at akka.http.impl.engine.client.HostConnectionPoolSpec$$anonfun$testSet$1$$anonfun$apply$mcV$sp$1$$anon$1.<init>(HostConnectionPoolSpec.scala:89)
[info]   at akka.http.impl.engine.client.HostConnectionPoolSpec$$anonfun$testSet$1$$anonfun$apply$mcV$sp$1.apply(HostConnectionPoolSpec.scala:85)
[info]   at akka.http.impl.engine.client.HostConnectionPoolSpec$$anonfun$testSet$1$$anonfun$apply$mcV$sp$1.apply(HostConnectionPoolSpec.scala:85)
[info]   at org.scalatest.OutcomeOf$class.outcomeOf(OutcomeOf.scala:85)
[info]   at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:22)
[info]   ...
[info] - should complete a simple request/response cycle with a chunked request and response *** FAILED ***
```

chbatey:
https://jenkins.akka.io:8498/job/akka-http-nightly-against-akka-master-on-2.12/359//consoleFull

raboof:
Possibly https://jenkins.akka.io:8498/job/akka-http-nightly-against-akka-master-on-2.12/367//consoleFull

jrudolph:
Fixed by #1625.

[...]

PR #1625, opened 2017-12-14
=htc #1610 fix "Cannot push port ... twice" in NewHostConnectionPool

jrudolph:


akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/NewHostConnectionPool.scala
```diff
@@ -301,7 +300,7 @@ private[client] object NewHostConnectionPool {
           }
           def isClosed: Boolean = requestOut.isClosed || responseIn.isClosed
 
-          def onPush(): Unit = {
+          def onPush(): Unit = try {
```
raboof: What does `try` achieve here?

raboof:
Refs #1610

akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/NewHostConnectionPool.scala
```diff
@@ -344,11 +343,23 @@ private[client] object NewHostConnectionPool {
...
             withSlot(_.debug("Connection cancelled"))
 
+          /** Helper that makes sure requestOut is pulled before pushing */
+          private def emitRequest(request: HttpRequest): Unit =
+            if (requestOut.isAvailable) requestOut.push(request)
+            else
+              requestOut.setHandler(new OutHandler {
+                def onPull(): Unit = {
+                  requestOut.push(request)
+                  requestOut.setHandler(connection) // implicit assumption is that connection was the handler before
```
raboof: Perhaps we should make this assumption explicit and fail fast by moving this inner `OutHandler` class to its own type and checking this existing handler is not of that type before doing `requestOut.setHandler(new ...)` above?

akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/NewHostConnectionPool.scala
jrudolph: Oops, looks like a left-over. Weird that it compiles.

akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/NewHostConnectionPool.scala
jrudolph: There's no way to get the current handler for `SubSourceOutlet` (see https://github.com/akka/akka/issues/24158 for other inconsistencies between `SubSourceOutlet` and regular ones). I'll update the comment.

jrudolph:
> Would be nice to have a test that more reliably reproduces this than #1610, but seems reasonable!

It is completely reproducible on Scala 2.12.

raboof:
Needs MiMa exclusions for these internals:
```
[info] akka-http-core: found 2 potential binary incompatibilities while checking against com.typesafe.akka:akka-http-core_2.11:10.0.11 
[error]  * method onShutdown(akka.http.impl.engine.client.pool.SlotContext)Unit in trait akka.http.impl.engine.client.pool.SlotState#BusyState is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.impl.engine.client.pool.SlotState#BusyState.onShutdown")
[error]  * synthetic method akka$http$impl$engine$client$pool$SlotState$BusyState$$super$onShutdown(akka.http.impl.engine.client.pool.SlotContext)Unit in trait akka.http.impl.engine.client.pool.SlotState#BusyState is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.impl.engine.client.pool.SlotState#BusyState.akka$http$impl$engine$client$pool$SlotState$BusyState$$super$onShutdown")
```

jrudolph:
Triggered nightlies to see if it is fixed for good.

PR #1628, opened 2017-12-14
=htc try to keep pool running when a single slot fails

jrudolph:
This is a best effort try not having to kill a pool if a single slot gets
into a weird state. In the best case, that will allow the pool to keep going
in the presence of bugs that are only triggered rarely. On the other hand,
the risk is that we keep a pool alive that might be in a weird state leading
to more problem further down the road.


jrudolph:
PLS BUILD

jrudolph:
This commit was already (accidentally) submitted, reviewed and merged in #1625.

Issue #1624, opened 2017-12-14
Consolidate the remainder of java/scala content in akka-http


richardimaoka:
Continued from #1290 as the conversation of #1290 got too long.

## Description: 

Like we did on akka/akka#23052  "consolidate java/scala content", let's consolidate the java and scala contents in akka-http, too.

Below is the procedure described by @raboof.

The process is as follows:

* Mention in this issue which pages you're planning to merge 
  * Either the Akka maintainers or @richard-imaoka can update the list below
* Compare a file in `docs/src/main/paradox/java/...` with the same one under `docs/src/main/paradox/scala/....` They should cover the same contents, but one is for Java and the other is for Scala. The 'compare files' feature of IntelliJ IDEA is a nice tool for this.
* Update the file in the `scala` folder to contain both the Scala and Java content. You may use:
`@scala` and `@java` directives for inline text
`@@@ div { .group-scala }` for blocks
[code blocks](http://developer.lightbend.com/docs/paradox/latest/features/groups.html) become tabbed
* Replace the 'java' version of the page with a symlink:
  * Remove the old version (`git rm akka-docs/src/main/paradox/java/...`)
  * Create a symlink (`cd akka-docs/src/main/paradox/java; ln -s ../scala/... ...; cd -`)
  * Add the symlink to git (`git add akka-docs/src/main/paradox/java/...`)
Test your changes (`sbt akka-docs/paradox`, see CONTRIBUTING.md for details, check both the Java and Scala versions)

You could look at https://github.com/akka/akka/pull/23039/files as an example - it's probably best to contribute PR's on a page-per-page basis at least at first. Feel free to let us know if you have any questions/problems!

richardimaoka:
Under **docs/src/main/paradox/java/http/**
* [ ] routing-dsl/index.md (@raboof #1600)


Directives:
* [ ] routing-dsl/directives/cookie-directives/index.md (@raboof #1608)
* [ ] routing-dsl/directives/scheme-directives/index.md (@raboof #1608)
* [ ] routing-dsl/directives/cache-condition-directives/index.md (@raboof #1608)
* [ ] routing-dsl/directives/caching-directives/cache.md (@raboof, first #1579)
* [ ] routing-dsl/directives/caching-directives/index.md
* [ ] routing-dsl/directives/form-field-directives/formFieldList.md
* [ ] routing-dsl/directives/form-field-directives/formField.md
[...]

PR #1630, opened 2017-12-14
Remove '-java' postfix from references

raboof:


docs/src/main/paradox/java/http/routing-dsl/directives/security-directives/index.md
```diff
@@ -20,7 +20,7 @@
 
 @@@
 
-<a id="authentication-vs-authorization-java"></a>
+<a id="authentication-vs-authorization"></a>
```
jrudolph: Do we still need those at all?

raboof: only when they don't correspond 1-to-1 with the heading I believe... would be nice to go through those but then I'd like to have https://github.com/lightbend/paradox/issues/120 first

jrudolph: I see, makes sense.

ktoso:
LGTM

ktoso:
#1624

PR #1634, opened 2017-12-14
Consolidate basic directives

raboof:


docs/src/main/paradox/scala/http/routing-dsl/directives/basic-directives/tprovide.md
```diff
@@ -1,3 +1,10 @@
+@@@ div { .group-java }
+
+The `tprovide` directive is not available in the Java API.
```
ktoso: True :)

ktoso:
LGTM, it seems at least ;) Much stuff

PR #1635, opened 2017-12-14
=pro #1633 run class scanner only once per run

### LINKED ISSUE #1633: Unidoc directive is somewhat slow
After I've seen quite some slowdown of paradox I've profiled the reason and it seems that the class scanner of the new unidoc directives is quite slow. I attached the flame graph. Probably, the directive needs to somehow cache the results of a previous run.

[paradox.svg.gz](https://github.com/akka/akka-http/files/1559843/paradox.svg.gz)

jrudolph:
Fixes #1633.


jrudolph:
PLS BUILD

Issue #436, opened 2016-10-26
Improve docs structure for stable release


ktoso:
Tasks:

 * [x] Go through page tree and see if it makes sense and otherwise add tasks here
 * [x] Clean up document structure as shown in the top-level TOC
 * [x] Figure out where to put migration notes per version (release notes or somewhere else?)

jonas:
Paradox uses the title of the top level page as the site title, thus docs/src/main/Paradox/index.md should be titled "Akka HTTP".

jrudolph:
First steps have been done in #1638. I guess we'll keep on the look for further improvements.

PR #1638, opened 2017-12-14
#463 Restructuring of documentation TOC

jrudolph:
Refs #436 

The idea is to start restructuring the top-level TOC to give the most important topics the best spots. The "chapter numbers" are mostly a cosmetic feature that feels to give the docs an ordered appearance (and hopefully prevents adding more entries lightly). I also added some other smaller "features".

screenshot from 2017-12-22 16 07 19

### DIFF
--- a/docs/src/main/paradox/scala/http/routing-dsl/index.md
+++ b/docs/src/main/paradox/scala/http/routing-dsl/index.md
@@ -1,6 +1,6 @@
-# High-level Server-Side API
+# Routing DSL
 
-In addition to the @ref[Low-Level Server-Side API](../server-side/low-level-api.md) Akka HTTP provides a very flexible "Routing DSL" for elegantly
+In addition to the @ref[Core Server API](../server-side/low-level-api.md) Akka HTTP provides a very flexible "Routing DSL" for elegantly
 defining RESTful web services. It picks up where the low-level API leaves off and offers much of the higher-level
 functionality of typical web servers or frameworks, like deconstruction of URIs, content negotiation or
 static content serving.
--- a/docs/src/main/paradox/scala/http/security.md
+++ b/docs/src/main/paradox/scala/http/security.md
@@ -1,4 +1,4 @@
-# Security Announcements
+# ! Security Announcements !
 
 ## Receiving Security Advisories
 The best way to receive any and all security announcements is to subscribe to the [Akka security list](https://groups.google.com/forum/#!forum/akka-security).
--- a/docs/src/main/paradox/scala/http/server-side/low-level-api.md
+++ b/docs/src/main/paradox/scala/http/server-side/low-level-api.md
@@ -1,6 +1,6 @@
-# Low-Level Server-Side API
+# Core Server API
 
-The low-level server-side API is scoped with a clear focus on the essential functionality of an HTTP/1.1 server:
+The core Server API is scoped with a clear focus on the essential functionality of an HTTP/1.1 server:
 
  * Connection management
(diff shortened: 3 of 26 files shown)


docs/src/main/paradox/scala/http/06extensions.md
```diff
@@ -0,0 +1,3 @@
+# 6. Extensions ➕
```
2m: The extensions icon does not render here. Neither Chromium, nor IntelliJ. That is a rare plus you found. :)

Nevertheless, I like the icons. :)

docs/src/main/paradox/scala/http/06extensions.md
```diff
@@ -0,0 +1,3 @@
+# 6. Extensions
+
+This page list well-known third-part extensions of Akka HTTP. See the Akka HTTP tag on scaladex for more info. // TODO: add link
```
ktoso: Link to https://akka.io/community/  

or https://index.scala-lang.org/search?topics=akka-http

ktoso:
Release notes last? Or do something so they don't continue growing into forever in the TOC?

ktoso:
Not sure the utf characters make sense though hm, looks a bit silly

jrudolph:
> Release notes last? Or do something so they don't continue growing into forever in the TOC?

Yep, I guess at some point we should fold them into subpages like "10.0.x release note" etc.

> Not sure the utf characters make sense though hm, looks a bit silly

I removed them again for now. We can still add them later on if we (unlikely enough) happen to agree on something. ;)

ktoso:
> I removed them again for now. We can still add them later on if we (unlikely enough) happen to agree on something. ;) 

ok, sorry ;-)

jonas:
I'm not sure about the Dingbats. It's an extremely limited "icon set" and many of the proposed icons give no clue at all, which IMO makes them pointless. For example, security often uses a shield icon, extensions/plugins a jigsaw puzzle piece.

jrudolph:
@jonas, I see. Let's not go there for now.

docs/src/main/paradox/scala/http/routing-dsl/index.md
```diff
@@ -1,4 +1,4 @@
-# High-level Server-Side API
+# Routing DSL
```
ktoso: Oh yeah great rename

docs/src/main/paradox/scala/http/server-side/low-level-api.md
```diff
@@ -1,6 +1,6 @@
-# Low-Level Server-Side API
+# Core Server API
```
ktoso: Also good rename, I always thought not many people get the "low / high level" wording we used

docs/src/main/paradox/scala/http/security.md
```diff
@@ -1,4 +1,4 @@
-# Security Announcements
+# ! Security Announcements !
```
ktoso: good ;) I can live with that ;)

jrudolph:
A few more changes here:

 * removed the numbering of top-level files
 * copied contributing.md into contributing section, I think it's nice to have that inside the documentation (not sure how to handle that duplication, probably we could link to the documentation from the contributing.md)
 * A bit of page title cleanup (Server-side API => Server API, Low-level Server-Side API => Core Server API, High-level server-side API => Routing DSL)
 * Filled out technologies page
 * Moved some sections under "Tips And Tricks"
 * Reference section that links to reference information at the bottom

jrudolph:
Solved conflicts.

PR #1641, opened 2017-12-15
PathMatcher DSL docs for Java

raboof:


docs/src/main/paradox/scala/http/routing-dsl/path-matchers.md
```diff
@@ -13,29 +13,48 @@ left until, in most cases, it eventually has been consumed completely.
...
-This will match paths like `foo/bar/X42/edit` or `foo/bar/X/create`.
+This will match paths like `foo/bar/X42/edit` or @scala[`foo/bar/X/create`]@java[`foo/bar/X37/create`].
 
 @@@ note
 The path matching DSL describes what paths to accept **after** URL decoding. This is why the path-separating
 slashes have special status and cannot simply be specified as part of a string! The string "foo/bar" would match
 the raw URI path "foo%2Fbar", which is most likely not what you want!
 @@@
 
+@@@ div { .group-scala }
```
jrudolph: Ok, that still means that we're still missing lots of info for Java but adding at least the above part is an improvement I guess ;)

Issue #1645, opened 2017-12-18
Prepare for 10.1.x


jrudolph:
In previous tickets we discussed and planned the upcoming 10.1.0 release:

 * Future Akka HTTP Versioning And Compatibility (#1305)
 * Prepare for Akka 2.5 (#821)
 * Run tests also against Akka 2.5 (#826)
 * Depend on Akka 2.5 directly, and call it 10.1.x (#1151)

This ticket should track what needs to be done concretely in preparation for the release.

 * [x] Branch off release-10.0.x branch
 * [x] Make sure that releasing still works for upcoming 10.0.x maintenance versions (#1729, #1921)
 * [x] Setup nightly builds for 10.0.x against akka 2.4 and 2.5
 * [x] Build against Akka 2.5 by default (= remove support for Akka 2.4) #1652
 * [x] Set new client pool as the default #1657
 * [x] Make sure version considerations and compatibility guarantee docs are up to date (see also #1106)
 * [x] Remove Akka 2.4 references from documentation #1652 #1728
 * [x] Go through deprecations and make them happen.
   * [x] 2.4.x deprecations addressed in #1656 
   * [x] 10.0.1 and 10.0.2 deprecations in #1668
   * [x] < 10.0.11 deprecations in #1689
   *  ~~ 10.0.11 deprecations~~
 * [x] Adapt Java example to Akka 2.5 (`AbstractActor` changes, see #992)

jonas:
@jrudolph Can you clarify what "go through deprecations and make them happen" means?

Does this only apply to deprecations with version before `10.0.0`, for example there are several deprecations from Akka version `2.4.5`.

jonas:
I suggest adding this ticket to the list: https://github.com/akka/akka-http/issues/992

jrudolph:
> Does this only apply to deprecations with version before 10.0.0, for example there are several deprecations from Akka version 2.4.5.

Definitely for the old ones but I would suggest to also remove the newly deprecated methods (even those from 10.0.11). Affected would be users that update to 10.1.0 which use a third-party library that was built against an Akka HTTP version < 10.0.11. I'd rather be a bit optimistic about it and either hope for library authors to push a compatible version if needed or in the unlikely case that the impact is too big, treat it as a regression that we could still fix in 10.1.1 by reintroducing the deprecated methods. WDYT?

jonas:
I only had the guts to remove deprecations from Akka 2.4.x in my "hygge hakking" session.

[...]

PR #1656, opened 2017-12-19
Remove deprecations introduced in 2.4.x

jonas:
Refs #1645


raboof:
(looks like the jenkins failure is MiMa)

akka-http-core/src/main/java/akka/http/javadsl/model/HttpMethod.java
```diff
@@ -42,14 +42,13 @@ public final String name() {
 
     /**
      * Returns the entity acceptance level for this method.
-     * @deprecated Use {@link #getRequestEntityAcceptance} instead, which returns {@link akka.http.javadsl.model.RequestEntityAcceptance}.
      */
-    @Deprecated
-    public abstract akka.http.scaladsl.model.RequestEntityAcceptance requestEntityAcceptance();
+    public abstract akka.http.javadsl.model.RequestEntityAcceptance requestEntityAcceptance();
```
raboof: Subtle - so if you'd want to support both akka-http 10.0 and 10.1 you'd have to use the now-deprecated `getRequestEntityAcceptance` - that makes sense and is in line with the comment on the 'old' `requestEntityAcceptance()`, so 👍

akka-http-core/src/main/java/akka/http/javadsl/model/HttpMethods.java
```diff
@@ -27,18 +27,17 @@ private HttpMethods() {}
 
     /**
      * Create a custom method type.
-     * @deprecated Use {@link  #createCustom} instead.
+     * @deprecated Use {@link #custom} instead. Will be removed in Akka HTTP 11.0.
      */
     @Deprecated
-    public static HttpMethod custom(String value, boolean safe, boolean idempotent, akka.http.scaladsl.model.RequestEntityAcceptance requestEntityAcceptance) {
-        return akka.http.scaladsl.model.HttpMethod.custom(value, safe, idempotent, requestEntityAcceptance);
+    public static HttpMethod createCstom(String value, boolean safe, boolean idempotent, akka.http.javadsl.model.RequestEntityAcceptance requestEntityAcceptance) {
```
raboof: Typo: missing 'u'

akka-http-core/src/main/java/akka/http/javadsl/model/headers/HttpChallenge.java
```diff
@@ -17,22 +17,6 @@
 
     public abstract Map<String, String> getParams();
 
-    /**
-     * @deprecated Use constructor with optional realm parameter instead.
-     */
-    @Deprecated
-    public static HttpChallenge create(String scheme, String realm) {
```
jrudolph: Do we know what's the reasoning behind deprecating those? They seem like useful auxiliary constructors.

akka-http-core/src/test/scala/akka/http/javadsl/JavaInitializationSpec.scala
```diff
@@ -1,66 +0,0 @@
...
-    def =!=(expected: String) = {
-      self should !==(null)
-      self.toString shouldBe expected
-    }
-  }
-
-  "EntityTagRange" should {
-    "initializes the right field" in {
-      akka.http.scaladsl.model.headers.EntityTagRange.`*` =!= "*"
-      akka.http.javadsl.model.headers.EntityTagRanges.ALL =!= "*"
```
jrudolph: Shouldn't we keep testing the constants that still work?

akka-http-core/src/main/java/akka/http/javadsl/model/headers/HttpChallenge.java
ktoso: The idea was to "move to Paths everywhere", but I don't mind keeping them

akka-http-core/src/test/scala/akka/http/javadsl/JavaInitializationSpec.scala
ktoso: This was some crazy edge case so tests best to keep AFAICS

jonas:
I've addressed all comments.

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
```diff
@@ -278,8 +276,7 @@ object HttpEntity {
    *
    * If the given `chunkSize` is -1 the default chunk size is used.
    */
-  @deprecated("Use `fromPath` instead", "2.4.5")
-  def apply(contentType: ContentType, file: File, chunkSize: Int = -1): UniversalEntity =
+  def fromFile(contentType: ContentType, file: File, chunkSize: Int = -1): UniversalEntity =
```
jrudolph: :+1:

akka-http-core/src/test/scala/akka/http/javadsl/JavaInitializationSpec.scala
```diff
@@ -19,46 +19,40 @@ class JavaInitializationSpec extends WordSpec with Matchers {
     "initializes the right field" in {
       akka.http.scaladsl.model.headers.EntityTagRange.`*` =!= "*"
       akka.http.javadsl.model.headers.EntityTagRanges.ALL =!= "*"
-      akka.http.javadsl.model.headers.EntityTagRange.ALL =!= "*"
```
jrudolph: :+1:

Weird that they weren't all in the same order because the bug so much depended on the order of initialization...

akka-http-core/src/main/java/akka/http/javadsl/model/headers/HttpChallenge.java
```diff
@@ -17,20 +17,12 @@
 
     public abstract Map<String, String> getParams();
 
-    /**
-     * @deprecated Use constructor with optional realm parameter instead.
-     */
-    @Deprecated
     public static HttpChallenge create(String scheme, String realm) {
```
jrudolph: :+1:

PR #1659, opened 2017-12-19
=doc Consolidate method directive docs #1624

jlprat:
Refs: #1624
Creates symlink at directory level


jlprat:
The main difference was that `RejectionHandler` wasn't a link for the Java part.

PR #1661, opened 2017-12-19
Run DontLeakActorsOnFailingConnectionSpecs against new client pool + fix unexpected connection signal reception

raboof:
Still does not succeed reliably, but a starting point for further work

### DIFF
--- a/akka-http-tests/src/test/scala/akka/http/scaladsl/server/DontLeakActorsOnFailingConnectionSpecs.scala
+++ b/akka-http-tests/src/test/scala/akka/http/scaladsl/server/DontLeakActorsOnFailingConnectionSpecs.scala
@@ -6,6 +6,12 @@ package akka.http.scaladsl.server
 
 import java.util.concurrent.{ CountDownLatch, TimeUnit }
 
+import com.typesafe.config.ConfigFactory
+
+import scala.concurrent.Await
+import scala.concurrent.duration._
+import scala.util.{ Failure, Success, Try }
+
 import akka.actor.ActorSystem
 import akka.event.Logging
 import akka.http.scaladsl.Http
@@ -14,18 +20,18 @@ import akka.stream.ActorMaterializer
 import akka.stream.scaladsl.{ Sink, Source }
 import akka.stream.testkit.Utils.assertAllStagesStopped
 import akka.testkit.{ TestKit, SocketUtil }
-import com.typesafe.config.ConfigFactory
+
 import org.scalatest.{ BeforeAndAfterAll, Matchers, WordSpecLike }
 
-import scala.util.{ Failure, Success, Try }
+abstract class DontLeakActorsOnFailingConnectionSpecs(poolImplementation: String) extends WordSpecLike with Matchers with BeforeAndAfterAll {
 
-class DontLeakActorsOnFailingConnectionSpecs extends WordSpecLike with Matchers with BeforeAndAfterAll {
-
-  val config = ConfigFactory.parseString("""
+  val config = ConfigFactory.parseString(s"""
     akka {
-      # disable logs (very noisy tests - 100 exepected errors)
+      # disable logs (very noisy tests - 100 expected errors)
       loglevel = OFF
       stdout-loglevel = OFF
+
+      http.host-connection-pool.pool-implementation = $poolImplementation
     }""").withFallback(ConfigFactory.load())
   implicit val system = ActorSystem("DontLeakActorsOnFailingConnectionSpecs", config)
   import system.dispatcher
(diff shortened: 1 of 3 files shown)


jrudolph:
This diff seems to help:

```diff
diff --git a/akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/SlotState.scala b/akka-http-core/s
index 5847b28..cd6e6a7 100644
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/SlotState.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/SlotState.scala
@@ -154,6 +154,12 @@ private[pool] object SlotState {
       ctx.dispatchFailure(ongoingRequest, cause)
       Unconnected
     }
+    override def onConnectionCompleted(ctx: SlotContext): SlotState = {
+      ctx.debug("Connection completed.")
+      ctx.closeConnection()
+      ctx.dispatchFailure(ongoingRequest, new IllegalStateException("Connection was completed when waiting for
+      Unconnected
+    }
   }
 
   case object PreConnecting extends ConnectedState with IdleState with WithRequestDispatching {
@@ -168,6 +174,19 @@ private[pool] object SlotState {
       Unconnected
     }
 
+    override def onConnectionFailed(ctx: SlotContext, cause: Throwable): SlotState = {
+      ctx.debug("Connection failed.")
+      ctx.closeConnection()
+      Unconnected
+    }
+
+    override def onConnectionCompleted(ctx: SlotContext): SlotState = {
+      ctx.debug("Connection completed.")
+      ctx.closeConnection()
+      Unconnected
+    }
+
     override def onNewRequest(ctx: SlotContext, requestContext: RequestContext): SlotState =
       Connecting(requestContext)
   }
```

I guess the `onConnectionCompleted` events are due to the well-known, cancel vs. failure race that can convert a failure to completion :disappointed:

jrudolph:
I think the `temporaryServerHostnameAndPort` approach here is a bit dangerous as well as other tests running in parallel could grab the port and stupid things could happen. Maybe convert it to use a well--enough-known unused host:port like `127.0.0.14:23529`.

jrudolph:
Let's merge this so we can go ahead and make the switch to the new pool afterwards. I'll probably adapt my other new client-pool PR to resolve conflicts and maybe add a few other changes.

akka-http-tests/src/test/scala/akka/http/scaladsl/server/DontLeakActorsOnFailingConnectionSpecs.scala
```diff
@@ -45,7 +45,8 @@ abstract class DontLeakActorsOnFailingConnectionSpecs(poolImplementation: String
       assertAllStagesStopped {
         val reqsCount = 100
         val clientFlow = Http().superPool[Int]()
-        val (host, port) = SocketUtil.temporaryServerHostnameAndPort()
+        val host = "127.0.0.46" // unlikely to be used host / port
```
raboof: This won't work on non-linux machines right? I'm not sure how this is any better compared to `SocketUtil.temporaryServerHostnameAndPort` though?

jrudolph: `SocketUtil.temporaryServerHostnameAndPort` will offer a port that is likely to be used afterwards which will lead to hard to diagnose behavior when other tests are running at the same time.

Maybe there are other addresses that are know to be inaccessible?

jrudolph: Thinking about it, the whole idea of this test is to connect to a port that is *unavailable*, so choosing from the localhost block should still work on systems that don't react on those loopback addresses. I guess the detail error message may be different ("not routable" vs. "connection refused"). Can you try if it's problem for you, @raboof?

jrudolph: Ah, you probably don't got such a machine either... I guess, let's react when someone complains.

raboof: Why would ports returned by `SocketUtil.temporaryServerHostnameAndPort` be likely to be used afterwards? 
```
(for (_ ← Range(0, 10000)) yield {
        SocketUtil.temporaryServerHostnameAndPort()
      }).groupBy(identity _).collect { case (pair, s) if s.size > 1 ⇒ pair }
```
This only turned up a duplicate after a number of attempts. I haven't done the math but that seems fairly plausibly random to me (birthday paradox).

That said, it could be nice to reserve some ports for this kind of use and make sure `temporaryServerHostnameAndPort` stays away from those as well.

jrudolph: > This only turned up a duplicate after a number of attempts. I haven't done the math but that seems fairly plausibly random to me (birthday paradox).

That's good to hear but not really evidence that it would be the same behavior on any other system. We just don't control the allocation procedure.

> That said, it could be nice to reserve some ports for this kind of use and make sure temporaryServerHostnameAndPort stays away from those as well.

But that would require that the OS also takes part in the reservation otherwise it could still offer addresses to other users. By now I think that using `temporaryServerHostnameAndPort` is an unreliable anti-pattern that we should try to avoid (and replace by using `0` when binding servers etc).

jrudolph: Hmm when using an address that isn't known I indeed see a bit weird behavior. It seems the first few connection attempts come back quickly with a "No route to host" error. Afterwards the next connection attempts seem to be stuck a bit longer however eventually running into the connection timeout. I wonder where this behavior comes from.

raboof: I completely agree about avoiding `temporaryServerHostnameAndPort` and just binding to `0` where possible ;)

Issue #1623, opened 2017-12-14
scaladsl.model.HttpResponse.withStatus return javadsl.model.HttpResponse


carstenlenz:
Hi.

the following code snippet will not compile:
```scala
import akka.http.scaladsl.model.{HttpResponse, StatusCodes}

val response: HttpResponse = HttpResponse().withStatus(StatusCodes.NotFound)
```
because the `.withStatus()` method return a `javadsl.model.HttpResponse`.

**Expectation**: is should return `scaladsl.model.HttpResponse` and the code should compile.

Opening this after a short discussion in the Gitter channel by request of @johanandren

jrudolph:
Thanks, we seem to be indeed missing overloads that return the more specific Scala types. Weird, that no one noticed so far.

carstenlenz:
Yes - that is why I thought that there is a surely a reason for having it this way. ;)

jrudolph:
Fixed by #1664.

PR #1664, opened 2017-12-19
=htc Fix return type for `withStatus` methods #1623

jlprat:
Refs: #1623
Changed the return type to the corresponding Scala DSL one

I think this is Binary compatible


jlprat:
I was also surprised, but I trust MiMa. I guess as the return type is a subclass it works at binary level.

jrudolph:
> I was also surprised, but I trust MiMa. I guess as the return type is a subclass it works at binary level.

I think this isn't a feature of the JVM / bytecode but bridges are added by the scala compiler to make it work similar [as Java does](https://stackoverflow.com/questions/18655541/how-covariant-method-overriding-is-implemented-using-bridging-technique-in-java).

jrudolph:
Thanks, @jlprat.

Issue #1644, opened 2017-12-17
HTTPS cached connection pool


rraziel:
I've been using this for a while, to have a flow going towards a host (this is a reverse proxy application), which works fine:
```java
Http.get(actorSystem).cachedHostConnectionPool(someConnectHttp);
```

However when `someConnectHttp` is an HTTPS target, the scheme is ignored.
I noticed there is a `cachedHostConnectionPoolHttps`, but it takes some settings and a logging adapter.

Is there an HTTPS equivalent to the simple `cachedHostConnectionPool`? And if not, shouldn't there be one (or couldn't it at least handle HTTPS targets)?

jrudolph:
Thanks for the report @rraziel. It seems there's indeed an overload missing in the javadsl which just takes a `host` and uses default (Https and other) settings otherwise.

jrudolph:
Fixed by #1665.

ktoso:
Forgot to close the issue?

jrudolph:
Woops, thanks ;)

PR #1665, opened 2017-12-19
Add convenience Java DSL `Http.cachedHostConnectionPoolHttps` #1644 

bpg:


bpg:
The API should be covered with existing documentation, also don't think it requires any specific unit tests.

raboof:
OK TO TEST

akka-http-core/src/main/scala/akka/http/javadsl/Http.scala
```diff
@@ -505,7 +505,7 @@ class Http(system: ExtendedActorSystem) extends akka.actor.Extension {
   /**
    * Same as [[cachedHostConnectionPool]] but with HTTPS encryption.
    *
-   * The given [[ConnectionContext]] will be used for encryption on the connection.
+   * The current default client-side [[ConnectionContext]] will be used for encryption on the connection.
```
jrudolph: That's not quite correct. `ConnectHttp` can contain a custom context which will then be used. Maybe change the comment to

> When a [[HttpConnectionContext]] is defined in [[ConnectHttp]] this will be used, otherwise the default client-side context will be used.

Same below.

jrudolph: I went ahead and applied the suggested doc change.

jrudolph:
PLS BUILD

jrudolph:
Thanks, @bpg!

PR #1668, opened 2017-12-20
!htp Remove deprecated methods for 10.0.1 10.0.2 #1645

jlprat:
Refs: #1645
Methods are mostly fixed typos with 1to1 replacements, or simplified overloads.
Done for versions 10.0.1 and 10.0.2 only.


jlprat:
failure was: https://github.com/akka/akka-http/issues/1669

jlprat:
PLS BUILD

jlprat:
There is another cluster of deprecated methods done on 10.0.6
And then a final batch done on 10.0.11

Issue #1650, opened 2017-12-18
Path matcher DSL docs for Java API


raboof:
docs/src/main/paradox/scala/http/routing-dsl/path-matchers.md contains an introduction for both Scala and Java, but is way more detailed for the Scala API. It would be great if someone could add similar details and examples for the Java API!

jlprat:
I consolidated the first part under #1624
However, the Combinators and Modifiers part is missing. Also a consolidation on wording wouldn't hurt I guess :)

PR #1673, opened 2017-12-20
=doc Consolidate path directives docs #1624

jlprat:
Refs #1624
Converts directoty to symlink
Consolidation of `path-directives.md`, that was only in the Java DSL, with `path-matchers.md`.


docs/src/main/paradox/scala/http/routing-dsl/path-matchers.md
```diff
@@ -135,7 +135,78 @@ PathMatcher[L: Tuple](prefix: Path, extractions: L): PathMatcher[L]
 : Matches and consumes the given path prefix and extracts the given list of extractions.
 If the given prefix is empty the returned matcher matches always and consumes nothing.
 
+@@@
+
+@@@ div { .group-java }
+
+A path matcher is a description of a part of a path to match. The simplest path matcher is `PathMatcher.segment` which
```
jlprat: This section was the one copied over from `path-directives.md`

raboof:
I guess perhaps `path-matchers.md` could now be unified even further, but this is already great!

jlprat:
Agree, `path-matchers.md` can be unified further by including the path combinators (equivalents of `|` and `~`).
I'll create a ticket for it.

jlprat:
Ah, you've done it already: #1650

PR #1674, opened 2017-12-20
=doc Consolidate respondWithHeaders directive docs #1624

jlprat:
Refs #1624
Consolidation of docs
Fix the first snippet that wrong for both Scala and Java


jrudolph:
Thanks, @jlprat.

PR #1679, opened 2017-12-21
Add more Java Api

2m:
These were discovered while working on a play-ws Akka Http backend.

### DIFF
--- a/akka-http-core/src/main/java/akka/http/javadsl/model/Authority.java
+++ b/akka-http-core/src/main/java/akka/http/javadsl/model/Authority.java
@@ -1,0 +1,43 @@
+/**
+ * Copyright (C) 2009-2017 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.http.javadsl.model;
+
+import java.nio.charset.Charset;
+
+/**
+ * Represents a hostname, port and user info.
+ */
+public abstract class Authority {
+
+  /**
+   * Host in a URI
+   */
+  public abstract Host host();
+
+  /**
+   * A port number that may be `0` to signal the default port of for scheme.
+   * In general what you want is not the value of this field but {@link Uri::port::}
+   */
+  public abstract int port();
+
+  /**
+   * The percent decoded userinfo. According to https://tools.ietf.org/html/rfc3986#section-3.2.1
+   * the "user:password" syntax is deprecated and implementations are encouraged to ignore any characters
+   * after the colon (`:`). Therefore, it is not guaranteed that future versions of this class will
+   * preserve full userinfo between parsing and rendering (even if it might do so right now).
+   */
+  public abstract String userinfo();
+
+  /**
+   * Returns a Authority created by parsing the given string representation.
+   */
+  public static Authority create(String authority) {
+    return akka.http.scaladsl.model.Uri.Authority$.MODULE$.parse(
+      akka.parboiled2.ParserInput$.MODULE$.apply(authority),
+      Charset.forName("UTF8"),
+      akka.http.scaladsl.model.Uri$ParsingMode$Relaxed$.MODULE$
+    );
+  }
+}
--- a/akka-http-core/src/main/java/akka/http/javadsl/model/ContentTypes.java
+++ b/akka-http-core/src/main/java/akka/http/javadsl/model/ContentTypes.java
@@ -28,6 +28,13 @@ public final class ContentTypes {
   public static final ContentType.WithCharset TEXT_CSV_UTF8 =
           akka.http.scaladsl.model.ContentTypes.text$divcsv$u0028UTF$minus8$u0029();
 
+  public static final ContentType.Binary NO_CONTENT_TYPE =
+          akka.http.scaladsl.model.ContentTypes.NoContentType();
+
+  public static ContentType parse(String contentType) {
+    return ContentType$.MODULE$.parse(contentType).right().get();
+  }
+
   public static ContentType.Binary create(MediaType.Binary mediaType) {
     return ContentType$.MODULE$.apply((akka.http.scaladsl.model.MediaType.Binary) mediaType);
   }
--- a/akka-http-core/src/main/java/akka/http/javadsl/model/HttpHeader.java
+++ b/akka-http-core/src/main/java/akka/http/javadsl/model/HttpHeader.java
@@ -4,6 +4,8 @@
 
 package akka.http.javadsl.model;
 
+import akka.http.scaladsl.model.IllegalHeaderException;
+
 /**
  * The base type representing Http headers. All actual header values will be instances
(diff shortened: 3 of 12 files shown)


akka-http-core/src/main/java/akka/http/javadsl/model/ContentTypes.java
```diff
@@ -28,6 +28,13 @@ private ContentTypes() { }
   public static final ContentType.WithCharset TEXT_CSV_UTF8 =
           akka.http.scaladsl.model.ContentTypes.text$divcsv$u0028UTF$minus8$u0029();
 
+  public static final ContentType.Binary NO_CONTENT_TYPE =
+          akka.http.scaladsl.model.ContentTypes.NoContentType();
```
jrudolph: Weird indentation but it seems to be consistent with the rest of the file ;)

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
```diff
@@ -172,6 +172,9 @@ sealed trait HttpEntity extends jm.HttpEntity {
   override def toStrict(timeoutMillis: Long, materializer: Materializer): CompletionStage[jm.HttpEntity.Strict] =
     toStrict(timeoutMillis.millis)(materializer).toJava
 
+  /** Java API */
+  override def withContentType(contentType: jm.ContentType): HttpEntity =
+    withContentType(contentType.asInstanceOf[ContentType])
```
jrudolph: We usually use `JavaMappings` and `.asScala`.

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
```diff
@@ -164,6 +166,14 @@ sealed trait HttpMessage extends jm.HttpMessage {
   }
   /** Java API */
   def addHeaders(headers: JIterable[jm.HttpHeader]): Self = mapHeaders(_ ++ headers.asScala.asInstanceOf[Iterable[HttpHeader]])
+  /** Java API */
+  def withHeaders(headers: JIterable[jm.HttpHeader]): Self =
+    withHeaders(headers.asScala.toIndexedSeq.asInstanceOf[immutable.Seq[HttpHeader]])
```
jrudolph: How about `.asScala.toVector`?

akka-http-core/src/main/java/akka/http/javadsl/model/Authority.java
```diff
@@ -0,0 +1,43 @@
...
+   * Returns a Authority created by parsing the given string representation.
+   */
+  public static Authority create(String authority) {
+    return akka.http.scaladsl.model.Uri.Authority$.MODULE$.parse(
+      akka.parboiled2.ParserInput$.MODULE$.apply(authority),
+      Charset.forName("UTF8"),
+      akka.http.scaladsl.model.Uri$ParsingMode$Relaxed$.MODULE$
+    );
+  }
+}
```
ktoso: hah weird that it was missing indeed

akka-http-core/src/main/java/akka/http/javadsl/model/HttpHeader.java
```diff
@@ -44,4 +44,22 @@
...
+    public static HttpHeader parse(String name, String value) {
+      final akka.http.scaladsl.model.HttpHeader.ParsingResult result =
+        akka.http.scaladsl.model.HttpHeader.parse(name, value,
+          akka.http.impl.model.parser.HeaderParser$.MODULE$.DefaultSettings());
+
+      if (result instanceof akka.http.scaladsl.model.HttpHeader$ParsingResult$Ok) {
+        return ((akka.http.scaladsl.model.HttpHeader$ParsingResult$Ok) result).header();
+      }
+      else {
+        throw new IllegalArgumentException("Unable to parse header [" + name + "] with value [" + value + "]");
```
ktoso: I wonder if safe to include the value here hm... If it's some auth header etc?

jrudolph:
Still need mima fixes.

akka-http-core/src/main/java/akka/http/javadsl/model/HttpHeader.java
2m: Good point. Remove header name and value from the exception message.

2m:
PLS BUILD

2m:
Failure was https://github.com/akka/akka-http/issues/1685

jrudolph:
PLS BUILD

jrudolph:
PLS BUILD

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
```diff
@@ -164,6 +166,16 @@ sealed trait HttpMessage extends jm.HttpMessage {
   }
   /** Java API */
   def addHeaders(headers: JIterable[jm.HttpHeader]): Self = mapHeaders(_ ++ headers.asScala.asInstanceOf[Iterable[HttpHeader]])
+  /** Java API */
+  def withHeaders(headers: JIterable[jm.HttpHeader]): Self = {
+    import JavaMapping.Implicits._
+    withHeaders(headers.asScala.toVector.map(_.asScala))
+  }
+  /** Java API */
+  def toStrict(timeoutMillis: Long, ec: Executor, materializer: Materializer): CompletionStage[Self] = {
```
jrudolph: Good that it was as simple as that in the end :)

akka-http-core/src/main/java/akka/http/javadsl/model/HttpHeader.java
```diff
@@ -44,4 +44,22 @@
...
+    public static HttpHeader parse(String name, String value) {
+      final akka.http.scaladsl.model.HttpHeader.ParsingResult result =
+        akka.http.scaladsl.model.HttpHeader.parse(name, value,
+          akka.http.impl.model.parser.HeaderParser$.MODULE$.DefaultSettings());
+
+      if (result instanceof akka.http.scaladsl.model.HttpHeader$ParsingResult$Ok) {
+        return ((akka.http.scaladsl.model.HttpHeader$ParsingResult$Ok) result).header();
+      }
+      else {
+        throw new IllegalArgumentException("Unable to parse header");
```
jrudolph: In the best case, we would also support the ParsingResult in some way on the Java side. If we don't want to do that we could at least throw an `IllegalHeaderException` or `ParsingError` with the `ErrorInfo` in the error. Then the user can decide whether to expose only the `summary` of the error info or the full information.

2m: Alright. Switched to `IllegalHeaderException`

jrudolph: Cool, thanks.

PR #1692, opened 2017-12-22
Consolidate host directives

raboof:


project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=1.0.4
+sbt.version=1.1.0-RC2
```
jrudolph: Sneaky, do we really want that here? ;)

docs/src/main/paradox/java/http/routing-dsl/directives/alphabetically.md
```diff
@@ -37,7 +37,7 @@
...
 |@ref[extractMaterializer](basic-directives/extractMaterializer.md)                                                   | Extracts the @unidoc[Materializer] from the @unidoc[RequestContext]                                                                                                                                                               |
-|@ref[extractHost](host-directives/extractHostName.md)                                                                | Extracts the hostname part of the Host request header value                                                                                                                                                         |
+|@ref[extractHostName](host-directives/extractHost.md)                                                                | Extracts the hostname part of the Host request header value                                                                                                                                                         |
```
jrudolph: There doesn't seem to be any `extractHostName` directive (any more?). I guess we can use `extractHost` now also for Java.

docs/src/main/paradox/scala/http/routing-dsl/directives/host-directives/index.md
```diff
@@ -7,7 +7,7 @@ contained in incoming requests as well as extracting its value for usage in inne
 
 @@@ index
 
-* [extractHost](extractHost.md)
 * [host](host.md)
+* @scala[[extractHost](extractHost.md)]@java[[extractHostName](extractHost.md)]
```
jrudolph: Why change the order?

project/build.properties
ktoso: I think this has a chance to fix the javac NPE issue we've seen. Perhaps as different mini PR though?

raboof: Oops, that wasn't intentional here :)

docs/src/main/paradox/java/http/routing-dsl/directives/alphabetically.md
raboof: ah even better, I didn't realize ;)

PR #1696, opened 2017-12-22
=doc Consolidate security directives docs #1624

jlprat:
Refs #1624
Create symlink for directory
Fixes some Java Signatures
Unifies anchor links


docs/src/main/paradox/scala/http/routing-dsl/directives/security-directives/authorize.md
```diff
@@ -12,8 +12,8 @@ Applies the given authorization check to the request.
 
 ## Description
 
-The user-defined authorization check can either be supplied as a `=> Boolean` value which is calculated
-just from information out of the lexical scope, or as a function `RequestContext => Boolean` which can also
+The user-defined authorization check can either be supplied as a @scala[`=> Boolean`]@java[`() -> Boolean`] value which is calculated
```
jlprat: This could also be written as `Supplier<Boolean>`, personally I like the lambda way...

jrudolph: Do Java programmers understand the lambda syntax? Because I think you cannot really use the types like this in Java code, right?

jlprat: Well, indeed this syntax is not correct. It should rather be `Supplier<Boolean>` but it is used as: `() -> true`. So this syntax is something really in between. I will change it to the official names, so `Supplier<Boolean>`

jlprat: Changed, also for the other one, now is `Function<RequestContext,CompletionStage<Boolean>>`

jrudolph: Cool, I guess it makes sense to follow common notation.

PR #1704, opened 2017-12-22
=doc Consolidate custom-directives docs #1624

jlprat:
Refs: #1624


jlprat:
2 sections are missing on the Java docs:
- Transforming Directives
- Directives from Scratch

First one makes not much sense IMO as there is no support for Java DSL. Second one is also something we don't really support much for the Java DSL.

PR #1705, opened 2017-12-22
Convert old java-only content to symlinks

raboof:
Consolidation was already done but symlink was never added (or perhaps
the old content was brought back by a merge error)


jrudolph:
Nice, I just wondered about that page.

PR #1707, opened 2017-12-22
=doc Consolidate alphabetical ordre and by-trait #1624

jlprat:
Refs: #1624
Alphabetical is not consolidated but just unified


jlprat:
Uses ⇒ instead of => for functions in Scala.
Uses proper names for Java Lambdas (UnaryOperator, Function and so)

PR #1710, opened 2017-12-22
Publish only one documentation tree

raboof:
And switch between java and scala versions in javascript


jrudolph:
Great that we managed to get this in. Can you rebase once again, @raboof?

akka-http-bench-jmh/src/main/scala/akka/http/scaladsl/unmarshalling/sse/LineParserBenchmark.scala
```diff
@@ -22,10 +22,10 @@ class LineParserBenchmark {
 
   // @formatter:off
   @Param(Array(
-    "1024",     // around 1   KB event
```
jrudolph: What happened here? I think I've seen that before?

docs/src/main/paradox/.htaccess
```diff
@@ -10,3 +10,6 @@ RedirectMatch 301 ^(.*)/formFieldList.html$ $1/formFieldSeq.html
 RedirectMatch 301 ^(.*)/extractHostName.html$ $1/extractHost.html
 
 RedirectMatch 301 ^(.*)/release-notes.html$ $1/release-notes/
+
+RedirectMatch 301 ^(.*)/scala/http/(.*)$ $1/$2?language=scala
```
jrudolph: Ooh, really nice, if that works ;)

raboof: looks like: https://doc.akka.io/docs/akka-http/10.0.11+154-91b7071b+20171222-1715/scala/http/routing-dsl/index.html

jrudolph: Very cool, just trying is always the best solution. :+1:

jrudolph:
PLS BUILD

raboof:
PLS BUILD

jrudolph:
We have a liftoff :rocket:

docs/src/main/paradox/client-side/host-level.md
```diff
@@ -191,6 +191,6 @@ the pool will "pull out" single requests when capacity is available on one of th
 @@@
 
 @@@ div { .group-java }
-For now, please see the Scala examples in @ref[Scala Host-Level Client API](../../../scala/http/client-side/host-level.md#examples).
+For now, please see the Scala examples in [Scala Host-Level Client API](../client-side/host-level.html#examples?language=scala).
```
jlprat: This might break the algolia search, AFAIK. We need to reconfigure where to read the language from, right?

raboof: I'd say algolia could be language-agnostic now - but indeed we might have to do something to make them aware of the new locations of things? I'm not sure how that works?

jlprat: There is a JS snippet on the pages initializing the widget. Right now it determines the language by checking `window.location.pathname` if this contains `/java/` or `java.html`.
We need to check `window.location.search` instead and should contain `language=java`.
The other thing though, is how Algolia does the indexing. Because if I'm on the Java page and I search for something, I want to see directly the Java on my result, not the Scala one.

raboof: paradox already remembers the 'current language' in a cookie, so it might not be necessary for algolia to be aware of the language (anymore)?

ktoso: Oh interesting point!

Issue #1691, opened 2017-12-22
Unsupported content type is not logged


oleg-smith:
Unsupported content type should be logged in any case. In logs below it is not clear what was incorrect content type:

`akka.http.scaladsl.unmarshalling.Unmarshaller$UnsupportedContentTypeException: Unsupported Content-Type, supported: application/json
	at akka.http.scaladsl.unmarshalling.Unmarshaller$UnsupportedContentTypeException$.apply(Unmarshaller.scala:158) ~[akka-http_2.11-10.0.11.jar:na]
	at akka.http.scaladsl.unmarshalling.Unmarshaller$EnhancedFromEntityUnmarshaller$$anonfun$forContentTypes$extension$1$$anonfun$apply$24$$anonfun$apply$25.apply(Unmarshaller.scala:114) ~[akka-http_2.11-10.0.11.jar:na]
	at akka.http.scaladsl.unmarshalling.Unmarshaller$EnhancedFromEntityUnmarshaller$$anonfun$forContentTypes$extension$1$$anonfun$apply$24$$anonfun$apply$25.apply(Unmarshaller.scala:111) ~[akka-http_2.11-10.0.11.jar:na]
`

raboof:
Hello Oleg,

I agree it would be useful to log the `Content-Type` that was not supported here, would you be up to preparing a PR?

oleg-smith:
@raboof sure

kell18:
@jrudolph Hi! Any updates here? :)

oleg-smith:
@jrudolph could you merge?

raboof:
The PR currently has some conflicts, do you want to pick those up or should we?

kell18:
@raboof probably better you

jrudolph:
Finally! fixed by #2512.

PR #1713, opened 2017-12-25
Added logging unsupported content type

oleg-smith:


jlprat:
OK TO TEST

oleg-smith:
@jlprat Do you think the binary compatibility problem can be ignored?

jlprat:
Sorry I'm not close to a computer now, I will take a look later, but as a general rule binary compatible problems can only be "ignored" if the class is labeled as `@InternalApi` or `@ApiMayChange`. There are other cases when adding methods though.

jlprat:
Hi @oleg-smith 
Sorry for the delay, but I was away from a computer. IMO this change is binary incompatible the way it is. For example, the Exception `UnsupportedContentTypeException` now has a different constructor and any 3rd party library built upon previous versions of Akka HTTP will fail if we introduce this change.
I would try to implement this change in a compatible way, for example, offering 2 constructors and some default value for `contentType`.

jlprat:
One extra thing, is there an issue for this PR?

oleg-smith:
Yes, #1691

oleg-smith:
Will implement it in compatible way.

jrudolph:
PLS BUILD

oleg-smith:
@jlprat @jrudolph can this incompatibilty be ignored?
`[error]  * the type hierarchy of object akka.http.scaladsl.server.UnsupportedRequestContentTypeRejection is different in current version. Missing types {scala.runtime.AbstractFunction1}
[error]    filter with: ProblemFilters.exclude[MissingTypesProblem]("akka.http.scaladsl.server.UnsupportedRequestContentTypeRejection$")`

oleg-smith:
@jlprat I have finished task.

akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/Unmarshaller.scala
```diff
@@ -154,7 +148,40 @@ object Unmarshaller
         s"Right failure: ${right.getMessage}, " +
         s"Left failure: ${left.getMessage}")
 
+  /**
+   * Signals that unmarshalling failed because the entity content-type did not match one of the supported ranges.
+   * This error cannot be thrown by custom code, you need to use the `forContentTypes` modifier on a base
+   * [[akka.http.scaladsl.unmarshalling.Unmarshaller]] instead.
+   */
+  final case class UnsupportedContentTypeException(
```
raboof: Do we still want this to be a case class?

akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/Unmarshaller.scala
```diff
@@ -154,7 +148,40 @@ object Unmarshaller
...
+    supported:   Set[ContentTypeRange],
+    contentType: Option[ContentType]   = None)
+    extends RuntimeException(supported.mkString(
+      s"Unsupported Content-Type [$contentType], supported: ", ", ", "")) {
+
+    //constructors and copy methods added to cover binary compatibility
+    def this(supported: Set[ContentTypeRange]) = this(supported, None)
```
raboof: do we want to mark these as deprecated as they're only here for binary compat?

akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/Unmarshaller.scala
```diff
@@ -154,7 +148,40 @@ object Unmarshaller
...
         s"Left failure: ${left.getMessage}")
 
+  /**
+   * Signals that unmarshalling failed because the entity content-type did not match one of the supported ranges.
+   * This error cannot be thrown by custom code, you need to use the `forContentTypes` modifier on a base
+   * [[akka.http.scaladsl.unmarshalling.Unmarshaller]] instead.
+   */
+  final case class UnsupportedContentTypeException(
+    supported:   Set[ContentTypeRange],
+    contentType: Option[ContentType]   = None)
```
raboof: Is the only reason to have this as an option binary compatibility? That's kind of unfortunate... perhaps we should consider introducing a singleton `ContentType` to use in that case?

akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/Unmarshaller.scala
oleg-smith: Yes, as moving it to non-case class could cause more binary incompatibility issues

akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/Unmarshaller.scala
oleg-smith: no, the reason is `Content-Type` is not always available

akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/Unmarshaller.scala
oleg-smith: ok

oleg-smith:
@jlprat @raboof can merge?

jlprat:
Hi @oleg-smith!
Thanks for your contribution and welcome to Akka HTTP! In order to be merged, we need first 2 people approving the changes. After this, the PR will be merged.
I'll try to take a look later today to this PR.

jrudolph:
I think we haven't pulled it off so far to extend a case class in such a way. I think we need to be quite sure that we don't introduce binary incompatibilities (though mima says it's fine as far as I can see).

oleg-smith:
@jrudolph what do you mean by that, specifically? How is case class extended except addind case class methods of previous versions?

oleg-smith:
@raboof Do you think it can be merged

akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/Unmarshaller.scala
ktoso: Using an default value is a bad ideas as with those it becomes even harder to keep any bin compat as one later loses the capability to do overloads properly. Please remove the default value, we need to provide methods that achieve the same effect, which I think have been added.

akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/Unmarshaller.scala
```diff
@@ -154,7 +148,46 @@ object Unmarshaller
...
+  }
+
   object UnsupportedContentTypeException {
-    def apply(supported: ContentTypeRange*): UnsupportedContentTypeException = UnsupportedContentTypeException(Set(supported: _*))
+
+    @deprecated("for binary compatibility")
+    def apply(supported: ContentTypeRange*): UnsupportedContentTypeException =
+      new UnsupportedContentTypeException(Set(supported: _*), None)
+
+    @deprecated("for binary compatibility")
```
ktoso: yes, for compatibility but not really deprecated, the apply is perfectly fine as is.

ktoso:
I can try pushing this over the finish line, the binary-compatibility aspect is pretty tricky in this one. You've done a good job about it @oleg-smith but I'm still worried about a few things, the default param and the way we added the field etc.

akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/Unmarshaller.scala
raboof: We would indeed have to add some additional explicit methods (like `equals`), but it would avoid adding yet another automatically-generated `copy` that we'd have to handle when we change this type... OK though.

akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/Unmarshaller.scala
raboof: Perhaps we could make the naming more descriptive here? `actual`?

akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/Unmarshaller.scala
oleg-smith: should i remove? as i was asked to add `deprecated` directly before

lightbend-cla-validator:
Hi @oleg-smith,

akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/Unmarshaller.scala
oleg-smith: @ktoso removed the default value

oleg-smith:
Signed the CLA

oleg-smith:
@raboof @ktoso Does it look ok?

raboof:
(verified you have indeed signed the CLA, not sure why the github bot isn't picking it up yet)

oleg-smith:
@ktoso could you check when you have a chance?

raboof:
PLS BUILD

jlprat:
I'm quite busy at the moment, I'll try to take a look, but don't wait for me :)

oleg-smith:
@raboof Providing the other guys are really busy, can we merge it? :)

akka-http-core/src/test/scala/akka/http/scaladsl/GracefulTerminationSpec.scala
```diff
@@ -122,7 +122,7 @@ class GracefulTerminationSpec extends WordSpec with Matchers with BeforeAndAfter
 
     "in-flight request responses should include Connection: close and connection should be closed" in new TestSetup {
       val r1 = makeRequest() // establish connection
-      val time: FiniteDuration = 3.seconds
+      val time: FiniteDuration = 6.seconds
```
johanandren: This is unrelated, revert.

akka-http/src/main/scala/akka/http/scaladsl/common/StrictForm.scala
```diff
@@ -125,10 +126,10 @@ object StrictForm {
           }
 
         tryUnmarshalToQueryForm.fast.recoverWith {
-          case Unmarshaller.UnsupportedContentTypeException(supported1) ⇒
+          case UnsupportedContentTypeException(supported1, _) ⇒
```
johanandren: if this is required we still have a binary compatibility problem, I think we need the single param unapply or else it could potentially break usage like the old one.

akka-http/src/main/scala/akka/http/scaladsl/server/Rejection.scala
```diff
@@ -111,10 +110,39 @@ final case class InvalidOriginRejection(allowedOrigins: immutable.Seq[SHttpOrigi
...
+final case class UnsupportedRequestContentTypeRejection(
+  supported:   Set[ContentTypeRange],
+  contentType: Option[ContentType]   = None)
+  extends jserver.UnsupportedRequestContentTypeRejection
+  with Rejection {
+
   override def getSupported: java.util.Set[model.ContentTypeRange] =
     scala.collection.mutable.Set(supported.map(_.asJava).toVector: _*).asJava // TODO optimise
+
+  @deprecated("for binary compatibility")
```
johanandren: All these should have a version as well, so `@deprecated("Something something", since = "10.1.6")`, they should also rather mention what to use instead. So, in this case "Use the two parameter constructor instead".

For the copy I guess the alternative now is to create a new object manually, so same there.

akka-http/src/main/scala/akka/http/scaladsl/server/Rejection.scala
```diff
@@ -111,10 +110,39 @@ final case class InvalidOriginRejection(allowedOrigins: immutable.Seq[SHttpOrigi
...
+  def copy$default$1(supported: Set[ContentTypeRange]) =
+    new UnsupportedRequestContentTypeRejection(supported, this.contentType)
+
+  @deprecated("for binary compatibility")
+  def copy(
+    supported:   Set[ContentTypeRange] = this.supported,
+    contentType: Option[ContentType]   = this.contentType) =
+    UnsupportedRequestContentTypeRejection(supported, contentType)
```
johanandren: This is the new two-param copy so it shouldn't be deprecated?

akka-http/src/main/scala/akka/http/javadsl/server/Rejections.scala
```diff
@@ -376,8 +376,18 @@ object Rejections {
...
     s.MalformedHeaderRejection(headerName, errorMsg, cause.asScala)
 
+  def unsupportedRequestContentType(
+    supported:   java.lang.Iterable[MediaType],
+    contentType: Optional[ContentType]): UnsupportedRequestContentTypeRejection =
+    s.UnsupportedRequestContentTypeRejection(
+      supported = supported.asScala.map(m ⇒ scaladsl.model.ContentTypeRange(m.asScala)).toSet,
+      contentType = contentType.asScala.map(_.asScala))
+
+  // for backwards compatibility
```
johanandren: should be marked deprecated as well

akka-http/src/main/scala/akka/http/scaladsl/server/RejectionHandler.scala
```diff
@@ -257,7 +257,8 @@ object RejectionHandler {
       }
       .handleAll[UnsupportedRequestContentTypeRejection] { rejections ⇒
         val supported = rejections.flatMap(_.supported).mkString(" or ")
-        rejectRequestEntityAndComplete((UnsupportedMediaType, "The request's Content-Type is not supported. Expected:\n" + supported))
+        val unsupportedContentType = rejections.find(_.contentType.isDefined).map(_.contentType)
```
johanandren: This will be an `Option[Option[ContentType]]` now. I think you can just do a `rejections.flatMap(_.contentType)` instead.

akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/Unmarshaller.scala
```diff
@@ -47,38 +49,44 @@ object Unmarshaller
...
 
   /**
-   * Helper for creating a "super-unmarshaller" from a sequence of "sub-unmarshallers", which are tried
-   * in the given order. The first successful unmarshalling of a "sub-unmarshallers" is the one produced by the
-   * "super-unmarshaller".
-   */
+    * Helper for creating a "super-unmarshaller" from a sequence of "sub-unmarshallers", which are tried
+    * in the given order. The first successful unmarshalling of a "sub-unmarshallers" is the one produced by the
+    * "super-unmarshaller".
+    */
```
johanandren: This is not the scaladoc style we use, please revert

akka-http-core/src/test/scala/akka/http/scaladsl/GracefulTerminationSpec.scala
oleg-smith: This test fails due to lack of time on Jenkins. How can I make the unrelated test to pass otherwise?

akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/Unmarshaller.scala
oleg-smith: doesn't `validatePullRequest` format code properly?

akka-http-core/src/test/scala/akka/http/scaladsl/GracefulTerminationSpec.scala
johanandren: Then it should fail for all other PR validation and our nightlies on Jenkins as well, which I'm pretty sure it doesn't.

akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/Unmarshaller.scala
johanandren: Could be that it doesn't reformat scaladoc, not sure.

akka-http-core/src/test/scala/akka/http/scaladsl/GracefulTerminationSpec.scala
oleg-smith: No, it should not - it can fail from time to time depending on agent.

oleg-smith:
@raboof yeah, will do

akka-http/src/main/scala/akka/http/scaladsl/common/StrictForm.scala
oleg-smith: isn't binary compatibility checked automatically so nobody guesses about it?

akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/Unmarshaller.scala
raboof: I am also surprised that scalariform doesn't update this indentation, but I just checked and indeed it doesn't.

akka-http/src/main/scala/akka/http/scaladsl/common/StrictForm.scala
raboof: Unfortunately there are some cases that are not yet caught by MiMa, and this is one of them: https://github.com/lightbend/migration-manager/issues/40

raboof:

PR #1732, opened 2018-01-02
=pro disable parallel test execution to make tests more reliable

jrudolph:


jrudolph:
I'm still experimenting with the settings... The ones that I thought should work, don't.

jrudolph:
PLS BUILD ALL

jrudolph:
Approximately doubles the full execution time of all tests. I guess that time is well-spent for the nightlies and may still be good enough for PR validation. Full time for PR validation would afterwards be ~ 8 minutes.

project/ValidatePullRequest.scala
```diff
@@ -84,6 +84,9 @@ object ValidatePullRequest extends AutoPlugin {
   val validatePullRequest = taskKey[Unit]("Validate pull request")
   val additionalTasks = taskKey[Seq[TaskKey[_]]]("Additional tasks for pull request validation")
 
+  // The set of (top-level) files or directories to watch for build changes.
+  val BuildFilesAndDirectories = Set("project", "build.sbt")
```
ktoso: good catch

build.sbt
```diff
@@ -44,7 +44,8 @@ inThisBuild(Def.settings(
   testOptions += Tests.Argument(TestFrameworks.JUnit, "-q", "-v"),
   Dependencies.Versions,
   Formatting.formatSettings,
-  shellPrompt := { s => Project.extract(s).currentProject.id + " > " }
+  shellPrompt := { s => Project.extract(s).currentProject.id + " > " },
+  concurrentRestrictions in Global += Tags.limit(Tags.Test, 1)
```
ktoso: yeah sounds good, until we stabilize things; weird that it started failing so much on our infra recently after the upgrades

jrudolph: No, it runs only on my machine right now. I guess the much bigger amount of tests after the client pool merges changed the execution order in a way that makes it more likely somehow.

PR #1734, opened 2018-01-02
Check the entity body first

raboof:
Perhaps gives a more useful error message for #1063, which I haven't been able
to reproduce locally

### DIFF
--- a/docs/src/test/java/docs/http/javadsl/server/directives/FutureDirectivesExamplesTest.java
+++ b/docs/src/test/java/docs/http/javadsl/server/directives/FutureDirectivesExamplesTest.java
@@ -133,10 +133,10 @@ public class FutureDirectivesExamplesTest extends JUnitRouteTest {
       .assertStatusCode(StatusCodes.InternalServerError())
       .assertEntity("An error occurred: / by zero");
     // opened the circuit-breaker 
-    
+
     testRoute(route).run(HttpRequest.GET("/divide/10/0"))
-          .assertStatusCode(StatusCodes.ServiceUnavailable())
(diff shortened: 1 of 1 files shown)


akka-http-tests/src/test/java/VolatileTest.java
```diff
@@ -0,0 +1,4 @@
+package PACKAGE_NAME;
+
+public class VolatileTest {
```
jrudolph: ?

jrudolph:
Yep, good idea.

akka-http-tests/src/test/java/VolatileTest.java
raboof: woops didn't intend to add that one :)

docs/src/test/scala/docs/http/scaladsl/server/directives/FutureDirectivesExamplesSpec.scala
```diff
@@ -40,7 +40,7 @@ class FutureDirectivesExamplesSpec extends RoutingSpec {
       path("divide" / IntNumber / IntNumber) { (a, b) =>
         onComplete(divide(a, b)) {
           case Success(value) => complete(s"The result was $value")
-          case Failure(ex)    => complete((InternalServerError, s"An error occurred: ${ex.getMessage}"))
+          case Failure(ex)    => complete((InternalServerError, s"An error occurred: $ex"))
```
johanandren: So that will add the exception class in addition to the message?

raboof: exactly

Issue #1724, opened 2018-01-01
java.lang.IllegalStateException: Cannot open connection when slot still has an open connection


abhijith-vr:
Build.sbt

```scala
lazy val akkaHttpVersion = "10.1.0-RC1"
lazy val akkaVersion    = "2.5.8"
```

Code - 

```scala
val QueueSize = 100
  //A create http client pool to core server
  val poolClientFlow = Http().cachedHostConnectionPool[Promise[HttpResponse]](ip, port)

  val queue =
    Source.queue[(HttpRequest, Promise[HttpResponse])](QueueSize, OverflowStrategy.dropNew)
      .via(poolClientFlow)
      .toMat(Sink.foreach({
        case ((Success(resp), p)) => p.success(resp)
        case ((Failure(e), p)) => p.failure(e)
      }))(Keep.left)
      .run()

  def queueRequest(request: HttpRequest): Future[HttpResponse] = {
    val responsePromise = Promise[HttpResponse]()
    queue.offer(request -> responsePromise).flatMap {
      case QueueOfferResult.Enqueued => responsePromise.future
      case QueueOfferResult.Dropped => Future.failed(new RuntimeException("Queue overflowed. Try again later."))
      case QueueOfferResult.Failure(ex) => Future.failed(ex)
      case QueueOfferResult.QueueClosed => Future.failed(new RuntimeException("Queue was closed (pool shut down) while running the request. Try again later."))
    }
  }
```

**Settings**

```
host-connection-pool {
   # The maximum number of parallel connections that a connection pool to a
    # single host endpoint is allowed to establish. Must be greater than zero.
    max-connections = 200

    # The minimum number of parallel connections that a pool should keep alive ("hot").
    # If the number of connections is falling below the given threshold, new ones are being spawned.
    # You can use this setting to build a hot pool of "always on" connections.
    # Default is 0, meaning there might be no active connection at given moment.
    # Keep in mind that `min-connections` should be smaller than `max-connections` or equal
    min-connections = 50

    # The maximum number of times failed requests are attempted again,
    # (if the request can be safely retried) before giving up and returning an error.
    # Set to zero to completely disable request retries.
    max-retries = 5
  
   # The maximum number of open requests accepted into the pool across all
    # materializations of any of its client flows.
    # Protects against (accidentally) overloading a single pool with too many client flow materializations.
    # Note that with N concurrent materializations the max number of open request in the pool
    # will never exceed N * max-connections * pipelining-limit.
[...]

Issue #1721, opened 2017-12-31
FAILED. LegacyPoolImplementation on AkkaHttpEngineNoNetwork should create a new connection when previous one was closed regularly between requests


jlprat:
Encountered in #1719
https://jenkins.akka.io:8498/job/pr-validator-akka-http/2020/testReport/junit/akka.http.impl.engine.client/HostConnectionPoolSpec/LegacyPoolImplementation_on_AkkaHttpEngineNoNetwork_should_create_a_new_connection_when_previous_one_was_closed_regularly_between_requests/

jrudolph:
https://jenkins.akka.io:8498/job/akka-http-nightly-against-akka-2.5/13/testReport/akka.http.impl.engine.client/HostConnectionPoolSpec/LegacyPoolImplementation_on_AkkaHttpEngineNoNetwork_should_create_a_new_connection_when_previous_one_failed_between_requests/

jrudolph:
Probably very similar to #1653

jrudolph:
Test disabled in #1735.

Issue #1653, opened 2017-12-18
HostConnectionPoolSpec.LegacyPoolImplementation on PassThrough should create a new connection when previous one failed between requests


raboof:
https://jenkins.akka.io:8498/job/pr-validator-akka-http/1915/testReport/junit/akka.http.impl.engine.client/HostConnectionPoolSpec/LegacyPoolImplementation_on_PassThrough_should_create_a_new_connection_when_previous_one_failed_between_requests/
```
java.lang.AssertionError: assertion failed: timeout (4500 milliseconds) during expectMsgClass waiting for class akka.http.impl.engine.client.HostConnectionPoolSpec$$anonfun$testSet$1$SetupWithServerProbes$2$ServerConnection
	at scala.Predef$.assert(Predef.scala:170)
	at akka.testkit.TestKitBase$class.expectMsgClass_internal(TestKit.scala:508)
	at akka.testkit.TestKitBase$class.expectMsgType(TestKit.scala:481)
	at akka.testkit.TestKit.expectMsgType(TestKit.scala:882)
	at akka.http.impl.engine.client.HostConnectionPoolSpec$$anonfun$testSet$1$SetupWithServerProbes$2.expectNextConnection(HostConnectionPoolSpec.scala:444)
	at akka.http.impl.engine.client.HostConnectionPoolSpec$$anonfun$testSet$1$$anonfun$apply$mcV$sp$16$$anon$12.<init>(HostConnectionPoolSpec.scala:269)
	at akka.http.impl.engine.client.HostConnectionPoolSpec$$anonfun$testSet$1$$anonfun$apply$mcV$sp$16.apply(HostConnectionPoolSpec.scala:259)
	at akka.http.impl.engine.client.HostConnectionPoolSpec$$anonfun$testSet$1$$anonfun$apply$mcV$sp$16.apply(HostConnectionPoolSpec.scala:259)
```

jrudolph:
On release-10.0 branch:

https://jenkins.akka.io:8498/view/Akka-http/job/akka-http-nightly-release-10.0-against-akka-master/lastCompletedBuild/testReport/akka.http.impl.engine.client/HostConnectionPoolSpec/LegacyPoolImplementation_on_PassThrough_should_create_a_new_connection_when_previous_one_failed_between_requests/

jrudolph:
[...]

PR #1735, opened 2018-01-02
Fix connection closing issue #1724

jrudolph:
Likely the bug happened in the Idle state. That's the one state that
didn't explicitly close the connection when the signal came. On the
other hand, the assumption may have been that on reception of a
`onConnectionCompleted` / `onConnectionFailed` signal, a connection was
already thought to be 'closed'.

This will also make the code simpler because SlotContext.closeConnection()
doesn't need to be exposed to the state handlers.

In some way it also shows the redundancy of the whole state handling code
structure as the code doesn't do much more than showing what expected signals
are in which states. The only side-effects left for the states are pushing
requests and results. But even those are never delayed so there's little logic
left in SlotStates.

Refs #1724, #1653, #1721.

### DIFF
--- a/akka-http-core/src/test/scala/akka/http/impl/engine/client/HostConnectionPoolSpec.scala
+++ b/akka-http-core/src/test/scala/akka/http/impl/engine/client/HostConnectionPoolSpec.scala
@@ -259,12 +259,33 @@ class HostConnectionPoolSpec extends AkkaSpec(
         conn2.pushResponse(HttpResponse(entity = "response"))
         expectResponseEntityAsString() shouldEqual "response"
       }
+      "create a new connection when previous one was closed regularly between requests without sending a `Connection: close` header first" in new SetupWithServerProbes {
+        pushRequest(HttpRequest(uri = "/simple"))
+
+        val conn1 = expectNextConnection()
+        val req = conn1.expectRequest()
+        conn1.pushResponse(HttpResponse(entity = req.uri.path.toString))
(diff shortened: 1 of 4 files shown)


jrudolph:
Added another commit disabling flaky LegacyPoolImplementation tests as reported in #1653 and #1721 so we can focus on the new implementation.

Became a bit of a grab bag but the commits depend on each other.

akka-http-core/src/test/scala/akka/http/impl/engine/client/HostConnectionPoolSpec.scala
```diff
@@ -259,12 +260,35 @@ class HostConnectionPoolSpec extends AkkaSpec(
...
+        conn2.pushResponse(HttpResponse(entity = "response"))
+        expectResponseEntityAsString() shouldEqual "response"
+      }
       "create a new connection when previous one failed between requests" in new SetupWithServerProbes {
+        pendingIn(targetImpl = LegacyPoolImplementation) // flaky test, no reason to debug old client pool issues for now
         pushRequest(HttpRequest(uri = "/simple"))
 
         val conn1 = expectNextConnection()
         val req = conn1.expectRequest()
-        conn1.pushResponse(HttpResponse(headers = headers.Connection("close") :: Nil, entity = req.uri.path.toString))
```
jrudolph: No need to announce that the connection will be closed afterwards, just failing should still work.

PR #1738, opened 2018-01-03
Configure sbt-header and update Lightbend copyright headers

jonas:
Refs #1255

### DIFF
--- a/akka-http-caching/src/main/java/akka/http/caching/javadsl/Cache.java
+++ b/akka-http-caching/src/main/java/akka/http/caching/javadsl/Cache.java
@@ -1,6 +1,7 @@
-/**
- * Copyright (C) 2017 Lightbend Inc. <http://www.lightbend.com/>
+/*
+ * Copyright (C) 2017-2018 Lightbend Inc. <https://www.lightbend.com>
  */
+
 package akka.http.caching.javadsl;
 
 import akka.annotation.ApiMayChange;
--- a/akka-http-core/src/main/scala/akka/http/impl/util/StageLoggingWithOverride.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/util/StageLoggingWithOverride.scala
@@ -1,9 +1,7 @@
-/**
- * Copyright (C) 2017 Lightbend Inc. <http://www.lightbend.com>
- *
- * Copied and adapted from akka-remote
- * https://github.com/akka/akka/blob/c90121485fcfc44a3cee62a0c638e1982d13d812/akka-remote/src/main/scala/akka/remote/artery/StageLogging.scala
+/*
+ * Copyright (C) 2017-2018 Lightbend Inc. <https://www.lightbend.com>
  */
+
 package akka.http.impl.util
 
 import akka.annotation.InternalApi
@@ -17,6 +15,9 @@ import akka.event.NoLogging
 // but we need a few more changes to be able to override the default logger. So for now we keep it here.
 /**
  * INTERNAL API
+ *
+ * Copied and adapted from akka-remote
+ * https://github.com/akka/akka/blob/c90121485fcfc44a3cee62a0c638e1982d13d812/akka-remote/src/main/scala/akka/remote/artery/StageLogging.scala
  */
 @InternalApi
 private[akka] trait StageLoggingWithOverride { self: GraphStageLogic ⇒
--- a/akka-http-core/src/main/scala/akka/http/impl/util/StageLoggingWithOverride.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/util/StageLoggingWithOverride.scala
@@ -1,5 +1,8 @@
 /*
  * Copyright (C) 2017-2018 Lightbend Inc. <https://www.lightbend.com>
+ *
+ * Copied and adapted from akka-remote
+ * https://github.com/akka/akka/blob/c90121485fcfc44a3cee62a0c638e1982d13d812/akka-remote/src/main/scala/akka/remote/artery/StageLogging.scala
  */
 
 package akka.http.impl.util
@@ -15,9 +18,6 @@ import akka.event.NoLogging
 // but we need a few more changes to be able to override the default logger. So for now we keep it here.
 /**
  * INTERNAL API
- *
- * Copied and adapted from akka-remote
- * https://github.com/akka/akka/blob/c90121485fcfc44a3cee62a0c638e1982d13d812/akka-remote/src/main/scala/akka/remote/artery/StageLogging.scala
  */
 @InternalApi
 private[akka] trait StageLoggingWithOverride { self: GraphStageLogic ⇒
--- a/build.sbt
+++ b/build.sbt
@@ -79,7 +79,8 @@ lazy val parsing = project("akka-parsing")
   .settings(
     scalacOptions := scalacOptions.value.filterNot(Set("-Xfatal-warnings", "-Xlint", "-Ywarn-dead-code").contains), // disable warnings for parboiled code
     scalacOptions += "-language:_",
-    unmanagedSourceDirectories in ScalariformKeys.format in Test := (unmanagedSourceDirectories in Test).value
+    unmanagedSourceDirectories in ScalariformKeys.format in Test := (unmanagedSourceDirectories in Test).value,
+    includeFilter in (Compile, headerCreate) := "LogHelper.scala"
   )
   .enablePlugins(ScaladocNoVerificationOfDiagrams)
   .disablePlugins(MimaPlugin)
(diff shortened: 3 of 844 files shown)


project/CopyrightHeader.scala
```diff
@@ -0,0 +1,29 @@
...
+
+  override def requires = HeaderPlugin
+  override def trigger = allRequirements
+
+  override def projectSettings = Seq(Compile, Test).flatMap { config =>
+    inConfig(config)(
+      Seq(
+        includeFilter in headerCreate := (includeFilter in unmanagedSources).value,
+        excludeFilter in headerCreate := (excludeFilter in unmanagedSources).value,
+        unmanagedSources in headerCreate := Defaults.collectFiles(
```
jonas: @hseeberger As mentioned in https://github.com/sbt/sbt-header/pull/147 I could not get `excludeFilter.in(unmanagedSources.in(headerCreate))` to work. It could be the lack of configuration scope however, the [sbt-header code](https://github.com/sbt/sbt-header/blob/fa041e5ca0f5c1ac4ca5c558abfb1384a568b4fc/src/main/scala/de/heikoseeberger/sbtheader/HeaderPlugin.scala#L95) doesn't seem to use `Defaults.collectFiles` or something similar to apply these filters.

project/CopyrightHeader.scala
```diff
@@ -0,0 +1,29 @@
...
+          unmanagedSourceDirectories,
+          includeFilter in headerCreate,
+          excludeFilter in headerCreate
+        ).value,
+        headerLicense := Some(HeaderLicense.Custom(
+          """|Copyright (C) 2009-2018 Lightbend Inc. <https://www.lightbend.com>
```
jonas: Not sure whether to use `http://` or `https://` and ending `/` or not.

project/CopyrightHeader.scala
```diff
@@ -0,0 +1,43 @@
...
+            includeFilter in headerCreate,
+            excludeFilter in headerCreate
+          ).value,
+          unmanagedSources in headerCheck := Defaults.collectFiles(
+            unmanagedSourceDirectories,
+            includeFilter in headerCreate,
+            excludeFilter in headerCreate
+          ).value,
+          headerLicense := Some(HeaderLicense.Custom(
+            """|Copyright (C) 2009-2018 Lightbend Inc. <https://www.lightbend.com>
```
jonas: [rebased so recommenting]

Please confirm whether using `https://` is OK.

project/CopyrightHeader.scala
```diff
@@ -0,0 +1,43 @@
...
+  override def requires = HeaderPlugin
+  override def trigger = allRequirements
+
+  override def projectSettings = Def.settings(
+    Seq(Compile, Test).flatMap { config =>
+      inConfig(config)(
+        Seq(
+          includeFilter in headerCreate := (includeFilter in unmanagedSources).value,
+          excludeFilter in headerCreate := (excludeFilter in unmanagedSources).value,
+          unmanagedSources in headerCreate := Defaults.collectFiles(
```
jonas: @hseeberger (Sorry to ping you twice) As mentioned in https://github.com/sbt/sbt-header/pull/147 I could not get `excludeFilter.in(unmanagedSources.in(headerCreate))` to work. It could be the lack of configuration scope however, the [sbt-header code](https://github.com/sbt/sbt-header/blob/fa041e5ca0f5c1ac4ca5c558abfb1384a568b4fc/src/main/scala/de/heikoseeberger/sbtheader/HeaderPlugin.scala#L95) doesn't seem to use `Defaults.collectFiles` or something similar to apply these filters. Is this a bug?

build.sbt
```diff
@@ -57,6 +57,7 @@ lazy val root = Project(
   .settings(
     // Unidoc doesn't like macros
     unidocProjectExcludes := Seq(parsing),
+    unmanagedSources in (Compile, headerCreate) := (baseDirectory.value / "project").**("*.scala").get,
```
jonas: Manage headers as part of the root project.

project/CopyrightHeader.scala
hseeberger: Probably a Bug/missing feature.

project/CopyrightHeader.scala
raboof: Yes, please! :)

raboof:
Hmm I wonder if we want to update all years, or try to preserve them (using https://github.com/sbt/sbt-header#custom-comment-creators)

akka-http-caching/src/main/java/akka/http/caching/javadsl/Cache.java
```diff
@@ -1,6 +1,7 @@
-/**
- * Copyright (C) 2017 Lightbend Inc. <http://www.lightbend.com/>
+/*
+ * Copyright (C) 2009-2018 Lightbend Inc. <https://www.lightbend.com>
```
jlprat: AFAIU the year range was meant to keep track of when the file was created. For example, `Cache` was created on 2017, so this year should be 2017-2018. Is this something legal? Or just a good thing to have?

ktoso: A nice to have AFAIR. We can use the automatic template hm

jrudolph: > AFAIU the year range was meant to keep track of when the file was created. For example, Cache was created on 2017, so this year should be 2017-2018. Is this something legal? Or just a good thing to have?

That's one way to use/interpret them but it seems that Akka uses the convention to use the same headers everywhere (e.g. https://github.com/akka/akka/commit/6c7085252a4bb8e5761af3a2c08e7f006e657c8a where everything was changed to 2017). It would make sense to keep it persistent between Akka  sub-projects. When it comes to history, there's always the commit history as the ultimate truth.

build.sbt
```diff
@@ -91,6 +93,11 @@ lazy val httpCore = project("akka-http-core")
   .settings(Dependencies.httpCore)
   .settings(VersionGenerator.versionSettings)
   .enablePlugins(BootstrapGenjavadoc)
+  .settings(
+    excludeFilter in (Compile, headerCreate) :=
+      "LastEventId.*" || "ServerSentEvent.*" || "Base64Parsing.scala" || "StringBuilding.scala",
```
jrudolph: Hmm, that seems quite brittle to me. The plugin shouldn't touch any files that are not yet copyright Lightbend. Can we do that somehow?

jonas: It looks like we can match on the previous header text and update accordingly. Will give it a try.

jonas: Access to existing headers was added by @raboof in https://github.com/sbt/sbt-header/pull/142 and has not yet been released. Since it simplifies a lot I suggest we wait until a new release is out.

jonas: sbt-header 4.1.0 has been released. I'll update the PR to get rid of the excludes.

jonas:
The years are now preserved. Additional text in the Lightbend copyright headers are not preserved however. There was one such header.

This also detected one Typesafe copyright header and several missing headers (mainly in `*Spec` files).

jonas:
PLS BUILD

akka-http-core/src/main/scala/akka/http/impl/util/StageLoggingWithOverride.scala
```diff
@@ -1,9 +1,7 @@
-/**
- * Copyright (C) 2017 Lightbend Inc. <http://www.lightbend.com>
- *
- * Copied and adapted from akka-remote
```
jonas: Additional text is not carried over. Not sure if that would be a requirement.

jrudolph: Would be nice if that could be prevented. The problem is that the commit which changes all the headers is so big that it's likely that changes like this would be lost forever. As they are likely copyright-related, maybe we shouldn't take chances there? Could we use a pattern like

```
"(?s).* Copyright \\([Cc]\\) (\\d{4}(-\\d{4})?) (Lightbend|Typesafe) Inc[^\n]*\n((?:\\s|\\*|\\*/)*)".r
```

to prevent those headers to be changed? Or to make sure that it doesn't touch the existing text?

jonas:
Yes, I agree. Will try to get it done in the next day or two.

jonas:
Only Lightbend copyright notices are touched as of the latest version.

jrudolph:
Great stuff, thanks a lot, @jonas. It will help a lot in upcoming years!

ktoso:
Late to the party but this is fantastic, thanks a lot :)

PR #1740, opened 2018-01-03
Circuitbreaker directive test failures fix

### LINKED ISSUE #1419: FAILED: scaladsl FutureDirectivesSpec onCompleteWithBreaker
akka.http.scaladsl.server.directives.FutureDirectivesSpec.The `onCompleteWithBreaker` directive should stop failing fast when the circuit breaker closes

This seems to be different from #1063 which is for the javadsl.

```
Error Message

Request was rejected with rejection CircuitBreakerOpenRejection(akka.pattern.CircuitBreakerOpenException: Circuit Breaker is open; calls are failing fast)

Stacktrace

org.scalatest.exceptions.TestFailedException: Request was rejected with rejection CircuitBreakerOpenRejection(akka.pattern.CircuitBreakerOpenException: Circuit Breaker is open; calls are failing fast)
	at akka.http.scaladsl.testkit.TestFrameworkInterface$Scalatest$class.failTest(TestFrameworkInterface.scala:24)
	at akka.http.scaladsl.server.RoutingSpec.failTest(RoutingSpec.scala:19)
	at akka.http.scaladsl.testkit.RouteTestResultComponent$RouteTestResult.rawResponse(RouteTestResultComponent.scala:56)
	at akka.http.scaladsl.testkit.RouteTestResultComponent$RouteTestResult.response(RouteTestResultComponent.scala:39)
	at akka.http.scaladsl.testkit.RouteTest$class.response(RouteTest.scala:62)
	at akka.http.scaladsl.server.RoutingSpec.response(RoutingSpec.scala:19)
	at akka.http.scaladsl.testkit.RouteTest$class.responseAs(RouteTest.scala:71)
	at akka.http.scaladsl.server.RoutingSpec.responseAs(RoutingSpec.scala:19)
	at akka.http.scaladsl.server.directives.FutureDirectivesSpec$$anonfun$2$$anonfun$apply$mcV$sp$9$$anon$4$$anonfun$15.apply(FutureDirectivesSpec.scala:93)
	at akka.http.scaladsl.server.directives.FutureDirectivesSpec$$anonfun$2$$anonfun$apply$mcV$sp$9$$anon$4$$anonfun$15.apply(FutureDirectivesSpec.scala:93)
	at scala.util.DynamicVariable.withValue(DynamicVariable.scala:58)
	at akka.http.scaladsl.testkit.RouteTest$$anonfun$check$1.apply(RouteTest.scala:57)
	at akka.http.scaladsl.testkit.RouteTest$$anonfun$check$1.apply(RouteTest.scala:57)
	at akka.http.scaladsl.testkit.RouteTestResultComponent$RouteTestResult.$tilde$greater(RouteTestResultComponent.scala:50)
	at akka.http.scaladsl.server.directives.FutureDirectivesSpec$$anonfun$2$$anonfun$apply$mcV$sp$9$$anon$4.<init>(FutureDirectivesSpec.scala:92)
	at akka.http.scaladsl.server.directives.FutureDirectivesSpec$$anonfun$2$$anonfun$apply$mcV$sp$9.apply(FutureDirectivesSpec.scala:89)
	at akka.http.scaladsl.server.directives.FutureDirectivesSpec$$anonfun$2$$anonfun$apply$mcV$sp$9.apply(FutureDirectivesSpec.scala:89)
	at org.scalatest.OutcomeOf$class.outcomeOf(OutcomeOf.scala:85)
```

johanandren:
Fixes #1419 and #1063

### DIFF
--- a/akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/FutureDirectivesSpec.scala
+++ b/akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/FutureDirectivesSpec.scala
@@ -86,19 +86,18 @@ class FutureDirectivesSpec extends RoutingSpec with Inside with TestKitBase {
           inside(rejection) {
             case CircuitBreakerOpenRejection(_) ⇒
           }
-        }, 500.millis)
+        }, breakerResetTimeout / 2)
     }
     "stop failing fast when the circuit breaker closes" in new TestWithCircuitBreaker {
       openBreaker()
       // observe that it opened
       awaitAssert(breaker.isOpen should ===(true))
-      Thread.sleep(breakerResetTimeout.toMillis)
       // since this is timing sensitive, try a few times to observe the breaker closed
       awaitAssert({
         Get() ~> onCompleteWithBreaker(breaker)(Future.successful(1)) { echoComplete } ~> check {
           responseAs[String] shouldEqual "Success(1)"
         }
-      }, 500.millis)
+      }, breakerResetTimeout + 1.second)
     }
     "catch an exception in the success case" in new TestWithCircuitBreaker {
       Get() ~> onCompleteWithBreaker(breaker)(Future.successful("ok")) { throwTestException("EX when ") } ~> check {
--- a/akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/FutureDirectivesSpec.scala
+++ b/akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/FutureDirectivesSpec.scala
@@ -14,7 +14,7 @@ import org.scalatest.Inside
 
 import scala.concurrent.duration._
 
-class FutureDirectivesSpec extends RoutingSpec with Inside {
+class FutureDirectivesSpec extends RoutingSpec with Inside with TestKitBase {
 
   class TestException(msg: String) extends Exception(msg)
   object TestException extends Exception("XXX")
@@ -80,18 +80,25 @@ class FutureDirectivesSpec extends RoutingSpec with Inside {
     }
     "fail fast if the circuit breaker is open" in new TestWithCircuitBreaker {
       openBreaker()
-      Get() ~> onCompleteWithBreaker(breaker)(Future.successful(1)) { echoComplete } ~> check {
-        inside(rejection) {
-          case CircuitBreakerOpenRejection(_) ⇒
-        }
-      }
+      // since this is timing sensitive, try a few times to observe the breaker open
+      awaitAssert(
+        Get() ~> onCompleteWithBreaker(breaker)(Future.successful(1)) { echoComplete } ~> check {
+          inside(rejection) {
+            case CircuitBreakerOpenRejection(_) ⇒
+          }
+        }, 500.millis)
     }
     "stop failing fast when the circuit breaker closes" in new TestWithCircuitBreaker {
       openBreaker()
-      Thread.sleep((breakerResetTimeout + 200.millis.dilated).toMillis)
-      Get() ~> onCompleteWithBreaker(breaker)(Future.successful(1)) { echoComplete } ~> check {
-        responseAs[String] shouldEqual "Success(1)"
-      }
+      // observe that it opened
+      awaitAssert(breaker.isOpen should ===(true))
+      Thread.sleep(breakerResetTimeout.toMillis)
+      // since this is timing sensitive, try a few times to observe the breaker closed
+      awaitAssert({
+        Get() ~> onCompleteWithBreaker(breaker)(Future.successful(1)) { echoComplete } ~> check {
+          responseAs[String] shouldEqual "Success(1)"
+        }
+      }, 500.millis)
     }
     "catch an exception in the success case" in new TestWithCircuitBreaker {
       Get() ~> onCompleteWithBreaker(breaker)(Future.successful("ok")) { throwTestException("EX when ") } ~> check {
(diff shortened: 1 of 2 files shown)


docs/src/test/java/docs/http/javadsl/server/directives/FutureDirectivesExamplesTest.java
```diff
@@ -138,12 +139,21 @@ public void testOnCompleteWithBreaker() throws InterruptedException {
...
+      {
+        awaitAssert(
+            FiniteDuration.create(300, TimeUnit.MILLISECONDS),
+            () -> {
+              testRoute(route).run(HttpRequest.GET("/divide/8/2"))
+                  .assertEntity("The result was 4");
+              return null;
+            });
+      }
+    };
```
johanandren: A bit messy in a code sample but I guess that's life in Java-land.

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/FutureDirectivesSpec.scala
```diff
@@ -80,18 +80,25 @@ class FutureDirectivesSpec extends RoutingSpec with Inside {
...
-          case CircuitBreakerOpenRejection(_) ⇒
-        }
-      }
+      // since this is timing sensitive, try a few times to observe the breaker open
+      awaitAssert(
+        Get() ~> onCompleteWithBreaker(breaker)(Future.successful(1)) { echoComplete } ~> check {
+          inside(rejection) {
+            case CircuitBreakerOpenRejection(_) ⇒
+          }
+        }, 500.millis)
```
jrudolph: I think that can be even longer. Everything that is still significant shorter than the `breakerResetTimeout` would do. Maybe half the `breakerResetTimeout`?

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/FutureDirectivesSpec.scala
```diff
@@ -80,18 +80,25 @@ class FutureDirectivesSpec extends RoutingSpec with Inside {
...
-      }
+      // observe that it opened
+      awaitAssert(breaker.isOpen should ===(true))
+      Thread.sleep(breakerResetTimeout.toMillis)
+      // since this is timing sensitive, try a few times to observe the breaker closed
+      awaitAssert({
+        Get() ~> onCompleteWithBreaker(breaker)(Future.successful(1)) { echoComplete } ~> check {
+          responseAs[String] shouldEqual "Success(1)"
+        }
+      }, 500.millis)
```
jrudolph: Do we even need the sleep still? We could just succeed once the breaker is reset by increasing the timeout here to `breakerResetTimeout` plus a bit.

docs/src/test/java/docs/http/javadsl/server/directives/FutureDirectivesExamplesTest.java
jrudolph: Jep, unfortunate that testkit is not available and not easily mixed in Java.

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/FutureDirectivesSpec.scala
johanandren: Good idea

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/FutureDirectivesSpec.scala
johanandren: Yeah, I just thought it wasn't very useful to hit it while we know it will fail, either way is fine, I'll change.

johanandren: One thing that may be a bit surprising with that is that the `breakerResetTimeout` is already dilated, and then `assertAwait` further dilates the max time.

jrudolph:
I wonder if it would make sense only testing the state of the circuit breaker inside of the awaitAsserts instead of running whole requests. That would still test the routing layer sufficiently imo.

johanandren:
I did that first, but then thought that it would be more realistic if the actual directive logic would be what triggers the circuit breaker to open.

jrudolph:
> I did that first, but then thought that it would be more realistic if the actual directive logic would be what triggers the circuit breaker to close.

That yes, but one request should be enough to get there. Afterwards we just need to wait until the breaker has registered the failure.

johanandren:
Now you lost me. We open it, verify that it is open in an `awaitAssert`, then wait for the close-timeout, trigger a request that when it succeeds closes the cb again. You say we don't need to do the request to trigger the close?

jrudolph:
> Now you lost me. We open it, verify that it is open in an awaitAssert, then wait for the close-timeout, trigger a request that when it succeeds closes the cb again. You say we don't need to do the request to trigger the close?

Hmm, I didn't think too deeply about how the breaker actually works. Is a request necessary to move it back from open to half-closed? Or will that change based on the timeout alone?

johanandren:
When the timeout hits it moves to half-open, the first request coming through is let through without failing fast (others are failed fast still), when that first request succeeds it toggles the breaker back to closed. 

I guess we could assert that it becomes half-open but I still like observing it from the outside better since it is more black-boxey.

jrudolph:
I see.

jrudolph:
Thanks, @johanandren. Glad to see that one finally gone for good.

PR #1744, opened 2018-01-03
Setup travis for snapshot publishing

### LINKED ISSUE #1254: Publish nightly snapshots
Nightly snapshots are good to have in general but especially for akka-http because we don't do any release candidates, so our releases are only tested by the project's automated tests and enthusiast that build from source.

Publishing the snapshots would help there, because we could people instantly point to a nightly with a fix which might help with catching regressions early on.

2m:
We will need to set up a [cron job](https://docs.travis-ci.com/user/cron-jobs/) on the master branch to start publishing snapshots. The artifacts will end up [here](https://bintray.com/akka/snapshots/com.typesafe.akka%3Aakka-http_2.11) or under a different view [here](https://dl.bintray.com/akka/snapshots/com/typesafe/akka/).

They will be possible to be consumed by users by adding
```
resolvers += Resolver.bintrayRepo("akka", "snapshots")
```
to the build.

I also cleaned up `.travis.yml` a bit and moved the nested bash conditional logic to a separete sbt command for whitesource.

If this looks good for an approach, I will ammend documentation.

Fixes #1254

### DIFF
--- a/project/Publish.scala
+++ b/project/Publish.scala
@@ -21,12 +21,17 @@ object Publish extends AutoPlugin {
   import bintray.BintrayPlugin
   import bintray.BintrayPlugin.autoImport._
 
+  object CliOptions {
+    val publishRepo = CliOption("akka.publish.repo", "maven")
+  }
+
   override def trigger = allRequirements
   override def requires = BintrayPlugin
 
   override def projectSettings = Seq(
     bintrayOrganization := Some("akka"),
-    bintrayPackage := "com.typesafe.akka:akka-http_2.11"
+    bintrayPackage := "com.typesafe.akka:akka-http_2.11",
+    bintrayRepository := CliOptions.publishRepo.get
   )
 }
 
--- a/travis.sbt
+++ b/travis.sbt
@@ -1,0 +1,32 @@
+commands ++= Seq(
+  // check policies, if on master also upload
+  Command.command("whitesourceCi") { state =>
+    val extracted = Project.extract(state)
+    import extracted._
+
+    if (sys.env.getOrElse("TRAVIS_SECURE_ENV_VARS", "") == "true") {
+      val stateWithCredentials =
+        append(credentials += Credentials("whitesource", "whitesourcesoftware.com", "", System.getenv("WHITESOURCE_KEY")), state)
+
+      runTask(whitesourceCheckPolicies, stateWithCredentials)
+
+      if (sys.env.getOrElse("TRAVIS_BRANCH", "") == "master" && sys.env.getOrElse("TRAVIS_EVENT_TYPE", "") == "push") {
+        runTask(whitesourceUpdate, stateWithCredentials)
+      }
+    }
+
+    state
+  },
+
+  // publish snapshot to bintray snapshots repo
+  Command.command("publishCi") { state =>
+    val extracted = Project.extract(state)
+    import extracted._
+
+    if (sys.env.getOrElse("TRAVIS_EVENT_TYPE", "") == "cron") {
+      runAggregated(ThisBuild / publish, state)
+    }
+
+    state
+  }
+)
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,18 +23,15 @@ cache:
     - $HOME/.sbt/boot
 
 script:
-  # need to override as the default is to test
-  - sbt -J-XX:ReservedCodeCacheSize=256m +update
   # make 'git branch' work again
   - git branch -f "$TRAVIS_BRANCH" && git checkout "$TRAVIS_BRANCH"
-  # check policies, if on master also upload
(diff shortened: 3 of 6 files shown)


travis.sbt
```diff
@@ -0,0 +1,32 @@
...
+    state
+  },
+
+  // publish snapshot to bintray snapshots repo
+  Command.command("publishCi") { state =>
+    val extracted = Project.extract(state)
+    import extracted._
+
+    if (sys.env.getOrElse("TRAVIS_EVENT_TYPE", "") == "cron") {
+      runAggregated(ThisBuild / publish, state)
```
jonas: Very nice. Does this do the same as `+publish`?

travis.sbt
```diff
@@ -0,0 +1,32 @@
+commands ++= Seq(
```
jrudolph: Why a command and no task? Wouldn't a task be easier to use and maintain?

travis.sbt
```diff
@@ -0,0 +1,32 @@
...
+    import extracted._
+
+    if (sys.env.getOrElse("TRAVIS_SECURE_ENV_VARS", "") == "true") {
+      val stateWithCredentials =
+        append(credentials += Credentials("whitesource", "whitesourcesoftware.com", "", System.getenv("WHITESOURCE_KEY")), state)
+
+      runTask(whitesourceCheckPolicies, stateWithCredentials)
+
+      if (sys.env.getOrElse("TRAVIS_BRANCH", "") == "master" && sys.env.getOrElse("TRAVIS_EVENT_TYPE", "") == "push") {
+        runTask(whitesourceUpdate, stateWithCredentials)
```
jrudolph: Shouldn't that run only on releases?

travis.sbt
2m: At first, I had state changes in both of the commands (adding credentials and changing repo name), therefore I went with a command. Now only the first command does that, therefore the second could be a task. However, having conditional statements would require it to be a DynTask. So I stuck with commands.

travis.sbt
2m: I am not sure what the policy is. @raboof?

travis.sbt
2m: `runAggregated` runs across all subprojects. The cross `+` is still in the `.travis.yml` as `+publishCi`.

travis.sbt
jrudolph: What's wrong with dyntask?

2m: Nothing in particular. It would work for the `publishCi` but not for the `whitesourceCi` which changes the state.

jrudolph: But why does it need to update the state? Why not just add the credentials anyway?

2m:
With the hint from https://github.com/sbt/sbt/issues/3850 I was able to make it work without the system property: https://github.com/2m/akka-http/commits/wip-travis-to-sbt-2m

It is cleaner in once sense but uses another not-really-documented SBT api.

project/Publish.scala
```diff
@@ -21,12 +21,17 @@ object Publish extends AutoPlugin {
...
+    val publishRepo = CliOption("akka.publish.repo", "maven")
+  }
+
   override def trigger = allRequirements
   override def requires = BintrayPlugin
 
   override def projectSettings = Seq(
     bintrayOrganization := Some("akka"),
-    bintrayPackage := "com.typesafe.akka:akka-http_2.11"
+    bintrayPackage := "com.typesafe.akka:akka-http_2.11",
```
raboof: I know it isn't part of this PR, but is `_2.11` correct here?

2m: Unfortunately, it is. This is the original name of the bintray package that akka-http is deployed to. We can not rename it as it is connected to the way artifacts sync to jcenter.

raboof:
This is pending finding a nice way to periodically clean up bintray

raboof:
I guess we could make a script that gets the versions from bintray (https://bintray.com/docs/api/#_get_package), looks at the version naming numbers to see which are 'old' snapshots and deletes (https://bintray.com/docs/api/#url_delete_version) those?

2m:
I would like to have deletion added to sbt-bintray plugin, so it is not an ad-hoc solution. But the logic would be pretty much like the one you described.

raboof:
Added https://github.com/sbt/sbt-bintray/issues/154

travis.sbt
```diff
@@ -0,0 +1,32 @@
+commands ++= Seq(
+  // check policies, if on master also upload
+  Command.command("whitesourceCi") { state =>
+    val extracted = Project.extract(state)
+    import extracted._
+
+    if (sys.env.getOrElse("TRAVIS_SECURE_ENV_VARS", "") == "true") {
+      val stateWithCredentials =
+        append(credentials += Credentials("whitesource", "whitesourcesoftware.com", "", System.getenv("WHITESOURCE_KEY")), state)
```
raboof: I think this is no longer needed since https://github.com/lightbend/sbt-whitesource/pull/29

2m:
Superseded by #2365

Issue #1743, opened 2018-01-03
Allow response to start before request is finished


raboof:
Currently `WaitingForEndOfRequestEntity` and `WaitingForResponse` are explicitly separate steps in the new client pool, meaning when a response comes in while the request is not yet finished the connection will fail.

We should add test coverage for this and implement it.

PR #1746, opened 2018-01-03
Accept response before request has been sent completely

raboof:
Perhaps a bit naive, but might work? I guess this does open up the
possibility for misbehaving clients to forget to finish requests,
but I'm not sure what would be the right level to fix that.

The test should probably be extended further

Refs #1743, #1930


raboof:
Yes, perhaps stashing it until we find a real-world scenario where this is really reasonable and needed might make sense :+1:

jrudolph:
Also see this related discussion: https://github.com/akka/akka-http/commit/98e4dd6616fd882bd14d46a3bff3d152414cad9a#r27930353

akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/NewHostConnectionPool.scala
```diff
@@ -324,19 +325,21 @@ private[client] object NewHostConnectionPool {
...
 
-            // bit of a HACK to make sure onRequestEntityCompleted will end up in the right place
-            state = nextState
+            // bit of a HACK: pushing the request may cause a onRequestEntityCompleted event on the current thread.
 
+            // To accomodate this we first do an 'early' update of the state:
+            state = nextState
+            // Then execute the action that might cause the 'inline' state change:
             connection.pushRequest(request)
+            // And then return the possibly-again-updated state so we don't overwrite it afterwards:
```
ktoso: thanks for the comments, good to make it more understandable while we're here ;)

akka-http-core/src/test/scala/akka/http/impl/engine/client/ConnectionPoolSpec.scala
```diff
@@ -409,6 +411,61 @@ abstract class ConnectionPoolSpec(poolImplementation: PoolImplementation) extend
...
+      sourceQueue.complete()
+      val bytes = response.entity.dataBytes.runReduce(_ ++ _)
+      Await.result(bytes, 3.seconds) should be(ByteString("lala"))
+    }
+
+    /**
+     * Currently failing the 'outgoing request' part of the connection may also fail the 'incoming reply' part of the connection.
+     * In the future we may want to disconnect those and allow the server we connect to to choose how to handle the failure
+     * of the request entity.
+     */
```
ktoso: Hmmm yeah I wonder about that, I guess we could but would rather have someone request for it specifically

akka-http-core/src/test/scala/akka/http/impl/engine/client/ConnectionPoolSpec.scala
```diff
@@ -409,6 +411,61 @@ abstract class ConnectionPoolSpec(poolImplementation: PoolImplementation) extend
...
+      val sourceQueue = Await.result(sourceQueuePromise.future, 3.seconds)
+      val response = Await.result(responseFuture, 1.second.dilated)
+      response.headers should contain(RawHeader("Req-Host", s"$serverHostName:$serverPort"))
+
+      sourceQueue.offer(HttpEntity.Chunk("lala")).futureValue should be(QueueOfferResult.Enqueued)
+      sourceQueue.offer(LastChunk).futureValue should be(QueueOfferResult.Enqueued)
+      sourceQueue.complete()
+      val bytes = response.entity.dataBytes.runReduce(_ ++ _)
+      Await.result(bytes, 3.seconds) should be(ByteString("lala"))
+    }
```
ktoso: good test

akka-http-core/src/test/scala/akka/http/impl/engine/client/HostConnectionPoolSpec.scala
```diff
@@ -173,7 +173,7 @@ class HostConnectionPoolSpec extends AkkaSpec(
       "time out and reconnect when a request is not handled in time" in pending
       "time out when connection cannot be established" in pending
       "fail a request if the request entity fails" in new SetupWithServerProbes {
-        val reqBytesOut = pushChunkedRequest()
+        val reqBytesOut = pushChunkedRequest(numRetries = 0)
```
ktoso: was this needed?

akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/SlotState.scala
```diff
@@ -255,13 +265,30 @@ private[pool] object SlotState {
...
+    ongoingResponse:              HttpResponse,
+    waitingForEndOfRequestEntity: Boolean) extends ConnectedState with BusyWithResultAlreadyDetermined {
 
     override def onResponseEntityCompleted(ctx: SlotContext): SlotState =
-      if (ctx.willCloseAfter(ongoingResponse) || ctx.isConnectionClosed)
+      if (waitingForEndOfRequestEntity)
+        WaitingForEndOfRequestEntity
+      // TODO can we be *sure* that by skipping to Unconnected if ctx.willCloseAfter(ongoingResponse)
+      // we can't get a connection closed event from the 'previous' connection later?
+      else if (ctx.willCloseAfter(ongoingResponse) || ctx.isConnectionClosed)
```
ktoso: not entirely sure but presence of a close header kind of implies that after this request a close is expected... though you may have dug deeper into the state machine here and see something I don't?

ktoso:
General comment on https://github.com/akka/akka-http/pull/1746#issuecomment-387646118
note that this is really only for error cases, but yeah, the spec does encourage clients to monitor for replies while still sending the request, so in general we're good I think

raboof:
I agree this is mostly for error scenarios (in #1930 it was reported to be hit when the Amazon ELB is rejecting requests), but a common enough error scenario to support :smile:

akka-http-core/src/test/scala/akka/http/impl/engine/client/HostConnectionPoolSpec.scala
raboof: The test succeeded before, but would keep retrying after this particular test had already finished, making the logging for the *next* test very confusing to read :).

akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/NewHostConnectionPool.scala
raboof: yeah this had me puzzled for a while ;)

akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/SlotState.scala
raboof: I agree the presence of a close header implies that after this request a close is expected.

What I'm afraid of is that since this `Slot` will be re-used for the next connection, there might be a race condition in which we'll get the `onConnectionCompleted` of the 'previous' connection while we've already moved on to the 'next' connection.

This is pretty much unrelated to this PR but just an observation I ran into, though.

akka-http-core/src/test/scala/akka/http/impl/engine/client/ConnectionPoolSpec.scala
raboof: I tried it for a bit but it didn't seem straightforward, so I agree we should probably leave it be until we find a strong use case for it.

ktoso:
Lookin good in general it seems, I think you can merge at will if you want

akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/SlotState.scala
jrudolph: Good point and it certainly plagued me before as well (maybe the legacy pool as well). It is prevented by using `withSlot` inside the `Connection` which first checks if the connection is still the current one of the slot.

PR #1749, opened 2018-01-04
Add HTTP status code 421

aarongut:
Status code document at https://tools.ietf.org/html/rfc7540#section-9.1.2


lightbend-cla-validator:
Hi @aarongut,

aarongut:
CLA signed

ktoso:
OK TO TEST

ktoso:
Hmm, sounds good; I think I'd want to link to the RFC from the code since it's not as widely known I think?

aarongut:
Would that be consistent with the other codes? It is listed in the wikipedia page, and the only code with a comment is 420, which is Twitter specific.

ktoso:
Clarity is more important here than being consistent with not documenting a status code ;) I mean, it seems this one was defined only recently in the HTTP/2 RFC? On my phone and could not look very deep into it but adding s comment should not hurt I think :)

jrudolph:
Thanks, @aarongut.

PR #1750, opened 2018-01-07
Replaces triple equals assertions with shouldEqual (#1136)

nikom1337:
I've replaced almost all of the triple equals assertions with shouldEqual from the code base. Some of the test cases stopped passing after that change so I've fixed them. 

The last triple equals assertion is located in `akka.http.impl.engine.ws.Utf8CodingSpecs` test class. But fixing that probably will take more time because the failure cause was not clear to me. 

Should I create a new issue for Utf8CodingSpecs as a bug or continue it in #1136?

Refs: #1136

### DIFF
--- a/docs/src/test/scala/docs/http/scaladsl/server/ExceptionHandlerExamplesSpec.scala
+++ b/docs/src/test/scala/docs/http/scaladsl/server/ExceptionHandlerExamplesSpec.scala
@@ -17,15 +17,15 @@ object MyExplicitExceptionHandler {
   import StatusCodes._
   import Directives._
 
-  object MyApp extends App {
+  val myExceptionHandler = ExceptionHandler {
+    case _: ArithmeticException =>
+      extractUri { uri =>
+        println(s"Request to $uri could not be handled normally")
+        complete(HttpResponse(InternalServerError, entity = "Bad numbers, bad result!!!"))
+      }
+  }
 
-    val myExceptionHandler = ExceptionHandler {
-      case _: ArithmeticException =>
-        extractUri { uri =>
-          println(s"Request to $uri could not be handled normally")
-          complete(HttpResponse(InternalServerError, entity = "Bad numbers, bad result!!!"))
-        }
-    }
+  object MyApp extends App {
 
     implicit val system = ActorSystem()
     implicit val materializer = ActorMaterializer()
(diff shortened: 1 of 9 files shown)


jlprat:
OK TO TEST

jlprat:
Thanks for the PR!
I would create a separate issue for the utf8 case as it probably is hiding an underlying bug.

jlprat:
Failure was #1751

PLS BUILD

docs/src/test/scala/docs/http/scaladsl/server/ExceptionHandlerExamplesSpec.scala
```diff
@@ -17,15 +17,15 @@ object MyExplicitExceptionHandler {
...
-  object MyApp extends App {
+  val myExceptionHandler = ExceptionHandler {
+    case _: ArithmeticException =>
+      extractUri { uri =>
+        println(s"Request to $uri could not be handled normally")
+        complete(HttpResponse(InternalServerError, entity = "Bad numbers, bad result!!!"))
+      }
+  }
 
-    val myExceptionHandler = ExceptionHandler {
```
jrudolph: This was written like this by purpose to allow copy/pasting the docs.

docs/src/test/scala/docs/http/scaladsl/server/ExceptionHandlerExamplesSpec.scala
```diff
@@ -52,16 +52,16 @@ object MyImplicitExceptionHandler {
   import StatusCodes._
   import Directives._
 
-  object MyApp extends App {
+  implicit def myExceptionHandler: ExceptionHandler =
```
jrudolph: Here as well.

docs/src/test/scala/docs/http/scaladsl/server/directives/ParameterDirectivesExamplesSpec.scala
```diff
@@ -128,15 +128,15 @@ class ParameterDirectivesExamplesSpec extends RoutingSpec with PredefinedFromStr
...
     }
 
     Get("/?color=blue&city=Chicago") ~> Route.seal(route) ~> check {
-      responseAs[String] === "The color is 'blue' and the city is Chicago."
+      responseAs[String] shouldEqual "The color is 'blue' and the city is Chicago."
     }
 
     Get("/?color=blue&city=Chicago&city=Boston") ~> Route.seal(route) ~> check {
-      responseAs[String] === "The color is 'blue' and the cities are Chicago, Boston."
+      responseAs[String] shouldEqual "The color is 'blue' and the cities are Boston, Chicago."
```
jrudolph: Oops. Great to fix those.

nikom1337:
I've created new issue #1760 regarding `Utf8CodingSpecs` test class.

docs/src/test/scala/docs/http/scaladsl/server/ExceptionHandlerExamplesSpec.scala
nikom1337: As far as i remember the problem with previous implementation was that App class is extending a DelayedInit. Because of that in the both tests handlers are not initialized. e.g MyExplicitExceptionHandler.MyApp.myExceptionHandler is null.

jlprat: @jrudolph if you want to have it copy paste-able, maybe the whole snippet can be wrapped in an `object`. As `App` extends `DelayedInit` it causes line 82 to blow with an exception as it's not initialized.

jrudolph: I see. Maybe then it's ok the way it is now.

PR #1759, opened 2018-01-08
Fix a few warnings

jrudolph:
- akka.stream.impl.ConstantFun => akka.util.ConstantFun
 - ''' => '\''
 - Left.a => Left.value
 - Remove unused imports

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
@@ -7,7 +7,7 @@ package akka.http.impl.engine.client
 import akka.NotUsed
 import akka.annotation.InternalApi
 import akka.http.scaladsl.settings.{ ClientConnectionSettings, ParserSettings }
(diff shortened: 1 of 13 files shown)


jrudolph:
PLS BUILD

jrudolph:
Something broke but I'm not yet sure what. Will investigate later.

jrudolph:
Observed problem was #1786.

jrudolph:
Removed the problematic change..

jrudolph:
Missed one change that should rather be fixed in https://github.com/akka/akka-http/pull/1787/files.

akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
```diff
@@ -7,7 +7,7 @@ package akka.http.impl.engine.client
 import akka.NotUsed
 import akka.annotation.InternalApi
 import akka.http.scaladsl.settings.{ ClientConnectionSettings, ParserSettings }
-import akka.stream.impl.ConstantFun
+import akka.util.ConstantFun
```
ktoso: good :)

Issue #1769, opened 2018-01-12
Support OAuth2 access token in URI query component as defined in RFC6750


mwegrz:
Currently OAuth2 authentication in Akka HTTP supports access tokens only in `Authorization: Bearer ..` header. [RFC6750 in section 2.3](https://tools.ietf.org/html/rfc6750#section-2.3) allows (but does not recommend) including access tokens also in the URI query component. Akka HTTP might support this use case and leave the decision what method to use to the client making the call.

jrudolph:
Implemented in #1771.

PR #1771, opened 2018-01-12
Support OAuth2 access token in URI query component as defined in RFC6750

mwegrz:
Implements https://github.com/akka/akka-http/issues/1769

If `Authorization` header is missing check `access_token` URI query parameter instead.


mwegrz:
Ready to test.

jrudolph:
OK TO TEST

mwegrz:
Thx @jrudolph. The test failed but I think the failure is not related to this PR. Could you retest?

jrudolph:
PLS BUILD

jrudolph:
I added a commit, slightly changing the implementation not to run the authenticator twice but instead falling back on the parameter credentials when the header cannot be found. @mwegrz, what do you think about that?

ktoso:
Tests have been run on the latest commit if that's what you're asking?

mwegrz:
Sorry, I overlooked that. All good then.

jrudolph:
Thanks a lot @mwegrz.

PR #1772, opened 2018-01-14
=doc change the formation of the & operator

satansk:
using `&` instead of *&*


jlprat:
OK TO TEST

jlprat:
Thanks @satansk !

PR #1779, opened 2018-01-15
ActorSystem leaks in tests

johanandren:
Seen locally that the two ConnectionSpec variations leak an actor system each


johanandren:
Hmm, that is so weird. On MacOS that fails consistently, something with how connections are closed/kept hanging around?

jrudolph:
Would be good to know what's going on but let's close this one.

PR #1780, opened 2018-01-15
Silence noisy debug logs from ConnectionPoolSpec and HostConnectionPoolSpec for successful tests

jrudolph:
But show them after a test failed. If that actually works we can use it in other tests as well.


jrudolph:
I'm rerunning nightlies to see if we get a bit more information for remaining failures.

PR #1787, opened 2018-01-16
=htc #1786 fix customMediaTypes behavior

### LINKED ISSUE #1786: Enabling ParsingSettings.customMediaTypes breaks predefined media type parsing
In the following code snippet `ParsingSettings.customMediaTypes` is compared to a somewhat arbitrary value to check if behavior should be enabled or not. This is fine as an optimization but providing an equivalent function should not change the behavior. 

https://github.com/akka/akka-http/blob/3c966af2255b58b02fbdacd6abec89c47f78ce48/akka-http-core/src/main/scala/akka/http/impl/model/parser/CommonActions.scala#L59

Providing an equivalent function (i.e. one that always returns `None`) makes RequestParserCRLFSpec and many other tests fail.

This was discovered while migrating `akka.stream.impl.ConstantFun`  to `akka.util.ConstantFun`.

jrudolph:
Previously, defining customMediaType in ParserSettings broke parsing of
predefined media types or custom ad-hoc media-types with charsets. This had
two reasons:

 * when custom media types were defined, we never parsed into our own
   predefined ones any more
 * areCustomMediaTypesDefined was defined the wrong way round: just checking
   that the value was changed from the magic one doesn't mean there might
   still be none defined

Fixes #1786.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/model/parser/CommonActions.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/model/parser/CommonActions.scala
@@ -6,15 +6,14 @@ package akka.http.impl.model.parser
 
 import akka.annotation.InternalApi
 import akka.http.impl.util._
-import akka.http.scaladsl.model.MediaType.Binary
 import akka.http.scaladsl.model._
-import akka.stream.impl.ConstantFun
 
 /** INTERNAL API */
 @InternalApi
 private[parser] trait CommonActions {
 
   def customMediaTypes: MediaTypes.FindCustom
+  def areNoCustomMediaTypesDefined: Boolean
 
   type StringMapBuilder = scala.collection.mutable.Builder[(String, String), Map[String, String]]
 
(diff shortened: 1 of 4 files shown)


akka-http-core/src/main/scala/akka/http/impl/model/parser/CommonActions.scala
```diff
@@ -33,29 +32,27 @@ private[parser] trait CommonActions {
...
-          customMediaTypes(mainLower, subType) getOrElse fallbackMediaType(subType, params, mainLower)
-        else MediaTypes.getForKey((mainLower, subLower)) match {
-          case Some(registered) ⇒ if (params.isEmpty) registered else registered.withParams(params)
-          case None ⇒
-            if (charsetDefined)
-              MediaType.customWithOpenCharset(mainLower, subType, params = params, allowArbitrarySubtypes = true)
-            else
-              fallbackMediaType(subType, params, mainLower)
-        }
+        if (areNoCustomMediaTypesDefined) // try to prevent closure creation for usual case without custom media types
```
ktoso: good

Issue #1773, opened 2018-01-15
FAILED: akka.http.impl.engine.client.HostConnectionPoolSpec.LegacyPoolImplementation on AkkaHttpEngineTCP should fail a request if the connection stream fails while waiting for request entity bytes


jrudolph:
https://jenkins.akka.io:8498/job/pr-validator-akka-http/2097/testReport/junit/akka.http.impl.engine.client/HostConnectionPoolSpec/LegacyPoolImplementation_on_AkkaHttpEngineTCP_should_fail_a_request_if_the_connection_stream_fails_while_waiting_for_request_entity_bytes/

```
Error Message

Expected OnNext(_), yet no element signaled during 4500 milliseconds

Stacktrace

java.lang.AssertionError: Expected OnNext(_), yet no element signaled during 4500 milliseconds
	at akka.stream.testkit.TestSubscriber$ManualProbe.expectNext(StreamTestKit.scala:325)
	at akka.stream.testkit.TestSubscriber$ManualProbe.expectNext(StreamTestKit.scala:316)
	at akka.stream.testkit.TestSubscriber$Probe.requestNext(StreamTestKit.scala:753)
	at akka.http.impl.engine.ws.ByteStringSinkProbe$$anon$1.expectBytes(ByteStringSinkProbe.scala:70)
	at akka.http.impl.engine.ws.ByteStringSinkProbe$$anon$1.expectUtf8EncodedString(ByteStringSinkProbe.scala:80)
	at akka.http.impl.engine.client.HostConnectionPoolSpec$$anonfun$testSet$1$$anonfun$apply$mcV$sp$11$$anon$7.<init>(HostConnectionPoolSpec.scala:200)
	at akka.http.impl.engine.client.HostConnectionPoolSpec$$anonfun$testSet$1$$anonfun$apply$mcV$sp$11.apply(HostConnectionPoolSpec.scala:193)
	at akka.http.impl.engine.client.HostConnectionPoolSpec$$anonfun$testSet$1$$anonfun$apply$mcV$sp$11.apply(HostConnectionPoolSpec.scala:193)
```

2m:
https://jenkins.akka.io:8498/job/akka-http-nightly-release-10.0-against-akka-master/32/testReport/junit/akka.http.impl.engine.client/HostConnectionPoolSpec/LegacyPoolImplementation_on_AkkaHttpEngineTCP_should_fail_a_request_if_the_connection_stream_fails_while_waiting_for_request_entity_bytes/

[...]

Issue #1765, opened 2018-01-11
FAILED: LegacyPoolImplementation on PassThrough create a new connection when previous one failed between requests 


ktoso:
```

[info] - should create a new connection when previous one failed between requests *** FAILED ***
[info]   java.lang.AssertionError: assertion failed: timeout (3 seconds) during expectMsgClass waiting for class akka.http.impl.engine.client.HostConnectionPoolSpec$SetupWithServerProbes$2$ServerConnection
[info]   at scala.Predef$.assert(Predef.scala:219)
[info]   at akka.testkit.TestKitBase.expectMsgClass_internal(TestKit.scala:508)
[info]   at akka.testkit.TestKitBase.expectMsgType(TestKit.scala:481)
[info]   at akka.testkit.TestKitBase.expectMsgType$(TestKit.scala:481)
[info]   at akka.testkit.TestKit.expectMsgType(TestKit.scala:890)
[info]   at akka.http.impl.engine.client.HostConnectionPoolSpec$SetupWithServerProbes$2.expectNextConnection(HostConnectionPoolSpec.scala:444)
[info]   at akka.http.impl.engine.client.HostConnectionPoolSpec$$anon$12.<init>(HostConnectionPoolSpec.scala:269)
[info]   at akka.http.impl.engine.client.HostConnectionPoolSpec.$anonfun$testSet$17(HostConnectionPoolSpec.scala:259)
[info]   at org.scalatest.OutcomeOf.outcomeOf(OutcomeOf.scala:85)
[info]   at org.scalatest.OutcomeOf.outcomeOf$(OutcomeOf.scala:83)
```

https://jenkins.akka.io:8498/job/akka-http-nightly-release-10.0-against-akka-master-on-2.12/27//consoleFull

I thought we don't run the Legacy one anymore?

2m:
https://jenkins.akka.io:8498/job/akka-http-nightly-against-akka-master-on-2.12/396/testReport/junit/akka.http.impl.engine.client/HostConnectionPoolSpec/LegacyPoolImplementation_on_AkkaHttpEngineTCP_should_fail_a_request_if_the_connection_stream_fails_while_waiting_for_request_entity_bytes/

2m:
[...]

PR #1788, opened 2018-01-16
=htc disable LegacyPoolImplementation in HostConnectionPoolSpec

jrudolph:
These are basically new tests for the old implementation. Let's disable
those and focus on improving reliability of the new one.


jrudolph:
Fixes #1773, #1765.

Issue #1793, opened 2018-01-17
Uri.Path ?/ operator


tg44:
We can write something like this now:
`Path("abc") / "def"` which will be `abc/def`
But if we have a tailing / we will get:
`Path("abc/") / "def"` => `abc//def` which is not so handy

If you use akka-http-core as a http client too this small thing is really irritating... So I wrote a small helper to solve this problem, but it would be better in the official lib.
(PR coming.)

jrudolph:
Fixed by #1794.

PR #1794, opened 2018-01-17
Uri.Path ?/ operator #1793

tg44:
solves #1793

### DIFF
--- a/akka-http-core/src/test/scala/akka/http/scaladsl/model/UriSpec.scala
+++ b/akka-http-core/src/test/scala/akka/http/scaladsl/model/UriSpec.scala
@@ -256,6 +256,10 @@ class UriSpec extends WordSpec with Matchers {
       Path("/abc/def").endsWithSlash shouldBe false
       Path("/abc/def/").endsWithSlash shouldBe true
     }
+    "support the `?/` operator" in {
+      Path("abc") ?/ "def" shouldEqual Path("abc/def")
+      Path("abc/") ?/ "def" shouldEqual Path("abc/def")
+    }
     "support the `dropChars` modifier" in {
       Path./.dropChars(0) shouldEqual Path./
       Path./.dropChars(1) shouldEqual Empty
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/Uri.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/Uri.scala
@@ -499,6 +499,13 @@ object Uri {
     def reverse: Path = reverseAndPrependTo(Path.Empty)
     def reverseAndPrependTo(prefix: Path): Path
     def /(segment: String): Path = this ++ Path.Slash(segment :: Path.Empty)
+    def ?/(segment: String): Path = {
+      if (this.endsWithSlash) {
+        this + segment
+      } else {
+        this / segment
+      }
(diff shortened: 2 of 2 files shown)


jlprat:
OK TO TEST

jlprat:
Hi @tg44  Thanks for your contribution!

akka-http-core/src/test/scala/akka/http/scaladsl/model/UriSpec.scala
```diff
@@ -256,6 +256,10 @@ class UriSpec extends WordSpec with Matchers {
       Path("/abc/def").endsWithSlash shouldBe false
       Path("/abc/def/").endsWithSlash shouldBe true
     }
+    "support the `?/` operator" in {
+      Path("abc") ?/ "def" shouldEqual Path("abc/def")
+      Path("abc/") ?/ "def" shouldEqual Path("abc/def")
+    }
```
jlprat: Nitpicking, but I would also add the "annoying" case here, maybe more as documentation than because it's needed.

tg44:
Docs:
It's not a path matcher. More like a builder. I searched references for the Uri.Path./ function, and it's only used in tests. I think it is just a helper for the http-client part, where you need an uri to start a new request.
If it should be mentioned somewhere I think [this place](https://doc.akka.io/docs/akka-http/current/common/uri-model.html) is the suitable one.
(but the / and + only mentioned [in there](https://doc.akka.io/api/akka-http/10.1.0-RC1/akka/http/scaladsl/model/Uri$$Path.html) so I think we don't really need to document this)

Java:
[It is implemented already :D](https://github.com/akka/akka-http/blob/master/akka-http-core/src/main/scala/akka/http/impl/model/JavaUri.scala#L73)

Tests:
This is what I want and this 2 case is totally cover it. (I can add more, but I think this is pointless :D )

jlprat:
Good for the Java part already done! 😀
Regarding docs, https://doc.akka.io/docs/akka-http/current/routing-dsl/path-matchers.html#modifiers here you can find the explanation for `/`, there is where I would add this new "operator" `?/`

jlprat:
Regarding test, yes I was nitpicking, not really needed.

jrudolph:
@jlprat 

> Regarding docs, https://doc.akka.io/docs/akka-http/current/routing-dsl/path-matchers.html#modifiers here you can find the explanation for /, there is where I would add this new "operator" ?/

The docs are about `PathMatcher./`, not `Path./`, right?

jlprat:
Yes, I did the code check on mobile phone, maybe I misunderstood something or overlooked something.

tg44:
Super :D Anything other from me?

jlprat:
But yes, now I realise my error, sorry for that. This refers to Path building not PathMatcher.


Sorry again!

tg44:
Two things; 
- 1st there is an extra line break which fails the format check in the build, and 
- 2nd I personally hate when sb use def-s and if-s without bracers but with newlines :D So I think the two "readable" and "error safe" option is the one I provided originally or this oneliner:
`def ?/(segment: String): Path = if (this.endsWithSlash) this + segment else this / segment` (but ofc it's just personal taste)

jrudolph:
Great, thanks @tg44.

PR #1800, opened 2018-01-19
Use HTTP 431 when header exceeds value or name max length

marcospereira:
HTTP status code 431 is more specific and described in [RFC 6585](https://tools.ietf.org/html/rfc6585#section-5).

Related to https://github.com/playframework/playframework/pull/8162#discussion_r161883824.


jlprat:
OK TO TEST

PR #1801, opened 2018-01-19
Add `Tls-session-info` header for http/2

### LINKED ISSUE #1432: Http2 provide Tls Session Info header
I tried out http2 and found that the Tls Session Info header was not being provided. This works fine if I don't use the Java agent which makes the negotiation fall back to using http/1.1.

greenhat:
Fixes #1432

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpAttributes.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpAttributes.scala
@@ -30,6 +30,11 @@ private[akka] object HttpAttributes {
   private[akka] def remoteAddress(address: InetSocketAddress): Attributes =
     Attributes(RemoteAddress(address))
 
+  /**
+   * INTERNAL API
+   * Internally used TLS session info in the HTTP pipeline.
+   */
+  @InternalApi
   private[akka] final case class TLSSessionInfo(session: SSLSession) extends Attribute
 
   private[akka] def tlsSessionInfo(session: SSLSession): Attributes =
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpAttributes.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpAttributes.scala
@@ -5,6 +5,7 @@
 package akka.http.impl.engine.server
 
 import java.net.InetSocketAddress
+import javax.net.ssl.SSLSession
 
 import akka.annotation.InternalApi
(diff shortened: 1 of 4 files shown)


jlprat:
OK TO TEST

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpAttributes.scala
```diff
@@ -29,6 +30,11 @@ private[akka] object HttpAttributes {
   private[akka] def remoteAddress(address: InetSocketAddress): Attributes =
     Attributes(RemoteAddress(address))
 
+  private[akka] final case class TLSSessionInfo(session: SSLSession) extends Attribute
```
ktoso: needs comments and annotations `@InternalApi` (for tools, in the future) and `/** INTERNAL API */` (for java, since private[] is *public* in Java)

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -717,6 +718,24 @@ class Http2ServerSpec extends AkkaSpec("""
         remoteAddressHeader.address.getPort shouldBe thePort
       }
 
+      "expose Tls-Session-Info" in new TestSetup with RequestResponseProbes with AutomaticHpackWireSupport {
+        override def settings: ServerSettings =
+          super.settings.withParserSettings(super.settings.parserSettings.withIncludeTlsSessionInfoHeader(true))
```
ktoso: no need for the `super`?

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpAttributes.scala
greenhat: @ktoso Thanks! Done.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -717,6 +718,24 @@ class Http2ServerSpec extends AkkaSpec("""
         remoteAddressHeader.address.getPort shouldBe thePort
       }
 
+      "expose Tls-Session-Info" in new TestSetup with RequestResponseProbes with AutomaticHpackWireSupport {
```
jrudolph: :+1: for adding tests :)

PR #1803, opened 2018-01-21
Add HTTP status code 103

Philippus:


ktoso:
(make sure has a ticket or the PR is assigned to the milestone)

jrudolph:
Looks good, I wonder if we should mark it as `@ApiMayChange` for now since the RFC is still experimental?

ktoso:
Good catch, yeah lets. Would you add that annotation please @Philippus ?

Philippus:
the RFC is experimental but the HTTP status code is already in the official registry: 
https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml
But happy to add it if you want of course.

jlprat:
We can remove the `@ApiMayChange` at any time :)

akka-http-core/src/main/scala/akka/http/scaladsl/model/StatusCode.scala
```diff
@@ -104,6 +105,9 @@ object StatusCodes extends ObjectRegistry[Int, StatusCode] {
   val SwitchingProtocols = reg(i(101)("Switching Protocols", "The server is switching protocols, because the client requested the switch."))
   val Processing         = reg(i(102)("Processing", "The server is processing the request, but no response is available yet."))
 
+  @ApiMayChange
+  val EarlyHints         = reg(i(103)("Early Hints", "The server is likely to send a final response with the header fields included in the response.")) // RFC 8297
```
raboof: The text is a bit puzzling to me, but that's how the RFC also describes it, so OK :).

akka-http-core/src/main/java/akka/http/javadsl/model/StatusCodes.java
```diff
@@ -19,6 +19,7 @@ private StatusCodes() {}
     public static final StatusCode CONTINUE = akka.http.scaladsl.model.StatusCodes.Continue();
     public static final StatusCode SWITCHING_PROTOCOLS = akka.http.scaladsl.model.StatusCodes.SwitchingProtocols();
     public static final StatusCode PROCESSING = akka.http.scaladsl.model.StatusCodes.Processing();
+    public static final StatusCode EARLY_HINTS = akka.http.scaladsl.model.StatusCodes.EarlyHints();
```
jlprat: I guess `@ApiMayChange` also needs to go here :)

jrudolph:
unrelated failure

PLS BUILD

jrudolph:
Thanks, @Philippus!

PR #1806, opened 2018-01-23
Suggest valid max-open-requests values on config error

raboof:


akka-http-core/src/main/scala/akka/http/impl/settings/ConnectionPoolSettingsImpl.scala
```diff
@@ -44,11 +44,21 @@ private[akka] final case class ConnectionPoolSettingsImpl(
...
   require(maxRetries >= 0, "max-retries must be >= 0")
-  require(maxOpenRequests > 0 && (maxOpenRequests & (maxOpenRequests - 1)) == 0, "max-open-requests must be a power of 2 > 0")
+  require(maxOpenRequests > 0, "max-open-requests must be a power of 2 > 0.")
+  require((maxOpenRequests & (maxOpenRequests - 1)) == 0, "max-open-requests must be a power of 2. " + suggestPowerOfTwo(maxOpenRequests))
   require(pipeliningLimit > 0, "pipelining-limit must be > 0")
   require(idleTimeout >= Duration.Zero, "idle-timeout must be >= 0")
 
   override def productPrefix = "ConnectionPoolSettings"
+
+  private def suggestPowerOfTwo(around: Int): String = {
```
jrudolph: Shouldn't we use a similarly advanced way of figuring out like the above bit-fiddling, just for consistence? ;)

How about something like that [SO answer](https://stackoverflow.com/a/10143264/7647) using [Integer.numberOfLeadingZeros](https://docs.oracle.com/javase/8/docs/api/java/lang/Integer.html#numberOfLeadingZeros-int-).

raboof: ;)

Philippus: it doesn't give correct answers for `around` values that are already powers of two..

raboof: Heh ;). The point is a bit moot since it is never called with `around` values that are already powers of two.. or do you have an elegant proposal for a replacement?

Philippus: they all involve an extra `if` I'm afraid. :)

akka-http-core/src/main/scala/akka/http/impl/settings/ConnectionPoolSettingsImpl.scala
```diff
@@ -52,11 +52,11 @@ private[akka] final case class ConnectionPoolSettingsImpl(
   override def productPrefix = "ConnectionPoolSettings"
 
   private def suggestPowerOfTwo(around: Int): String = {
-    @annotation.tailrec
-    def findAbove(n: Int = 2): Int = if (n > around) n else findAbove(n * 2)
+    val firstBit = 31 - Integer.numberOfLeadingZeros(around)
+
+    val below = 1 << firstBit
+    val above = 1 << (firstBit + 1)
```
Philippus: or `2 << firstBit`

PR #1808, opened 2018-01-23
Update JMH

ktoso:
Seems we were on a pretty old one... may not just work with changing the number, trying though...


ktoso:
Seems happy then, nice

PR #1813, opened 2018-01-26
Log the timed out request

### LINKED ISSUE #1810: Server request timeouts are silent
When the server request timeout triggers (`akka.http.server.request-timeout`, https://github.com/akka/akka-http/blob/master/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala#L302), we produce a `503` response with a nice message, but the logging is silent.

While this could happen during normal operation, it is also something that might point to a problem, so it might make sense to log at INFO level when this timeout is hit (and leave it up to monitoring solutions to give more detailed insight)?

greenhat:
Closes #1810.


raboof:
OK TO TEST

greenhat:
@raboof Thanks. Fixed.

PR #1814, opened 2018-01-26
Fix nesting WrapDirective's in docs

raboof:
This [does not work](https://github.com/lightbend/paradox/issues/185)

In this case this made Scala-specific content pop up at https://doc.akka.io/docs/akka-http/current/routing-dsl/case-class-extraction.html?language=java


jrudolph:
Let's update this with the deeper nesting you found out.

raboof:
Seems to work indeed!

selection_289

Issue #386, opened 2016-10-12
Multipart-Form-Data in akka-http, missing quotes in Content-Disposition field


ktoso:
<a href="https://github.com/KrzyMucha"><img src="https://avatars.githubusercontent.com/u/14311593?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [KrzyMucha](https://github.com/KrzyMucha)**
_Wednesday Oct 12, 2016 at 08:05 GMT_
_Originally opened as https://github.com/akka/akka/issues/21656_
---

When we use akka-http for http request with following code:

```
Multipart
            .FormData(Multipart.FormData.BodyPart("file",
                HttpEntity.Strict(MediaTypes.`image/jpeg`, ByteString("test")), Map("filename" -> "somefile.jpeg"))).toEntity()
```

Then Http request is as follows

```
User-Agent: [akka-http/2.4.8]
Transfer-Encoding: [chunked]
Host: [localhost:8089]
Content-Type: [multipart/form-data; boundary=xWtacaNwjOWLKFExaR8fOv6Q; charset=UTF-8]

--xWtacaNwjOWLKFExaR8fOv6Q
Content-Type: image/jpeg
Content-Disposition: form-data; filename=somefile.jpeg; name=file

test
--xWtacaNwjOWLKFExaR8fOv6Q--
```

Parameters **filename** and **file** are without quotation marks. However following HTTP specification this params should be quoted. As result some servers (ex. "django" app) are not able to see attached files.

ktoso:
<a href="https://github.com/ktoso"><img src="https://avatars.githubusercontent.com/u/120979?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [ktoso](https://github.com/ktoso)**
_Wednesday Oct 12, 2016 at 08:10 GMT_

----

Please open the issue on github.com/akka/akka-http instead, sinve the project has moved.

AFAIR there is no need to quote these according to the https://tools.ietf.org/html/rfc6266 spec

```
     filename-parm       = "filename" "=" value
                         | "filename*" "=" ext-value

       value                   = token | quoted-string
```

notice the `|` there.

Let's please continue in the right repo.

ktoso:
<a href="https://github.com/KrzyMucha"><img src="https://avatars.githubusercontent.com/u/14311593?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [KrzyMucha](https://github.com/KrzyMucha)**
[...]

Issue #1066, opened 2017-04-27
Figure out how to deal with RFC 5987 header parameters


jrudolph:
[RFC 5987](https://tools.ietf.org/html/rfc5987) describes a syntax that allows to specify values of header parameters including an encoding like this

```
filename*=utf-8''%c2%a3%20and%20%e2%82%ac%20rates.csv
```

With the change in https://github.com/akka/akka-http/pull/695 we would not interpret that notation but would parse it as is as a parameter `filename*` with the value `utf-8''%c2%a3%20and%20%e2%82%ac%20rates.csv`.

However, consistent to what we do otherwise, we should interpret the value and decode it, i.e. storing it as the key-value-pair `"filename" -> "£ and € rates.csv"`. When rendering we should create both a `filename` param with non-ascii characters replaced by placeholders or just skipped and a `filename*` parameter with utf-8 encoding.

raboof:
Interpreting and decoding the value would be 'lossy', as it loses the information of the language (if provided), but that seems reasonable to me in this case.

PR #1815, opened 2018-01-26
Encoded file name is not handled properly

jkugiya:
It seems we are not able to download the file with the name containing the encoded characters with the correct name in current server version. I have used the `filename*` parameter with **Content-Disposition** header to satisfy this request. The older version of Akka server had returned a response like the following.

```
// complete(HttpResponse(200, entity = "foo", headers = immutable.Seq(`Content-Disposition`(ContentDispositionTypes.attachment, Map("filename*" -> "%E3%82%A2%E3%83%83%E3%82%AB.txt")))))

Content-Dispsition: attachment; filename*=utf-8''%E3%82%A2%E3%83%83%E3%82%AB.txt
```

In the current version, server returns the following response.

```
Content-Dispsition: attachment; filename*="utf-8''%E3%82%A2%E3%83%83%E3%82%AB.txt"
```

In Firefox I was able to download with the correct file name "アッカ.txt", but I could not download with the correct name on Edge or Internet Explorer. According to the old issue(#386), it seems that the `filename` parameter had used quoted-string in the old RFC([RFC2616](https://tools.ietf.org/html/rfc2616)), but since the `filename*` parameter is not to quoted in the new RFC([RFC6266](https://tools.ietf.org/html/rfc6266)), it is necessary not be quoted.

This PR fixes not to quote the `filename*` parameter like the previous version of Akka. I ran the following code at the console and it worked correctly on Edge and Internet Explorer.

```scala
import akka.actor.ActorSystem
import akka.stream._
import akka.http.scaladsl.Http
import akka.http.scaladsl.Http.ServerBinding
import akka.http.scaladsl.model.headers._
import akka.http.scaladsl.server.Directives._
import akka.stream.ActorMaterializer
import scala.io.StdIn


implicit val system: ActorSystem = ActorSystem()
implicit val m: Materializer = ActorMaterializer()
implicit val executionContext = system.dispatcher

val route =
  path("foo.md") {
    get {
      complete(HttpResponse(200, entity = "foo", headers = immutable.Seq(`Content-Disposition`(ContentDispositionTypes.attachment, Map("filename" -> "%E3%82%A2%E3%83%83%E3%82%AB.txt")))))
    }
  } ~
  path("bar.md") {
    get {
      complete(HttpResponse(200, entity = "bar", headers = immutable.Seq(`Content-Disposition`(ContentDispositionTypes.attachment, Map("filename*" -> "utf-8''%E3%82%A2%E3%83%83%E3%82%AB.txt")))))
    }
  }

def start() = {
  val f = Http().bindAndHandle(route, "localhost", 8080)
  () => f.flatMap(_.unbind()).foreach(_ => system.terminate)
}
```

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
@@ -397,7 +397,10 @@ final case class `Content-Disposition`(dispositionType: ContentDispositionType, 
   extends jm.headers.ContentDisposition with RequestResponseHeader {
   def renderValue[R <: Rendering](r: R): r.type = {
     r ~~ dispositionType
-    params foreach { case (k, v) ⇒ r ~~ "; " ~~ k ~~ '=' ~~#! v }
+    params foreach {
+      case (k, v) if k == "filename*" ⇒ r ~~ "; " ~~ k ~~ '=' ~~# v
+      case (k, v)                     ⇒ r ~~ "; " ~~ k ~~ '=' ~~#! v
+    }
     r
   }
   protected def companion = `Content-Disposition`
--- a/akka-http-core/src/test/scala/akka/http/impl/model/parser/HttpHeaderSpec.scala
+++ b/akka-http-core/src/test/scala/akka/http/impl/model/parser/HttpHeaderSpec.scala
@@ -194,6 +194,8 @@ class HttpHeaderSpec extends FreeSpec with Matchers {
         `Content-Disposition`(ContentDispositionTypes.attachment, Map("name" → "field1", "filename" → "file/txt"))
       "Content-Disposition: attachment; name=\"field1\"; other=\"\"" =!=
         `Content-Disposition`(ContentDispositionTypes.attachment, Map("name" → "field1", "other" → ""))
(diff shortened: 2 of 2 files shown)


raboof:
OK TO TEST

jrudolph:
Thanks for the suggestion, @jkugiya. It looks good in general. We haven't finally decided about how to deal with those kinds of parameters. We tend to introduce changes which would automatically handle parameters with on US-ASCII correctly by auto-generating those "filename*" parameters. That would probably conflict with the proposed change here. So, we might want to await the resolution of #1066.

jkugiya:
Thank you for your feedback. I understood the circumstances. In the previous version(2.4.11), the behavior had been what we expected about downloading files, but is not in current version. So we have a workarounds as following.

```scala
  val fileName = ???
  val encodedFileName = java.net.URLEncoder.encode(fileName, "UTF8")
  val header = RawHeader("Content-Disposition", s"attachment; filename*=utf-8''${encodedFileName}")
```

I will try to check what kind of conflict there are.

jrudolph:
Yes, using `RawHeader` is the right workaround for now.

I'm closing this one for now. We would like to have automatic rendering for (a selection of?) parameters which are known to contain non-ASCII characters, so that you do not have to worry about what to put inside and have Akka HTTP render both the regular `filename` and the `filename*` params automatically.

Issue #1812, opened 2018-01-26
"Illegal character ' ' in chunk start" for some pages when parsing response with akka-http


SakulK:
This is my example/test page: http://archiwum.teatrszekspirowski.pl/
I attached a stacktrace which isn't all that useful but I tried to debug the problem and found out that it is in HttpMessageParser.parseChunk.parseSize, there is no case for whitespace characters and an error is thrown:
https://github.com/akka/akka-http/blob/40bd7a2d14ecd92eb1527d43a3f56d51e2c17e23/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpMessageParser.scala#L246
I tried a dirty hack by simply adding a case like this:
`case c if CharacterClasses.WSP(c) ⇒ parseSize(cursor + 1, size)`
And it does 'fix' the problem but I don't know how to actually do it properly. Also found a bug report with the same issue in spray:
https://github.com/spray/spray/issues/1063
I can create a pull request with a fix if someone can give me advice what would be the correct way of fixing this issue.
This is the mentioned stacktrace:
>  akka.http.scaladsl.model.EntityStreamException:
> 
>  Illegal character ' ' in chunk start
> 	at akka.http.impl.engine.parsing.HttpMessageParser$$anonfun$chunkedEntity$1$$anonfun$2.applyOrElse(HttpMessageParser.scala:329)
> 	at akka.http.impl.engine.parsing.HttpMessageParser$$anonfun$chunkedEntity$1$$anonfun$2.applyOrElse(HttpMessageParser.scala:327)
> 	at akka.stream.impl.fusing.Collect$$anon$2$$anonfun$10.apply(Ops.scala:212)
> 	at akka.stream.impl.fusing.SupervisedGraphStageLogic.withSupervision(Ops.scala:173)
> 	at akka.stream.impl.fusing.Collect$$anon$2.onPush(Ops.scala:214)
> 	at akka.stream.impl.fusing.GraphInterpreter.processPush(GraphInterpreter.scala:499)
> 	at akka.stream.impl.fusing.GraphInterpreter.processEvent(GraphInterpreter.scala:462)
> 	at akka.stream.impl.fusing.GraphInterpreter.execute(GraphInterpreter.scala:368)
> 	at akka.stream.impl.fusing.GraphInterpreterShell.runBatch(ActorGraphInterpreter.scala:571)
[...]

PR #1819, opened 2018-01-29
Supporting illegal whitespace characters in chunk size #1812

SakulK:


akka-http-core/src/test/scala/akka/http/impl/engine/parsing/RequestParserSpec.scala
```diff
@@ -255,6 +255,16 @@ abstract class RequestParserSpec(mode: String, newLine: String) extends FreeSpec
...
+      "with incorrect but harmless whitespace after chunk size" in new Test {
+        Seq(
+          start,
+          """0  
+            |
+            |""") should generalMultiParseTo(
+            Right(baseRequest.withEntity(Chunked(`application/pdf`, source(LastChunk)))))
+        closeAfterResponseCompletion shouldEqual Seq(false)
+      }
+
```
ktoso: thanks for the test 👍

akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpMessageParser.scala
```diff
@@ -243,6 +243,7 @@ private[http] trait HttpMessageParser[Output >: MessageOutput <: ParserOutput] {
           case ';' if cursor > offset ⇒ parseChunkExtensions(size.toInt, cursor + 1)()
           case '\r' if cursor > offset && byteChar(input, cursor + 1) == '\n' ⇒ parseChunkBody(size.toInt, "", cursor + 2)
           case '\n' if cursor > offset ⇒ parseChunkBody(size.toInt, "", cursor + 1)
+          case c if CharacterClasses.WSP(c) ⇒ parseSize(cursor + 1, size)
```
ktoso: perhaps we should add some comment that we know this is illegal in the spec but we've seen it in the wild?

raboof:
OK TO TEST

akka-http-core/src/test/scala/akka/http/impl/engine/parsing/RequestParserSpec.scala
```diff
@@ -255,6 +255,16 @@ abstract class RequestParserSpec(mode: String, newLine: String) extends FreeSpec
         closeAfterResponseCompletion shouldEqual Seq(false)
       }
 
+      "with incorrect but harmless whitespace after chunk size" in new Test {
+        Seq(
+          start,
+          """0
```
jrudolph: Maybe encode the whitespace as `\u0020` so it's clearer.

jrudolph:
> The worst thing that could happen is the method spinning for a while so that's not nice

It's not really spinning, it's still parsing and making progress. The method itself is `@tailrec` so that's fine.

> If we're worried about that aspect we should hide this under a relaxed mode flag.

I'd say it's fine like this

ktoso:
Merge once validated

ktoso:
Damn those formatting failures...! Thanks @jrudolph for fixing those.
Ready to go then.

Thanks for the fix @SakulK !

ktoso:
Resolves #1812

SakulK:
Thanks for your help @ktoso @jrudolph :)

PR #1822, opened 2018-01-29
 =htc #380 TLS should close connection when instructed by stream completion

### LINKED ISSUE #380: HttpEntity.CloseDelimited toStrict times out (on the client-side) when using SSL

jrudolph:
Thanks @nodefactory-bk for the reproducer.

Fixes #380.


jrudolph:
Unrelated: header check failed.

Issue #1811, opened 2018-01-26
Ability to access current response timeout


kag0:
It's currently possible to modify the request timeout with the `withRequestTimeout` directive which internally uses the `Timeout-Access` header and `TimeoutAccess` trait. However, it's not possible to access the currently set request timeout. If we added a method like `getRequestTimeout` to the `TimeoutAccess` trait then it would be easy to add an `extractRequestTimeout` directive to retrieve that value.

[related SO](https://stackoverflow.com/questions/48433026/how-to-get-the-current-server-request-timeout-in-akka-http)

jrudolph:
Seems simple enough I agree.

kag0:
took a quick pass at a pr https://github.com/akka/akka-http/pull/1828

jlprat:
The corresponding PR is merged, so I guess we can close thin one.
And it's actually released under 10.1.0 if I'm not mistaken

jrudolph:
:+1:

Am 14.03.2018 19:49 schrieb "Josep Prat" <notifications@github.com>:

PR #1828, opened 2018-01-29
Ability to access current response timeout (#1811)

kag0:
The currently set response timeout for a request can now be accessed through the `TimeoutAccess` trait, or through directives.


akka-http/src/main/scala/akka/http/scaladsl/server/directives/TimeoutDirectives.scala
```diff
@@ -16,6 +15,16 @@ import scala.concurrent.duration.Duration
  */
 trait TimeoutDirectives {
 
+  def extractRequestTimeout: Directive1[Duration] = Directive { inner ⇒ ctx ⇒
+    val timeout = ctx.request.header[`Timeout-Access`] match {
+      case Some(t) ⇒ t.timeoutAccess.getTimeout
+      case _ ⇒
+        ctx.log.warning("withRequestTimeout was used in route however no request-timeout is set!")
+        Duration.Inf
```
kag0: Anyone have input on how this error case should be handled?

akka-http/src/main/scala/akka/http/scaladsl/server/directives/TimeoutDirectives.scala
```diff
@@ -16,6 +15,16 @@ import scala.concurrent.duration.Duration
  */
 trait TimeoutDirectives {
 
+  def extractRequestTimeout: Directive1[Duration] = Directive { inner ⇒ ctx ⇒
+    val timeout = ctx.request.header[`Timeout-Access`] match {
+      case Some(t) ⇒ t.timeoutAccess.getTimeout
+      case _ ⇒
+        ctx.log.warning("extractRequestTimeout was used in route however no request-timeout is set!")
+        Duration.Inf
```
kag0: Anyone have input on how this error case should be handled?

ktoso:
OK TO TEST

akka-http-core/src/main/scala/akka/http/scaladsl/TimeoutAccess.scala
```diff
@@ -12,6 +12,16 @@ import akka.http.scaladsl.model.{ HttpResponse, HttpRequest }
...
 
+  /**
+   * Returns the currently set timeout.
+   * The timeout period is measured as of the point in time that the end of the request has been received,
+   * which may be in the past or in the future!
+   *
+   * Due to the inherent raciness it is not guaranteed that the returned timeout was applied before
+   * the previously set timeout has expired!
+   */
+  def getTimeout: Duration
```
jrudolph: Let's call it just `timeout` for scala.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/TimeoutDirectives.scala
jrudolph: `Duration.Inf` is fine, no need to log a warning. Let's document it on the directive.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/TimeoutDirectives.scala
```diff
@@ -16,6 +15,16 @@ import scala.concurrent.duration.Duration
  */
 trait TimeoutDirectives {
 
+  def extractRequestTimeout: Directive1[Duration] = Directive { inner ⇒ ctx ⇒
```
jrudolph: Needs a documentation page in the docs and Scaladocs.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/TimeoutDirectives.scala
kag0: Is it a common case that the timeout access header is missing? What situation should we tell users to expect it?

kag0:
Just a quick bump on this @jrudolph.
Is there anything I can do to improve here? Looks like the jenkins validation raised issues with binary compatibility. Any pointers on how to resolve that?

ktoso:
How to handle binary compatibility is explained here in the [CONTRIBUTING.md](https://github.com/akka/akka-http/blob/ad113f8a83d00edc6333202f074148ef9e8c54ea/CONTRIBUTING.md#binary-compatibility) :) I'll have a look and advice what to do in this case.

docs/src/test/java/docs/http/javadsl/server/directives/TimeoutDirectivesExamplesTest.java
```diff
@@ -177,4 +177,31 @@ public void testRequestTimeoutCustomResponseCanBeAddedSeparately() {
         assert (StatusCodes.ENHANCE_YOUR_CALM.equals(statusCode));
         //#withRequestTimeoutResponse
     }
+
+    @Test
+    public void extractRequestTimeout(){
+        //#extractRequestTimeout
+        Duration timeout1 = Duration.create(500, TimeUnit.MILLISECONDS);
+        Duration timeout2 = Duration.create(1000, TimeUnit.MILLISECONDS);
+        Route route =
```
jrudolph: We use 2 spaces as indentation for Java doc snippets as well for better readability.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/TimeoutDirectives.scala
jrudolph: Ah, good point. I think there's currently no way the header would be missing. Maybe we can just fail in that case because it should be seldom enough?

kag0: I think failing or continuing with `Duration.Inf` would both be fine, maybe bump the log to error level.
What would failure look like here? Completing with an exception?

jrudolph: Maybe, let's keep it for now. We might want to allow switching off request timeouts completely for performance reasons later on, so not requiring that this header is always available is maybe the best solution.

jrudolph:
Fixed Mima issues.

PR #1834, opened 2018-02-02
=cache remove `asScala` usage that fails in doc generation

jrudolph:


raboof:
How can I reproduce this failing doc generation? `sbt -Dakka.genjavadoc.enabled=true javaunidoc:doc` and `sbt clean paradox` on master didn't fail for me

jrudolph:
It was `scalaunidoc:doc` that failed before.

akka-http-caching/src/main/scala/akka/http/caching/javadsl/CachingSettings.scala
```diff
@@ -21,7 +20,7 @@ abstract class CachingSettings private[http] () { self: CachingSettingsImpl ⇒
   def lfuCacheSettings: LfuCacheSettings
 
   def withLfuCacheSettings(newSettings: LfuCacheSettings): CachingSettings =
-    self.copy(lfuCacheSettings = newSettings.asScala)
+    self.copy(lfuCacheSettings = newSettings.asInstanceOf[LfuCacheSettingsImpl]) // crashes with `asScala` in doc generation
```
raboof: I think we've seen this before - another solution is to move the two imports inside this function, so:
```
  def withLfuCacheSettings(newSettings: LfuCacheSettings): CachingSettings = {
    import akka.http.caching.CacheJavaMapping.Implicits._

    self.copy(lfuCacheSettings = newSettings.asScala)
  }
```
Just checked, this works and is perhaps slightly neater. Weird though.

jrudolph: Nice, I had a try to cleanup and simplify JavaMapping a bit in the last days, and I found also lots of weird issues. It's a bit crazy how you stop trying to debug those implicit issues at some point and just try around until it works...

jrudolph: Will change it to your suggestions.

PR #1839, opened 2018-02-07
=ht2 fix failure in H2SpecIntegration

jrudolph:


akka-http2-support/src/test/scala/akka/http/h2spec/H2SpecIntegrationSpec.scala
```diff
@@ -143,8 +143,4 @@ class H2SpecIntegrationSpec extends AkkaSpec(
     def executable =
       System.getProperty("h2spec.path").ensuring(_ != null, "h2spec.path property not defined")
   }
-
-  override protected def afterTermination(): Unit = {
-    binding.unbind().futureValue
-  }
```
ktoso: hmm why? we just rely on the system being killed I guess now then?

jrudolph: Sorry, didn't explain anywhere. It seems that the unbind may fail because the actorsystem is already closed down when `afterTermination` is called which can make the unbind call fail.

ktoso: I see, thanks!

PR #1844, opened 2018-02-08
add grpc content type / media type

ktoso:


akka-http-core/src/main/scala/akka/http/scaladsl/model/MediaType.scala
```diff
@@ -334,6 +334,8 @@ object MediaTypes extends ObjectRegistry[(String, String), MediaType] {
   val `application/javascript`                                                    = awoc("javascript", "js")
   val `application/json`                                                          = awfc("json", HttpCharsets.`UTF-8`, "json")
   val `application/json-patch+json`                                               = awfc("json-patch+json", HttpCharsets.`UTF-8`)
+  val `application/grpc+proto`                                                    = abin("grpc+proto", NotCompressible)
+  val `application/grpc+json`                                                     = abin("grpc+json", NotCompressible)
```
jrudolph: Has the json support anywhere been specified yet? Is it really a binary, incompressible one? Binary might be ok, because I guess the framing would still be binary? Should we really include it already if it isn't clear how it will be specified?

ktoso:
good point, will remove the json one. it's not entirely clear yet

Issue #1575, opened 2017-11-29
Memory leak in RequestTimeoutSupport with 2.5.6


johanandren:
Request timeout support creates one `AsyncCallback` per request. Pre 2.5.6 this was fine, but with the new `invokeWithFeedback` infrastructure in place it now keeps a reference to each `AsyncCallback`.  akka/akka#24046 introduces a way to release the callback, but I think we'll need to find a way to not create multiple callbacks here to not force a dependency on Akka 2.5.7.

Perhaps the callback can be created once in `RequestTimeoutSupport` and reused throughout its life?

johanandren:
With the current iteration of the fix there is no memory leak, but I think we could still avoid allocating one callback per request and create that once per materialization instead.

ktoso:
> avoid allocating one callback per request and create that once per materialization instead.

Solved in   #1848

PR #1848, opened 2018-02-09
Removing creation of callback per grabbed request

kelebra:
Moved creation of `AsyncCallback` one level up so that it won't be created each time `grab(requestIn)` is invoked.

Addresses comment on issue #1575


raboof:
OK TO TEST

ktoso:
Thanks!

Issue #1843, opened 2018-02-08
Support trailing request and response headers in HTTP/2


jrudolph:
Challenges:

 * incoming request: Http/2 allows trailers on entities with or without content length set, however, our model only allows trailers in Chunked requests (potential solution: always throw away length info and create chunked ones to be prepared for trailers?)
 * `Http2SubStream` does not support trailers so far in its model which models entities as `Source[ByteString]`

jrudolph:
The problem that we cannot support content-length and trailers at the same time can maybe solved differently for requests and responses:

 * for requests, we can choose not to change trailer behavior for now (i.e. keep ignoring them) since grpc does not require request trailers
 * for responses, make sure that we support trailers when an `HttpEntity.Chunked` is given, for that we need to change `Http2SubStream` (or create another variant) to support `ChunkStreamPart` elements instead of plain `ByteString`

Changes are necessary
 * in `ResponseRendering` to pass the `ChunkStreamPart` to the `Http2SubStream`
 * in `Http2Multiplexer` to actually produce `ParsedHeadersFrame` for `LastChunk` if trailers are defined (Or should frame generation live in `ResponseRendering` as well and `Http2SubStream` should deliver frames directly? Anyway it needs to work with buffer management in `Http2Multiplexer`.)

raboof:
Sending headers is done in #1857, let's track receiving headers in #1869

PR #1857, opened 2018-02-15
Support for trailing response headers in HTTP/2

raboof:
Refs #1843

### DIFF
--- a/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/FrameEvent.scala
+++ b/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/FrameEvent.scala
@@ -16,6 +16,7 @@ sealed trait FrameEvent { self: Product ⇒
 }
 sealed trait StreamFrameEvent extends FrameEvent { self: Product ⇒
   def streamId: Int
+  def endStream: Boolean
 }
 
 final case class GoAwayFrame(lastStreamId: Int, errorCode: ErrorCode, debug: ByteString = ByteString.empty) extends FrameEvent {
@@ -45,15 +46,21 @@ final case class HeadersFrame(
 final case class ContinuationFrame(
   streamId:   Int,
   endHeaders: Boolean,
-  payload:    ByteString) extends StreamFrameEvent
+  payload:    ByteString) extends StreamFrameEvent {
+  def endStream = false
+}
 
 case class PushPromiseFrame(
   streamId:            Int,
   endHeaders:          Boolean,
   promisedStreamId:    Int,
-  headerBlockFragment: ByteString) extends StreamFrameEvent
+  headerBlockFragment: ByteString) extends StreamFrameEvent {
+  def endStream = false
+}
 
-final case class RstStreamFrame(streamId: Int, errorCode: ErrorCode) extends StreamFrameEvent
+final case class RstStreamFrame(streamId: Int, errorCode: ErrorCode) extends StreamFrameEvent {
+  def endStream = false
+}
 final case class SettingsFrame(settings: immutable.Seq[Setting]) extends FrameEvent
 final case class SettingsAckFrame(acked: immutable.Seq[Setting]) extends FrameEvent
 
@@ -62,13 +69,17 @@ case class PingFrame(ack: Boolean, data: ByteString) extends FrameEvent {
 }
 final case class WindowUpdateFrame(
   streamId:            Int,
-  windowSizeIncrement: Int) extends StreamFrameEvent
+  windowSizeIncrement: Int) extends StreamFrameEvent {
+  def endStream = false
+}
 
 final case class PriorityFrame(
   streamId:         Int,
   exclusiveFlag:    Boolean,
   streamDependency: Int,
-  weight:           Int) extends StreamFrameEvent
+  weight:           Int) extends StreamFrameEvent {
+  def endStream = false
+}
 
 final case class Setting(
   identifier: SettingIdentifier,
@@ -84,7 +95,9 @@ final case class UnknownFrameEvent(
   tpe:      FrameType,
   flags:    ByteFlag,
   streamId: Int,
-  payload:  ByteString) extends StreamFrameEvent
+  payload:  ByteString) extends StreamFrameEvent {
+  def endStream = false
+}
 
 final case class ParsedHeadersFrame(
   streamId:      Int,
--- a/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/FrameEvent.scala
+++ b/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/FrameEvent.scala
@@ -16,7 +16,6 @@ sealed trait FrameEvent { self: Product ⇒
 }
 sealed trait StreamFrameEvent extends FrameEvent { self: Product ⇒
   def streamId: Int
-  def endStream: Boolean
 }
 
 final case class GoAwayFrame(lastStreamId: Int, errorCode: ErrorCode, debug: ByteString = ByteString.empty) extends FrameEvent {
@@ -46,21 +45,14 @@ final case class HeadersFrame(
 final case class ContinuationFrame(
   streamId:   Int,
   endHeaders: Boolean,
-  payload:    ByteString) extends StreamFrameEvent {
-  def endStream = false
-}
-
+  payload:    ByteString) extends StreamFrameEvent
 case class PushPromiseFrame(
   streamId:            Int,
   endHeaders:          Boolean,
   promisedStreamId:    Int,
-  headerBlockFragment: ByteString) extends StreamFrameEvent {
-  def endStream = false
-}
+  headerBlockFragment: ByteString) extends StreamFrameEvent
 
-final case class RstStreamFrame(streamId: Int, errorCode: ErrorCode) extends StreamFrameEvent {
-  def endStream = false
-}
+final case class RstStreamFrame(streamId: Int, errorCode: ErrorCode) extends StreamFrameEvent
 final case class SettingsFrame(settings: immutable.Seq[Setting]) extends FrameEvent
 final case class SettingsAckFrame(acked: immutable.Seq[Setting]) extends FrameEvent
 
@@ -69,17 +61,13 @@ case class PingFrame(ack: Boolean, data: ByteString) extends FrameEvent {
 }
 final case class WindowUpdateFrame(
   streamId:            Int,
-  windowSizeIncrement: Int) extends StreamFrameEvent {
-  def endStream = false
-}
+  windowSizeIncrement: Int) extends StreamFrameEvent
 
 final case class PriorityFrame(
   streamId:         Int,
   exclusiveFlag:    Boolean,
   streamDependency: Int,
-  weight:           Int) extends StreamFrameEvent {
-  def endStream = false
-}
+  weight:           Int) extends StreamFrameEvent
 
 final case class Setting(
   identifier: SettingIdentifier,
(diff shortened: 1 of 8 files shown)


akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Multiplexer.scala
```diff
@@ -108,8 +124,14 @@ private[http2] trait Http2MultiplexerSupport { logic: GraphStageLogic with Stage
         def bufferedBytes: Int = buffer.size
 
         override def onPush(): Unit = {
-          val newData = inlet.grab()
-          buffer ++= newData
+          inlet.grab() match {
```
raboof: (this is not very nice, I'm going to see if subclassing might help here)

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/FrameEvent.scala
```diff
@@ -16,6 +16,7 @@ sealed trait FrameEvent { self: Product ⇒
 }
 sealed trait StreamFrameEvent extends FrameEvent { self: Product ⇒
   def streamId: Int
+  def endStream: Boolean
```
jrudolph: Doesn't seem to be used?

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
```diff
@@ -21,9 +21,25 @@ import scala.concurrent.{ ExecutionContext, Future }
 /**
  * Represents one direction of an Http2 substream.
  */
-private[http2] final case class Http2SubStream(
+private[http2] sealed trait Http2SubStream {
+  type T
```
jrudolph: I think the type can be removed if you make it a `val data: Source[_, Any]` below. I think the generic part is only used for cancellation.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
```diff
@@ -21,9 +21,25 @@ import scala.concurrent.{ ExecutionContext, Future }
...
+  val data: Source[T, Any]
+  def streamId: Int
+}
+
+private[http2] final case class ByteHttp2SubStream(
+  initialHeaders: ParsedHeadersFrame,
+  data:           Source[ByteString, Any]
+) extends Http2SubStream {
+  type T = ByteString
+  def streamId: Int = initialHeaders.streamId
```
jrudolph: `streamId` can be implemented in `Http2SubStream`.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Multiplexer.scala
```diff
@@ -43,45 +45,57 @@ private[http2] trait Http2MultiplexerSupport { logic: GraphStageLogic with Stage
     new Http2Multiplexer with OutHandler with StateTimingSupport with LogSupport {
       outlet.setHandler(this)
 
-      class OutStream(
+      class OutStream[T](
         val streamId:           Int,
-        private var maybeInlet: Option[SubSinkInlet[ByteString]],
+        private var maybeInlet: Option[SubSinkInlet[T]],
```
jrudolph: I think the extra type parameter can be removed if you use `SubSinkInlet[_]` here. We match on the runtime time of the elements anyway below.

raboof: Yeah might be nice to try out both variants (I somewhat avoided the runtime match in #1858)

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Multiplexer.scala
```diff
@@ -108,8 +123,14 @@ private[http2] trait Http2MultiplexerSupport { logic: GraphStageLogic with Stage
...
         override def onPush(): Unit = {
-          val newData = inlet.grab()
-          buffer ++= newData
+          inlet.grab() match {
+            case newData: ByteString          ⇒ buffer ++= newData
+            case HttpEntity.Chunk(newData, _) ⇒ buffer ++= newData
+            case HttpEntity.LastChunk(_, headers) ⇒
+              val headerPairs = new VectorBuilder[(String, String)]()
+              ResponseRendering.renderHeaders(headers, headerPairs, None, log)
+              trailer = Some(ParsedHeadersFrame(streamId, endStream = true, headerPairs.result(), None))
```
jrudolph: Maybe add another method for that in `ResponseRendering` that directly renders a set of headers into a `ParsedHeadersFrame`.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Multiplexer.scala
```diff
@@ -154,9 +175,16 @@ private[http2] trait Http2MultiplexerSupport { logic: GraphStageLogic with Stage
             info.closeStream()
             outStreams.remove(sub.streamId)
           } else {
-            val subIn = new SubSinkInlet[ByteString](s"substream-in-${sub.streamId}")
-            info.registerIncomingData(subIn)
-            sub.data.runWith(subIn.sink)(subFusingMaterializer)
+            sub match {
```
jrudolph: When the type params are gone, I think no match is even necessary any more. (I think there's still value in keeping the two subclasses of Http2SubStream).

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Multiplexer.scala
```diff
@@ -244,15 +272,18 @@ private[http2] trait Http2MultiplexerSupport { logic: GraphStageLogic with Stage
...
+        def enqueueOutStream(outStream: OutStream[_]): Unit =
           if (connectionWindowLeft == 0) become(WaitingForConnectionWindow(immutable.TreeSet(outStream.streamId)))
           else {
             require(outStream.canSend)
 
             val maxBytesToSend = currentMaxFrameSize min connectionWindowLeft
             val frame = outStream.nextFrame(maxBytesToSend)
             outlet.push(frame)
-            connectionWindowLeft -= frame.payload.size
+            frame match {
```
jrudolph: I guess one consequence of putting it like this is that we will wait for space in the window before sending the trailing `HEADER` frame while that wouldn't be necessary (because control frames are not accounted for with windows).

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Multiplexer.scala
raboof: Hmm, I don't think `sub.data.runWith(subIn.sink)` will be allowed when `subIn` is a `SubSinkInlet[_]` and `data` is a `Source[_, Any]`, since the compiler has no way to verify those `_`'s line up, right?

jrudolph: This seems to work:

```scala
val subIn = new SubSinkInlet[Any](s"substream-in-${sub.streamId}")
info.registerIncomingData(subIn)
sub.data.runWith(subIn.sink)(subFusingMaterializer)
```

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Multiplexer.scala
```diff
@@ -291,15 +310,21 @@ private[http2] trait Http2MultiplexerSupport { logic: GraphStageLogic with Stage
...
-            // now empty
-            become(Idle)
-          else
-            become(WaitingForNetworkToSendData(sendableOutstreams.filterNot(_ == chosenId))) // TODO: use Set instead
+          frame match {
+            case DataFrame(_, _, payload) ⇒ connectionWindowLeft -= payload.size
+            case _                        ⇒
+          }
+
+          outStream.endStreamIfPossible() match {
```
jrudolph: I think that's a viable solution. Would be nice if the duplication with the above case could be removed.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
```diff
@@ -21,12 +21,22 @@ import scala.concurrent.{ ExecutionContext, Future }
...
  * Represents one direction of an Http2 substream.
  */
-private[http2] final case class Http2SubStream(
-  initialHeaders: ParsedHeadersFrame,
-  data:           Source[ByteString, Any]) {
+private[http2] sealed trait Http2SubStream {
+  val initialHeaders: ParsedHeadersFrame
+  val data: Source[_, Any]
   def streamId: Int = initialHeaders.streamId
 }
```
ktoso: Would personally prefer spelling out that `Any` but `_` can remain I guess

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Multiplexer.scala
```diff
@@ -45,45 +46,53 @@ private[http2] trait Http2MultiplexerSupport { logic: GraphStageLogic with Stage
...
-        private var maybeInlet: Option[SubSinkInlet[ByteString]],
+        private var maybeInlet: Option[SubSinkInlet[_]],
         var outboundWindowLeft: Int,
-        private var buffer:     ByteString                       = ByteString.empty,
-        var upstreamClosed:     Boolean                          = false,
-        var endStreamSent:      Boolean                          = false
+        private var buffer:     ByteString                 = ByteString.empty,
+        var upstreamClosed:     Boolean                    = false,
+        var endStreamSent:      Boolean                    = false,
+        var trailer:            Option[ParsedHeadersFrame] = None
```
ktoso: would at some point change all the Option's here to OptionVals, but keeping as is for now is fine

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/RequestParsing.scala
```diff
@@ -69,10 +69,12 @@ private[http2] object RequestParsing {
...
           if (tlsSessionInfoHeader.isDefined) headers += tlsSessionInfoHeader.get
 
-          val entity =
-            if (subStream.data == Source.empty || contentLength == 0) HttpEntity.Empty
-            else if (contentLength > 0) HttpEntity.Default(contentType, contentLength, subStream.data)
-            else HttpEntity.Chunked.fromData(contentType, subStream.data)
+          val entity = subStream match {
+            case s if s.data == Source.empty || contentLength == 0 ⇒ HttpEntity.Empty
+            case ByteHttp2SubStream(_, data) if contentLength > 0  ⇒ HttpEntity.Default(contentType, contentLength, data)
+            case ByteHttp2SubStream(_, data)                       ⇒ HttpEntity.Chunked.fromData(contentType, data)
```
jrudolph: Maybe add a comment `/* contentLength undefined */` so it's clear why this case may happen.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/ResponseRendering.scala
```diff
@@ -56,13 +56,31 @@ private[http2] object ResponseRendering {
 
       val headers = ParsedHeadersFrame(streamId, endStream = response.entity.isKnownEmpty, headerPairs.result(), None)
 
-      Http2SubStream(
-        headers,
-        response.entity.dataBytes
-      )
+      response.entity match {
+        case HttpEntity.Chunked(_, chunks) ⇒
```
jrudolph: Nitpick: the lines could be collapsed here.

jrudolph:
Done from my side. Please merge when you think it's ready, @raboof.

raboof:
:+1: I'm happy with it

jrudolph:
Yay! :fireworks:

PR #1858, opened 2018-02-15
Supporting trailing response headers in HTTP/2

raboof:
Variation on #1857, this time subclassing OutStream

### DIFF
--- a/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Multiplexer.scala
+++ b/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Multiplexer.scala
@@ -233,8 +233,8 @@ private[http2] trait Http2MultiplexerSupport { logic: GraphStageLogic with Stage
       private def getOrCreateStreamFor(stream: Http2SubStream): OutStream[_] = outStreams.get(stream.streamId) match {
         case None ⇒
           val newOne = stream match {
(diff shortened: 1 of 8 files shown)


akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Multiplexer.scala
```diff
@@ -132,17 +150,37 @@ private[http2] trait Http2MultiplexerSupport { logic: GraphStageLogic with Stage
           pushControlFrame(RstStreamFrame(streamId, Http2Protocol.ErrorCode.INTERNAL_ERROR))
         }
       }
+      private class ByteOutStream(streamId: Int, maybeInlet: Option[Http2MultiplexerSupport.this.SubSinkInlet[ByteString]], outboundWindowLeft: Int)
+        extends OutStream[ByteString](streamId, maybeInlet, outboundWindowLeft) {
+        override def onData(bytes: ByteString): ByteString = bytes
```
jrudolph: Interesting idea as well. Would be nice to get rid of the type parameter. Maybe it could work by referencing the inlet in the overridden method? Maybe as a 

```scala
override def grabBytes(): ByteString = grab(maybeInlet.get)
```

but not sure if it's worth it when the runtime match would probably work out good enough.

raboof:
(the additional safety doesn't seem to weigh up against the added complexity, so retiring this attempt in favor of #1857)

PR #1860, opened 2018-02-15
Allows Sec-WebSocket-Extensions in requests

### LINKED ISSUE #1859: Sec-WebSocket-Extensions not allowed in requests
From the [spec](https://tools.ietf.org/html/rfc6455#page-58):

>  The |Sec-WebSocket-Extensions| header field is used in the WebSocket
>    opening handshake.  It is initially sent from the client to the
>    server, and then subsequently sent from the server to the client, to
>    agree on a set of protocol-level extensions to use for the duration
>    of the connection.
> 
>    The |Sec-WebSocket-Extensions| header field MAY appear multiple times
>    in an HTTP request (which is logically the same as a single
>    |Sec-WebSocket-Extensions| header field that contains all values.
>    However, the |Sec-WebSocket-Extensions| header field MUST NOT appear
>    more than once in an HTTP response.

Lagom migrated to an Akka-HTTP based backend and recently (there's not deadline) WS connections couldn't be established. We noticed the following in the logs:

```
2018-02-15T18:49:11.508Z [warn] akka.actor.ActorSystemImpl [sourceThread=application-akka.actor.default-dispatcher-7, akkaSource=akka.actor.ActorSystemImpl(application), sourceActorSystem=application, akkaTimestamp=18:49:11.507UTC] - HTTP header 'Sec-WebSocket-Extensions: permessage-deflate; client_max_window_bits' is not allowed in requests
``` 

caused in [HttpRequestRenderFactory](https://github.com/akka/akka-http/blob/f3e83935ffc27bc94f586d9387b54b499c6250d4/akka-http-core/src/main/scala/akka/http/impl/engine/rendering/HttpRequestRendererFactory.scala#L97). The root cause seems to be that `Sec-WebSocket-Extensions` extends `ResponseHeader` instead of `RequestResponseHeader`.

ignasi35:
Fixes #1859

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
@@ -766,7 +766,7 @@ private[http] object `Sec-WebSocket-Extensions` extends ModeledCompanion[`Sec-We
  */
 @InternalApi
 private[http] final case class `Sec-WebSocket-Extensions`(extensions: immutable.Seq[WebSocketExtension])
-  extends ResponseHeader {
+  extends RequestResponseHeader {
   require(extensions.nonEmpty, "Sec-WebSocket-Extensions.extensions must not be empty")
   import `Sec-WebSocket-Extensions`.extensionsRenderer
   protected[http] def renderValue[R <: Rendering](r: R): r.type = r ~~ extensions
--- a/akka-http-core/src/test/scala/akka/http/scaladsl/model/headers/HeaderSpec.scala
+++ b/akka-http-core/src/test/scala/akka/http/scaladsl/model/headers/HeaderSpec.scala
@@ -7,7 +7,8 @@ package akka.http.scaladsl.model.headers
 import akka.http.impl.util._
 import org.scalatest._
 import java.net.InetAddress
-import akka.http.scaladsl.model._
+
+import akka.http.scaladsl.model.{ headers, _ }
 
 class HeaderSpec extends FreeSpec with Matchers {
   "ModeledCompanion should" - {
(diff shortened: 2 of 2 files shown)


akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
```diff
@@ -766,7 +766,7 @@ private[http] object `Sec-WebSocket-Extensions` extends ModeledCompanion[`Sec-We
  */
 @InternalApi
 private[http] final case class `Sec-WebSocket-Extensions`(extensions: immutable.Seq[WebSocketExtension])
-  extends ResponseHeader {
+  extends RequestResponseHeader {
```
ktoso: LGTM, quoting spec that @ignasi35 mentions in ticket for fellow reviewers:

`It is initially sent from the client to the server, and then subsequently sent from the server to the client, to agree on a set of protocol-level extensions to use for the duration of the connection. `

ignasi35: Also relevant from the spec but not tested or addressed in this PR:

> The |Sec-WebSocket-Extensions| header field MAY appear multiple times
> in an HTTP request (which is logically the same as a single
> |Sec-WebSocket-Extensions| header field that contains all values.
> However, the |Sec-WebSocket-Extensions| header field MUST NOT appear
> more than once in an HTTP response.

I made no effort in asserting that the header may appear several times and that values are preserved when that happens.

raboof:
PLS BUILD

raboof:
OK TO TEST

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
ktoso: I'm ok with not enforcing this strickly, would be costly on insertions

jrudolph:
LGTM. Thanks, @ignasi35.

TimMoore:
Would it be possible to backport this to 10.0.x?

ktoso:
ping @raboof in case I forget

ktoso:
cc @patriknw, just FYI

PR #1862, opened 2018-02-16
Update docs to reference new jetty-alpn version

raboof:


jrudolph:
Can we automate with paradox variables?

raboof:
Hmm, we could remove the duplication (the same version being mentioned twice), but not really beyond that I think?

ktoso:
> Can we automate with paradox variables?

Doing the same as we have with the akka version should be doable right; would be an improvement I guess just in the way of remembering we have that version in there somewhere even ;)

jrudolph:
> Hmm, we could remove the duplication (the same version being mentioned twice), but not really beyond that I think?

I meant feeding the version from our build file where we mention the version number for running the HTTP/2 tests. Wouldn't that be possible?

raboof:
Ah I didn't realize we use it in tests - I'll give it a shot!

Issue #1731, opened 2018-01-02
Add headers / getHeaders method to HttpMessage


jrudolph:
Should basically look and work like the existing optional-result `header` / `getHeader` but return a collection with all the headers of a class.

Make sure to adapt the examples / documentation added in https://github.com/akka/akka-http/pull/1719.

jlprat:
I'll take this issue

PR #1863, opened 2018-02-16
+htc Add headers and getHeaders method to HttpMessage #1731

jlprat:
Refs: #1731
Java implementation is duplicated but I can't make type checker happy otherwise
Update examples in docs

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
@@ -120,6 +120,14 @@ sealed trait HttpMessage extends jm.HttpMessage {
     }
   }
 
+  /** Returns all the headers of the given type **/
+  def headers[T >: Null <: jm.HttpHeader: ClassTag]: immutable.Seq[T] = {
+    val clazz = classTag[T].runtimeClass.asInstanceOf[Class[T]]
+    headers.collect {
+      case h if clazz.isInstance(h) ⇒ h.asInstanceOf[T]
+    }
+  }
+
   /**
    * Returns true if this message is an:
    *  - HttpRequest and the client does not want to reuse the connection after the response for this request has been received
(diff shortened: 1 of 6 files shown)


akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
```diff
@@ -160,6 +168,12 @@ sealed trait HttpMessage extends jm.HttpMessage {
       case _                 ⇒ Optional.empty()
     }
   /** Java API */
+  def getHeaders[T <: jm.HttpHeader](headerClass: Class[T]): JIterable[T] = {
+    headers.collect {
+      case h if headerClass.isInstance(h) ⇒ h.asInstanceOf[T]
+    }.asJava
+  }
```
jlprat: As I mentioned, I'm not really happy to duplicate this bit. But I can't make compiler happy if I just call `headers[T]` as the type T is slightly different on the Scala counterpart.

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
```diff
@@ -120,6 +120,14 @@ sealed trait HttpMessage extends jm.HttpMessage {
     }
   }
 
+  /** Returns all the headers of the given type **/
+  def headers[T >: Null <: jm.HttpHeader: ClassTag]: immutable.Seq[T] = {
```
jrudolph: Here's a simpler implementation:

```
  /** Returns all the headers of the given type **/
  def headers[T <: jm.HttpHeader: ClassTag]: immutable.Seq[T] =
    headers.collect { case h: T ⇒ h }
```

The `>: Null` type bound is only needed for `header` because of its (arguably useful) usage of `OptionVal`.

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
jrudolph: With the `>: Null` type bound gone, this can be:

```scala
  /** Java API */
  def getHeaders[T <: jm.HttpHeader](headerClass: Class[T]): JIterable[T] =
    headers[T](ClassTag[T](headerClass)).asJava
```

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
jlprat: Aha!! Thanks for that, I wasn't sure why the need for `>: Null`. I'll fix it later :)

jrudolph:
I also suggested a change for the body of the Scala method btw. ;)

jlprat:
Ops, missed that one. Will happen in an hour or so 🙂

jlprat:
Done :)

jrudolph:
Thanks, @jlprat!

PR #1864, opened 2018-02-17
[WIP] Fix Uri.apply didn't normalize port (#1763)

### LINKED ISSUE #1763: Potential inconsistency in Uri API
The different smart constructors to create an `akka.http.scaladsl.model.Uri` have a different behavior regarding the port normalization.

The `apply` methods don't normalize the `Authority` port, unlike the `from` method:
```scala
scala> import akka.http.scaladsl.model.Uri
import akka.http.scaladsl.model.Uri

scala> Uri("http://www.google.fr:80")
res0: akka.http.scaladsl.model.Uri = http://www.google.fr:80

scala> Uri.from(scheme = "http", host = "www.google.fr", port= 80)
res1: akka.http.scaladsl.model.Uri = http://www.google.fr
```
As a side note `Spray` was normalizing the port on all `Uri` smart constructors.

jiminhsieh:
Fix #1763.
This commit should fix the example of #1763. I am still investigating other `apply` constructors, so I labeled it as WIP. This issue somehow related to https://github.com/akka/akka/issues/19650.

### DIFF
--- a/akka-http-core/src/test/scala/akka/http/scaladsl/model/UriSpec.scala
+++ b/akka-http-core/src/test/scala/akka/http/scaladsl/model/UriSpec.scala
@@ -500,7 +500,7 @@ class UriSpec extends WordSpec with Matchers {
 
       // empty host
       Uri("http://:8000/foo") shouldEqual Uri("http", Authority(Host.Empty, 8000), Path / "foo")
-      Uri("http://:80/foo") shouldEqual Uri("http", Authority(Host.Empty, 80), Path / "foo")
+      Uri("http://:80/foo") shouldEqual Uri("http", Authority(Host.Empty, 0), Path / "foo")
     }
 
     "properly complete a normalization cycle" in {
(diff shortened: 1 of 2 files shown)


jlprat:
OK TO TEST

akka-http-core/src/test/scala/akka/http/scaladsl/model/UriSpec.scala
```diff
@@ -500,7 +500,7 @@ class UriSpec extends WordSpec with Matchers {
 
       // empty host
       Uri("http://:8000/foo") shouldEqual Uri("http", Authority(Host.Empty, 8000), Path / "foo")
-      Uri("http://:80/foo") shouldEqual Uri("http", Authority(Host.Empty, 80), Path / "foo")
```
jiminhsieh: Although I changed this line to pass the unit test, I did have a question about this case. Should we expect the port number of  
```scala
Uri("http", Authority(Host.Empty, 80), Path / "foo")
```
to be normalize . In Spary, this would be normalized. Due to this issue (https://github.com/akka/akka/issues/19650) and this PR (https://github.com/akka/akka/pull/20490), this would not be normalized. In https://github.com/akka/akka/pull/20490, it removed this https://github.com/akka/akka/blob/v2.4.4/akka-http-core/src/main/scala/akka/http/scaladsl/model/Uri.scala#L222. The only solution I came up is restoring this line then make a similar method but without normalizing the port number for the method `copy`.

jrudolph: Thanks a lot, @jiminhsieh for going digging through the history. I think so far the changes are pretty straight-forward. So, let's merge this one and continue the discussion on a new PR/issue.

jrudolph: Btw I also wonder whether these tests shouldn't actually return a different result:

https://github.com/akka/akka-http/blob/944e19475450afcbefbc7dac9914aa7d0470526f/akka-http-core/src/test/scala/akka/http/scaladsl/model/UriSpec.scala#L762-L765

i.e. partly revert the changes done back then.

PR #1872, opened 2018-02-20
=pro build with Scala 2.12 by default

jrudolph:
I think it's time to switch the default. I started changing the nightlies to use explicit versions when building.


jrudolph:
> Late LGTM; will we run jenkins on 2.11 though?

The nightlies will also run with 2.11.

ktoso:
Good, that sounds like a fair balance

PR #1877, opened 2018-02-21
Commonized code in PredefinedToResponseMarshallers

catalin-ursachi:
Addressed a FIXME in `PredefinedToResponseMarshallers`; plus some minor refactoring.


akka-http/src/main/scala/akka/http/scaladsl/marshalling/PredefinedToResponseMarshallers.scala
```diff
@@ -79,36 +79,12 @@ trait PredefinedToResponseMarshallers extends LowPriorityToResponseMarshallerImp
...
+  implicit def fromEntityStreamingSupportAndByteStringMarshaller[T, M](s: EntityStreamingSupport, m: ToByteStringMarshaller[T]): ToResponseMarshaller[Source[T, M]] =
+    fromEntityStreamingSupportAndByteStringMarshaller(null, s, m)
 
-          HttpResponse(entity = HttpEntity(s.contentType, marshalledElements))
-        }) :: Nil
-      }
-    }
-  }
+  implicit def fromEntityStreamingSupportAndByteStringMarshaller[T: ClassTag, M](implicit s: EntityStreamingSupport, m: ToByteStringMarshaller[T]): ToResponseMarshaller[Source[T, M]] =
+    fromEntityStreamingSupportAndByteStringSourceMarshaller(s, m.map(Source.single))
```
catalin-ursachi: Is this fine, or will wrapping each `ByteString` into a `Source.single`, only to be `mapConcat`ed immediately afterwards, add an overhead?

If it isn't fine, then we could at least commonize the "bestMarshallingPerElement" logic, and keep everything else duplicated.

akka-http/src/main/scala/akka/http/scaladsl/marshalling/PredefinedToResponseMarshallers.scala
```diff
@@ -117,18 +93,20 @@ trait LowPriorityToResponseMarshallerImplicits {
...
+  private[marshalling] def fromEntityStreamingSupportAndByteStringSourceMarshaller[T: ClassTag, M](s: EntityStreamingSupport, m: Marshaller[T, Source[ByteString, _]]): ToResponseMarshaller[Source[T, M]] = {
     Marshaller[Source[T, M], HttpResponse] { implicit ec ⇒ source ⇒
       FastFuture successful {
         Marshalling.WithFixedContentType(s.contentType, () ⇒ {
           val availableMarshallingsPerElement = source.mapAsync(1) { t ⇒ m(t)(ec) }
 
           // TODO optimise such that we pick the optimal marshalling only once (headAndTail needed?)
           // TODO, NOTE: this is somewhat duplicated from Marshal.scala it could be made DRYer
-          val bestMarshallingPerElement = availableMarshallingsPerElement mapConcat { marshallings ⇒
+          val bestMarshallingPerElement = availableMarshallingsPerElement map { marshallings ⇒
```
catalin-ursachi: The rest of the changes here should just simplify the code a bit (unless I missed something about why it was the way it was before).

akka-http/src/main/scala/akka/http/scaladsl/marshalling/PredefinedToResponseMarshallers.scala
```diff
@@ -137,18 +115,17 @@ trait LowPriorityToResponseMarshallerImplicits {
                   case Marshalling.WithFixedContentType(`best`, marshal) ⇒ marshal
                   case Marshalling.WithOpenCharset(`bestMT`, marshal)    ⇒ () ⇒ marshal(bestCS)
                 }
-            }).toList
+            }
 
             // TODO we could either special case for certrain known types,
             // or extend the entity support to be more advanced such that it would negotiate the element content type it
             // is able to render.
-            if (selectedMarshallings.isEmpty) throw new NoStrictlyCompatibleElementMarshallingAvailableException[T](s.contentType, marshallings)
```
catalin-ursachi: Was it intentional that this logic (introduced by https://github.com/akka/akka-http/commit/238e7fb) should apply to `fromEntityStreamingSupportAndEntityMarshaller` but not `fromEntityStreamingSupportAndByteStringMarshaller`?

jrudolph:
OK TO TEST

catalin-ursachi:
I imagine the test failure is unrelated?

raboof:
PLS BUILD

jrudolph:
@ktoso do you remember about this logic? Care to review?

ktoso:
Will review in my morning!

jrudolph:
Cool, thanks, @ktoso.

akka-http/src/main/scala/akka/http/scaladsl/marshalling/PredefinedToResponseMarshallers.scala
```diff
@@ -79,36 +79,12 @@ trait PredefinedToResponseMarshallers extends LowPriorityToResponseMarshallerImp
...
-                  case Marshalling.WithOpenCharset(`bestMT`, marshal)    ⇒ () ⇒ marshal(bestCS)
-                }
-            }).toList
-          }
-          val marshalledElements: Source[ByteString, M] =
-            bestMarshallingPerElement.map(_.apply()) // marshal!
-              .via(s.framingRenderer)
+  @deprecated("This method exists only for the purpose of binary compatibility, it used to be implicit.", "10.1.0")
+  def fromEntityStreamingSupportAndByteStringMarshaller[T, M](s: EntityStreamingSupport, m: ToByteStringMarshaller[T]): ToResponseMarshaller[Source[T, M]] =
+    fromEntityStreamingSupportAndByteStringMarshaller(null, s, m)
```
ktoso: hm not enjoying passing nulls around

ktoso: Isn't this calling itself and wound cause a n infinite recursion..?

akka-http/src/main/scala/akka/http/scaladsl/marshalling/PredefinedToResponseMarshallers.scala
ktoso: I think here it's likely fine... since it's going to materialize a stream in any case

akka-http/src/main/scala/akka/http/scaladsl/marshalling/PredefinedToResponseMarshallers.scala
ktoso: so that's the classtag... configusing to pass the null... would it now blow up for existing users?

catalin-ursachi: I believe it won't; the method that uses the class tag, `NoStrictlyCompatibleElementMarshallingAvailableException`, can cope with a null class tag (because the same overloading with `null` was used when `fromEntityStreamingSupportAndEntityMarshaller` was changed.

But I agree it's unideal; I could change the private method to take an `Option[ClassTag[T]]`?

catalin-ursachi: On reflection, I'm not entirely sure using an option would be a good idea; since the sole user of the `ClassTag[T]`, `NoStrictlyCompatibleElementMarshallingAvailableException`, was written to explicitly handle a `null` (see commit https://github.com/akka/akka-http/commit/238e7fb553948978ab9d6cac6aa5f164b62de66d).

Happy to make any chance you'd prefer, though.

catalin-ursachi:
@ktoso - any thoughts on what changes I should make, if any?

raboof:
PLS BUILD

ktoso:
I'm OK with this though not super sure it is worth the complexity;
2nd reviewer please call the shot on this one

akka-http/src/main/scala/akka/http/scaladsl/marshalling/PredefinedToResponseMarshallers.scala
raboof: This might have been influenced by the 'somewhat duplicated' code in Marshal.scala? Did you look at the test coverage for this feature, did it seem sufficient to give confidence the behavior didn't change?

akka-http/src/main/scala/akka/http/scaladsl/marshalling/PredefinedToResponseMarshallers.scala
raboof: The null is indeed a bit dirty but I guess acceptable as it's a 'feature' of `NoStrictlyCompatibleElementMarshallingAvailableException` anyway... might warrant a comment though? Do/can we have a test that checks this situation?

akka-http/src/main/scala/akka/http/scaladsl/marshalling/PredefinedToResponseMarshallers.scala
catalin-ursachi: As far as I can tell, this code is covered by `EntityStreamingSpec.scala`.

akka-http/src/main/scala/akka/http/scaladsl/marshalling/PredefinedToResponseMarshallers.scala
catalin-ursachi: I've added a test case (which also tests `NoStrictlyCompatibleElementMarshallingAvailableException` being thrown by `fromEntityStreamingSupportAndByteStringMarshaller`).

akka-http/src/main/scala/akka/http/scaladsl/marshalling/PredefinedToResponseMarshallers.scala
```diff
@@ -137,18 +115,17 @@ trait LowPriorityToResponseMarshallerImplicits {
...
                 }
-            }).toList
+            }
 
             // TODO we could either special case for certrain known types,
             // or extend the entity support to be more advanced such that it would negotiate the element content type it
             // is able to render.
-            if (selectedMarshallings.isEmpty) throw new NoStrictlyCompatibleElementMarshallingAvailableException[T](s.contentType, marshallings)
-            else selectedMarshallings
+            selectedMarshalling.getOrElse(throw new NoStrictlyCompatibleElementMarshallingAvailableException[T](s.contentType, marshallings))
```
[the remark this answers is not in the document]
catalin-ursachi: Well, now that it's an option it'd have to be:
```
if (selectedMarshalling.isDefined) {
  selectedMarshalling.get
} else {
  throw ...
}
```
Can change to that if you prefer, of course.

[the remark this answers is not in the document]
johanandren: I don't think the allocation overhead here matters enough.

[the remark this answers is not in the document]
catalin-ursachi: (Just out of curiosity, what do you mean here? Does the `Option` not get allocated unless we call `getOrElse` on it?)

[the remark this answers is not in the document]
johanandren: `getOrElse` causes a lambda allocation for the by-name parameter.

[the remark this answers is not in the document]
catalin-ursachi: Ah, got it; thanks ☺

raboof:
PLS BUILD

raboof:
Thanks for this contribution!

PR #1880, opened 2018-02-23
Require Akka 2.5.10 and use the new version checker utility

### LINKED ISSUE #1538: Fail fast on Akka version mismatch
When getting an older Akka version pulled into a project than what is supported by/works with Akka HTTP you get missing method errors etc but would be nice if we instead could check on load of the extension and fail fast with a nice error message saying "You have Akka 2.x.y in your class path but need at least Akka version 2.a.b".

johanandren:
Fixes #1538 

See issue for discussion if it should really be done.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
@@ -54,6 +54,7 @@ class HttpExt private[http] (private val config: Config)(implicit val system: Ex
   with DefaultSSLContextCreation {
 
   akka.http.Version.check(system.settings.config)
+  akka.AkkaVersion.require("akka-http", akka.http.Version.supportedAkkaVersion)
 
   import Http._
 
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -19,7 +19,6 @@ object Dependencies {
   val h2specUrl = s"https://github.com/summerwind/h2spec/releases/download/v${h2specVersion}/${h2specName}.zip"
   val alpnAgentVersion = "2.0.7"
 
-  lazy val akkaVersion = settingKey[String]("The version of Akka to use.")
   lazy val scalaTestVersion = settingKey[String]("The version of ScalaTest to use.")
   lazy val specs2Version = settingKey[String]("The version of Specs2 to use")
   lazy val scalaCheckVersion = settingKey[String]("The version of ScalaCheck to use.")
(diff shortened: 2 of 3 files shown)


johanandren:
Ach, some issue with the `@unidoc` directive. No idea why it hit me here.

johanandren:
@raboof Any clue why this happens? The paradox directive does not find `akka.util.ByteString` any more after the 2.5.10 upgrade (see the build fail). I have verified it is not among the class names passed to the directive.

raboof:
@johanandren the error is `Error writing content for page migration-guide/migration-from-spray.html: 2 matches found for ByteString, but not javadsl/scaladsl: akka.protobuf.ByteString, akka.util.ByteString`, which means that it did find `akka.util.ByteString`, but it now also finds `akka.protobuf.ByteString` and doesn't know which one to pick. 

@richardimaoka added searching by fqdn in https://github.com/akka/akka/pull/24563/files#diff-c858760d32fe74f2dcabf0ea7bc4aaecR32 , I guess we'd want to add that and use it here as well (until we move that directive to its own plugin).

jrudolph:
> @richardimaoka added searching by fqdn in https://github.com/akka/akka/pull/24563/files#diff-c858760d32fe74f2dcabf0ea7bc4aaecR32 , I guess we'd want to add that and use it here as well (until we move that directive to its own plugin).

Another solution for often used names like `ByteString` could either be an explicit configured disambiguation option ("ByteString should always mean akka.util.ByteString") or a manual exclusion list where we could put classes from "impl" packages or from "akka.proto".

johanandren:
Sorry, I was perhaps a bit unclear, in addition to the CI-server fail, I also tried with the fqcn `akka.utils.ByteString` but that is also not found, and when I look at the classnames passed to the paradox-extension-tag-tingie `akka.utils.ByteString` is not among them for some reason.

johanandren:
Haha, no sorry, now I misread _both_ of your comments. Just ignore me, thanks.

johanandren:
Something I'd like to add though, why did it stop working now all of a sudden, both of those two classes was surely on the class path before as well?

raboof:
Hmm, that's indeed strange..

johanandren:
Giving this one up for now, created a separate issue about upgrading Akka version and dealing with the directive.

raboof:
We didn't see this problem before because akka-stream didn't depend on akka-protobuf before, it does since https://github.com/akka/akka/pull/24230

PR #1894, opened 2018-03-02
Fix several warnings across the project

jlprat:
Removes unused private methods
Converts unnecessary `var` to `val`
Adds new overloads without unnecessary implicts
Removes unnecessary default values for parameters
Removes unnecessary `implicit val`s
Migrates deprecated methods to new ones

### DIFF
--- a/akka-http-testkit/src/main/scala/akka/http/scaladsl/testkit/RouteTest.scala
+++ b/akka-http-testkit/src/main/scala/akka/http/scaladsl/testkit/RouteTest.scala
@@ -146,7 +146,6 @@ trait RouteTest extends RequestBuilding with WSTestRequestBuilding with RouteTes
                                  executionContext: ExecutionContext,
                                  materializer:     Materializer,
                                  routingLog:       RoutingLog,
-                                 rejectionHandler: RejectionHandler = RejectionHandler.default,
                                  exceptionHandler: ExceptionHandler = null) =
       new TildeArrow[RequestContext, Future[RouteResult]] {
         type Out = RouteTestResult
--- a/akka-http/src/main/scala/akka/http/javadsl/server/Route.scala
+++ b/akka-http/src/main/scala/akka/http/javadsl/server/Route.scala
@@ -50,8 +50,42 @@ trait Route {
    *    will be already be handled using the implicitly given [[RejectionHandler]] and [[ExceptionHandler]] (or
    *    the default handlers if none are given or can be found implicitly).
    *  - Consequently, no route alternatives will be tried that were combined with this route.
+   * @deprecated Use the variant without [[Materializer]]
    */
+  @Deprecated
   def seal(system: ActorSystem, materializer: Materializer): Route
+
+  /**
+   * Seals a route by wrapping it with default exception handling and rejection conversion.
+   *
+   * A sealed route has these properties:
+   *  - The result of the route will always be a complete response, i.e. the result of the future is a
+   *    ``Success(RouteResult.Complete(response))``, never a failed future and never a rejected route. These
+   *    will be already be handled using the implicitly given [[RejectionHandler]] and [[ExceptionHandler]] (or
+   *    the default handlers if none are given or can be found implicitly).
+   *  - Consequently, no route alternatives will be tried that were combined with this route.
+   */
+  def seal(system: ActorSystem): Route
+
+  /**
+   * Seals a route by wrapping it with explicit exception handling and rejection conversion.
+   *
+   * A sealed route has these properties:
+   *  - The result of the route will always be a complete response, i.e. the result of the future is a
+   *    ``Success(RouteResult.Complete(response))``, never a failed future and never a rejected route. These
+   *    will be already be handled using the implicitly given [[RejectionHandler]] and [[ExceptionHandler]] (or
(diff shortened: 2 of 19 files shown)


akka-http-testkit/src/main/scala/akka/http/scaladsl/testkit/RouteTest.scala
```diff
@@ -146,7 +146,6 @@ trait RouteTest extends RequestBuilding with WSTestRequestBuilding with RouteTes
                                  executionContext: ExecutionContext,
                                  materializer:     Materializer,
                                  routingLog:       RoutingLog,
-                                 rejectionHandler: RejectionHandler = RejectionHandler.default,
```
jrudolph: Weird that it wouldn't picked up somewhere implicitly?

akka-http/src/main/scala/akka/http/javadsl/server/Route.scala
```diff
@@ -50,9 +50,23 @@ trait Route {
...
+   * Seals a route by wrapping it with default exception handling and rejection conversion.
+   *
+   * A sealed route has these properties:
+   *  - The result of the route will always be a complete response, i.e. the result of the future is a
+   *    ``Success(RouteResult.Complete(response))``, never a failed future and never a rejected route. These
+   *    will be already be handled using the implicitly given [[RejectionHandler]] and [[ExceptionHandler]] (or
+   *    the default handlers if none are given or can be found implicitly).
+   *  - Consequently, no route alternatives will be tried that were combined with this route.
+   */
+  def seal(system: ActorSystem): Route
```
jrudolph: Is that method really needed?

akka-http-testkit/src/main/scala/akka/http/scaladsl/testkit/RouteTest.scala
jrudolph: Ah, yes, probably not because rejections are handled specially.

akka-http/src/main/scala/akka/http/javadsl/server/Route.scala
jlprat: well, if the other one with an `ActorSystem` and `Materializer` was needed, this one is as well...
I kind of like this minimalist way to seal a route.
What's your idea? Just deprecate the one with `ActorSystem` and `Materializer`?

jrudolph: Yep, I don't think either one is necessary.

jlprat: Leaving the `seal` variant with `RoutingSettings`, `ParserSettings`, `RejectionHandler`, `ExceptionHandler` as the only way to do it, right?

jlprat: Then, we'll might want to update this doc:
https://doc.akka.io/docs/akka-http/current/routing-dsl/routes.html?language=java#sealing-a-route

jlprat: However, I still like to be able to seal the route in a concise way.

jrudolph: I yes right, can't we just remove the parameter?

jlprat: You mean remove without deprecating the old method?
If so, it's a non binary compatible change AFAIU

jrudolph: No, actually deprecating is the right step. However, instead of introducing a new `seal(ActorSystem)`, we can directly create this one:

```scala
  override def seal(): Route = RouteAdapter {
    // TODO: scaladsl.server.Route.seal shouldn't require an implicit RoutingSettings when it can retrieve it itself
    akka.http.scaladsl.server.directives.BasicDirectives.extractSettings { implicit settings ⇒
      scaladsl.server.Route.seal(delegate)
    }
  }
```

Can you create a ticket to fix the scaladsl `Route.seal` not to require implicit parameters any more?

jlprat: Ah, yes, good idea. I'll change the `seal` to have no parameters.

I will also create the ticket you mention :+1:

jrudolph: Cool, thanks a lot, @jlprat.

jlprat:
Applied feedback. I also updated the doc pages where the deprecated method was used with the new parameter-less variant.

jlprat:
Rebase done

docs/src/test/java/docs/http/javadsl/server/directives/SecurityDirectivesExamplesTest.java
```diff
@@ -222,7 +222,7 @@ public void testAuthenticateOrRejectWithChallenge() {
       authenticateOrRejectWithChallenge(myUserPassAuthenticator, userName ->
         complete("Authenticated!")
       )
-    ).seal(system(), materializer());
+    ).seal();
```
jrudolph: :+1: much nicer now ;)

PR #1897, opened 2018-03-05
Mark a lot more internal classes as such

jrudolph:
I went through the current [API docs](https://doc.akka.io/api/akka-http/10.1.0-RC2/akka/http/javadsl/index.html) which should exclude private stuff and so found a few more classes that should be marked as internal.

### DIFF
--- a/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/FrameEvent.scala
+++ b/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/FrameEvent.scala
@@ -4,6 +4,7 @@
 
 package akka.http.impl.engine.http2
 
+import akka.annotation.InternalApi
 import akka.http.impl.engine.http2.Http2Protocol.ErrorCode
 import akka.http.impl.engine.http2.Http2Protocol.FrameType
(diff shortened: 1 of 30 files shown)


build.sbt
```diff
@@ -56,7 +56,7 @@ lazy val root = Project(
   .disablePlugins(BintrayPlugin, MimaPlugin)
   .settings(
     // Unidoc doesn't like macros
-    unidocProjectExcludes := Seq(parsing),
+    unidocProjectExcludes := Seq(parsing, httpJmhBench),
```
ktoso: oh, yes good

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/FrameEvent.scala
```diff
@@ -4,90 +4,99 @@
...
+@InternalApi
+private[http2] sealed trait FrameEvent { self: Product ⇒
   def frameTypeName: String = productPrefix
 }
-sealed trait StreamFrameEvent extends FrameEvent { self: Product ⇒
-  def streamId: Int
-}
+/** INTERNAL API */
+@InternalApi
+private[http2] object FrameEvent {
```
ktoso: yes nice

Issue #1895, opened 2018-03-04
`@unidoc` links to Scala objects are broken in documentation


shkoder:
On https://doc.akka.io/docs/akka-http/current/common/caching.html#frequency-biased-lfu-cache links to `LfuCache` are broken for both Scala and Java - leading to 404

jrudolph:
Probably a problem in the unidoc directive when there's only an object but no companion class?

/cc @raboof

jrudolph:
I renamed the ticket since all links to objects seem to be broken.

raboof:
There seem to be a couple of problems here:

1) In Scaladoc, we link to the `LfuCache` class instead of the companion object, but that class is `private[caching]` and thus is not included in the scaladoc. We rely on https://github.com/lukehutch/fast-classpath-scanner to list the classes and this does not allow us to distinguish between objects and classes. Short term solution might be to allow linking to objects explicitly, with `LfuCache$`.

2) In Javadoc, `LfuCache` is not included at all

jrudolph:
I see. Thanks for the analysis, @raboof. So, it seems LfuCache is a particularly bad case. Would it work if we would move the `private[caching] class LfuCache` somewhere else?

jrudolph:
Should be fixed by #1903 and #1904.

PR #1903, opened 2018-03-06
Allow linking to case objects with `@unidoc`

raboof:
Fixes the scaladoc part of #1895. It would be even nicer to try and autodetect
whether we should link to the object, and produce a compile-time error when
linking to a class does not have a companion object or vice-versa, but it seems
this is not supported by fast-classpath-scanner at this point

### DIFF
--- a/project/ParadoxSupport.scala
+++ b/project/ParadoxSupport.scala
@@ -8,7 +8,6 @@ import java.io.{File, FileNotFoundException}
 
 import sbt._
 import Keys._
-import com.lightbend.paradox._
 import com.lightbend.paradox.markdown._
 import com.lightbend.paradox.sbt.ParadoxPlugin.autoImport._
 import org.pegdown.Printer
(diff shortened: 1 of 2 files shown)


project/ParadoxSupport.scala
```diff
@@ -59,8 +56,18 @@ object ParadoxSupport {
       }
     }
 
+    private def baseClassName(label: String) = {
+      val labelWithoutGenerics = label.split("\\[")(0)
+      if (labelWithoutGenerics.endsWith("$")) labelWithoutGenerics.init
```
jrudolph: Learned something here as I didn't know about `init` ;) Do you know where that terminology comes from?

raboof: Not sure where it *originally* came from, I knew of it from Haskell... (init/last are kind of the "ugly cousins" of head/tail I suppose :) )

jrudolph: Hehe, cool, thanks for the explanation :)

PR #1904, opened 2018-03-06
Add javadoc generation for akka-http-caching

raboof:
Fixes the 'other' half of #1895


jrudolph:
Why do have to do that there but not anywhere else? Wouldn't it be automatically enabled when `-Dakka.genjavadoc.enabled=true`?

raboof:
AFAICS we *do* have that everywhere else (for those subprojects that should be documented).

jrudolph:
Ah, how could I miss that?

PR #865, opened 2017-02-14
WIP - generate pull request validation report

jrudolph:

PR #1907, opened 2018-03-07
Make travis perform mimaValidation?

ktoso:
This way the mima checks are done in parallel

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ sudo: false
 
 scala:
   - "2.12.4"
+  - "2.11.12"
 
 jdk:
   - oraclejdk8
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -4,6 +4,7 @@ sudo: false
 
 scala:
   - "2.12.4"
+  - "2.11.12"
```
ktoso: This way the mima checks are done in parallel

ktoso: Though now I need the sbt task to pick that scala version... I don't want to do + since that'd duplicate work

.travis.yml
```diff
@@ -24,7 +25,7 @@ cache:
 
 script:
   # need to override as the default is to test
-  - sbt -J-XX:ReservedCodeCacheSize=256m +update
+  - sbt -J-XX:ReservedCodeCacheSize=256m update mimaReportBinaryIssues
```
ktoso: @2m how can I make sbt do `++{the-scalaversion}`? I want it to be done in parallel, which is why the setting above and not just + here

2m: Do `++$TRAVIS_SCALA_VERSION`

ktoso:
This is expected to fail, since there is an incompat on 2.12

ktoso:
And it did:

perfect: 

itunes

We could remove the mimacheck from the jenkins now, speeding up validation

jrudolph:
Should be fast enough on jenkins when things are already compiled and also it will integrate with #865 later so I'd say we leave it there for now.

PR #1908, opened 2018-03-07
!htc #56 change return type of unbind to Future[Done]

jrudolph:
Should be binary compatible as long as no one tried to operate on the
previous result value (Unit). The type might have bled into user code
so it might be source incompatible.


jlprat:
Cool, you can change this if you want as well: https://github.com/akka/akka-http/blob/master/akka-http/src/main/scala/akka/http/scaladsl/server/HttpApp.scala#L114
This extra conversion to `Done` wouldn't be needed anymore.

jrudolph:
> Cool, you can change this if you want as well

Good point :+1: done.

jrudolph:
I added a priliminary note to the release notes.

ktoso:
Okey :)

PR #1912, opened 2018-03-07
=pro also allow building for 2.13.0-M3

jrudolph:
This will allow running the nightlies for 2.13. Let's see if that helps for anything or if it just produces spurious failures.


jrudolph:
Just merging so I can trigger the nightly with 2.13.0-M3. If it doesn't work we can revert later.

PR #1917, opened 2018-03-08
=pro fix javadoc generation

jrudolph:
Yes, moving lines around is the only thing necessary here...


jrudolph:
Going ahead on this one.

PR #1919, opened 2018-03-08
[DO NOT MERGE] Additional fixes for allowing compilation with 2.13.0-M3

jrudolph:
These additional changes on top of #1918 shouldn't be merged, but can be used to release a version for 2.13.0-M3.

 * Eta Expansion fixes could be fine but are a bit ugly
 * Removing `@tailrec` to make the code compile but then tail optimization actually doesn't apply which might lead to stack overflows.


jrudolph:
Closing directly, but kept for reference.

jrudolph:
You can use these extra fixes if another release should be necessary with Scala 2.13.0-M3.

Issue #1922, opened 2018-03-08
Travis will try to release twice


jrudolph:
Because since https://github.com/akka/akka-http/commit/9a35a0734536daf4e0f602515d78a14793a583d8 two version numbers are now active on travis also for releasing which happens with `+ publish`. One solution could be to use `++$TRAVIS_SCALA_VERSION publish` in `.travis.yml` to publish for both Scala versions in parallel. One concern with this solution could be that we need to maintain the Scala versions in `.travis.yml` actively which seems somewhat error prone.

This time I just cancelled one of the two jobs.

raboof:
#1923 makes sure each build publishes for its respective scala version, that seems sufficient and the burden of keeping the list of scala versions up-to-date in the .travis.yml seems acceptable.

PR #1923, opened 2018-03-08
=pro Publish for 2.12.x and 2.11.x in parallel with travis

jrudolph:
This will ignore whatever Scala versions are defined in our sbt build.

Problem with the existing version was that it tried building with 2.13.0-M3 and also tried to run multiple deployment builds because of the added scala version (see #1922).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ script:
 deploy:
   provider: script
   script:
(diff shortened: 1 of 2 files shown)


jrudolph:
Merging directly as there's nothing to verify on jenkins.

PR #1925, opened 2018-03-09
add monad syntax and tests

winitzki:
Support the Scala `for`/`yield` syntax in route directives, simplify route code.

## Writing routes in `for`/`yield` syntax

The default Akka-HTTP style is to use nested functions for defining routes, for example:

```scala
val myRoute: Route = get {
  path("test" / Segment) { userId =>
    val user = userById(userId)
    extractLog { log =>
      extractUri { uri =>
        parameter('x, 'y, 'z) { (x, y, z) =>
          log.info(s"Got URI $uri with parameters $x, $y, $z") // log this before parsing JSON
          entity(as[MyCaseClass]) { input =>
            // `computeAsFuture` is our business logic function returning a `Future`
            onSuccess(computeAsFuture(input, user, x, y, z)) { result =>
              complete(MyResultCaseClass(result))
            }
          }
        }
      }
    }
  }
}
```

Each of the functions `get`, `path`, `extractLog`, `entity` etc. are defined by the Akka-HTTP library as values of type `Directive[L]`.
A `Directive[L]` has a `tapply()` method that takes as an argument a function of type `L => Route` and returns a `Route`.
Therefore, each directive in a route requires you to write one more nested function of type `L => Route`.

It also follows that a `Directive[L]` is equivalent to a value of type `(L => Route) => Route`, which is known in functional programming as the "continuation monad".
Being a monad, this type can be used in `for`/`yield` syntax, which is idiomatic in Scala, and avoids having to write several nested functions.
The only missing part is the methods `map`, `flatMap`, and `withFilter` as required by the `for`/`yield` syntax.

The class `Directive[L]` already has methods `tmap`, `tflatMap`, and `tfilter` but their type signatures are not suitable because of the type class constraint on `L` and because of using magnets.

This PR adds syntax helpers that enable the `for`/`yield` syntax for routes, so that the route above can be rewritten like this:
 
```scala
import DirectiveMonad._

val myRoute: Route = for {
  _         <- get
  userId    <- path("test" / Segment)
  user      =  userById(userId)
  log       <- extractLog
  uri       <- extractUri
  (x, y, z) <- parameter('x, 'y, 'z)
  _         =  log.info(s"Got URI $uri with parameters $x, $y, $z") // log this before parsing JSON
  input     <- entity(as[MyCaseClass])
  result    <- onSuccess(computeAsFuture(input, user, x, y, z))
} yield MyResultCaseClass(result) // no need for `complete()`
```

[All Akka-HTTP directives](https://doc.akka.io/docs/akka-http/current/routing-dsl/directives/alphabetically.html) can be used without any changes in this syntax.
Directive operations such as `&` can be used as well, in the right-hand side of `<-`.

The new syntax is added using a zero-allocation wrapper, `WrapDirective[L]`, which can be created out of a `Directive[L]` or a `Directive1[L]` automatically via an implicit conversion.
There are also implicit conversions from `WrapDirective[L]` back to `Directive[L]` and `Directive1[L]` for convenience.

Some comments:

- Directives that appear to have no arguments (such as `get`) need to be used with the syntax `_ <- get` (these directives are of type `Directive0` and have an argument of type `Unit`).
- Ordinary (non-directive) computations within the `for` block are written as `a = b` instead of `val a = b`.
- Computations returning `Unit`, such as `log.info(...)`, are written as `_ = log.info(...)`.
- Filtering conditions such as `if x.length > 0` are permitted within the `for` block as usual; the route will be rejected if the condition fails.
- To return a `complete(xyz)`, just write `yield xyz`, where `xyz` must be a marshallable value (a string, a case class, or an http response.)
- For convenience, `yield` can be also used on a `Route`, so that `yield complete(xyz)`, `yield reject(...)`, etc. are also supported.
- Directive methods, such as `recover()`, `&` etc., can be used on a `WrapDirective` (i.e. on the result of `for/yield`) as well.

The advantages of this syntax:

- the `for`/`yield` syntax is a standard, idiomatic Scala syntax for sequential computations
- code is easier to read and refactor because it is not deeply nested
- directives become composable in the usual way, as monadic values
- further enrichment can be done using monad transformers

Example of refactoring several routes with a common prefix:

```scala
val commonDirectives = for {
  _ <- get
  _ <- pathPrefix("my_app")
} yield ()

val route1: Route = ???
val route2: Route = ???

val route: Route = for {
  _ <- commonDirectives
} yield route1 ~ route2
```

### DIFF
--- a/akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/DirectiveMonadSpec.scala
+++ b/akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/DirectiveMonadSpec.scala
@@ -5,7 +5,7 @@
 package akka.http.scaladsl.server.directives
 
 import akka.http.scaladsl.model.StatusCodes
-import akka.http.scaladsl.server.{Directive, Directive1, Route, RoutingSpec, ValidationRejection}
+import akka.http.scaladsl.server.{ Directive, Directive1, Route, RoutingSpec, ValidationRejection }
 import akka.http.scaladsl.server.directives.DirectiveMonad._
 
 class DirectiveMonadSpec extends RoutingSpec {
--- a/akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/DirectiveMonadSpec.scala
+++ b/akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/DirectiveMonadSpec.scala
@@ -1,15 +1,15 @@
 package akka.http.scaladsl.server.directives
 
 import akka.http.scaladsl.server.directives.DirectiveMonad._
-import akka.http.scaladsl.server.{Route, RoutingSpec}
+import akka.http.scaladsl.server.{ Route, RoutingSpec }
 
 class DirectiveMonadSpec extends RoutingSpec {
 
   "extract method and 1 path segment and return string" in {
     val route: Route = for {
-      _ <- get
-      segment <- path("some" / Segment)
-      method <- extractMethod
+      _ ← get
+      segment ← path("some" / Segment)
+      method ← extractMethod
     } yield s"method: ${method.value}, segment: $segment"
     Get("/some/1") ~> route ~> check {
       responseAs[String] shouldEqual "method: GET, segment: 1"
@@ -18,9 +18,9 @@ class DirectiveMonadSpec extends RoutingSpec {
 
   "extract 2 path segments and return string" in {
     val route: Route = for {
-      _ <- get
-      _ <- pathPrefix("test")
-      (segment1, segment2) <- path("some" / Segment / Segment)
+      _ ← get
+      _ ← pathPrefix("test")
+      (segment1, segment2) ← path("some" / Segment / Segment)
     } yield {
       s"segment1: $segment1, segment2: $segment2"
     }
@@ -31,9 +31,9 @@ class DirectiveMonadSpec extends RoutingSpec {
 
   "extract 3 path segments and return string" in {
     val route: Route = for {
-      _ <- get
-      (segment1, segment2, segment3) <- path("test" / "some" / Segment / Segment / Segment)
-      method <- extractMethod
+      _ ← get
+      (segment1, segment2, segment3) ← path("test" / "some" / Segment / Segment / Segment)
+      method ← extractMethod
     } yield s"method: ${method.value}, segment1: $segment1, segment2: $segment2, segment3: $segment3"
     Get("/test/some/1/2/3") ~> route ~> check {
       responseAs[String] shouldEqual "method: GET, segment1: 1, segment2: 2, segment3: 3"
--- a/akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/DirectiveMonadSpec.scala
+++ b/akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/DirectiveMonadSpec.scala
@@ -1,0 +1,43 @@
+package akka.http.scaladsl.server.directives
+
+import akka.http.scaladsl.server.directives.DirectiveMonad._
+import akka.http.scaladsl.server.{Route, RoutingSpec}
+
+class DirectiveMonadSpec extends RoutingSpec {
+
+  "extract method and 1 path segment and return string" in {
+    val route: Route = for {
+      _ <- get
+      segment <- path("some" / Segment)
+      method <- extractMethod
+    } yield s"method: ${method.value}, segment: $segment"
+    Get("/some/1") ~> route ~> check {
+      responseAs[String] shouldEqual "method: GET, segment: 1"
+    }
+  }
+
+  "extract 2 path segments and return string" in {
+    val route: Route = for {
+      _ <- get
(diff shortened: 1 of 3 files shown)


ktoso:
Hmm, very interesting proposal, thanks!

I may not get to review it today but will soon

ktoso:
OK TO TEST

pocorall:
As I wrote in #1970, I hope to use `for` / `yield` in the middle of `Directive1` chaining. For example:

```scala
def optionalCookieSession: Directive1[Option[(Session, User)]] =
  (for {
    sessionId <- cookie(SessionCookieName).map(_.value)
    if sessionId != ""
    SessionSearchResult(Some(session)) <- onSuccess(loginService ? FindSessionById(sessionId))
    UserSearchResult(Some(user)) <- onSuccess(userService ? FindUserById(session.userId))
  } yield Some(session -> user))
    .recover(_ => provide(None))
```

I've tested the code above with this PR. But it shows the following error:

```
value recover is not a member of akka.http.scaladsl.server.directives.WrapDirective[Some[(com.xxx.Session, com.xxx.User)]]
 possible cause: maybe a semicolon is missing before `value recover'?
       .recover(_ => provide(None))
```

, clearly because of the return type of the `for` / `yield` is `WrapDirective`, not `Directive1`. Then, how about implement `map` and `withFilter` in `akka.http.scaladsl.server.Directive` directly?

winitzki:
I'll see if I can support the case described by @pocorall .

raboof:
Sounds good!

winitzki:
Sorry, I was swamped with work until now. Let me take up this PR once again and revise the code.

lightbend-cla-validator:
Hi @winitzki,

winitzki:
I agreed to the CLA.

winitzki:
In my view, routes should not necessarily be trees. The logic of a route is more or less linear: extract most of the code should be extracting various data from the request, then there might be some filtering based on validation, then there are calls to some functions to compute responses, and finally a response is returned. I think the API for routes and the code would have been a lot simpler, had Akka-HTTP used the monadic `for`/`yield` from the beginning. There would have been no need for `Directive0`, `Directive1`, Directive`, and no need for special-casing tuples of 1,2,3,..., elements.

ktoso:
Hi @winitzki,
I don't mean to be too dismisive of the PR, I think it's a great idea and showcase, but we have to weight it against the, what I call, "confusion factor" (for new people) i.e. "whoa there's so many APIs and styles, which one should I use".

> In my view, routes should not necessarily be trees.  The logic of a route is more or less linear: extract most of the code should be extracting various data from the request, then there might be some filtering based on validation, some logging or performance metrics, then there are calls to some functions to compute responses, and finally a response is returned. 

They are trees however, and what you explained is a single traversal of such tree -- an depth-first-ish (*-ish* since rejections) traversal of the routes tree is indeed what routes basically do at runtime. However, in face of rejections, the traversal is not as simple as you painted it here. Even a `path(...) { get {} ~ post {} }` relies on this mechanism, same as do top level multiple pathPrefix directives.

> had Akka-HTTP used the monadic for/yield from the beginning. There would have been no need for Directive0, Directive1, Directive`, and no need for special-casing tuples of 1,2,3,..., elements.

That could not have been the solution -- the reason we did so is in order to have feature parity with Java which is a hard, non negotiable, requirement for all of our APIs. An abstracted, using hlists, version of directives *existed* in Spray which became Akka HTTP, and one of the tradeoffs we had to make was this abstraction, among other things.

As for the general monad argument -- yes and no, I don't think it enchances composability over how routes already are able to compose. I do see that the API proposal can be nice, however the tradeoff is that we'd confuse newcomers even more, by providing yet another style inside the core library. This is not to say that the PR isn't cool and a great idea, I think it is, however I'd wage it against the increased "choices" issue which we continue to face on a day to day basis.

Having all that said... I wonder if you would be okey with releasing this directive as a separate library, and we would link to it in the official docs -- would this be a nice compromise? You'd also get exposure in the community that it's your cool monadic akka routes library etc :-)

winitzki:
I understand about the Java compatibility requirement. So perhaps another person in your team could look over this and weigh in, and if there are 3 people all saying that this functionality belongs outside the core library, so be it. I will make a small add-on library then.

patriknw:
Hi @winitzki 

This is really impressive work and I understand that some people will find it attractive to define the routes with for comprehension. However, we wouldn't like to add confusion about how to use the DSL by adding another way to define the same thing. Questions like "when to use what?". Therefore I suggest that you release this as a small external library and we would be happy to include a reference to it from the Akka HTTP documentation.

ktoso:
Hi there, seems we're in agreement that this is very cool but best not in the main library -- however we're happy to link to your project once you send us a link or PR such PR that would add such docs :-)

Closing for now, thanks and I hope we can get that separate mini project running :-)

winitzki:
Thank you for the useful input that helped me understand where this API should be going. We are already using this for our production code at my day job. Perhaps once we see a few more use cases that require more development, I'll make this into a separate library. Especially, I would like to figure out whether monad transformers could be harnessed to generate swagger annotations or other stuff automatically from the route code.

Issue #996, opened 2017-03-28
Allow client setting overrides for host / host-patterns


jrudolph:
It would be nice if settings could be customized per host. To achieve that a section in the configuration is created with per-host overrides that should be chosen over the global the defaults. When the client pool is used these overrides should be applied automatically.

Suggestion how it could look like:

```
akka.http.host-connection-pool.per-host-override {
  "akka.io" = { # can use same things as in global `host-connection-pool` section
    max-connections = 10
  }

 "*.example.com" = { # allow `*` to apply overrides for all subdomains
 }
}
```

Related to https://github.com/akka/akka-http/issues/995.

anilgursel:
One may also want to set different `akka.http.client.idle-timeout` (or other configuration) for different hosts.  How should the configuration be specified in that case?  I believe it would look like below:

```
akka.http { 
  host-connection-pool {
    max-connections = 20 # default for all other hosts
    per-host-override {
      "akka.io" = { # can use same things as in global `host-connection-pool` section
        max-connections = 10
      }
    }
  }

  client {
    idle-timeout = 20 s # default for all other hosts
    per-host-override {
      "akka.io" = { # can use same things as in global `host-connection-pool` section
        idle-timeout = 10 s
      }
    }
  }
} 
```

An alternative would be to specify host at the top level as below.  I believe it makes it easier to see what all overrides are there for a specific host:

```
akka.http {
  host-connection-pool.max-connections = 20
  client.idle-timeout = 20 s
} 

per-host-override."akka.io" = {
  akka.http {
    host-connection-pool.max-connections = 10
    client.idle-timeout = 10 s
  } 
}
```

jrudolph:
Closed in #2574.

PR #1926, opened 2018-03-09
Allow client setting overrides for host / host-patterns #996

Falmarri:
This change uses nulls as the default parameter for the connection pool
settings because it was extremely ugly to try to create a placeholder
object for the default connection pool that would then need to be
configured later.

### DIFF
--- a/akka-http-core/src/main/resources/reference.conf
+++ b/akka-http-core/src/main/resources/reference.conf
@@ -384,6 +384,20 @@ akka.http {
     client = {
       # no overrides by default, see `akka.http.client` for default values
     }
+
+    # Add client specific overrides to host-connectioin-pool. Allows exact and prefixed urls
+    # eg:
+    #
+    # "akka.io" = { # can use same things as in global `host-connection-pool` section
+    #   max-connections = 10
+    # }
+
+    # "*.example.com" = { # allow `*` to apply overrides for all subdomains, including `example.com`
+    # }
+    #
+    per-host-override {
+
+    }
   }
 
   # Modify to tweak default parsing settings.
--- a/akka-http-core/src/main/resources/reference.conf
+++ b/akka-http-core/src/main/resources/reference.conf
@@ -385,19 +385,18 @@ akka.http {
       # no overrides by default, see `akka.http.client` for default values
     }
 
-    # Add client specific overrides to host-connectioin-pool. Allows exact and prefixed urls
+    # Add client specific overrides to host-connection-pool. Allows exact and prefixed urls
     # eg:
-    #
-    # "akka.io" = { # can use same things as in global `host-connection-pool` section
-    #   max-connections = 10
-    # }
+    # per-host-override = [
+    #   "akka.io" = { # can use same things as in global `host-connection-pool` section
+    #     max-connections = 10
+    #   },
 
-    # "*.example.com" = { # allow `*` to apply overrides for all subdomains, including `example.com`
-    # }
-    #
-    per-host-override {
+    #   "*.example.com" = { # allow `*` to apply overrides for all subdomains, including `example.com`
+    #   }
+    #]
+    per-host-override = []
 
-    }
   }
 
   # Modify to tweak default parsing settings.
--- a/akka-http-core/src/main/scala/akka/http/impl/settings/ConnectionPoolSettingsImpl.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/settings/ConnectionPoolSettingsImpl.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2017-2018 Lightbend Inc. <https://www.lightbend.com>
+ * Copyright (C) 2017-2019 Lightbend Inc. <https://www.lightbend.com>
  */
 
 package akka.http.impl.settings
--- a/akka-http-core/src/main/scala/akka/http/impl/settings/ConnectionPoolSettingsImpl.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/settings/ConnectionPoolSettingsImpl.scala
@@ -6,7 +6,7 @@ package akka.http.impl.settings
 
 import akka.annotation.InternalApi
 import akka.http.impl.util.{ SettingsCompanion, _ }
-import akka.http.scaladsl.settings.{ ClientConnectionSettings, ConnectionPoolSettings, PoolImplementation }
+import akka.http.scaladsl.settings._
 import com.typesafe.config.Config
 
 import scala.collection.immutable
@@ -25,7 +25,8 @@ private[akka] final case class ConnectionPoolSettingsImpl(
   connectionSettings:                ClientConnectionSettings,
   poolImplementation:                PoolImplementation,
   responseEntitySubscriptionTimeout: Duration,
-  hostMap:                           immutable.Seq[(Regex, ConnectionPoolSettings)])
+  hostOverrides:                     immutable.Seq[(Regex, ConnectionPoolSettings)],
+  hostOverrideImpl:                  HostOverride)
   extends ConnectionPoolSettings {
 
   require(maxConnections > 0, "max-connections must be > 0")
@@ -40,7 +41,7 @@ private[akka] final case class ConnectionPoolSettingsImpl(
   override def productPrefix = "ConnectionPoolSettings"
 
   def withUpdatedConnectionSettings(f: ClientConnectionSettings ⇒ ClientConnectionSettings): ConnectionPoolSettingsImpl =
-    copy(connectionSettings = f(connectionSettings))
+    copy(connectionSettings = f(connectionSettings), hostOverrides = hostOverrides.map { case (k, v) ⇒ k -> v.withUpdatedConnectionSettings(f) })
 
   private def suggestPowerOfTwo(around: Int): String = {
     val firstBit = 31 - Integer.numberOfLeadingZeros(around)
@@ -94,7 +95,8 @@ object ConnectionPoolSettingsImpl extends SettingsCompanion[ConnectionPoolSettin
         case "new"    ⇒ PoolImplementation.New
       },
       c getPotentiallyInfiniteDuration "response-entity-subscription-timeout",
-      List.empty
+      List.empty,
+      NoOpHostOverride
     )
   }
 
--- a/akka-http-core/src/main/scala/akka/http/impl/settings/ConnectionPoolSettingsImpl.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/settings/ConnectionPoolSettingsImpl.scala
@@ -25,7 +25,7 @@ private[akka] final case class ConnectionPoolSettingsImpl(
   connectionSettings:                ClientConnectionSettings,
   poolImplementation:                PoolImplementation,
   responseEntitySubscriptionTimeout: Duration,
-  hostMap:                           immutable.Map[Regex, ConnectionPoolSettings])
+  hostMap:                           immutable.Seq[(Regex, ConnectionPoolSettings)])
   extends ConnectionPoolSettings {
 
   require(maxConnections > 0, "max-connections must be > 0")
@@ -94,7 +94,7 @@ object ConnectionPoolSettingsImpl extends SettingsCompanion[ConnectionPoolSettin
         case "new"    ⇒ PoolImplementation.New
       },
       c getPotentiallyInfiniteDuration "response-entity-subscription-timeout",
-      Map.empty
+      List.empty
     )
   }
 
--- a/akka-http-core/src/main/scala/akka/http/impl/settings/ConnectionPoolSettingsImpl.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/settings/ConnectionPoolSettingsImpl.scala
@@ -9,7 +9,9 @@ import akka.http.impl.util.{ SettingsCompanion, _ }
 import akka.http.scaladsl.settings.{ ClientConnectionSettings, ConnectionPoolSettings, PoolImplementation }
 import com.typesafe.config.Config
 
+import scala.collection.immutable
 import scala.concurrent.duration.Duration
+import scala.util.matching.Regex
 
 /** INTERNAL API */
 @InternalApi
@@ -22,7 +24,8 @@ private[akka] final case class ConnectionPoolSettingsImpl(
   idleTimeout:                       Duration,
   connectionSettings:                ClientConnectionSettings,
   poolImplementation:                PoolImplementation,
-  responseEntitySubscriptionTimeout: Duration)
+  responseEntitySubscriptionTimeout: Duration,
+  hostMap:                           immutable.Map[Regex, ConnectionPoolSettings])
   extends ConnectionPoolSettings {
 
   require(maxConnections > 0, "max-connections must be > 0")
@@ -50,6 +53,33 @@ private[akka] final case class ConnectionPoolSettingsImpl(
 }
 
 object ConnectionPoolSettingsImpl extends SettingsCompanion[ConnectionPoolSettingsImpl]("akka.http.host-connection-pool") {
+
+  private[akka] def hostRegex(pattern: String): Regex = {
+    val regexPattern = if (pattern.startsWith("regex:")) {
+      pattern.stripPrefix("regex:")
+    } else {
+      pattern.stripPrefix("glob:").map {
+        case '*'   ⇒ ".*"
+        case '?'   ⇒ "."
+        case '.'   ⇒ "\\."
+        case other ⇒ other.toString
+      }.mkString
+    }
+
+    // If the pattern starts with a wildcard, we want to match subdomains, as well as the raw domain
+    // so *.example.com should match www.example.com as well as example.com. So we replacing any leading
+    // (.*\.) pattern to allow for the beginning of the string, as well as an arbitrary subdomain. But it
+    // won't match something like thisexample.com
+    val p = if (regexPattern.startsWith(".*\\.")) {
+      s"(^|.*\\.)${regexPattern.drop(4)}"
+    } else {
+      regexPattern
+    }
+
+    p.r
+
+  }
+
   def fromSubConfig(root: Config, c: Config) = {
     new ConnectionPoolSettingsImpl(
       c getInt "max-connections",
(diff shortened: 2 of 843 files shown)


ktoso:
OK TO TEST

ktoso:
Refs #996  // please refer to the ticket in PR comments

akka-http-core/src/main/scala/akka/http/scaladsl/settings/ConnectionPoolSettings.scala
```diff
@@ -76,6 +76,35 @@ abstract class ConnectionPoolSettings extends js.ConnectionPoolSettings { self:
...
 
 object ConnectionPoolSettings extends SettingsCompanion[ConnectionPoolSettings] {
+
+  /**
+   * We want to support globbing on hostnames, but I didn't want to write an entire globbing engine. This assumes that
+   * *.host.name will match host.name (without the leading period) because that was most intuitive to me. I'm not a huge
+   * fan of this, open to suggestions in this PR
+   */
+  private[akka] def hostMatches(host: String, pattern: String): Boolean = {
+    pattern.replaceFirst("^*\\.", "*.?").replace(".", raw"\.").replace("*", ".*").r.pattern.matcher(host).matches()
```
ktoso: how about a simple `host endsWith myThingy` (so no * globbing but simple suffix matching)?

ktoso:
I like the idea, however please make sure that existing tests pass:

```
Test Result (51 failures / +51)

    akka.http.impl.engine.client.ClientCancellationSpec.Http client connections must support cancellation in pooled outgoing connection
    akka.http.impl.engine.client.LegacyConnectionPoolSpec.The single-request client infrastructure should transform absolute request URIs into relative URIs plus host header
    akka.http.impl.engine.client.LegacyConnectionPoolSpec.The single-request client infrastructure should support absolute request URIs without path component
    akka.http.impl.engine.client.LegacyConnectionPoolSpec.The single-request client infrastructure should support absolute request URIs with a double slash path component
    akka.http.impl.engine.client.LegacyConnectionPoolSpec.The single-request client infrastructure should produce an error if the request does not have an absolute URI
    akka.http.impl.engine.client.NewConnectionPoolSpec.The single-request client infrastructure should transform absolute request URIs into relative URIs plus host header
    akka.http.impl.engine.client.NewConnectionPoolSpec.The single-request client infrastructure should support absolute request URIs without path component
    akka.http.impl.engine.client.NewConnectionPoolSpec.The single-request client infrastructure should support absolute request URIs with a double slash path component
    akka.http.impl.engine.client.NewConnectionPoolSpec.The single-request client infrastructure should produce an error if the request does not have an absolute URI
    akka.http.javadsl.HttpExtensionApiSpec.The Java HTTP extension should create a cached connection pool (with a ConnectToHttp and a materializer)
```

akka-http-core/src/main/scala/akka/http/scaladsl/settings/ConnectionPoolSettings.scala
Falmarri: I think the globbing would be worthwhile, but I'm ok with doing an endswith to start. Are you suggesting ONLY do an endswith? Or are you saying that if  it starts with `*.`, then do an endswith?

Falmarri:
@ktoso whoops, sorry about that

Falmarri:
OK, only thing left is still getting an NPE on .superPool call, because it needs to get the pipeline value for the mapAsyncUnordered call before it has access to a request object to get the host. I have to think about how to handle that. (open to suggestions)

Falmarri:
@ktoso agreed. i got busy with other things, and i'm not sure on how to handle the superPool mapAsyncUnordered parameter in a decent way. Once that's decided on, this is good to go imo (unless you know of a better way than the null default parameters)

Falmarri:
@ktoso can you confirm whether these tests are related to my changes, or unrelated?

raboof:
PLS BUILD

raboof:
PLS BUILD

Falmarri:
@raboof binary compatibility is exactly the reason i went with null. i'm totally open to alternatives though, since it is indeed ugly.

raboof:
Perhaps the approach in https://github.com/akka/akka-http/compare/optionalCachedHostConnectionPoolWithBinCompat might be something to consider... not quite sure myself though. Thoughts?

(it is not strictly source compatible, but should be OK in most cases and easy to update in others...)

raboof:
Nice find!

This does mean the settings are really 'overrides': `ConnectionPoolSettings(system).withMaxConnections(12)` will still fall back to the configured `maxConnections` if any config overrides are configured for this host.

With this API, if I understand correctly you cannot set custom settings for a host in code when it is also configured in some config override. Since this is probably rare it's not a big problem if this is slightly cumbersome, but I do think it should be possible.

We could open up `forHost` and allow `ConnectionPoolSettings(system).forHost(host).withMaxConnections(12)`, or provide a separate 

```
object ConnectionPoolSettings {
  ...
  def forHost(host: String)(implicit system: ActorSystem): ConnectionPoolSettings =
  ...
}
```

In any case we should probably document this and not forget to include the `javadsl` as well.

Falmarri:
@raboof I'm not sure I understand the ask. 

> With this API, if I understand correctly you cannot set custom settings for a host in code when it is also configured in some config override.

This isn't correct. It will only use the configured overrides if you create the object with ConnectionPoolSettings.forDefault(), otherwise it will always return itself from .forHost, regardless of the configuration.

And since .forDefault is private to akka, ALL ConnectionPoolSettings created by users will be respected and never defer to configuration

akka-http-core/src/main/scala/akka/http/impl/settings/ConnectionPoolSettingsImpl.scala
```diff
@@ -63,7 +68,9 @@ object ConnectionPoolSettingsImpl extends SettingsCompanion[ConnectionPoolSettin
         case "legacy" ⇒ PoolImplementation.Legacy
         case "new"    ⇒ PoolImplementation.New
       },
-      c getPotentiallyInfiniteDuration "response-entity-subscription-timeout"
+      c getPotentiallyInfiniteDuration "response-entity-subscription-timeout",
+      Map.empty
     )
   }
+
```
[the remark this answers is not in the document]
Falmarri: My thought was that the config override wouldn't be visible in code. Basically I was thinking that if users supply a connection pool setting, that would always be taken as whatever was passed. the config overrides would only be available if you don't pass in an explicit setting (and thus it doesn't make sense to allow configuring the overrides from anywhere other than configuration).

I'm open to changing it, but I personally find it confusing to allow host-specific overrides when you're passing an explicitly configured setting in your code

akka-http-core/src/test/scala/akka/http/scaladsl/settings/ConnectionPoolSettingsSpec.scala
```diff
@@ -39,6 +40,84 @@ class ConnectionPoolSettingsSpec extends AkkaSpec {
...
+              |  per-host-override {
+              |    "akka.io" = { # can use same things as in global `host-connection-pool` section
+              |      max-connections = 47
+              |    }
+              |
+              |   "*.example.com" = { # allow `*` to apply overrides for all subdomains
+              |      max-connections = 34
+              |   }
+              |
+              |   "*example2.com" = { # allow `*` to apply overrides for all subdomains
```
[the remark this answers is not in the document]
Falmarri: Yes, bad copy/paste

akka-http-core/src/main/scala/akka/http/impl/settings/ConnectionPoolSettingsImpl.scala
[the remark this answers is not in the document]
ktoso: I don't see what you mean with the extremely confusing. Configs should always be possible to be changed using either way since people may have different needs and use cases. Does not matter if we strongly encourage the config way, methods should exist.

raboof:
@Falmarri ah, indeed I hadn't understood.

So: 
* if you configure `host-overrides` in the config, these will be taken when the user doesn't provide other settings
* If you create your own `ConnectionPoolSettings` from config (`ConnectionPoolSettings(system.settings.config)`), it will *ignore* the `host-overrides` from the config, and only load the root defaults

This indeed answers my previous question about overriding settings: because the user will never have access to an `ConnectionPoolSettings` instance with host-overrides, there can never be confusion between whether the host-overrides or overrides in code take precedence.

However, exposing `withHostMap`/`withHostOverride` will indeed lead to confusion, as it breaks that invariant: now you can have a `ConnectionPoolSettings` which has custom value for parameter X, but also a host-override for host Y. When connecting to host Y, should it use the customized value set on the root `ConnectionPoolSettings`, or the value in the host-override for Y?

Falmarri:
@raboof Yes that was my thinking. After discussing with @ktoso on gitter I went ahead with the `withHostMap`/`withHostOverride`, but my concerns still apply, that it's not immediately clear whether the config overrides will apply or not without knowing how the object was created.

>When connecting to host Y, should it use the customized value set on the root ConnectionPoolSettings, or the value in the host-override for Y?

I guess I didn't even go that far, but that's even more confusing, yes. Right now I believe it will take the full override for Y, regardless of what was set in the custom `ConnectionPoolSettings`

ktoso:
I have to admit I probably did not get the full extent of the confusion you meant there @Falmarri. I'm not entirely sure now what style to leave... Leaving it up to you and @raboof to decide :)

Falmarri:
@ktoso no worries =]

Falmarri:
@ktoso @raboof I reverted to the previous version of the PR that didn't expose modifying the overrides from code. Not sure what jenkins is complaining about though

raboof:
PLS BUILD

raboof:
Hm, looks like it can't merge this PR to master - could you try rebasing?

akka-http-core/src/main/resources/reference.conf
```diff
@@ -384,6 +384,20 @@ akka.http {
     client = {
       # no overrides by default, see `akka.http.client` for default values
     }
+
+    # Add client specific overrides to host-connectioin-pool. Allows exact and prefixed urls
```
jrudolph: "host-connection-pool"

akka-http-core/src/main/scala/akka/http/impl/settings/ConnectionPoolSettingsImpl.scala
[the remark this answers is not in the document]
jrudolph: Hmm, I see what you mean, but actually our convention is to have things modeled in our settings classes. 

For `singleRequest` it also makes sense to create only one object of settings with all the overrides.

akka-http-core/src/main/scala/akka/http/impl/settings/ConnectionPoolSettingsImpl.scala
```diff
@@ -50,6 +53,33 @@ private[akka] final case class ConnectionPoolSettingsImpl(
...
+    } else {
+      pattern.stripPrefix("glob:").map {
+        case '*'   ⇒ ".*"
+        case '?'   ⇒ "."
+        case '.'   ⇒ "\\."
+        case other ⇒ other.toString
+      }.mkString
+    }
+
+    // If the pattern starts with a wildcard, we want to match subdomains, as well as the raw domain
```
jrudolph: Is that a common assumption? I see how it can be useful but e.g. wildcard TLS certificates are *not* matched that way.

akka-http-core/src/main/scala/akka/http/impl/settings/ConnectionPoolSettingsImpl.scala
```diff
@@ -22,7 +24,8 @@ private[akka] final case class ConnectionPoolSettingsImpl(
   idleTimeout:                       Duration,
   connectionSettings:                ClientConnectionSettings,
   poolImplementation:                PoolImplementation,
-  responseEntitySubscriptionTimeout: Duration)
+  responseEntitySubscriptionTimeout: Duration,
+  hostMap:                           immutable.Map[Regex, ConnectionPoolSettings])
```
jrudolph: Why is that a map? It seems the mapping feature isn't used anywhere (but `find` instead on key-value pairs). Could be just a `Seq` of key-value pairs.

akka-http-core/src/main/scala/akka/http/scaladsl/settings/ConnectionPoolSettings.scala
```diff
@@ -76,6 +96,34 @@ abstract class ConnectionPoolSettings extends js.ConnectionPoolSettings { self:
 }
 
 object ConnectionPoolSettings extends SettingsCompanion[ConnectionPoolSettings] {
+
   override def apply(config: Config) = ConnectionPoolSettingsImpl(config)
+
+  private[akka] def forDefault(system: ActorSystem): ConnectionPoolSettings = {
```
jrudolph: Let's make it public and call it `withOverrides`.

jrudolph:
We also need more documentation for the feature.

akka-http-core/src/main/scala/akka/http/scaladsl/settings/ConnectionPoolSettings.scala
raboof: I think we had some back-and-forth on this one. The issue with adding a `withOverrides` is it becomes ambiguous what takes precedence when you both update individual settings from code and update settings for a how with `withOverrides`.

In its current state, we resolve this ambiguity by making `forDefault`/`withOverrides` private, so when creating a `ConnectionPoolSettings` from code it never takes into account overrides and always chooses the 'global default' settings overridden with the settings from code.

jrudolph: > I think we had some back-and-forth on this one. The issue with adding a `withOverrides` is it becomes ambiguous what takes precedence when you both update individual settings from code and update settings for a how with `withOverrides`.

I suspected that but didn't expand enough of the hidden communication in this PR. What I had not understood so far is the significance of doing the overriding not in our code but using config, so that `ConnectionPoolSettings.withOverrides().withMaxConnections(15)` would not apply at all if there's an override block for the target host in the config (because the settings for that host would be copied from the default config in the file).

That's bad indeed. But I also don't like the current solution which introduces another special case to ignore some parts of the config (the overrides) when the user creates a settings object.

Do we have any alternative ideas how to fix that problem? One idea would be to change all the `ConnectionPoolSettings.withXYZ` methods to also apply the change to the configured overrides but I guess that's the case of the ambiguity where it's unclear whether that manual change should actually override the overrides or just set a new default if no override was specified (which isn't represented in our code).

raboof: Say we have:
```
ConnectionPoolSettings.withOverrides()
  .withMaxConnections(42)
```
And there is a custom configuration configured for `foo.com`, then we'd expect to see `42` used when these `Settings` are used for `foo.com` as well, right? Doesn't that mean `withMaxConnections` would have to be adapted so the new value is applied not only to the global defaults, but also to all `Settings` in the `hostMap`?

raboof: (ah didn't see your last message before writing mine, these overlap nicely)

jrudolph:
@Falmarri 

> Does jenkins keep a stateful workspace?

Indeed, something was broken there and I wiped the workspace.

akka-http-core/src/main/scala/akka/http/scaladsl/settings/ConnectionPoolSettings.scala
raboof: I think it's reasonable to assume if you do `.withMaxConnections(42)` from code, this is the value you want to use in all cases when using this `Settings` object, regardless of earlier configuration.

I see 2 ways to achieve that behavior:
* adapting all `withXxx` methods to also apply the new setting to all host-specific overrides
* removing `hostMap` from the `Settings`, and solving this earlier: we could introduce a `ConnectionPoolSettings.forHost(hostname)` which would get the per-host defaults for this hostname (with a fallback to the configured defaults).

The advantage of the former approach is it's fairly straightforward to do, but not very elegant.

The advantage of the latter approach is it seems a bit clearer. The disadvantage is it might be hard to implement in a backwards-compatible way, and it's possible able to use a `Settings` for host `X` when actually connecting to host `Y`.

WDYT?

jrudolph: > Doesn't that mean `withMaxConnections` would have to be adapted so the new value is applied not only to the global defaults, but also to all `ConnectionPoolSettings` instances in the `hostMap`?

Then the order of changing default settings and override would matter.

I.e.

```
ConnectionPoolSettings.withOverrides()
  .withMaxConnections(42)
  .changeSettingsForHost("example.com") { settings =>
     settings.withMaxConnections(51)
  }
```

would first set all values for `withMaxConnections` to 42 and then again change it only for "example.com". But if you would swap those around, `withMaxConnections` would override the setting for "example.com".

I guess we could accept that. The good thing would be that we don't need the special loading case.

akka-http-core/src/main/scala/akka/http/impl/settings/ConnectionPoolSettingsImpl.scala
Falmarri: I guess I hadn't thought about it too much. It just seemed intuitive to me, but yeah, I dunno

akka-http-core/src/main/scala/akka/http/impl/settings/ConnectionPoolSettingsImpl.scala
Falmarri: Honestly I don't know why it's a map. If I made it a seq, would that address your concern about order? Does the typesafe config allow ordered retrieval in this way?

akka-http-core/src/main/scala/akka/http/scaladsl/settings/ConnectionPoolSettings.scala
Falmarri: So what was the outcome here? Personally I think exposing the overrides in code adds a significant amount of complexity for very marginal benefit (imo). Not just implementation complexity, since that can easily be hidden from users. But just complexity in what it's supposed to do, how to use it, etc.

Falmarri:
I just thought of another use case. What about cascading overrides? For example:

```
akka.http.host-connection-pool {
  max-connections = 7

  per-host-override : [
    {
      "*.io" = {
        client = {
          user-agent-header = "something specific"
        }
    },
    {
      "akka.io" = { # can use same things as in global `host-connection-pool` section
        max-connections = 47
      }
    }
  ]
}
```
So that akka.io would get both the user agent and the max-connections, but `other.io` would only get the user agent.

I'm going to push a version of this PR with that implement. That might simplify some of the confusion about which host override would take place as well

ennru:
Rebased into https://github.com/akka/akka-http/pull/2574
Closing this one.

Issue #1898, opened 2018-03-05
Have `Route.seal` not require any implicit parameters


jlprat:
As discussed in https://github.com/akka/akka-http/pull/1894#discussion_r172199319 we can change the `seal` method in `Route` to not require any implicit parameters.

jlprat:
I was looking at this issue, and we could remove all implicits except the exception and rejection handler. If we would remove those, this code would not be possible: https://doc.akka.io/docs/akka-http/current/routing-dsl/rejections.html#customising-rejection-http-responses

Is that what you had in mind @jrudolph?

ktoso:
Thanks!

jlprat:
Solved by https://github.com/akka/akka-http/commit/5826729506bec56a271ec25bcb2c79e983f0b409

PR #1928, opened 2018-03-12
Remove not needed implicits from Route.seal #1898

jlprat:
Refs: #1898
Initial attempt to remove not needed implicits
MiMa is not happy about the removal of the defaults
If deafaults stay, Scalac complaints about having multiple overloaded alternatives of method seal define default arguments.

### DIFF
--- a/akka-http/src/main/scala/akka/http/scaladsl/server/Route.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/server/Route.scala
@@ -5,13 +5,14 @@
 package akka.http.scaladsl.server
 
 import akka.NotUsed
+import akka.http.scaladsl.model.{ HttpRequest, HttpResponse }
+import akka.http.scaladsl.server.directives.BasicDirectives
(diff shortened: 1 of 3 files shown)


jlprat:
Asking @jrudolph and @ktoso for help, masters of Binary Compatibility.
Do you know if there is a trick to do this in a binary compatible way?

The only other alternative I can imagine is provide a `null` default value for `ParserSettings` and extract that with a directive if `null`

jlprat:
Ready to review :)

akka-http/src/main/scala/akka/http/scaladsl/server/Route.scala
```diff
@@ -32,14 +33,21 @@ object Route {
    *    using the ``~`` on routes or the [[Directive.|]] operator on directives.
    */
   def seal(route: Route)(implicit
-    routingSettings: RoutingSettings,
+    routingSettings: RoutingSettings = null,
```
ktoso: nice <3 (yes the null trick is ok)

PR #1931, opened 2018-03-14
Append newline to every element when streaming CSV (and NDJSON).

odd:
By using `Flow[ByteString].mapConcat(_ :: ByteString("\n") :: Nil)` instead of
`Flow[ByteString].intersperse(ByteString("\n"))` as frame renderer elements coming
from the source can be streamed to the client immediately, instead of
having to wait for the next element to become ready (or the stream to
finish) before streaming the current element to the client.


johanandren:
OK TO TEST

johanandren:
flatMapConcat does a sub materialization for every subgraph emitted, I'm not sure this is a good idea performance wise. If you/we don't care about trailing newline we can just do a `.map(bytestring => bytestring ++ newLineByteString)` instead.

odd:
Yes, that would be even better.

docs/src/test/scala/docs/http/scaladsl/server/directives/JsonStreamingExamplesSpec.scala
```diff
@@ -85,10 +85,9 @@ class JsonStreamingExamplesSpec extends RoutingSpec {
     // {"example":43}
     // ...
     // {"example":1000}
-    val sep = ByteString("\n")
 
     implicit val jsonStreamingSupport = EntityStreamingSupport.json()
-      .withFramingRenderer(Flow[ByteString].mapConcat(_ :: sep :: Nil))
+      .withFramingRenderer(Flow[ByteString].map(bs => bs ++ ByteString("\n")))
```
johanandren: Create a single `ByteString` instance outside of the stream and reuse that instead of allocating a new one for every `map` call.

johanandren:
And tests are passing, ofc. ;)

odd:
According to the CSV specifications I have found (mainly https://tools.ietf.org/html/rfc4180 and https://en.wikipedia.org/wiki/Comma-separated_values) each record (including the last) should in fact end with a newline (or rather according to RFC4180, a CRLF pair). The previous implementation did not enforce this (since the last element passed to the client did not end in a newline), whereas the new one does.

For NDJSON (at least according to https://tools.ietf.org/html/rfc7464 and http://jsonlines.org) it is permitted to have a trailing newline at the end of the stream.

jrudolph:
> as frame renderer elements coming
> from the source can be streamed to the client immediately, instead of
> having to wait for the next element to become ready

Is there a deeper reason why intersperse would hold back something? Maybe that's something that should be fixed on the akka-stream side as well?

odd:
@jrudolph I believe that the problem isn't that `intersperse` holds back anything but rather that without the trailing newline the data doesn't get sent to the client. When using `intersperse` you only get the trailing newline when the next element becomes available (or the source is finished). But by unsing `map` as in this PR, the newline is always appended to each element, the element is therefore sent directly to the client (without having to wait for the next element). Why the elements are not sent until they have the trailing newline I do not know, perhaps it has to do with the chunked transfer encoding being used for the streaming response (although according to specification https://tools.ietf.org/html/rfc7230#section-4.1 each chunk should be terminated by CRLF rather than a newline).

ktoso:
Pretty sure this does not have anything to do with "data not being sent"; for Akka this is transparent -- "just bytes"; Rather, could it be that your client is awaiting the \n and only then parses things?

johanandren:
`intersperse` holds back elements because it needs to know if the element is the last or there are more ones to select the right element to inject (or if there should be none at all, in case of no end-inject).

johanandren:
Ups. Strike that, I looked at the `intersperse` impl and that should not be the case.

raboof:
PLS BUILD

raboof:
@johanandren I'm not sure I follow your last comment, it seems to me that indeed `intersperse` has to wait until it observes the 'next' element before it can decide to output the newline for the 'previous' element?

I think this change makes sense, and the side-effect of also emitting a final newline seems like a feature more than a bug. The only problem is that it is a change in behavior from how it used to work - should we make it optional?

In terms of performance, I guess `bs ++ newline` could potentially be expensive... it would be neat to have a stage that emits a constant value after forwarding each upstream element but I don't think we have such a thing right now.

johanandren:
Ah yes, I think I thought about holding up the element, but the separator is held up until the next element comes in.

raboof:
Added issue #2033 to track this change

Issue #1849, opened 2018-02-09
Allow using HTTP/2 without SSL


jroper:
While browsers might not speak HTTP/2 without SSL, gRPC clients (eg the official grpc-java client) are easy to configure to not use SSL, and this is very convenient for ad-hoc testing, because it means you don't need to go and create certificates and ensure the client trusts them and so on and so on. This may be particularly useful for people that want to get started quickly, eg if they wanted to validate that Akka HTTP/Akka gRPC is right for them, as it will allow them to defer worrying about configuring SSL till after they've proved that the gRPC side of things works, when they've got more time to burn.

jroper:
There's no spec, there's just different configurable options that the official grpc client/server supports:

https://github.com/grpc/grpc-java/blob/master/netty/src/main/java/io/grpc/netty/NegotiationType.java

doofin:
Still not working without ssl

raboof:
@doofin that is correct: since #1934 we do support 'http2 without TLS with preexisting knowledge', but that is of somewhat limited use (it does not allow using insecure http2 and insecure regular HTTP on the same port) and has not been released yet. h2c support is being worked on in #1966.

raboof:
Nowadays we support both h2c with prior knowledge and h2c via the Upgrade header, so I think it's high time we close this

PR #1934, opened 2018-03-14
Allow Http2 without negotiation/TLS

raboof:
Refs #1849. Still needs docs, tests and Java API.

### DIFF
--- a/akka-http2-support/src/main/scala/akka/http/scaladsl/Http2.scala
+++ b/akka-http2-support/src/main/scala/akka/http/scaladsl/Http2.scala
@@ -35,6 +35,49 @@ class Http2Ext(private val config: Config)(implicit val system: ActorSystem)
 
   val http = Http(system)
 
+  /**
+   * Handle requests using HTTP/2 immediately, without any TLS or negotiation layer.
+   */
+  def bindAndHandleRaw(
+    handler:   HttpRequest ⇒ Future[HttpResponse],
+    interface: String, port: Int = DefaultPortForProtocol,
+    settings:    ServerSettings = ServerSettings(system),
+    parallelism: Int            = 1,
+    log:         LoggingAdapter = system.log)(implicit fm: Materializer): Future[ServerBinding] = {
+    val effectivePort = if (port >= 0) port else 80
+
+    val serverLayer: Flow[ByteString, ByteString, Future[Done]] = Flow.fromGraph(StreamUtils.fuseAggressive(
+      Flow[HttpRequest]
+        .watchTermination()(Keep.right)
+        // FIXME: parallelism should maybe kept in track with SETTINGS_MAX_CONCURRENT_STREAMS so that we don't need
+        // to buffer requests that cannot be handled in parallel
+        .via(Http2Blueprint.handleWithStreamIdHeader(parallelism)(handler)(system.dispatcher))
+        .joinMat(Http2Blueprint.serverStack(settings, log))(Keep.left)))
+
+    val connections = Tcp().bind(interface, effectivePort, settings.backlog, settings.socketOptions, halfClose = false, settings.timeouts.idleTimeout)
+
+    connections.mapAsyncUnordered(settings.maxConnections) {
+      incoming: Tcp.IncomingConnection ⇒
+        try {
+          serverLayer.addAttributes(Http.prepareAttributes(settings, incoming)).joinMat(incoming.flow)(Keep.left)
+            .run().recover {
+              // Ignore incoming errors from the connection as they will cancel the binding.
+              // As far as it is known currently, these errors can only happen if a TCP error bubbles up
+              // from the TCP layer through the HTTP layer to the Http.IncomingConnection.flow.
+              // See https://github.com/akka/akka/issues/17992
+              case NonFatal(ex) ⇒
+                Done
+            }(ExecutionContexts.sameThreadExecutionContext)
+        } catch {
+          case NonFatal(e) ⇒
+            log.error(e, "Could not materialize handling flow for {}", incoming)
+            throw e
+        }
+    }.mapMaterializedValue {
+      _.map(tcpBinding ⇒ ServerBinding(tcpBinding.localAddress)(() ⇒ tcpBinding.unbind()))(fm.executionContext)
+    }.to(Sink.ignore).run()
+  }
+
   def bindAndHandleAsync(
     handler:   HttpRequest ⇒ Future[HttpResponse],
     interface: String, port: Int = DefaultPortForProtocol,
@@ -91,7 +134,7 @@ class Http2Ext(private val config: Config)(implicit val system: ActorSystem)
     val connections = Tcp().bind(interface, effectivePort, settings.backlog, settings.socketOptions, halfClose = false, settings.timeouts.idleTimeout)
 
     connections.mapAsyncUnordered(settings.maxConnections) {
-      case incoming: Tcp.IncomingConnection ⇒
+      incoming: Tcp.IncomingConnection ⇒
         try {
           fullLayer().addAttributes(Http.prepareAttributes(settings, incoming)).joinMat(incoming.flow)(Keep.left)
             .run().recover {


akka-http2-support/src/main/scala/akka/http/javadsl/Http2.scala
```diff
@@ -0,0 +1,60 @@
...
+import scala.concurrent.Future
+
+object Http2 extends ExtensionId[Http2] with ExtensionIdProvider {
+  override def get(system: ActorSystem): Http2 = super.get(system)
+  def lookup() = Http
+  def createExtension(system: ExtendedActorSystem): Http2 = new Http2(system)
+}
+
+class Http2(system: ExtendedActorSystem) extends Extension {
+  import akka.dispatch.ExecutionContexts.{ sameThreadExecutionContext ⇒ ec }
```
ktoso: please pass it explicitly rather than import like this; we want to avoid using it accidentally

akka-http2-support/src/main/scala/akka/http/javadsl/Http2.scala
```diff
@@ -0,0 +1,60 @@
...
+import akka.http.impl.util.JavaMapping.Implicits._
+import akka.japi.Function
+import akka.stream.Materializer
+
+import scala.compat.java8.FutureConverters._
+import scala.concurrent.Future
+
+object Http2 extends ExtensionId[Http2] with ExtensionIdProvider {
+  override def get(system: ActorSystem): Http2 = super.get(system)
+  def lookup() = Http
```
ktoso: Http2

akka-http2-support/src/main/scala/akka/http/javadsl/Http2.scala
```diff
@@ -0,0 +1,60 @@
...
+  def createExtension(system: ExtendedActorSystem): Http2 = new Http2(system)
+}
+
+class Http2(system: ExtendedActorSystem) extends Extension {
```
ktoso: final

akka-http2-support/src/main/scala/akka/http/javadsl/Http2.scala
```diff
@@ -0,0 +1,60 @@
...
+  import akka.dispatch.ExecutionContexts.{ sameThreadExecutionContext ⇒ ec }
+
+  import language.implicitConversions
+  private implicit def javaModelIsScalaModel[J <: AnyRef, S <: J](in: Future[J]): Future[S] = in.asInstanceOf[Future[S]]
```
ktoso: fun one ;) ok

akka-http2-support/src/main/scala/akka/http/javadsl/Http2.scala
```diff
@@ -0,0 +1,60 @@
...
+import akka.http.javadsl.model.{ HttpRequest, HttpResponse }
+import akka.http.javadsl.settings.ServerSettings
+import akka.http.impl.util.JavaMapping.Implicits._
+import akka.japi.Function
+import akka.stream.Materializer
+
+import scala.compat.java8.FutureConverters._
+import scala.concurrent.Future
+
+object Http2 extends ExtensionId[Http2] with ExtensionIdProvider {
```
ktoso: If we're introducing it, then please mirror methods from scaladsl's Http2 or mark with huge FIXME's that it's not mirroring all methods yet

akka-http2-support/src/main/scala/akka/http/scaladsl/Http2.scala
```diff
@@ -35,6 +35,49 @@ class Http2Ext(private val config: Config)(implicit val system: ActorSystem)
 
   val http = Http(system)
 
+  /**
+   * Handle requests using HTTP/2 immediately, without any TLS or negotiation layer.
+   */
+  def bindAndHandleRaw(
```
ktoso: Raw sounds a bit weird; `bindAndHandlePlainText`?  AFAIR that's also wording used in GRPC

akka-http2-support/src/main/scala/akka/http/scaladsl/Http2.scala
```diff
@@ -91,7 +134,7 @@ class Http2Ext(private val config: Config)(implicit val system: ActorSystem)
     val connections = Tcp().bind(interface, effectivePort, settings.backlog, settings.socketOptions, halfClose = false, settings.timeouts.idleTimeout)
 
     connections.mapAsyncUnordered(settings.maxConnections) {
-      case incoming: Tcp.IncomingConnection ⇒
+      incoming: Tcp.IncomingConnection ⇒
```
ktoso: good

akka-http2-support/src/main/scala/akka/http/scaladsl/Http2.scala
```diff
@@ -35,6 +35,49 @@ class Http2Ext(private val config: Config)(implicit val system: ActorSystem)
...
+   */
+  def bindAndHandleRaw(
+    handler:   HttpRequest ⇒ Future[HttpResponse],
+    interface: String, port: Int = DefaultPortForProtocol,
+    settings:    ServerSettings = ServerSettings(system),
+    parallelism: Int            = 1,
+    log:         LoggingAdapter = system.log)(implicit fm: Materializer): Future[ServerBinding] = {
+    val effectivePort = if (port >= 0) port else 80
+
+    val serverLayer: Flow[ByteString, ByteString, Future[Done]] = Flow.fromGraph(StreamUtils.fuseAggressive(
```
ktoso: we can remove `StreamUtils.fuseAggressive` now since Akka HTTP 10.1 requires 2.5 and this is a No-Op on 2.5+

ktoso:
Some comments, looks like a good start though. We can merge this as we continue working on these things and polish remaining API / deduplicate later

akka-http2-support/src/main/scala/akka/http/scaladsl/Http2.scala
raboof: That's indeed what `io.grpc.netty` uses, but I found `PlainText` a bit weird since even though there's no encryption, it's still a binary protocol. 

This approach is described in the spec as [starting 'with prior knowledge'](http://httpwg.org/specs/rfc7540.html#known-http), but `bindAndHandleWithPriorKnowledge` is weird because it's the client that has prior knowledge, not the server.

raboof:
Should be good to go now I think

raboof:
Should be good to go now I think

akka-http2-support/src/main/scala/akka/http/scaladsl/Http2.scala
patriknw: would it be possible to make it a configuration property or a special `connectionContext` instead?

I'd image that you might want to easily switch between this and the real thing, perhaps running (some) tests with this and full integration tests and production with the real thing.

ktoso: > That's indeed what io.grpc.netty uses, but I found PlainText a bit weird since even though there's no encryption, it's still a binary protocol.

Yeah, I've heard this argument but the general feel it conveys seems right "though shalt not use paintext in prod" ;-) Crypto wise I think plaintext is referred to non encrypted, regardless if binary or not?

ktoso: Ah, I meant this comment; Yeah trying to achieve this with a connection context would be preferable indeed, could you give it a look @raboof ?

ktoso: Basically it's the same as we have in Akka HTTP/1, where we have the plain and https one; the same could work here basically, even if we strongly recommend the https one

johanandren:
Also wondering if we could have it as a context instead, did you give that some thought and landed on it being impossible/bad @raboof ?

raboof:
Extending the `ConnectionContext` to allow this switch seems suitable. Right now the `ConnectionContext` is just `isSecure` (and a default port). To add http2 to that we would like to support the following variations:
* secure (h2 can be negotiated if enabled in the config)
* insecure (no h2)
* insecure (h2 can be negotiated (not implemented yet))
* insecure (h2 only (introduced in this PR))

I suppose we could add a `http2` option to `ConnectionContext`:  with possible values `never`, `negotiated` or `always`, set it to `never` or `negotiated` by default based on `settings.previewServerSettings.enableHttp2`? So configuration in code trumps that configuration option? WDYT?

johanandren:
Sounds (naively) good to me.

ktoso:
Yes this sounds okey. Note that the new method on connection context we should directly add on `ConnectionContext` I guess. Should throw if HTTP2 is requested but the support is not on classpath in `Http()`.

`Http2()` will have to be relaxed to accept any connection context, not just Https since we allow the plaintext one: 

```
  def bindAndHandleAsync(
    handler:   HttpRequest ⇒ Future[HttpResponse],
    interface: String, port: Int = DefaultPortForProtocol,
    httpsContext: HttpsConnectionContext, // relax this
```

patriknw:
ah, I forgot that there were more work remaining here. let's not include this in the release. we can always use a snapshot for grpc if we need it earlier than next release.

docs/src/test/java/docs/http/javadsl/Http2Test.java
```diff
@@ -32,5 +36,12 @@ void testBindAndHandleAsync() {
         toHostHttps("127.0.0.1", 8443).withCustomHttpsContext(httpsConnectionContext),
         materializer);
     //#bindAndHandleAsync
+
+    //#bindAndHandleRaw
+
+    // TODO is there a reason there is no `bindAndHandle` that accepts
+    // a `ConnectionContext` in the javadsl?
```
raboof: Is there? Should we add it?

akka-http-core/src/main/mima-filters/10.0.11.backwards.excludes
```diff
@@ -86,3 +86,8 @@ ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.javadsl.TimeoutA
 ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.scaladsl.TimeoutAccess.timeout")
 ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.scaladsl.TimeoutAccess.getTimeout")
 
+# Constructors of classes that should have had @DoNotInherit in #1934
+# Should probably move to 10.1.1.backwards.includes since this is not expected to be merged into 10.1.1
```
ktoso: yes ;-) please do

docs/src/test/java/docs/http/javadsl/Http2Test.java
ktoso: In Java connection context is carried in ConnectHttp, the DSL that kept our sanity since otherwise too many overloads were needed:

```java
  def bindAndHandle(
    handler:      Flow[HttpRequest, HttpResponse, _],
    connect:      ConnectHttp,
```


```java
abstract class ConnectHttp {
  def host: String
  def port: Int

  def isHttps: Boolean
  def connectionContext: Optional[HttpsConnectionContext]
```

akka-http-core/src/main/scala/akka/http/scaladsl/UseHttp2.scala
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * Specify whether to support HTTP/2: never, negotiated, or always.
+ */
+sealed trait UseHttp2 extends akka.http.javadsl.UseHttp2 {
+  override def asScala = this.asInstanceOf[UseHttp2]
+}
+object Never extends UseHttp2
+object Negotiated extends UseHttp2
+object Always extends UseHttp2
```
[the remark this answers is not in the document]
raboof: Agreed about moving them to a companion object.

`UseHttp2` is indeed not a very nice name. Secure/insecure and no/negotiated/always http2 are orthogonal though:

* insecure no: http h1
* insecure negotiated: h2c (http + Upgrade + h1/h2)
* insecure always: 'raw' h2 ('preexisting knowledge')
* secure no: https + h1
* secure negotiated: https + ALPN + h1/h2
* secure always: https + h2

So it might make sense to keep them as 2 separate fields instead of an 'enum' of 6?

akka-http-core/src/main/scala/akka/http/scaladsl/ConnectionContext.scala
```diff
@@ -28,18 +30,29 @@ object ConnectionContext {
...
+
   // for binary-compatibility, since 2.4.7
   def https(
     sslContext:          SSLContext,
     enabledCipherSuites: Option[immutable.Seq[String]],
     enabledProtocols:    Option[immutable.Seq[String]],
     clientAuth:          Option[TLSClientAuth],
     sslParameters:       Option[SSLParameters]) =
-    new HttpsConnectionContext(sslContext, None, enabledCipherSuites, enabledProtocols, clientAuth, sslParameters)
+    new HttpsConnectionContext(sslContext, sslConfig = None, enabledCipherSuites, enabledProtocols, clientAuth, sslParameters, http2 = Negotiated)
```
raboof: Not *sure* this is the correct default yet (between `Never` and `Negotiated`), but right now there's no functional difference yet, so we can still revisit this.

akka-http-core/src/main/scala/akka/http/scaladsl/ConnectionContext.scala
```diff
@@ -28,18 +30,29 @@ object ConnectionContext {
...
     clientAuth:          Option[TLSClientAuth]         = None,
-    sslParameters:       Option[SSLParameters]         = None) =
-    new HttpsConnectionContext(sslContext, sslConfig, enabledCipherSuites, enabledProtocols, clientAuth, sslParameters)
+    sslParameters:       Option[SSLParameters]         = None,
+    http2:               UseHttp2                      = UseHttp2.Negotiated) =
+    new HttpsConnectionContext(sslContext, sslConfig, enabledCipherSuites, enabledProtocols, clientAuth, sslParameters, http2)
   //#https-context-creation
 
+  // for binary-compatibility, since 1.10.2
+  def https(
```
johanandren: `@deprecated` ?

akka-http2-support/src/main/scala/akka/http/scaladsl/Http2.scala
```diff
@@ -27,23 +27,85 @@ import scala.util.Success
...
   // FIXME: won't having the same package as top-level break osgi?
 
   private[this] final val DefaultPortForProtocol = -1 // any negative value
 
   val http = Http(system)
 
+  /**
+   * Handle requests using HTTP/2 immediately, without any TLS or negotiation layer.
+   */
+  private def bindAndHandleRaw(
```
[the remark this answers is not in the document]
johanandren: It's private so it doesn't matter so much

akka-http-core/src/main/scala/akka/http/javadsl/ConnectHttp.scala
```diff
@@ -24,9 +27,10 @@ abstract class ConnectHttp {
     if (connectionContext.isPresent) connectionContext.get()
     else fallbackContext
 
-  override def toString = s"ConnectHttp($host,$port,$isHttps,$connectionContext)"
+  override def toString = s"ConnectHttp($host,$port,$isHttps,$connectionContext,$http2)"
 }
 
+@DoNotInherit
 object ConnectHttp {
```
ktoso: can't inherit object anyway ;-)

akka-http-core/src/main/scala/akka/http/javadsl/ConnectionContext.scala
```diff
@@ -58,21 +60,25 @@ object ConnectionContext {
     scaladsl.ConnectionContext.noEncryption()
 }
 
+@DoNotInherit
 abstract class ConnectionContext {
   def isSecure: Boolean
   def sslConfig: Option[AkkaSSLConfig]
+  def http2: UseHttp2
```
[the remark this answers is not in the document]
ktoso: yes / no?

akka-http-core/src/main/scala/akka/http/scaladsl/ConnectionContext.scala
```diff
@@ -28,18 +30,31 @@ object ConnectionContext {
...
     enabledProtocols:    Option[immutable.Seq[String]] = None,
     clientAuth:          Option[TLSClientAuth]         = None,
-    sslParameters:       Option[SSLParameters]         = None) =
-    new HttpsConnectionContext(sslContext, sslConfig, enabledCipherSuites, enabledProtocols, clientAuth, sslParameters)
+    sslParameters:       Option[SSLParameters]         = None,
+    http2:               UseHttp2                      = UseHttp2.Negotiated) =
+    new HttpsConnectionContext(sslContext, sslConfig, enabledCipherSuites, enabledProtocols, clientAuth, sslParameters, http2)
   //#https-context-creation
 
+  // for binary-compatibility, since 1.10.2
```
ktoso: 10.1.2

akka-http-core/src/main/scala/akka/http/scaladsl/ConnectionContext.scala
```diff
@@ -28,18 +30,31 @@ object ConnectionContext {
...
+  @deprecated
```
ktoso: deprecated should have since and reason filled in, should also have `@Deprecated`

akka-http-core/src/main/scala/akka/http/scaladsl/ConnectionContext.scala
```diff
@@ -28,18 +30,31 @@ object ConnectionContext {
...
+    sslContext:          SSLContext,
+    sslConfig:           Option[AkkaSSLConfig],
+    enabledCipherSuites: Option[immutable.Seq[String]],
+    enabledProtocols:    Option[immutable.Seq[String]],
+    clientAuth:          Option[TLSClientAuth],
+    sslParameters:       Option[SSLParameters]) =
+    new HttpsConnectionContext(sslContext, sslConfig, enabledCipherSuites, enabledProtocols, clientAuth, sslParameters, http2 = Negotiated)
+
   // for binary-compatibility, since 2.4.7
+  @deprecated
```
ktoso: same as above

akka-http-core/src/main/scala/akka/http/scaladsl/ConnectionContext.scala
ktoso: Negotiated is ok

akka-http-core/src/main/scala/akka/http/scaladsl/ConnectionContext.scala
```diff
@@ -50,8 +65,19 @@ final class HttpsConnectionContext(
...
   val enabledProtocols:    Option[immutable.Seq[String]] = None,
   val clientAuth:          Option[TLSClientAuth]         = None,
-  val sslParameters:       Option[SSLParameters]         = None)
-  extends akka.http.javadsl.HttpsConnectionContext with ConnectionContext {
+  val sslParameters:       Option[SSLParameters]         = None,
+  http2:                   UseHttp2                      = Negotiated)
+  extends akka.http.javadsl.HttpsConnectionContext(http2) with ConnectionContext {
+
+  // for binary-compatibility, since 2.10.2
+  def this(
```
ktoso: mark deprecated?

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
```diff
@@ -95,17 +96,15 @@ class HttpExt private[http] (private val config: Config)(implicit val system: Ex
...
-      )
+      Flow[HttpRequest]
+        .watchTermination()(Keep.right)
+        .viaMat(handler)(Keep.left)
+        .watchTermination() { (termWatchBefore, termWatchAfter) ⇒
+          // flag termination when the user handler has gotten (or has emitted) termination
+          // signals in both directions
+          termWatchBefore.flatMap(_ ⇒ termWatchAfter)(ExecutionContexts.sameThreadExecutionContext)
+        }
+        .joinMat(baseFlow)(Keep.left)
```
ktoso: good cleanup :)

akka-http2-support/src/main/scala/akka/http/scaladsl/Http2.scala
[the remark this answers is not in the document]
ktoso: Still is a bit weird ;) Internal etc; Raw does not mean much to me here hm

docs/src/main/paradox/server-side/http2.md
```diff
@@ -39,7 +40,20 @@ Java
 
 Note that `bindAndHandle` currently does not support HTTP/2, you must use `bindAndHandleAsync`.
 
-## Testing with cURL
+### HTTP/2 without HTTPS
+
+While un-encrypted connections are allowed by HTTP/2, clients that support this are rare.
```
ktoso: not sure about this claim; rather that it is not recommended to be used for real outside of testing?

docs/src/test/java/docs/http/javadsl/Http2Test.java
```diff
@@ -32,5 +41,13 @@ void testBindAndHandleAsync() {
         toHostHttps("127.0.0.1", 8443).withCustomHttpsContext(httpsConnectionContext),
         materializer);
     //#bindAndHandleAsync
+
+    //#bindAndHandleRaw
```
ktoso: remove the Raw?

akka-http-core/src/main/scala/akka/http/scaladsl/ConnectionContext.scala
johanandren: Didn't we say there is no need for the Java deprecation over in the Akka repo?

raboof: I figured as it's in the `scaladsl` package (and the javadsl doesn't inherit from the scaladsl) it might be fine here?

Issue #1471, opened 2017-10-10
Show off dynamically updating routes


raboof:
One of our strong points is how the Routing DSL allows both efficient static and highly dynamic routes.

We don't always make that very clear, see for example #1470 . Perhaps we should add a section/example in the docs somewhere? Perhaps a server that changes/adds routes based on incoming requests, like a "http mock server server" where you can dynamically add mock endpoints and associated responses, behaviour etc? Should be fairly simple to do, right?

ktoso:
Good ticket, triaged; People sometimes ask about that expecting it to be super hard but it's very simple :)

PR #1935, opened 2018-03-14
Add dynamically updating routes example (#1471)

tuleism:
Ref #1471.

### DIFF
--- a/docs/src/main/paradox/routing-dsl/index.md
+++ b/docs/src/main/paradox/routing-dsl/index.md
@@ -69,6 +69,38 @@ the Routing DSL will look like:
 
 @@@
 
+## Dynamic Routing Example
+
+As the routes are evaluated for each request, it is possible to make changes at runtime. Please note that every access
+may happen on a separated thread, so any shared mutable state must be thread safe.
+
+The following is an Akka HTTP route definition that allows dynamically adding new or updating mock endpoints with
+associated request-response pairs at runtime.
+
+Scala
+:  @@snip [HttpServerExampleSpec.scala]($test$/scala/docs/http/scaladsl/HttpServerExampleSpec.scala) { #dynamic-routing-example }
+
+Java
+:  @@snip [HttpServerDynamicRoutingExampleTest.java]($test$/java/docs/http/javadsl/HttpServerDynamicRoutingExampleTest.java) { #dynamic-routing-example }
+
+For example, let's say we do a POST request with body:
+
+```json
+{
+  "path": "test",
+  "requests": [
+    {"id": 1},
+    {"id": 2}
+  ],
+  "responses": [
+    {"amount": 1000},
+    {"amount": 2000}
+  ]
+}
+```
+
+Subsequent POST request to `/test` with body `{"id": 1}` will be responded with `{"amount": 1000}`.
+
 ## Handling HTTP Server failures in the High-Level API
 
 There are various situations when failure may occur while initialising or running an Akka HTTP server.
--- a/docs/src/test/java/docs/http/javadsl/HttpServerDynamicRoutingExampleTest.java
+++ b/docs/src/test/java/docs/http/javadsl/HttpServerDynamicRoutingExampleTest.java
@@ -1,0 +1,103 @@
+/*
+ * Copyright (C) 2009-2018 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package docs.http.javadsl;
+
+import akka.NotUsed;
+import akka.actor.ActorSystem;
+import akka.http.javadsl.ConnectHttp;
+import akka.http.javadsl.Http;
+import akka.http.javadsl.marshallers.jackson.Jackson;
+import akka.http.javadsl.model.HttpRequest;
+import akka.http.javadsl.model.HttpResponse;
+import akka.http.javadsl.model.StatusCodes;
+import akka.http.javadsl.server.AllDirectives;
+import akka.http.javadsl.server.Route;
+import akka.stream.ActorMaterializer;
+import akka.stream.javadsl.Flow;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.JsonNode;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+public class HttpServerDynamicRoutingExampleTest extends AllDirectives {
+
+  public static void main(String[] args) throws Exception {
+    // boot up server using the route as defined below
+    ActorSystem system = ActorSystem.create("routes");
+
+    final Http http = Http.get(system);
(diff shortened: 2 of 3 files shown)


docs/src/test/scala/docs/http/scaladsl/HttpServerExampleSpec.scala
```diff
@@ -699,5 +699,49 @@ class HttpServerExampleSpec extends WordSpec with Matchers
...
+          val mapping = mock.requests.zip(mock.responses).toMap
+          state.put(mock.path, mapping)
+          complete("ok")
+        }
+      }
+    }
+
+    // dynamic routing based on current state
+    val dynamicRoute: Route = ctx => {
+      val routes = state.map { kv =>
```
raboof: Perhaps use `case (path, response)` notation here?

docs/src/test/scala/docs/http/scaladsl/HttpServerExampleSpec.scala
```diff
@@ -699,5 +699,49 @@ class HttpServerExampleSpec extends WordSpec with Matchers
...
+    implicit val materializer = ActorMaterializer()
+
+    val state = TrieMap.empty[String, Map[JsObject, JsObject]]
 
+    // fixed route to update state
+    val fixedRoute: Route = post {
+      pathSingleSlash {
+        entity(as[MockDefinition]) { mock =>
+          val mapping = mock.requests.zip(mock.responses).toMap
+          state.put(mock.path, mapping)
```
raboof: While `TrieMap` is indeed thread-safe, it might be a better example to replace the `Map`?

jlprat:
OK TO TEST

tuleism:
@raboof 
- I've addressed your comments on the Scala code.
- I've added a Java version of the example. This is mostly copied from other examples as I'm not familiar with Java. Please see if it makes sense.
- I also change the type to `JsValue` to be more general than Json object, and change the documentation to be more focused on the route definition.

docs/src/test/java/docs/http/javadsl/HttpServerDynamicRoutingExampleTest.java
```diff
@@ -0,0 +1,111 @@
...
+
+    final Http http = Http.get(system);
+    final ActorMaterializer materializer = ActorMaterializer.create(system);
+
+    //In order to access all directives we need an instance where the routes are define.
+    HttpServerDynamicRoutingExampleTest app = new HttpServerDynamicRoutingExampleTest();
+
+    final Flow<HttpRequest, HttpResponse, NotUsed> routeFlow = app.createRoute().flow(system, materializer);
+    final CompletionStage<ServerBinding> binding = http.bindAndHandle(routeFlow,
+      ConnectHttp.toHost("localhost", 8080), materializer);
```
raboof: I'd say it should be sufficient to just end the `main` method after `http.bindAndHandle(routeFlow,
ConnectHttp.toHost("localhost", 8080), materializer);` and leave it up to the user to kill the application, though to make that convenient we should add sbt-revolver to our build

docs/src/test/scala/docs/http/scaladsl/HttpServerExampleSpec.scala
```diff
@@ -699,5 +699,50 @@ class HttpServerExampleSpec extends WordSpec with Matchers
...
+    implicit val format = jsonFormat3(MockDefinition)
+
+    var state = Map.empty[String, Map[JsValue, JsValue]]
+
+    // fixed route to update state
+    val fixedRoute: Route = post {
+      pathSingleSlash {
+        entity(as[MockDefinition]) { mock =>
+          val mapping = mock.requests.zip(mock.responses).toMap
+          state += mock.path -> mapping
```
raboof: perhaps even `state = state + (mock.path -> mapping)`?

(your example is indeed equally safe, but that might not be obvious to people newer to Scala)

tuleism:
@raboof 
Both new comments are addressed.

docs/src/test/scala/docs/http/scaladsl/HttpServerExampleSpec.scala
```diff
@@ -699,5 +699,50 @@ class HttpServerExampleSpec extends WordSpec with Matchers
...
+    import spray.json._
 
+    implicit val system = ActorSystem()
+    implicit val materializer = ActorMaterializer()
+
+    //#dynamic-routing-example
+    case class MockDefinition(path: String, requests: Seq[JsValue], responses: Seq[JsValue])
+    implicit val format = jsonFormat3(MockDefinition)
+
+    var state = Map.empty[String, Map[JsValue, JsValue]]
```
ktoso: hmm mutating this Map from potentially multiple threads -- since one handler flow is materialized for every connection, so there's many at once. Should this be made volatile for the example?

tuleism: Yes, done!

docs/src/test/java/docs/http/javadsl/HttpServerDynamicRoutingExampleTest.java
```diff
@@ -0,0 +1,102 @@
...
+    //In order to access all directives we need an instance where the routes are define.
+    HttpServerDynamicRoutingExampleTest app = new HttpServerDynamicRoutingExampleTest();
+
+    final Flow<HttpRequest, HttpResponse, NotUsed> routeFlow = app.createRoute().flow(system, materializer);
+
+    http.bindAndHandle(routeFlow, ConnectHttp.toHost("localhost", 8080), materializer);
+  }
+
+  //#dynamic-routing-example
+  private Map<String, Map<JsonNode, JsonNode>> state = new HashMap<>();
```
johanandren: This also needs to be made thread safe - `final private ... = new ConcurrentHashMap<>();`

docs/src/main/paradox/routing-dsl/index.md
```diff
@@ -69,6 +69,36 @@ the Routing DSL will look like:
 
 @@@
 
+## Dynamic Routing Example
```
johanandren: Would be good with an introductory paragraph saying that as the routes are evaluated for each request it is possible to make it change at runtime. However every access may happen on a separate thread so any shared mutable state must be thread safe.

tuleism:
@johanandren I've fixed for the Java example and the introductory paragraph.

docs/src/main/paradox/routing-dsl/index.md
```diff
@@ -69,6 +69,38 @@ the Routing DSL will look like:
...
+
+As the routes are evaluated for each request, it is possible to make changes at runtime. Please note that every access
+may happen on a separated thread, so any shared mutable state must be thread safe.
```
johanandren: Great

ktoso: Good note, thanks

PR #1936, opened 2018-03-14
=htp Discard Entity in Default RejectionHandler #38

jlprat:
Refs: #38
Default RejectionHandler is rejecting now the request entity by default.
Not knowing a better way to check if the `Source` has been discarded, it's counting elements emitted in test

### DIFF
--- a/akka-http-tests/src/test/scala/akka/http/scaladsl/server/DiscardEntityDefaultRejectionHandlerSpec.scala
+++ b/akka-http-tests/src/test/scala/akka/http/scaladsl/server/DiscardEntityDefaultRejectionHandlerSpec.scala
@@ -9,6 +9,7 @@ import akka.http.scaladsl.model.HttpEntity
 import akka.http.scaladsl.model.StatusCodes.NotFound
 import akka.stream.scaladsl.Source
 import akka.util.ByteString
+import org.scalatest.concurrent.Eventually._
 import org.scalatest.concurrent.ScalaFutures
 
 class DiscardEntityDefaultRejectionHandlerSpec extends RoutingSpec with ScalaFutures {
@@ -17,20 +18,24 @@ class DiscardEntityDefaultRejectionHandlerSpec extends RoutingSpec with ScalaFut
     complete("bar")
   }
 
+  private val numElems = 1000
+  @volatile
   private var elementsEmitted = 0
   private def gimmeElement(): ByteString = {
     elementsEmitted = elementsEmitted + 1
     ByteString("Foo")
   }
 
-  private val ThousandElements: Stream[ByteString] = Stream.continually(gimmeElement()).take(1000)
+  private val ThousandElements: Stream[ByteString] = Stream.continually(gimmeElement()).take(numElems)
   private val RequestToNotHandled = Get("/bar", HttpEntity(`text/plain(UTF-8)`, Source[ByteString](ThousandElements)))
 
   "Default RejectionHandler" should {
     "rejectEntity by default" in {
       RequestToNotHandled ~> Route.seal(route) ~> check {
         status shouldBe NotFound
-        elementsEmitted should be > 1
+        eventually {
+          elementsEmitted shouldBe numElems
+        }
       }
     }
   }
--- a/akka-http-tests/src/test/scala/akka/http/scaladsl/server/DiscardEntityDefaultRejectionHandlerSpec.scala
+++ b/akka-http-tests/src/test/scala/akka/http/scaladsl/server/DiscardEntityDefaultRejectionHandlerSpec.scala
@@ -1,0 +1,38 @@
+/*
+ * Copyright (C) 2009-2018 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.http.scaladsl.server
+
+import akka.http.scaladsl.model.ContentTypes.`text/plain(UTF-8)`
+import akka.http.scaladsl.model.HttpEntity
+import akka.http.scaladsl.model.StatusCodes.NotFound
+import akka.stream.scaladsl.Source
+import akka.util.ByteString
+import org.scalatest.concurrent.ScalaFutures
+
+class DiscardEntityDefaultRejectionHandlerSpec extends RoutingSpec with ScalaFutures {
+
+  private val route = path("foo") {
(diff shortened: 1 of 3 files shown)


jlprat:
Will take a look at the failures

ktoso:
Failure seems related, will you have time to look into it? https://jenkins.akka.io:8498/job/pr-validator-akka-http/2339/testReport/akka.http.scaladsl.server/DiscardEntityDefaultRejectionHandlerSpec/Default_RejectionHandler_should_rejectEntity_by_default/

akka-http-tests/src/test/scala/akka/http/scaladsl/server/WithoutSizeLimitSpec.scala
```diff
@@ -49,7 +49,9 @@ class WithoutSizeLimitSpec extends WordSpec with Matchers with RequestBuilding w
 
       val (hostName, port) = SocketUtil.temporaryServerHostnameAndPort()
 
-      EventFilter[EntityStreamSizeException](occurrences = 1).intercept {
+      // It went from 1 occurrence to 2 after discarding the entity, I think is due to the retrying nature of `handleRejections`
+      // that causes one Exception for the original entity and another one from the rejected one.
+      EventFilter[EntityStreamSizeException](occurrences = 2).intercept {
```
jlprat: As mentioned in the comment, now there are 2 occurrences of the exception

jlprat:
Ready for review!

akka-http-tests/src/test/scala/akka/http/scaladsl/server/DiscardEntityDefaultRejectionHandlerSpec.scala
```diff
@@ -0,0 +1,41 @@
...
+import org.scalatest.concurrent.Eventually._
+import org.scalatest.concurrent.ScalaFutures
+
+class DiscardEntityDefaultRejectionHandlerSpec extends RoutingSpec with ScalaFutures {
+
+  private val route = path("foo") {
+    complete("bar")
+  }
+
+  private var elementsEmitted = 0
```
johanandren: Consumed from a stream so needs to be `@volatile` (I think that is enough since there will never be any actual concurrent access)

akka-http-tests/src/test/scala/akka/http/scaladsl/server/DiscardEntityDefaultRejectionHandlerSpec.scala
```diff
@@ -0,0 +1,41 @@
...
+
+  private val ThousandElements: Stream[ByteString] = Stream.continually(gimmeElement()).take(1000)
+  private val RequestToNotHandled = Get("/bar", HttpEntity(`text/plain(UTF-8)`, Source[ByteString](ThousandElements)))
+
+  "Default RejectionHandler" should {
+    "rejectEntity by default" in {
+      RequestToNotHandled ~> Route.seal(route) ~> check {
+        status shouldBe NotFound
+        eventually {
+          elementsEmitted should be > 1
```
johanandren: Hmm. So when the rejection handler is triggered, it will still have the client send the entire body? Why not check exactly that rather than >1 then?

jlprat: As far as I understand, all the bytes will be discarded, so yes, all the bytes should be sent from the client.
You are right, I can check for the exact number.

akka-http-tests/src/test/scala/akka/http/scaladsl/server/DiscardEntityDefaultRejectionHandlerSpec.scala
jlprat: Right!

akka-http-tests/src/test/scala/akka/http/scaladsl/server/DiscardEntityDefaultRejectionHandlerSpec.scala
johanandren: I know early response is problematic, but this feels a bit weird, let's say I was going to do an upload with a gigabyte of data but used the wrong path...

jlprat: mmm maybe I misunderstood the problem stated in https://github.com/akka/akka-http/issues/38.
If the entity is not discarded, the connections will stall, right?

jlprat: Please let me know if I got it wrong

johanandren: Yeah, not sure what is right here, just struck me that it sounds like weird behavior. Gut feeling is that we'd rather want to cancel, but I guess then the entire connection is torn down?

johanandren:
Now the GitHub UI ate my concerns, so re-adding:

With `discardEntityBytes`, we consume all data from the client. Imagine a case where I POST 1 Gb of data with the wrong path, this will mean my client uploads all 1 Gb before getting a `NotFound` back, this doesn't feel quite right.

jlprat:
And following to this, concern, AFAIU, not discarding the entity will cause the connection to stall waiting for those bytes to be consumed. Am I right?

johanandren:
Yeap, so neither is a good idea as far as I can see.

jlprat:
Open to suggestions :)

johanandren:
Hoping someone more into the details of HTTP in general could chime in. Is that early response always bad? Would cancelling the request body break things?

jlprat:
@ktoso Now that you're back, WDYT about this?

johanandren:
This comment and the linked RFC is interesting: https://github.com/akka/akka-http/pull/1746#issuecomment-387646118

jlprat:
One important fact about the RFC is that they use *SHOULD* for this, so clients might decide to not implement it and still be compliant, don't they?

jlprat:
Rebased as Github complained about it

Issue #687, opened 2016-12-26
Relax user-agent parsing to fit reality


gaydenko:
Hi! I get:
```
Illegal request header: Illegal 'user-agent' header: Invalid input '/', expected OWS, 'EOI', tchar, product-or-comment or comment (line 1, column 184): Mozilla/5.0 (Linux; U; Android 5.0.2; en-us; Redmi Note 3 Build/LRX22G) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/53.0.2785.146 Mobile Safari/537.36 XiaoMi/MiuiBrowser/8.4.6
```
So, the parser is too strict to fit reality.

viktorklang:
# Enables/disables the logging of warning messages in case an incoming
    # message (request or response) contains an HTTP header which cannot be
    # parsed into its high-level model class due to incompatible syntax.
    # Note that, independently of this settings, akka-http will accept messages
    # with such headers as long as the message as a whole would still be legal
    # under the HTTP specification even without this header.
    # If a header cannot be parsed into a high-level model instance it will be
    # provided as a `RawHeader`.
    # If logging is enabled it is performed with the configured
    # `error-logging-verbosity`.
    illegal-header-warnings = on

http://doc.akka.io/docs/akka-http/current/scala/http/configuration.html

gaydenko:
I would prefer to keep other warnings not-suppressed.

viktorklang:
? The warning is logged due to a spec violation.

gaydenko:
Yes, I'm saying just about that. When millions devices don't follow a specification, I guess it is more handy for software to follow reality rather the (just still not updated) specification :)

ktoso:
I do agree that the world of http is a mess. The user agent warnings are
very exposed to a mess that is the world of crappy clients, thus we could
change the handling as well as optional settings for specifically that
header.

jrudolph:
I agree we might relax it if it really happens for lots of people. Is the `XiaoMi/MiuiBrowser` really as prominent as you say? Could we also file an issue with them to fix their user-agent header as well?

gaydenko:
@jrudolph Sorry, I guess it may be difficult to get some statistics related to Xiaomi spread. It is one of the most popular China manufacturer, and the browser is the official **stock** one.

I get the warning as log file viewer rather than as device owner :)

jrudolph:
I see, even if they would fix it now the problem right might still persist for years to come.

[...]

PR #1942, opened 2018-03-16
add setting ignore-illegal-header-for #687

shnmorimoto:
Ref #687


ktoso:
OK TO TEST

ktoso:
Another PR from the hackathon, please review @raboof @johanandren ? :)

raboof:
Thanks for your contribution @shnmorimoto, this is indeed something people run into a lot, and it would be great if we could give them more fine-grained tools to suppress such warnings.

You might have noticed the builds failed, this is because of our automated binary incompatibility checks:
```
[error]  * abstract method getIgnoreIllegalHeaderFor()scala.collection.immutable.Set in class akka.http.javadsl.settings.ParserSettings is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.javadsl.settings.ParserSettings.getIgnoreIllegalHeaderFor")
[error]  * method copy(java.lang.String,java.lang.String)akka.http.scaladsl.model.ErrorInfo in class akka.http.scaladsl.model.ErrorInfo does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.scaladsl.model.ErrorInfo.copy")
[error]  * method this(java.lang.String,java.lang.String)Unit in class akka.http.scaladsl.model.ErrorInfo does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.scaladsl.model.ErrorInfo.this")
[error]  * method apply(java.lang.String,java.lang.String)akka.http.scaladsl.model.ErrorInfo in object akka.http.scaladsl.model.ErrorInfo does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.scaladsl.model.ErrorInfo.apply")
[error]  * abstract method ignoreIllegalHeaderFor()scala.collection.immutable.Set in class akka.http.scaladsl.settings.ParserSettings is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.scaladsl.settings.ParserSettings.ignoreIllegalHeaderFor")
```

The warnings on `ParserSettings` can be ignored because `ParserSettings` is marked `@DoNotInherit` (we have an issue open to [automate this logic](https://github.com/lightbend/migration-manager/issues/160)). You can add them to the list of ignored warnings as described [in CONTRIBUTING.md](https://github.com/akka/akka-http/blob/master/CONTRIBUTING.md#binary-compatibility).

The warnings on `ErrorInfo` are a bit more tricky: these pop up because `ErrorInfo` is a `case class`, and the automatically-generated methods on this case class changed because the fields changed. I think it would be fine to turn `ErrorInfo` into a regular class and hand-implemented the missing API's (perhaps private?) You can see something similar has been done to `RouterSettings` in ababace00363d446cb6c4442020d2509a8e7c295.

Would you be up for fixing those?

shnmorimoto:
@raboof Thank you for your review.

I understood about the reason for the builds failed.

I will check the reference code and I will fix it.

shnmorimoto:
@raboof cc: @ktoso I fix binary incompatibility.

akka-http-core/src/main/mima-filters/10.0.11.backwards.excludes
```diff
@@ -86,3 +86,5 @@ ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.javadsl.TimeoutA
 ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.scaladsl.TimeoutAccess.timeout")
 ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.scaladsl.TimeoutAccess.getTimeout")
 
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.javadsl.settings.ParserSettings.getIgnoreIllegalHeaderFor")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.scaladsl.settings.ParserSettings.ignoreIllegalHeaderFor")
```
raboof: Shouldn't we introduce a `10.1.0.backwards.excludes`?

akka-http-core/src/main/resources/reference.conf
```diff
@@ -386,6 +386,9 @@ akka.http {
     # `error-logging-verbosity`.
     illegal-header-warnings = on
 
+    # TODO comment
```
raboof: Describe the setting here

akka-http-core/src/main/scala/akka/http/scaladsl/model/ErrorInfo.scala
```diff
@@ -12,12 +12,40 @@ import StatusCodes.ClientError
...
  */
-final case class ErrorInfo(summary: String = "", detail: String = "") {
+final class ErrorInfo(
+  val summary:         String = "",
+  val detail:          String = "",
+  val errorHeaderName: String = ""
+) extends scala.Product with scala.Serializable with scala.Equals with java.io.Serializable {
   def withSummary(newSummary: String) = copy(summary = newSummary)
   def withSummaryPrepended(prefix: String) = withSummary(if (summary.isEmpty) prefix else prefix + ": " + summary)
+  def withHeaderNamePrepend(headerName: String) = setErrorHeaderName(errorHeaderName = headerName)
```
raboof: This doesn't actually prepend the header name, right?

akka-http-core/src/main/scala/akka/http/scaladsl/model/ErrorInfo.scala
```diff
@@ -12,12 +12,40 @@ import StatusCodes.ClientError
...
   def withFallbackSummary(fallbackSummary: String) = if (summary.isEmpty) withSummary(fallbackSummary) else this
   def formatPretty = if (summary.isEmpty) detail else if (detail.isEmpty) summary else summary + ": " + detail
   def format(withDetail: Boolean): String = if (withDetail) formatPretty else summary
+
+  private[akka] def copy(summary: String = summary, detail: String = detail): ErrorInfo = {
+    new ErrorInfo(summary, detail, errorHeaderName)
+  }
+
+  private[akka] def setErrorHeaderName(errorHeaderName: String): ErrorInfo = new ErrorInfo(summary, detail, errorHeaderName)
```
raboof: Perhaps make this public and call it `withErrorHeaderName`?

akka-http-core/src/main/scala/akka/http/scaladsl/model/ErrorInfo.scala
```diff
@@ -26,6 +54,10 @@ object ErrorInfo {
    * Used for example when catching exceptions generated by the header value parser, which doesn't provide
    * summary/details information but structures its exception messages accordingly.
    */
+  private[akka] def apply(summary: String = "", detail: String = ""): ErrorInfo = new ErrorInfo(summary, detail, "")
```
raboof: You introduced these new methods between the existing scaladoc documentation and the method they document below - can you reorder that?

akka-http-core/src/main/mima-filters/10.0.11.backwards.excludes
shnmorimoto: move configuration to 10.1.0.backwards.excludes.

akka-http-core/src/main/resources/reference.conf
shnmorimoto: add comment.

akka-http-core/src/main/scala/akka/http/scaladsl/model/ErrorInfo.scala
shnmorimoto: Surely you are right.　rename method.

akka-http-core/src/main/scala/akka/http/scaladsl/model/ErrorInfo.scala
shnmorimoto: Surely you are right. delete this method and rename `withHeaderNamePrepend` method to `withErrorHeaderName`.

akka-http-core/src/main/scala/akka/http/scaladsl/model/ErrorInfo.scala
shnmorimoto: I made a mistake.  reorder methods. thanks.

shnmorimoto:
I'd like to consult one point. 
Should We check the header name in settings sensitive? Or insensitive?

when set 'user-agent' in settings, ignore warn.
but set 'User-agent' in settings, logged warn now.

shnmorimoto:
Thanks! I will fix it.

akka-http-core/src/main/scala/akka/http/scaladsl/model/ErrorInfo.scala
```diff
@@ -12,15 +12,45 @@ import StatusCodes.ClientError
  * repeating anything present in the message itself (in order to not open holes for XSS attacks),
  * while the detail can contain additional information from any source (even the request itself).
  */
-final case class ErrorInfo(summary: String = "", detail: String = "") {
+final class ErrorInfo(
```
ktoso: please leave it as a `case class`, no reason to change IMHO; we should mark it as `@InternalApi` perhaps even

shnmorimoto: Thank you for your review.
I put case class back. and add annotation `@InternalAPI` as described below.

```scala
/**
 * INTERNAL API
 * Two-level model of error information.
 * The summary should explain what is wrong with the request or response *without* directly
 * repeating anything present in the message itself (in order to not open holes for XSS attacks),
 * while the detail can contain additional information from any source (even the request itself).
 */
@InternalApi private[akka] final case class ErrorInfo(summary: String = "", detail: String = "", errorHeaderName: String = "") {
  def withSummary(newSummary: String) = copy(summary = newSummary)
  def withSummaryPrepended(prefix: String) = withSummary(if (summary.isEmpty) prefix else prefix + ": " + summary)
  def withErrorHeaderName(headerName: String) = copy(errorHeaderName = headerName.toLowerCase)
  def withFallbackSummary(fallbackSummary: String) = if (summary.isEmpty) withSummary(fallbackSummary) else this
  def formatPretty = if (summary.isEmpty) detail else if (detail.isEmpty) summary else summary + ": " + detail
  def format(withDetail: Boolean): String = if (withDetail) formatPretty else summary
}

/** INTERNAL API */
@InternalApi private[akka] object ErrorInfo {
  /**
   * Allows constructing an `ErrorInfo` from a single string.
   * Used for example when catching exceptions generated by the header value parser, which doesn't provide
   * summary/details information but structures its exception messages accordingly.
   */
  def fromCompoundString(message: String): ErrorInfo = message.split(": ", 2) match {
    case Array(summary, detail) ⇒ apply(summary, detail)
    case _                      ⇒ ErrorInfo("", message)
  }
}
```
I check binary compatible with  `mimaReportBinaryIssues` sbt command.
I still receive a following error log.
```
[error]  * method copy(java.lang.String,java.lang.String)akka.http.scaladsl.model.ErrorInfo in class akka.http.scaladsl.model.ErrorInfo does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.scaladsl.model.ErrorInfo.copy")
[error]  * method this(java.lang.String,java.lang.String)Unit in class akka.http.scaladsl.model.ErrorInfo does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.scaladsl.model.ErrorInfo.this")
[error]  * method apply(java.lang.String,java.lang.String)akka.http.scaladsl.model.ErrorInfo in object akka.http.scaladsl.model.ErrorInfo does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.scaladsl.model.ErrorInfo.apply")
```
Should I add three method(copy, this, apply) to the mima-filters configuration?

ktoso: Yes, the check fail is expected and yes it should be added to the excludes :)

ktoso: akka-http-core/src/main/mima-filters/10.1.0.backwards.excludes

shnmorimoto: Thanks! I understand.
I put case class back. And add some methods to 10.1.0.backwards.excludes.

ktoso:
Thanks for the continued work, we'll look into this soon -- on Monday most likely

akka-http-core/src/main/scala/akka/http/scaladsl/model/ErrorInfo.scala
```diff
@@ -5,22 +5,26 @@
...
 
 /**
+ * INTERNAL API
  * Two-level model of error information.
  * The summary should explain what is wrong with the request or response *without* directly
  * repeating anything present in the message itself (in order to not open holes for XSS attacks),
  * while the detail can contain additional information from any source (even the request itself).
  */
-final case class ErrorInfo(summary: String = "", detail: String = "") {
+@InternalApi private[akka] final case class ErrorInfo(summary: String = "", detail: String = "", errorHeaderName: String = "") {
```
johanandren: We cannot simply make it an internal class as far as I can see, it is part of the public API in `IllegalRequestException` for example.

Keeping this binary compatible while adding a field requires quite a bit of gymnastics I'm afraid, if completely possible at all: we need a two param constructor, a two param `apply`, etc.

johanandren: Sorry to give you conflicting feedback btw, I see you got directed to annotate like this, maybe @ktoso know something I don't know.

shnmorimoto: Thank you for review.

I understand the reason for can't change internal class.
I missed it is part of the public API. And thank you for pointing that out.

How about the following method to keep case class and binary compatible? Although a bit tricky.

```scala
final case class ErrorInfo(summary: String = "", detail: String = "") {
  var errorHeaderName = ""
  def withSummary(newSummary: String) = copy(summary = newSummary).withErrorHeaderName(errorHeaderName)
  def withSummaryPrepended(prefix: String) = withSummary(if (summary.isEmpty) prefix else prefix + ": " + summary)
  def withErrorHeaderName(headerName: String) = {
    errorHeaderName = headerName.toLowerCase
    this
  }
  def withFallbackSummary(fallbackSummary: String) = if (summary.isEmpty) withSummary(fallbackSummary) else this
  def formatPretty = if (summary.isEmpty) detail else if (detail.isEmpty) summary else summary + ": " + detail
  def format(withDetail: Boolean): String = if (withDetail) formatPretty else summary
}
```

Or should we rethink from the idea of adding headerName to ErrorInfo.
WDYT @johanandren @ktoso ?

johanandren: I don't think we can have a `var` in there because of concurrency. Otoh, maybe we can make the constructor internal, and keep the rest of the binary signature intact somehow? The type and accessors needs to be public but there is not really a reason for anyone outside of the library to construct instances.

shnmorimoto: I indeed did not consider concurrency.
Thank you for your explanation.
I understood the reason why I should not use `var`. 

And I'm sorry. I missed your proposal that you wrote above.

Does the meaning that you wrote match in the following?

1. add two param `apply` method and mark internal. As described below.

```scala
object ErrorInfo {
  /**
   * Allows constructing an `ErrorInfo` from a single string.
   * Used for example when catching exceptions generated by the header value parser, which doesn't provide
   * summary/details information but structures its exception messages accordingly.
   */
  def fromCompoundString(message: String): ErrorInfo = message.split(": ", 2) match {
    case Array(summary, detail) ⇒ apply(summary, detail)
    case _                      ⇒ ErrorInfo("", message)
  }

  /** INTERNAL API */
  @InternalApi private[akka] def apply(summary: String, detail: String): ErrorInfo = {
    ErrorInfo(summary, detail)
  }
}
```

2. `copy` and `this` methods are still binary incompatible. I still don't understand well this...
Should we add two methods to mima-filters configuration? Or add methods just like `apply`?

johanandren: `this`/constructor can be solved with auxiliary constructor, but copy method is impossible to get around I think, since it is generated for case classes. Maybe there is no way around breaking bin comp. I think that if users touch this it will be the unapply, could we keep a two param unapply somehow so that wouldn't be incompatible at least?

shnmorimoto: Thank you for replay.

I understand that how to fix bin comp for `this method` and `copy method`.
And indeed it seems better to consider about `unapply method`, too.

I defined unapply method as below.

```scala
object ErrorInfo {
  /** INTERNAL API */
  @InternalApi private[akka] def apply(summary: String, detail: String): ErrorInfo = {
    ErrorInfo(summary, detail, "")
  }

  def unapply(info: ErrorInfo): Option[(String, String)] = Some((info.summary, info.detail))
  ...
```

But it did not work well.

```scala
scala> val info = ErrorInfo("summary", "detail")

info: akka.http.scaladsl.model.ErrorInfo = ErrorInfo(summary,detail,)

scala> val ErrorInfo(summary, detail) = info
<console>:18: error: wrong number of arguments for pattern akka.http.scaladsl.model.ErrorInfo(summary: String,detail: String,errorHeaderName: String)
       val ErrorInfo(summary, detail) = info
```

`unapply method` generated By case class and `unapply method` I defined both have the same arguments, although the return value is different. 
Therefore, I think that it is probably impossible to decide which is used.

For now, I pushed fixed code without unapply.

akka-http-core/src/test/scala/akka/http/scaladsl/model/headers/HeaderSpec.scala
```diff
@@ -23,7 +23,7 @@ class HeaderSpec extends FreeSpec with Matchers {
             CacheDirectives.`s-maxage`(1000)))
       }
       "failing parse run" in {
-        val Left(List(ErrorInfo(summary, detail))) = headers.`Last-Modified`.parseFromValueString("abc")
+        val Left(List(ErrorInfo(summary, detail, ""))) = headers.`Last-Modified`.parseFromValueString("abc")
```
johanandren: This original one would be how I expect it to be used by users if any users use it - the two param unapply - 
``case ErrorInfo(summary, detail) => ...` and that will break because of the binary and source incomp. introduced.

It may be a very rare use case though, I'm not sure.

shnmorimoto: Thank you for pointing out.

I think that the possibility can not be denied as you wrote about such usecase.

I tried several methods in addition to try to solve incomp, but I could not solve it.
Maybe `unapply` generated by case class can not override and can it be defined in multiple.

Should we change ErrorInfo to normal class for to solve incomp？

raboof: Changing it from a case class to a normal class (with unapply/copy methods) sounds fine to me!

johanandren: Yeah, maybe that is the only viable option

shnmorimoto: Thanks for the replay.

I changed it and pushed it.

ktoso: Ok, thanks for fixing it :)

shnmorimoto:
Thank you for review and fix comment. 

I have no experience with testing logging behavior.
So, I will check if there is any test code to be helpful. And I will try to add such tests.

raboof:
PLS BUILD

raboof:
@shnmorimoto sounds great! for adding a test that tests logging behavior https://doc.akka.io/docs/akka/current/testing.html#expecting-log-messages might be useful.

shnmorimoto:
Thank you for the document!
I am going to add that tests on this weekend.

shnmorimoto:
I added a test and push it.

shnmorimoto:
Sorry to push some times.
I fixed a bit about the test.

shnmorimoto:
@ktoso I will be glad if you review at your convenience.

ktoso:
Looking great now, thanks! Updated branch and can merge after validation

ktoso:
Thanks a lot and sorry this took such a long time to merge! :-)

shnmorimoto:
@ktoso @raboof @johanandren 
Thank you for merge.
And thank you for advising me on the review.

PR #1945, opened 2018-03-17
fix file-io-dispatcher #1879

shkoder:
* use akka.stream.materializer.blocking-io-dispatcher for file IO

Ref #1879

Issue #1879, opened 2018-02-22
The file directives relies on broken streams config


johanandren:
The file directive has config (https://github.com/akka/akka-http/blob/6cc1ec970fdfae4e257169158bbc66c48449d93b/akka-http/src/main/resources/reference.conf#L42) which uses `akka.stream.blocking-io-dispatcher` and that was deprecated (because it was incorrect) in akka/akka#24440 

I actually think the file directive should not set a dispatcher attribute at all (because it uses FileIO.fromPath(file.toPath) which uses the IODispatcher and that can be globally configured now)

ktoso:
(please add tags)

shkoder:
I'll take a look into this

shkoder:
#1945 introduces a basic fix while #1946 removes the dispatcher attribute. Removal of the dispatcher attribute leads to debatable consequences, that's why it's a separate PR.

PR #1946, opened 2018-03-17
do not set dispatcher for file directives and remove file-io-dispatcher config #1879

shkoder:
ref #1879

Removing dispatcher attributes in the `FileAndResourceDirectives` has the following consequences:
 - `akka.stream.materializer.blocking-io-dispatcher` is used by default
 - there are no means to have separate dispatchers for the file directives and `akka.stream.materializer.blocking-io-dispatcher`
- `file-io-dispatcher` config is not really used. Removing all the related code breaks binary compatibility in the public API (`RoutingSettings`), so respective methods are marked as deprecated

As these consequences are debatable, it is a separate PR from the basic fix in #1945


ktoso:
OK TO TEST

akka-http-core/src/main/scala/akka/http/scaladsl/settings/RoutingSettings.scala
```diff
@@ -36,7 +38,8 @@ abstract class RoutingSettings private[akka] () extends akka.http.javadsl.settin
   override def withRangeCountLimit(rangeCountLimit: Int): RoutingSettings = self.copy(rangeCountLimit = rangeCountLimit)
   override def withRangeCoalescingThreshold(rangeCoalescingThreshold: Long): RoutingSettings = self.copy(rangeCoalescingThreshold = rangeCoalescingThreshold)
   override def withDecodeMaxBytesPerChunk(decodeMaxBytesPerChunk: Int): RoutingSettings = self.copy(decodeMaxBytesPerChunk = decodeMaxBytesPerChunk)
-  override def withFileIODispatcher(fileIODispatcher: String): RoutingSettings = self.copy(fileIODispatcher = fileIODispatcher)
+  @deprecated("binary compatibility method. Use `akka.stream.materializer.blocking-io-dispatcher` to configure the dispatcher")
+  override def withFileIODispatcher(fileIODispatcher: String): RoutingSettings = self
```
ktoso: has to be annotated using Java's `@Deprecated` as well; 
Scala annotation must include `, since = "10.1.1"` part

docs/src/test/java/docs/http/javadsl/server/directives/BasicDirectivesExamplesTest.java
```diff
@@ -218,27 +218,18 @@ public void testWithSettings() {
...
-      // ContentTypes.APPLICATION_JSON, source
-      final Source<ByteString, Object> source =
-        FileIO.fromPath(Paths.get("example.json"))
-          .mapMaterializedValue(completionStage -> (Object) completionStage);
-      return complete(
-        HttpResponse.create()
-          .withEntity(HttpEntities.create(ContentTypes.APPLICATION_JSON, source))
-      );
-    });
+        .withFileGetConditional(false);
```
ktoso: why this?

akka-http-core/src/main/scala/akka/http/impl/settings/RoutingSettingsImpl.scala
```diff
@@ -17,7 +17,7 @@ private[http] final case class RoutingSettingsImpl(
   rangeCountLimit:          Int,
   rangeCoalescingThreshold: Long,
   decodeMaxBytesPerChunk:   Int,
-  fileIODispatcher:         String) extends akka.http.scaladsl.settings.RoutingSettings {
+  fileIODispatcher:         String  = "") extends akka.http.scaladsl.settings.RoutingSettings {
```
ktoso: can be made a method instead in order to keep compatibility; also mark it as deprecated then

ktoso:
@johanandren please review as you remember context there perhaps more.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectives.scala
```diff
@@ -66,11 +65,8 @@ trait FileAndResourceDirectives {
       if (file.isFile && file.canRead)
         conditionalFor(file.length, file.lastModified) {
           if (file.length > 0) {
-            withRangeSupportAndPrecompressedMediaTypeSupportAndExtractSettings { settings ⇒
-              complete {
-                HttpEntity.Default(contentType, file.length,
-                  FileIO.fromPath(file.toPath).withAttributes(ActorAttributes.dispatcher(settings.fileIODispatcher)))
```
shkoder: remove the dispatcher attribute rendering the `file-io-dispatcher` setting effectively unused

akka-http/src/main/resources/reference.conf
```diff
@@ -36,8 +36,4 @@ akka.http.routing {
   # The maximum number of bytes per ByteString a decoding directive will produce
   # for an entity data stream.
   decode-max-bytes-per-chunk = 1m
-
-  # Fully qualified config path which holds the dispatcher configuration
-  # to be used by FlowMaterialiser when creating Actors for IO operations.
-  file-io-dispatcher = ${akka.stream.blocking-io-dispatcher}
```
shkoder: remove the now unused `file-io-dispatcher` from the `reference.conf`

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectives.scala
```diff
@@ -210,10 +202,9 @@ trait FileAndResourceDirectives {
 }
 
 object FileAndResourceDirectives extends FileAndResourceDirectives {
-  private val withRangeSupportAndPrecompressedMediaTypeSupportAndExtractSettings =
+  private val withRangeSupportAndPrecompressedMediaTypeSupport =
     RangeDirectives.withRangeSupport &
-      CodingDirectives.withPrecompressedMediaTypeSupport &
-      BasicDirectives.extractSettings
```
shkoder: no need to extract settings as the settings were only used to get the file io dispatcher config

akka-http-core/src/main/scala/akka/http/scaladsl/settings/RoutingSettings.scala
```diff
@@ -36,7 +39,9 @@ abstract class RoutingSettings private[akka] () extends akka.http.javadsl.settin
   override def withRangeCountLimit(rangeCountLimit: Int): RoutingSettings = self.copy(rangeCountLimit = rangeCountLimit)
   override def withRangeCoalescingThreshold(rangeCoalescingThreshold: Long): RoutingSettings = self.copy(rangeCoalescingThreshold = rangeCoalescingThreshold)
   override def withDecodeMaxBytesPerChunk(decodeMaxBytesPerChunk: Int): RoutingSettings = self.copy(decodeMaxBytesPerChunk = decodeMaxBytesPerChunk)
-  override def withFileIODispatcher(fileIODispatcher: String): RoutingSettings = self.copy(fileIODispatcher = fileIODispatcher)
+  @deprecated("binary compatibility method. Use `akka.stream.materializer.blocking-io-dispatcher` to configure the dispatcher", since = "10.1.1")
+  @Deprecated
+  override def withFileIODispatcher(fileIODispatcher: String): RoutingSettings = self
```
shkoder: removal of the `file-io-dispatcher` config makes the whole `*FileIODispatcher` family of methods unnecessary, but they need to stay because of the binary compatibility. The implementation changes to no op.

docs/src/test/java/docs/http/javadsl/server/directives/BasicDirectivesExamplesTest.java
shkoder: as `withFileIODispatcher()` gets deprecated in this PR and it does not actually change settings, the example now demonstrates how to use `withSettings` just with another routing setting - `fileGetConditional` which looks to be a good replacement.

raboof:
This looks like an interesting improvement, sorry to leave this hanging for so long. Unfortunately it seems to have gained some merge conflicts since then, @shkoder would you like to fix those?

shkoder:
sure, I'll rebase

PR #1950, opened 2018-03-19
Support conversion between scaladsl.Uri and javadsl.Uri

### LINKED ISSUE #1949: Support conversion between scaladsl.Uri and javadsl.Uri
Before Akka Http 1.1.0 we could convert a Javadsl URI to its Scaladsl counterpart via
```
uri.asInstanceOf[akka.http.impl.model.JavaUri].uri`
```
This is now forbidden as JavaUri was made private.

Scaladsl to Javadsl conversion used to be possible with
```
JavaUri(sUri)
```

Akka Http should add an official way to convert between these types.

See https://github.com/akka/alpakka/issues/847

ennru:
Adds `javadsl.model.Uri.asScala()`
Adds `static javadsl.model.Uri.create(scaladsl.model.Uri)`
Removes the internal conversion method from `akka.http.impl.util.Util`.

Fixes #1949

### DIFF
--- a/akka-http-core/src/main/java/akka/http/javadsl/model/Uri.java
+++ b/akka-http-core/src/main/java/akka/http/javadsl/model/Uri.java
@@ -172,7 +172,7 @@ public abstract class Uri {
   public abstract String getPathString();
 
   /**
-   * Returns the scaladsl view of this instance
+   * Returns the Scala DSL representation of this Uri.
    */
   public abstract akka.http.scaladsl.model.Uri asScala();
 
@@ -193,7 +193,7 @@ public abstract class Uri {
   }
 
   /**
-   * Returns an Uri wrapping the scaladsl representation.
+   * Returns the Java DSL representation of a Scala DSL Uri.
    */
   public static Uri create(akka.http.scaladsl.model.Uri uri) {
     return new JavaUri(uri);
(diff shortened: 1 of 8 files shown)


ennru:
The new method is marked as possible incompatibility
*abstract method asScala()akka.http.scaladsl.model.Uri in class akka.http.javadsl.model.Uri is present only in current version*

Ok to add to create `10.1.0.backwards.excludes` with it?

ktoso:
Yes it's fine to add since the model is assumed "we may add methods there, do not inherit from it"

akka-http-core/src/main/java/akka/http/javadsl/model/Uri.java
```diff
@@ -186,6 +192,13 @@ public static Uri create(String uri) {
     return new JavaUri(akka.http.scaladsl.model.Uri.apply(uri));
   }
 
+  /**
+   * Returns an Uri wrapping the scaladsl representation.
```
ktoso: don't explain the implementation details :-)
"Returns the Java DSL representation of an Scala DSL Uri" :)

akka-http-core/src/main/java/akka/http/javadsl/model/Uri.java
```diff
@@ -171,6 +171,12 @@
    */
   public abstract String getPathString();
 
+  /**
+   * Returns the scaladsl view of this instance
```
ktoso: same as bellow, explain that it mirrors the APIs, no impl details in such methods pls

PR #1960, opened 2018-03-23
Revert "!pro make akka-actor / akka-stream a provided dependency (#1582)"

ktoso:
This reverts commit dc708850193244eb3c67f34d77bc4401c2fa36a3.

We accidentally backported the change to the build file which made akka-stream not depended on automatically; This is fine and documented in 10.1, however such change should not exist on 10.0.

This PR reverts this commit

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -75,7 +75,7 @@ lazy val root = Project(
   )
 
 lazy val parsing = project("akka-parsing")
-  .addAkkaModuleDependency("akka-actor", "provided")
+  .addAkkaModuleDependency("akka-actor")
   .settings(Dependencies.parsing)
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -87,15 +87,15 @@ lazy val parsing = project("akka-parsing")
 
 lazy val httpCore = project("akka-http-core")
   .dependsOn(parsing)
-  .addAkkaModuleDependency("akka-stream", "provided")
```
ktoso: this is what breaks the 10.0 dependency semantics -- we moved to "explicitly depend on akka" in 10.1, however this should not be backported to 10.0

PR #1961, opened 2018-03-23
Always render the custom header from the example

raboof:


ktoso:
LGTM, what was the rationale for this though? Likely some discussion I missed?

PR #1964, opened 2018-03-27
=doc document client-side for streaming JSON

ktoso:
It was pointed out on akka-user that we didn't document the client-side for consuming streaming JSON responses nicely; I added those docs.

`Re: [akka-user][deprecated] Akka Http and consuming a response of "application/stream+json" type`


ktoso:
As this is likely difficult to visualise how it renders on the page, here a screenshot: 

Scala: 
json_support_ _akka_http

Java:
json_support_ _akka_http

ktoso:
I would be interested in including this in the upcoming release if not much discussion about the additional docs :)

mwkohout:
thanks ktoso!

PR #1974, opened 2018-04-03
Use new sbt-api-mappings plugin version

2m:
New version does not override apiMappings when they come from dependencies. So scaladoc links to akka classes now go to https://doc.akka.io/api/akka/2.5.11/ and links to java dependency (like typesafe-config) classes go to sonatype mirror.


raboof:
PLS BUILD

PR #1981, opened 2018-04-06
Remove unused imports by scalafix

jiminhsieh:


jlprat:
OK TO TEST

raboof:
Right, so there's a manual step in deciding which imports actually do need to be removed. Would be neat if we could somehow mark those.

PR #1993, opened 2018-04-19
make sure static method style works for AllDirectives

ktoso:
Expected to fail, yet should work.
We need to fix this as we document this style as being an available one to use.

Reported in: https://discuss.lightbend.com/t/why-are-routing-directives-not-static-methods-in-java/757/3

Likely a problem on Scala 2.12 and static forwarders, which is weird since it is a class + companion so it should have worked

### DIFF
--- a/akka-http-tests/src/test/java/akka/http/javadsl/server/JavaRouteTest.java
+++ b/akka-http-tests/src/test/java/akka/http/javadsl/server/JavaRouteTest.java
@@ -162,6 +162,13 @@ public class JavaRouteTest extends JUnitRouteTest {
   }
 
   @Test
+  public void allowUsingStaticMethodStyleForDirectives() {
+    Route route = AllDirectives.path("hello", () -> complete("world"));
+    runRoute(route, HttpRequest.GET("/hello"))
+      .assertEntity("world")
+  }
+
+  @Test
   public void requiredParamCauses404OnSealedRoute() {
     runRoute(route, HttpRequest.GET("/cookies"))
       .assertStatusCode(StatusCodes.NOT_FOUND)


ktoso:
Issue: https://github.com/akka/akka-http/issues/1994

ktoso:
I have not yet, would welcome help in doing so -- trying to focus tomorrow on only GDPR before my week off

raboof:
I'll boil this down to a minimal example and check with the Scala team

ktoso:
Hah nice catch... that's easy to miss: In JavaDSL we have `AllDirectives` class and `Directives extends AllDirectives` I can't remember why we did this change in name there to be honest... Was a while since I looked at JavaDSL so I has assumed "companion style naming".

I can confirm it works correctly when `Directives.path` is used. As follow up for this I suppose we should do some more clarification in docs about this?

raboof:
`Directives` is marked `@InternalApi` though, so that suggests we don't want to recommend it? Or do we want to open it up? We seem to recommend `extends AllDirectives`, but I agree importing static methods seems nicer...

ktoso:
> Directives is marked @InternalApi though, so that suggests we don't want to recommend it? Or do we want to open it up? 

That's a mistake or ancient thingy I think.

> We seem to recommend extends AllDirectives, but I agree importing static methods seems nicer...

yet we also say: `public class JavaTestServer extends AllDirectives { // or import static Directives.*;`


Seems this was not cleaned up I guess. Esp. for java the static ones make a lot of sense.

I can amend these in this PR

raboof:
The naming is still weird though... I guess that's due to naming clashes?

Should we update more of the docs to add explicit static imports instead of extending `AllDirectives`? Nicer to read, no multiple inheritance limitations, and somewhat more consistent since the object is called `Directives` at the Scala side as well?

ktoso:
Yeah we can do that, makes much sense in Java anyway.

raboof:
Do you want to pick that up or shall I take this one on?

Issue #2001, opened 2018-04-30
ArrayIndexOutOfBoundsException in HttpHeaderParser


gregbeech:
This is a _terrible_ bug report because other than the exception itself I don't have the details of what caused this error and no reproducing case. We've seen it a couple of times in our production environment, likely as a result of people sending deliberately malformed header content.

I realise it's going to be very hard to investigate without a repro so feel free to close this, but thought I'd report it anyway.

```
java.lang.ArrayIndexOutOfBoundsException: 762
    at akka.http.impl.engine.parsing.HttpHeaderParser.parseHeaderValue(HttpHeaderParser.scala:194)
    at akka.http.impl.engine.parsing.HttpHeaderParser.parseHeaderLine(HttpHeaderParser.scala:130)
    at akka.http.impl.engine.parsing.HttpMessageParser.parseHeaderLines(HttpMessageParser.scala:125)
    at akka.http.impl.engine.parsing.HttpMessageParser.parseHeaderLines$(HttpMessageParser.scala:116)
    at akka.http.impl.engine.parsing.HttpRequestParser$$anon$1.parseHeaderLines(HttpRequestParser.scala:43)
    at akka.http.impl.engine.parsing.HttpRequestParser$$anon$1.parseMessage(HttpRequestParser.scala:76)
    at akka.http.impl.engine.parsing.HttpMessageParser.startNewMessage(HttpMessageParser.scala:100)
    at akka.http.impl.engine.parsing.HttpMessageParser.startNewMessage$(HttpMessageParser.scala:98)
    at akka.http.impl.engine.parsing.HttpRequestParser$$anon$1.startNewMessage(HttpRequestParser.scala:43)
    at akka.http.impl.engine.parsing.HttpMessageParser.$anonfun$startNewMessage$1(HttpMessageParser.scala:101)
    at akka.http.impl.engine.parsing.HttpMessageParser.$anonfun$startNewMessage$1$adapted(HttpMessageParser.scala:101)
    at akka.http.impl.engine.parsing.HttpMessageParser.$anonfun$continue$2(HttpMessageParser.scala:265)
    at akka.http.impl.engine.parsing.HttpMessageParser.run$1(HttpMessageParser.scala:64)
    at akka.http.impl.engine.parsing.HttpMessageParser.parseBytes(HttpMessageParser.scala:78)
    at akka.http.impl.engine.parsing.HttpMessageParser.parseBytes$(HttpMessageParser.scala:62)
[...]

PR #2002, opened 2018-04-30
Fuzz test the header parsing code

raboof:
Attempt to reproduce #2001, but didn't find it
(but haven't tested large numbers of test inputs yet)


jrudolph:
We would probably need something more like Google's https://github.com/google/oss-fuzz where each target project needs to provide a set of seed inputs and fuzzing starts from changing those. With completely random input it's pretty unlikely that the test finds a way into the more interesting states of the parser.

PR #2004, opened 2018-05-02
Commonize marshalling selection code

catalin-ursachi:
Just a straightforward commonization of duplicated code.


akka-http/src/main/scala/akka/http/scaladsl/marshalling/PredefinedToResponseMarshallers.scala
```diff
@@ -105,24 +106,13 @@ trait LowPriorityToResponseMarshallerImplicits {
           val availableMarshallingsPerElement = source.mapAsync(1) { t ⇒ m(t)(ec) }
 
           // TODO optimise such that we pick the optimal marshalling only once (headAndTail needed?)
```
catalin-ursachi: I can't really see how this could be achievable. So long as `m.apply` returns a `List[Marshalling[_]]` for each input, it seems inevitable that it'll have to go through the list and find the first adequate marshaller. For that purpose, the code below looks as efficient as could be.

akka-http/src/main/scala/akka/http/scaladsl/marshalling/PredefinedToResponseMarshallers.scala
```diff
@@ -105,24 +106,13 @@ trait LowPriorityToResponseMarshallerImplicits {
...
-                  case Marshalling.WithOpenCharset(`bestMT`, marshal)    ⇒ () ⇒ marshal(bestCS)
-                }
-            }
-
-            // TODO we could either special case for certrain known types,
+            // TODO we could either special case for certain known types,
             // or extend the entity support to be more advanced such that it would negotiate the element content type it
             // is able to render.
-            selectedMarshalling.getOrElse(throw new NoStrictlyCompatibleElementMarshallingAvailableException[T](s.contentType, marshallings))
+            selectMarshallingForContentType(marshallings, s.contentType)
```
catalin-ursachi: `Marshal.scala` also does:
```
  orElse {
    marshallings collectFirst { case Marshalling.Opaque(marshal) ⇒ marshal }
  }
```
(https://github.com/akka/akka-http/pull/2004/files#diff-05f3ff570222d55bae3d318c06467f23R65)

Is there any reason why that's not also done here?

akka-http/src/main/scala/akka/http/scaladsl/marshalling/Marshal.scala
```diff
@@ -16,6 +16,18 @@ object Marshal {
 
   final case class UnacceptableResponseContentTypeException(supported: Set[ContentNegotiator.Alternative])
     extends RuntimeException with NoStackTrace
+
+  private[marshalling] def selectMarshallingForContentType[T](marshallings: Seq[Marshalling[T]], contentType: ContentType): Option[() ⇒ T] = {
+    contentType match {
+      case best @ (_: ContentType.Binary | _: ContentType.WithFixedCharset | _: ContentType.WithMissingCharset) ⇒
```
catalin-ursachi: `best` is a bit of a misnomer here, anyway. Could get rid of it and simply have:
```
  contentType match {
    case _: ContentType.Binary | _: ContentType.WithFixedCharset | _: ContentType.WithMissingCharset ⇒
      marshallings collectFirst { case Marshalling.WithFixedContentType(`contentType`, marshal) ⇒ marshal }
    case ContentType.WithCharset(mediaType, charset) ⇒
      marshallings collectFirst {
        case Marshalling.WithFixedContentType(`contentType`, marshal) ⇒ marshal
        case Marshalling.WithOpenCharset(`mediaType`, marshal)    ⇒ () ⇒ marshal(charset)
      }
  }
```

raboof:
OK TO TEST

raboof:
PLS BUILD

akka-http/src/main/scala/akka/http/scaladsl/marshalling/Marshal.scala
```diff
@@ -16,6 +16,18 @@ object Marshal {
...
+        marshallings collectFirst { case Marshalling.WithFixedContentType(`best`, marshal) ⇒ marshal }
+      case best @ ContentType.WithCharset(bestMT, bestCS) ⇒
+        marshallings collectFirst {
+          case Marshalling.WithFixedContentType(`best`, marshal) ⇒ marshal
+          case Marshalling.WithOpenCharset(`bestMT`, marshal)    ⇒ () ⇒ marshal(bestCS)
+        }
+    }
```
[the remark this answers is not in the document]
raboof: I agree 'best' just adds confusion here, that proposed change looks neat!

akka-http/src/main/scala/akka/http/scaladsl/marshalling/PredefinedToResponseMarshallers.scala
raboof: Looks like it'd make sense to do that here as well. @jrudolph WDYT?

jrudolph: I agree, otherwise `Opaque` marshallers won't work here.

catalin-ursachi: Okay; I'll have to add a test for this as well, right?

akka-http/src/main/scala/akka/http/scaladsl/marshalling/PredefinedToResponseMarshallers.scala
jrudolph: Agreed. The issue is that the current generic streaming support doesn't seem to be as useful as we thought initially. The same functionality would probably be simpler and faster if implemented directly in the target marshallers. For example, we could provide a `ToEntityMarshaller[Source[T]]` in akka-http-marshallers-spray-json directly.

akka-http/src/main/scala/akka/http/scaladsl/marshalling/PredefinedToResponseMarshallers.scala
jrudolph: Would be good, indeed. Would also be fine to create a ticket to add the test later. This PR would still be an improvement.

catalin-ursachi: Okay; added tests that check that an opaque marshaller will be used if and only if a non-opaque marshaller isn't available.

akka-http/src/main/scala/akka/http/scaladsl/marshalling/PredefinedToResponseMarshallers.scala
catalin-ursachi: Okay; I've removed the TODO - is that right?

jrudolph: That's fine

akka-http-tests/src/test/scala/akka/http/scaladsl/server/EntityStreamingSpec.scala
```diff
@@ -190,6 +190,74 @@ class EntityStreamingSpec extends RoutingSpec with ScalaFutures {
     }
   }
 
+  "opaque-marshaller-example" in {
```
jrudolph: Did that one fail before?

catalin-ursachi: Yup (I wrote the test first and it was failing).

jrudolph: Great, thanks!

jrudolph:
PLS BUILD

jrudolph:
Thanks a lot, @catalin-ursachi!

PR #2006, opened 2018-05-04
WIP: Multipart form directive

daddykotex:
Given a list of FileField, the directive will look for parts with a name equal to the FileField#fieldName.
These parts will be streamed on disk in the file returned by FileField#fileNameF call. All the other,
non-binary and strict part will be gathered in a Map[String, List[String]] similar to what
FormFieldDirectives#formFieldMultiMap returns.

All other parts will be discarded.

Given `Seq("file" -> (_ => new File("/tmp/content.txt")))` and the following request:

```
POST /submit
Content-Type: multipart/form-data; boundary=------------------------boundary
--------------------------boundary
Content-Disposition: form-data; name="name"
Content-Type: text/plain; charset=UTF-8

name_value
--------------------------boundary
Content-Disposition: form-data; name="version"
Content-Type: text/plain; charset=UTF-8

version_value
--------------------------boundary
Content-Disposition: form-data; name="unannounced"; filename="dicarded.txt"
Content-Type: text/plain; charset=UTF-8

discarded
--------------------------boundary
Content-Disposition: form-data; name="file"; filename="content.txt"
Content-Type: text/plain; charset=UTF-8

file content
--------------------------boundary--
```

The map would be: `Map("name" -> List("name_value"), "version" -> List("version_value"))`
The list would be `List(FileInfo("file", "content.txt", Content.Type.`text/plain`) -> File("/tmp/content.txt"))`

### DIFF
--- a/akka-http-caching/src/main/java/akka/http/caching/javadsl/Cache.java
+++ b/akka-http-caching/src/main/java/akka/http/caching/javadsl/Cache.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2017-2018 Lightbend Inc. <https://www.lightbend.com>
+ * Copyright (C) 2017-2019 Lightbend Inc. <https://www.lightbend.com>
  */
 
 package akka.http.caching.javadsl;
--- a/akka-http-caching/src/main/scala/akka/http/caching/CacheJavaMapping.scala
+++ b/akka-http-caching/src/main/scala/akka/http/caching/CacheJavaMapping.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2017-2018 Lightbend Inc. <https://www.lightbend.com>
+ * Copyright (C) 2017-2019 Lightbend Inc. <https://www.lightbend.com>
  */
 
 package akka.http.caching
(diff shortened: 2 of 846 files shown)


daddykotex:
Travis gave:
```
[error]  * abstract method formAndFiles(scala.collection.immutable.Seq)akka.http.scaladsl.server.Directive in trait akka.http.scaladsl.server.directives.MultipartDirectives is inherited by class Directives in current version.
[error]    filter with: ProblemFilters.exclude[InheritedNewAbstractMethodProblem]("akka.http.scaladsl.server.Directives.formAndFiles")
```
I'm not sure how new code makes this binary incompatible.

johanandren:
The reason it is potentially binary incompatible is that `akka.http.scaladsl.server.directives.MultipartDirectives` is a trait which users may have subclassed/mixed into their own classes, where they could potentially already have a conflicting method.

Not sure how we have done with new directive methods before. @raboof ?

raboof:
OK TO TEST

raboof:
Hmm, in https://github.com/akka/akka-http/pull/1828/files#diff-390e61628477725ff016449aa751dea4R25 we ignored it, though I'm not completely sure whether that was an oversight

daddykotex:
I'm sorry I thought I would receive email for this PR, but I did not.

Ok, I understand the binary compatibility issue. But I don't know how to fix it, if I don't add it to the `Directives` trait, then it will hurt usability IMO.

daddykotex:
Also if this is something that makes senses, I can add documentation regarding this new `directive`.

PR #2007, opened 2018-05-04
Add JMH benchmark for header parsing

raboof:


akka-http-bench-jmh/src/main/scala/akka/http/impl/engine/HeaderParserBenchmark.scala
```diff
@@ -0,0 +1,63 @@
...
+
+  @TearDown
+  def tearDown(): Unit = {
+    Await.result(system.terminate(), 5.seconds)
+  }
+
+  @Benchmark
+  def bench_parse_headers(): Unit = {
+    val next = parser.parseHeaderLine(requestBytes, firstHeaderStart)()
+    parser.parseHeaderLine(requestBytes, next)()
```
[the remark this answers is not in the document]
raboof: Judging by the results it didn't, but added a check to make sure :+1: .

akka-http-bench-jmh/src/main/scala/akka/http/impl/engine/HeaderParserBenchmark.scala
```diff
@@ -58,6 +58,7 @@ class HeaderParserBenchmark {
   @Benchmark
   def bench_parse_headers(): Unit = {
     val next = parser.parseHeaderLine(requestBytes, firstHeaderStart)()
-    parser.parseHeaderLine(requestBytes, next)()
+    val last = parser.parseHeaderLine(requestBytes, next)()
+    require(last == 84, s"Expected to read until 84, but read until $last")
```
ktoso: this is still incorrect, you have to return the value (then it is blackholed) or blackhole it by using an explicit one.

Refer to https://github.com/ktoso/sbt-jmh/blob/d20ad9145da005deba2442055cdb53216c339e56/plugin/src/sbt-test/sbt-jmh/run/src/main/scala/org/openjdk/jmh/samples/JMHSample_09_Blackholes.scala#L85 

`measureRight*` vs `measureWrong`

ktoso: You can also skim http://java-performance.info/jmh/ which is a softer introduction

richdougherty:
Thanks @raboof, I can see this will help with #2005. :smile: I'll rebase to pick this up.

We'll also want to test performance when we a custom media type for `application/json` but receive a value that _doesn't_ match the media-type, e.g. `text/plain`. I can add this case if you want.

akka-http-bench-jmh/src/main/scala/akka/http/impl/engine/HeaderParserBenchmark.scala
```diff
@@ -0,0 +1,65 @@
...
+    )
+  }
+
+  @TearDown
+  def tearDown(): Unit = {
+    Await.result(system.terminate(), 5.seconds)
+  }
+
+  @Benchmark
+  def bench_parse_headers(): Int = {
```
ktoso: good now it is correct, though keeping that `require()` will have quite an impact there since allocates functions for the thingy doesn't it? I would suggest removing the require but LGTM otherwise

richdougherty: If I want to check a value I usually put it into a field and check it on teardown as a sanity check. That should work, right @ktoso, JMH guru?

ktoso: I guess, though could impact the benchmark since GCs are not forced between iterations AFAIR, so this could cause more garbage. IMHO when building benchmark have the checks you need, but then remove them to avoid accidental noise

ktoso:
LGTM

ktoso:
Assign PR to milestone for trackability when was merged when merging such things please?

Issue #1768, opened 2018-01-11
Support maxKeepAliveRequests on akka http server


kstokoz:
It would be great if akka http server would provide configuration options to set maxKeepAliveRequests per connection, so that after connection served that amount of requests akka http would force connection close.

It is very useful when using load-balance, for web service, load-balance always use connection based strategy, and when new node added in, server need a chance to re-balance the load.
Currently keepalive connection will be kept for ever, so there is no new connection established, on going connection can not be re-balanced.

When running in kubernetes load balancer would not see how many nodes running in the pod (they can be potentially hidden behind single cluster ip), so scaling up would be useless as traffic would not be rebalanced.

Many http server already support such option:
 - nginx: http://nginx.org/en/docs/http/ngx_http_core_module.html#keepalive_requests
 - tomcat: https://tomcat.apache.org/tomcat-7.0-doc/config/http.html 
 - apache: https://httpd.apache.org/docs/2.4/mod/core.html#maxkeepaliverequests
 - jetty (not supported): https://bugs.eclipse.org/bugs/show_bug.cgi?id=437636

I tried as suggested in jetty bug to send connection close header in my route and it does work.
I cannot control which exactly connection i will close with it as it is abstracted away by akka http and i`m not sure how sending connection close header will affect request pipelining.

It would be great if akka http would support such configuration option out of the box.

jrudolph:
Yes, good idea.

(That said, imo load balancing decisions should be taken at the load balancer not at the upstream backend. The load balancer is free at any time to close an upstream connection and that's the usual solution for these kinds of problems. Having such a limit could still be worthwhile for force thumb clients to reconnect regularly.)

zsedem:
Opened a PR for this
[#2016](https://github.com/akka/akka-http/pull/2016)

iamthiago:
Hi guys, I just came across this situation using `cachedHostConnection` on kubernetes. Would `Http().single` be too bad for several requests? I'm trying to think in a workaround as the above PR is still on it's way to get merged.

zsedem:
Http().singleRequest does use the same host connection pool

iamthiago:
@zsedem so changing from `cachedHostConnection` to `singleRequest` would make no difference?!

zsedem:
Just quoting this: https://doc.akka.io/docs/akka-http/current/client-side/request-level.html#request-level-client-side-api

[...]

Issue #1226, opened 2017-06-23
DNS positive-ttl is not honored by Akka Http Client


wojda:
We have found that under some circumstances, Akka's Http client is not honoring the positive-ttl expiry value, not picking up new DNS entries.

It looks as if under load, and using the default http connection pool, the client will never try resolving again the DNS entry if the connection is not closed, regardless of the positive-ttl value.

**Steps to reproduce:**
0. Using akka-http 10.0.6 and akka-core 2.5.2
1. DNS resolves `test.com` to `server_A` with `ip_A`
2. Run akka http application with following settings:
`dns.inet-address {
	positive-ttl = 30s
	negative-ttl = 30s
}`
3. Run load continuously to this akka http app which does requests to `test.com`
4. Change DNS entry (in `/etc/hosts`, for instance) to point to `ip_B`. NOTE: `server_A` with `ip_A` is still running.
5. Wait for positive-ttl dns cache expiry (30 seconds, in this example)

**Expected behaviour:**
- DNS cache expires, every new request should be sent to `ip_B`.

**Current behaviour:**
- New requests after DNS expiry time are still going to `ip_A`. 
- The only way to have the akka http application to pick up the new DNS entry is by restarting.

jrudolph:
That seems to be the case because akka's DNS resolver is based on JVM's `InetAddress.getAllByName` which introduces another layer of caching.

You can already observe the behavior by just using `java.net.InetAddress.getAllByName("...")` and changing `/etc/hosts` entries in between.

It seems the JVM DNS caching layer is configured using `java.security.Security` properties which are defined in a security file if a `SecurityManager` is installed, otherwise it can be overridden (or turned of in this case) by setting this JVM property using `-Dsun.net.inetaddr.ttl=0`. Can you see if that works for you?

jrudolph:
See also https://docs.oracle.com/javase/8/docs/technotes/guides/net/properties.html#nct and http://www.myhowto.org/java/42-understanding-host-name-resolution-and-dns-behavior-in-java/ and https://stackoverflow.com/questions/1256556/any-way-to-make-java-honor-the-dns-caching-timeout-ttl#1256609

wojda:
Thank you @jrudolph for a quick response. The problem is not related to JVM's DNS resolver unfortunately, that would be something easy to fix. 
[...]

Issue #1886, opened 2018-02-26
Failed: ClientServerSpec.The low-level HTTP infrastructure should produce a useful error message when connecting to a HTTP endpoint over HTTPS


2m:
```
java.lang.AssertionError: timeout (3 seconds) waiting for 6 messages on WarningFilter(None,Right(Perhaps this was an HTTPS request sent to an HTTP endpoint),false)
	at akka.testkit.EventFilter.intercept(TestEventListener.scala:114)
	at akka.http.scaladsl.ClientServerSpec$$anonfun$1$$anonfun$apply$mcV$sp$16$$anonfun$apply$mcV$sp$17.apply$mcV$sp(ClientServerSpec.scala:719)
	at akka.http.scaladsl.ClientServerSpec$$anonfun$1$$anonfun$apply$mcV$sp$16$$anonfun$apply$mcV$sp$17.apply(ClientServerSpec.scala:713)
	at akka.http.scaladsl.ClientServerSpec$$anonfun$1$$anonfun$apply$mcV$sp$16$$anonfun$apply$mcV$sp$17.apply(ClientServerSpec.scala:713)
	at akka.stream.testkit.Utils$.assertAllStagesStopped(Utils.scala:34)
	at akka.http.scaladsl.ClientServerSpec$$anonfun$1$$anonfun$apply$mcV$sp$16.apply$mcV$sp(ClientServerSpec.scala:713)
	at akka.http.scaladsl.ClientServerSpec$$anonfun$1$$anonfun$apply$mcV$sp$16.apply(ClientServerSpec.scala:713)
	at akka.http.scaladsl.ClientServerSpec$$anonfun$1$$anonfun$apply$mcV$sp$16.apply(ClientServerSpec.scala:713)
	at org.scalatest.OutcomeOf$class.outcomeOf(OutcomeOf.scala:85)
	at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
	at org.scalatest.Transformer.apply(Transformer.scala:22)
	at org.scalatest.Transformer.apply(Transformer.scala:20)
...
```

https://jenkins.akka.io:8498/job/akka-http-nightly-against-akka-master-on-2.11/458/testReport/junit/akka.http.scaladsl/ClientServerSpec/The_low_level_HTTP_infrastructure_should_produce_a_useful_error_message_when_connecting_to_a_HTTP_endpoint_over_HTTPS/

jrudolph:
[...]

PR #2016, opened 2018-05-15
add new option to close connections after a period of time #1768

zsedem:
# Use case
In kubernetes when a service replica count changes. Our akka-http-clients needs some kind of eventual rebalancing between pods, so the new pod gets some of the requests.
With this aproach the DNS is resolved again after the client reconnects, so it might creates a workaround for #1226


lightbend-cla-validator:
Hi @zsedem,

raboof:
OK TO TEST

raboof:
@zsedem thanks for this contribution! As you may have noticed our automated binary compatibility checker flagged this change as potentially binary incompatible. Because this class is marked `@DoNotInherit`, however, adding the method is safe and you can add an exception as documented in our CONTRIBUTING.md.

akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/NewHostConnectionPool.scala
```diff
@@ -324,10 +326,19 @@ private[client] object NewHostConnectionPool {
...
+            val max = (settings.maxConnectionKeepAliveTime / 10).toMillis
+            () ⇒ random.nextLong() % max
+          }
           def openConnection(): Future[Http.OutgoingConnection] = {
             if (connection ne null) throw new IllegalStateException("Cannot open connection when slot still has an open connection")
 
             connection = logic.openConnection(this)
+            connection.outgoingConnection.foreach(_ ⇒ if (settings.maxConnectionKeepAliveTime.isFinite()) {
+
+              disconnectAt = Instant.now().getEpochSecond + settings.maxConnectionKeepAliveTime.toMillis + keepAliveDurationFuzziness()
```
raboof: Hmm is this thread-safe? `connection.outgoingConnection.foreach` might not be strictly needed here and make things simpler?

akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/NewHostConnectionPool.scala
```diff
@@ -349,7 +360,13 @@ private[client] object NewHostConnectionPool {
 
           def dispatchResponseResult(req: RequestContext, result: Try[HttpResponse]): Unit = logic.dispatchResponseResult(req, result)
 
-          def willCloseAfter(res: HttpResponse): Boolean = logic.willClose(res)
+          def willCloseAfter(res: HttpResponse): Boolean = {
+            logic.willClose(res) || keepAliveTimeApplies()
```
raboof: This will make the `slot` assume the connection will be closed after this response has been delivered - but will it also actually make sure the connection is closed?

akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/NewHostConnectionPool.scala
zsedem: I guess you are right, that `foreach` does not do much

akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/NewHostConnectionPool.scala
zsedem: After testing it manually it worked like, it would close, but now i thinking about it. I might used a too low idle limit, which means _that_ closed it actually. I will get back to this after doing some manual testing with the new parameter and `idle-timeout`

zsedem:
Well testing is not a trivial for this, I was hoping that i might get some tips here.

akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/NewHostConnectionPool.scala
```diff
@@ -349,7 +360,13 @@ private[client] object NewHostConnectionPool {
...
+          }
+
+          def keepAliveTimeApplies(): Boolean = if (settings.maxConnectionKeepAliveTime.isFinite()) {
+            Instant.now().toEpochMilli - disconnectAt > 0
```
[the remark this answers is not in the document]
zsedem: absolutely

raboof:
@zsedem for testing, perhaps you could give the utlitity also used in #1221 a try? It's a bit experimental still so it'd be good to get feedback on it ;)

akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/NewHostConnectionPool.scala
zsedem: With `idle-timeout` set to `infinite` it is still works

zsedem: I tested with this server: https://gist.github.com/zsedem/7fbab5e62335c907aeafcda3da04de47

raboof:
Failure was  #1886

raboof:
PLS BUILD

akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/NewHostConnectionPool.scala
```diff
@@ -325,10 +327,18 @@ private[client] object NewHostConnectionPool {
 
           def settings: ConnectionPoolSettings = _settings
 
+          private lazy val keepAliveDurationFuzziness: () ⇒ Long = {
+            val random = new Random()
```
raboof: This constructs a new `Random` instance for each connection - could we do that less often?

zsedem: this random is created once per slot.
And slots are created for the pool here _once_ 
https://github.com/akka/akka-http/blob/master/akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/NewHostConnectionPool.scala#L70

raboof: ah, you're right, the `() ⇒ random.nextLong() % max` is created below, I misread :+1:

docs/src/main/paradox/common/timeouts.md
```diff
@@ -80,3 +80,10 @@ Tweaking it should rarely be required, but it allows erroring out the connection
 is unable to be established for a given amount of time.
 
 it can be configured using the `akka.http.client.connecting-timeout` setting.
+
+### Connection Keep Alive timeout
```
raboof: "keep alive" sounds like "extend the life of a connection that would otherwise be closed", while this feature is actually "reduce the life of a connection that would otherwise remain open".

Perhaps we should consider renaming it to "Max connection age" of "Max connection lifetime" (in the setting name as well)? WDYT?

zsedem: Also the "timeout" word does not makes any sense here, so yes lets rename it. (I will have time tomorrow)
max-connection-lifetime sounds fancier :+1:

raboof:
Hmm that failure looks unrelated

raboof:
PLS BUILD

johanandren:
Sorry for the slow progress reviewing @zsedem ! Vacation times.

kstokoz:
Any updates on this issue?

zsedem:
Any updates on this issue?

gregrivera1:
@zsedem or @kstokoz Our team is also affected by this. How did you work around this issue? Did you use a different client or is there some other workaround? We need to decide if akka-http is going to work for us, and wanted to hear other's thoughts.

kstokoz:
I'm waiting on the decision for this pull request, meanwhile, we send connection close header manually for every 100 requests. It does not allow to fine-grained closing connections which were used longer etc. but it works to balance traffic to an acceptable degree.

zsedem:
I simply applied this patch to `10.1.1` and released it into our own artifactory. 
We use it in one of our services, since I opened this PR. (In kubernetes one pod has 300-500 QPS each of the 2 upstream)
There might be an issue with graceful stop, but it does not seem to be related to the patch, because we seen it with the official version too.

gregrivera1:
Thanks @zsedem and @kstokoz for your response. If I am understanding the problem correctly, this problem is specific to using the akka http client to make requests. In theory, is it possible to continue to use akka http to serve requests, and a separate library to make rest calls that is not akka http? 

A rewrite of our system in a different technology is not feasible at this point. But if we did a small rewrite to use a java rest client, it would not be the end of the world. I am trying to determine if there are any other options but we may resort to maintaining our own hosted version as it seems to be the simplest solution. Other than the manual maintenance of course.

zsedem:
@gregrivera1 https://github.com/akka/akka-http/pull/2016#issuecomment-444214333

zsedem:
@gregrivera1 This is an akka-http client bug client and server are completely independent, however they integrate nicely, when you have to stream your requests and responses, but other than that if you have small http request bodies I think there is no advantage of using akka-http client just because you use akka-http server (and vica-versa).

If you choose another http client the important thing is to not block in the akka default threads (use a separate executioncontext for blocking, while waiting http client response)

Also this could be fixed on the server side (in the upstream services) pretty easily if you just send `Connection: close` header in the response randomly. For my company its not an option unfortunately.

zsedem:
For @Maintainers I happy to contribute in any way :+1: 

I do not think this is the perfect implementation and I happy to reimplement it in any other way, which I cannot see right now.

gregrivera1:
Thanks @zsedem! I will wait for the PR, I have some old spray services that are in need of migration and I would be able to reuse most of what exists.

raboof:
(checked that the github 'squash and merge' left the `Author: Adam Oliver Zsigmond <zsedem@gmail.com>` attribution intact)

raboof:
Thanks for this contribution, good to have it in finally!

zsedem:
:fireworks:

PR #2018, opened 2018-05-15
Make ErrorInfo field public in ExceptionWithErrorInfo

kag0:
Tiny change
It would be useful if the exception that wraps `ErrorInfo` exposed that info without needing to pattern match every implementation to get it.
The `override`s in the implementing case classes aren't super pretty, so ideas on how to make that go away are welcome.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/ErrorInfo.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/ErrorInfo.scala
@@ -67,41 +67,41 @@ object ErrorInfo {
 }
 
 /** Marker for exceptions that provide an ErrorInfo */
-abstract class ExceptionWithErrorInfo(info: ErrorInfo) extends RuntimeException(info.formatPretty)
+abstract class ExceptionWithErrorInfo(val info: ErrorInfo) extends RuntimeException(info.formatPretty)
 
-case class IllegalUriException(info: ErrorInfo) extends ExceptionWithErrorInfo(info)
+case class IllegalUriException(override val info: ErrorInfo) extends ExceptionWithErrorInfo(info)
 object IllegalUriException {
   def apply(summary: String, detail: String = ""): IllegalUriException = apply(ErrorInfo(summary, detail))
 }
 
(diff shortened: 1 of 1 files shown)


raboof:
When is the `ErrorInfo` object useful? It's 3 strings (summary, detail and errorHeaderName), are these useful to have separately instead of the `getMessage`?

kag0:
Useful for example when building custom error responses where a user might want to return or log the summary and/or detail separately. Especially consider the comment on the class 

> The summary should explain what is wrong ... *without* directly repeating anything present in the message itself (in order to not open holes for XSS attacks), while the detail can contain additional information from any source (even the request itself).

`getMessage` always contains output from `formatPretty` and therefore potentially parts of the message content.

johanandren:
I would have expected the binary compatibility check to bark on this given that the exception is open for user extension (although maybe that is extremely unlikely?)

raboof:
@kag0 it seems this PR got kind of buried, sorry about that.

Yes, splitting the error in a part that can securely be shared and a part that is only for internal logging makes sense.

Does this mean the default exception handlers for these exceptions also need to be updated to only use the 'safe' part of the error? Or is that already taken care of somehow?

raboof:
Hmm, looking at the default exception handler, https://github.com/akka/akka-http/blob/master/akka-http/src/main/scala/akka/http/scaladsl/server/ExceptionHandler.scala#L52, it seems it already has a special case for `IllegalRequestException`, where it determines based on `settings.verboseErrorMessages` whether to show the full or the short error message.

Perhaps it would be better to widen that exception to work for every `ExceptionWithErrorInfo`, and keep the full error in the `message` to make sure all of it is logged?

One downside of this approach is we're showing the full message by default, and rely on a special case to show the 'safe' message. It would be nicer to be 'safe by default', but I guess that might be too big of a change at this point - or at least beyond the scope of this PR.

kag0:
I'd lean towards defaulting to the safe option. We could sub in some default message if there is no safe message available.

raboof:
Unless I'm mistaken this seems superseded by #2314

PR #2020, opened 2018-05-16
Implement reverseProxy directive (#240)

kag0:
Work in progress to resolve https://github.com/akka/akka-http/issues/240
Lots of credit to https://github.com/akka/akka-http/pull/324

still need to

- [ ] implement `Via` header (how important is this actually for a first version?)
- [x] add tests
- [ ] add more magnets
- [ ] provide javadsl (how important is this for a first version?)

### DIFF
--- a/akka-http-caching/src/main/java/akka/http/caching/javadsl/Cache.java
+++ b/akka-http-caching/src/main/java/akka/http/caching/javadsl/Cache.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2017-2018 Lightbend Inc. <https://www.lightbend.com>
+ * Copyright (C) 2017-2019 Lightbend Inc. <https://www.lightbend.com>
  */
 
 package akka.http.caching.javadsl;
--- a/akka-http-caching/src/main/scala/akka/http/caching/CacheJavaMapping.scala
+++ b/akka-http-caching/src/main/scala/akka/http/caching/CacheJavaMapping.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2017-2018 Lightbend Inc. <https://www.lightbend.com>
+ * Copyright (C) 2017-2019 Lightbend Inc. <https://www.lightbend.com>
  */
 
 package akka.http.caching
--- a/akka-http-caching/src/main/scala/akka/http/caching/LfuCache.scala
+++ b/akka-http-caching/src/main/scala/akka/http/caching/LfuCache.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2017-2018 Lightbend Inc. <https://www.lightbend.com>
+ * Copyright (C) 2017-2019 Lightbend Inc. <https://www.lightbend.com>
  */
 
 package akka.http.caching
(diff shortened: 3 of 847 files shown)


kag0:
Is there a way to run `mimaReportBinaryIssues` locally for a specific scala version?

raboof:
Does `sbt ++2.12.6 mimaReportBinaryIssues` do the trick?

kag0:
Is the current binary incompatibility failure on 2.11 OK? If not, why isn't it present on 2.12, and what's the right way to fix?

kag0:
quick bump on this.
is there still interest in getting this feature in?

odd:
I have successfully been using this directive in my latest project, so I'm quite interested in seeing it included in a future release.

kag0:
@odd glad to hear that. If there is any feedback you have while you're using it, definitely let me know and I'll try to integrate it here.

Gabology:
What would be needed to get some traction on this? Would personally be very interested but for me Java DSL is a must as I don't code in Scala.

raboof:
PLS BUILD

mushtaq:
This is the exact thing I was looking for. Is there anything blocking this pull request. Thanks!

kag0:
> Is the current binary incompatibility failure on 2.11 OK? If not, why isn't it present on 2.12, and what's the right way to fix?

I'm still in need of an answer to this question. Maybe a member can chime in as to whether there is anything aside from that that needs improvement.

kag0:
@Gabology I added java support, does the api look appropriate for you?

kag0:
@raboof I'm not sure how the java dsl directive type hierarchy works. Everything is an abstract class with `BasicDirectives` at the bottom. So should I be making `BasicDirectives extends ReverseProxyDirectives`?

raboof:
@kag0 We recommend people to import the directives with `akka.http.scaladsl.server.Directives._` (scala) or `import static akka.http.javadsl.server.Directives.*;` (java).

The Scala `Directives` mixes in all actual directive implementations as traits [like you already found](https://github.com/akka/akka-http/pull/2020/files#diff-9ef83b15ae379f9f080ae5a1a25ab3d6R40).

Java unfortunately does not have multiple inheritance like this, so instead we create a 'chain' of implementations all extending each other. To add your directives to the 'bottom of the chain' like in Scala, extend your new API's at https://github.com/akka/akka-http/blob/master/akka-http/src/main/scala/akka/http/javadsl/server/Directives.scala#L14, and have them extend the `FramedEntityStreamingDirectives` to connect to the rest of the chain.

Does that make sense?

raboof:
(travis still catches a compile error when compiling with 2.11, btw https://travis-ci.org/akka/akka-http/builds/478845890#L639)

kag0:
That makes sense, took care of the 2.11 issue as well.

raboof:
Closing due to inactivity - feel free to reopen when work on this continues!

kag0:
Looking back through here and it seems there's a lot of interest and even some SWIM using it in production. I should get back on the horse and get this merged.

Where did we leave off? Seems like things should be compiling on 2.11, we have the java api, we might feel OK ignoring the MiMa error. Anything I'm missing?

PR #2024, opened 2018-05-18
Update sbt and sbt-plugins

jiminhsieh:
* sbt-jmh - bump to 0.3.4 for GraalVM
* sbt-paradox-akka - use the same version at Akka
* sbt-header - use the same version at Akka


ktoso:
OK TO TEST

Issue #2025, opened 2018-05-18
Missing parameter with required value directive in Java API


nikom1337:
Hi,

I've found that Scala API has one directive to specify require value of parameter
```
parameters('color, 'action ! "true") { (color) =>
    complete(s"The color is '$color'.")
  }
```
but there is no Java API version of it. 

There is also some [inconsistency](https://github.com/akka/akka-http/blob/b84d8d905339eddb9e5fb6af1f8f2d555e858a7b/akka-http/src/main/scala/akka/http/scaladsl/server/directives/ParameterDirectives.scala#L150-L158) related to parameter directives, most of the filters are rejecting request with clearly defined Rejection (e.g MalformedRequestContentRejection), requiredFilter is using empty rejection.

I would like to suggest something like that:
```
    private def requiredFilter[T](paramName: String, fsou: FSOU[T], requiredValue: Any): Directive0 =
      extractRequestContext flatMap { ctx ⇒
        import ctx.executionContext
        import ctx.materializer
        onComplete(fsou(ctx.request.uri.query().get(paramName))) flatMap {
          case Success(value) if value == requiredValue ⇒ pass
          case Success(value)                           ⇒ reject(InvalidRequiredValueForQueryParamRejection(paramName, requiredValue.toString, value.toString))
          case _                                        ⇒ reject(MissingQueryParamRejection(paramName))
        }
      }
```

jlprat:
Closed via fd94b5d

Issue #1766, opened 2018-01-11
FAILED: NewConnectionPoolSpec: handle 500 pipelined requests with connection termination


ktoso:
```
[info] - be able to handle 500 pipelined requests with connection termination *** FAILED ***
[info]   java.lang.RuntimeException: Failed at pipeliningLimit=2, poolFlow=Flow(FlowShape(MapAsyncUnordered.in(1208931770),MapAsyncUnordered.out(1545925407)))
[info]   at akka.http.impl.engine.client.ConnectionPoolSpec$$anonfun$38$$anon$24$$anonfun$39.apply(ConnectionPoolSpec.scala:495)
[info]   at akka.http.impl.engine.client.ConnectionPoolSpec$$anonfun$38$$anon$24$$anonfun$39.apply(ConnectionPoolSpec.scala:469)
[info]   at scala.collection.Iterator$class.foreach(Iterator.scala:891)
[info]   at scala.collection.AbstractIterator.foreach(Iterator.scala:1334)
[info]   at akka.http.impl.engine.client.ConnectionPoolSpec$$anonfun$38$$anon$24.<init>(ConnectionPoolSpec.scala:469)
[info]   at akka.http.impl.engine.client.ConnectionPoolSpec$$anonfun$38.apply(ConnectionPoolSpec.scala:457)
[info]   at akka.http.impl.engine.client.ConnectionPoolSpec$$anonfun$38.apply(ConnectionPoolSpec.scala:457)
[info]   at org.scalatest.OutcomeOf$class.outcomeOf(OutcomeOf.scala:85)
[info]   at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:22)
[info]   ...
[info]   Cause: java.util.concurrent.TimeoutException: Futures timed out after [35 seconds]
[info]   at scala.concurrent.impl.Promise$DefaultPromise.ready(Promise.scala:223)
[info]   at scala.concurrent.impl.Promise$DefaultPromise.result(Promise.scala:227)
[info]   at scala.concurrent.Await$$anonfun$result$1.apply(package.scala:190)
[info]   at scala.concurrent.BlockContext$DefaultBlockContext$.blockOn(BlockContext.scala:53)
[info]   at scala.concurrent.Await$.result(package.scala:190)
[...]

PR #2026, opened 2018-05-18
Adds missing parameter with required value directive for Java API (#2025)

nikom1337:
and updates documentation

Refs: #2025

### DIFF
--- a/akka-http/src/main/scala/akka/http/javadsl/server/Rejections.scala
+++ b/akka-http/src/main/scala/akka/http/javadsl/server/Rejections.scala
@@ -13,6 +13,7 @@ import java.util.Optional
 import java.util.function.{ Function ⇒ JFunction }
 import java.lang.{ Iterable ⇒ JIterable }
 
+import akka.annotation.DoNotInherit
 import akka.http.scaladsl
 import akka.japi.Util
 import akka.pattern.CircuitBreakerOpenException
@@ -37,6 +38,7 @@ trait CustomRejection extends akka.http.scaladsl.server.Rejection
  * Rejection created by method filters.
  * Signals that the request was rejected because the HTTP method is unsupported.
  */
+@DoNotInherit
 trait MethodRejection extends Rejection {
   def supported: HttpMethod
 }
@@ -45,6 +47,7 @@ trait MethodRejection extends Rejection {
  * Rejection created by scheme filters.
  * Signals that the request was rejected because the Uri scheme is unsupported.
  */
+@DoNotInherit
 trait SchemeRejection extends Rejection {
   def supported: String
 }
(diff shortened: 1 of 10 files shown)


jlprat:
OK TO TEST

akka-http/src/main/scala/akka/http/javadsl/server/Rejections.scala
```diff
@@ -57,6 +57,16 @@ trait MissingQueryParamRejection extends Rejection {
   def parameterName: String
 }
 
+/**
+ * Rejection created by parameter filters.
+ * Signals that the request was rejected because a query parameter value was not equal to required one.
+ */
+trait InvalidRequiredValueForQueryParamRejection extends Rejection {
```
raboof: Perhaps mark with `@DoNotInherit`?

akka-http/src/main/scala/akka/http/scaladsl/server/RejectionHandler.scala
```diff
@@ -200,6 +200,10 @@ object RejectionHandler {
         case MissingQueryParamRejection(paramName) ⇒
           complete((NotFound, "Request is missing required query parameter '" + paramName + '\''))
       }
+      .handle {
+        case InvalidRequiredValueForQueryParamRejection(paramName, requiredValue, actualValue) ⇒
+          complete((NotFound, s"Request is missing required value '$requiredValue' for query parameter '$paramName', actualValue '$actualValue'"))
```
raboof: I think echo'ing a user-provided value in a response is considered a potential security risk (harmless on its own, but could be used in elaborate attach contexts), so perhaps best not to include the `actualValue`?

akka-http/src/main/scala/akka/http/javadsl/server/Rejections.scala
nikom1337: I can mark it as a `@DoNotInherit`, but what with the rest of a traits in Rejections file? I think we should mark all traits or none to keep a consistency.

akka-http/src/main/scala/akka/http/scaladsl/server/RejectionHandler.scala
nikom1337: actualValue is not included in error message anymore + I used new rejectRequestEntityAndComplete method to handle InvalidRequiredValueForQueryParamRejection

akka-http/src/main/scala/akka/http/javadsl/server/Rejections.scala
jlprat: In my opinion, I would put `@DoNotInherit` to all of them

jlprat:
Failure was #1766 

PLS BUILD

jlprat:
@raboof are the changes satisfying your comments? Can we merge?

PR #2029, opened 2018-05-22
Added ability to set https proxy host and port through application.conf

### LINKED ISSUE #995: Allow to configure ClientTransport with application.conf

Scalahansolo:
Re opened this RP after rebasing against master. (Original PR #1425)

Meant to fix #995

Right now, this does not allow for setting auth parameters in the settings as the HttpCredentials class in akka has a few different implementation, including Basic/Oauth/ect. As it includes a few I was able to come up with a concise implementation to get that set. If it would be desired, I can add the ability to add Basic auth at the very least in the settings, or... it can be kept as it and auth must always be passed directly to the httpsProxy method.


raboof:
OK TO TEST

Issue #1852, opened 2018-02-12
Resurrect OSGI support


jrudolph:
OSGI support regularly breaks during releasing (or even used to break the released artifacts without notice), which is the worst time for breakage. We therefore removed OSGI support while releasing 10.1.0-RC1.

If someone cares enough for OSGI support, please step up as a maintainer of Akka HTTP OSGI support. We'll need help with:

 * adding back sbt OSGI support fixing the problem that happened during the last release
 * integrate OSGI checks with the PR validation and/or nightly builds checking that
   * releasing will still work at any time
   * OSGI metadata is correct, i.e. we need executable tests that verify that the emitted metadata is correct and works in an actual OSGI setup
  * ongoing commitment to maintain the OSGI setup and help diagnose and fix OSGI release issues

oheger:
Hi, just to let you know, I have started working on this - especially on automated OSGi integration tests. I have created a repository for doing some experiments [here](https://github.com/oheger/akka-http-osgi-test).
The basic idea is to have a project that depends on akka and akka-http. The sbt build collects the (transitive) dependencies and copies them in a directory. From there they are picked up by a [PaxExam](https://ops4j1.jira.com/wiki/spaces/PAXEXAM4/overview) test and deployed in an Apache Felix OSGi container. The test then starts an HTTP server and executes a request against it.
The advantage of this approach is that not only akka-http but all its dependencies are checked for OSGi compliance (namely the akka bundles). Here I currently face some problems because the OSGi meta data for akka-streams and ssl-config-core is broken (I have created PRs for these issues).
If you agree to this approach, we can discuss how to integrate this test in the build of akka-http; I haven't yet looked into this.

ktoso:
This sounds pretty good; I skimmed the code and I assume it does the right thing... Thanks for linking to the PaxExam page, it seems like it would do the job.

Definitely submit a PR and let's try to integrate soon

ennru:
Future versions of Akka and Akka HTTP will not support OSGi.

PR #2030, opened 2018-05-23
Re-enabled OSGi support (#1852).

oheger:
The commits that removed OSGi support are reverted. The
Require-Capability header has been fixed.


lightbend-cla-validator:
Hi @oheger,

ktoso:
What about the tests you mentioned? I would not want to enable these without in the same PR introducing tests that confirm that this actually works.

oheger:
Yes, I am working on an integration of my experimental project into the akka-http build. However, I am having a hard time as I am not very experienced with sbt. The artifacts behave differently when they are referenced as a project dependency than as an external dependency. Will have to dig deeper.
Also be warned that the tests will not be successful with the current state of dependencies. The modules akka-stream and ssl-config-core have invalid OSGi meta data in their current versions. Only when newer versions are referenced with these problems fixed it will be possible to deploy the akka-http bundles with all their dependencies in an OSGi container which is required for the test.

oheger:
I have now an initial version of the integration test that works as expected. Note that the changes in build.sbt are more or less a proof of concept - there is for sure much room to improve this and make the code cleaner.

When executing the test (by running the osgiTest task in the new akka-http-osgi-tests project) it collects the required dependencies, builds OSGi bundles, and deploys them in a Felix OSGi container. Then it executes a simple test that sends a request against an HTTP server.

Because of the problems with dependencies mentioned before the test currently fails with a ClassNotFoundException. This is due to the fact that some of the bundles could not be started correctly. The (pretty verbose) log of PaxExam shows some errors from Felix about requirements that could not be resolved.

One more thing about PaxExam: The tool is great for writing OSGi integration tests and the only one with this functionality I am aware of. However, it also has its oddities. One problem you might encouter is that it uses a mechanism to resolve some internal dependencies based on the libraries used by Apache Maven. I think, as long as the Maven Central repository can be reached directly via the internet, this should not cause trouble. But if you are behind a corporate proxy, it may be necessary to have a working Maven configuration (a ~/.m2/settings.xml file) that defines the corresponding network options.

ktoso:
Very exciting! We'll dig into it shortly :)

build.sbt
```diff
@@ -185,6 +191,44 @@ lazy val httpTests = project("akka-http-tests")
...
+lazy val copyDeps = taskKey[Unit]("Copies all dependencies in a special directory")
+
+lazy val osgiTest = taskKey[Unit]("Executes OSGi integration tests")
+
+lazy val httpOsgiTests = project("akka-http-osgi-tests")
+  .settings(Dependencies.httpOsgiTests)
+  .settings(
+    fork := true,
+    copyDeps := {
+      val httpFile = (http / Compile / OsgiKeys.bundle).value
```
raboof: This is quite a bit of logic, I wonder if it'd make sense to move it to `project/OSGiTesting.scala`.

build.sbt
```diff
@@ -185,6 +191,44 @@ lazy val httpTests = project("akka-http-tests")
...
+      IO.copy(testDependencies
+        .filter(f => f.getName.contains("scalatest") || f.getName.contains("scalactic") ||
+          f.getName.contains("scala-reflect") || f.getName.contains("scala-xml") ||
+          f.getName.contains("scala-compiler"))
+        .map(f => (f, new File(depsTarget, f.getName))))
+    },
+    osgiTest in Test := Def.sequential(
+      copyDeps,
+      test in Test
+    ).value
```
raboof: I wonder if we could add `copyDeps` as a precondition of `test in Test` for the `akka-http-osgi-tests` project, that way we might not need to introduce the separate `osgiTest` task?

project/OSGi.scala
```diff
@@ -0,0 +1,90 @@
...
+package akka
+
+import com.typesafe.sbt.osgi.OsgiKeys
+import com.typesafe.sbt.osgi.SbtOsgi._
+import sbt.Keys._
+import sbt._
+
+object OSGi {
+
+  // The included osgiSettings that creates bundles also publish the jar files
```
raboof: with `The included osgiSettings` you mean the upstream defaults from the sbt-osgi plugin? But we modify it to go with the other jars? OK I guess.

build.sbt
oheger: As I said, there is certainly much room for improvement. You guys know for sure better how logic can be organized in an sbt build so that it can be easier maintained.

project/OSGi.scala
oheger: The OSGi.scala file was mainly restored to the state before support for OSGi was dropped. This comment was part of the original content. I understand it that the default settings of the OSGi plugin are adapted to generate jars with an OSGi manifest in the default location.

johanandren:
Should we merge this and do the sbt-cleanup on top?

ktoso:
> Should we merge this and do the sbt-cleanup on top?

Sounds good but with adding the testing to travis

johanandren:
I'll merge then and we can continue with that separately.

ktoso:
Ok cool, though please do follow up shortly, as we were hoping to release here very soon and this may be a potentially scary change.

PR #2034, opened 2018-05-28
[+doc] Document entity discarding in default rejection handler

jlprat:
Document that default `RejectionHandler` is not discarding the entity.
Also added release notes for version 10.1.1 that were missing (oops!)

### DIFF
--- a/docs/src/main/paradox/release-notes/10.1.x.md
+++ b/docs/src/main/paradox/release-notes/10.1.x.md
@@ -1,5 +1,66 @@
 # 10.1.x Release Notes
 
+## 10.1.2
+
+10.1.2 is the thrid release in the 10.1.x series of Akka HTTP.
+
+#### Fixes in akka-http-core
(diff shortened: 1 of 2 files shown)


docs/src/main/paradox/release-notes/10.1.x.md
```diff
@@ -1,5 +1,66 @@
...
+ * Minor correction in rejections.md ([#1924](https://github.com/akka/akka-http/issues/1924))
+ * Cache docs: more explicit imports, create cache once ([#1955](https://github.com/akka/akka-http/issues/1955))
+ * Better caching example for Java 
+
+#### Infrastructure and build fixes
+
+ * Fully remove OSGi support ([#1943](https://github.com/akka/akka-http/issues/1943))
+   * Which was already removed and discussed in 10.1.0
+    
+
```
ktoso: Thanks a lot for adding it!!! My bad for missing it I think...

PR #2048, opened 2018-06-01
docs: fix broken links of linking to Scala object

jiminhsieh:
All of Scala API links of `Route` are broken at the document.


raboof:
Looking at https://github.com/akka/akka-http/pull/1903, I wonder if linking to `@unidoc[Route$]` would work? Unfortunate that we currently don't detect this at compile time!

jiminhsieh:
> linking to @unidoc[Route$] would work?

I tried it locally. It does not work for Scala API. And `LfuCache` only has Scala version.` Route` is not in that case. It has scaladsl and javadsl.

If you'd like to fix the unidoc directive, I can close it and open an issue to record this.

raboof:
Let's keep this open for a bit: I'd prefer to fix it in the unidoc plugin, but if we fail to do so quickly we should face that fact and merge this first ;).

jiminhsieh:
> fix unidoc later

How about open an issue for records at least? Someone may help to solve the root cause.

BTW, if I don't misunderstand, this will need to fix at `project/ParadoxSupport.scala`, right?

jlprat:
OK TO TEST

jiminhsieh:
I found we might use `@unidoc[Route] @java[trait]@scala[object]` to solve this issue. It should be working. But I also found that `@scala[object]` is not working. `@scala[object]` will miss `$` too.

The examples of overview at scala version docs have the same issue. 
https://doc.akka.io/docs/akka-http/10.1.3/common/http-model.html?language=scala#overview

jiminhsieh:
The problem of this issue is that `Route` is the different type of class at scaladsl and javadsl. At scaladsl, `Route` is an object. At javadsl, `Route` is a trait.

`HttpMethods ` has the same issue at https://doc.akka.io/docs/akka-http/10.1.3/common/http-model.html?language=scala#overview. At scaladsl, `HttpMethods` is an object. At javadsl, `Route` is java final class.

Could anyone give me some suggesion? I have ideas on how to fix, but I think it's really hacky. Thanks! :)

docs/src/main/paradox/introduction.md
```diff
@@ -111,7 +111,7 @@ JSON support is possible in `akka-http` by the use of Jackson, an external artif
 for details).
 @@@
 
-The @unidoc[Route] created using the Route DSL is then "bound" to a port to start serving HTTP requests:
+The @scala[@scaladoc[Route](akka.http.scaladsl.server.Route$)]@java[@javadoc[Route](akka.http.javadsl.server.Route)] created using the Route DSL is then "bound" to a port to start serving HTTP requests:
```
ktoso: This is not correct for Scala. In Scala the `Route` we talk about here is: 
`  type Route = RequestContext ⇒ Future[RouteResult]`
as defined on: https://github.com/akka/akka-http/blob/master/akka-http/src/main/scala/akka/http/scaladsl/server/package.scala#L11

ktoso: We can link to that instead I propose, just to the source location

ktoso:
Commented about the ScalaDSL, the link should be different

for HttpMethods etc the aproach you did is good :)

docs/src/main/paradox/introduction.md
jiminhsieh: Ah, I got it. Thanks for the correction!

> We can link to that instead I propose, just to the source location

I am thinking of linking to the API document: https://doc.akka.io/api/akka-http/current/akka/http/scaladsl/server/index.html#Route=akka.http.scaladsl.server.RequestContext=%3Escala.concurrent.Future[akka.http.scaladsl.server.RouteResult]

ktoso: That sounds okey, please do and then we merge, thanks

jiminhsieh: I went through all of `@unidoc`s . I also found some other small issues. After fixing those and linking to the correct `Route`, I will push soon. :)

ktoso:
Seems good now, thanks!

PR #2053, opened 2018-06-06
remove misplaced directives

shkoder:
* removes offending scala/java directives

### DIFF
--- a/docs/src/main/paradox/common/http-model.md
+++ b/docs/src/main/paradox/common/http-model.md
@@ -346,8 +346,8 @@ Java
 :   @@snip [CustomHeaderExampleTest.java]($test$/java/docs/http/javadsl/CustomHeaderExampleTest.java) { #header-value-pf }
 
 One can also directly extend @unidoc[CustomHeader] which requires less boilerplate, however that has the downside of
-@scala[matching against @unidoc[RawHeader]]@java[having to deal with converting @unidoc[HttpHeader]] instances scala[not working out-of-the-box, thus limiting its usefulness in the routing layer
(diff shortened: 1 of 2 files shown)


docs/src/main/paradox/common/http-model.md
```diff
@@ -346,8 +346,8 @@ Java
 :   @@snip [CustomHeaderExampleTest.java]($test$/java/docs/http/javadsl/CustomHeaderExampleTest.java) { #header-value-pf }
 
 One can also directly extend @unidoc[CustomHeader] which requires less boilerplate, however that has the downside of
-@scala[matching against @unidoc[RawHeader]]@java[having to deal with converting @unidoc[HttpHeader]] instances scala[not working out-of-the-box, thus limiting its usefulness in the routing layer
-of Akka HTTP]@java[to your custom one]. For only rendering such header however it would be enough.
+matching against @unidoc[RawHeader] having to deal with converting @unidoc[HttpHeader] instances not working out-of-the-box, thus limiting its usefulness in the routing layer
```
raboof: Originally the text here was "matching against `RawHeader` instances not working out-of-the-box, thus limiting its usefulness in the routing layer of Akka HTTP" for Scala and "having to deal with converting `HttpHeader` instances to your custom one" for Java.

I think that's technically correct, and the new sentence introduced in this PR does not make much sense.

I agree the original is not super easy to read either though, perhaps we can come up with a new way to explain this?

shkoder: indeed, I've actually removed this paragraph now as there is a note below describing just that

PR #2058, opened 2018-06-08
Add release 10.1.2 release notes, update to Akka 2.5.13

raboof:


ignasi35:
Bitting nails until this is out! (https://github.com/playframework/playframework/issues/8464) :-)

project/AkkaDependency.scala
```diff
@@ -14,7 +14,7 @@ object AkkaDependency {
   // else if akka version is "default", then the hard coded default will be used (jenkins doesn't allow empty values for config axis)
   // else if akka.version is anything else, then the given version will be used
 
-  val defaultAkkaVersion = "2.5.12"
+  val defaultAkkaVersion = "2.5.13"
```
patriknw: Is this needed? Did we add something that is required for Akka HTTP? Otherwise we should be stay a behind on the Akka versions so that it still works with older versions if someone prefers that. This is even more important now when it is not a transitive dependency.

How was it? We added that mininum version check? Is that automatically picking up the version number from here or is it defined somewhere else also?

raboof: Ah, updating the Akka version was still mentioned in RELEASING.md as a release step but indeed that might not make sense anymore (and would even be somewhat dangerous).

I don't see a version check (though I also remember discussing adding one). 10.1.1 still had version 2.5.11, perhaps we should even revert to that?

patriknw: I think it's more important to get this release out now than testing again with another version. Seems things have been tested with `2.5.12` so let's stick with that.

I think this should be followed up afterwards. Lookin to the version checking again, and document this aspect in RELEASING.

patriknw: and here is the PR for the version checking that was closed without merging https://github.com/akka/akka-http/pull/1880

ktoso:
Branch can be deleted I assume? Please do so so we don't have too many confusing branches in the main repo

PR #2059, opened 2018-06-10
doc Correct pipelining-limit flag name

richardimaoka:
The doc page has a wrong name - `akka.http.server.pipelining-limit` is correct.

https://github.com/akka/akka-http/blob/12e7c98b7f003ed0bfe3ed9ddf51dc2e360a6c00/akka-http-core/src/main/resources/reference.conf#L98


jlprat:
OK TO TEST

Issue #2060, opened 2018-06-11
Enhance Caching Documentation


herzrasen:
I implemented Caching in my application today and found it quite hard to get it running without checking the TestSuite at https://github.com/akka/akka-http/blob/master/akka-http-caching/src/test/scala/akka/http/scaladsl/server/directives/CachingDirectivesSpec.scala

Maybe it is just me or wouldn't it be helpful to 

- add a `keyerFunction` and
- show the mixing in of `CachingDirectives` where `cache()` is used?

johanandren:
Sounds like a good idea to me!

herzrasen:
I could provide a PR in some days if it helps.

johanandren:
Sounds excellent!

PR #2068, opened 2018-06-14
improve caching code sample

herzrasen:
This PR is meant to improve the code snippets provided for the caching module as mentioned in #2060

### DIFF
--- a/docs/src/test/scala/docs/http/scaladsl/server/directives/CachingDirectivesExamplesSpec.scala
+++ b/docs/src/test/scala/docs/http/scaladsl/server/directives/CachingDirectivesExamplesSpec.scala
@@ -22,7 +22,7 @@ class CachingDirectivesExamplesSpec extends RoutingSpec {
     //#cache
     import akka.http.scaladsl.server.RequestContext
     import akka.http.scaladsl.model.Uri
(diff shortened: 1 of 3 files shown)


johanandren:
OK TO TEST

johanandren:
CI failure is just the code formatting check, if you make sure to compile before checking in the code will be automatically formatted for you, so that needs a separate commit now.

herzrasen:
@johanandren ok sorry, forgot that. recompiled the project and committed the updated sources.

docs/src/test/scala/docs/http/scaladsl/server/directives/CachingDirectivesExamplesSpec.scala
```diff
@@ -4,6 +4,7 @@
 
 package docs.http.scaladsl.server.directives
 
+import akka.http.scaladsl.model.HttpMethods
```
jlprat: Usually is better to have the imports within the examples, so they can be copy-pasted and run without needing to guess the right import.

herzrasen: @jlprat you are absolutely right. In this case the import is unused. The first implementation checked if the Request was an HTTP GET. Forgot the remove it.

Issue #2066, opened 2018-06-13
One2OneBidiFlow$OutputTruncationException exception


idzivinskyi:
With migrating to new akka-http (val akka = "2.5.13", val akkaHttp = "10.1.2") I got an error:

```
[system-akka.actor.default-dispatcher-3] ERROR a.a.ActorSystemImpl - Internal server error, sending 500 response
akka.http.impl.util.One2OneBidiFlow$OutputTruncationException: Inner flow was completed without producing result elements for 1 outstanding elements
```

Steps to reproduce:
1. Use code below
2. Open in chrome (or your favorite browser) http://localhost:8080
3. Reload tab 2 times

```
import akka.actor.{Actor, ActorRef, ActorSystem, Props}
import akka.http.scaladsl.Http
import akka.http.scaladsl.server.{Directives, Route}
import akka.pattern._
import akka.stream.ActorMaterializer
import akka.util.Timeout

import scala.concurrent.duration._
import scala.concurrent.{ExecutionContext, Future}
import scala.io.StdIn

object TestOne2OneBidiFlow extends App with Directives {

  implicit val system: ActorSystem = ActorSystem("system")

  implicit val executionContext: ExecutionContext = system.dispatcher

  implicit val materializer: ActorMaterializer = ActorMaterializer()

  implicit val timeout: Timeout = 2 seconds

  class Foo extends Actor {
    override def receive: Receive = {
      case _ =>
        Thread.sleep(1000)
        sender() ! Some(0)
    }
  }

  val foo: ActorRef = system.actorOf(Props(new Foo()))

  val route: Route = get {
    val future = foo ? "test"

    onSuccess(future) { _ =>
      complete("pong")
    }
  }

  val server: Future[Http.ServerBinding] = Http().bindAndHandle(Route.handlerFlow(route), "localhost", 8080)

  server.foreach { b =>
    println(s"Server started on ${b.localAddress.toString}")
  }

  println(s"Server online at http://localhost:8080/\nPress RETURN to stop...")
  StdIn.readLine()
  server.foreach(_.terminate(1 second))
}
```

Question: How handle this case properly?

johanandren:
The trigger for the logged error is not actually reloading the page, but reloading before the previous one had completed, right?

raboof:
Hello @idzivinskyi,

The main problem is in the parameter to `get`:

```
{
    val future = foo ? "test"

    onSuccess(future) { _ =>
      complete("pong")
    }
}
```

[...]

PR #2069, opened 2018-06-14
Avoid eagerly failing connections when request closes

raboof:
Refs #2066

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/server/ServerTerminator.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/server/ServerTerminator.scala
@@ -203,6 +203,11 @@ private[http] final class GracefulTerminatorStage(settings: ServerSettings)
           pendingUserHandlerResponse = false
           push(toNet, response)
         }
+
+        override def onUpstreamFinish(): Unit = {
+          // don't finish the whole bidi stage, just propagate the completion:
+          complete(toNet)
+        }
       })
       setHandler(toUser, new OutHandler {
         override def onPull(): Unit = {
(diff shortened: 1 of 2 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/server/ServerTerminator.scala
```diff
@@ -216,6 +221,11 @@ private[http] final class GracefulTerminatorStage(settings: ServerSettings)
           pendingUserHandlerResponse = true
           push(toUser, request)
         }
+
+        override def onUpstreamFinish(): Unit = {
+          // don't finish the whole bidi stage, just propagate the completion:
+          complete(toUser)
+        }
```
raboof: Do we have to keep track and stop the stage when both flows have completed?

johanandren: That was my first thought as well, if we keep the bidistage alive here, then when _does_ it actually stop

johanandren: Maybe do some debug printing in the stop callback and verify that something else makes it stop?

raboof: `postStop` seems to be called fine. Also https://doc.akka.io/docs/akka/current/stream/stream-customize.html#custom-processing-with-graphstage says "`completeStage()` is equivalent to closing all output ports and cancelling all input ports.", it sounds like this should be sufficient.

raboof:
Verified with a load test (also had to merge in #2070)

akka-http-core/src/main/scala/akka/http/impl/engine/server/ServerTerminator.scala
johanandren: Thanks for verifying!

ktoso:
PLS BUILD

ktoso:
Failure was https://github.com/akka/akka-http/issues/2076 seems unrelated... running again

PR #2070, opened 2018-06-14
=htc #2067 fix memory leak in server terminators

### LINKED ISSUE #2067: Memory leak in 10.1.2
With application heap set to 4G and load at 60 requests per second, server managed to survive about an hour. After that Old Gen was full, Full GC started and paused on timeout with zero reclaimed space. 

Heap dump was taken from server while it still was running under load - from 3GB taken memory, 2.9GB of objects were with strong reachability from GC roots.

From dump, it seems that `MasterServerTerminator` which retains a lot of `HashSet`, `AtomicReference`, `Promises`, `Success`, `GracefulTerminatorStage$ConnectionTerminator`, `HttpRequestParser`, `HttpHeaderParser` or some structure of these all.

Here are the top lines of "Objects by Class" as CSV dump extracted from application heap dump few minutes before it crashed.
```
"Class","Objects","Shallow Size","Retained Size"
"java.util.concurrent.atomic.AtomicReference","4692187","75074992","3059793288"
"scala.collection.immutable.HashSet$HashTrieSet","38461","923064","3059064264"
"scala.collection.immutable.HashSet[]","38461","1333608","3059058624"
"akka.http.impl.engine.server.MasterServerTerminator","1","24","3058762912"
"akka.http.impl.engine.server.MasterServerTerminator$AliveConnectionTerminators","1","16","3058762856"
"scala.collection.immutable.HashSet$HashSet1","126836","3044064","3056645720"
"akka.http.impl.engine.server.GracefulTerminatorStage$ConnectionTerminator","119883","1918128","3053744448"
"scala.concurrent.impl.Promise$DefaultPromise","488574","7817184","3052676704"
"scala.util.Success","486245","7779920","3050509744"
"akka.http.impl.engine.server.GracefulTerminatorStage$$anon$1$$Lambda$1206","119883","2877192","3047993424"
"akka.http.impl.engine.server.GracefulTerminatorStage$$anon$1","119883","12467832","3031698456"
"akka.http.impl.engine.parsing.HttpRequestParser$$anon$1","119883","14385960","1421439848"
"akka.http.impl.engine.parsing.HttpHeaderParser","120011","8640792","1176968680"
"java.lang.Object[]","4704982","316041640","713139749"
"byte[]","5129188","461609984","461609984"
"char[]","120503","421740256","421740256"
"java.lang.String","5004569","120109656","407229632"
"akka.stream.impl.fusing.ActorGraphInterpreter$BatchingActorInputBoundary","239811","32614296","372096144"
"akka.stream.impl.fusing.ActorGraphInterpreter$ActorOutputBoundary","239811","24940344","341388552"
"akka.stream.impl.fusing.GraphInterpreter$Connection[]","4325465","154683888","327745176"
"short[]","120075","185302776","185302776"
"akka.stream.impl.fusing.GraphInterpreter$Connection","4322526","172901040","173061224"
...
```

Maybe I forgot to amend the code during migration?

ktoso:
We forgot to remove the terminators once a connection has terminated.

Resolves #2067

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/server/ServerTerminator.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/server/ServerTerminator.scala
@@ -77,6 +77,31 @@ private[http] final class MasterServerTerminator(log: LoggingAdapter) extends Se
     }
   }
 
+  /**
+   * Removes a previously registered per-connection terminator.
+   * Terminators must remove themselves like this once their respective connection is closed,
+   * otherwise they would leak and remain in the set indefinitely.
+   *
+   * If called with terminator that is not registered, false is returned but it is not an error.
+   *
+   * @return true if the terminator has been successfully removed.
+   */
+  @tailrec def removeConnection(terminator: ServerTerminator)(implicit ec: ExecutionContext): Boolean = {
+    terminators.get() match {
+      case v @ AliveConnectionTerminators(ts) ⇒
+        terminators.compareAndSet(v, v.copy(ts = ts - terminator)) ||
+          removeConnection(terminator) // retry
+
+      case Terminating(deadline) ⇒
+        // the `terminator` that we are being called with can only be one that already was registered,
+        // due to the register call happening during materialization, and the remove call happening during
+        // connection stream completion. Since we are in Terminating state, this means `terminate()` was called,
+        // and all existing terminators were invoked to `terminate()`. So if this is an existing one, we must not call
+        // terminate on it, as it would be the 2nd invocation.
+        false
+    }
+  }
+
   // If a connection attempts to register once termination has started, it will immediately be rejected (though
   // since termination also implies unbinding such new connections should not really happen).
   def terminate(timeout: FiniteDuration)(implicit ex: ExecutionContext): Future[HttpTerminated] = {
@@ -92,7 +117,9 @@ private[http] final class MasterServerTerminator(log: LoggingAdapter) extends Se
         if (terminators.compareAndSet(v, Terminating(timeout.fromNow))) {
           // cause the termination for all connections
           val connectionsTerminated = Future.sequence(ts.map { t ⇒
-            t.terminate(timeout)
+            // termination in general always succeeds, but we make sure here in order
+            // to not accidentally short-circuit terminating the other connection-terminators -- all must be terminated
+            t.terminate(timeout).recover { case _ ⇒ HttpServerTerminated }
           })
           val serverTerminated = connectionsTerminated.map(_ ⇒ HttpServerTerminated)
           termination.completeWith(serverTerminated)
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/server/ServerTerminator.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/server/ServerTerminator.scala
@@ -6,24 +6,21 @@ package akka.http.impl.engine.server
 
 import java.util.concurrent.atomic.AtomicReference
 
-import akka.Done
-import akka.actor.{ ActorSystem, Terminated }
-import akka.annotation.{ DoNotInherit, InternalApi }
+import akka.actor.ActorSystem
+import akka.annotation.InternalApi
 import akka.event.LoggingAdapter
 import akka.http.scaladsl.Http
 import akka.http.scaladsl.Http.{ HttpConnectionTerminated, HttpServerTerminated, HttpTerminated }
-import akka.http.scaladsl.settings.{ RoutingSettings, ServerSettings }
-import akka.http.scaladsl.model.{ HttpRequest, HttpResponse, StatusCodes }
-import akka.stream.scaladsl.BidiFlow
-
-import scala.concurrent.duration._
+import akka.http.scaladsl.model.{ HttpRequest, HttpResponse }
+import akka.http.scaladsl.settings.ServerSettings
 import akka.stream._
+import akka.stream.scaladsl.BidiFlow
 import akka.stream.stage._
 import akka.util.PrettyDuration
 
 import scala.annotation.tailrec
-import scala.concurrent.{ ExecutionContext, Future, Promise }
 import scala.concurrent.duration.{ Deadline, FiniteDuration }
+import scala.concurrent.{ ExecutionContext, Future, Promise }
 import scala.util.{ Failure, Success }
 
 /**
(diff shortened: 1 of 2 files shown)


ktoso:
Building a test will be rather hacky but I can do it... WDYT?

ktoso:
> wondering a bit about contention on the Terminators thing, especially with many short lived connections but that's not a show stopper in any way.

Yeah good point it likely impacts it a bit... I'll open a ticket about it though, I guess we could do it in an actor instead of the atomics dance so it would likely impact the actual hot path less.

ktoso:
Ticketified https://github.com/akka/akka-http/issues/2071

akka-http-core/src/main/scala/akka/http/impl/engine/server/ServerTerminator.scala
```diff
@@ -77,6 +77,27 @@ private[http] final class MasterServerTerminator(log: LoggingAdapter) extends Se
...
+   * @return true if the terminator has been successfully removed.
+   */
+  @tailrec def removeConnection(terminator: ServerTerminator)(implicit ec: ExecutionContext): Boolean = {
+    terminators.get() match {
+      case v @ AliveConnectionTerminators(ts) ⇒
+        terminators.compareAndSet(v, v.copy(ts = ts - terminator)) ||
+          removeConnection(terminator) // retry
+
+      case Terminating(deadline) ⇒
+        terminator.terminate(deadline.timeLeft)(ec)
```
raboof: I think this line needs to be removed - we don't want to trigger this terminator when removing it, right?

akka-http-core/src/main/scala/akka/http/impl/engine/server/ServerTerminator.scala
```diff
@@ -77,6 +77,27 @@ private[http] final class MasterServerTerminator(log: LoggingAdapter) extends Se
...
+        false // termination is in progress already, we did not register but immediately issue termination
```
raboof: Not sure if this comment makes sense here

ktoso: right thanks, removing

akka-http-core/src/main/scala/akka/http/impl/engine/server/ServerTerminator.scala
ktoso: Hm, I had thought it's fine, but you're right I think, here's the problem case which confirms this should be removed I guess:

- remove is issued when the connection stream completes/fails.
- the phase becomes `Terminating()` when terminate is issued on the master terminator
- this is when the user calls terminate on the server binding
- existing connections are still alive, the terminate() call results in all existing connections be called terminate on; this is a CAS into the `Terminating(deadline)` container; val connectionsTerminated = Future.sequence(ts.map { t ⇒ t.terminate(timeout) })
- if they remain alive and the deadline triggers, the the stream is killed and the terminate enters here, with an terminator that already was called on with terminate... this is a 2nd call on the same instance and should not happen. -- you're right, removing this invocation here

akka-http-core/src/main/scala/akka/http/impl/engine/server/ServerTerminator.scala
```diff
@@ -77,6 +77,31 @@ private[http] final class MasterServerTerminator(log: LoggingAdapter) extends Se
     }
   }
 
+  /**
+   * Removes a previously registered per-connection terminator.
+   * Terminators must remove themselves like this once their respective connection is closed,
+   * otherwise they would leak and remain in the set indefinitely.
+   *
+   * If called with terminator that is not registered, false is returned but it is not an error.
```
raboof: I don't think we actually fulfill this promise: we return 'false' only when the connection is not removed because the server (possibly including this connection) is already being terminated.

I'm not sure we need to either, it might be fine to just return `Unit`?

akka-http-core/src/main/scala/akka/http/impl/engine/server/ServerTerminator.scala
```diff
@@ -77,6 +77,31 @@ private[http] final class MasterServerTerminator(log: LoggingAdapter) extends Se
...
+   *
+   * @return true if the terminator has been successfully removed.
+   */
+  @tailrec def removeConnection(terminator: ServerTerminator)(implicit ec: ExecutionContext): Boolean = {
```
viktorklang: Is this ExecutionContext used?

akka-http-core/src/main/scala/akka/http/impl/engine/server/ServerTerminator.scala
ktoso: True, good I'll simplify to be a simple `: Unit` return type

akka-http-core/src/main/scala/akka/http/impl/engine/server/ServerTerminator.scala
ktoso: It's not used, right; all other methods here needed one, though true that this one not; removing param

akka-http-core/src/main/scala/akka/http/impl/engine/server/ServerTerminator.scala
```diff
@@ -92,7 +111,9 @@ private[http] final class MasterServerTerminator(log: LoggingAdapter) extends Se
         if (terminators.compareAndSet(v, Terminating(timeout.fromNow))) {
           // cause the termination for all connections
           val connectionsTerminated = Future.sequence(ts.map { t ⇒
-            t.terminate(timeout)
+            // termination in general always succeeds, but we make sure here in order
+            // to not accidentally short-circuit terminating the other connection-terminators -- all must be terminated
+            t.terminate(timeout).recover { case _ ⇒ HttpServerTerminated }
```
raboof: Perhaps log in this case?

ktoso: okey

akka-http-core/src/main/scala/akka/http/impl/engine/server/ServerTerminator.scala
```diff
@@ -92,7 +111,13 @@ private[http] final class MasterServerTerminator(log: LoggingAdapter) extends Se
         if (terminators.compareAndSet(v, Terminating(timeout.fromNow))) {
           // cause the termination for all connections
           val connectionsTerminated = Future.sequence(ts.map { t ⇒
-            t.terminate(timeout)
+            // termination in general always succeeds, but we make sure here in order
+            // to not accidentally short-circuit terminating the other connection-terminators -- all must be terminated
+            t.terminate(timeout).recover {
+              case ex ⇒
+                log.warning("Ignoring termination failure of {}, failure was: {}", t, ex.getMessage)
```
patriknw: what will `t.toString` be here? a Future with some collection? All connections? Will that be a good log message?

PR #2072, opened 2018-06-14
+doc docs for graceful termination

ktoso:


docs/src/main/paradox/server-side/graceful-termination.md
```diff
@@ -0,0 +1,75 @@
...
+## Graceful termination using `ServerTerminator`
+
+Akka HTTP provides two APIs to "stop" the server, either of them are available via the
+@java[@javadoc[ServerBinding](akka.http.javadsl.ServerBinding)]
+@scala[@scaladoc[ServerBinding](akka.http.scaladsl.Http$$ServerBinding)]
+obtained from starting the server (by using any of the `bind...` methods on the
+@java[@javadoc[Http](akka.http.javadsl.Http)]@scala[@scaladoc[Http](akka.http.scaladsl.HttpExt)] extension).
+
+The first method, called `unbind()` causes the server to *stop accepting new connections*, however any existing
+connections that are still being used will remain active.
```
johanandren: "until the client chooses to close them"?

docs/src/main/paradox/server-side/graceful-termination.md
```diff
@@ -0,0 +1,75 @@
...
+It only unbinds the port on which the http server has been listening. This allows applications to finish streaming any
```
johanandren: "HTTP server"

docs/src/main/paradox/server-side/graceful-termination.md
```diff
@@ -0,0 +1,75 @@
...
+
+Step 2, in flight requests will be handled. If a request is "in-flight" (being handled by user code), it is given `hardDeadline` time to complete.
+ 
+- if user code emits a response within the timeout, then this response is sent to the client
+- if it is a streaming response, it is also mandated that it shall complete within the deadline, and if it does not
+  the connection will be terminated regardless of status of the streaming response. This is because such response could be infinite,
+  which could trap the server in a situation where it could not terminate if it were to wait for a response to "finish".
+    - existing streaming responses must complete before the deadline as well.
+      When the deadline is reached the connection will be terminated regardless of status of the streaming responses.
+- if user code does not reply with a response within the deadline, we produce a special @java[`akka.http.javadsl.settings.ServerSettings.getTerminationDeadlineExceededResponse`]@scala[`akka.http.scaladsl.settings.ServerSettings.terminationDeadlineExceededResponse`]  HTTP response (e.g. `503 Service Unavailable`)
```
johanandren: superfluos comma and double space after class name

docs/src/main/paradox/server-side/graceful-termination.md
```diff
@@ -0,0 +1,75 @@
...
+
+A better and more graceful solution to terminate an Akka HTTP server is to use the
+@java[@javadoc[ServerBinding.terminate(Duration)](akka.http.javadsl.ServerBinding#terminate-java.time.Duration-)]
+@scala[@scaladoc[ServerBinding.terminate(FiniteDuration)](akka.http.scaladsl.Http$$ServerBinding#terminate%28FiniteDuration%29:Future[HttpTerminated])]
+method, which not only performs the unbinding, but also
+handles replying to new incoming requests with (configurable) "terminating" HTTP responses.
+It also allows setting a deadline after which any connections that are still alive will be shut down forcefully.
+More precisely, termination works by following these steps:
+
+Step 1, the server port is unbound and no new connections will be accepted (same as invoking `unbind()`).
```
johanandren: When reading through the steps it seems to me they are in fact not at all steps but just different aspects of the termination, so maybe skip the "Step N," part?

docs/src/main/paradox/server-side/graceful-termination.md
ktoso: ok :)

docs/src/main/paradox/server-side/graceful-termination.md
ktoso: yup

docs/src/main/paradox/server-side/graceful-termination.md
ktoso: thanks

docs/src/main/paradox/server-side/graceful-termination.md
ktoso: Time wise they are executed in such order... so while not explicit steps kind of in this order hm... should I remove `step`? In scaladoc it was like: 

```
   * 2) Handle in-flight request:
   * - if a request is "in-flight" (being handled by user code), it is given `hardDeadline` time to complete,
```

johanandren: Step one is a step, and then 2 but 3 is just what happens as soon as step 1 has, and same for 4 I think. Felt it was somewhat confusing.

ktoso: Ok, I'll remove the steps

docs/src/main/paradox/server-side/graceful-termination.md
```diff
@@ -0,0 +1,76 @@
+# Graceful termination
+
+## Graceful termination using `ServerTerminator`
+
+Akka HTTP provides two APIs to "stop" the server, either of them are available via the
```
patriknw: either vs both?

docs/src/test/java/docs/http/javadsl/server/HttpServerExampleDocTest.java
```diff
@@ -321,6 +322,35 @@ void discardEntityManuallyCloseConnections() {
...
+
+    ServerBinding serverBinding = binding.toCompletableFuture().get(3, TimeUnit.SECONDS);
+
+    // ...
+    // once ready to terminate the server, invoke terminate:
+    CompletionStage<HttpTerminated> onceAllConnectionsTerminated =
+        serverBinding.terminate(Duration.ofSeconds(3));
+
+    // once all connections are terminated,
+    HttpTerminated httpTerminated = onceAllConnectionsTerminated.toCompletableFuture().get();
```
patriknw: blocking `get()` (also without timeout) ?

docs/src/test/java/docs/http/javadsl/server/HttpServerExampleDocTest.java
```diff
@@ -321,6 +322,35 @@ void discardEntityManuallyCloseConnections() {
...
+
+    // - you can invoke coordinated shutdown to tear down the rest of the system:
+    CoordinatedShutdown.get(system).run();
```
patriknw: I'd rather register a task to one of the early phases of `CoordinatedShutdown` and that task would call `serverBinding.terminate`.

docs/src/test/java/docs/http/javadsl/server/HttpServerExampleDocTest.java
```diff
@@ -321,6 +322,35 @@ void discardEntityManuallyCloseConnections() {
...
+    //#graceful-termination
+    ActorSystem system = ActorSystem.create();
+    Materializer materializer = ActorMaterializer.create(system);
+
+
+    CompletionStage<ServerBinding> binding = Http.get(system).bindAndHandle(
+        Directives.complete("Hello world!").flow(system, materializer),
+        ConnectHttp.toHost("localhost", 8080), materializer);
+
+    ServerBinding serverBinding = binding.toCompletableFuture().get(3, TimeUnit.SECONDS);
```
patriknw: Don't do this at home. We should not show such code in documentation, it will be copied.

docs/src/test/java/docs/http/javadsl/server/HttpServerExampleDocTest.java
ktoso: On purpose did not do callbacks since this is "where you start the server" where you often have a blocking read line or something anyway. But ok, can go back to callbacks

docs/src/test/java/docs/http/javadsl/server/HttpServerExampleDocTest.java
ktoso: Not entirely sure where exactly to hook it... can do that as part of https://github.com/akka/akka-http/issues/1210 ?

docs/src/test/java/docs/http/javadsl/server/HttpServerExampleDocTest.java
ktoso: We do show this in many places in the "start a server and wait for binding" so I did so too.
Can go with nesting but it makes it rather weird tbh

docs/src/test/java/docs/http/javadsl/server/HttpServerExampleDocTest.java
ignasi35: If this is just an example, I guess adding a task in the `service-unbind` phase that invokes `serverBinding.terminate` should do.

I think it's what we'd do in Play [instead of using `unbind`](https://github.com/playframework/playframework/pull/8406/files#diff-fdb7a2606b698a5146eb4e044fc6f2deR403).

docs/src/test/java/docs/http/javadsl/server/HttpServerExampleDocTest.java
patriknw: ok

docs/src/test/java/docs/http/javadsl/server/HttpServerExampleDocTest.java
patriknw: this is not when starting the server, why would you terminate right after you have started it. Probably called from elsewhere, an actor, an incoming request, management...

docs/src/test/java/docs/http/javadsl/server/HttpServerExampleDocTest.java
ktoso: Like so then: 


```
    val binding: Future[Http.ServerBinding] =
        Http().bindAndHandle(routes, "127.0.0.1", 8080)

    // ...
    // add a hook for coordinated shutdown
    CoordinatedShutdown(system).addTask(CoordinatedShutdown.PhaseServiceUnbind, "terminate-http")(() ⇒ {
      binding.flatMap { bound ⇒
        bound.terminate(3.seconds)
      } map { _ ⇒ Done }
    })

    // terminating the system will invoke coordinated shutdown,
    // and thus also apply the "terminate-http" phase in the right order
    system.terminate()

```

patriknw: Something like https://github.com/akka/akka-http/issues/1210#issuecomment-338825745 but PhaseServiceUnbind and using terminate instead of unbind.

Perhaps not show CoordinatedShutdown at all yet.
and the blocking on onceAllConnectionsTerminated, just a comment saying that you can terminate the system when onceAllConnectionsTerminated has completed would be enough for now, and then do the proper CoordinatedShutdown

Nothing of this is blocking the release though

ktoso: > Perhaps not show CoordinatedShutdown at all yet.

Yeah, I think so. And next section will explain it properly and we'll likely hook it in by default by then anyway

ktoso: Addressing like the last @patriknw comment suggests then

patriknw: > terminating the system will invoke coordinated shutdown,

this is not correct, one would have to call `CoordinatedShutdown(system).run()`
we might adjust that in the future, see https://github.com/akka/akka/issues/25213

I wouldn't show the `system.terminate()` or  `CoordinatedShutdown(system).run()` in the same snippet as the startup, since it makes no sense to shutdown immediately after starting, it's from elsewhere.

PR #2074, opened 2018-06-14
Fix some warnings

jlprat:
Went through some no-brainer warning fixes within the project

### DIFF
--- a/akka-http-core/src/test/scala/akka/http/scaladsl/GracefulTerminationSpec.scala
+++ b/akka-http-core/src/test/scala/akka/http/scaladsl/GracefulTerminationSpec.scala
@@ -73,7 +73,7 @@ class GracefulTerminationSpec extends WordSpec with Matchers with BeforeAndAfter
 
       // termination should kill all connections within the deadline and complete its whenTerminated by then as well
       // (we give it a second wiggle room)
-      val terminated: Http.HttpTerminated = Await.result(serverBinding.whenTerminated, 3.seconds)
+      val _: Http.HttpTerminated = Await.result(serverBinding.whenTerminated, 3.seconds)
     }
 
     "provide whenTerminated future that completes once server has completed termination (existing connection, no user reply)" in new TestSetup {
@@ -86,8 +86,8 @@ class GracefulTerminationSpec extends WordSpec with Matchers with BeforeAndAfter
 
       r1.futureValue.status should ===(StatusCodes.ServiceUnavailable)
 
-      val terminated1: Http.HttpTerminated = Await.result(terminateFuture, 2.seconds)
-      val terminated2: Http.HttpTerminated = Await.result(serverBinding.whenTerminated, 2.seconds)
+      Await.result(terminateFuture, 2.seconds)
+      Await.result(serverBinding.whenTerminated, 2.seconds)
     }
 
     "provide whenTerminated future that completes once server has completed termination (existing connection, user reply)" in new TestSetup {
@@ -100,7 +100,7 @@ class GracefulTerminationSpec extends WordSpec with Matchers with BeforeAndAfter
 
       r1.futureValue.status should ===(StatusCodes.OK)
 
-      val terminated: Http.HttpTerminated = Await.result(serverBinding.whenTerminated, 3.seconds)
+      val _: Http.HttpTerminated = Await.result(serverBinding.whenTerminated, 3.seconds)
     }
     "provide whenTerminated future that completes once server has completed termination (existing connection, user reply, terminate, no reply)" in new TestSetup {
       val r1 = makeRequest() // establish connection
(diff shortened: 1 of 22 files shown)


akka-http-core/src/test/scala/akka/http/scaladsl/GracefulTerminationSpec.scala
```diff
@@ -162,7 +162,7 @@ class GracefulTerminationSpec extends WordSpec with Matchers with BeforeAndAfter
         r2.futureValue.status should ===(StatusCodes.EnhanceYourCalm) // the injected 503 response
         r2.futureValue.entity.toStrict(1.second).futureValue.data.utf8String should ===("Chill out, man!")
 
-        val terminated: Http.HttpTerminated = Await.result(serverBinding.whenTerminated, 3.seconds)
+        val _: Http.HttpTerminated = Await.result(serverBinding.whenTerminated, 3.seconds)
```
johanandren: If we really need to type-ascribe it we can do it on the right side instead and skip the val part.

akka-http-core/src/test/scala/akka/http/scaladsl/util/FastFutureSpec.scala
```diff
@@ -176,8 +177,12 @@ class FastFutureSpec extends FreeSpec with Matchers {
     testLazily()
   }
 
-  def failF: PartialFunction[Any, Nothing] = PartialFunction(_ ⇒ throw TheException)
```
johanandren: These looks suspiciously like there was a reason they were made like this, but I don't know it. Was the old one more performant/less wasteful as it doesn't really do any matching perhaps?

akka-http-tests/src/test/scala/akka/http/scaladsl/CustomMediaTypesSpec.scala
```diff
@@ -42,7 +42,7 @@ class CustomMediaTypesSpec extends AkkaSpec with ScalaFutures
       val routes = extractRequest { r ⇒
         complete(r.entity.contentType.toString + " = " + r.entity.contentType.getClass)
       }
-      val binding = Http().bindAndHandle(routes, host, port, settings = serverSettings)
+      val _ = Http().bindAndHandle(routes, host, port, settings = serverSettings)
```
johanandren: We either keep it or don't val it at all, I'd say?

akka-http2-support/src/main/scala/akka/http/scaladsl/Http2.scala
```diff
@@ -41,11 +39,12 @@ final class Http2Ext(private val config: Config)(implicit val system: ActorSyste
...
-    interface: String, port: Int = DefaultPortForProtocol,
-    settings:    ServerSettings = ServerSettings(system),
-    parallelism: Int            = 1,
-    log:         LoggingAdapter = system.log)(implicit fm: Materializer): Future[ServerBinding] = {
+    handler:     HttpRequest ⇒ Future[HttpResponse],
+    interface:   String,
+    port:        Int,
+    settings:    ServerSettings,
+    parallelism: Int,
+    log:         LoggingAdapter)(implicit fm: Materializer): Future[ServerBinding] = {
```
johanandren: Default value of log was removed, was it never used?

akka-http-core/src/test/scala/akka/http/scaladsl/util/FastFutureSpec.scala
jlprat: It's deprecated in Scala 2.12, hence the change.

akka-http-core/src/test/scala/akka/http/scaladsl/GracefulTerminationSpec.scala
jlprat: I think we can just skip the `val` part, I was doing some "search and replace"

akka-http2-support/src/main/scala/akka/http/scaladsl/Http2.scala
jlprat: Not for this private method

akka-http-tests/src/test/scala/akka/http/scaladsl/CustomMediaTypesSpec.scala
jlprat: You are right, no need to keep the `val`. I was playing with some search-and-replace expressions :)

akka-http-core/src/test/scala/akka/http/scaladsl/util/FastFutureSpec.scala
jlprat: https://www.scala-lang.org/api/current/scala/PartialFunction$.html#apply[A,B](f:A=%3EB):PartialFunction[A,B]

> (Since version 2.12.5) For converting an ordinary function f to a partial function pf, use val pf: PartialFunction[A, B] = { case x => f(x) }. For creating a new PartialFunction, use an explicit type annotation instead, like in val pf: PartialFunction[Int, String] = { case 1 => "one" }.

johanandren: 👍 thanks for the rationale

jlprat: I also learned it during this warning fix :)

Issue #1966, opened 2018-03-27
h2c support: upgrade insecure connections to http2


raboof:
There are 2 mechanisms for supporting http/2 on insecure connections: with prior knowledge (#1934) or with h2c using the HTTP Upgrade mechanism (http://httpwg.org/specs/rfc7540.html#discover-http).

We're unsure how widely used h2c is and are not currently planning to implement it, but if anyone would express interest or contribute it that would be great!

satansk:
Let me try this :)

dsilvasc:
It looks like h2c is the transport that the [grpc-java](https://github.com/grpc/grpc-java) client uses when connecting to a localhost server.

If you try to use Akka HTTP as a gRPC proxy, even with `akka.http.server.preview.enable-http = on` in your `application.conf`, Akka HTTP fails to handle gRPC requests with:

```
[my-system-akka.actor.default-dispatcher-9] [akka.actor.ActorSystemImpl(my-system)] Illegal request, responding with status '501 Not Implemented': Unsupported HTTP method: PRI
```

Documentation on the `PRI` method:
https://http2.github.io/http2-spec/#rfc.section.11.6

> This method is never used by an actual client. This method will appear to be used when an HTTP/1.1 server or intermediary attempts to parse an HTTP/2 connection preface.

raboof:
It depends on what you set as https://grpc.io/grpc-java/javadoc/io/grpc/netty/NegotiationType.html : TLS is 'normal' HTTPS HTTP2, PLAINTEXT is #1934 and PLAINTEXT_UPGRADE is h2c.

@dsilvasc if you see that message in an Akka HTTP server it suggests the connection is 'raw' HTTP2 as supported by #1934, and not using the h2c upgrade mechanism.

raboof:
@satansk did you have any luck with this so far?

satansk:
@raboof not really, but I'm still working on this.

raboof:
@satansk OK that's great! I'm also planning on having a look, so I'll share and we can compare notes.

raboof:
[...]

PR #2075, opened 2018-06-15
WIP complete h2c

johanandren:
Refs #1966

### DIFF
--- a/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/framing/Http2FrameParsing.scala
+++ b/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/framing/Http2FrameParsing.scala
@@ -10,15 +10,36 @@ import akka.stream.Attributes
 import akka.stream.impl.io.ByteStringParser
 import akka.stream.stage.GraphStageLogic
 import Http2Protocol.FrameType._
-import Http2Protocol.{ ErrorCode, Flags, FrameType, SettingIdentifier }
+import Http2Protocol.{ErrorCode, Flags, FrameType, SettingIdentifier}
 import akka.annotation.InternalApi
-
 import FrameEvent._
+
+import scala.annotation.tailrec
+
+/** INTERNAL API */
+@InternalApi
+private[http] object Http2FrameParsing {
+
+  def readSettings(payload: ByteStringParser.ByteReader): immutable.Seq[Setting] = {
+    @tailrec def readSettings(read: List[Setting]): immutable.Seq[Setting] =
+      if (payload.hasRemaining) {
+        val id = payload.readShortBE()
+        val value = payload.readIntBE()
+        if (isKnownId(id)) readSettings(Setting(SettingIdentifier.byId(id), value) :: read)
+        else readSettings(read)
+      } else read.reverse
+
+    readSettings(Nil)
+  }
+
+}
+
 
 /** INTERNAL API */
 @InternalApi
 private[http2] class Http2FrameParsing(shouldReadPreface: Boolean) extends ByteStringParser[FrameEvent] {
   import ByteStringParser._
+  import Http2FrameParsing._
 
   abstract class Step extends ParseStep[FrameEvent]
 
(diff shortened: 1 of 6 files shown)


akka-http2-support/src/main/scala/akka/http/impl/engine/http2/framing/Http2FrameParsing.scala
```diff
@@ -10,15 +10,36 @@ import akka.stream.Attributes
...
+    @tailrec def readSettings(read: List[Setting]): immutable.Seq[Setting] =
+      if (payload.hasRemaining) {
+        val id = payload.readShortBE()
+        val value = payload.readIntBE()
+        if (isKnownId(id)) readSettings(Setting(SettingIdentifier.byId(id), value) :: read)
+        else readSettings(read)
+      } else read.reverse
+
+    readSettings(Nil)
+  }
```
johanandren: Needed to be opened up to be able to parse the settings frame out of the `Http2-Settings` header that comes with the upgrade

akka-http2-support/src/main/scala/akka/http/scaladsl/Http2.scala
```diff
@@ -103,33 +105,79 @@ final class Http2Ext(private val config: Config)(implicit val system: ActorSyste
...
+
+            val serverLayer: Flow[ByteString, ByteString, Future[Done]] = Flow.fromGraph(
+              Flow[HttpRequest]
+                .watchTermination()(Keep.right)
+                .merge(injectedRequest)
+                .via(Http2Blueprint.handleWithStreamIdHeader(effectiveParallelism)(handler)(system.dispatcher))
+                .joinMat(Http2Blueprint.serverStack(settings, log))(Keep.left))
+
+
+            // TODO do not respond until the potential http2 request entity was consumed?
```
johanandren: I tried this but it deadlocked with nghttp, I'm wondering if the do-not-send-http2 frames part is more about the client than what the server does?

johanandren:
Meh formatting...

johanandren:
Tests green locally, not sure what jenkins is up to with missing http2 test result xml

akka-http2-support/src/main/scala/akka/http/scaladsl/Http2.scala
```diff
@@ -103,33 +105,65 @@ final class Http2Ext(private val config: Config)(implicit val system: ActorSyste
...
-            immutable.Seq[HttpHeader](
-              ConnectionUpgradeHeader,
-              UpgradeHeader,
-              // TODO add attributes from the request
-              UpgradeToOtherProtocolHeader(serverLayer)
+
+        log.debug("Got h2c upgrade request from HTTP/1.1 to HTTP2")
+
+        // https://http2.github.io/http2-spec/#Http2SettingsHeader 3.2.1 HTTP2-Settings Header Field
+        val upgradeSettings = req.headers.collect {
```
raboof: As the spec promises exactly 1 header here, `collect` can become `find` + `map`?

akka-http2-support/src/main/scala/akka/http/scaladsl/Http2.scala
```diff
@@ -103,33 +105,65 @@ final class Http2Ext(private val config: Config)(implicit val system: ActorSyste
...
+
+            val injectedRequest =
+              if (req.method != HttpMethods.OPTIONS)
+                // inject the actual upgrade request with a stream identifier of 1
+                // https://http2.github.io/http2-spec/#rfc.section.3.2
+                Source.single(req.addHeader(Http2StreamIdHeader(1)))
+              else {
+                // No 100% sure about this but from the spec:
+                // If concurrency of an initial request with subsequent requests is important, an OPTIONS request
+                // can be used to perform the upgrade to HTTP/2, at the cost of an additional round trip.
```
raboof: I think the point here is that an `OPTIONS` request will be empty and not 'congest' the first request. I suspect "After the empty line that terminates the 101 response, the server can begin sending HTTP/2 frames. These frames MUST include a response to the request that initiated the upgrade. The HTTP/1.1 request that is sent prior to upgrade is assigned a stream identifier of 1" still holds, even for OPTIONS requests - so we can remove the special case here but perhaps see if we should accept OPTIONS requests in the generated route

akka-http2-support/src/main/scala/akka/http/scaladsl/Http2.scala
raboof: Ah, you quote 'A server MUST NOT upgrade the connection to HTTP/2 if this header field is not present or if more than one is present' below, :+1: then!

akka-http2-support/src/main/scala/akka/http/scaladsl/Http2.scala
```diff
@@ -103,33 +105,65 @@ final class Http2Ext(private val config: Config)(implicit val system: ActorSyste
...
+                )
+              )
             )
-          )
-        )
+
+          case _ ⇒
+
+            // A server MUST NOT upgrade the connection to HTTP/2 if this header field is not present or if more than one is present
+            log.debug("Invalid upgrade request (http2-settings header missing or repeated)")
```
raboof: Perhaps it should still return `handler(req)` though?

akka-http2-support/src/main/scala/akka/http/scaladsl/Http2.scala
johanandren: Ok, I'll remove the special handling then.

akka-http2-support/src/main/scala/akka/http/scaladsl/Http2.scala
raboof: I think it'd make sense to log something (debug seems fine, perhaps info), but continue to handle the request as a normal non-http2 request.

PR #2081, opened 2018-06-19
Fix scheme validation logic to apply rule from start instead of end

### LINKED ISSUE #2080: Uri.from fails for scheme ending in digit
Constructing a `Uri` via `Uri.from(scheme = "s3", path = "image.png")` fails with following exception

```
Invalid URI scheme, unexpected character at pos 1 ('3')
akka.http.scaladsl.model.IllegalUriException: Invalid URI scheme, unexpected character at pos 1 ('3')
	at akka.http.scaladsl.model.IllegalUriException$.apply(ErrorInfo.scala:74)
	at akka.http.scaladsl.model.Uri$.fail(Uri.scala:821)
	at akka.http.scaladsl.model.Uri$.normalizeScheme(Uri.scala:776)
	at akka.http.scaladsl.model.Uri$.apply(Uri.scala:228)
	at akka.http.scaladsl.model.Uri$.from(Uri.scala:244)
```

Compared to that `Uri("s3:image.png")` works as expected. Per [RFC 2396][1], Appendix A: 

> scheme        = alpha *( alpha | digit | "+" | "-" | "." )

The scheme should start with alphabet but can end with digit. Looking at [Uri.normalizeScheme][2] it appears that it applies the rule in reverse!

[1]: http://www.ietf.org/rfc/rfc2396.txt
[2]: https://github.com/akka/akka-http/blob/master/akka-http-core/src/main/scala/akka/http/scaladsl/model/Uri.scala#L767

chetanmeh:
`Uri.normalizeScheme` applies the rule to check for. first letter to be alpha from last instead of start. Change that to apply rule from start

Fixes #2080

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/Uri.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/Uri.scala
@@ -765,10 +765,10 @@ object Uri {
     else sb.toString
 
   private[http] def normalizeScheme(scheme: String): String = {
-    @tailrec def verify(ix: Int = scheme.length - 1, allowed: CharPredicate = ALPHA, allLower: Boolean = true): Int =
-      if (ix >= 0) {
+    @tailrec def verify(ix: Int = 0, allowed: CharPredicate = ALPHA, allLower: Boolean = true): Int =
+      if (ix < scheme.length) {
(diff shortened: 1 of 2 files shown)


raboof:
OK TO TEST

Issue #1883, opened 2018-02-24
Http/2 client-side support


mimani:
Is there any plan to provide HTTP/2 client side support?

I am not sure if this is asked earlier as this seems quite common but could not found when I [searched](https://github.com/akka/akka-http/search?p=1&q=client+http+2&type=Issues&utf8=%E2%9C%93) it. I see there is [server side](https://github.com/akka/akka-http/issues/244) implementation but client side was out of scope that time.

jrudolph:
It seems we haven't tracked it so far indeed. We are planning to provide a client implementation at some point. Nothing definite but it might be later this year.

mushtaq:
@jrudolph is this going to be part of 10.2?

raboof:
No, we expect to release 10.2.0 earlier than implementing HTTP/2 client-side support

mtraul:
Is work for this being done already ? If not, I would love to contribute. Require Http/2 support for Apple APNS support

raboof:
> Is work for this being done already ? If not, I would love to contribute.

No, we haven't started on this - it would be great if you could contribute! While we hope a lot of the work on the server-side implementation can be re-used when implementing the client side, this might still be a non-trivial feature: for example, since HTTP/2 allows multiplexing many parallel requests over a single TCP connection that introduces some interesting opportunities - and challenges!

> Require Http/2 support for Apple APNS support

:+1:

jrudolph:
I had a quick try at this today in #3166. The goal for it would be to be able to get some base for a potential implementation into the code base that would help with contributions. Maybe you'd like to have a look at that and play with it, @mtraul? There's a big list of things that still need to be done, maybe I can finish it to a point where it already makes sense to merge it and then we can ticketify the list of remaining issues.

pete-proton:
thanks, @jrudolph ! are we any closer after this PR https://github.com/akka/akka-http/pull/3166 has been merged?

jrudolph:
Yes, we are closer but not yet far enough to announce something more publicly (won't happen for 10.2.0 but maybe later in the series, we still need to figure out where to put it in our internal roadmap).

The client-side is now somewhat working but no public API is available yet. If you want to play with it, you can use the internal API. https://github.com/jrudolph/akka-http/blob/cd5e40b2b555e5ef6195692dec7c8d67be9464a0/akka-http2-support/src/test/scala/akka/http/scaladsl/Http2ClientApp.scala is a working example that should work with 10.2.0-RC1.

[...]

PR #2085, opened 2018-06-20
h2c via 'Upgrade' mechanism

raboof:
This allows HTTP/1.1 and HTTP/2 to coexist on the same insecure port.
This feature is not widely supported among clients: browsers
generally do not support h2c at all, curl and nghttp somewhat support
it, grpc-java does not.

Unfortunately it is not covered by h2spec, not in the latest version
either.

It would be nice to have some tests, but perhaps we should do that
when we start adding support at the client side? #1883

Refs #1966


raboof:
PLS BUILD

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
```diff
@@ -292,17 +292,14 @@ class HttpExt private[http] (private val config: Config)(implicit val system: Ex
     parallelism:       Int               = 1,
     log:               LoggingAdapter    = system.log)(implicit fm: Materializer): Future[ServerBinding] = {
     val http2enabled = settings.previewServerSettings.enableHttp2
-    if (http2enabled && connectionContext.isSecure && connectionContext.http2 != Never) {
-      log.debug("Binding server using HTTP/2...")
-      Http2Shadow.bindAndHandleAsync(handler, interface, port, connectionContext, settings, parallelism, log)(fm)
-    } else if (http2enabled && !connectionContext.isSecure && connectionContext.http2 == Always) {
-      // We do not support HTTP/2 negotiation for insecure connections (h2c), https://github.com/akka/akka-http/issues/1966
+    if (http2enabled && connectionContext.http2 != Never) {
```
raboof: This part of the change is nontrivial because `connectionContext.http2` by default is `Negotiated`, so after this change the code path at https://github.com/akka/akka-http/pull/2085/files#diff-f0a57b2d6e8f16da4ca0286a011211e1R109 will be taken, which might have a performance impact.

Given the limited support for h2c `Upgrade`, I think we should switch the default to `Never` for insecure connections and `Negotiated` only for secure connections.

WDYT?

johanandren: Sounds like it makes sense.

akka-http-core/src/test/scala/akka/http/impl/engine/rendering/ResponseRendererSpec.scala
```diff
@@ -612,8 +612,8 @@ class ResponseRendererSpec extends FreeSpec with Matchers with BeforeAndAfterAll
             Source.single(ctx),
             renderer.named("renderer")
               .map {
-                case ResponseRenderingOutput.HttpData(bytes)      ⇒ bytes
-                case _: ResponseRenderingOutput.SwitchToWebSocket ⇒ throw new IllegalStateException("Didn't expect websocket response")
+                case ResponseRenderingOutput.HttpData(bytes)          ⇒ bytes
+                case _: ResponseRenderingOutput.SwitchToOtherProtocol ⇒ throw new IllegalStateException("Didn't expect protocol switch response")
```
ktoso: good

akka-http2-support/src/main/scala/akka/http/scaladsl/Http2.scala
```diff
@@ -94,16 +101,72 @@ final class Http2Ext(private val config: Config)(implicit val system: ActorSyste
...
+        // https://http2.github.io/http2-spec/#Http2SettingsHeader 3.2.1 HTTP2-Settings Header Field
+        val upgradeSettings = req.headers.collect {
+          case raw: RawHeader if raw.lowercaseName == Http2SettingsHeader.name ⇒
+            Http2SettingsHeader.parse(raw.value)
+        }
+
+        upgradeSettings match {
+          // Must be exactly one
+          case immutable.Seq(Success(settingsFromHeader)) ⇒
+            // TODO remove duplication?
```
ktoso: apply todo or remove; which duplication exactly?

ktoso: (it's not too crazy code... can be repeated I think)

akka-http2-support/src/main/scala/akka/http/scaladsl/model/http2/Http2SettingsHeader.scala
```diff
@@ -0,0 +1,35 @@
...
+import akka.stream.impl.io.ByteStringParser
+import akka.util.ByteString
+
+import scala.collection.immutable
+import scala.util.Try
+
+/**
+ * Internal API
+ */
+@InternalApi object Http2SettingsHeader {
```
ktoso: `private[akka]` I know it gets boring with all the markers... ;-) Eventually we'll get the gains from the annotation tho!

raboof:
PLS BUILD

docs/src/main/paradox/server-side/http2.md
```diff
@@ -42,13 +42,45 @@ Note that `bindAndHandle` currently does not support HTTP/2, you must use `bindA
 
 ### HTTP/2 without HTTPS
 
-While un-encrypted connections are allowed by HTTP/2, this is [generally discouraged](https://http2.github.io/faq/#does-http2-require-encryption).
+While un-encrypted connections, also known as 'h2c', are allowed by HTTP/2, this
+is [generally discouraged](https://http2.github.io/faq/#does-http2-require-encryption)
+and not typically supported by browsers.
```
jroper: I would remove the link to say it's generally discouraged.

By default, on Kubernetes, when you use HTTP/2 or gRPC, communication is unencrypted, with ingress TLS typically being terminated at the ingress proxy. When using Istio, the sidecar terminates TLS, and communicates unencrypted locally to the application. It is possible to make the sidecar communicate using TLS with the app, but that doesn't make sense, one of the major reasons for using Istio is so that the applications don't need to worry about certificates, Istio manages it all for you. So, in these environments, which is where many people will be using gRPC, unencrypted will be the norm. It's only on the web, with browsers, where there has been a push for many years to make all communications (not just HTTP/2) encrypted, that HTTP/2 requires TLS, but that's got nothing to do with HTTP/2, and everything to do with the web and the way it's evolving.

jrudolph: > So, in these environments, which is where many people will be using gRPC, unencrypted will be the norm.

That sounds reasonable in theory. Is it already happening or being prepared somewhere?

I could imagine that an HTTP/2 proxy sidecar will be its own can of worms. Is it specified which protocol it is going to use (h2c using the `Upgrade` header like in this PR or the plain form that just assumes HTTP/2 which is already implemented)?

jroper: I haven't looked in detail enough to answer those questions.... not sure.

raboof: I think in such situations h2c by preexisting knowledge rather than via the `Upgrade` is be more common. Makes sense to be more conservative in our assertions and remove the blanket "generally discouraged" claim though.

raboof:
Let's close this until we revisit this topic

raboof:
(added a simple test last night, probably good to have some more though. We can exercise the feature in Akka gRPC using the Netty-based client, though even there I don't think I'd recommend it for the typical cases)

Issue #1170, opened 2017-05-29
Javadsl RouteDirectives.route() with no args will throw java.lang.UnsupportedOperationException: empty.reduceLeft


jrudolph:
This should be illegal. We should probably 

 1. deprecate [that signature](https://github.com/akka/akka-http/blob/501663adb9a2925933377822896387268f3a25f1/akka-http/src/main/scala/akka/http/javadsl/server/directives/RouteDirectives.scala#L43-L43):

```scala
@varargs def route(alternatives: Route*): Route 
```

 2. and add another one

```scala
@varargs def route(first: Route, alternatives: Route*): Route 
```
 
 3. add a better runtime exception if someone uses the existing one in the wrong way

PR #2100, opened 2018-07-01
Add new method for routes chaining in Java DSL (#1170)

nikom1337:
I've deprecated both methods: RouteDirectives.route() and Directives.route() which was for fixing Scala's generated invalid  Java varargs method signatures.

```
// These are repeated here since sometimes (?) the Scala compiler won't actually generate java-compatible
// signatures for varargs methods, making them show up as Seq<Object> instead of T... in Java.`
```

Does anyone know if this problem still occurs? It was implemented more than 2 years ago.
https://github.com/akka/akka-http/commit/0fc1448ea86ac82630799c337981ed21b1f41081#diff-fdb294dd15eda1cb28e11e7aacfb9a24R23

Maybe we can just remove it :)

Adding new method required different method name to make compiler happy (Ambiguous method call).

Refs: #1170


ktoso:
OK TO TEST

akka-http/src/main/scala/akka/http/javadsl/server/Directives.scala
```diff
@@ -23,6 +23,7 @@ object Directives extends AllDirectives {
   // These are repeated here since sometimes (?) the Scala compiler won't actually generate java-compatible
   // signatures for varargs methods, making them show up as Seq<Object> instead of T... in Java.
 
+  @Deprecated
   @varargs override def route(alternatives: Route*): Route =
```
ktoso: javadoc is inherited from RouteDirectives right? okey then (to make sure the deprecation reason shows up

akka-http/src/main/scala/akka/http/javadsl/server/directives/RouteDirectives.scala
```diff
@@ -38,14 +37,29 @@ abstract class RouteDirectives extends RespondWithDirectives {
...
    * Java-specific call added so you can chain together multiple alternate routes using comma,
    * rather than having to explicitly call route1.orElse(route2).orElse(route3).
+   * @deprecated Use the `RouteDirectives.routes` method instead.
    */
+  @Deprecated
   @CorrespondsTo("concat")
   @varargs def route(alternatives: Route*): Route = RouteAdapter {
     import akka.http.scaladsl.server.Directives._
 
+    require(alternatives.nonEmpty, "Chaining empty list of routes is illegal.")
```
ktoso: please as `if (alternatives.isEmpty) throw ...` since require() allocates a closure and is used quite a bit nowadays so we don't want to add too much overhead there; even though we're deprecating it

akka-http/src/main/scala/akka/http/javadsl/server/directives/RouteDirectives.scala
```diff
@@ -38,14 +37,29 @@ abstract class RouteDirectives extends RespondWithDirectives {
...
+
     alternatives.map(_.delegate).reduce(_ ~ _)
   }
 
+  /**
+   * Java-specific call added so you can chain together multiple alternate routes using comma,
```
ktoso: Let's rephrase at the same time I guess; let's remove `Java-specific call added so you can` and say more like "Used tochain multiple alternate routes ..."

jlprat:
Regarding:
> Does anyone know if this problem still occurs? It was implemented more than 2 years ago.
0fc1448#diff-fdb294dd15eda1cb28e11e7aacfb9a24R23

It depends for which Scala this was built for. We still support Scala 2.11.x series, so unless this was only happening on 2.10.x, I guess is still needed.

akka-http/src/main/scala/akka/http/javadsl/server/Directives.scala
nikom1337: Yes, javadocs will be taken from a superclass.

nikom1337:
Regarding the `Directives.route()`, maybe it's better to keep it for now. It can be removed whenever RouteDirectives.route()` will be removed. 

@jlprat do you think i should provide the same workaround for a new method?

nikom1337:
We have three options (to keep Kotlin users happy and don't break binary compatibility):
- close this PR and postpone fixing issue until major release, then we can just replace old method signature with new one and break binary compatibility
- change name from `routes` to `chain`, `concat` or `join`, but it will break the naming convention between Scala and Java API :(
- leave Java API as it is 

sorry for delaying but it's very hot in Berlin O_________o

nikom1337:
I was wrong about the second point, `concat` can be used as a method name for routes chaining. It will bring Java and Scala API's closer together.

akka-http/src/main/scala/akka/http/javadsl/server/directives/RouteDirectives.scala
```diff
@@ -38,14 +37,30 @@ abstract class RouteDirectives extends RespondWithDirectives {
   /**
    * Java-specific call added so you can chain together multiple alternate routes using comma,
    * rather than having to explicitly call route1.orElse(route2).orElse(route3).
+   * @deprecated Use the `RouteDirectives.concat` method instead.
    */
+  @Deprecated
```
johanandren: We usually use the scala `@deprecated` which takes an explanation and a `since` version which makes it easy to see when the deprecation happened. Does that not end up in the Java code (I think someone claimed that elsewhere) or why are we using the Java annotation + javadoc-tag here instead?

The since "1.10.4" information would be useful regardless.

johanandren: Ah, I saw @jlprat s comment now, maybe ignore the part about the `since` then

akka-http/src/main/scala/akka/http/javadsl/server/directives/RouteDirectives.scala
```diff
@@ -38,14 +37,31 @@ abstract class RouteDirectives extends RespondWithDirectives {
...
+      throw new IllegalArgumentException("Chaining empty list of routes is illegal.")
+
     alternatives.map(_.delegate).reduce(_ ~ _)
   }
 
+  /**
+   * Used to chain multiple alternate routes using comma,
+   * rather than having to explicitly call route1.orElse(route2).orElse(route3).
+   */
+  @CorrespondsTo("concat")
```
jrudolph: Do we need the `@CorrespondsTo` when they have the same name?

ktoso: Nope, it's only when names are different

nikom1337:
@jrudolph thanks for reviewing :) I've already removed unnecessary @CorrespondsTo annotation.

PR #2101, opened 2018-07-02
Clarifying not all abrupt client go-aways are clean 

### LINKED ISSUE #1179: Server does not fail flow when client abruptly disconnects
The docs [says that the flow running a connection will see a failure if client abrupt disconnect](http://doc.akka.io/docs/akka-http/10.0.7/scala/http/low-level-server-side-api.html#connection-failures) however this sample which accepts a request but never responds does never see it (I'm guessing the stream completes rather than fails, but haven't verified):

```
  implicit val system = ActorSystem()
  implicit val materializer = ActorMaterializer()
  implicit val executionContext = system.dispatcher

  val (host, port) = ("localhost", 8080)
  val serverSource = Http().bind(host, port)
  
  val httpEcho = Flow[HttpRequest]
    .recover[HttpRequest] {
      case ex =>
        // handle the failure somehow
        println("Handling req failure somehow")
        throw ex
    }.mapAsync(1) { request =>
      // never complete future, causing client to timeout and close conn
      // or give us time to kill the client etc.
      Promise().future
    }

  serverSource
    .runForeach { con =>
      con.handleWith(httpEcho)
    }
```

We should fix docs or behaviour.

johanandren:
Fixes #1179

### DIFF
--- a/docs/src/main/paradox/server-side/low-level-api.md
+++ b/docs/src/main/paradox/server-side/low-level-api.md
@@ -218,6 +218,9 @@ Java
 
 The third type of failure that can occur is when the connection has been properly established,
 however afterwards is terminated abruptly – for example by the client aborting the underlying TCP connection.
+(Note that this is when the TCP connection is abruptly terminated in an orderly fasion, if the client just goes away,
+there will be no such failure and it will instead be detected through the @ref[idle timeout](../common/timeouts.md#timeouts))
+
 To handle this failure we can use the same pattern as in the previous snippet, however apply it to the connection's Flow:
 
 Scala


docs/src/main/paradox/server-side/low-level-api.md
```diff
@@ -218,6 +218,9 @@ Java
 
 The third type of failure that can occur is when the connection has been properly established,
 however afterwards is terminated abruptly – for example by the client aborting the underlying TCP connection.
+(Note that this is when the TCP connection is abruptly terminated in an orderly fasion, if the client just goes away,
```
ktoso: Full new sentence in () is a bit weird, just a normal one is fine

ktoso: `abruptly` sounds at odds to me to `orderly fasion` (note also typo, `fashion`), this needs more wording to explain properly I feel

docs/src/main/paradox/server-side/low-level-api.md
```diff
@@ -218,6 +218,9 @@ Java
...
+there will be no such failure and it will instead be detected through the @ref[idle timeout](../common/timeouts.md#timeouts))
```
ktoso: Missing . at end of this sentence;

docs/src/main/paradox/server-side/low-level-api.md
johanandren: The reason I didn't do a regular sentence is that then the following sentence becomes weird.

johanandren: As it refers back to the clean termination sentence.

docs/src/main/paradox/server-side/low-level-api.md
```diff
@@ -226,6 +227,12 @@ Scala
 Java
 :   @@snip [HttpServerExampleDocTest.java]($test$/java/docs/http/javadsl/server/HttpServerExampleDocTest.java) { #connection-stream-failure-handling }
 
+
+Note that this is when the TCP connection is closed correctly, if the client just goes away, for example because of
+a network failure, it will not be seen as this kind of stream failure. It will instead be detected through the
+@ref[idle timeout](../common/timeouts.md#timeouts)).
+
```
ktoso: better!

Issue #2103, opened 2018-07-03
perform null check when constructing a RawHeader instance


laszlovandenhoek:
Currently, this code:
```
import scala.collection.immutable
import akka.actor.ActorSystem
import akka.http.scaladsl._
import akka.http.scaladsl.model._
import akka.http.scaladsl.model.headers._

implicit val actorSystem = ActorSystem()

val badHeader = RawHeader("NullHeader",null)

Http().singleRequest(HttpRequest(uri ="http://example.com", headers = immutable.Seq(badHeader)))
```
...will throw a `NullPointerException` from an obscure place - [see for yourself in this Scastie](https://scastie.scala-lang.org/gYTZ0znQS0aKvbrVgGhSeA).

We should add proper null checks when constructing `RawHeader` instances, catching problems early.

jlprat:
Nulls are (or maybe I should say "should") not that common in Scala. However, this is not the case for Java where nulls are more present. I think it makes sense to have this protection

ktoso:
Fine to add IMHO; usually "should not happen"™ but when it does it is good to fail early

PR #2104, opened 2018-07-03
Perform null check when constructing RawHeader

laszlovandenhoek:
This resolves issue #2103.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
@@ -668,6 +668,8 @@ final case class Range(rangeUnit: RangeUnit, ranges: immutable.Seq[ByteRange]) e
 }
 
 final case class RawHeader(name: String, value: String) extends jm.headers.RawHeader {
+  if (name == null) throw new IllegalArgumentException("name must not be null")
+  if (value == null) throw new IllegalArgumentException("value must not be null")
   def renderInRequests = true
   def renderInResponses = true
   val lowercaseName = name.toRootLowerCase
--- a/akka-http-core/src/test/scala/akka/http/scaladsl/model/headers/HeaderSpec.scala
+++ b/akka-http-core/src/test/scala/akka/http/scaladsl/model/headers/HeaderSpec.scala
@@ -209,4 +209,15 @@ class HeaderSpec extends FreeSpec with Matchers {
       }
     }
   }
+  "RawHeader should" - {
+    "check for valid arguments" - {
+      "successful parse run" in {
+        RawHeader("foo", "bar").toString shouldEqual "foo: bar"
+      }
+      "failing parse run" in {
+        an[IllegalArgumentException] should be thrownBy RawHeader(null, "bar")
+        an[IllegalArgumentException] should be thrownBy RawHeader("foo", null)
+      }
+    }
+  }
 }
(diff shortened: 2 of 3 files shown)


akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
```diff
@@ -668,6 +668,8 @@ final case class Range(rangeUnit: RangeUnit, ranges: immutable.Seq[ByteRange]) e
 }
 
 final case class RawHeader(name: String, value: String) extends jm.headers.RawHeader {
+  require(name != null, "name must not be null")
```
jlprat: @ktoso I remember your comment about `require` allocating a closure in another PR. So I guess we should use here as well `if(name == null) throw ...`.

johanandren: True here as well, please if-throw instead.

ktoso: Ah good catch @jlprat! thanks! :)

jlprat: It got me by surprise on the other PR, so I tried to remember this one :)

johanandren:
One thought though, is there a reason RawHeader is specifically likely to get nulls passed, compared to any other header, or any other method in the Java APIs? If not, then shouldn't we either add _everywhere_ or nowhere to be consistent about it?

jlprat:
@johanandren I can't imagine why specifically `RawHeader` is more prone to it. One thing that came to my mind as well, is why `null` for the value shouldn't be considered as the empty string? In the end, it represents the absence of value. And I've seen headers that don't have any meaningful value, so they are left empty.

ktoso:
> One thing that came to my mind as well, is why null for the value shouldn't be considered as the empty string? In the end, it represents the absence of value. And I've seen headers that don't have any meaningful value, so they are left empty.

-1 for such treatment of nulls; null is wrong and we should throw then. I would suspect if a null ended up there it is a programming error and not an intended thing

ktoso:
Raw headers are perhaps a bit more likely to get nulls since people may want to expose "pass through" or "anything i don't understand" on as raw headers, perhaps without making sure the values are valid; I'm ok with the null checks here and not in all other places for that reason

jlprat:
>-1 for such treatment of nulls;

Agree, this was my old Java dev in me :laughing:

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
laszlovandenhoek: There's quite a few similar `require` calls in the file. Should those be changed too?

ktoso: Yeah, good idea

raboof:
@laszlovandenhoek are you still interested in updating the `require`s in this PR? if not I can have a look at it

laszlovandenhoek:
@raboof I updated the `require`s. I also added a test (as requested by Konrad) and rebased on `master` (because who likes merge commits).

Actually, I had the fix ready in code, but didn't push it because I had trouble running the tests locally (still do, actually, but that's probably me, so nvm), and subsequently forgot all about it.

Thanks for the nudge!

laszlovandenhoek:
Also, annoyingly, [it's its, not it's](http://www.its-not-its.info/).

laszlovandenhoek:
PLS BUILD

Issue #2097, opened 2018-06-29
Routing: PathMatcher.repeat(…) incorrectly matches an empty pattern


rrwright:
A route that contains the following path matcher:
`path("node" / PathMatchers.JavaUUID.repeat(1, Int.MaxValue, ",")) { uuids =>  … }`

correctly matches paths like:
`http://localhost:8080/node/5fcab910-b9b9-43de-b102-4eb3be66ff4c,00000000-0000-0000-0000-000000000002`

but incorrectly still matches paths which contain no UUIDs, like:
`http://localhost:8080/node/`

The minimum value passed in to the `PathMatcher.repeat(…)` method is not respected. 

I can see why here:
https://github.com/akka/akka-http/blob/master/akka-http/src/main/scala/akka/http/scaladsl/server/PathMatcher.scala#L85

but not being able to understand why there is a hardcoded value of `1` passed in to `rec` here kept me from submitting a patch: 
https://github.com/akka/akka-http/blob/master/akka-http/src/main/scala/akka/http/scaladsl/server/PathMatcher.scala#L83

raboof:
At first glance looks like an oversight/off-by-one - try it and make sure to check with tests?

Prithvirajbilla:
Hi, I'd like to pick this up. Thanks

PR #2105, opened 2018-07-04
Fixes bug in PathMatcher.repeat incorrectly matching empty pattern

Prithvirajbilla:
Fixes akka/akka-http#2097


ktoso:
OK TO TEST

akka-http/src/main/scala/akka/http/scaladsl/server/PathMatcher.scala
```diff
@@ -80,13 +80,13 @@ abstract class PathMatcher[L](implicit val ev: Tuple[L]) extends (Path ⇒ PathM
     new PathMatcher[lift.Out]()(lift.OutIsTuple) {
       require(min >= 0, "`min` must be >= 0")
       require(max >= min, "`max` must be >= `min`")
-      def apply(path: Path) = rec(path, 1)
+      def apply(path: Path) = rec(path, 0)
       def rec(path: Path, count: Int): Matching[lift.Out] = {
```
Prithvirajbilla: This piece of code is very hard to reason about. So to benefit the reviewer, I am explaining the change in words.

`rec(path: Path, count: Int)` is recursive path matcher, in which path corresponds to the provided/remaining path and count corresponds to number of times the expected pattern appears in the path.

so, `count` should start with 0, instead of 1. (which is the root cause for the bug)

akka-http/src/main/scala/akka/http/scaladsl/server/PathMatcher.scala
```diff
@@ -80,13 +80,13 @@ abstract class PathMatcher[L](implicit val ev: Tuple[L]) extends (Path ⇒ PathM
...
         def done = if (count >= min) Matched(path, lift()) else Unmatched
-        if (count <= max) {
+        if (count < max) {
```
Prithvirajbilla: check if count is less than max (instead of less than or equal to), because if **count is already max value**, no need to pathmatcher for expected pattern and run below logic.

akka-http/src/main/scala/akka/http/scaladsl/server/PathMatcher.scala
```diff
@@ -80,13 +80,13 @@ abstract class PathMatcher[L](implicit val ev: Tuple[L]) extends (Path ⇒ PathM
...
           self(path) match {
             case Matched(remaining, extractions) ⇒
-              def done1 = if (count >= min) Matched(remaining, lift(extractions)) else Unmatched
+              def done1 = if (count + 1 >= min) Matched(remaining, lift(extractions)) else Unmatched
```
Prithvirajbilla: here checking if **count+1 >= min instead of count >= min**, because we `self(path)` already matched a pattern and we need to increment count. (as mentioned earlier count denotes number of times the expected pattern appeared in the path till now. )

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/PathDirectivesSpec.scala
```diff
@@ -350,6 +350,26 @@ class PathDirectivesSpec extends RoutingSpec with Inside {
...
+      val test = testFor(pathPrefix(IntNumber.repeat(min = 0, max = 0, separator = ".")) { echoCaptureAndUnmatchedPath })
+      "accept [/qux]" inThe test("List():qux")
+      "accept [/1qux]" inThe test("List():1qux")
+    }
+    {
+      val test = testFor(pathPrefix(IntNumber.repeat(min = 0, max = 1, separator = ".")) { echoCaptureAndUnmatchedPath })
+      "accept [/xyz]" inThe test("List():xyz")
+      "accept [/1xyz]" inThe test("List(1):xyz")
+      "accept [/1.xyz]" inThe test("List(1):.xyz")
+      "accept [/1.2.xyz]" inThe test("List(1):.2.xyz")
```
johanandren: What about when it is not a prefix but the full path, would ´IntNumber.repeat(0, 1, ".")` reject for this one then? Could you check if there is a test covering that and add if not?

Prithvirajbilla: I don't there is a test which matches below test content. 
```
val test = testFor(path(IntNumber.repeat(min = 0, max = 1, separator = ".")) { echoCaptureAndUnmatchedPath }) //path instead of pathPrefix

"reject [/1.2.xyz]" inThe test()
..... so on
``` 
I will add it to this PR. 

Thanks.

Prithvirajbilla: Added these above tests. Thanks.

Prithvirajbilla:
Unrelated failure (#2108)

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/PathDirectivesSpec.scala
johanandren: Thanks, good to verify that it wasn't a case that broke with this fix.

johanandren:
PLS BUILD

Prithvirajbilla:
I don't see any merge button to merge this change. Also, cc: @ktoso, @jlprat - because they wanted to see if other members see the merge button.

jlprat:
I think it was a misunderstanding. It's all good, the people who should see the merge button can see them. Please notice that only committers of the project can do a merge.
If you are interested in know better how the whole PR review process goes you can read https://github.com/akka/akka-http/blob/master/CONTRIBUTING.md#general-workflow
Thanks for the PR @Prithvirajbilla

Prithvirajbilla:
oh! okay.

welcome!

akka-http/src/main/scala/akka/http/scaladsl/server/PathMatcher.scala
jrudolph: Maybe rename `count` to `alreadyFound`?

Thanks, good explanation. I guess one of the problems of the complexity here is matching the separator between the elements (as so often when you try to do that). Do you match the separator first (if it's not the first element) and then the element? Or do you match an element first and then try to see if there's a separator and only go on in that case.

I tried turning that logic around (only matching separator if it's not the first element), wdyt about that:

```scala
def apply(path: Path) = matchNext(path, 0)
def matchNext(path: Path, alreadyFound: Int): Matching[lift.Out] = {
  def done = if (alreadyFound >= min) Matched(path, lift()) else Unmatched

  def matchSeparatorIfNeeded(path: Path): Matching[Unit] =
    if (alreadyFound >= 1) separator(path) else Matched(path, ())
  def matchElement(start: Path): Matching[lift.Out] =
    self(start)
      .andThen { (remaining, extractions) ⇒
        matchNext(remaining, alreadyFound + 1)
          .map(result ⇒ lift(extractions, result))
      }
      .orElse(done)

  if (alreadyFound < max)
    matchSeparatorIfNeeded(path)
      .andThen { (remaining, _) ⇒ matchElement(remaining) }
      .orElse(done)
  else done
}
```

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/PathDirectivesSpec.scala
```diff
@@ -350,6 +350,39 @@ class PathDirectivesSpec extends RoutingSpec with Inside {
       "accept [/1.2.3/]" inThe test("List(1, 2, 3):/")
       "accept [/1.2.3./]" inThe test("List(1, 2, 3):./")
     }
+    {
```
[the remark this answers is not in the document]
ktoso: Yup, looks good 👍  Thanks @Prithvirajbilla

PR #2112, opened 2018-07-05
Allow using cleartext HTTP/2 from Java

### LINKED ISSUE #2110: No way to create a unencrypted HTTP/2 server with Java
The juggling going on with the `ConnectHttp` in the Java version of `bindAndHandleAsync` will loose the Http2 flag when the context is not encrypted (`effectiveConnectionContext` always returns the `fallbackContext` which doesn't have http2 set since `ConnectHttpImpl` always has an empty `connectionContext`)

johanandren:
Fixes  #2110

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/javadsl/Http.scala
+++ b/akka-http-core/src/main/scala/akka/http/javadsl/Http.scala
@@ -6,25 +6,28 @@ package akka.http.javadsl
 
 import java.net.InetSocketAddress
 import java.util.Optional
+
 import akka.http.impl.util.JavaMapping
 import akka.http.impl.util.JavaMapping.HttpsConnectionContext
 import akka.http.javadsl.model.ws._
-import akka.http.javadsl.settings.{ ConnectionPoolSettings, ClientConnectionSettings, ServerSettings }
-import akka.{ NotUsed, stream }
+import akka.http.javadsl.settings.{ClientConnectionSettings, ConnectionPoolSettings, ServerSettings}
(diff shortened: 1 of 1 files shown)


johanandren:
yea yea, formatting, I know...

ktoso:
Would propose renaming the commit though ;-)

raboof:
I have to admit I don't quite understand how this is intended to fit together. Should `ConnectHttp.effectiveHttpContext` be deprecated? Can/should we fix that one instead of writing an alternative inside `Http`?

johanandren:
Yeah,  I could definitely move the same logic inside ConnectHttp instead, transferring the HTTP/2 flag there. I haven't gotten my head around the `ConnectHttp` design, so I may very well be doing it in a backwards way.

johanandren:
What I find most confusing is that a `ConnectHttp` has a `UseHttp2` getter, which in the concrete subclasses are fields, while at the same time `ConnectHttp` has an optional `HttpsConnectionContext` which in turn _also_ has a `UseHttp2` field. The plot thickens!

The problem in this case is that if there is no https context - in other words what we have is a `ConnectHttpImpl` calling `effectiveHttpContext(fallback)` will just return the fallback, and the `ConnectHttpImpl` http2 flag is lost.

You have any idea what would be a good solution @raboof ?

johanandren:
I think maybe Chris is providing a cleaner solution for this in #2149

chbatey:
Sorry for the duplication, didn't look at existing issues while trying to make this work yesterday

Issue #2111, opened 2018-07-05
Support wildcard for Sec-WebSocket-Protocol


chbatey:
Currently netty (https://github.com/netty/netty/issues/1374) has this option and is the default for play. 

Related play issues: https://github.com/playframework/playframework/issues/7943 and https://github.com/playframework/playframework/issues/7895

There's a discussion that play should offer an API for it but having a `*` has the default seems sensible for people migrating from netty to akka http as a backend for play/lagom.

raboof:
The proposed change would be a workaround for the fact that some clients use the list of 'subprotocols' not to select a subprotocol, but to send metadata in the request. That would be fine, but apparently some client libraries check that the server selects one subprotocol from the list (it's not obvious from the spec that the server has to choose?), and this change makes the client happy by just arbitrarily returning one of the values.

I'd really prefer fixing this by having the 'downstream' implementation (i.e. Play) call `handleWebSocketMessagesForOptionalProtocol(handler, requestedProtocols.headOption)` instead of adding support for the current `handleWebSocketMessagesForProtocol(handler, "*")`. Or are there situations where that is not reasonable to expect?

PR #2113, opened 2018-07-06
Support all subprotocols via wildcard

chbatey:
Refs #2111

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/ws/Handshake.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/ws/Handshake.scala
@@ -25,6 +25,7 @@ import akka.util.OptionVal
 @InternalApi
 private[http] object Handshake {
   val CurrentWebSocketVersion = 13
+  val SubProtocolWildcard = "*"
 
   object Server {
     /**
@@ -109,9 +110,9 @@ private[http] object Handshake {
 
             def handle(handler: Either[Graph[FlowShape[FrameEvent, FrameEvent], Any], Graph[FlowShape[Message, Message], Any]], subprotocol: Option[String]): HttpResponse = {
               require(
-                subprotocol.forall(chosen ⇒ clientSupportedSubprotocols.contains(chosen)),
+                subprotocol.forall(chosen ⇒ chosen == SubProtocolWildcard || clientSupportedSubprotocols.contains(chosen)),
                 s"Tried to choose invalid subprotocol '$subprotocol' which wasn't offered by the client: [${requestedProtocols.mkString(", ")}]")
-              buildResponse(key.get, handler, subprotocol)
+              buildResponse(key.get, handler, subprotocol, clientSupportedSubprotocols)
             }
 
             def handleFrames(handlerFlow: Graph[FlowShape[FrameEvent, FrameEvent], Any], subprotocol: Option[String]): HttpResponse =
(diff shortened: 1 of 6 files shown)


akka-http/src/main/scala/akka/http/javadsl/server/directives/WebSocketDirectives.scala
```diff
@@ -66,6 +68,8 @@ abstract class WebSocketDirectives extends SecurityDirectives {
    * the request is rejected with an [[UnsupportedWebSocketSubprotocolRejection]] rejection.
    *
    * To support several subprotocols you may chain several `handleWebSocketMessagesForOptionalProtocol` routes.
+   *
+   * To support any subprotocol `*` can be used.
```
ktoso:
```
`"*"` can be used 
```

docs/src/test/scala/docs/http/scaladsl/server/directives/WebSocketDirectivesExamplesSpec.scala
```diff
@@ -137,6 +137,29 @@ class WebSocketDirectivesExamplesSpec extends RoutingSpec {
...
+    val wsClient = WSProbe()
+    WS("/all-sub-protocols", wsClient.flow, Seq("cat", "dog")) ~> route ~> check {
+      expectWebSocketUpgradeWithProtocol { protocol =>
+        protocol shouldEqual "cat"
+        wsClient.sendMessage("ping")
+        wsClient.expectMessage("ping")
+        wsClient.sendCompletion()
+        wsClient.expectCompletion()
+      }
+    }
```
ktoso: test obviously implemented by a cat lover 🐈  ;-)

akka-http-core/src/main/scala/akka/http/impl/engine/ws/Handshake.scala
```diff
@@ -144,15 +145,20 @@ private[http] object Handshake {
...
+        selectSubProtocol(requestedProtocols, subprotocol).toList :::
           List(
             UpgradeHeader,
             ConnectionUpgradeHeader,
             `Sec-WebSocket-Accept`.forKey(key),
             UpgradeToWebSocketResponseHeader(handler)))
+
+    // subprotocol has been validated to be either * or in the list
+    private[akka] def selectSubProtocol(requestedProtocols: immutable.Seq[String], subprotocol: Option[String]): Option[`Sec-WebSocket-Protocol`] = {
+      subprotocol.map(sp ⇒ `Sec-WebSocket-Protocol`(Seq(if (SubProtocolWildcard.eq(sp)) requestedProtocols.head else sp)))
```
raboof: Why not `sp == SubProtocolWildcard`? Weems obscure to rely on object equality here since it'd assume the `"*"` passed in by e.g. Play is interned (or refers to this constant)?

akka-http/src/main/scala/akka/http/javadsl/server/directives/WebSocketDirectives.scala
ktoso: It's a bit unfortunate that we have `Optional[String]` and not a select function here hm...

chbatey:
An alternative would be to offer a directive that accepts a function from `Seq[Protocol] -> Option[Protocol]` from the user which always matches but makes the user pick the protocol rather than us just selecting the first.

akka-http/src/main/scala/akka/http/javadsl/server/directives/WebSocketDirectives.scala
raboof: Actually I think passing in the selected subprotocol as a parameter rather than a function here makes sense, since you have to have to provide a `Flow` which implements the subprotocol you select, so subprotocol selection must be done before creating the `Flow` anyway.

chbatey:
Yeah @raboof that makes a lot more sense, was being stupid! Play uses the low level API but can do it there as well.

raboof:
Should be fixed upstream

PR #2117, opened 2018-07-16
make route concatentation lazy and save some allocations

swachter:
This pull request is a follow up on the proposal [Route concatenation operator should use value class and possibly be by name in its second argument](https://discuss.lightbend.com/t/route-concatenation-operator-should-use-value-class-and-possibly-be-by-name-in-its-second-argument/1540).
It makes two slight modifications to the `RouteWithConcatenation` class that provides the `~` extension method to routes:

1. The first modification makes `RouteWithConcatenation` a value class.
1. The second modification makes the argument to the `~` method a by-name argument.

Both modifications require binary compatibility exception rules.

### DIFF
--- a/akka-http/src/main/scala/akka/http/scaladsl/server/RouteConcatenation.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/server/RouteConcatenation.scala
@@ -34,12 +34,12 @@ trait RouteConcatenation {
 
 object RouteConcatenation extends RouteConcatenation {
 
-  class RouteWithConcatenation(route: Route) {
+  class RouteWithConcatenation(val route: Route) extends AnyVal {
     /**
      * Returns a Route that chains two Routes. If the first Route rejects the request the second route is given a
      * chance to act upon the request.
      */
-    def ~(other: Route): Route = { ctx ⇒
+    def ~(other: ⇒ Route): Route = { ctx ⇒
       import ctx.executionContext
       route(ctx).fast.flatMap {
         case x: RouteResult.Complete ⇒ FastFuture.successful(x)
--- a/akka-http/src/main/mima-filters/10.1.3.backwards.excludes
+++ b/akka-http/src/main/mima-filters/10.1.3.backwards.excludes
@@ -2,3 +2,12 @@
 
 # Removal of deprecated methods in 10.0.1 and 10.0.2
 ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.javadsl.server.Route.asScala")
+
+# Make RouteWithConcatentation a value class
+ProblemFilters.exclude[IncompatibleResultTypeProblem]("akka.http.scaladsl.server.RouteConcatenation._enhanceRouteWithConcatenation")
+ProblemFilters.exclude[IncompatibleResultTypeProblem]("akka.http.scaladsl.server.Directives._enhanceRouteWithConcatenation")
+ProblemFilters.exclude[FinalClassProblem]("akka.http.scaladsl.server.RouteConcatenation$RouteWithConcatenation")
+ProblemFilters.exclude[IncompatibleResultTypeProblem]("akka.http.scaladsl.server.HttpApp._enhanceRouteWithConcatenation")
+
+# Change argument of "~" method to be a by name argument
+ProblemFilters.exclude[IncompatibleMethTypeProblem]("akka.http.scaladsl.server.RouteConcatenation#RouteWithConcatenation.~")


swachter:
The binary compatibility issues may prevent the pull request from being merged. Another option could be to integrate these changes only for the coming Scala 2.13.

johanandren:
OK TO TEST

swachter:
Maybe we can have the discussion about lazyness right here?

I think that the `~` operator should be lazy because of the principle of least surprise.  Most of the akka-http routing logic is also lazy (cf. comment of @sirthias at [Spray issue 780](https://github.com/spray/spray/issues/780).

I guess that performance-wise it makes not much difference. Some applications would benefit because unmatched routes have not to be built whereas other applications would suffer from lazyness if route matches often happen at later alternatives. In these cases however, a reordering of alternatives seems to be in order anyway.

johanandren:
I'm on the fence, while I agree there is also the risk that users already depend on the eager behavior in weird ways, and changing it could break their code. 

That's why I proposed a separate route that explicitly is lazy, available for anyone to use, and not breaking anyones code.

Let's see what the rest of the team thinks. ping @akka/akka-http-team

jypma:
Somewhat unrelated, but perhaps we should pull into the discussion that forgetting to add `~` between routes is an often-made mistake; perhaps an alternative syntax could take care of both safety and lazy evaluation in one shot?

More on-topic, are there any established real-life cases where eagerly allocating these routes has shown to be the performance bottleneck?

ktoso:
> perhaps an alternative syntax could take care of both safety and lazy evaluation in one shot?

It exists. `concat(route, route, route)`

swachter:
@ktoso An alternative syntax could be `|`. The `|` symbol reflects that combined routes are alternatives (disregarding rejections that are collected over all alternatives.)

ktoso:
> @ktoso An alternative syntax could be |. The | symbol reflects that combined routes are alternatives (disregarding rejections that are collected over all alternatives.)

We will not add more glyphic operators to this DLS, if anything it could be a named combinator like `concatLazily` or similar.

akka-http/src/main/scala/akka/http/scaladsl/server/RouteConcatenation.scala
```diff
@@ -34,12 +34,12 @@ trait RouteConcatenation {
...
 object RouteConcatenation extends RouteConcatenation {
 
-  class RouteWithConcatenation(route: Route) {
+  class RouteWithConcatenation(val route: Route) extends AnyVal {
     /**
      * Returns a Route that chains two Routes. If the first Route rejects the request the second route is given a
      * chance to act upon the request.
      */
-    def ~(other: Route): Route = { ctx ⇒
+    def ~(other: ⇒ Route): Route = { ctx ⇒
```
ktoso: wait wait wait... :-)

This causes quite a lot of new allocations in each and every route combination. Before merging such change it would have to be made sure that it does not terribly degrade performance.

Also, not sure if the binary compatibility ignores applies hare can actually be made safely.

I would not want to merge this change at this stage of the project.
If anything, an alternate combinator could be offered.

akka-http/src/main/scala/akka/http/scaladsl/server/RouteConcatenation.scala
```diff
@@ -34,12 +34,12 @@ trait RouteConcatenation {
 
 object RouteConcatenation extends RouteConcatenation {
 
-  class RouteWithConcatenation(route: Route) {
+  class RouteWithConcatenation(val route: Route) extends AnyVal {
```
ktoso: the value class idea is nice; can we do a separate PR with just this?

ktoso: At the same time we could mark it also as `@DoNotInherit`

akka-http/src/main/scala/akka/http/scaladsl/server/RouteConcatenation.scala
swachter: @ktoso @johanandren I implemented a small performance test that compares 4 variants of concat operators (with/without value class, with/without by name argument). Preliminary results can be found here: https://github.com/swachter/akka-route-concat 

For sure, the measurements should be run with more iterations and forks. I will do this as soon as time permits. The results indicate that the concatenation operator that is implemented using a value class and uses a by-name argument for the second route has the highest throughput.

johanandren: Note that it is not only about throughput but also about memory pressure, which may not show up in the throughput of a micro bench but will make life miserable in general in a larger app. So make sure to look at those numbers from JMH a swell.

swachter: @johanandren Can you please tell me how to measure memory pressure. (I did not find a corresponding sbt-jmh option.)

johanandren: That'd be `-prof gc`. Also as you said, please run with warmup enough that you see the result triggering JIT and stabilize (5-10 at least I'd guess), and then more iterations.

swachter: @johanandren Thanks for the information. I ran the measurements with the following args: `sbt> jmh:run -i 20 -wi 20 -f 5 -prof gc` and update the results (cf. https://github.com/swachter/akka-route-concat). For convenience I repeat my summary here:

The concatenation combinator that is implemented by a value class and whose 2nd argument is by-name (i.e. route1) has the highest throughput and the smallest normalized allocation rate for all considered structures of routing trees. The difference in performance between the various concatenation operators gets more pronounced for deeper and broader routing trees.

If you agree with this result then I propose to add a new route concatenation operator (because of binary compatibility) that is implemented by a value class and has a by-name second argument.

Directives can be combined with the `|` combinator. In order to align routes with directives I would also offer the `|` symbol for that new combinator. An alias for the new combinator could be `or`.

jrudolph: I basically agree with performance and the laziness argument. One change between spray and akka-http was that we chose to embrace the laziness in the routing tree even more and made `Directive0.apply` lazy (because that's what most people tripped over). Following up on that making even more parts of the routing structure lazy seems just consistent.

I'm actually not particularly concerned about performance (though testing it definitely makes sense). Routing is a high-level feature and we know that it is not super fast. The directive structure is very flexible but it doesn't come for free. In fact, right now the most expensive part is not the routing but the marshalling and content type negotiation code, so that's where you would have to start optimizing to improve performance.

On the other hand, by now, we usually recommend `concat` instead of `~`. The problem is that we cannot do the same change for `concat` because there are no repeated by-name parameters (`=> Route*`). So, we would either have to introduce another inconsistency or would have to provide overloads of `concat` for different arities.

There's another argument against doing anything and that is about why adding this laziness would actually make a large difference. As I said before `Directive0.apply` is now lazy. That means the only place where adding laziness to `~` is where *creating* the directive passed to `~` is expensive or evaluated unexpectedly eagerly. @swachter, do you have any examples of where that would be the case?

jrudolph:
I'm closing this for now. The only way to get something like this in would be if it was binary compatible, with tests, and consistent over `~` and `concat` and hopefully also for the javadsl.

It's not clear if it's really worth the effort. For the time being, if you need lazyness a simple workaround is surround the directive application with `pass { /* now lazy inside */ }`.

Thanks for the suggestion and the discussion, @swachter.

Issue #340, opened 2016-09-20
Extract akka-http-model, make akka-http-core depend on it


ktoso:
Thanks to that Play could use it for their typed headers feature (and later on entire model perhaps).
The model should include the parsing things as well.

It's ok that model depends on akka-streams, we like the entities being modeled as streams.

Discussed Play roadmap with @wsargent and this would be a great win for both of the projects we think.

viktorklang:
:+1:

-- 
Cheers,
√

ktoso:
Hi @wsargent @gmethvin @schmitch,
so I looked into this and we talked it over in the team as well.
When looking at what akka-http would actually become we found that it would contain very little classes, mostly just the `Http` "entry point object" and some server internals (around 20 classes in total, all of it).

We were thinking with @patriknw that we're not sure the split is that important, rationale:
- it would make a lot of sense if splitting (core / model) would allow Play to avoid some additional dependencies... but we don't add any dependencies in akka-http-core – it is just the parsing / model / ~20 server classes.
- it would be tricky, but doable, with regards to splitting up where we get settings from - mostly it's about the parsers which get settings depending on if they're "server" or "client", so it's weird to talk about "server" or "client" in the model module itself, but in practice it makes sense so that's what we have now.

Do you think the split is really needed and would speed up adoption of the model by Play, or not so much (sounded useful, but not that important once you look at it)?
We need to decide this very soon, as we're trying to release a stable version as-soon-as-possible.

Have a look at what's inside akka-http-core and let me know,
if you strongly feel about the split and it really would contribute to quicker adoption we could do it, but if not we feel it's not enough bang-for-buck. Esp if we'll be considering the akka-http backend more seriously...

Open to suggestions, please comment!

schmitch:
Somehow I lost the issue...

Actually the first thing I actually wanted to use in play wasn't the model, I actually worked on the Uri Parsing for our netty backend and we looked into [scala-uri](https://github.com/NET-A-PORTER/scala-uri), however that pulls in `spray-json` and `parboiled` which pulls in `shapeless`, so that means we either need to shim it like akka did it or maybe reuse the akka stuff here.

I actually raised that to the typesafe/lightbend people and I guess they enlarged it since we also have and issue to use [typesafe headers](https://github.com/playframework/playframework/issues/5811) which we might can reuse from akka-http.

[...]

PR #2118, opened 2018-07-16
[WIP] Initial attempt to extract model module

jlprat:
Really rough Work In Progress. However, feedback is really welcomed!

MiMa will complain, and the tests are not compiling yet.

While pulling the `model` out, I needed to move some more (too much?) classes over as well.
I also have a problem I don't understand with Implicits and `JavaMapping`. They are somehow not recognised and need to be defined as `val`s locally.


akka-http-core/src/main/scala/akka/http/impl/util/JavaMappingCore.scala
```diff
@@ -0,0 +1,35 @@
...
+
+package akka.http.impl.util
+
+import akka.annotation.InternalApi
+import akka.http.javadsl.{ ConnectionContext, HttpConnectionContext, HttpsConnectionContext, settings ⇒ js }
+import akka.http.{ javadsl ⇒ jdsl, scaladsl ⇒ sdsl }
+
+/** INTERNAL API */
+@InternalApi
+private[http] object JavaMappingCore extends JavaMappings {
```
jlprat: To avoid moving all those classes to the `akka-http-model` module, I split the `JavaMapping` into 2 different files.

akka-http-core/src/main/scala/akka/http/javadsl/ClientTransport.scala
```diff
@@ -33,6 +33,7 @@ abstract class ClientTransport {
  */
 @ApiMayChange
 object ClientTransport {
+  implicit val ct = akka.http.impl.util.JavaMappingCore.ClientTransport
```
jlprat: For some reason, if I import `JavaMappingCore._` the implicit objects are not picked, I need to manually put them into scope

akka-http-model/src/main/scala/akka/http/impl/util/JavaMapping.scala
```diff
@@ -66,7 +65,7 @@ private[http] trait JavaMapping[_J, _S] extends J2SMapping[_J] with S2JMapping[_
 
 /** INTERNAL API */
 @InternalApi
-private[http] object JavaMapping {
+private[http] trait JavaMappings {
```
jlprat: In an attempt to re-use the implicits defined here, I renamed it and converted to a trait, so the counter part in `akka-http-core` can extend it as well.

jlprat:
Refs #340 BTW

ktoso:
Heads up / fair warning: that such move may end up being too disruptive and may not get in.
Please give some time for initial review before spending a lot of time on it :)

jlprat:
Sure, I spend some time doing this move back and forth. I think it's still quite binary compatible, but the whole `JavaMapping` doesn't work as it should. Also the cut is not clean, and I understand this might be one of the reasons why it might not get in. I wanted to give it a try and see how coupled `core` and `model` were. The answer seems to be, quite coupled.
I will wait for some reviews to spend more time on this one :)

build.sbt
```diff
@@ -99,13 +99,22 @@ lazy val parsing = project("akka-parsing")
...
   .settings(AutomaticModuleName.settings("akka.http.core"))
-  .dependsOn(parsing)
+  .dependsOn(httpModel)
   .addAkkaModuleDependency("akka-stream", "provided")
   .addAkkaModuleDependency("akka-stream-testkit", "test")
   .settings(Dependencies.httpCore)
   .settings(VersionGenerator.versionSettings)
   .enablePlugins(BootstrapGenjavadoc)
 
+lazy val httpModel = project("akka-http-model")
```
naderghanbari: - Maybe this could come before `httpCore`:
  - a tiny bit more sequential in terms of dependency, `httpModel` depends on `parsing` and `httpCore` depends on `httpModel`.

- Should the `root` project aggregate `httpModel` as well?

jlprat:
Shall we close this one?

akka-http-core/src/main/scala/akka/http/javadsl/ClientTransport.scala
jrudolph: It's probably because of name clashes. Maybe renaming the instances would be enough for imports to work.

jrudolph:
> Shall we close this one?

Yep, maybe close it for now. It's already a great first step. I've also lying a branch around somewhere where I tried to disentangle things a bit more to achieve cross-compilation with scalaJS. It was lots of work and in the end the interesting parts didn't work with scalaJS anyway (it seems parboiled generates too complicated code for the header parsers).

Issue #2122, opened 2018-07-20
HTTP Strict Transport Security (HSTS) header support is not conform spec


sindlar:
HSTS support in akka-http is not conform spec:
- Trailing semicolons are not supported; see [here](https://github.com/spotify/web-api/issues/655) for an occurrence of this bug in the wild.
- A fixed order of directives is assumed.

See below for a summary of the HSTS header definition.

# HSTS header definition
The [HSTS spec, Section 6](https://tools.ietf.org/html/rfc6797#section-6) defines the HSTS header value through the following ABNF syntax:
```
     Strict-Transport-Security = "Strict-Transport-Security" ":"
                                 [ directive ]  *( ";" [ directive ] )
     directive                 = directive-name [ "=" directive-value ]
     directive-name            = token
     directive-value           = token | quoted-string
```

From the [ABNF spec, Section 2](https://tools.ietf.org/html/rfc2616#section-2), the following parts are relevant:
```
   *rule
      The character "*" preceding an element indicates repetition. The
      full form is "<n>*<m>element" indicating at least <n> and at most
      <m> occurrences of element. Default values are 0 and infinity so
      that "*(element)" allows any number, including zero; "1*element"
      requires at least one; and "1*2element" allows one or two.
   [rule]
      Square brackets enclose optional elements; "[foo bar]" is
      equivalent to "*1(foo bar)".
```

Thus, the HSTS header value syntax can be rewritten to
```
     Strict-Transport-Security = "Strict-Transport-Security" ":"
                                 0*1( directive )  0*∞( ";" 0*1( directive ) )
```

In words, this means the value of the HSTS header has
- an optional directive,
- followed by any number (0 to infinity) of the following successive elements
  - semicolon
  - an optional directive

In addition to this syntax, the [HSTS spec, Section 6.1](https://tools.ietf.org/html/rfc6797#section-6.1) imposes some requirements, summarized below insofar as relevant to this issue:
- The order of appearance of directives is not significant.
- All directives MUST appear only once in an STS header field. Directives are either optional or required, as stipulated in their definitions.
- Unrecognized directives must be ignored.
- The 'max-age' directive is REQUIRED.
- The 'includeSubDomains' directive is OPTIONAL.

PR #2123, opened 2018-07-20
Implement HSTS header support conform spec #2122

sindlar:
Fixes issue [#2122](https://github.com/akka/akka-http/issues/2122) so that HSTS header support is conform spec. Specifically:
- Trailing semicolons are allowed.
- Order of directives is not important.
- 'max-age' directive must occur exactly once.
- 'includeSubDomains' directive may not occur more than once.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/model/parser/SimpleHeaders.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/model/parser/SimpleHeaders.scala
@@ -202,7 +202,7 @@ private[parser] trait SimpleHeaders { this: Parser with CommonRules with CommonA
           | ignoreCase("max-age=") ~ `delta-seconds` ~> (MaxAge(_))
       )
     def ignoredDirective = rule {
-      token ~ optional(ws("=") ~ word) ~> ((k: String, v: Option[String]) => IgnoredDirective(k + v.getOrElse("")))
+      token ~ optional(ws("=") ~ word) ~> ((k: String, v: Option[String]) ⇒ IgnoredDirective(k + v.getOrElse("")))
     }
 
     rule {
--- a/akka-http-core/src/main/scala/akka/http/impl/model/parser/SimpleHeaders.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/model/parser/SimpleHeaders.scala
@@ -194,8 +194,18 @@ private[parser] trait SimpleHeaders { this: Parser with CommonRules with CommonA
   // http://tools.ietf.org/html/rfc7231#section-7.4.2
   def server = rule { products ~ EOI ~> (Server(_)) }
 
-  def `strict-transport-security` = rule {
-    ignoreCase("max-age=") ~ `delta-seconds` ~ optional(ws(";") ~ ignoreCase("includesubdomains") ~ push(true)) ~ zeroOrMore(ws(";") ~ token0) ~ EOI ~> (`Strict-Transport-Security`(_, _))
+  // https://tools.ietf.org/html/rfc6797#section-6
+  def `strict-transport-security` = {
+    def directives =
+      rule(
+        ignoreCase("includesubdomains") ~ push(IncludeSubDomains)
+          | ignoreCase("max-age=") ~ `delta-seconds` ~> (MaxAge(_))
+          | word ~> (IgnoredDirective(_))
+      )
+
+    rule {
+      oneOrMore(directives).separatedBy(oneOrMore(ws(";"))) ~ zeroOrMore(ws(";")) ~ EOI ~> (`Strict-Transport-Security`(_: _*))
+    }
   }
 
   // http://tools.ietf.org/html/rfc7230#section-3.3.1
--- a/akka-http-core/src/main/scala/akka/http/impl/model/parser/SimpleHeaders.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/model/parser/SimpleHeaders.scala
@@ -206,7 +206,8 @@ private[parser] trait SimpleHeaders { this: Parser with CommonRules with CommonA
     }
 
     rule {
-      oneOrMore(directives | ignoredDirective).separatedBy(oneOrMore(ws(";"))) ~ zeroOrMore(ws(";")) ~ EOI ~> (`Strict-Transport-Security`(_: _*))
+      oneOrMore(directives | ignoredDirective).separatedBy(oneOrMore(ws(";"))) ~ zeroOrMore(ws(";")) ~ EOI ~>
+        (`Strict-Transport-Security`.fromDirectives(_: _*))
     }
   }
 
--- a/akka-http-core/src/main/scala/akka/http/impl/model/parser/SimpleHeaders.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/model/parser/SimpleHeaders.scala
@@ -200,11 +200,13 @@ private[parser] trait SimpleHeaders { this: Parser with CommonRules with CommonA
       rule(
         ignoreCase("includesubdomains") ~ push(IncludeSubDomains)
           | ignoreCase("max-age=") ~ `delta-seconds` ~> (MaxAge(_))
-          | word ~> (IgnoredDirective(_))
       )
+    def ignoredDirective = rule {
+      token ~ optional(ws("=") ~ word) ~> ((k: String, v: Option[String]) => IgnoredDirective(k + v.getOrElse("")))
+    }
 
     rule {
-      oneOrMore(directives).separatedBy(oneOrMore(ws(";"))) ~ zeroOrMore(ws(";")) ~ EOI ~> (`Strict-Transport-Security`(_: _*))
+      oneOrMore(directives | ignoredDirective).separatedBy(oneOrMore(ws(";"))) ~ zeroOrMore(ws(";")) ~ EOI ~> (`Strict-Transport-Security`(_: _*))
     }
   }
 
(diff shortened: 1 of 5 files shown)


ktoso:
OK TO TEST

akka-http-core/src/main/scala/akka/http/impl/model/parser/SimpleHeaders.scala
```diff
@@ -194,8 +194,21 @@ private[parser] trait SimpleHeaders { this: Parser with CommonRules with CommonA
...
+          | ignoreCase("max-age=") ~ `delta-seconds` ~> (MaxAge(_))
+      )
+    def ignoredDirective = rule {
+      token ~ optional(ws("=") ~ word) ~> ((k: String, v: Option[String]) ⇒ IgnoredDirective(k + v.getOrElse("")))
+    }
+
+    rule {
+      oneOrMore(directives | ignoredDirective).separatedBy(oneOrMore(ws(";"))) ~ zeroOrMore(ws(";")) ~ EOI ~>
+        (`Strict-Transport-Security`.fromDirectives(_: _*))
+    }
```
johanandren: In the spec it says `[ directive ]  *( ";" [ directive ] )` unless I'm really tired this morning and am missing something, that doesn't mean that a trailing semicolon is allowed. It means a directive followed by zero or more semicolon prefixed directives.

Also, you didn't mention adding accepting ignoring unknown directives, is this really a good idea? I see only a note about non-compliant UAs ignoring future/unknown directives in the spec.

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/StrictTransportSecurityDirective.scala
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright (C) 2009-2018 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.http.scaladsl.model.headers
+
+sealed abstract class StrictTransportSecurityDirective
+case class IgnoredDirective(value: String) extends StrictTransportSecurityDirective
```
johanandren: Make the case classes final to protect from user extension (makes bincomp evolution easier)

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/StrictTransportSecurityDirective.scala
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright (C) 2009-2018 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.http.scaladsl.model.headers
+
+sealed abstract class StrictTransportSecurityDirective
```
johanandren: mark as `/** Not for user extension */ @DoNotInherit`

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
```diff
@@ -864,6 +864,16 @@ final case class Server(products: immutable.Seq[ProductVersion]) extends jm.head
 // https://tools.ietf.org/html/rfc6797
 object `Strict-Transport-Security` extends ModeledCompanion[`Strict-Transport-Security`] {
   def apply(maxAge: Long, includeSubDomains: Option[Boolean]) = new `Strict-Transport-Security`(maxAge, includeSubDomains.getOrElse(false))
+
+  def fromDirectives(directives: StrictTransportSecurityDirective*) = {
+    val maxAgeDirectives = directives.filter(_.isInstanceOf[MaxAge])
+    require(maxAgeDirectives.size == 1, "exactly one 'max-age' directive required")
```
johanandren: Require allocates a lambda on every call, please use an `if (exp) throw IllegalArgumentEx(...)` instead to save some GC for this an the other validation both.

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
```diff
@@ -864,6 +864,16 @@ final case class Server(products: immutable.Seq[ProductVersion]) extends jm.head
 // https://tools.ietf.org/html/rfc6797
 object `Strict-Transport-Security` extends ModeledCompanion[`Strict-Transport-Security`] {
   def apply(maxAge: Long, includeSubDomains: Option[Boolean]) = new `Strict-Transport-Security`(maxAge, includeSubDomains.getOrElse(false))
+
+  def fromDirectives(directives: StrictTransportSecurityDirective*) = {
+    val maxAgeDirectives = directives.filter(_.isInstanceOf[MaxAge])
```
johanandren: `directives.collect { case m: MaxAge => m }`, and lift the partial function out to the header object to make sure only a single instance of it is used/needed

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
```diff
@@ -864,6 +864,16 @@ final case class Server(products: immutable.Seq[ProductVersion]) extends jm.head
...
+    require(maxAgeDirectives.size == 1, "exactly one 'max-age' directive required")
+
+    val includeSubDomainsDirectives = directives.filter(_.equals(IncludeSubDomains))
```
johanandren: Since we don't use anything but the count of instances `.count(_ eq IncludeSubdomains)` would avoid creating a collection instance, make sure to use `eq` since instance equality is enough, and same here about caching the lambda.

johanandren:
We need Java API also btw.

akka-http-core/src/main/scala/akka/http/impl/model/parser/SimpleHeaders.scala
sindlar: Hi, thanks for your comments!

Going by the ABNF spec, `[ directive ] *( ";" [ directive ] )` can be rewritten to `0*1( directive )  0*∞( ";" 0*1( directive ) )`. Especially note that the directives are *optional*, and therefore in my understanding the following strings are all syntactically valid:
* directive1; directive2
* directive1;; directive2;
* ; directive2
* ;;;
However, they are not conform spec, because the spec requires at least the `max-age` directive to be present. Nevertheless, I think a trailing semicolon is syntactically allowed!

I'm not sure I understand your point about ignored directives. [Section 6.1 of the spec](https://tools.ietf.org/html/rfc6797#section-6.1) mentions:

> 5.  If an STS header field contains directive(s) not recognized by
>     the UA, the UA MUST ignore the unrecognized directives, and if
>     the STS header field otherwise satisfies the above requirements
>     (1 through 4), the UA MUST process the recognized directives.

In my implementation, ignored directives are parsed into instances of `IgnoredDirective`, and filtered out in `Strict-Transport-Security#fromDirectives`. Does this clarify it for you?

johanandren: Yes, you are right, the square brackets is what my monday morning eyes didn't catch, thanks for clarifying.

johanandren: And point 5. Good, then this is the expected behavior per spec!

sindlar:
Hi @johanandren ,

You remarked "We need Java API also btw." Does that still need addressing? I wasn't sure, because the Scala model (`final case class `Strict-Transport-Security`(maxAge: Long, includeSubDomains: Boolean = false) extends jm.headers.StrictTransportSecurity`) hasn't changed.

akka-http-core/src/main/scala/akka/http/impl/model/parser/SimpleHeaders.scala
raboof: (really weird spec though... but yeah)

akka-http-core/src/test/scala/akka/http/scaladsl/model/headers/HeaderSpec.scala
```diff
@@ -84,18 +84,34 @@ class HeaderSpec extends FreeSpec with Matchers {
...
         headers.`Strict-Transport-Security`.parseFromValueString("max-age=30; includeSubDomains") shouldEqual Right(headers.`Strict-Transport-Security`(30, true))
-        headers.`Strict-Transport-Security`.parseFromValueString("max-age=30; includeSubDomains; preload") shouldEqual Right(headers.`Strict-Transport-Security`(30, true))
+        headers.`Strict-Transport-Security`.parseFromValueString("includeSubDomains; max-age=30") shouldEqual Right(headers.`Strict-Transport-Security`(30, true))
       }
-      "successful parse run with additional values" in {
+      "successful parse run with ignored directives" in {
         headers.`Strict-Transport-Security`.parseFromValueString("max-age=30; includeSubDomains; preload; dummy") shouldEqual
           Right(headers.`Strict-Transport-Security`(30, true))
-        headers.`Strict-Transport-Security`.parseFromValueString("max-age=30; includeSubDomains; dummy; preload") shouldEqual
+        headers.`Strict-Transport-Security`.parseFromValueString("max-age=30; includeSubDomains; foo=bar; preload") shouldEqual
```
raboof: Why not test `foo=bar`?

akka-http-core/src/main/scala/akka/http/impl/model/parser/SimpleHeaders.scala
raboof: I'm not entirely sure about those `IgnoredDirective`s: indeed the spec says the UA must ignore unrecognized directives, but even if Akka HTTP doesn't recognize them, the application could, so I'm not sure it's the responsibility of Akka HTTP to filter those out. OTOH we don't keep arbitrary fields in e.g. `HttpCookie` so this is consistent & fine.

Issue #2127, opened 2018-07-26
MarshallingTestUtils should get the timeout from config


MelaHub:
MarshallingTestUtils uses hardcoded 1 and 2 seconds for timeout, in case of heavy tests this causes the tests to fail with Request was neither completed nor rejected within 1 second.

MatshallingTestUtils should get the timeout from conf - e.g. akka.http.testkit.marshalling.timeout

johanandren:
Sounds good! Would you be up for a PR adding that @MelaHub ?

MelaHub:
Sure thing @johanandren ! I will need some help though :)

johanandren:
Let us know what you need help with and we'll help as best as we can! (It's vacation times so the team is a bit thin though, replies may not be super-fast the coming weeks)

MelaHub:
Just to start...how can I assign the issue to myself?  I'm going to start to work on this in the next few days - can't wait to contribute!

jlprat:
Unfortunately, issues can only be assigned to members of the organization. But I'll label it accordingly.

jlprat:
So feel free to start working on it 🙂

MelaHub:
Thank you @jlprat ! Now I have something in place but I have no permission to push the branch and open the pr :(

jlprat:
Close with https://github.com/akka/akka-http/commit/72d4f9615adaba5eb62896550dbdb4a5f898d865

PR #2130, opened 2018-07-27
Refactored out the 1 seconds marshalling timeout from MarshallingTestUtils so it can be configured (#2127)

MelaHub:
@ktoso this is the PR on the timeout (we were chatting about it in gitter) - I think this should work but you were mentioning about requiring an implicit actor system so I'm not sure if I'm missing something here

### DIFF
--- a/akka-http-testkit/src/main/resources/reference.conf
+++ b/akka-http-testkit/src/main/resources/reference.conf
@@ -1,0 +1,1 @@
+akka.http.testkit.marshalling.timeout = 1 s
--- a/akka-http-tests/src/test/scala/akka/http/scaladsl/marshalling/MarshallingSpec.scala
+++ b/akka-http-tests/src/test/scala/akka/http/scaladsl/marshalling/MarshallingSpec.scala
@@ -4,26 +4,30 @@
 
 package akka.http.scaladsl.marshalling
 
-import scala.collection.immutable
-import scala.collection.immutable.ListMap
-import org.scalatest.{ BeforeAndAfterAll, FreeSpec, Matchers }
-import akka.util.ByteString
 import akka.actor.ActorSystem
+import akka.http.impl.util._
+import akka.http.scaladsl.marshallers.xml.ScalaXmlSupport._
+import akka.http.scaladsl.model.HttpCharsets._
+import akka.http.scaladsl.model.MediaTypes._
+import akka.http.scaladsl.model._
+import akka.http.scaladsl.model.headers._
+import akka.http.scaladsl.testkit.MarshallingTestUtils
 import akka.stream.ActorMaterializer
 import akka.stream.scaladsl.Source
-import akka.http.scaladsl.testkit.MarshallingTestUtils
-import akka.http.scaladsl.marshallers.xml.ScalaXmlSupport._
-import akka.http.impl.util._
-import akka.http.scaladsl.model._
-import headers._
-import HttpCharsets._
-import MediaTypes._
 import akka.testkit.TestKit
+import akka.util.ByteString
+import com.typesafe.config.ConfigFactory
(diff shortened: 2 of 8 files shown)


jlprat:
OK TO TEST

akka-http-testkit/src/main/resources/application.conf
```diff
@@ -0,0 +1 @@
+akka.http.testkit.marshalling.timeout = 1
```
johanandren: This one should be `reference.conf` rather than `application.conf` as it is a default value and then users will use their `application.conf` to override it if they wish. 

The hoconf file format allows for actual time with units, so that isn't implicit, so if you specify this as "1 s" instead. Here's an example how to read a duration field: https://github.com/akka/akka-http/blob/1d1bfa3f796f3d008676c14e10bcd8daff42dd81/akka-http-tests/src/multi-jvm/scala/akka/http/AkkaHttpServerLatencyMultiNodeSpec.scala#L154

akka-http-tests/src/test/scala/akka/http/scaladsl/marshalling/MarshallingSpec.scala
```diff
@@ -25,6 +26,8 @@ class MarshallingSpec extends FreeSpec with Matchers with BeforeAndAfterAll with
   implicit val materializer = ActorMaterializer()
   import system.dispatcher
 
+  val marshallingTimeout = 1.second
```
johanandren: Shouldn't these come from the config now/marshalling test utils now?

MelaHub: Yes indeed - as these were tests I just set a default timeout but indeed they should come from conf - I've moved around things a bit so they're fetched from conf directly

lightbend-cla-validator:
Hi @MelaHub,

johanandren:
I verify that the CLA is signed and this is just the CLA-validator not being quite awake.

akka-http-testkit/src/main/resources/reference.conf
```diff
@@ -0,0 +1 @@
+akka.http.testkit.marshalling.timeout = 1 s
```
johanandren: We usually prefix those with a little bit of docs comment giving an explanation, in this case maybe it is evident what it does from the name, so not sure it is needed. Add if you agree or leave it be if not :)

johanandren:
Fixes #2127

jlprat:
I just realized you committed using two different users `MelaHub` and `Mela`. I guess your local git config has some different data than your github account.
@MelaHub do you want to squash your commits to have a proper user or shall we merge anyway and see to which of the both users the commits are attributed?

MelaHub:
Mmm weird, I only have one github account - not sure how it managed to get the Mela one. I'm going to squash, hold on.

MelaHub:
Squashed! I still don't get where it's getting the Mela from but I guess it's fine anyways.

akka-http-testkit/src/main/resources/reference.conf
MelaHub: Ouch I totally missed your comment! I think it's pretty clear though, I cannot think of any documentation that wouldn't just repeat the name of the param

jlprat:
I think I know where it might come from.
If you run:`git config --list` while in the `akka-http` repo, you'll see:
```
...
user.email=xxxx
...
```
Is that e-mail the same e-mail you use for your github account?

MelaHub:
yeah, it's the same as my github account - I also have a bitbucket account called Mela that somehow was interfering even if I was committing with the github one. Anyways, now it looks like the committing user is the correct one.

jlprat:
Thanks a lot @MelaHub and congrats on your first Akka HTTP PR!

Issue #1979, opened 2018-04-06
Improved handle APIs for websockets that automatically accumulate messages etc


ktoso:
This is a good idea by @jrudolph to improve handling websockets:

https://github.com/akka/akka-http/pull/611#issuecomment-275113296

----------

I think the change does what it advertises. However, I'd like to experiment with alternative APIs, instead, before people start to use `toStrict` (which requires to deal with futures) when an alternative could solve the issue in a simpler way. 

What I'd like to try first is adding a method to `UpgradeToWebSocket`:

```scala
def handleStrictMessages(Graph[FlowShape[StrictMessage, Message], Any], ...)
def handleStrictTextMessages(Graph[FlowShape[TextMessage.Strict, Message], Any], ...)
def handleStrictBinaryMessages(Graph[FlowShape[BinaryMessage.Strict, Message], Any], ...)
```

where `StrictMessage` is either a new hierarchy of WS messages or a new subtype. All of these should automatically "do the right thing", i.e. accumulating streamed messages with a configured timeout and maximum size and ignore messages of the wrong shape. We could probably reuse your `toStrict` logic for it. Is that something you would like to help with, @rbudzko?

ktoso:
Made it a new issue since I think the previous PR is already improving status quo and should be merged (adding toStrict), since it has then timeouts and is "the same as" HttpEntity methods.

johanandren:
Sounds like a good idea 👍

ulasakdeniz:
I would like to give this a try.

PR #2133, opened 2018-07-31
Add support for strict websocket message handling #1979

ulasakdeniz:
As #1979 suggests, this PR
- adds `handleStrictMessages`, `handleStrictTextMessages` and `handleStrictBinaryMessages` methods to `UpgradeToWebSocket`
- defines `StrictMessage` and `StreamedMessage` as new subtypes of `Message`
- also adds directives to `WebSocketDirectives`.

This helps users to
- operate with `StrictMessage` by using akka-streams API.
- avoid [scenarios that “work fine” in tests but fail in production](https://github.com/akka/akka-http/issues/65#issuecomment-245641551).

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/ws/UpgradeToWebSocket.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/ws/UpgradeToWebSocket.scala
@@ -7,13 +7,11 @@ package akka.http.scaladsl.model.ws
 import java.lang.Iterable
 import scala.collection.immutable
 import akka.NotUsed
+import akka.actor.ActorSystem
 import akka.stream._
 import akka.http.impl.util.JavaMapping
 import akka.http.javadsl.{ model ⇒ jm }
 import akka.http.scaladsl.model.HttpResponse
-
-import scala.concurrent.Future
-import scala.concurrent.duration.FiniteDuration
 
 /**
  * A custom header that will be added to an WebSocket upgrade HttpRequest that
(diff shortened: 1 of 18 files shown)


jlprat:
OK TO TEST

akka-http-core/src/main/scala/akka/http/scaladsl/model/ws/UpgradeToWebSocket.scala
```diff
@@ -54,6 +57,31 @@ trait UpgradeToWebSocket extends jm.ws.UpgradeToWebSocket {
     subprotocol: Option[String]                   = None): HttpResponse =
     handleMessages(scaladsl.Flow.fromSinkAndSource(inSink, outSource), subprotocol)
 
+  def handleStrictMessages(
```
johanandren: I don't think this name quite matches what it actually does, just reading the name I'd expect non-strict messages are filtered, but in fact it turns non-strict messages into strict ones. It would be good if we could convey the `toStrict` change in the name here as well somehow.

No good name ideas from the top of my head though.

One idea that could be worth exploring, to also avoid having a combinatory explosion, would be adding some form of builder instead, the API could look something like:
```
handleWsMessages().forProto(subProto).toStrict().handleWith(flow)
handleWsMessages().forProto(subProto).only[Binary].handleWith(flow)
handleWsMessages().only[Text].handleWith(sink, source)
handleWsMessages().forProto(subProto).only[Binary].handleWith(actorRef) // for #2158
```

akka-http-core/src/main/scala/akka/http/scaladsl/model/ws/UpgradeToWebSocket.scala
```diff
@@ -54,6 +57,31 @@ trait UpgradeToWebSocket extends jm.ws.UpgradeToWebSocket {
...
+    handlerFlow: Graph[FlowShape[StrictMessage, Message], Any],
+    subprotocol: Option[String]                                = None)(implicit mat: Materializer, timeout: FiniteDuration): HttpResponse =
+    handleMessages(scaladsl.Flow[Message].mapAsync(1) {
+      case streamed: TextMessage.Streamed   ⇒ streamed.toStrict(timeout)
+      case streamed: BinaryMessage.Streamed ⇒ streamed.toStrict(timeout)
```
johanandren: I think we also need a max-bytes limit in addition to the timeout here, with a fast client I could potentially OOM the server by just streaming a huge amount of streamed data. On the entity side this is protected by the `akka.http.parsing.max-content-length` so maybe we need either to use that or a corresponding new setting. Thoughts @raboof @jrudolph ?

akka-http-core/src/main/scala/akka/http/scaladsl/model/ws/UpgradeToWebSocket.scala
johanandren: Ups, strike the `toStrict` there, and strictify with default values, but maybe have a `maxStrictSize(bytes)`/`toStrictTimeout(4.seconds)` to allow customization per route.

ulasakdeniz:
Thanks for the review @johanandren, I'll try to apply your comments when I have time.

jlprat:
Hi @ulasakdeniz 
Do you want to come back to this PR? It's almost there!

Thanks!

ulasakdeniz:
Hi @jlprat 
Sure, I'll start with exploring the builder approach.

ulasakdeniz:
Hello,
I tried builder approach together with maxStrictSize/toStrictTimeout suggested by @johanandren and added 2 new settings for these values. Didn’t add scaladoc to everywhere yet.

Thanks

docs/src/test/scala/docs/http/scaladsl/server/directives/WebSocketDirectivesExamplesSpec.scala
```diff
@@ -107,6 +107,119 @@ class WebSocketDirectivesExamplesSpec extends RoutingSpec {
...
+  "handle-only-binary-messages-builder" in {
+    //#handle-only-binary-messages-builder
+    def greeter: Flow[StrictMessage, Message, Any] =
+      Flow[StrictMessage].map {
+        case BinaryMessage.Strict(data) => BinaryMessage(ByteString("Binary message received: ") ++ data)
+      }
+    val timeout: FiniteDuration = 1.second
+    val websocketRoute =
+      path("greeter") {
+        handleWsMessages { builder =>
```
johanandren: I don't think a `Directive1` makes sense for the builder, because it is not composable with other directives which this implies, better with just a method that returns the builder, and the builder then returns a `Route` which is the end-station, so to speak, of the routing tree.

ennru:
Hi @ulasakdeniz is this something you'd like to warm up to discuss it further? That would be great, if not we might as well close this PR.

raboof:
> Hi @ulasakdeniz is this something you'd like to warm up to discuss it further? That would be great, if not we might as well close this PR.

let's close it due to inactivity for now - though if someone wants to pick this back up that'd still be interesting!

ulasakdeniz:
Hi @ennru, sorry I missed the notification for your message. And yes @raboof it's good that you closed it, it's been a long time and I've lost track of this PR.

Issue #2145, opened 2018-08-14
Default Http().bindAndHandleAsync default parallelism=1 doesn't make sense for HTTP/2


jrudolph:
`parallelism` defaults to `1` because in HTTP/1 we assume that pipelining is switched off so handling one request at a time *per connection* is enough.

For HTTP/2, however, it doesn't really make sense as we want to allow multiplexing multiple requests per connection. It's somewhat weird that this never came up while developing and testing HTTP/2. I guess the reason is that during HTTP/2 testing we never tested running multiple `HttpRequest => Future[HttpResponse]` at the same time but only tested streaming out the response in parallel which works.

We should

 * document that you need to adapt this setting with HTTP/2
 * change samples to use a more sensible value
 * add a warning if the HTTP/2 server is run with the default of 1

ktoso:
(Please mark in progress since has PR)

jrudolph:
Fixed in #2146.

PR #2146, opened 2018-08-14
Introduce HTTP/2 max-concurrent-streams settings

jrudolph:
Refs #2145 

The idea is to change the default for the `bindAndHandleAsync` `parallelism` parameter to `0`, which means "take the value from the configuration".

We then take different values for HTTP/1 and HTTP/2:

 * for HTTP/1 take `pipelining-limit` because that's even stronger than parallelism
 * for HTTP/2 take a new setting `http2.max-concurrent-streams`

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/javadsl/settings/Http2ServerSettings.scala
+++ b/akka-http-core/src/main/scala/akka/http/javadsl/settings/Http2ServerSettings.scala
@@ -4,9 +4,11 @@
 
 package akka.http.javadsl.settings
 
+import akka.annotation.DoNotInherit
 import akka.http.scaladsl
 import com.typesafe.config.Config
 
+@DoNotInherit
 trait Http2ServerSettings { self: scaladsl.settings.Http2ServerSettings ⇒
   def getRequestEntityChunkSize: Int = requestEntityChunkSize
   def withRequestEntityChunkSize(newRequestEntityChunkSize: Int): Http2ServerSettings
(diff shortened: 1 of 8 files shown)


akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
```diff
@@ -283,28 +283,36 @@ class HttpExt private[http] (private val config: Config)(implicit val system: Ex
...
+      Http2Shadow.bindAndHandleAsync(handler, interface, port, connectionContext, definitiveSettings, parallelism, log)(fm)
     } else {
-      if (http2enabled)
+      if (http2Enabled)
         log.debug("The akka.http.server.preview.enable-http2 flag was set, " +
           "but a plain HttpConnectionContext (not Https) was given, binding using plain HTTP...")
 
-      bindAndHandle(Flow[HttpRequest].mapAsync(parallelism)(handler), interface, port, connectionContext, settings, log)
+      val definitiveParallelism =
+        if (parallelism != 0) parallelism
```
ktoso: `> 0`?

akka-http-core/src/main/scala/akka/http/javadsl/settings/Http2ServerSettings.scala
```diff
@@ -16,6 +16,9 @@ trait Http2ServerSettings { self: scaladsl.settings.Http2ServerSettings ⇒
 
   def getIncomingStreamLevelBufferSize: Int = incomingStreamLevelBufferSize
   def withIncomingStreamLevelBufferSize(newIncomingStreamLevelBufferSize: Int): Http2ServerSettings
+
+  def getMaxConcurrentStreams: Int = maxConcurrentStreams
+  def withMaxConcurrentStreams(newValue: Int): Http2ServerSettings
```
johanandren: One thought, I don't remember the exact HTTP/2 details, but, isn't there a way this should be sent to the client to inform it about how many sub streams it may use to the server?

ktoso: Hm indeed it can be set as SETTING frame:

```
5.1.2 Stream Concurrency

A peer can limit the number of concurrently active streams using the SETTINGS_MAX_CONCURRENT_STREAMS parameter (see Section 6.5.2) within a SETTINGS frame. The maximum concurrent streams setting is specific to each endpoint and applies only to the peer that receives the setting. That is, clients specify the maximum number of concurrent streams the server can initiate, and servers specify the maximum number of concurrent streams the client can initiate.

```

I guess this is more like a default value for us for now hm... The setting we don't respect yet I guess

jrudolph: That's correct, in future we could announce and enforce this value with HTTP/2.

raboof:
There is a follow-up issue to this one, https://github.com/akka/akka-http/issues/2168

PR #2148, opened 2018-08-15
Ignore case for MediaType.hashCode

### LINKED ISSUE #2144: MediaType discrepency between hashCode and equals: cannot use Set
Media types are case insensitive. Currently, `MediaType`'s [equals](https://github.com/akka/akka-http/blob/f3e83935ffc27bc94f586d9387b54b499c6250d4/akka-http-core/src/main/scala/akka/http/scaladsl/model/MediaType.scala#L53) method correctly ignores case, but its [hashCode](https://github.com/akka/akka-http/blob/f3e83935ffc27bc94f586d9387b54b499c6250d4/akka-http-core/src/main/scala/akka/http/scaladsl/model/MediaType.scala#L59) is calculated on the raw value. This presents a problem for when parsed incoming media types are looked up in a `HashSet` or `HashMap`: since the hash codes differ for differently-cased `MediaType` instances for the same media type, the set will look in the wrong bucket and not ever invoke `equals()`. 

I think a simple fix could be to change `hashCode()` to be `value.toLowerCase().hashCode()` and perhaps cache that.

jypma:
If MediaType instances are used with a hashset or hashmap,
equivalent (equal) objects should go into the same buckets. Since
MediaType's equals() method correctly is case insensitive, do should its
hashCode.

Fixes #2144.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/MediaType.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/MediaType.scala
@@ -56,7 +56,7 @@ sealed abstract class MediaType extends jm.MediaType with LazyValueBytesRenderab
       case _            ⇒ false
     }
 
-  override def hashCode(): Int = value.hashCode
+  override def hashCode(): Int = value.toLowerCase.hashCode
 
   /**
    * JAVA API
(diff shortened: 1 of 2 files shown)


raboof:
PLS BUILD

ktoso:
Thanks!

akka-http-core/src/main/scala/akka/http/scaladsl/model/MediaType.scala
```diff
@@ -56,7 +56,7 @@ sealed abstract class MediaType extends jm.MediaType with LazyValueBytesRenderab
       case _            ⇒ false
     }
 
-  override def hashCode(): Int = value.hashCode
+  override def hashCode(): Int = value.toLowerCase.hashCode
```
hepin1989: val

ktoso: Nah, def is fine. No reason to increase the weight of the object for this

PR #2149, opened 2018-08-15
Preserve http2 option when going from javadsl to scaladsl

chbatey:
The only information passed from javadsl to the scaladsl in Http
was the context which for Http connections was always the default one
meaning that setting http2 always had no effect.

This seed the easiest way not to break the user facing API

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/javadsl/ConnectHttp.scala
+++ b/akka-http-core/src/main/scala/akka/http/javadsl/ConnectHttp.scala
@@ -9,6 +9,7 @@ import java.util.Optional
 
 import akka.annotation.{ DoNotInherit, InternalApi }
 import akka.http.javadsl.model.Uri
+import akka.http.scaladsl.HttpConnectionContext
 import akka.http.scaladsl.UseHttp2.Negotiated
 
 @DoNotInherit
@@ -23,7 +24,7 @@ abstract class ConnectHttp {
   final def effectiveHttpsConnectionContext(fallbackContext: HttpsConnectionContext): HttpsConnectionContext =
     connectionContext.orElse(fallbackContext)
 
-  final def effectiveConnectionContext(fallbackContext: ConnectionContext): ConnectionContext =
+  def effectiveConnectionContext(fallbackContext: ConnectionContext): ConnectionContext =
     if (connectionContext.isPresent) connectionContext.get()
     else fallbackContext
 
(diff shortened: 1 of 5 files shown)


johanandren:
I think this refs #2110 ?

akka-http-core/src/main/scala/akka/http/javadsl/ConnectHttp.scala
```diff
@@ -167,6 +168,9 @@ final class ConnectHttpImpl(val host: String, val port: Int, val http2: UseHttp2
   def isHttps: Boolean = false
 
   def connectionContext: Optional[HttpsConnectionContext] = Optional.empty()
+
+  override def effectiveConnectionContext(fallbackContext: ConnectionContext): ConnectionContext =
+    HttpConnectionContext(http2.asScala)
```
raboof: So this always ignores the `host`, `port`, `connectionContext` and the `fallbackContext` parameter? That seems odd..

raboof: ah `connectionContext` is the HTTPS context so that might make sense to ignore here.

chbatey: :spades: 

This did take me a couple of hours :confused: 

I find the API very confusing bewtween the ConnectHttp and Context.

the `connectionContext` is only Https so a user can't ever have called `withContext` and set a custom one here.

I think that is a restriction in the API that we could remove but it wouldn't be binary/source compatible so this was a quick way to make it work without breaking any existing users.

chbatey: The host and port are ignored here but when the delegation to the scaladsl from the javadsl they are taken from the ConnectHttp and passed in

raboof: Yeah seems to make sense, just trying to digest the change ;)

chbatey:
I prefer @jrudolph's solution. Do you want to take another look @raboof ?

akka-http-core/src/main/scala/akka/http/javadsl/ConnectHttp.scala
```diff
@@ -21,11 +21,15 @@ abstract class ConnectHttp {
...
-    connectionContext.orElse(fallbackContext)
+    connectionContext
+      .orElse(fallbackContext)
+      .withHttp2(http2)
 
   final def effectiveConnectionContext(fallbackContext: ConnectionContext): ConnectionContext =
-    if (connectionContext.isPresent) connectionContext.get()
-    else fallbackContext
+    connectionContext
+      .map[ConnectionContext](c ⇒ c: ConnectionContext) // covariance, anyone?
```
chbatey: Indeed :(

jrudolph:
PLS BUILD

jrudolph:
... no more editing in Github I'll check it locally ...

jrudolph:
Compilation failed with a Scala 2.11 type inference problem... Hopefully fixed now :)

jrudolph:
Hey, scalac 2.11, you are a pain (though, I know having to deal with Java existentials is not your fault). Great, however, that these bugs have been fixed in 2.12!

Issue #2151, opened 2018-08-16
t:akka-core connection issue


Edoin:
I am having issue with `akka-http` connecting to `wss://s2.ripple.com:443` in using

`val webSocketFlow = Http().webSocketClientFlow(WebSocketRequest("wss://s2.ripple.com:443"))`

I am having this debug logs with errors.

```
[DEBUG] [08/16/2018 18:23:50.360] [default-akka.actor.default-dispatcher-2] [akka://default/system/IO-TCP/selectors/$a] Executing [WorkerForCommand(Connect(s2.ripple.com:443,None,List(),Some(10 seconds),true),Actor[akka://default/system/StreamSupervisor-0/$$a#-1464307790],<function1>)]
[DEBUG] [08/16/2018 18:23:50.380] [default-akka.actor.default-dispatcher-2] [akka://default/system/IO-TCP/selectors/$a/0] Resolving s2.ripple.com before connecting
[DEBUG] [08/16/2018 18:23:50.408] [default-akka.actor.default-dispatcher-4] [akka://default/system/IO-DNS] Resolution request for s2.ripple.com from Actor[akka://default/system/IO-TCP/selectors/$a/0#-1362893195]
[DEBUG] [08/16/2018 18:23:50.425] [default-akka.actor.default-dispatcher-4] [akka://default/system/IO-TCP/selectors/$a/0] Attempting connection to [s2.ripple.com/169.55.164.20:443]
[DEBUG] [08/16/2018 18:23:50.824] [default-akka.actor.default-dispatcher-4] [akka://default/system/IO-TCP/selectors/$a/0] Connection established to [s2.ripple.com:443]
[DEBUG] [08/16/2018 18:23:50.838] [default-akka.actor.default-dispatcher-11] [akka://default/system/IO-TCP/selectors/$a/0] [Actor[akka://default/system/StreamSupervisor-0/$$a#-1464307790]] registered as connection handler
[DEBUG] [08/16/2018 18:23:50.845] [default-akka.actor.default-dispatcher-11] [akka://default/system/IO-TCP/selectors/$a/0] Wrote [236] bytes to channel
[DEBUG] [08/16/2018 18:23:51.439] [default-akka.actor.default-dispatcher-2] [akka://default/system/IO-TCP/selectors/$a/0] Read [4656] bytes.
[...]

PR #2152, opened 2018-08-16
Accept websocket headers in parts

raboof:
Refs #2151.

I would've liked to add a test that checks whether this still works when the
second message is sent partially, but I couldn't figure out who to convince
WebSocketClientSpec to exchange multiple messages even in the simple case...

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/ws/WebSocketClientBlueprint.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/ws/WebSocketClientBlueprint.scala
@@ -24,7 +24,7 @@ import akka.http.impl.engine.parsing.ParserOutput.{ MessageStartError, NeedMoreD
 import akka.http.impl.engine.parsing.{ HttpHeaderParser, HttpResponseParser, ParserOutput }
 import akka.http.impl.engine.rendering.{ HttpRequestRendererFactory, RequestRenderingContext }
 import akka.http.impl.engine.ws.Handshake.Client.NegotiatedWebSocketSettings
-import akka.http.impl.util.StreamUtils
+import akka.http.impl.util.{ SingletonException, StreamUtils }
 import akka.stream.impl.fusing.GraphStages.SimpleLinearGraphStage
 
 /** INTERNAL API */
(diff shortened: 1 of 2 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/ws/WebSocketClientBlueprint.scala
```diff
@@ -65,13 +65,23 @@ private[http] object WebSocketClientBlueprint {
...
+          val parser: HttpResponseParser = new HttpResponseParser(settings.parserSettings, HttpHeaderParser(settings.parserSettings, log)) {
             var first = true
             override def handleInformationalResponses = false
             override protected def parseMessage(input: ByteString, offset: Int): StateResult = {
               if (first) {
-                first = false
-                super.parseMessage(input, offset)
+                try {
+                  // If we're called recursively then that's a next message
+                  first = false
```
jrudolph: Would it also work to just switch the two statements?

```scala
val result = super.parseMessage(input, offset)
first = false
result
```

akka-http-core/src/test/scala/akka/http/impl/engine/ws/WebSocketClientSpec.scala
```diff
@@ -43,6 +44,23 @@ class WebSocketClientSpec extends FreeSpec with Matchers with WithMaterializerSp
       sendWSFrame(Protocol.Opcode.Text, ByteString("Message 1"), fin = true)
       expectMaskedFrameOnNetwork(Protocol.Opcode.Text, ByteString("Message 1"), fin = true)
     }
+
+    "establish connection with header values sent in parts" in new TestSetup with ClientEchoes {
```
jrudolph: :+1:

akka-http-core/src/main/scala/akka/http/impl/engine/ws/WebSocketClientBlueprint.scala
raboof: No, apparently `parseMessage` recursively calls itself (and then you want 'first' to be false for the recursive call) - I haven't looked into how/why in detail but that's what happened :smile:

This is a risk: what if `NotEnoughDataException` is thrown while processing the next element in that recursive call? Since the recursive call will just does `emit(RemainingBytes(input.drop(offset))); terminate()` I don't think that can happen though.

Issue #2155, opened 2018-08-17
Use travis stages to call whitesource only once per build


raboof:
I suspect the failures we're having are related to the fact that we're invoking whitesource at roughly the same time for both the 2.11 and the 2.12 version.

If we update our travis config to check whitesource only once that might mitigate the problem. The chance that this would miss a licensing problem that is present only on 2.11 or only on 2.12 seems rather low.

jrudolph:
> I suspect the failures we're having are related to the fact that we're invoking whitesource at roughly the same time for both the 2.11 and the 2.12 version.

Ugh, good point. Will be interesting to see if we can improve the situation :+1:

jrudolph:
I guess that was fixed by now?

PR #2157, opened 2018-08-20
Split off whitesource check to separate stage

raboof:
Refs #2155


jrudolph:
screenshot from 2018-08-21 10-35-05

"Build config didn't create any jobs"

raboof:
Ugh, and that is actually correct :blush: will fix

jrudolph:
So, (as before), we only use travis in PR validation for compilation and mima checks?

ktoso:
> So, (as before), we only use travis in PR validation for compilation and mima checks?

Yeah AFAIK

PR #2159, opened 2018-08-21
Add MiMa configuration

raboof:
Seems we have missed some in the past. Should we go through them and see if
there are any we want to fix retroactively?


jrudolph:
I think they are a blast from the past where releasing a new 10.0.x release would again mark all deprecation removals in 10.1.x as incompatible changes.

Therefore, I think the right solution would be to reorganize mima checks in a way that this cannot happen.

Then, we need to make sure no actual incompatibilities have been introduced in the meantime.

Finally, we seem to need to automate more of the release process to rule out these kind of mistakes ;)

RELEASING.md
```diff
@@ -3,6 +3,7 @@
 1. Communicate that a release is about to be released in [Gitter Akka Dev Channel](https://gitter.im/akka/dev), so that no new Pull Requests are merged
 1. Add a release notes entry in docs/src/main/paradox/release-notes/. As a helper run
 `scripts/commits-for-release-notes.sh <last-version-tag>` which will output a list of commits grouped by submodule, and the closed issues for this milestone.
+1. Add the released version to `project/MiMa.scala` to the `mimaPreviousArtifacts` key.
```
jrudolph: The problem with putting that step here is, that you cannot commit it yet, because mima cannot find the newly created release yet. On the other hand, I guess that putting it in line 4 might increase the probability that the releaser actually reads the line :)

raboof: Ah of course...

raboof:
Added https://github.com/akka/akka-http/issues/2160 to consider getting the versions from git tags

jrudolph:
> Therefore, I think the right solution would be to reorganize mima checks in a way that this cannot happen.

I.e. all excludes from `10.0.11` should be valid for all 10.0.x versions because from the view of mima 10.0.12 reintroduced all those incompatible methods again.

I guess one way to implement it, would be to allow a `10.0.x.backwards.excludes` file whose excludes would then always be added to the latest registered 10.0.x version. Maybe we can prototype that in akka-http and then add it to MIMA proper?

jrudolph:
> Then, we need to make sure no actual incompatibilities have been introduced in the meantime.

It seems this is the only one if I only add the new 10.1.x versions:

```
[info] akka-http: found 1 potential binary incompatibilities while checking against com.typesafe.akka:akka-http_2.12:10.1.0 
[error]  * abstract method seal(akka.http.javadsl.server.RejectionHandler,akka.http.javadsl.server.ExceptionHandler)akka.http.javadsl.server.Route in interface akka.http.javadsl.server.Route is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.javadsl.server.Route.seal")
```

project/MiMa.scala
```diff
@@ -34,7 +34,13 @@ object MiMa extends AutoPlugin {
           "10.0.8",
           "10.0.9",
           "10.0.10",
-          "10.0.11"
+          "10.0.11",
+          "10.0.12",
+          "10.0.13",
+          "10.1.1",
```
jrudolph: 10.1.0 is missing

jrudolph:
> I guess one way to implement it, would be to allow a 10.0.x.backwards.excludes file whose excludes would then always be added to the latest registered 10.0.x version. Maybe we can prototype that in akka-http and then add it to MIMA proper?

An alternative would be to define a "fork". In our case the fork would be 10.0.x, in that case we collect all exclusions for 10.0.x and apply them to all 10.0.x versions where the default behavior of mima is only to collect all "older" (by version ordering) exclusions.

https://github.com/lightbend/migration-manager/blob/e54f3914741b7f528da5507e515cc84db60abdd5/core/src/main/scala/com/typesafe/tools/mima/core/ProblemReporting.scala#L22

jrudolph:
> An alternative would be to define a "fork". In our case the fork would be 10.0.x, in that case we collect all exclusions for 10.0.x and apply them to all 10.0.x versions where the default behavior of mima is only to collect all "older" (by version ordering) exclusions.

E.g. like this:

```scala
    mimaBackwardIssueFilters ~= { filters =>
      val oldVersions = filters.keys.filterNot(_.startsWith("10.1"))
      val collectedFilters = oldVersions.map(filters).reduce(_ ++ _)

      filters.filterKeys(_ startsWith "10.1") + ("10.0.13" -> collectedFilters)
    }
```

raboof:
Let's do only #2161 for now ;)

PR #2161, opened 2018-08-21
Register versions to perform MiMa checks against

raboof:


jrudolph:
I think we can just add an exclusion for the single failure we get for this change:
https://github.com/akka/akka-http/commit/5826729506bec56a271ec25bcb2c79e983f0b409

Seems to be between 10.1.0 and 10.1.1.

jrudolph:
> I think we can just add an exclusion for the single failure we get for this change:
> 5826729

Strictly speaking it's an incompatibility, because the `Route` interface might be extended by someone but actually, you would usually extend `RouteAdapter`, instead. Maybe we can also mark `Route` as `@DoNotInherit` and add comments how to write Routes instead.

(No blocker for 10.1.4 imo, no one has noticed the incompatibility so far and we cannot fix it anymore anyway).

akka-http/src/main/mima-filters/10.1.0.backwards.excludes
```diff
@@ -0,0 +1,2 @@
+# https://github.com/akka/akka-http/commit/5826729506bec56a271ec25bcb2c79e983f0b409
```
jrudolph: :+1: Maybe add a comment that `Route` is marked as `@DoNotInherit`

akka-http/src/main/mima-filters/10.1.1.backwards.excludes
```diff
@@ -0,0 +1,4 @@
+# Only show up on scala 2.11:
```
jrudolph: I guess these ones are ok because 1. PredefinedToResponseMarshallers/LowPriorityToResponseMarshallerImplicits are de-facto `@DoNotInherit` and 2. Java users should probably use the Scala 2.12 version anyway?

PR #2164, opened 2018-08-21
=pro Fix Mima rules for forked release branches

jrudolph:
In these cases like looking from 10.1.x to 10.0.x, when branching, we
defined a set of mima filters between 10.1 and 10.0. However, from Mima's
perspective a new 10.0.x version will reintroduce the incompatibilities.

For that reason, we collect all old filters (i.e. filters for released versions of
forked branches) and only keep an entry for the latest version of the forked
branch.

In our case, all filters for 10.0.x are collected and then registered as a
single entry for 10.0.13 (mima already reuses newer filters for older versions
so that this covers all older versions).

An alternative approach would be to collect mima filters for forked branches
when doing the branch in a single file like `10.0.x.backwards.excludes`.


jrudolph:
@raboof Would be good to merge that soon, so we can 10.0.14 to the list of versions to check against.

PR #2174, opened 2018-08-23
=htp provide exceptions as failed futures in mapRouteResultFuture, fixes #2173

### LINKED ISSUE #2173: BasicDirective.mapRouteResultFuture should have exceptions converted into failed Futures
`mapRouteResultFuture` should be the go-to directive when you want to handle any kind of result of the inner route, `Completed`, `Rejected` or failed with an exception. However, we currently catch exceptions and convert them to failed futures only at some few selected places (maybe to avoid sprinkling exception handling all over?).

E.g.

```scala
val logResults = BasicDirectives.mapRouteResultFuture { res =>
  println(s"Result was $res")
  res
}
```

use like this:

```scala
logResults {
  get { throw new IllegalStateException("errrrrrorrr") }
}
```

does not currently log the exception because the `mapRouteResultFuture` handler is never called.

jrudolph:


johanandren:
Fixes #2173

Issue #2137, opened 2018-08-06
decodeRequest does not respect withSizeLimit


tewe:
When using [decodeRequest](https://doc.akka.io/docs/akka-http/current/routing-dsl/directives/coding-directives/decodeRequest.html) to handle `Content-Encoding: gzip` any [withSizeLimit](https://doc.akka.io/docs/akka-http/current/routing-dsl/directives/misc-directives/withSizeLimit.html) applies to the compressed size, not the uncompressed size. This might facilitate DoS attacks.

The reason might be that `decodeRequestWith` uses `transformDataBytes` which creates a new `HttpEntity` that is not limitable.

TheEmacsShibe:
I tested this and it can indeed crash a server. The following shell commands can create a large enough gzip file and send it to the server:

```
dd if=/dev/zero bs=1M count=6000 | pv | gzip -9 > foo.gz
curl -v -i http://example.com/post-endpoint -H'Content-Encoding: gzip' --data-binary @foo.gz
``

jrudolph:
Thanks for the report. We'll look into it.

jrudolph:
The idea used to be that by basing everything on streaming, these kind of issues are less likely. However, there are two main parts where streams are loaded into memory:

 * `toStrict` (in all variants)
 * `PredefinedFromEntityUnmarshallers.byteStringUnmarshaller` which blindly collects everything it gets and which is the base unmarshaller for basically all other strict unmarshallers like those used for the `formData` directive.

`toStrict` docs at least contain a warning that this might not be a good idea but for the unmarshallers it is unexpected.

I think we might want limits at different stages:

 * `withSizeLimit` / `max-content-length` as input limits
 * `decodeRequest` and everything that potentially inflates data
 * `toStrict` should have a limit
 * `PredefinedFromEntityUnmarshallers.byteStringUnmarshaller` and other marshallers that might collect data should have limits

I propose the following changes:
 * add config settings for each of those cases with a reasonable default limit (which isn't unlimited) like 10MB per request
 * change `decodeRequest` to add another stage which honors the limit
   * maybe add a variant of `decodeRequest` that allows to change limit as a shortcut
 * same for `toStrict` directives
 * limiting the unmarshallers is the hardest because they are static and currently allow no possibility for configuration. With a bit of fiddling we could change the internals to pass configuration when they are invoked. Adding limits to `decodeRequest` and `toStrict` would restrict the problem for now, so maybe changing the unmarshallers could be postponed.
[...]

PR #2185, opened 2018-08-31
Limit the size of a decoded stream

raboof:
refs #2137


akka-http/src/main/resources/reference.conf
```diff
@@ -37,6 +37,13 @@ akka.http.routing {
   # for an entity data stream.
   decode-max-bytes-per-chunk = 1m
 
+  # Maximum content length after applying a decoding directive. When the directive
+  # decompresses, for example, an entity compressed with gzip, the resulting stream can be much
+  # bigger than the max-content-length. Like with max-content-length, this is not necessarilly a
+  # problem when consuming the entity in a streaming fashion, but does risk high memory use
+  # when the entity is made strict or marshalled into an in-memory object.
+  decode-max-size = 8m
```
johanandren: Should it default to the value of max-content-size to try to not break applications for users that increased that?

akka-http/src/main/scala/akka/http/scaladsl/server/directives/CodingDirectives.scala
```diff
@@ -82,14 +82,20 @@ trait CodingDirectives {
...
+                      ErrorInfo("The request's encoding is corrupt", e.getMessage))
+                })
+                recovered match {
+                  case c: HttpEntity.Chunked ⇒ c.copy(chunks = HttpEntity.limitableChunkSource(c.chunks))
+                  case e: HttpEntity         ⇒ e
+                }
+              }
             }
-          }
+          } & withSizeLimit(settings.decodeMaxSize)
```
johanandren: Ok, so this together with the limitableChunkSource is the fix right? My brain is tired: Is it only ever streamed for chunked, not a large non-chunked entity?

akka-http/src/main/resources/reference.conf
raboof: I'm not sure that is possible: if `max-content-size` is updated in `application.conf`, AFAICS referring to `max-content-size` from this place in the `reference.conf` still wouldn't point to the new value if I'm reading https://github.com/lightbend/config#note-about-resolving-substitutions-in-referenceconf-and-applicationconf right.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/CodingDirectives.scala
raboof: A large non-chunked entity should still be restricted by `withSizeLimit` - but I'll make sure the tests cover both cases.

akka-http/src/main/resources/reference.conf
johanandren: Yes, you are right. It wouldn't work.

johanandren: What about solving that when reading it then, leave this undefined/empty and use the other one unless this has a specific value?

raboof: That could work - leaving it undefined would confuse config-checker, but we could introduce another 'magic value' here and handle it specially in code. Do we have any other places where we do something similar?

akka-http/src/main/scala/akka/http/scaladsl/server/directives/CodingDirectives.scala
jrudolph: The output of `transformDataBytes` will be a chunked entity (but maybe it's still a brittle assumption). I think we should change `tranformDataBytes`, instead, to ensure that the result is always limitable. WDYT?

akka-http-tests/src/test/scala/akka/http/scaladsl/server/SizeLimitSpec.scala
```diff
@@ -0,0 +1,246 @@
...
+    val binding = Http().bindAndHandle(route, "localhost", port = 0).futureValue
+
+    "accept a small request" in {
+      val response = Http().singleRequest(Post(s"http:/${binding.localAddress}/noDirective", entityOfSize(maxContentLength))).futureValue
+      response.status shouldEqual StatusCodes.OK
+      response.entity.dataBytes.runReduce(_ ++ _).futureValue.utf8String shouldEqual (s"Got request with entity of $maxContentLength characters")
+    }
+
+    "reject a small request that decodes into a large entity" in {
+      val data = ByteString.fromString("0" * (maxContentLength + 1))
```
jrudolph: `decodeMaxSize + 1`?

akka-http/src/main/scala/akka/http/scaladsl/server/directives/CodingDirectives.scala
jrudolph: Ultimately, it's the `HttpEntity.withSizeLimit` that is broken right now (it's debatable whether it was a good idea to add it there but we cannot change that right now :) ).

What if we would always add the limitable stage when `withSizeLimit` is called (at least if we cannot verify whether the last stage in that source already is a Limitable which isn't possible using public streams API afaik). That might mean that we have multiple such stages in the source e.g. in the default case with `decodeRequest` after this PR. For a slight performance overhead it would still be the safer option.

The performance overhead would be less than currently if no limits are defined in the first place because then we would never add the stage.

raboof: Hmm that sounds interesting, tried that! I'm not too familiar with how exactly Attributes work, so I'd appreciate if you could double-check I'm not introducing any problems there.

akka-http-tests/src/test/scala/akka/http/scaladsl/server/SizeLimitSpec.scala
```diff
@@ -63,7 +63,8 @@ class SizeLimitSpec extends WordSpec with Matchers with RequestBuilding with Bef
     "not accept entities bigger than configured with akka.http.parsing.max-content-length" in {
       // It went from 1 occurrence to 2 after discarding the entity, I think is due to the retrying nature of `handleRejections`
       // that causes one Exception for the original entity and another one from the rejected one.
-      EventFilter[EntityStreamSizeException](occurrences = 2).intercept {
+      // Then from 2 to 4 since the change to add a `Limitable` whenever `withSizeLimit` is applied
+      EventFilter[EntityStreamSizeException](occurrences = 4).intercept {
```
jrudolph: Actually, we should fix `Limitable` not to log anything (by using failStage instead of throwing).

akka-http-tests/src/test/scala/akka/http/scaladsl/server/SizeLimitSpec.scala
```diff
@@ -0,0 +1,248 @@
...
+      zippedData.size should be <= maxContentLength
+      data.size should be > decodeMaxSize
+
+      val response = Http().singleRequest(request).futureValue
+      response.status shouldEqual StatusCodes.OK
+      response.entity.dataBytes.runReduce(_ ++ _).futureValue.utf8String shouldEqual (s"Got request with entity of ${decodeMaxSize + 1} characters")
+    }
+
+    // This is not entirely obvious: the 'withoutSizeLimit' inside the decodeRequest
+    // will also reset the size limit outside the decodeRequest.
```
jrudolph: I find the comment is a bit misleading. If you use `withoutSizeLimit`, you basically waive your guarantees, and that works correctly and only inside the directive.

The non-obvious / counter-intuitive behavior might be that if decoding actually makes data smaller, then applying a tighter limit that would be passed after decoding could still catch on the original bigger data. This would probably be rare but when it happens it might be difficult to debug, because you may not find out where the final limit came from.

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
```diff
@@ -392,7 +392,7 @@ object HttpEntity {
       if (contentType == this.contentType) this else copy(contentType = contentType)
 
     override def withSizeLimit(maxBytes: Long): HttpEntity.Default =
-      copy(data = data withAttributes Attributes(SizeLimit(maxBytes, Some(contentLength))))
+      copy(data = limitableByteSource(data) withAttributes Attributes(SizeLimit(maxBytes, Some(contentLength))))
```
jrudolph: For `Strict.withSizeLimit` in line 337 you can remove the extra `limitableByteSource` because `withSizeLimit` is called afterwards (whether delegating to Default is the right approach for Strict entities is another question...).

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
```diff
@@ -609,7 +609,7 @@ object HttpEntity {
           case Some(limit: SizeLimit) if limit.isDisabled ⇒
           // "no limit"
           case Some(SizeLimit(bytes, cl @ Some(contentLength))) ⇒
-            if (contentLength > bytes) throw EntityStreamSizeException(bytes, cl)
+            if (contentLength > bytes) failStage(EntityStreamSizeException(bytes, cl))
```
jrudolph: :+1:

akka-http-tests/src/test/scala/akka/http/scaladsl/server/SizeLimitSpec.scala
```diff
@@ -0,0 +1,244 @@
...
+
+    val binding = Http().bindAndHandle(route, "localhost", port = 0).futureValue
+
+    "accept small POST requests" in {
+      Http().singleRequest(Post(s"http:/${binding.localAddress}/noDirective", entityOfSize(maxContentLength)))
+        .futureValue.status shouldEqual StatusCodes.OK
+    }
+
+    "not accept entities bigger than configured with akka.http.parsing.max-content-length" in {
+      // It went from 1 occurrence to 2 after discarding the entity, I think is due to the retrying nature of `handleRejections`
```
jrudolph: Comment doesn't apply to anything any more ;)

jrudolph: Though, it would be interesting to know how all these occurrence did happen, not sure it really was because of `handleRejections`.

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/MiscDirectivesSpec.scala
```diff
@@ -99,10 +99,8 @@ class MiscDirectivesSpec extends RoutingSpec {
         status shouldEqual StatusCodes.OK
       }
 
-      EventFilter[EntityStreamSizeException](occurrences = 1).intercept {
-        Post("/abc", entityOfSize(501)) ~> Route.seal(route) ~> check {
-          status shouldEqual StatusCodes.BadRequest
-        }
+      Post("/abc", entityOfSize(501)) ~> Route.seal(route) ~> check {
+        status shouldEqual StatusCodes.BadRequest
```
jrudolph: We could check instead that there's a reasonable message with an explanation.

akka-http/src/main/resources/reference.conf
```diff
@@ -37,6 +37,13 @@ akka.http.routing {
   # for an entity data stream.
   decode-max-bytes-per-chunk = 1m
 
+  # Maximum content length after applying a decoding directive. When the directive
+  # decompresses, for example, an entity compressed with gzip, the resulting stream can be much
+  # bigger than the max-content-length. Like with max-content-length, this is not necessarilly a
+  # problem when consuming the entity in a streaming fashion, but does risk high memory use
+  # when the entity is made strict or marshalled into an in-memory object.
```
jrudolph: Good explanation.

akka-http/src/main/scala/akka/http/scaladsl/coding/Decoder.scala
```diff
@@ -13,13 +13,19 @@ import headers.HttpEncoding
...
+import scala.util.control.NonFatal
 
 trait Decoder {
   def encoding: HttpEncoding
 
   def decodeMessage(message: HttpMessage): message.Self =
     if (message.headers exists Encoder.isContentEncodingHeader)
-      message.transformEntityDataBytes(decoderFlow).withHeaders(message.headers filterNot Encoder.isContentEncodingHeader)
+      message.transformEntityDataBytes(decoderFlow.recover {
+        case NonFatal(e) ⇒
```
jrudolph: Not sure if that catch all exception handler is really warranted (but you didn't create it anyway, so it's fine for now but maybe create another ticket to investigate if we can be more precise, so that not all other upstream errors are converted here.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/CodingDirectives.scala
```diff
@@ -80,16 +80,7 @@ trait CodingDirectives {
...
-            effectiveDecoder.decodeMessage(request).mapEntity { entity ⇒
-              entity.transformDataBytes(Flow[ByteString].recover {
-                case NonFatal(e) ⇒
-                  throw IllegalRequestException(
-                    StatusCodes.BadRequest,
-                    ErrorInfo("The request's encoding is corrupt", e.getMessage))
-              })
-            }
-          }
+          mapRequest(effectiveDecoder.decodeMessage(_)) & withSizeLimit(settings.decodeMaxSize)
```
jrudolph: Much nicer here now!

akka-http/src/main/resources/reference.conf
jrudolph: Maybe add another comment, that this limit (like `max-content-length`) can be overridden on a case-by-case basis using `withSizeLimit`.

akka-http/src/main/resources/reference.conf
jrudolph: We do break applications anyway where uncompressed data is bigger than either the new limit or - if we go with this proposal - whatever `max-content-length` was set to. So, I'd vote for keeping it simple and explicit as it is here.

The good thing is, that `withSizeLimit` now works for overriding both limits.

jrudolph:
I created https://github.com/akka/akka-http/issues/2192 to improve the error message later on to point to the actual place where the limit was set.

jrudolph:
Good to go from my side!

jrudolph:
I'm going to split it up manually in two pieces, one fixing HttpEntity.withSizeLimit and the other adding the new limit.

jrudolph:
Superceded by #2196.

Issue #2184, opened 2018-08-31
Request timeout response is out of band


johanandren:
Seen in the forums here: https://discuss.lightbend.com/t/akka-http-mapresponse-ignores-some-responses/1975/6

Intuitively request timeouts happens while in the routing tree, but actually that's not how they are implemented, which means that for example `mapResponse` does not also transform the request timeout response if the timeout hits.

Can we improve, and make `mapResponse` also map timeout responses out of the box, or something else that makes it more intuitive combining a route with the timeout?

If not, can we use documentation to make it more clear that it happens at a lower level?

jrudolph:
> Can we improve, and make mapResponse also map timeout responses out of the box, or something else that makes it more intuitive combining a route with the timeout?

The idea was that we have a low-level, low-effort safe-guard against long-running requests that will trigger even under load. The rationale was that any high-level (= routing) measures may under load try to do "too much" and delay the timeout. (A counter point would be that the way it is built, the timeout may have triggered and processing still cannot be cancelled, so even more work is done.)

jrudolph:
Historically, there was a `Timeout` message in spray that your request handler actor could react to. If it didn't in a while, a `TimeoutTimeout` message was generated and a default timeout response was produced by the lower level.

PR #2186, opened 2018-08-31
Allow a maximum size parameter for HttpEntity 'toStrict'

raboof:
Refs #2137


akka-http/src/main/scala/akka/http/javadsl/server/directives/BasicDirectives.scala
```diff
@@ -327,4 +327,16 @@ abstract class BasicDirectives {
     D.toStrictEntity(timeout) { inner.get.delegate }
   }
 
+  /**
+   * WARNING: This will read the entire request entity into memory and effectively disable streaming.
+   *
+   * Extracts the [[akka.http.javadsl.server.RequestContext]] itself with the strict HTTP entity,
+   * or fails the route if unable to drain the entire request body within the timeout.
```
johanandren: I think this one is tricky, as it fails the request the same way as timeout, there was some recent discussions around this described in #2184, so it does fail the request but not quite the `route`

akka-http/src/main/scala/akka/http/scaladsl/server/directives/BasicDirectives.scala
```diff
@@ -362,6 +362,24 @@ trait BasicDirectives {
...
+   */
+  def toStrictEntity(timeout: FiniteDuration, maxBytes: Long): Directive0 =
+    Directive { inner ⇒ ctx ⇒
+      import ctx.{ executionContext, materializer }
+
+      ctx.request.entity.toStrict(timeout, maxBytes).flatMap { strictEntity ⇒
+        val newCtx = ctx.mapRequest(_.copy(entity = strictEntity))
+        inner(())(newCtx)
+      }
+    }
```
johanandren: Shouldn't the `maxBytes`-parameter-less overload pick a max-bytes up from config to make that safe as well?

jrudolph: Yep, I'm for a limited default.

akka-http/src/main/scala/akka/http/javadsl/server/directives/BasicDirectives.scala
jrudolph: > as it fails the request the same way as timeout,

Why do you think so? Wouldn't it just fail in the route right here?

The low-level request timeouts instead are handled on the core layer and never reach the higher layer.

raboof: Ah this might be fine indeed: whoever consumes the stream (typically some unmarshaller) will fail, which will then e.g. lead to an `MalformedRequestContentRejection` and be processed like usual.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/BasicDirectives.scala
raboof: :+1: for the directive - not sure it's worth it for `HttpEntity.toStrict` though, since it's not obvious how to pass in the configuration settings there?

akka-http/src/main/scala/akka/http/scaladsl/server/directives/BasicDirectives.scala
```diff
@@ -344,7 +344,7 @@ trait BasicDirectives {
     toStrictEntity(timeout) & extract(_.request.entity.asInstanceOf[HttpEntity.Strict])
 
   /**
-   * WARNING: This will read the entire request entity into memory regardless of size and effectively disable streaming.
+   * WARNING: This will read the entire request entity into memory and effectively disable streaming.
```
jrudolph: Needs a comment that it is now governed by `to-strict-max-bytes` setting.

raboof:
PLS BUILD

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
```diff
@@ -72,9 +72,24 @@ sealed trait HttpEntity extends jm.HttpEntity {
...
+  /**
+   * Collects all possible parts and returns a potentially future Strict entity for easier processing.
+   * The Future is failed with an TimeoutException if the stream isn't completed after the given timeout,
+   * or with a FooException when the end of the entity is not reached within the maximum number of bytes.
+   */
+  def toStrict(timeout: FiniteDuration, maxBytes: Long)(implicit fm: Materializer): Future[HttpEntity.Strict] = contentLengthOption match {
+    case Some(contentLength) if contentLength > maxBytes ⇒
+      FastFuture.failed(new EntityStreamException(new ErrorInfo("Request too large", s"Request of size $contentLength was longer than the maximum of $maxBytes")))
+    case _ ⇒
+      // TODO can we trust the content-length or should we check the actual length anyway?
```
jrudolph: WDYM? `ToStrict` will check the actual size later, doesn't it?

akka-http-core/src/test/scala/akka/http/scaladsl/model/HttpEntitySpec.scala
```diff
@@ -99,6 +100,38 @@ class HttpEntitySpec extends FreeSpec with MustMatchers with BeforeAndAfterAll {
         }.getMessage must be("HttpEntity.toStrict timed out after 100 milliseconds while still waiting for outstanding data")
       }
     }
+    "support toStrict with a max size" - {
```
jrudolph: :+1:

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
raboof: Yes, right now I chose the 'safe' path. I wondered whether there could be cases where we could decide early that it's not needed to add the `ToStrict`. I guess the only case is `Strict`, which might not be worth special-casing. I'll remove the comment ;)

raboof:
PLS BUILD

akka-http/src/main/scala/akka/http/javadsl/server/directives/BasicDirectives.scala
johanandren: Ok, thanks for clarifying!

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
```diff
@@ -72,9 +72,23 @@ sealed trait HttpEntity extends jm.HttpEntity {
    */
   def toStrict(timeout: FiniteDuration)(implicit fm: Materializer): Future[HttpEntity.Strict] =
     dataBytes
-      .via(new akka.http.impl.util.ToStrict(timeout, contentType))
+      .via(new akka.http.impl.util.ToStrict(timeout, maxBytes = None, contentType))
```
jrudolph: So, we don't apply a default limit here for now? Could we use `fm.system.settings.config` to get at the config setting directly?

raboof: Hmm and move the setting to `akka.http.server.parsing`/`akka.http.parsing` then I guess... :+1: I'll make the change :smile:

jrudolph: Could we also put a copy of the `routing` default value into the akka-http-core reference.conf?

raboof: Wouldn't that be confusing? if we move it to `parsing` we don't have this problem (since `parsing` is already in akka-http-core)

jrudolph: Yeah, maybe let's keep it simpler and put it in parsing.

akka-http-core/src/main/resources/reference.conf
```diff
@@ -413,6 +413,10 @@ akka.http {
     # programmatically via `withSizeLimit`.)
     max-content-length = 8m
 
+    # The maximum number of bytes to allow when reading the entire entity into memory with `toStrict`
+    # (which is used by the `toStrictEntity` and `extractStrictEntity` directives)
+    max-to-strict-bytes = 8m
```
jrudolph: Seems in good company there :+1:

PR #2188, opened 2018-09-04
fix LfuCache ttl can be equal to tti

tg44:
If you don't want to use tti you need to write some ugly `time-to-idle = 5 minutes - 1 seconds` to the config. So I fixed the requirement.

(Maybe if tti is not finite we should use ttl for both values. That would make easier setup in some use-cases. But that can be misleading too, so I did the minimal effort "fix" only.)

(Or maybe we should play with boolalgebra and change the whole thing to:
`!(timeToLive.isFinite && timeToIdle.isFinite && timeToLive < timeToIdle)` bcs the following implementation only set the desired finite values...)

### DIFF
--- a/akka-http-caching/src/main/scala/akka/http/caching/LfuCache.scala
+++ b/akka-http-caching/src/main/scala/akka/http/caching/LfuCache.scala
@@ -69,8 +69,8 @@ object LfuCache {
   private def expiringLfuCache[K, V](maxCapacity: Long, initialCapacity: Int,
                                      timeToLive: Duration, timeToIdle: Duration): LfuCache[K, V] = {
     require(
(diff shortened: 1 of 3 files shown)


jrudolph:
OK TO TEST

jonas:
The documentation currently states:
```markdown
Time-based entry expiration is enabled when a time-to-live and/or time-to-idle
expirations are set to a finite duration. The former provides an
upper limit to the time period an entry is allowed to remain in the cache while
the latter limits the maximum time an entry is kept without having been
accessed, ie. either read or updated. If both values are finite the time-to-live
has to be strictly greater than the time-to-idle.
```
Please update it to reflect the change by removing the last sentence.

Otherwise, I don't remember the motivation for `TTL > TTI`, so looks good to me.

PR #2196, opened 2018-09-05
Follow up to #2185

jrudolph:
I mostly split this up in three separate commits. Regarding the HttpEntity changes, I cleaned up a few things around where the limitable stages were added before.


jrudolph:
@raboof, here's the diff to your original PR:

https://github.com/akka/akka-http/compare/wipDecodeRequestWithMaxSize...jrudolph:for-diff-2185

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
```diff
@@ -576,29 +566,31 @@ object HttpEntity {
   object LastChunk extends LastChunk("", Nil)
 
   /**
-   * Turns the given source into one that respects the `withSizeLimit` calls when used as a parameter
-   * to entity constructors.
+   * Deprecated: no-op, not explicitly needed any more.
    */
+  @deprecated("Not needed explicitly any more. ", "10.1.4")
```
johanandren: 10.1.4 already released so should be 10.1.5, right?

raboof: CONTRIBUTING.md says "src/main/mima-filters/<last-released-version>.backwards.excludes" but this always confuses me so I'm not entirely confident...

raboof: In contrast to MiMa :laughing:

jrudolph:
PLS BUILD

PR #2199, opened 2018-09-05
Add 10.1.5 release notes

raboof:


docs/src/main/paradox/release-notes/10.1.x.md
```diff
@@ -1,5 +1,20 @@
 # 10.1.x Release Notes
 
+## 10.1.5
+
+10.1.5 is the sixth release in the 10.1.x series of Akka HTTP.
+
+It is mainly a mitigation for the [DoS vulnerability when using `decodeRequest`](https://akka.io/blog/news/2018/08/30/akka-http-dos-vulnerability-found)
```
jrudolph: I think we need to add some migration notes:

### Migration Notes

To avoid excessive memory usage we introduced two new limits that apply per default:

 * `akka.http.routing.decode-max-size`: This limit applies when you use `decodeRequest` to limit the amount of decompressed data. The default limit is 8 megabytes.
 * `akka.http.parsing.max-to-strict-bytes`: This limit applies when you use `HttpEntity.toStrict` or the `toStrictEntity` directives (and related). It will only collect up to the given amount data and fail otherwise. The default limit is 8 megabytes.

Depending on your application requirements, you may want to change these settings.

```
### Migration Notes

To avoid excessive memory usage we introduced two new limits that apply per default:

 * `akka.http.routing.decode-max-size`: This limit applies when you use `decodeRequest` to limit the amount of decompressed data. The default limit is 8 megabytes.
 * `akka.http.parsing.max-to-strict-bytes`: This limit applies when you use `HttpEntity.toStrict` or the `toStrictEntity` directive (and related directives). It will only collect up to the given amount data and fail otherwise. The default limit is 8 megabytes.

Depending on your application requirements, you may want to change these settings.
```

jlprat: Maybe add this in here as well? https://github.com/akka/akka-http/blob/master/docs/src/main/paradox/migration-guide/migration-guide-10.1.x.md

jlprat: I can do it in a separate PR if you like

raboof: Good one, I'll add it here

Issue #2054, opened 2018-06-06
Java DSL: completeWithFuture is missing unwrapCompletionException


jypma:
In modern Java, a `CompletionStage`'s failure will always be a `CompletionException`. That doesn't go very well with writing route `ExceptionHandlers`, since one would always be matching `CompletionException`. 

In order to work around this, `FutureDirectives` has a [`unwrapCompletionException`](https://github.com/akka/akka-http/blob/f3e83935ffc27bc94f586d9387b54b499c6250d4/akka-http/src/main/scala/akka/http/javadsl/server/directives/FutureDirectives.scala#L110) that auto-unwraps failed futures, so exception handlers can deal with the _actual_ exception. However, this unwrapping is missing from `completeWithFuture` (multiple variants, in [javadsl/RouteDirectives.scala](https://github.com/akka/akka-http/blob/f3e83935ffc27bc94f586d9387b54b499c6250d4/akka-http/src/main/scala/akka/http/javadsl/server/directives/RouteDirectives.scala#L232) ).

Now, any exception thrown by a `CompletionStage` passed to `completeWithFuture` will always be `CompletionException`, and not be handled correctly by any applied exception handler.

A quick fix would be to add `unwrapCompletionException` to `RouteDirectives` where appropriate.

raboof:
Sounds good, I remember this being quite cumbersome.

jlprat:
I will attempt to tackle this one :)

jlprat:
Closed by https://github.com/akka/akka-http/commit/8bd0beca16d33ba4a62fe10c7be885d172530d32

PR #2205, opened 2018-09-06
Fail with unwrapped exception when using CompletionStage #2054

jlprat:
Refs: #2054
Unwraps the original exception that caused the `CompletionStage` to fail.
This original exception can then be used on the custom `ExceptionHandler`.

Please note that the unwrapping is only done for the methods that take a `CompletionStage` not for the ones getting a Scala `Future`.

### DIFF
--- a/akka-http/src/main/scala/akka/http/javadsl/server/directives/RouteDirectives.scala
+++ b/akka-http/src/main/scala/akka/http/javadsl/server/directives/RouteDirectives.scala
@@ -4,7 +4,7 @@
 
 package akka.http.javadsl.server.directives
 
-import java.util.concurrent.CompletionStage
+import java.util.concurrent.{ CompletionException, CompletionStage }
 
 import akka.dispatch.ExecutionContexts
 import akka.http.javadsl.marshalling.Marshaller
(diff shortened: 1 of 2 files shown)


jlprat:
Travis failure is because 10.1.5 artifacts are not yet in Maven Central

akka-http/src/main/scala/akka/http/javadsl/server/directives/RouteDirectives.scala
```diff
@@ -239,47 +239,56 @@ abstract class RouteDirectives extends RespondWithDirectives {
...
   /**
    * Completes the request by marshalling the given value into an http response.
    */
   @CorrespondsTo("complete")
   def completeWithFuture[T](value: CompletionStage[T], marshaller: Marshaller[T, HttpResponse]) = RouteAdapter {
-    D.complete(value.asScala.fast.map(v ⇒ ToResponseMarshallable(v)(marshaller)))
+    D.complete(value.asScala.fast.map(v ⇒ ToResponseMarshallable(v)(marshaller)).recover(unwrapCompletionException))
+  }
+
+  // TODO: This might need to be raised as an issue to scala-java8-compat instead.
```
jrudolph: :+1: Can you file an issue there as well?

jlprat: https://github.com/scala/scala-java8-compat/issues/120

Issue #2217, opened 2018-09-21
Doesn't support chunked HTTP/2 responses


ctoomey:
Chunked HTTP/2 responses are disallowed, and shouldn't be, by the following check in `akka.http.scaladsl.model.HttpResponse`:
```
  require(
    protocol == HttpProtocols.`HTTP/1.1` || !entity.isInstanceOf[HttpEntity.Chunked],
    "HTTP/1.0 responses must not have a chunked entity")
```
The check should instead be 
```
protocol != HttpProtocols.`HTTP/1.0` || !entity.isInstanceOf[HttpEntity.Chunked]
```

jrudolph:
The reasoning is that akka-http2-support can actually handle `HttpEntity.Chunked` by ignoring the chunk extensions but supporting the trailers using the usual HTTP/2 stream framing.

We already [make use](https://github.com/akka/akka-grpc/blob/master/runtime/src/main/scala/akka/grpc/internal/GrpcResponseHelpers.scala#L56) of this feature in akka-grpc.

Why does it work there? Because for outgoing responses, akka-http doesn't require that the `protocol` of a response is set correctly (the protocol field is just ignored).

So, as long as you don't set the protocol of a response to HTTP/2 you can already now use `Chunked` with HTTP/2.

ctoomey:
I needed it to work w/ HTTP/2 protocol as part of an alternate Play integration I'll be proposing separately that involves returning `EssentialAction`s so that Play's `Filter` chaining support can be leveraged like usual. So the `HttpResponse` generation from the `Result` is handled downstream by Play's `AkkaHttpServer` and it copies the protocol from the request to the response.

jrudolph:
Makes sense, thanks for providing the context, @ctoomey.

Issue #2108, opened 2018-07-04
Failed: GracefulTerminationSpec


johanandren:
https://jenkins.akka.io:8498/job/pr-validator-akka-http/2729/testReport/junit/akka.http.scaladsl/GracefulTerminationSpec/Graceful_termination_should_in_flight_request_responses_should_include_Connection__close_and_connection_should_be_closed/

```
 None was not equal to Some(Connection: close)
	at org.scalatest.MatchersHelper$.indicateFailure(MatchersHelper.scala:346)
	at org.scalatest.Matchers$AnyShouldWrapper.shouldBe(Matchers.scala:6864)
	at akka.http.scaladsl.GracefulTerminationSpec$$anon$6.<init>(GracefulTerminationSpec.scala:132)
	at akka.http.scaladsl.GracefulTerminationSpec.$anonfun$new$10(GracefulTerminationSpec.scala:123)
	at org.scalatest.OutcomeOf.outcomeOf(OutcomeOf.scala:85)
	at org.scalatest.OutcomeOf.outcomeOf$(OutcomeOf.scala:83)
	at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
	at org.scalatest.Transformer.apply(Transformer.scala:22)
	at org.scalatest.Transformer.apply(Transformer.scala:20)
	at org.scalatest.WordSpecLike$$anon$1.apply(WordSpecLike.scala:1078)
	at akka.http.impl.util.WithLogCapturing.$anonfun$withFixture$1(WithLogCapturing.scala:36)
	at akka.testkit.EventFilter.intercept(TestEventListener.scala:111)
	at akka.http.impl.util.WithLogCapturing.withFixture(WithLogCapturing.scala:34)
	at akka.http.impl.util.WithLogCapturing.withFixture$(WithLogCapturing.scala:21)
	at akka.http.scaladsl.GracefulTerminationSpec.withFixture(GracefulTerminationSpec.scala:30)
```

raboof:
Different output but same spec: https://jenkins.akka.io:8498/job/pr-validator-akka-http/2735/testReport/junit/akka.http.scaladsl/GracefulTerminationSpec/Graceful_termination_should_provide_whenTerminated_future_that_completes_once_server_has_completed_termination__existing_connection__user_reply__terminate__no_reply_/

raboof:
Again: https://jenkins.akka.io:8498/job/pr-validator-akka-http/2792/consoleFull

[...]

PR #2218, opened 2018-09-22
Enable HTTP/2 chunked responses #2217

ctoomey:
Refs. #2217

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
@@ -262,7 +262,7 @@ final class HttpRequest(
   HttpRequest.verifyUri(uri)
   require(entity.isKnownEmpty || method.isEntityAccepted, s"Requests with method '${method.value}' must have an empty entity")
   require(
-    protocol != HttpProtocols.`HTTP/1.0` || !entity.isInstanceOf[HttpEntity.Chunked],
+    protocol != HttpProtocols.`HTTP/1.0` || !entity.isChunked,
     "HTTP/1.0 requests must not have a chunked entity")
 
   type Self = HttpRequest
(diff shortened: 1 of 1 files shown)


johanandren:
Refs #2217

raboof:
OK TO TEST

jrudolph:
PLS BUILD

jrudolph:
> Does it really end up as 'chunked' on the wire (which would be wrong) or do we ignore the fact that it was presented as 'chunked' and stream it like a regular entity (which might be OK, but perhaps would warrant a test?)?

`Transfer-Encoding` doesn't really make sense for HTTP/2 as it defines it's own stream transport protocol. Instead, we (ab)use our model for `Chunked` entities to support trailers in HTTP/2 (#1857).

ctoomey:
Will you guys please merge this so it'll be in the next release?

Also it's not super clear to me from the discussion above -- are you saying the akka-grpc response generation is HTTP/2 conformant because though it uses chunking, it doesn't use the chunked transfer encoding which is the thing that's verboten?

jrudolph:
> are you saying the akka-grpc response generation is HTTP/2 conformant 

Yes, it uses the akka-http2-support implementation in a valid way.

> because though it uses chunking, 

It uses `HttpEntity.Chunked` which in akka-http's HTTP/2 implementation is interpreted differently than for HTTP/1.1.

> it doesn't use the chunked transfer encoding which is the thing that's verboten?

Exactly. akka-http2-support will translate `HttpEntity.Chunked` to regular HTTP/2 streams but ignores some part of the model (namely the `extension` field of `ChunkStreamPart`).

jrudolph:
Btw. it already works now in akka-grpc because you don't have to set a matching protocol for `HttpResponse`.

raboof:
Failure is #2108

raboof:
PLS BUILD

PR #2228, opened 2018-09-25
Remove another side-effect in pool Slot machine

jrudolph:


akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/NewHostConnectionPool.scala
```diff
@@ -375,9 +370,7 @@ private[client] object NewHostConnectionPool {
           def pushRequest(request: HttpRequest): Unit = {
             val newRequest =
               request.entity match {
-                case _: HttpEntity.Strict ⇒
-                  withSlot(_.onRequestEntityCompleted())
```
jrudolph: `onRequestEntityCompleted` will not be called for Strict entities any more.

akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/SlotState.scala
```diff
@@ -200,6 +196,13 @@ private[pool] object SlotState {
       Unconnected
     }
   }
+  final case class PushingRequestToConnection(ongoingRequest: RequestContext) extends ConnectedState with BusyState {
+    override def waitingForEndOfRequestEntity: Boolean = ???
```
raboof: This flag is used in `BusyState` to make sure that when the request is being failed while the response is already in flight, we correctly wait for the response to also finish before accepting new connections in this slot.

It seems like `false` would be a fine constant value in this case: when we haven't even dispatched the request yet, we don't need to wait for a response.

The fact that this does not fail any tests suggests we might want to cover some more corner cases there, though?

jrudolph: I guess it works because this state is so short-lived that no events are ever dispatched to it. To improve the modelling we would have to separate actual states from requesting side-effects.

But I agree, setting this to false and commenting it would be the right way.

PR #2231, opened 2018-09-26
=htc #1391 client pool: exponential backoff after failed connection attempts

### LINKED ISSUE #1391: Thread loops infinitely consuming CPU with non-zero `akka.http.host-connection-pool.min-connections` setting
This issue created as followup for 2-months-old thread in google groups: https://groups.google.com/forum/#!topic/akka-user/062Op2fyD-E

Sorry for absent of minimal example, only text description here.

**How to reproduce:**
1. Create temporary DNS record (we were doing this with Amazon Route53 service)
2. Spin up Actor System with `akka.http.host-connection-pool.min-connections` set to non-zero value
3. Make request (from the ActorSystem) to the temp. DNS name 
4. Delete DNS record
5. (Optional) Make any http request 

As a result - one of threads\actors starts to consume all available CPU by processing some inner-state actor messages.

jrudolph:
Fixes #1391.

It works but I might want to tweak the implementation a bit. The settings file describes the feature quite well.

TODO:
 * [x] add mima excludes
 * [x] add docs section
 * [x] add comment to config that setting only applies to new pool impl
 * [x] enforce non-zero backoff for connections that were created by min-connections (to avoid the loop if misconfigured)
 * [x] fix race in test where second connection is still attempted because failure only came back a bit later

### DIFF
--- a/akka-http-core/src/main/resources/reference.conf
+++ b/akka-http-core/src/main/resources/reference.conf
@@ -372,9 +372,14 @@ akka.http {
     #   - After 7th, between 3200ms and 6400ms
     #   - After 8th, between 5000ms and 10 seconds (max capped by max-connection-backoff, min by half of that)
     #   - After 9th, etc., stays between 5000ms and 10 seconds
+    #
+    # This setting only applies to the new pool implementation and is ignored for the legacy one.
     base-connection-backoff = 100ms
 
-    # Maximum backoff duration between failed connection attempts.
+    # Maximum backoff duration between failed connection attempts. For more information see the above comment for the
+    # `base-connection-backoff` setting.
+    #
+    # This setting only applies to the new pool implementation and is ignored for the legacy one.
     max-connection-backoff = 2 min
 
     # The time after which an idle connection pool (without pending requests)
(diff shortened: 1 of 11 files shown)


jrudolph:
(Real failures but I'd like to see how flaky they are)

PLS BUILD

akka-http-core/src/main/resources/reference.conf
```diff
@@ -365,6 +365,38 @@ akka.http {
...
+    #   - After 2nd, between  200ms and  400ms
+    #   - After 3rd, between  200ms and  400ms
+    #   - After 4th, between  400ms and  800ms
+    #   - After 5th, between  800ms and 1600ms
+    #   - After 6th, between 1600ms and 3200ms
+    #   - After 7th, between 3200ms and 6400ms
+    #   - After 8th, between 5000ms and 10 seconds (max capped by max-connection-backoff, min by half of that)
+    #   - After 9th, etc., stays between 5000ms and 10 seconds
+    #
+    # This setting only applies to the new pool implementation and is ignored for the legacy one.
```
patriknw: good docs

jrudolph:
PLS BUILD

jrudolph:
PLS BUILD

jrudolph:
Rebased and added a small docs paragraph. The potential cleanups were not only for this feature but for the pool in general so they won't be done in this PR.

schoeneu:
@jrudolph I was slightly surprised to see that this changed the default behaviour even for pools with `min-connections = 0`. Was this the intended effect?

PR #2243, opened 2018-10-05
add [openjdk11] to Travis build matrix

sullis:


johanandren:
OK TO TEST

jrudolph:
The reason for the failure is https://github.com/akka/akka/issues/25784 which will be fixed in Akka 2.5.18.

jrudolph:
Restarted the failed Travis openjdk11 build after merging #2312.

jrudolph:
This currently doesn't work. We don't want to run tests on Travis which happens if we add more configurations to the Travis build matrix.

Instead, we are going to setup an akka-http JDK11 nightly build.

Thanks anyway, @sullis.

PR #2249, opened 2018-10-10
Fix CsvSeq unmarshaller to include trailing empty string values in the result Seq[T]

maximskripnik:
Hello,

This PR solves the issue when `CsvSeq[T]` unmarshaller would ignore last empty string value (the one after trailing comma), i.e. would not pass it to underlying unmarshaller. This is agains csv semantic, so I think it is fair to assume that this is bug.

```scala
val values = Await.result(CsvSeq[String].apply(",value1,value2,"), Duration.Inf)
 // Current behaviour
assert(values == Seq("", "value1", "value2"))
// Expected behaviour
// assert(values == Seq("", "value1", "value2", ""))
```

This is my very first PR, so I am scared and sorry if I did something wrong, but the fix seems straightforward enough to try submitting it

### DIFF
--- a/akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/ParameterDirectivesSpec.scala
+++ b/akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/ParameterDirectivesSpec.scala
@@ -89,6 +89,11 @@ class ParameterDirectivesSpec extends FreeSpec with GenericRoutingSpec with Insi
         responseAs[String] shouldEqual "The parameters are Caplin, John"
       }
     }
+    "extract a number of names, including last empty name" in {
+      Get("/?names=Caplin,John,") ~> route ~> check {
+        responseAs[String] shouldEqual "The parameters are Caplin, John, "
+      }
+    }
   }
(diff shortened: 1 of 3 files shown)


johanandren:
OK TO TEST

raboof:
We don't seem to have this feature documented, do we? Should we?

maximskripnik:
@raboof If you mean documentation for `CsvSeq` in general, there is: https://doc.akka.io/docs/akka-http/current/routing-dsl/directives/parameter-directives/parameters.html#csv-parameter
But it covers only `String` params, not any `T` (as long as this `T` is unmarshallable from `String`).
This feature was added by [this commit](https://github.com/akka/akka-http/commit/204ce70807a68530a6f3ddc76a82b325770e105b), I believe, but it does not contain any updates in documentation.

We could add some samples for custom types to be used in `CsvSeq` in addition to `String`, if that is what you're suggesting, I think it would be nice to have

raboof:
Ah, missed that. I agree showing that it works for `T` would be nice but that's something for another PR.

Thanks for this one!

PR #2252, opened 2018-10-11
Fix limiting of chunk size in the NoCoding encoder

ollyw:
This fixes an issue where the limitByteChunksStage function used by the NoCoding encoder was losing data if larger than the maximum chunk size. The tests have also been fixed to cover this scenario.

I don't know if this is an issue that affects people using the akka-http in the wild, but is quite subtle. I came across the issue trying to test out chunked StreamRefs in akka-cluster in a WIP PR https://github.com/akka/akka/pull/25643. That PR copies the broken code from akka-http verbatim, so is likely to be also broken.

### DIFF
--- a/akka-http-tests/src/test/scala/akka/http/scaladsl/coding/CoderSpec.scala
+++ b/akka-http-tests/src/test/scala/akka/http/scaladsl/coding/CoderSpec.scala
@@ -5,7 +5,6 @@
 package akka.http.scaladsl.coding
 
 import java.io.{ OutputStream, InputStream, ByteArrayInputStream, ByteArrayOutputStream }
-import java.util
 import java.util.zip.DataFormatException
 import akka.NotUsed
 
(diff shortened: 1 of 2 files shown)


johanandren:
OK TO TEST

akka-http-tests/src/test/scala/akka/http/scaladsl/coding/CoderSpec.scala
```diff
@@ -113,20 +112,22 @@ abstract class CoderSpec extends WordSpec with CodecSpecSupport with Inspectors
     }
 
     "shouldn't produce huge ByteStrings for some input" in {
-      val array = new Array[Byte](10) // FIXME
```
jrudolph: Add least I left a FIXME :roll_eyes:

akka-http-tests/src/test/scala/akka/http/scaladsl/coding/CoderSpec.scala
```diff
@@ -113,20 +112,22 @@ abstract class CoderSpec extends WordSpec with CodecSpecSupport with Inspectors
...
-      util.Arrays.fill(array, 1.toByte)
+      val array = new Array[Byte](100007)
+      val random = ThreadLocalRandom.current()
```
jrudolph: Why random data? I guess it doesn't really matter but the original test tried (eh, very hard) to emulate a zip bomb which would use well-compressible data.

ollyw: I chose random data, as I wanted to detect frames being resend, etc, so each frame needed to be unique. I didn't fancy creating an array with 100007 bytes of data manually.

jrudolph: I see, thanks for the explanation and the fix!

PR #2253, opened 2018-10-11
Add Directive#tcollect and Directive1#collect

markus1189:
This commit adds two new methods on `Directives`:

- Directive#ttransform the tupled variant of Future#transform that
  allows mapping and filtering at the same time
- Directive1#transform the equivalent of Future#transform for
  Directive1, the specialized version of Directive#ttransform

As an example, instead of writing:

```
parameter("x".as[Int]).filter(x => x >= 0, someRejection).map(x => Positive(x))
```

we can now write this in one go using Directive1#transform:

```
parameter("x".as[Int]).transform({ case x if x >= 0 => Positive(x) }, someRejection)
```

# Review Notes:

- composing `PartialFunction`s is really awkward, had to hack around it in `Directive1#transform` because there we have to nest two of them
- I would prefer a curried version of `transform` that takes the partial function as the second argument because it looks better, but I went with the two-parameter style as done in `Directive1#filter`

### DIFF
--- a/akka-http/src/main/scala/akka/http/scaladsl/server/Directive.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/server/Directive.scala
@@ -83,6 +83,16 @@ abstract class Directive[L](implicit val ev: Tuple[L]) {
     Directive[L] { inner ⇒ tapply { values ⇒ ctx ⇒ if (predicate(values)) inner(values)(ctx) else ctx.reject(rejections: _*) } }
 
   /**
+   * If the given [[scala.PartialFunction]] is defined for the input, maps this directive with the given function,
+   * which can produce either a tuple or any other value.
+   * If it is not defined however, the returned directive will reject with the given rejections.
+   */
+  def tcollect[R](pf: PartialFunction[L, R], rejections: Rejection*)(implicit tupler: Tupler[R]): Directive[tupler.Out] =
+    Directive[tupler.Out] { inner ⇒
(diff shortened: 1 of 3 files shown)


markus1189:
Changed the name to `collect` because `transform` didn't fit as well.

jlprat:
OK TO TEST

jlprat:
I guess we would also need the Java counterpart for it.

markus1189:
@johanandren I will have a look at the docs

@jlprat Where would you add that in the codebase?  For example there seems to be no `tmap` in the Java API anyway (according to my grep skills :sweat_smile: )

markus1189:
@johanandren I agree that it would be good to have something in the docs, but I couldn't find mentions of `tmap`, `tflatMap` etc., so that covering only `tcollect` would be strange.

Maybe at another point and another pr?

johanandren:
Yeah, I was under the impression that we had more docs on it in this section https://doc.akka.io/docs/akka-http/current/routing-dsl/directives/custom-directives.html but I see that we really don't so you can skip that then.

johanandren:
To quick on the comment button, now I found at least `flatMap` and `tflatMap` in that section: https://doc.akka.io/docs/akka-http/current/routing-dsl/directives/custom-directives.html#flatmap-and-tflatmap

So adding a small section after that with some text about when it is useful to do `collect` and a little sample would be nice.

markus1189:
@johanandren  I added a section about `collect` ;)

@jlprat I don't see things like `map` in the Java file you posted.  I would assume that if that is not implemented it would not make sense to implement `collect` either?

PR #2255, opened 2018-10-13
Replace existing HTTP Status Code 425 with TOO_EARLY

Philippus:
I think this should be safe to just replace as `Unordered Collection` was only ever present in WebDAV drafts and should have no implementations, but of course it is a breaking change.
See https://tools.ietf.org/html/rfc8470 for details on `Too Early`.


johanandren:
Thanks, but we'll have to keep the existing name for binary compatibility, but I think it'd be fine to add an alias with the `TooEarly` and `TOO_EARLY` name (potentially deprecating the old one, so that we can remove it in the future). 

You can read about the bin comp guarantees/requirements here: https://doc.akka.io/docs/akka/current/common/binary-compatibility-rules.html

johanandren:
Yes, that's exactly what we'd need to do, and the same for Java. Sorry for not elaborating on that in the original feedback.

PR #2270, opened 2018-10-30
Add example showing consuming a stream of requests

### LINKED ISSUE #1836: Substream Source cannot be materialized more than once when reading from s3
There's a simple repro project for this at https://github.com/easel/akka-http-examples/blob/master/src/test/scala/io/github/easel/RecursiveAkkaHttpSpec.scala

Net-net, reading a series of files such as the SEC EDGAR files in the test fails after the first file when trying to parse individual lines. The same code works fine for a single file and for a sequence of other random urls. 

Looking at the headers, the only consistent difference I'm seeing off hand is that the files that fail haven't got a content-length header.

Happy to help however but some pointers where to start digging would be really helpful.

easel:
- Fixes #1836

### DIFF
--- a/docs/src/main/paradox/implications-of-streaming-http-entity.md
+++ b/docs/src/main/paradox/implications-of-streaming-http-entity.md
@@ -45,6 +45,29 @@ Scala
 Java
 :   @@snip [HttpClientExampleDocTest.java]($test$/java/docs/http/javadsl/HttpClientExampleDocTest.java) { #manual-entity-consume-example-2 }
 
+### Integrating with Akka Streams
+In some cases, it is necessary to process the results of a series of Akka HTTP calls as Akka Streams. In order
+to ensure that the HTTP Response Entity is consumed in a timely manner, the Akka HTTP stream for each request must
+be executed and completely consumed, then sent along for further processing.
+
+Failing to account for this behavior can result in seemingly non-deterministic failures due to complex interactions
+between http and stream buffering. This manifests as errors such as the following:
+
+```
+Response entity was not subscribed after 1 second. Make sure to read the response entity body or call `discardBytes()` on it.
+```
+
+This error indicatesy that the http response has been available for too long without being consumed. It can be 
+partially worked around by increasing the subscription timeout, but you will still run the risk of running into network 
+level timeouts and could still exceed the timeout under load so it's best to resolve the issue properly such as in 
+the examples below: 
+
+Scala
+:   @@snip [HttpClientExampleSpec.scala]($test$/scala/docs/http/scaladsl/HttpClientExampleSpec.scala) { #manual-entity-consume-example-3 }
+
+Java
+:   @@snip [HttpClientExampleDocTest.java]($test$/java/docs/http/javadsl/HttpClientExampleDocTest.java) { #manual-entity-consume-example-3 }
+
 ### Discarding the HTTP Response Entity (Client)
 
 Sometimes when calling HTTP services we do not care about their response payload (e.g. all we care about is the response code),
--- a/docs/src/main/paradox/implications-of-streaming-http-entity.md
+++ b/docs/src/main/paradox/implications-of-streaming-http-entity.md
@@ -57,7 +57,7 @@ between http and stream buffering. This manifests as errors such as the followin
 Response entity was not subscribed after 1 second. Make sure to read the response entity body or call `discardBytes()` on it.
 ```
 
-This error indicatesy that the http response has been available for too long without being consumed. It can be 
+This error indicates that the http response has been available for too long without being consumed. It can be 
 partially worked around by increasing the subscription timeout, but you will still run the risk of running into network 
 level timeouts and could still exceed the timeout under load so it's best to resolve the issue properly such as in 
 the examples below: 
--- a/docs/src/test/scala/docs/http/scaladsl/HttpClientExampleSpec.scala
+++ b/docs/src/test/scala/docs/http/scaladsl/HttpClientExampleSpec.scala
@@ -71,6 +71,57 @@ class HttpClientExampleSpec extends WordSpec with Matchers with CompileOnlySpec 
     //#manual-entity-consume-example-2
   }
 
+  "manual-entity-consume-example-3" in compileOnlySpec {
+    //#manual-entity-consume-example-3
+    import scala.concurrent.duration._
+    import scala.concurrent.Future
+
+    import akka.NotUsed
+    import akka.actor.ActorSystem
+    import akka.http.scaladsl.Http
+    import akka.http.scaladsl.model._
+    import akka.stream.ActorMaterializer
+    import akka.util.ByteString
+    import akka.stream.scaladsl.{ Flow, Sink, Source }
+
+    implicit val system = ActorSystem()
+    implicit val dispatcher = system.dispatcher
+    implicit val materializer = ActorMaterializer()
+
+    case class ExamplePerson(name: String)
+
+    def parse(line: ByteString): Option[ExamplePerson] =
+      line.utf8String.split(" ").headOption.map(ExamplePerson)
+
+    val requests: Source[HttpRequest, NotUsed] = Source
+      .fromIterator(() =>
+        Range(0, 10).map(i => HttpRequest(uri = Uri(s"https://localhost/people/$i"))).iterator
+      )
+
+    val processorFlow: Flow[Option[ExamplePerson], Int, NotUsed] =
+      Flow[Option[ExamplePerson]].map(_.map(_.name.length).getOrElse(0))
+
+    // Run and completely consume a single akka http request
+    def runRequest(req: HttpRequest): Future[Option[ExamplePerson]] =
+      Http()
+        .singleRequest(req)
+        .flatMap { response =>
+          response.entity.dataBytes
+            .runReduce(_ ++ _)
+            .map(parse)
+        }
+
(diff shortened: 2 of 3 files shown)


docs/src/test/scala/docs/http/scaladsl/HttpClientExampleSpec.scala
```diff
@@ -71,6 +71,50 @@ class HttpClientExampleSpec extends WordSpec with Matchers with CompileOnlySpec
...
+        }
+
+    // Run each akka http flow to completion, then continue processing
+    requests
+      .mapAsync(2)(runRequest)
+      .via(processorFlow)
+      .runWith(Sink.ignore)
+
+    //#manual-entity-consume-example-3
+  }
```
johanandren: I think example impls would be nice here as well, rather than the `???`s

docs/src/test/scala/docs/http/scaladsl/HttpClientExampleSpec.scala
```diff
@@ -71,6 +71,50 @@ class HttpClientExampleSpec extends WordSpec with Matchers with CompileOnlySpec
...
+        .flatMap(_.entity.toStrict(3.seconds))
+        .flatMap { e =>
+          e.dataBytes
+            .runFold(ByteString.empty) { case (acc, b) => acc ++ b }
+            .map(parse)
+        }
+
+    // Run each akka http flow to completion, then continue processing
+    requests
+      .mapAsync(2)(runRequest)
```
glammers1: why 2?

docs/src/test/scala/docs/http/scaladsl/HttpClientExampleSpec.scala
```diff
@@ -8,7 +8,7 @@ import akka.http.scaladsl.model.HttpRequest
 import akka.http.scaladsl.settings.ClientConnectionSettings
 import akka.http.scaladsl.settings.ConnectionPoolSettings
 import docs.CompileOnlySpec
-import org.scalatest.{ Matchers, WordSpec }
+import org.scalatest.{Matchers, WordSpec}
```
jlprat: This is wrongly formatted and broke the build. This should fix it
```suggestion
import org.scalatest.{ Matchers, WordSpec }
```

docs/src/test/scala/docs/http/scaladsl/HttpClientExampleSpec.scala
```diff
@@ -71,6 +71,50 @@ class HttpClientExampleSpec extends WordSpec with Matchers with CompileOnlySpec
...
+    import scala.concurrent.duration._
+    import scala.concurrent.Future
+
+    import akka.NotUsed
+    import akka.actor.ActorSystem
+    import akka.http.scaladsl.Http
+    import akka.http.scaladsl.model._
+    import akka.stream.ActorMaterializer
+    import akka.util.ByteString
+    import akka.stream.scaladsl.{Flow, Sink, Source}
```
jlprat: This is wrongly formatted and broke the build. This should fix it
```suggestion
    import akka.stream.scaladsl.{ Flow, Sink, Source }
```

docs/src/test/scala/docs/http/scaladsl/HttpClientExampleSpec.scala
glammers1: What do you think about having the same value for the Java example instead of 1?

docs/src/test/scala/docs/http/scaladsl/HttpClientExampleSpec.scala
jlprat: Build is failing because of this line. Please note the missing spaces after `{` and before `}`

docs/src/test/scala/docs/http/scaladsl/HttpClientExampleSpec.scala
easel: Done =p

jlprat:
There is a compilation error:
```
[error] /home/jenkins/workspace/workspace/pr-validator-akka-http/docs/src/test/scala/docs/http/scaladsl/HttpClientExampleSpec.scala:94:76: type mismatch;
[error]  found   : String("https://akka.org")
[error]  required: akka.http.scaladsl.model.HttpMethod
[error]     val requests: Source[HttpRequest, NotUsed] = Source.single(HttpRequest("https://akka.org"))
[error]                                                                            ^
```

docs/src/main/paradox/implications-of-streaming-http-entity.md
```diff
@@ -45,6 +45,17 @@ Scala
 Java
 :   @@snip [HttpClientExampleDocTest.java]($test$/java/docs/http/javadsl/HttpClientExampleDocTest.java) { #manual-entity-consume-example-2 }
 
+### Integrating with Akka Streams
+In some cases, it is necessary to process the results of a series of akka http calls as akka streams. In order
```
raboof:
```suggestion
In some cases, it is necessary to process the results of a series of Akka HTTP calls as Akka Streams. In order
```

docs/src/main/paradox/implications-of-streaming-http-entity.md
```diff
@@ -45,6 +45,17 @@ Scala
...
+to ensure that the HTTP Response Entity is consumed in a timely manner, the akka http stream for each request must
```
raboof:
```suggestion
to ensure that the HTTP Response Entity is consumed in a timely manner, the Akka HTTP stream for each request must
```

docs/src/test/scala/docs/http/scaladsl/HttpClientExampleSpec.scala
```diff
@@ -71,6 +71,50 @@ class HttpClientExampleSpec extends WordSpec with Matchers with CompileOnlySpec
...
+    val processorFlow: Flow[ExamplePerson, Long, NotUsed] = ???
+
+    // Run and completely consume a single akka http request
+    def runRequest(req: HttpRequest): Future[ExamplePerson] =
+      Http()
+        .singleRequest(req)
+        .flatMap(_.entity.toStrict(3.seconds))
+        .flatMap { e =>
+          e.dataBytes
+            .runFold(ByteString.empty) { case (acc, b) => acc ++ b }
```
raboof: Could this be more easily written as a `reduce`?

docs/src/test/scala/docs/http/scaladsl/HttpClientExampleSpec.scala
```diff
@@ -71,6 +71,50 @@ class HttpClientExampleSpec extends WordSpec with Matchers with CompileOnlySpec
...
+
+    val requests: Source[HttpRequest, NotUsed] = Source.single(HttpRequest("https://akka.org"))
+
+    val processorFlow: Flow[ExamplePerson, Long, NotUsed] = ???
+
+    // Run and completely consume a single akka http request
+    def runRequest(req: HttpRequest): Future[ExamplePerson] =
+      Http()
+        .singleRequest(req)
+        .flatMap(_.entity.toStrict(3.seconds))
```
raboof: Do we need the `toStrict` here, since we immediately consume the stream below?

docs/src/test/scala/docs/http/scaladsl/HttpClientExampleSpec.scala
raboof: Perhaps mention in a comment that this is a memory vs parallelism trade-off, since the more parallelism you allow, the more `ExamplePerson`s you will hold in memory?

easel:
Rebased and implemented comments per review, I think I caught everything. Let's see if CI agrees. Thanks @raboof @johanandren @glammers1 @jlprat

easel:
Added some more text to the docs indicating what can happen if you don't properly consume the stream, per discussion in #1836. This should help search engines pick up on the docs when people search for `Response entity was not subscribed after 1 second`.

docs/src/test/scala/docs/http/scaladsl/HttpClientExampleSpec.scala
```diff
@@ -71,6 +71,57 @@ class HttpClientExampleSpec extends WordSpec with Matchers with CompileOnlySpec
...
+    def runRequest(req: HttpRequest): Future[Option[ExamplePerson]] =
+      Http()
+        .singleRequest(req)
+        .flatMap { response =>
+          response.entity.dataBytes
+            .runReduce(_ ++ _)
+            .map(parse)
+        }
+
+    // Run each akka http flow to completion, then continue processing. You'll want to tune the `parralelism`
```
jrudolph:
```suggestion
    // Run each akka http flow to completion, then continue processing. You'll want to tune the `parallelism`
```

docs/src/main/paradox/implications-of-streaming-http-entity.md
```diff
@@ -45,6 +45,29 @@ Scala
...
+be executed and completely consumed, then sent along for further processing.
+
+Failing to account for this behavior can result in seemingly non-deterministic failures due to complex interactions
+between http and stream buffering. This manifests as errors such as the following:
+
+```
+Response entity was not subscribed after 1 second. Make sure to read the response entity body or call `discardBytes()` on it.
+```
+
+This error indicatesy that the http response has been available for too long without being consumed. It can be
```
jrudolph:
```suggestion
This error indicates that the http response has been available for too long without being consumed. It can be 
```

easel:
Thanks for the feedback and merge folks!

PR #2274, opened 2018-11-08
Use collision-resistent maps for formFieldMaps

raboof:


akka-http/src/main/scala/akka/http/scaladsl/server/directives/FormFieldDirectives.scala
```diff
@@ -104,11 +104,13 @@ object FormFieldDirectives extends FormFieldDirectives {
 
     _formFieldSeq.map {
       case seq ⇒
-        append(Map.empty, seq)
+        append(immutable.TreeMap.empty, seq)
```
jrudolph: Yep, I guess using TreeMap is fine for form fields. Having hundreds of those wouldn't be best practice anyway.

PR #2275, opened 2018-11-08
=pro update to spray-json 1.3.5

jrudolph:


jlprat:
I was looking at the occurrences of `order` in the docs and I couldn't find any that would refer to the JSON field ordering. This means, we didn't promise order, but we also mentioned that order is not to be trusted.
Maybe is worth stating it in the docs (maybe just on the release notes).

PR #2276, opened 2018-11-08
Collision-resistent maps in header parsing

raboof:


raboof:
Added parameters to the `Content-Type` in HeaderParserBenchmark, seems OK:
```
Master:

Cnt     Score     Error   Units
[info] HeaderParserBenchmark.bench_parse_headers                      no  thrpt    3  3598.170 ± 375.241  ops/ms
[info] HeaderParserBenchmark.bench_parse_headers                     yes  thrpt    3  3480.690 ± 209.644  ops/ms

Cnt     Score     Error   Units
[info] HeaderParserBenchmark.bench_parse_headers                      no  thrpt    3  3529.619 ± 729.152  ops/ms
[info] HeaderParserBenchmark.bench_parse_headers                     yes  thrpt    3  3565.951 ± 158.762  ops/ms


Branch:

Cnt     Score     Error   Units
[info] HeaderParserBenchmark.bench_parse_headers                      no  thrpt    3  3573.066 ±  64.581  ops/ms
[info] HeaderParserBenchmark.bench_parse_headers                     yes  thrpt    3  3643.082 ± 142.643  ops/ms

Cnt     Score      Error   Units
[info] HeaderParserBenchmark.bench_parse_headers                      no  thrpt    3  3502.156 ± 1586.796  ops/ms
[info] HeaderParserBenchmark.bench_parse_headers                     yes  thrpt    3  3608.040 ±  167.558  ops/ms
```

akka-http-core/src/test/scala/akka/http/impl/model/parser/HttpHeaderSpec.scala
```diff
@@ -141,12 +141,12 @@ class HttpHeaderSpec extends FreeSpec with Matchers {
       "Authorization: Fancy QWxhZGRpbjpvcGVuIHNlc2FtZQ==" =!=
         Authorization(GenericHttpCredentials("Fancy", "QWxhZGRpbjpvcGVuIHNlc2FtZQ=="))
       """Authorization: Fancy yes="n:o", nonce=42""" =!=
-        Authorization(GenericHttpCredentials("Fancy", Map("yes" → "n:o", "nonce" → "42"))).renderedTo(
-          """Fancy yes="n:o",nonce=42""")
-      """Authorization: Fancy yes=no,nonce="4\\2"""" =!=
-        Authorization(GenericHttpCredentials("Fancy", Map("yes" → "no", "nonce" → """4\2""")))
-      """Authorization: Other yes=no,empty=""""" =!=
-        Authorization(GenericHttpCredentials("Other", Map("yes" → "no", "empty" → "")))
+        Authorization(GenericHttpCredentials("Fancy", Map("nonce" → "42", "yes" → "n:o"))).renderedTo(
```
jlprat: I understand this change is because now using `TreeMap` the contents are sorted by key, right?

raboof: Correct. I don't think we ever promised to keep the order of the keys (and still don't), so in that sense this test is too specific.

akka-http-core/src/main/scala/akka/http/impl/model/parser/LinkHeader.scala
```diff
@@ -62,7 +64,7 @@ private[parser] trait LinkHeader { this: Parser with CommonRules with CommonActi
     }
   }
 
-  def `link-media-type` = rule { `media-type` ~> ((mt, st, pm) ⇒ getMediaType(mt, st, pm contains "charset", pm.toMap)) }
+  def `link-media-type` = rule { `media-type` ~> ((mt, st, pm) ⇒ getMediaType(mt, st, pm contains "charset", TreeMap(pm: _*))) }
```
jrudolph: This constructor is probably not super fast but it probably doesn't matter for that particular header.

Issue #17, opened 2016-09-08
Unmarshal() used with http entities is unsafe if it fails (it should toStrict)


ktoso:
<a href="https://github.com/ktoso"><img src="https://avatars.githubusercontent.com/u/120979?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [ktoso](https://github.com/ktoso)**
_Tuesday Jun 28, 2016 at 14:27 GMT_
_Originally opened as https://github.com/akka/akka/issues/20848_

----

Unmarshal must force a poper full draining of the http response body otherwise we might end up stalling connections. Reported by some users via https://github.com/akka/akka/issues/20192

I'm not 100% certain we do the right thing currently, so this ticket is about making sure that if an unmarshaller fails we do actually consume the entire incoming stream.

ktoso:
<a href="https://github.com/ktoso"><img src="https://avatars.githubusercontent.com/u/120979?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [ktoso](https://github.com/ktoso)**
_Tuesday Jun 28, 2016 at 14:28 GMT_

----

Current workaround is pretty much:

```scala
Unmarshal(x).to[Example].recoverWith {
  case ex =>
    x.entity.dataBytes.runWith(Sink.ignore).flatMap(_ => Future.failed(ex))
}
```

raboof:
The example above appears to be using spray-json, which should be safe because it [relies on Unmarshaller.byteStringUnmarshaller](https://github.com/akka/akka-http/blob/master/akka-http-marshallers-scala/akka-http-spray-json/src/main/scala/akka/http/scaladsl/marshallers/sprayjson/SprayJsonSupport.scala#L32) which [consumes the entire entity into memory before parsing](https://github.com/akka/akka-http/blob/master/akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/PredefinedFromEntityUnmarshallers.scala#L16).

Still, might be good to have an approach that would make sure this works for any `Unmarshaller`.

jrudolph:
#2281 added some extra note to the documenation but I think the real issue might lie in our own `Unmarshal` implementation that will not touch entities if the Accept headers don't match.

The question is whether this is basically by design of the `Unmarshal` API that it doesn't have side-effects if it fails (so that some other code could potential take over).

That would mean that we should probably provide another API for client purposes explicitly to "unmarshal or discard entity if unmarshalling isn't possible`.

PR #2281, opened 2018-11-20
Clarify unmarshallers should consume or cancel the entity body

raboof:
With this note I think we can demote #17 from 'bug' to 'enhancement'


raboof:
I agree it's a bit contrived, but I guess you could have something that reads a header from the stream, and then when there is an error in the header, returns a failed future without reading the rest of the stream.

PR #2282, opened 2018-11-20
Clarify client-side pipelining is not supported

raboof:
As discussed in #32

### DIFF
--- a/akka-http-core/src/main/resources/reference.conf
+++ b/akka-http-core/src/main/resources/reference.conf
@@ -347,22 +347,7 @@ akka.http {
     # Must be a power of 2 and > 0!
     max-open-requests = 32
 
-    # The maximum number of requests that are dispatched to the target host in
-    # batch-mode across a single connection (HTTP pipelining).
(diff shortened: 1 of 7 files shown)


akka-http-core/src/main/resources/reference.conf
```diff
@@ -347,22 +347,7 @@ akka.http {
...
-    # on a connection that still has a non-idempotent request in flight.
-    #
-    # Before increasing this value, make sure you understand the effects of head-of-line blocking.
-    # Using a connection pool, a request may be issued on a connection where a previous
-    # long-running request hasn't finished yet. The response to the pipelined requests may then be stuck
-    # behind the response of the long-running previous requests on the server. This may introduce an
-    # unwanted "coupling" of run time between otherwise unrelated requests.
-    #
-    # See http://tools.ietf.org/html/rfc7230#section-6.3.2 for more info.)
+    # Client-side pipelining is not currently supported. See https://github.com/akka/akka-http/issues/32
```
johanandren: Should we also log a warning if someone _does_ set the value to something else than 1?

PR #2283, opened 2018-11-20
toStrict the entity when parsing multiple form fields

raboof:
Fixes the 'simple case' ('A') of #73

To avoid consuming the entity for each field. Of course slurping in the
entire entity is not very elegant, but this is what both the form-data and
the multipart marshallers already do further down the line.

### DIFF
--- a/akka-http/src/main/scala/akka/http/scaladsl/common/StrictForm.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/common/StrictForm.scala
@@ -40,6 +40,9 @@ sealed abstract class StrictForm {
 }
 
 object StrictForm {
+  // TODO: make timeout configurable
+  val toStrictTimeout = 10.seconds
+
   sealed trait Field
   object Field {
     private[http] def fromString(value: String): Field = FromString(value)
@@ -115,7 +118,7 @@ object StrictForm {
         def tryUnmarshalToMultipartForm: Future[StrictForm] =
           for {
             multiPartFD ← multipartUM(entity).fast
-            strictMultiPartFD ← multiPartFD.toStrict(10.seconds).fast // TODO: make timeout configurable
+            strictMultiPartFD ← multiPartFD.toStrict(toStrictTimeout).fast
           } yield {
             new StrictForm {
               val fields = strictMultiPartFD.strictParts.map {
--- a/akka-http/src/main/scala/akka/http/scaladsl/server/directives/FormFieldDirectives.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/server/directives/FormFieldDirectives.scala
@@ -217,7 +217,7 @@ object FormFieldDirectives extends FormFieldDirectives {
     import akka.http.scaladsl.server.util.TupleOps._
 
     implicit def forTuple[T](implicit fold: FoldLeft[Directive0, T, ConvertFieldDefAndConcatenate.type]): FieldDefAux[T, fold.Out] =
-      fieldDef[T, fold.Out](fold(pass, _))
+      fieldDef[T, fold.Out](fold(toStrictEntity(StrictForm.toStrictTimeout), _))
 
     object ConvertFieldDefAndConcatenate extends BinaryPolyFunc {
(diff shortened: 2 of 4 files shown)


akka-http/src/main/scala/akka/http/scaladsl/common/StrictForm.scala
```diff
@@ -115,7 +118,7 @@ object StrictForm {
         def tryUnmarshalToMultipartForm: Future[StrictForm] =
           for {
             multiPartFD ← multipartUM(entity).fast
-            strictMultiPartFD ← multiPartFD.toStrict(10.seconds).fast // TODO: make timeout configurable
+            strictMultiPartFD ← multiPartFD.toStrict(toStrictTimeout).fast
```
jrudolph: Good :+1:

akka-http/src/main/scala/akka/http/scaladsl/common/StrictForm.scala
```diff
@@ -40,6 +40,9 @@ sealed abstract class StrictForm {
 }
 
 object StrictForm {
+  // TODO: make timeout configurable
+  val toStrictTimeout = 10.seconds
```
jrudolph: Not so good :-1: ;)

jrudolph: (not sure if there's an easy better solution, though)

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FormFieldDirectives.scala
```diff
@@ -217,7 +217,7 @@ object FormFieldDirectives extends FormFieldDirectives {
     import akka.http.scaladsl.server.util.TupleOps._
 
     implicit def forTuple[T](implicit fold: FoldLeft[Directive0, T, ConvertFieldDefAndConcatenate.type]): FieldDefAux[T, fold.Out] =
-      fieldDef[T, fold.Out](fold(pass, _))
+      fieldDef[T, fold.Out](fold(toStrictEntity(StrictForm.toStrictTimeout), _))
```
jrudolph: Why exactly here? I guess this helps for the `formField('a, 'b)` case. If we would put it in the directive directly, wouldn't it then also work for the nested formField directive tasks?

jrudolph:
Great that you are tackling this issue, @raboof. It's a long-standing nuisance.

> To avoid consuming the entity for each field. Of course slurping in the
> entire entity is not very elegant

There isn't any way around it for this style of API which assumes random access to form fields.

For streaming formField access we would need a completely different API.

akka-http/src/main/scala/akka/http/scaladsl/common/StrictForm.scala
raboof: Since we're dealing with directives it seems like it should be possible to read it from the config... something for another PR though?

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FormFieldDirectives.scala
raboof: Yeah, that would indeed also solve scenario `B`, at the expense of also inserting a `toStrict` in situations where it's not strictly necessary. Might be fine.

raboof:
This PR solves the `A` case, by applying `toStrict` in more places we could probably also fix the `B` case at the expense of doing too many `toStrict` calls in some cases. It's not entirely obvious to me how to fit this in with the magnets and recursive types, though.

I'd propose including this PR first and leaving the `B` case for later, given that this probably already helps for common cases.

jrudolph:
> doing too many `toStrict` calls in some cases.

Do you mean, because currently each single form field already tries to read the entity fully, we would `toStrict` multiple times even for the `formField(`singleField`) case? Or which case do you mean?

raboof:
> > doing too many `toStrict` calls in some cases.
> 
> Do you mean, because currently each single form field already tries to read the entity fully, we would `toStrict` multiple times even for the `formField(`singleField`) case? Or which case do you mean?

I meant if we'd insert a `toStrict` in front of any `formField`/`formFields` directive, it would also `toStrict` the entity even when there is no other nested `formField`/`formFields` directive in the hierarchy. That might not be a big deal, but it was also not obvious how to do it :smile:.

Issue #2287, opened 2018-11-21
HttpEntity.toStrict - JavaApi - is ignoring maxBytes


miroslavhornak:
akka-http version: 10.1.5

HttpEntity.scala:

Current (lines 185-187):
 ```
/** Java API */
 override def toStrict(timeoutMillis: Long, maxBytes: Long, materializer: Materializer): CompletionStage[jm.HttpEntity.Strict] =
    toStrict(timeoutMillis.millis)(materializer).toJava
```
Shall be (calling toStrict with maxBytes):
```
/** Java API */
  override def toStrict(timeoutMillis: Long, maxBytes: Long, materializer: Materializer): CompletionStage[jm.HttpEntity.Strict] =
    toStrict(timeoutMillis.millis, maxBytes)(materializer).toJava
```

to avoid unnecessary parsing of `config.getPossiblyInfiniteBytes("akka.http.parsing.max-to-strict-bytes")` with every single HttpResponse (on which toStrict() is called) ... 
Btw. parsing of that value (most probably default one in the config) (not specified in our own application.conf) throws a lot of exceptions that can be seen in any java profiler ... 

![screenshot 2018-11-21 at 16 33 47](https://user-images.githubusercontent.com/2237381/48851512-8c124c00-edab-11e8-8019-dd2887fa1088.png)

miroslavhornak:
Just a side note: exceptions reported by the java profiller come from the Typesafe Config project itself ... see: https://github.com/lightbend/config/blob/master/config/src/main/java/com/typesafe/config/impl/DefaultTransformer.java

jlprat:
Done by https://github.com/akka/akka-http/commit/e92456c9da594145b7b023832a8b49eb1e5d1fb0

PR #2290, opened 2018-11-22
Java API - call scala HttpEntity.toStrict that respects maxBytes #2287

miroslavhornak:


raboof:
refs #2287

raboof:
OK TO TEST

PR #2298, opened 2018-11-29
Support for scala 2.13.0-M5

raboof:
Updated just akka-http-core on 2.12/2.13 for now.

Unfortunately we have a number of places where we have both a `varargs` and a `immutable.Seq` overload for the same method. This is no longer valid on 2.13.0, and the only way I could think of to solve this in a binary compatible way was a rather weird trick: introduce a `VASeq` type which corresponds to `immutable.Seq` on 2.12 and `immutable.Iterable` on 2.13 (to avoid the ambiguity but at the same time be mostly source-compatible and keep intact the fact that this value is immutable).


akka-http-core/src/main/scala-2.13.0-M5/akka/http/ccompat/package.scala
```diff
@@ -0,0 +1,30 @@
+package akka.http
+
+import akka.http.impl.util.JavaMapping
+
+package object ccompat {
+
+  /**
+   * On scala 2.12, we often provide the same method both in 'immutable.Seq' and 'varargs' variation.
+   *
+   * On scala 2.13, this is no longer necessary, since varargs are treated as immutable. Therefore we need
```
jrudolph: necessary or possible?

jrudolph:
> Unfortunately we have a number of places where we have both a `varargs` and a `immutable.Seq` overload for the same method. This is no longer valid on 2.13.0

Is that documented somewhere? Would be nice if no overload was necessary at all and the compiler would just accept a Seq in the first place even without the `: _*` type ascription.

We could also say that we don't keep source compatibility and remove the `(Seq)` overloads in 2.13 (and deprecate them in 2.12).

akka-http-core/src/main/java/akka/http/impl/util/Util.java
```diff
@@ -35,7 +36,8 @@
         return (Source<U, scala.Unit>)(Object) p;
     }
     public static scala.collection.immutable.Map<String, String> convertMapToScala(Map<String, String> map) {
-        return emptyMap.$plus$plus(scala.collection.JavaConverters.mapAsScalaMapConverter(map).asScala());
+        return MapHelpers.convertMapToScala(map);
+        //emptyMap.<String>concat(scala.collection.JavaConverters.mapAsScalaMapConverter(map).asScala());
```
johanandren: Is JavaConverters removed in 2.13 or why is this needed?

akka-http-core/src/main/scala-2.12/akka/http/scaladsl/util/FastFuture.scala
```diff
@@ -0,0 +1,122 @@
...
+  def reduce[T, R >: T](futures: TraversableOnce[Future[T]])(op: (R, T) ⇒ R)(implicit executor: ExecutionContext): Future[R] =
+    if (futures.isEmpty) failed(new NoSuchElementException("reduce attempted on empty collection"))
+    else sequence(futures).fast.map(_ reduceLeft op)
+
+  def traverse[A, B, M[_] <: TraversableOnce[_]](in: M[A])(fn: A ⇒ Future[B])(implicit cbf: CanBuildFrom[M[A], B, M[B]], executor: ExecutionContext): Future[M[B]] =
+    in.foldLeft(successful(cbf(in))) { (fr, a) ⇒
+      val fb = fn(a.asInstanceOf[A])
+      for (r ← fr.fast; b ← fb.fast) yield r += b
+    }.fast.map(_.result())
+}
```
johanandren: IntelliJ claims this is identical to the other FastFuture? Oh wait, there is a third one...

project/Dependencies.scala
```diff
@@ -53,6 +53,8 @@ object Dependencies {
     val caffeine    = "com.github.ben-manes.caffeine" % "caffeine"                     % "2.6.2"
     val jsr305      = "com.google.code.findbugs"      % "jsr305"                       % "3.0.2"             % Provided // ApacheV2
 
+    val collectionCompat = "org.scala-lang.modules" %% "scala-collection-compat" % "0.2.1"
```
johanandren: But we wont be able to use it in the end, right?

akka-http/src/main/scala/akka/http/scaladsl/coding/Encoder.scala
```diff
@@ -38,7 +38,7 @@ trait Encoder {
     def encodeChunk(bytes: ByteString): ByteString = compressor.compressAndFlush(bytes)
     def finish(): ByteString = compressor.finish()
 
-    StreamUtils.byteStringTransformer(encodeChunk, finish)
+    StreamUtils.byteStringTransformer(encodeChunk, () ⇒ finish)
```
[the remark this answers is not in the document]
jrudolph: It might not be necessary in the next milestone of 2.13 :)

https://github.com/scala/scala/pull/7660

akka-http-core/src/test/scala/akka/http/impl/engine/server/HttpServerSpec.scala
```diff
@@ -1061,7 +1061,7 @@ class HttpServerSpec extends AkkaSpec(
       "have a programmatically set timeout handler" in assertAllStagesStopped(new RequestTimeoutTestSetup(400.millis) {
         send("GET / HTTP/1.1\r\nHost: example.com\r\n\r\n")
         val timeoutResponse = HttpResponse(StatusCodes.InternalServerError, entity = "OOPS!")
-        expectRequest().header[`Timeout-Access`].foreach(_.timeoutAccess.updateHandler(_ ⇒ timeoutResponse))
+        expectRequest().header[`Timeout-Access`].foreach(_.timeoutAccess.updateHandler((_: HttpRequest) ⇒ timeoutResponse))
```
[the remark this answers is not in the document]
raboof: I suspect this is needed because there are 2 `updateHandler` methods on `akka.http.scaladsl.TimeoutAccess`: one from the javadsl `void updateHandler(Function<HttpRequest, HttpResponse> handler);` and one from the scaladsl `def updateHandler(handler: HttpRequest ⇒ HttpResponse): Unit`. I suspect because the scaladsl HttpResponse also extends the javadsl one, this is now ambiguous?

akka-http-tests/src/test/scala/akka/http/scaladsl/marshalling/MarshallingSpec.scala
```diff
@@ -202,7 +202,10 @@ class MarshallingSpec extends FreeSpec with Matchers with BeforeAndAfterAll with
   override def afterAll() = TestKit.shutdownActorSystem(system)
 
   protected class FixedRandom extends java.util.Random {
-    override def nextBytes(array: Array[Byte]): Unit = "my-stable-boundary".getBytes("UTF-8").copyToArray(array)
+    override def nextBytes(array: Array[Byte]): Unit = {
```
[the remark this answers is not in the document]
raboof: It seems for some reason the implicit conversion `copyArrayToImmutableIndexedSeq` to `IndexesSeq[Byte]` that introduces the `IterableOnceOps` that contain `copyToArray(array)` does not trigger automatically for some reason - perhaps due to ambiguous implicits somewhere?

`copyToArray(array, start, length)` doesn't have that problem since it's defined on `ArrayOps`, which has the additional benefit that it doesn't copy the array twice (`copyArrayToImmutableIndexedSeq` has been deprecated because it copies the array, but that might not always be necessary, like here).

akka-http-core/src/main/java/akka/http/impl/util/Util.java
raboof: JavaConverters is fine, but converting to an immutable map was tricky.

This used to be done by creating an immutable map and calling `++`, but on 2.13 `++` is no longer available and `concat` is there instead - which doesn't work on 2.12. After some fiddling without a satisifying result I figured it'd be nicer to just have separate implementations and moved them to `MapHelpers`.

akka-http-caching/src/main/mima-filters/10.1.7.backwards.excludes
```diff
@@ -0,0 +1,2 @@
+# Scala 2.13.0-M5
+ProblemFilters.exclude[MissingTypesProblem]("akka.http.caching.scaladsl.CachingSettings$")
```
jrudolph: What happened here?

akka-http-tests/src/test/scala/akka/http/scaladsl/marshalling/MarshallingSpec.scala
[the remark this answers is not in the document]
jrudolph: Thanks for the explanation.

akka-http-core/src/test/scala/akka/http/impl/engine/server/HttpServerSpec.scala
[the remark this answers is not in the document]
jrudolph: Hmm, I guess that's good enough. We don't have to support source compatibility in all fringe cases.

project/MiMa.scala
```diff
@@ -27,10 +27,11 @@ object MiMa extends AutoPlugin {
   val currentFork = "10.1."
 
   override val projectSettings = Seq(
-    mimaPreviousArtifacts :=
+    mimaPreviousArtifacts := {
```
jrudolph: Looks good. Should we create a ticket to verify these settings once we release the second version for 2.13?

We could also just start to add versions here just to see if it already catches something?

akka-http-caching/src/main/mima-filters/10.1.7.backwards.excludes
jrudolph: I think that settings class doesn't seem to follow the usual pattern anyway. Shouldn't it derive from the public `akka.http.scaladsl.settings.SettingsCompanion`? Not really related to this one here. So, let's create a ticket for that and fix that afterwards.

akka-parsing/src/main/scala/akka/http/ccompat/since213macro.scala
```diff
@@ -0,0 +1,21 @@
...
+
+import scala.annotation.StaticAnnotation
+import scala.language.experimental.macros
+import scala.reflect.macros.blackbox.Context
+
+object since213macro {
+  def impl(c: Context)(annottees: c.Expr[Any]*): c.Expr[Any] = annottees match {
+    case Seq(method) ⇒
+      import c.universe._
+      if (scala.util.Properties.versionNumberString.startsWith("2.13"))
```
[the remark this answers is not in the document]
raboof: Created #2408 and added it there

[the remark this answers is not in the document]
jrudolph: Cool, thanks!

project/MiMa.scala
raboof: We had 10.1.7 in the list earlier, but I moved it out. The reason for that was that our new approach with the varargs was not binary compatible with our previous approach for the 2.13 artifacts. Adding exclusions for those would be possible, but I was afraid they would obscure possible 'real' problems on 2.11 and 2.12.

akka-http-core/src/main/java/akka/http/javadsl/model/Multiparts.java
```diff
@@ -5,16 +5,19 @@
 package akka.http.javadsl.model;
 
 import akka.stream.javadsl.Source;
+import scala.collection.immutable.HashMap;
```
jrudolph: Are these two added imports related to the change/necessary?

project/MiMa.scala
jrudolph: Yep, agreed.

Issue #2132, opened 2018-07-31
Incorrect behaviour for header parsing with akka-http-testkit


vhiairrassary:
It seems that header parsing with `akka-http-testkit` is behaving differently than a real akka HTTP server. I am not sure it is a bug but the behaviour is unexpected. Here is my failing test:
```scala
import akka.http.scaladsl.model.headers._
import akka.http.scaladsl.server.Directives._
import akka.http.scaladsl.testkit.ScalatestRouteTest
import org.scalatest._

class Test extends FlatSpec with Matchers with ScalatestRouteTest {
  val route =
    extractCredentials { creds =>
      complete {
        creds match {
          case Some(c) => "Found: " + c
          case _ => "No credentials"
        }
      }
    }

  it should "verify headers" in {
    RawHeader("Authorization", """Basic dXNlcjpwd2Q=""").toString() shouldBe "Authorization: Basic dXNlcjpwd2Q="

    Authorization(BasicHttpCredentials("user", "pwd")).toString() shouldBe "Authorization: Basic dXNlcjpwd2Q="
  }

  it should "test 1" in {
    Get("/") ~> addHeader(RawHeader("Authorization", """Basic dXNlcjpwd2Q=""")) ~>
      route ~> check {
      responseAs[String] shouldEqual """Found: Basic dXNlcjpwd2Q="""
    }
  }

  it should "test 2" in {
    Get("/") ~> addHeader(Authorization(BasicHttpCredentials("user", "pwd"))) ~>
      route ~> check {
      responseAs[String] shouldEqual """Found: Basic dXNlcjpwd2Q="""
    }
  }
}
```

From my understanding, the test `test 1` should not fail. If we launch a real server and perform `curl -H 'Authorization: Basic dXNlcjpwd2Q=' localhost:8080` then we get the expected result.

vhiairrassary:
@Prithvirajbilla Thank you for your help, this help me as I have now a workaround.
Should this be still considered as a bug or not?

Prithvirajbilla:
I would consider this as a bug. 

Strangely this test case is working with custom headers
```
import akka.http.scaladsl.model.headers._
import akka.http.scaladsl.server.Directives._
import akka.http.scaladsl.testkit.ScalatestRouteTest
import org.scalatest._

import scala.util.Try

class Test extends FlatSpec with Matchers with ScalatestRouteTest {
  final class ApiTokenHeader(val token: String) extends ModeledCustomHeader[ApiTokenHeader] {
[...]

PR #2300, opened 2018-12-02
#2132 Add method to parse raw header into one of model header

ashishkujoy:
#2132  Provide an overloaded addHeader method which may parse a given raw header to Model header like Authorization header


jlprat:
OK TO TEST

raboof:
Hmm, so `addHeader("Authorization", """Basic dXNlcjpwd2Q=""")` already works (and parses the string into a modeled header) and it's only when explicitly providing a `RawHeader` that the header is not parsed but passed along in `Raw` form?

I'm starting to think perhaps we should just make the documentation on `addHeader(header: HttpHeader)` more clear, and point people to `addHeader(headerName: String, headerValue: String)` when they want to provide the test headers as strings rather than 'modeled' (or explicitly raw) headers?

PR #2304, opened 2018-12-04
Overriden `getForKeyCaseInsensitive` in `HttpMethods` #2303

### LINKED ISSUE #2303: `getForKeyCaseInsensitive` not working on `HttpMethods`.
`HttpMethods.getForKeyCaseInsensitive("POST")` gives `None`, should be `Some(HttpMethod(GET))`

The reason is that when `HttpMethods` registers its members it uses upper case keys
https://github.com/akka/akka-http/blob/240d9ac8b2cb4a63a4dd2256b5705486d616f866/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMethod.scala#L57

... but when invoking `getForKeyCaseInsensitive` under the hood it applies `toRootLowerCase` on key
https://github.com/akka/akka-http/blob/240d9ac8b2cb4a63a4dd2256b5705486d616f866/akka-http-core/src/main/scala/akka/http/impl/util/ObjectRegistry.scala#L27

My recommendations are either store keys in lower case or override `getForKeyCaseInsensitive` and replace `toRootLowerCase` with `toUpperCase(Locale.ROOT)` 

```scala
override def getForKeyCaseInsensitive(key: String)(implicit conv: String <:< String) : Option[HttpMethod] =
    getForKey(key.toUpperCase(Locale.ROOT))
```

liljaylj:
Fixes #2303

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMethod.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMethod.scala
@@ -4,6 +4,8 @@
 
 package akka.http.scaladsl.model
 
+import java.util.Locale
+
 import akka.http.impl.util._
 import akka.http.javadsl.{ model ⇒ jm }
 import akka.http.scaladsl.model.RequestEntityAcceptance._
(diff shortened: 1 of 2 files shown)


lightbend-cla-validator:
Hi @liljaylj,

jlprat:
OK TO TEST

jlprat:
Thanks @liljaylj for you contribution!
I agree with @raboof I think adding a test in here would be helpful :)

Issue #1484, opened 2017-10-17
Error in stage [HttpRequestParser]


hamsterready:
Hi, I am using akka-http/10.0.9 and on production I see following exceptions (not very often, probably every 500k requests):

```
Error in stage [HttpRequestParser]: 762 
java.lang.ArrayIndexOutOfBoundsException: 762
        at akka.http.impl.engine.parsing.HttpHeaderParser.parseHeaderValue(HttpHeaderParser.scala:194)
        at akka.http.impl.engine.parsing.HttpHeaderParser.parseHeaderLine(HttpHeaderParser.scala:130)
        at akka.http.impl.engine.parsing.HttpMessageParser$class.parseHeaderLines(HttpMessageParser.scala:125)
        at akka.http.impl.engine.parsing.HttpRequestParser$$anon$1.parseHeaderLines(HttpRequestParser.scala:44)
        at akka.http.impl.engine.parsing.HttpRequestParser$$anon$1.parseMessage(HttpRequestParser.scala:77)
        at akka.http.impl.engine.parsing.HttpMessageParser$class.startNewMessage(HttpMessageParser.scala:100)
        at akka.http.impl.engine.parsing.HttpRequestParser$$anon$1.startNewMessage(HttpRequestParser.scala:44)
        at akka.http.impl.engine.parsing.HttpMessageParser$$anonfun$startNewMessage$1.apply(HttpMessageParser.scala:101)
        at akka.http.impl.engine.parsing.HttpMessageParser$$anonfun$startNewMessage$1.apply(HttpMessageParser.scala:101)
        at akka.http.impl.engine.parsing.HttpMessageParser$$anonfun$continue$2.apply(HttpMessageParser.scala:264)
        at akka.http.impl.engine.parsing.HttpMessageParser$$anonfun$continue$2.apply(HttpMessageParser.scala:264)
        at akka.http.impl.engine.parsing.HttpMessageParser$class.run$1(HttpMessageParser.scala:64)
        at akka.http.impl.engine.parsing.HttpMessageParser$class.parseBytes(HttpMessageParser.scala:78)
        at akka.http.impl.engine.parsing.HttpRequestParser$$anon$1.parseBytes(HttpRequestParser.scala:44)
        at akka.http.impl.engine.parsing.HttpMessageParser$class.parseSessionBytes(HttpMessageParser.scala:60)
[...]

PR #2306, opened 2018-12-06
=htc Fix broken multi-byte UTF-8 decoding if UTF-8 sequence is cut in the middle

jrudolph:
Refs #1484 (fixes the easy one of the two exceptions reported there).


akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpHeaderParser.scala
```diff
@@ -558,6 +558,7 @@ private[http] object HttpHeaderParser {
 
   @tailrec private def scanHeaderValue(hhp: HttpHeaderParser, input: ByteString, start: Int, limit: Int, log: LoggingAdapter,
                                        mode: IllegalResponseHeaderValueProcessingMode)(sb: JStringBuilder = null, ix: Int = start): (String, Int) = {
+    hhp.byteBuffer.clear()
```
jrudolph: The byteBuffer will only be filled here and is also cleared in `hhp.decodeByteBuffer`. The bug was, that `decodeByteBuffer` might not be reached if `byteAt` fails with a `NotEnoughDataException`.

PR #2307, opened 2018-12-06
Move RoutingSettings to akka-http module

renatocaval:
This PR moves the `RoutingSettings` to the `akka-http` module. 

This is currently in `akka-http-core` which is odd since the routing dsl is part of `akka-http` module. 

The build will fail because of MiMa.

`RoutingSettings` is public API, although they are marked with `@DoNotInherit` user may be relying on them. That could be a problem if someone is using that class without also using `akka-http` jar. That would be odd, but technically possible. 

Fot that we have three possible solutions: 
1. keep the old classes, mark them as deprecated and add new classes with a different name, eg: RoutingDslSettings?
1. add the exclusion to MiMa, although they are not private or internal API, they are not expected to be used. 
1. drop this PR and keep it as is

We have run into an issue where mixed versions of akka-http would cause an exception at runtime, for instance: `akka-http` v10.0.13 mixed with `akka-http-core` v10.1.5.

Moving the class to `akka-http` is not a real fix, but I would consider it just a maintenance thing. Just to keep the code base coherent. 

To solve the issue of having mixes `akka-http`, we should add a similar check like we now have in Akka.


akka-http-core/src/test/scala/akka/http/scaladsl/settings/SettingsEqualitySpec.scala
```diff
@@ -11,18 +11,7 @@ import org.scalatest.WordSpec
...
-    akka.http.routing {
-      verbose-error-messages = off
-      file-get-conditional = on
-      render-vanity-footer = yes
-      range-coalescing-threshold = 80
-      range-count-limit = 16
-      decode-max-bytes-per-chunk = 1m
-      decode-max-size = 8m
-      file-io-dispatcher = ${akka.stream.blocking-io-dispatcher}
-    }
```
renatocaval: this is the best example that that class was on the wrong place. 

In this test we check for the settings equality. The test works perfectly for all settings except for the RoutingSettings. To make it work, we need to define the settings in place because they are not in the reference.conf here.

renatocaval:
I tried to add a RoutingDslSettings, but that doesn't help much. The old RoutingSettings is passed to a few methods here and there so we can't trick MiMa on that. 

So, either we give up on this PR or we add deprecated methods everywhere.

raboof:
Would it be possible to move it to the akka-http project, but otherwise keep it binary compatible with the version that was in akka-http-core? that might be 'binary compatible' enough.

renatocaval:
Yes, that's what this PR is doing. I will fix MiMa and we can let it rest for a while why we think about it.

jlprat:
OK TO TEST

renatocaval:
@jlprat, the build won't pass because of MiMA.

jlprat:
@renatocaval Yes I know, it was just to ask Jenkins to pick it up. I doesn't matter if it stays red for some days.

raboof:
This seems safe enough to me. It may not be 'binary compatible' in the MiMa sense, but the reason we want to deliver bincompat is to make it possible for people to upgrade without fear, and here apart from the fact that any downstream projects that relied on RoutingSettings must also depend on akka-http (not just akka-http-core) they should be able to update just fine. Still curious why RoutingSettings was in core in the first place ;)

renatocaval:
@raboof I didn't dig further to find out the reason, but I believe it's a leftover of some refactoring. That's just a guess.

renatocaval:
For the record, 

Lagom 1.4.9 is using `akka-http-core` 10.0.14
Lagom 1.5.0 (upcoming) is using `akka-http-core` 10.1.5

At the end of the day, that's not an `akka-http` problem. But the fact that `RountingSettings` is on the "wrong" place make it fragile to for other libs or framework depending on it. And it breaks even between minor, bug fix, versions.

raboof:
Problems are expected when different versions of akka-http artifacts are mixed - even between minor, bug fix versions - so that cannot be a motivation for this change.

jlprat:
> I think the chance of someone depending on only akka-http-core and not akka-http but still using RoutingSettings are slim, so that justifies the 'binary incompatibility' in akka-http-core here.

@raboof is it worth adding this note in the migration guide? (https://github.com/akka/akka-http/blob/master/docs/src/main/paradox/migration-guide/migration-guide-10.1.x.md)

PR #2312, opened 2018-12-10
Update akka to 2.5.19

raboof:
Should fix compiling tests on java 11, #2243


raboof:
PLS BUILD

jlprat:
>Wow, we've been not updating for a while it seems.
Iirc, it's because we intentionally said that we shouldn't bump Akka dependency unless there was a case for it. So we don't force higher Akka versions on people using Akka HTTP.

raboof:
> Iirc, it's because we intentionally said that we shouldn't bump Akka dependency unless there was a case for it. So we don't force higher Akka versions on people using Akka HTTP.

I remember the same - this seems like a reasonable reason to bump it though, even if it is only a problem for us in tests.

jlprat:
> this seems like a reasonable reason to bump it though

Completely agree :+1:

PR #2314, opened 2018-12-10
Do not include error details in HTTP response

raboof:
Inspired by #2018

### DIFF
--- a/docs/src/main/paradox/routing-dsl/exception-handling.md
+++ b/docs/src/main/paradox/routing-dsl/exception-handling.md
@@ -94,6 +94,19 @@ please refer to @ref[the section above](exception-handling.md#exception-handling
 if the entity is not properly rejected or cancelled on the client side.
 @@@
 
+## Including sensitive data in exceptions
+
+To prevent certain types of attack, it is not recommended to include arbitrary invalid user input in the response.
+However, sometimes it can be useful to include it in the exception and logging for diagnostic reasons.
+In such cases, you can use exceptions that extend `ExceptionWithErrorInfo`, such as `IllegalHeaderException`:
+
+Scala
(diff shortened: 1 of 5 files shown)


raboof:
done

docs/src/main/paradox/routing-dsl/exception-handling.md
```diff
@@ -94,6 +94,19 @@ please refer to @ref[the section above](exception-handling.md#exception-handling
 if the entity is not properly rejected or cancelled on the client side.
 @@@
 
+## Including sensitive data in exceptions
```
jrudolph: :+1: for adding docs.

kag0:
@raboof this supersedes https://github.com/akka/akka-http/pull/2018 but imo we still need to make this change https://github.com/akka/akka-http/pull/2018/commits/a96e3fbc7cc10d58a2071050a314a0aafbc03cb, and the `formatPretty` [here](https://github.com/akka/akka-http/pull/2314/files#diff-a3c2fe9da784b62492e0c89750710e93R55) should probably be `format(settings.verboseErrorMessages)`.
The rationale being that we shouldn't log potentially sensitive data unless the user has configured it. I think logging sensitive data should be disabled by default, for example users who might get into trouble with compliance requirements not knowing that detail logging is enabled.

raboof:
@kag0 thanks for pointing that out. That seems like a separate consideration, given that `settings.verboseErrorMessages` is explicitly about the client output, not about the logging (https://github.com/akka/akka-http/blob/c733ab41e0cb843c9506e96ad9c7a825c2aa7468/akka-http/src/main/resources/reference.conf#L10). Could you open a separate ticket/PR?

PR #2323, opened 2018-12-15
Add Media types for RFC 7807

bigwheel:
https://tools.ietf.org/html/rfc7807

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/MediaType.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/MediaType.scala
@@ -338,6 +338,7 @@ object MediaTypes extends ObjectRegistry[(String, String), MediaType] {
   val `application/json`                                                          = awfc("json", HttpCharsets.`UTF-8`, "json")
   val `application/json-patch+json`                                               = awfc("json-patch+json", HttpCharsets.`UTF-8`)
   val `application/merge-patch+json`                                              = awfc("merge-patch+json", HttpCharsets.`UTF-8`)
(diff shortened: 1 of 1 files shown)


bigwheel:
Thank you for response!

I wrote it already ([this page](https://doc.akka.io/docs/akka-http/current/routing-dsl/rejections.html#customising-rejection-http-responses) was so helpful), and it works well as I expect.
However it is company private repository and I cannot access in weekend.

I'll share Monday it as gist.

bigwheel:
https://gist.github.com/bigwheel/14e521488527f35bf6efed19c2453861

```scala
case class ErrorResponseBody private(
    `type`: URI,
    title: String
)

// https://doc.akka.io/docs/akka-http/10.1.3/routing-dsl/rejections.html?language=scala#customising-rejection-http-responses
private val rfc7807RejectionHandler = RejectionHandler.default.mapRejectionResponse {
  case res @ HttpResponse(_, _, entity: HttpEntity.Strict, _) =>
    val message = entity.data.utf8String

    res.copy(
      entity = HttpEntity(
        CustomMediaTypes.`application/problem+json`,
        ErrorResponseBody(new URI("about:blank"), message).toJson.prettyPrint
      )
    )
  case x => x
}
```

Works well in my product.

kazufumi-nishida-www:
@jrudolph thanks!
I will suggest more proper code for akka-http in #2341 because https://github.com/akka/akka-http/pull/2323#issuecomment-447700620 code depends on spray-json and its still simple method, not RejectionHandler.
I probably create PR for #2341 in Decemeber.

PR #2331, opened 2018-12-19
Import directives instead of extending

raboof:


akka-http/src/main/java/akka/http/javadsl/server/HttpApp.java
```diff
@@ -28,7 +28,7 @@
  * It offers additional hooks to modify the default behavior.
  */
 @ApiMayChange
-public abstract class HttpApp extends AllDirectives {
+public abstract class HttpApp {
```
jlprat: Worth adding this to release notes, as it might break the code of whoever uses `HttpApp`.

jrudolph: Do we really want to do it here? It seems it is quite useful in that place because you can use `HttpApp` without adding much more imports.

jlprat: You are right, @jrudolph maybe we could keep it for `HttpApp`

raboof: It seems weird to recommend people to import, and then extend it ourselves in the 'standard example starting point'. I'm OK with closing this PR though.

jlprat: To be fair, `HttpApp` is still an experiment and is not the standard way to start a server. For example, the quickstart guides do not use `HttpApp`.

raboof:
Let's close this since it is not clear whether this is the direction to go in.

PR #2333, opened 2018-12-19
=pro when detecting master branch also check TRAVIS_BRANCH

jrudolph:
As Travis doesn't seem to check out the branch but a certain hash instead.


jrudolph:
> I slightly prefer solving this in `.travis.yml` like in [akka/alpakka/pull/1383/files](https://github.com/akka/alpakka/pull/1383/files) and keeping travis-specific things out of the scala codebase, but this is OK as well

Aha, even better. I agree! Changed the PR to do the same.

jrudolph:
Let's merge and see if it works.

jrudolph:
Seems to have worked :+1:

Issue #2330, opened 2018-12-19
Adding woff2 to MediaTypes


pme123:
Woff2 is not recognized as MediaType, because its entry is missing in MediaTypes.

In:
`akka.http.scaladsl.model.MediaTypes
`

Around Line 333 something like this should be added:

    val `application/font-woff2`  = abin("font-woff2", NotCompressible, "woff2")

raboof:
Shouldn't that be `font/woff2`? (https://dev.w3.org/webfonts/WOFF2/spec/)

A PR adding it would be appreciated, but for the time being you can add it to your application as a custom media type as described in https://doc.akka.io/docs/akka-http/current/common/http-model.html#registering-custom-media-types. Does that work for you?

pme123:
`font/woff2` I am not sure, I just found `application/font-woff2` and it solved my problem.

I use `akka-http` within `play` - Does this description work as well?

raboof:
I'm not sure how this works in Play - that might be a good question for https://discuss.playframework.com?

pme123:
ok, thanks for the support.

PR #2320, opened 2018-12-14
Deprecate unofficial Microsoft media-types and add the official ones

franktominc:
Hey guys, a first-time contributor here. Let's see how this goes

Closes #2264

PR #2334, opened 2018-12-19
#2330 Added application/font-woff2 media type for WOFF File Format 2.0

ashishkujoy:
#2330 Added application/font-woff2 media type. Refer
 https://tools.ietf.org/html/rfc8081#page-15


raboof:
OK TO TEST

ashishkujoy:
As per this RFC WOFF File Format 1 and 2 should be font/woff and font/woff2. As of now woff file format 1 is implemented as application/font-woff, so implemented application/font-woff2 to not break the existing contract and keep the implementation schematic.

raboof:
Perhaps we could deprecate application/font-woff and add font/woff and font/woff2 (similar to #2320)?

ashishkujoy:
> Perhaps we could deprecate application/font-woff and add font/woff and font/woff2 (similar to #2320)?

I created a new pull request with those change but as expected build is failing because of binary in compatibility. Refer pull request 2348

raboof:
(superseded by #2348)

Issue #2335, opened 2018-12-20
Runtime check that all akka-http modules have the same version


patriknw:
In same way as in Akka with `akka.util.ManifestInfo`

jrudolph:
Fixed by #2336.

PR #2336, opened 2018-12-20
Check that all akka-http modules have the same version, #2335

patriknw:
This is how the warning log looks like if you mess up:
```
[WARN] [12/20/2018 10:19:07.966] [run-main-0] [ManifestInfo(akka://helloAkkaHttpServer)] Detected possible incompatible versions on the classpath. Please note that a given Akka HTTP version MUST be the same across all modules of Akka HTTP that you are using, e.g. if you use [10.1.6+7-3560e987] all other modules that are released together MUST be of the same version. Make sure you're using a compatible set of libraries. Possibly conflicting versions [10.1.3, 10.1.6+7-3560e987] in libraries [akka-http-spray-json:10.1.3, akka-parsing:10.1.6+7-3560e987, akka-http-xml:10.1.6+7-3560e987, akka-http:10.1.6+7-3560e987, akka-http-core:10.1.6+7-3560e987]
```

Refs #2335

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
@@ -36,13 +36,14 @@ import com.typesafe.config.Config
 import com.typesafe.sslconfig.akka._
 import com.typesafe.sslconfig.akka.util.AkkaLoggerFactory
 import com.typesafe.sslconfig.ssl.ConfigSSLContextBuilder
-
 import scala.concurrent._
 import scala.util.Try
 import scala.util.control.NonFatal
 import scala.compat.java8.FutureConverters._
 import scala.concurrent.duration.{ Duration, FiniteDuration }
 import scala.concurrent.duration._
+
+import akka.util.ManifestInfo
 
 /**
  * Akka extension for HTTP which serves as the main entry point into akka-http.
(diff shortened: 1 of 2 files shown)


raboof:
`Cannot run program "/usr/lib/jvm/jdk8u192-b12/bin/java"`, `Took 25 sec on johannes`

jrudolph:
PLS BUILD

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
```diff
@@ -1108,8 +1109,26 @@ object Http extends ExtensionId[HttpExt] with ExtensionIdProvider {
...
+    "akka-http-tests",
+    "akka-http-marshallers-scala",
+    "akka-http-marshallers-java",
+    "akka-http-spray-json",
+    "akka-http-xml",
+    "akka-http-jackson"
+  )
+
+  def createExtension(system: ExtendedActorSystem): HttpExt = {
+    ManifestInfo(system).checkSameVersion("Akka HTTP", allModules, logWarning = true)
```
jrudolph: There are also checks above at 

https://github.com/akka/akka-http/blob/3560e98754e7bf0d684a9b316926737dfa04c6a8/akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala#L57-L58

Could you change it so that all the checks are done in one place?

raboof:
I think we need both, indeed ;)

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
patriknw: I moved the check to HttpExt, after the other version checks

jrudolph: Thanks.

PR #2337, opened 2018-12-20
=ht2 harden h2spec execution against spaces in the executable path

jrudolph:
The akka-http-nightly build is run from a path that contains the names of all the axis. The JVM axis is now "AdoptOpenJDK 8", i.e. containing a space. The h2spec needs to run an executable with full path that then includes a space which currently doesn't work.

Not sure if quoting works with the ProcessBuilder but let's try.


jrudolph:
No quoting didn't work, but using a `Seq[String]` to define the command line seems to be ok.

jrudolph:
H2SpecIntegration tests seems to work here with the pr-validator-akka-http task but we still need to check the nightly after merging.

PR #2348, opened 2018-12-20
#2330 Add Font as main media type

ashishkujoy:
#2330 Deprecated already existing application/font-woff media type and font/woff and font/woff2 media.
Refer pull request https://github.com/akka/akka-http/pull/2334

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/javadsl/model/MediaType.scala
+++ b/akka-http-core/src/main/scala/akka/http/javadsl/model/MediaType.scala
@@ -89,8 +89,6 @@ trait MediaType {
 
   def isVideo: Boolean
 
-  def isFont: Boolean
-
   /**
    * Creates a media-range from this media-type.
    */
--- a/akka-http-core/src/main/scala/akka/http/javadsl/model/MediaType.scala
+++ b/akka-http-core/src/main/scala/akka/http/javadsl/model/MediaType.scala
@@ -89,6 +89,8 @@ trait MediaType {
 
   def isVideo: Boolean
 
+  def isFont: Boolean
+
   /**
    * Creates a media-range from this media-type.
    */
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/MediaType.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/MediaType.scala
@@ -332,6 +332,7 @@ object MediaTypes extends ObjectRegistry[(String, String), MediaType] {
   @deprecated("This format is unofficial and should not be used. Use application/vnd.ms-excel instead.", "10.1.6")
   val `application/excel`                                                         = abin("excel", NotCompressible)
   val `application/font-woff`                                                     = abin("font-woff", NotCompressible, "woff")
+  val `application/font-woff2`                                                    = abin("font-woff2", NotCompressible, "woff2")
   val `application/gnutar`                                                        = abin("gnutar", NotCompressible, "tgz")
   val `application/java-archive`                                                  = abin("java-archive", NotCompressible, "jar", "war", "ear")
   val `application/javascript`                                                    = awoc("javascript", "js")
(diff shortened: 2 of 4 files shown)


raboof:
OK TO TEST

akka-http-core/src/main/scala/akka/http/javadsl/model/MediaType.scala
```diff
@@ -89,6 +89,8 @@ trait MediaType {
 
   def isVideo: Boolean
 
+  def isFont: Boolean
```
raboof: Hmm, indeed this produces a binary incompatibility because we did not mark `MediaType` as `DoNotInherit` (though I think it was meant that way).

The easiest way around this would be not to introduce `isFont` - it is not directly needed anywhere, is it?

akka-http-core/src/main/scala/akka/http/scaladsl/model/MediaType.scala
```diff
@@ -298,7 +307,10 @@ object MediaTypes extends ObjectRegistry[(String, String), MediaType] {
   private def registerFileExtensions[T <: MediaType](mediaType: T): T = {
     mediaType.fileExtensions.foreach { ext ⇒
       val lcExt = ext.toLowerCase
-      require(!extensionMap.contains(lcExt), s"Extension '$ext' clash: media-types '${extensionMap(lcExt)}' and '$mediaType'")
+      /* Remove this condition of already registered media type should application/font-woff media type only once this media type is removed */
```
raboof: Similar to https://github.com/akka/akka-http/pull/2320/files#diff-a3293e8a7bed4283fabbccebd3cb5800R333, I think we can remove the association of the `woff` file extension with the 'old' content type. That would be a change in behavior, but a desirable one AFAICS.

akka-http-core/src/main/scala/akka/http/scaladsl/model/MediaType.scala
```diff
@@ -428,6 +442,9 @@ object MediaTypes extends ObjectRegistry[(String, String), MediaType] {
   val `audio/xm`          = aud("xm", NotCompressible, "xm")
   val `audio/webm`        = aud("webm", NotCompressible)
 
+  val `font/woff`  = font("woff", NotCompressible, "woff")
```
raboof: Perhaps link to the RFC in a comment here?

ashishkujoy: Added

akka-http-core/src/main/scala/akka/http/javadsl/model/MediaType.scala
ashishkujoy: Not needed directly. I think it will be better if we add that as soon as possible, because it causing inconsistance in code

akka-http-core/src/main/scala/akka/http/scaladsl/model/MediaType.scala
```diff
@@ -73,6 +73,9 @@ object MediaType {
       override def isApplication = true
     }
 
+  def fontBinary(subType: String, comp: Compressibility, fileExtensions: String*): Binary =
```
raboof: for `application`, we have `applicationBinary`, `applicationWithFixedCharset` and `applicationWithOpenCharset`. For `font`, I don't think we need to call the method `fontBinary`, since at least for now all `font` subtypes are binary, like with `video` and similar below - so I would suggest calling this method `font` instead (and perhaps moving it down to below `video`)

raboof:
Failure was #2349

raboof:
PLS BUILD

ashishkujoy:
@raboof Why these commits are not merged? Is anything remaining?

raboof:
@ashishkujoy because the holidays happened :smile: - they're good to go indeed!

raboof:
Thanks again, nice contribution!

PR #2351, opened 2018-12-29
Rfc7807 rejection handlers

### LINKED ISSUE #2341: Provide RejectionHandlers to provide error messages in RFC 7807 format
I.e. in XML or JSON format. #2323 already adds the MediaTypes.

@bigwheel already shared a solution in https://github.com/akka/akka-http/pull/2323#issuecomment-447700620

bigwheel:
Resolve #2341 

I'm newbie as akka-http contributer, therefore I'm afraid of forgetting basic things.
Please tell me honest review.
For instance, there may be better names for variable "RejectionHandler.rfc7807Json" "RejectionHandler.rfc7807Xml".

### DIFF
--- a/akka-http/src/main/scala/akka/http/scaladsl/server/RejectionHandler.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/server/RejectionHandler.scala
@@ -282,6 +282,43 @@ object RejectionHandler {
       .result()
 
   /**
+   * [[https://tools.ietf.org/html/rfc7807 RFC 7807]] [[RejectionHandler]] instance.
+   */
+  final val rfc7807Json = default.mapRejectionResponse {
+    case res @ HttpResponse(_, _, ent: HttpEntity.Strict, _) ⇒
+      val title = ent.data.utf8String.replaceAll("\"", """\"""")
+
+      res.copy(
+        entity = HttpEntity(
+          MediaTypes.`application/problem+json`,
+          s"""{"type": "about:blank", "title": "$title"}"""
+        )
+      )
+    case x ⇒ x
+  }
+
+  /**
+   * [[https://tools.ietf.org/html/rfc7807 RFC 7807]] [[RejectionHandler]] instance.
+   */
+  final val rfc7807Xml = default.mapRejectionResponse {
+    case res @ HttpResponse(_, _, ent: HttpEntity.Strict, _) ⇒
+      val title = ent.data.utf8String
+
+      res.copy(
+        entity = HttpEntity(
+          ContentType(MediaTypes.`application/problem+xml`, HttpCharsets.`UTF-8`),
+          s"""<?xml version="1.0" encoding="UTF-8"?>
+             |<problem xmlns="urn:ietf:rfc:7807">
+             |  <type>about:blank</type>
+             |  <title>$title</title>
+             |</problem>
+           """.stripMargin
+        )
+      )
+    case x ⇒ x
+  }
+
+  /**
    * Filters out all TransformationRejections from the given sequence and applies them (in order) to the
    * remaining rejections.
    */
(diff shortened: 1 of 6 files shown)


bigwheel:
I signed the CLA.

jlprat:
OK TO TEST

bigwheel:
https://jenkins.akka.io:8498/job/pr-validator-akka-http/3096/console

Sorry I cannot understood why test failed. Can anyone help me ?

raboof:
@bigwheel I'm sorry, we're updating our Jenkins instances and moving into the new year, I think the failure is related to that. Let's give it another try ;)

raboof:
PLS BUILD

raboof:
(it seems the validator is confused, but I have manually confirmed that you have indeed signed the CLA)

bigwheel:
Test passed and I signed CLA already.
What else can I do for merge ?

johanandren:
Hi @bigwheel thanks for the contribution, for a merge the PR needs two reviews from the Akka HTTP team, you can read more about that here: https://github.com/akka/akka-http/blob/master/CONTRIBUTING.md#general-workflow

We'll try to review as soon as possible but we are very busy so please give us some time!

akka-http/src/main/scala/akka/http/scaladsl/server/RejectionHandler.scala
```diff
@@ -281,6 +281,43 @@ object RejectionHandler {
...
+   * [[https://tools.ietf.org/html/rfc7807 RFC 7807]] [[RejectionHandler]] instance.
+   */
+  final val rfc7807Json = default.mapRejectionResponse {
+    case res @ HttpResponse(_, _, ent: HttpEntity.Strict, _) ⇒
+      val title = ent.data.utf8String.replaceAll("\"", """\"""")
+
+      res.copy(
+        entity = HttpEntity(
+          MediaTypes.`application/problem+json`,
+          s"""{"type": "about:blank", "title": "$title"}"""
```
johanandren: I read the RFC as `type` being optional, and if missing it means `about:blank` not that it should be set to `about:blank`, so we can just skip it and only include the title, or, more useful, encode one for each error (since that is sort of the point with the RFC, that a machine can discern the various errors and deal with them)

>  When this member is not present, its value is assumed to be "about:blank".

Also, the `title`:

> It SHOULD NOT change from occurrence to occurrence of the problem

While the error text from above will change with details about the failure, I think maybe this default->HttpResponse->map-body path is not the right way to go, and that we should rather duplicate the rejections list from above and provide a proper title + description, alternatively abstract somehow over the title-description so we can re-use the same logic.

For example, for a `UnsupportedRequestEncodingRejection` I think the JSON should be something like:

```
{
  "type": "https://myserver.com/problems/unsupported-encoding",
  "title": "Unsupported request encoding",
  "description": "The request's Content-Encoding is not supported.",
  "expected": "application/xml"
}
```

I think since it then already becomes a bit more complicated json/xml it should perhaps not live here but come with the json and xml modules and use their infra for structuring the content.

bigwheel: I entirely agree.

> I think maybe this default->HttpResponse->map-body path is not the right way to go ...
> I think since it then already becomes a bit more complicated json/xml it should perhaps not live here but come with the json and xml modules and use their infra for structuring the content.

Agree, this PR is like a work around.

> come with the json and xml modules

Is this modules libraries like spray-json or scala-xml, right?
As a matter of fact, once I was thinking to add the libraries but I found akka-http depends on only akka-http-core and scala-library. I felt this would be kept explicitly so.
Then I kept my PR in work around without additional libraries.

bigwheel: You say so, `type` and `title` have problems.

Because our company API server has `type` property, I added `type` unnecessity.
About title, 

> It SHOULD NOT change from occurrence to occurrence of the problem

I know this and $title string can be changed depends on context, but description without title feels so odd, then I select title.

> we should rather duplicate the rejections list from above and provide a proper title + description, alternatively abstract somehow over the title-description so we can re-use the same logic.

Yes, I think this resolves all too, but It needs json/xml libraries otherwise we have to write json/xml dump codes.

johanandren: Yes, I was thinking it could go in the akka-http-spray-json, akka-http-xml (and maybe also akka-http-jackson for the java users). So some abstraction so that the most of the logic can go in Akka HTTP core and then minimal API to plug in turning it into JSON and XML perhaps.

bigwheel: It sounds good.
But this becomes a big deal for me (this is second PR) now :|

johanandren: Yes, sorry that it grew so much, I think it is necessary to be useful though. Great if you are up for it but if you think it became too much I can understand that.

bigwheel:
Whatever I do https://github.com/akka/akka-http/pull/2351#discussion_r247508818 or not, this PR seems not good to merge.
Then I close this once and create another PR if I do https://github.com/akka/akka-http/pull/2351#discussion_r247508818.

PR #2358, opened 2019-01-02
Be less clever about copyright years

raboof:


jrudolph:
Hmm, how does that help? Don't we still have to update everything and all open PRs anyway?

raboof:
This way we can update the 'current year' and all copyright headers in 1 PR. Then only PR's that add new files will also need to be updated, but others should be fine.

Before this change, after newyear all PR's would fail PR validation, since their headers became 'incorrect' overnight.

jrudolph:
Ah, cool makes sense. Thanks for the explanation.

PR #2359, opened 2019-01-03
Bring both contributing files (github and docs) on par

jlprat:
- Changed copyright year to 2019 in the additional guidelines section
- Added reference to internal api, api may change annotations on the doc version of it.


jlprat:
We could make one symlink from the other. However, the main title differs on both versions

jlprat:
Changed the doc page to link to the github page cc @raboof

PR #2365, opened 2019-01-03
Publish snapshots of every master commit

### LINKED ISSUE #1254: Publish nightly snapshots

2m:
To consume snapshots, the following resolver is to be used:

```scala
resolvers += Resolver.bintrayRepo("akka", "snapshots")
```

Fixes #1254

Supersedes #1744

### DIFF
--- a/notes/release-train-issue-template.md
+++ b/notes/release-train-issue-template.md
@@ -38,6 +38,7 @@ Wind down PR queue. There has to be enough time after the last (non-trivial) PR 
 ### Cutting the release
 
 - [ ] Make sure there are no stray staging repos on sonatype
+- [ ] Wait until [master build finished](https://travis-ci.org/akka/akka-http/builds/) after merging the release notes
 - [ ] Create a [new release](https://github.com/akka/akka-http/releases/new) with the next tag version (e.g. `v13.3.7`), title and release description linking to announcement, release notes and milestone.
 - [ ] Check that the Travis CI release build has executed successfully
 - [ ] Go to https://bintray.com/akka/maven/com.typesafe.akka:akka-http_2.11 and select the just released version
--- a/project/Publish.scala
+++ b/project/Publish.scala
@@ -30,7 +30,8 @@ object Publish extends AutoPlugin {
 
   override def projectSettings = Seq(
     bintrayOrganization := Some("akka"),
-    bintrayPackage := "com.typesafe.akka:akka-http_2.11"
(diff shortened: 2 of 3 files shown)


project/Publish.scala
```diff
@@ -30,7 +30,8 @@ object Publish extends AutoPlugin {
 
   override def projectSettings = Seq(
     bintrayOrganization := Some("akka"),
-    bintrayPackage := "com.typesafe.akka:akka-http_2.11"
+    bintrayPackage := "com.typesafe.akka:akka-http_2.11",
+    bintrayRepository := (if (isSnapshot.value) "snapshots" else "maven")
```
raboof: This does risk doing the non-snapshot publish twice: when you create the latest commit on master and then tag it, travis will start 2 jobs, but both will see the tag and select 'maven' here. Perhaps just add a note to the release train to wait for this?

2m: Yes. And such situation became especially common once we started committing release notes to documentation. After that PR is merged the tag is created usually very quickly afterwards, possibly making both builds release builds.

We have a [note about that in the Alpakka release train checklist](https://github.com/akka/alpakka/commit/2e295782a02dbdd0ff84b9299ea63040b360d25c#diff-5aa61b36f0395eb9d1fd1d14debe2e5cR39). Will add it to akka-http as well.

2m:
Release train issue template amended.

jrudolph:
PLS BUILD

notes/release-train-issue-template.md
```diff
@@ -38,6 +38,7 @@ Wind down PR queue. There has to be enough time after the last (non-trivial) PR
 ### Cutting the release
 
 - [ ] Make sure there are no stray staging repos on sonatype
+- [ ] Wait until [master build finished](https://travis-ci.org/akka/akka-http/builds/) after merging the release notes
```
jrudolph: Probably it's enough to wait for it to be started and has fetched the tags from git.

jrudolph:
Thanks, @2m!

jrudolph:
First run going on here: https://travis-ci.org/akka/akka-http/jobs/477749819

PR #2373, opened 2019-01-08
Set a more descriptive message to the RejectionError exception

julienrf:
The toString implementation inherited by java.lang.Throwable only prints
the class name followed by the result of the getMessage method (which,
by default, returns null). As a consequence, the rejection carried by a
RejectionError instance is not shown.

We pass the given rejection to the RuntimeException constructor so that
it is shown as the exception message.


raboof:
OK TO TEST

jrudolph:
Sounds useful. What's the scenario when `RejectionError`s are printed?

julienrf:
@jrudolph I often log exceptions:

~~~ scala
try {
  something()
} catch {
  case NonFatal(exception) => log.error("Oops", exception)
}
~~~

PR #2375, opened 2019-01-09
=ht2 use JDK9+ ALPN support if available, fixes #2280

### LINKED ISSUE #2280: Make ALPN for HTTP/2 over TLS work with java11
Just a simple hack - copying the existing `Http2AlpnSupport` to my code and changing the `enableForServer` method to use Java11 addition to `SSLEngine` seems to work (If I pass sslParameters that have had `sslParameters.setApplicationProtocols(Array("h2", "http/1.1"))` called.

This was just to prove that it could work - obviously a "real" solution needs to be made smartly to use java11 (or is it java9 and on) methods to access the ALPN, if not set these sslParameters by default as well.

``` 
/**
    * Enables server-side Http/2 ALPN support for the given engine.
    */
  def enableForServer(engine: SSLEngine, setChosenProtocol: String ⇒ Unit): SSLEngine = {
    engine.setHandshakeApplicationProtocolSelector { (_, options) =>
      val ret = if (options.contains("h2")) "h2" else "http/1.1"
      setChosenProtocol(ret)
      ret
    }
    engine
  }
```

jrudolph:


akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2AlpnSupport.scala
```diff
@@ -23,12 +27,32 @@ private[http] object Http2AlpnSupport {
...
+
+  def isAlpnSupportedByJDK: Boolean =
+    // ALPN is supported starting with JDK 9
+    JavaVersion.majorVersion >= 9
+
+  private type JDK9SSLEngine = {
+    def setHandshakeApplicationProtocolSelector(selector: BiFunction[SSLEngine, ju.List[String], String]): Unit
+  }
+  def jdkAlpnSupport(engine: SSLEngine, setChosenProtocol: String ⇒ Unit): SSLEngine = {
+    engine.asInstanceOf[JDK9SSLEngine].setHandshakeApplicationProtocolSelector { (_, protocols: ju.List[String]) ⇒
```
raboof: Ha, cool use of structural types. I guess we can't really test this automatically, can we?

akka-http-core/src/main/scala/akka/http/impl/util/JavaVersion.scala
```diff
@@ -0,0 +1,25 @@
+/**
+ * Copyright (C) 2009-2018 Lightbend Inc. <https://www.lightbend.com>
```
raboof: oh, 2019 ;)

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2AlpnSupport.scala
jrudolph: I configured nightlies with JDK 11 here: https://jenkins.akka.io:8498/job/akka-http-nightly/

jrudolph: We probably will have to revisit that code because it uses reflection which right now shows a warning:

```
[error] WARNING: An illegal reflective access operation has occurred
[error] WARNING: Illegal reflective access by scala.reflect.package$ (file:/tmp/sbt_8c5ade58/target/36b23483/scala-library-2.12.8.jar) to method sun.security.ssl.SSLEngineImpl.setHandshakeApplicationProtocolSelector(java.util.function.BiFunction)
[error] WARNING: Please consider reporting this to the maintainers of scala.reflect.package$
[error] WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
[error] WARNING: All illegal access operations will be denied in a future release
```

The alternative would be to somehow provide JDK 9+ binaries.

akka-http-core/src/main/scala/akka/http/impl/util/JavaVersion.scala
jrudolph: We are still in 2018 in akika-http but a newline was missing ;)

Issue #2277, opened 2018-11-08
Add support for overriding User-Agent headers by RawHeader in the http client


arempter:
Hello,
Currenlty it seems not possible to create http client request with raw header overriding User-Agent. Or when creating request like one below:

val request =HttpRequest(uri = "http://localhost:21000").addHeader(`User-Agent`("aws-sdk-java/1.11.358 Windows_7/6.1 Java_HotSpot(TM)_64-Bit_Server_VM/25.121-b13 java/1.8.0_121 scala/2.12.4"))

Above adds custom user agent string but it seems to be parsed by Akka client. 

In cases where I want to create proxy, and want to forward client user-agent, it will not work as the header value will change.
In akka http configuration there are some settings, to loosen parsing like:

server.parsing.modeled-header-parsing = off
client.parsing.modeled-header-parsing = off <- this one seems to have no effect

Some more details in:
https://discuss.lightbend.com/t/akka-http-client-disable-parsing-of-custom-user-agent-header/2659

Could you please consider adding possibility of overriding User-Agent in akka http client when making request?

Thanks,
Adam

PR #2379, opened 2019-01-11
Allow raw header User-Agent #2277

WalkerTR:
Fix for issue #2277

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/rendering/HttpRequestRendererFactory.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/rendering/HttpRequestRendererFactory.scala
@@ -87,10 +87,14 @@ private[http] class HttpRequestRendererFactory(
             if (x.renderInRequests) render(x)
             renderHeaders(tail, hostHeaderSeen, userAgentSeen, transferEncodingSeen)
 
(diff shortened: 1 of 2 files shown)


lightbend-cla-validator:
Hi @WalkerTR,

raboof:
OK TO TEST

akka-http-core/src/main/scala/akka/http/impl/engine/rendering/HttpRequestRendererFactory.scala
```diff
@@ -88,7 +88,7 @@ private[http] class HttpRequestRendererFactory(
             renderHeaders(tail, hostHeaderSeen, userAgentSeen, transferEncodingSeen)
 
           case x: RawHeader if (x is "content-type") || (x is "content-length") || (x is "transfer-encoding") ||
-            (x is "host") || (x is "user-agent") ⇒
```
jrudolph: This is not enough. You also need to adapt `userAgentSeen`, otherwise, the default user agent will be added as well.

WalkerTR:
Thanks @jrudolph for your feedback! I made the requested change.

I was also wondering whether we should extend this case to other `RawHeader`s, e.g. `Host`. Let me know if I need to implement this, I may even create a different PR.

WalkerTR:
Sure! I will create the PR for the `Host` header in about two weeks, since I am on holiday right now.

Issue #2382, opened 2019-01-14
Include path in truncation errors


johanandren:
When failing because of truncation, with `EntityStreamTruncation` or `ParsingException: Unexpected end of multipart entity` it would be useful if we could include the path of the request.

johanandren:
It would also be good to have the path in the `Sending an 2xx 'early' response before end of request was received` warning.

johanandren:
Out of these three the early response is the only one I can see how we can include path in.

PR #2383, opened 2019-01-14
Include path in early response warning

johanandren:
Refs #2382


akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -419,7 +419,7 @@ private[http] object HttpServerBluePrint {
           val isEarlyResponse = messageEndPending && openRequests.isEmpty
           if (isEarlyResponse && response.status.isSuccess)
             log.warning(
-              "Sending an 2xx 'early' response before end of request was received... " +
+              s"Sending an 2xx 'early' response before end of request for ${requestStart.uri.path} received... " +
```
jrudolph: Maybe print the full URI? The host could be relevant if someone uses multi-homing.

```suggestion
              s"Sending an 2xx 'early' response before end of request for ${requestStart.uri} received... " +
```

johanandren: Good point

PR #2389, opened 2019-01-15
Add akka-stream-testkit

rtfpessoa:
To run the example in this page I also needed the `akka-stream-testkit` dependency.
I believe it would make sense to have it in this page.


jrudolph:
OK TO TEST

Issue #358, opened 2016-09-28
Receiving overlarge request should return 413 response instead of 500


jypma:
Akka 2.4.10 is returning HTTP 500 instead of 413 when receiving an overlarge request (with chunked encoding):

```
09:26:13,337 ERROR [a.a.ActorSystemImpl] akka.actor.ActorSystemImpl(document-core) - Outgoing request stream error
EntityStreamSizeException: actual entity size (None) exceeded content length limit (52428800 bytes)! You can configure this by setting `akka.http.[server|client].parsing.max-content-length` or calling `HttpEntity.withSizeLimit` before materializing the dataBytes stream.
    at akka.http.scaladsl.model.HttpEntity$Limitable$$anon$1.onPush(HttpEntity.scala:620)
    at akka.stream.impl.fusing.GraphInterpreter.processPush(GraphInterpreter.scala:747)
    at akka.stream.impl.fusing.GraphInterpreter.execute(GraphInterpreter.scala:649)
    at akka.stream.impl.fusing.GraphInterpreterShell.runBatch(ActorGraphInterpreter.scala:471)
    at akka.stream.impl.fusing.GraphInterpreterShell.receive(ActorGraphInterpreter.scala:423)
    at akka.stream.impl.fusing.ActorGraphInterpreter.akka$stream$impl$fusing$ActorGraphInterpreter$$processEvent(ActorGraphInterpreter.scala:603)
    at akka.stream.impl.fusing.ActorGraphInterpreter$$anonfun$receive$1.applyOrElse(ActorGraphInterpreter.scala:618)
    at akka.actor.Actor$class.aroundReceive(Actor.scala:484)
    at akka.stream.impl.fusing.ActorGraphInterpreter.aroundReceive(ActorGraphInterpreter.scala:529)
    at akka.actor.ActorCell.receiveMessage(ActorCell.scala:526)
    at akka.actor.ActorCell.invoke_aroundBody0(ActorCell.scala:495)
    at akka.actor.ActorCell$AjcClosure1.run(ActorCell.scala:1)
    at org.aspectj.runtime.reflect.JoinPointImpl.proceed(JoinPointImpl.java:149)
    at akka.kamon.instrumentation.ActorMonitors$$anon$1$$anonfun$processMessage$1.apply(ActorMonitor.scala:59)
    at kamon.trace.Tracer$.withContext(TracerModule.scala:58)
[...]

PR #2395, opened 2019-01-21
Oversized requests should return 413 and not 500 #358

petermz:
Hi there!
To fix this bug I've followed the route suggested in the comments. A unit test was also patched.

### DIFF
--- a/akka-http/src/main/scala/akka/http/scaladsl/server/ExceptionHandler.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/server/ExceptionHandler.scala
@@ -51,6 +51,11 @@ object ExceptionHandler {
         ctx.request.discardEntityBytes(ctx.materializer)
         ctx.complete((status, info.format(settings.verboseErrorMessages)))
       }
+      case e: EntityStreamSizeException ⇒ ctx ⇒ {
+        ctx.log.error(e, ErrorMessageTemplate, e, RequestEntityTooLarge)
+        ctx.request.discardEntityBytes(ctx.materializer)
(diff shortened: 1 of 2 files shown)


johanandren:
OK TO TEST

akka-http/src/main/scala/akka/http/scaladsl/server/ExceptionHandler.scala
```diff
@@ -51,6 +51,11 @@ object ExceptionHandler {
         ctx.request.discardEntityBytes(ctx.materializer)
         ctx.complete((status, info.format(settings.verboseErrorMessages)))
       }
+      case e: EntityStreamSizeException ⇒ ctx ⇒ {
+        ctx.log.error(e, ErrorMessageTemplate, e, RequestEntityTooLarge)
+        ctx.request.discardEntityBytes(ctx.materializer)
+        ctx.complete((RequestEntityTooLarge, e.toString.format(settings.verboseErrorMessages)))
```
raboof: I think `e.getMessage` would be neater here compared to `e.toString`.

The `.format()` looks a bit like the `e.info.format()` below, but is in fact very different: this is https://www.scala-lang.org/api/current/scala/collection/immutable/StringOps.html#format(args:Any*):String which is not needed here since the exception message doesn't have a placeholder that needs to be filled.

```suggestion
        ctx.complete((RequestEntityTooLarge, e.getMessage))
```

petermz: I agree. Just pushed this change. Thanks!

PR #2396, opened 2019-01-21
replace logRequestResponse with logRequestResult

jcranky:
I guess there was a rename and the comment got left behind?


jlprat:
OK TO TEST

PR #2398, opened 2019-01-23
=htp Turn `Directive.SingleValueModifiers` into a value class

sirthias:
Just stumbled this (probably my) mistake from way back.
It appears to be the only occasion in the code base.

Trivial fix but unfortunately not fully binary compatible (or is it? Maybe at least "sufficiently compatible"?).


sirthias:
Mima failure, as expected.

Is there maybe some kind of "future" branch (for the next binary incompatible release) that would be a better target for this?

raboof:
I wonder if we could make this binary compatible by having the 'old' implementation as a deprecated (private?) non-implicit class and provide the 'new' one with a different name?

sirthias:
Awesome idea, Arnout!
I'll give it a try right away and see, whether we can convince Mima...

sirthias:
Thank you, Johannes! :)
I'm probably going to be around a little more in the coming months...

PR #2402, opened 2019-01-29
Disable to append charset=UTF-8 by default in `Content-Type: application/x-www-form-urlencoded` FormData

### LINKED ISSUE #91: Issue using Http().outgoingConnectionTls with HttpEntity and ContentType of application/x-www-form-urlencoded
<a href="https://github.com/TimGraf"></img></a> **Issue by [TimGraf](https://github.com/TimGraf)**
_Tuesday Jan 12, 2016 at 21:36 GMT_
_Originally opened as https://github.com/akka/akka/issues/19420_

---

I recently updated from Akka Streams and HTTP version 1.0 to version 2.0.1.  I have a sample service in which I was making an HTTPS connection to an external REST API with a content type of application/x-www-form-urlencoded for the submitted data.

In version 1.0 I used the following code.

``` scala
  def pnConnectionFlow()(implicit as: ActorSystem, mt: Materializer): Flow[HttpRequest, HttpResponse, Future[Http.OutgoingConnection]] = Http().outgoingConnectionTls(apiUrl.getHost)
  def pnFutureResponse(request: HttpRequest)(implicit as: ActorSystem, mt: Materializer): Future[HttpResponse]                         = Source.single(request).via(pnConnectionFlow).runWith(Sink.head)

   /*
       Sample curl command for the call

       curl -X POST --include 'https://neutrinoapi-phone-validate.p.mashape.com/phone-validate' \
       -H 'X-Mashape-Key: <api-key>' \
       -H 'Content-Type: application/x-www-form-urlencoded' \
       -H 'Accept: application/json' \
       -d 'country-code=us' \
      -d 'number=+18316561725'
    */
   def validate(numberString: String)(implicit as: ActorSystem, mt: Materializer, ec: ExecutionContext): Future[ValidateResponse] = {
     val pnRequest: HttpRequest = HttpRequest(POST, apiUrl.getPath)
       .withHeaders(
         RawHeader("X-Mashape-Key", apiKey),
         RawHeader("Accept", "application/json")
       )
       .withEntity(
         HttpEntity(ContentType(MediaTypes.`application/x-www-form-urlencoded`), s"country-code=us&number=${URLEncoder.encode(numberString, "UTF-8")}")
       )

     pnFutureResponse(pnRequest).flatMap { response =>
       response.status match {
         case OK =>
           Unmarshal(response.entity).to[ValidateResponse].flatMap {
             case validateResponse: ValidateResponse => Future.successful(validateResponse)
           }

         case _ =>
           val message = s"Phone number validation failed: ${response.status}"

           logger.debug(message)
           Future.failed(new Exception(message))
       }
     } recoverWith  {
       case error => Future.successful(createInvalidResponse())
     }
   }
```

In version 2.0.1 I tried using the following.

I tried creating the HTTP Entity with the following which does not work.

``` scala
val entity: RequestEntity     = HttpEntity(ContentType(MediaTypes.`application/x-www-form-urlencoded`, HttpCharsets.`UTF-8`), data)
```

The external service would reply with a 403.  So I ended up using the following which was to create an HTTP Entity with a custom binary content type.

``` scala
withContentType(MediaType.customBinary("application", "x-www-form-urlencoded", comp = NotCompressible))
```

It seems the ContentType using the MediaTypes.`application/x-www-form-urlencoded` does not work as expected.  For comparison here is the code I am using now.

``` scala
  def pnConnectionFlow()(implicit as: ActorSystem, mt: Materializer): Flow[HttpRequest, HttpResponse, Future[Http.OutgoingConnection]] = Http().outgoingConnectionTls(apiUrl.getHost)
  def pnFutureResponse(request: HttpRequest)(implicit as: ActorSystem, mt: Materializer): Future[HttpResponse]                         = Source.single(request).via(pnConnectionFlow).runWith(Sink.head)

/*
     Sample curl command for the call

     curl -X POST --include 'https://neutrinoapi-phone-validate.p.mashape.com/phone-validate' \
     -H 'X-Mashape-Key: <api-key>' \
     -H 'Content-Type: application/x-www-form-urlencoded' \
     -H 'Accept: application/json' \
     -d 'country-code=us' \
     -d 'number=+18316561725'
   */
  def validate(numberString: String)(implicit as: ActorSystem, mt: Materializer, ec: ExecutionContext): Future[ValidateResponse] = {
    val data: ByteString          = ByteString(s"country-code=us&number=${URLEncoder.encode(numberString, "UTF-8")}")
    val headers: List[HttpHeader] = List(RawHeader("X-Mashape-Key", apiKey), RawHeader("Accept", "application/json"))
    val entity: RequestEntity     = HttpEntity(data).withContentType(MediaType.customBinary("application", "x-www-form-urlencoded", comp = NotCompressible))
    val pnRequest: HttpRequest    = HttpRequest(POST, apiUrl.getPath, headers, entity)

    val responseWithTimeout = Future.firstCompletedOf(
      pnFutureResponse(pnRequest) ::
      after(serviceTimeout.second, as.scheduler)(Future.failed(new TimeoutException)) ::
      Nil
    )

    responseWithTimeout.flatMap { response =>
      response.status match {
        case OK =>
          Unmarshal(response.entity).to[ValidateResponse].flatMap {
            case validateResponse: ValidateResponse => Future.successful(validateResponse)
          }

        case _ =>
          val message = s"Phone number validation failed: ${response.status}"

          logger.info(message)
          Future.failed(new Exception(message))
      }
    } recoverWith  {
      case error => Future.successful(createInvalidResponse())
    }
  }
```

monkey-mas:
Closes #91

Currently `charset=UTF-8` is added to `Content-Type: application/x-www-form-urlencoded` by default. This mime-type, however, takes no charset parameter as IANA describes [its specification](https://www.iana.org/assignments/media-types/application/x-www-form-urlencoded). ~As [W3 application-x-www-form-urlencoded encoding algorithm](https://www.w3.org/TR/html5/sec-forms.html#application-x-www-form-urlencoded-encoding-algorithm) describes, `charset` is not expected.~ Thus we don't append `charset=UTF-8` in `Content-Type: application/x-www-form-urlencoded` FormData.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/FormData.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/FormData.scala
@@ -13,11 +13,12 @@ import akka.http.scaladsl.model.MediaTypes._
  */
 final case class FormData(fields: Uri.Query) {
   def toEntity: akka.http.scaladsl.model.RequestEntity =
-    toEntity(HttpCharsets.`UTF-8`)
+    toEntity(`application/x-www-form-urlencoded`.charset)
 
+  @deprecated("FormData always uses charset UTF-8 without appending the charset to 'Content-Type: application/x-www-form-urlencoded', use toEntity() instead.", "10.1.7")
   def toEntity(charset: HttpCharset): akka.http.scaladsl.model.RequestEntity = {
     val render: StringRendering = UriRendering.renderQuery(new StringRendering, this.fields, charset.nioCharset, CharacterClasses.unreserved)
-    HttpEntity(`application/x-www-form-urlencoded` withCharset charset, render.get)
+    HttpEntity(`application/x-www-form-urlencoded`, render.get)
   }
 }
 
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
@@ -278,6 +278,8 @@ object HttpEntity {
   implicit def apply(data: ByteString): HttpEntity.Strict = apply(ContentTypes.`application/octet-stream`, data)
   def apply(contentType: ContentType.NonBinary, string: String): HttpEntity.Strict =
     if (string.isEmpty) empty(contentType) else apply(contentType, ByteString(string.getBytes(contentType.charset.nioCharset)))
+  def apply(contentType: ContentType.WithFixedCharset, string: String): HttpEntity.Strict =
+    if (string.isEmpty) empty(contentType) else apply(contentType, ByteString(string.getBytes(contentType.charset.nioCharset)))
   def apply(contentType: ContentType, bytes: Array[Byte]): HttpEntity.Strict =
     if (bytes.length == 0) empty(contentType) else apply(contentType, ByteString(bytes))
   def apply(contentType: ContentType, data: ByteString): HttpEntity.Strict =
--- a/akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/FormFieldDirectivesSpec.scala
+++ b/akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/FormFieldDirectivesSpec.scala
@@ -126,9 +126,11 @@ class FormFieldDirectivesSpec extends RoutingSpec {
       val charset = HttpCharsets.`UTF-8`
       val render: StringRendering = UriRendering.renderQuery(new StringRendering, urlEncodedForm.fields, charset.nioCharset, CharacterClasses.unreserved)
       val streamingForm: RequestEntity = HttpEntity.Chunked(
-        `application/x-www-form-urlencoded` withCharset charset,
+        `application/x-www-form-urlencoded`,
         Source.single(ChunkStreamPart(render.get)).via(AllowMaterializationOnlyOnce())
       )
+
+      streamingForm.getContentType shouldEqual ContentTypes.`application/x-www-form-urlencoded`
       Post("/", streamingForm) ~> {
         formFields('firstName, "age".as[Int], 'sex.?, "VIP" ? false) { (firstName, age, sex, vip) ⇒
           complete(firstName + age + sex + vip)
--- a/akka-http-core/src/main/java/akka/http/javadsl/model/ContentTypes.java
+++ b/akka-http-core/src/main/java/akka/http/javadsl/model/ContentTypes.java
@@ -30,6 +30,8 @@ public final class ContentTypes {
           akka.http.scaladsl.model.ContentTypes.text$divcsv$u0028UTF$minus8$u0029();
 
   public static final ContentType.Binary APPLICATION_GRPC_PROTO = MediaTypes.APPLICATION_GRPC_PROTO.toContentType();
+  public static final ContentType.WithFixedCharset APPLICATION_X_WWW_FORM_URLENCODED =
+      akka.http.scaladsl.model.ContentTypes.application$divx$minuswww$minusform$minusurlencoded();
 
   public static final ContentType.Binary NO_CONTENT_TYPE =
           akka.http.scaladsl.model.ContentTypes.NoContentType();
--- a/akka-http-core/src/main/java/akka/http/javadsl/model/FormData.java
+++ b/akka-http-core/src/main/java/akka/http/javadsl/model/FormData.java
@@ -28,9 +28,13 @@ public final class FormData {
 
   /**
    * Converts this FormData to a RequestEntity using the given encoding.
+   *
+   * @deprecated FormData always uses charset UTF-8 without appending the charset to
+   *             'Content-Type: application/x-www-form-urlencoded', use toEntity() instead.
    */
+  @Deprecated
   public RequestEntity toEntity(HttpCharset charset) {
-    return HttpEntities.create(ContentTypes.create(MediaTypes.APPLICATION_X_WWW_FORM_URLENCODED, charset), fields.render(charset));
+    return HttpEntities.create(ContentTypes.APPLICATION_X_WWW_FORM_URLENCODED, fields.render(charset));
   }
 
   /**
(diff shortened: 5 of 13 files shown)


akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/FormFieldDirectivesSpec.scala
```diff
@@ -126,9 +126,11 @@ class FormFieldDirectivesSpec extends RoutingSpec {
       val charset = HttpCharsets.`UTF-8`
       val render: StringRendering = UriRendering.renderQuery(new StringRendering, urlEncodedForm.fields, charset.nioCharset, CharacterClasses.unreserved)
       val streamingForm: RequestEntity = HttpEntity.Chunked(
-        `application/x-www-form-urlencoded` withCharset charset,
+        `application/x-www-form-urlencoded` withMissingCharset,
```
monkey-mas: Technically, we can _still_ set `charset=UTF-8` in Content-Type using [Chunked(...)](https://github.com/akka/akka-http/blob/v10.1.7/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala#L496).

raboof:
OK TO TEST

jrudolph:
According to https://www.iana.org/assignments/media-types/application/x-www-form-urlencoded, this mime-type indeed takes no charset parameter. We should change the type of `application/x-www-form-urlencoded` from `WithOpenCharset` to `WithFixedCharset` and use UTF-8 as the fixed charset. WDYT?

monkey-mas:
Yeah, sounds good to me if this is the appropriate way of handling `application/x-www-form-urlencoded` in akka :)

akka-http-core/src/main/scala/akka/http/scaladsl/model/FormData.scala
```diff
@@ -13,11 +13,11 @@ import akka.http.scaladsl.model.MediaTypes._
  */
 final case class FormData(fields: Uri.Query) {
   def toEntity: akka.http.scaladsl.model.RequestEntity =
-    toEntity(HttpCharsets.`UTF-8`)
+    toEntity(`application/x-www-form-urlencoded`.charset)
 
   def toEntity(charset: HttpCharset): akka.http.scaladsl.model.RequestEntity = {
```
monkey-mas: We might deprecate `toEntity(HttpCharset)` as `FormData` has fixed charset UTF-8 now. WDYT?

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
```diff
@@ -278,6 +278,8 @@ object HttpEntity {
   implicit def apply(data: ByteString): HttpEntity.Strict = apply(ContentTypes.`application/octet-stream`, data)
   def apply(contentType: ContentType.NonBinary, string: String): HttpEntity.Strict =
     if (string.isEmpty) empty(contentType) else apply(contentType, ByteString(string.getBytes(contentType.charset.nioCharset)))
+  def apply(contentType: ContentType.WithFixedCharset, string: String): HttpEntity.Strict =
+    if (string.isEmpty) empty(contentType) else apply(contentType, ByteString(string.getBytes()))
```
monkey-mas: For the `apply(ContentType, ByteString)` part, which one do you think is better?;
`apply(contentType, ByteString(string.getBytes())`
OR
`apply(contentType, ByteString(string.getBytes(contentType.charset.nioCharset)))`

monkey-mas:
Umm, it seems like the test fails as this change breaks binary compatibility?

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
jrudolph: The second.

```suggestion
    if (string.isEmpty) empty(contentType) else apply(contentType, ByteString(string.getBytes(contentType.charset.nioCharset))))
```

akka-http-core/src/main/scala/akka/http/scaladsl/model/FormData.scala
jrudolph: Sounds good.

monkey-mas:
@jrudolph 
Thanks for your review!

It seems like Jenkins failed with some unrelated tests? 🤔 
https://jenkins.akka.io:8498/job/pr-validator-akka-http/3208/
https://jenkins.akka.io:8498/job/pr-validator-akka-http/3209/

johanandren:
PLS BUILD

monkey-mas:
@jrudolph Thanks for your review! Just fixed my commit message with a reference to IANA like you pointed it out :)

julienrf:
Upgrading to 10.1.8 lead to the following NSME:

~~~
java.lang.NoSuchMethodError: akka.http.scaladsl.model.MediaTypes$.application$divx$minuswww$minusform$minusurlencoded()Lakka/http/scaladsl/model/MediaType$WithOpenCharset;
        at akka.http.scaladsl.marshalling.PredefinedToEntityMarshallers.$init$(PredefinedToEntityMarshallers.scala:51)
        at akka.http.scaladsl.marshalling.Marshaller$.<init>(Marshaller.scala:81)
~~~

Do I understand correctly that the `akka-http` module was linked to an outdated version of the `akka-http-core` module?

julienrf:
Sorry, I found the culprit: I had a dependency on `akka-http-json`, which depends on `"akka-http" % "10.1.7"`. My build didn’t explicitly depend on `akka-http`, but only `"akka-http-core" % "10.1.8"`. So, the `akka-http-json` dependency was pulling a version of `akka-http` that didn’t work with my version of `akka-http-core`.

jrudolph:
Thanks for the update, @julienrf. We had to break binary compatibility for this one which isn't a problem if you use matched versions of akka-http and akka-http-core as recommended.

I now wonder how likely it will trip up other libraries. We haven't found a workaround, yet, how to keep a compatible value as well (because you cannot overload values).

PR #2407, opened 2019-01-30
Only provide a single fork for 2.11 and 2.12 source files

jrudolph:
Using the same trick as in akka.


raboof:
(failure is the header check, let's fix that in the other branch)

PR #2413, opened 2019-02-06
Don't hang in HttpAppSpec when privileged ports can be bound

jrudolph:


johanandren:
Who is running the test suite as root?

jrudolph:
The docker container I used while testing a new jenkins slave layout.

PR #2425, opened 2019-02-12
Jr/test auto formatter

jrudolph:


patriknw:
Has "Akka Jenkins" signed the CLA or how does that part work?

Will this work with Squash/merge button, giving attribution to the right person?

Will this trigger 2 PR validations?

jrudolph:
> Has "Akka Jenkins" signed the CLA or how does that part work?

I have no idea, I don't think so but at least the validator doesn't complain.

> Will this work with Squash/merge button, giving attribution to the right person?

Should be ok, usually, the first commit or the PR opener is attributed (can't remember).

> Will this trigger 2 PR validations?

Yes, would be nice to fix it but I haven't figured out how. It doesn't matter so much for akka-http because it's only a few minutes but it's not good for akka itself, I agree. One way to fix that would be to just include that step in the regular PR validation to be done before or after the actual build.

jrudolph:
The basic setup now works. I filed https://github.com/akka/akka-http/issues/2447 to investigate running the autoformatted as part of the regular PR validation.

PR #2426, opened 2019-02-14
=htc don't log AbruptTerminationException in client connections

jrudolph:
As it is expected when the ActorSystem shuts down while a client pool is running.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
@@ -96,7 +96,13 @@ private[http] object OutgoingConnectionBlueprint {
 
       val terminationFanout = b.add(Broadcast[HttpResponse](2))
 
-      val logger = b.add(Flow[ByteString].mapError { case t => log.error(t, "Outgoing request stream error"); t }.named("errorLogger"))
+      val logger = b.add(Flow[ByteString].mapError {
+        case ate: AbruptTerminationException => ate // don't log, because it's likely because of shutdown of the ActorSystem
+        case t =>
+          log.error(t, "Outgoing request stream error")
+          t
+      }.named("errorLogger"))
+
       val wrapTls = b.add(Flow[ByteString].map(SendBytes))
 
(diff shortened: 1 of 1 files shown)


johanandren:
But that is what the AbruptTerminationException is, same for arbitrary streams if you shut down the materializer or actor system. Wouldn't it hide if accidentally shutting down the system somewhere and using the client elsewhere and expecting it to be possible to keep using it?

jrudolph:
> But that is what the AbruptTerminationException is, same for arbitrary streams if you shut down the materializer or actor system. Wouldn't it hide if accidentally shutting down the system somewhere and using the client elsewhere and expecting it to be possible to keep using it?

I don't think interacting with a shut-down or shutting-down ActorSystem will lead to any kind of deterministic behavior, does it? In this case, the exception is shown on an idle pool while shutting down the system.

The whole business of `GraphInterpreterShell.tryAbort` seems futile anyway. It tries for a bit to propagate the `AbruptTerminationException` through the graph by starting from the `BatchingActorInputBoundary`. However, the only reason that some graphs still have those with subfusing is basically an implementation artefact.

johanandren:
Wasn't so much thinking of the case with shutting down an ActorSystem as when running it on a materializer that you shut down elsewhere (for example creating it from an actor as ActorRefFactory) and that actor shuts down. With this error in place at least you'd get a hint about that happening but if it just dies silently there is no clue. Not entirely sure that it matters, but maybe?

jrudolph:
> Wasn't so much thinking of the case with shutting down an ActorSystem as when running it on a materializer that you shut down elsewhere (for example creating it from an actor as ActorRefFactory) and that actor shuts down. With this error in place at least you'd get a hint about that happening but if it just dies silently there is no clue. Not entirely sure that it matters, but maybe?

Right, I didn't think about that `ActorRefFactory` case. It seems it's currently not possible to provide a materializer for the client APIs any more, so that might not be a concern?

johanandren:
Isn't the problem what materializer the returned flow for example from `cachedHostConnectionPool` is run with rather? Or is the blueprint materialized elsewhere somehow?

jrudolph:
> Isn't the problem what materializer the returned flow for example from `cachedHostConnectionPool` is run with rather? Or is the blueprint materialized elsewhere somehow?

I think you are right here, there are some APIs still where you can provide a materializer where the notice might be useful. I'll close it for now.

PR #2442, opened 2019-03-04
Link from 'configuration' to 'introduction'

raboof:
And add a link from Configuration to Introduction

### DIFF
--- a/docs/src/main/paradox/configuration.md
+++ b/docs/src/main/paradox/configuration.md
@@ -3,7 +3,7 @@
 Before configuring Akka HTTP it must be added to your project as described in
 @ref[the introduction](introduction.md#using-akka-http)
 
-Just like any other Akka module Akka HTTP is configured via [Lightbend Config](https://github.com/lightbend/config).
+Just like any other Akka module Akka HTTP is configured via [Typesafe Config](https://github.com/lightbend/config).
 Usually this means that you provide an `application.conf` which contains all the application-specific settings that
 differ from the default ones provided by the reference configuration files from the individual Akka modules.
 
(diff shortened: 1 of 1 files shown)


docs/src/main/paradox/configuration.md
```diff
@@ -1,6 +1,9 @@
 # Configuration
 
-Just like any other Akka module Akka HTTP is configured via [Typesafe Config](https://github.com/lightbend/config).
+Before configuring Akka HTTP it must be added to your project as described in
+@ref[the introduction](introduction.md#using-akka-http)
+
+Just like any other Akka module Akka HTTP is configured via [Lightbend Config](https://github.com/lightbend/config).
```
johanandren: Isn't the library still called "typesafe config" though?

raboof: Oops! Not sure what I was thinking.

docs/src/main/paradox/configuration.md
```diff
@@ -1,5 +1,8 @@
 # Configuration
 
+Before configuring Akka HTTP it must be added to your project as described in
+@ref[the introduction](introduction.md#using-akka-http)
```
jrudolph:
```suggestion
@ref[the introduction](introduction.md#using-akka-http).
```

raboof:
The rationale is that I wanted to point a user to the docs on adding the dependency, and it took me a few attempts to find it myself - so I wanted to add some more pointers in the right direction. Indeed better headings/subpages might help even more.

jrudolph:
I see, thanks, @raboof.

PR #2445, opened 2019-03-05
Update to latest scalatest 3.1.0-RC1

jrudolph:


project/Dependencies.scala
```diff
@@ -27,7 +27,7 @@ object Dependencies {
     crossScalaVersions := Seq("2.12.8", "2.11.12", "2.13.0-M5"),
     scalaVersion := crossScalaVersions.value.head,
     scalaCheckVersion := System.getProperty("akka.build.scalaCheckVersion", "1.14.0"),
-    scalaTestVersion := "3.0.6-SNAP6",
+    scalaTestVersion := "3.1.0-RC1",
```
raboof: Jumping over 3.0.6 because of accidental binary incompatibility there?

jrudolph: Ah, didn't see that a 3.0.6 final had been released. It was a bit weird that 3.0.6-SNAP6 was followed up by 3.1.0-SNAP7 (see https://gist.github.com/jrudolph/1b054191c235469592f62c7c61cb657d#file-analysis-txt-L197), so I thought 3.0.6 had been abandoned. Let's take that instead.

```suggestion
    scalaTestVersion := "3.0.6",
```

jlprat:
This twitter thread might explain something: https://twitter.com/bvenners/status/1101929144386908160

3.0.6 was binary incompatible and most probably, the one that should be used is 3.1.0

jrudolph:
I see, thanks @jlprat. I don't think binary compatibility matters for us and also just calling the same release with a different version name doesn't change anything anyway... Let's go with 3.0.6 until a final 3.1.0 has been released.

PR #2448, opened 2019-03-05
+htc Add `application/cbor` as predefined MediaType

sirthias:
Tiny addition of a new predefined `MediaType` that is used increasingly often.


sirthias:
Oh, man, there is Java API as well, of course.
This is not the first time that I completely forget about that... :-)
Sure, I'll add a Java version as well...

sirthias:
Ok, Java API added as well now.

jrudolph:
FTR: 
 * the specs: https://tools.ietf.org/html/rfc7049#section-7.3
 * the IANA registration: https://www.iana.org/assignments/media-types/application/cbor

PR #2450, opened 2019-03-06
Improve message parsing

jrudolph:
* Prevent a round-trip of throwing and catching exceptions after having read a full request (default case)
 * Prevent buffering single elements in the output buffer

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpMessageParser.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpMessageParser.scala
@@ -30,7 +30,11 @@ private[http] trait HttpMessageParser[Output >: MessageOutput <: ParserOutput] {
 
   import HttpMessageParser._
 
-  protected final val result = new ListBuffer[Output]
+  // Either:
+  //   - null: currently no output
+  //   - Output: one output element
+  //   - ListBuffer: several output elements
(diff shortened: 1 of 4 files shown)


jrudolph:
Just removing the exception is ~2% improvement.

akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpMessageParser.scala
```diff
@@ -74,25 +78,31 @@ private[http] trait HttpMessageParser[Output >: MessageOutput <: ParserOutput] {
...
   protected final def doPull(): Output =
-    if (result.nonEmpty) {
-      val head = result.head
-      result.remove(0) // faster than `ListBuffer::drop`
-      head
-    } else if (terminated) StreamEnd else NeedMoreData
+    result match {
+      case null ⇒ if (terminated) StreamEnd else NeedMoreData
+      case buffer: ListBuffer[Output] ⇒
+        val head = buffer.head
```
johanandren: What about this head vs indexed access at 0? Looks like the default head allocates two lambdas etc. and couldn't figure out if there was a specialized version somewhere.

jrudolph: The `ListBuffer` variant is already the slow path so I didn't care about optimizing that as well.

PR #2451, opened 2019-03-06
Speedup CharUtils.toLowerCase

jrudolph:
And fix broken (and unused) CharUtils.toUpperCase.

### DIFF
--- a/akka-parsing/src/main/scala/akka/parboiled2/CharUtils.scala
+++ b/akka-parsing/src/main/scala/akka/parboiled2/CharUtils.scala
@@ -170,13 +170,13 @@ object CharUtils {
    * Efficiently lower-cases the given character.
    * Note: only works for 7-bit ASCII letters.
    */
-  def toLowerCase(c: Char): Char = if (c >= 'A' && c <= 'Z') (c + 0x20 /* - 'A' + 'a' */).toChar else c
+  def toLowerCase(c: Char): Char = if (c >= 'A' && c <= 'Z') (c + 0x20 /* - 'A' + 'a' */ ).toChar else c
 
   /**
    * Efficiently upper-cases the given character.
    * Note: only works for 7-bit ASCII letters.
    */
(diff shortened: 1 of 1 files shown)


johanandren:
Any numbers on perf improvement?

jrudolph:
All these little performance improvements are between 0.5%-5% when running `wrk` in standard mode against `akka.http.scaladsl.TestServer`. In that benchmark, already ~30% of time is spent in the kernel.

This particular change is definitely one of the less important ones but it's easy to do and doesn't obfuscate the code. The cost saved here is that `CharPredicate extends (Char => Boolean)` which isn't specialized for some reason so the boolean results are boxed/unboxed.

akka-parsing/src/main/scala/akka/parboiled2/CharUtils.scala
```diff
@@ -170,13 +170,13 @@ object CharUtils {
    * Efficiently lower-cases the given character.
    * Note: only works for 7-bit ASCII letters.
    */
-  def toLowerCase(c: Char): Char = if (CharPredicate.UpperAlpha(c)) (c + 0x20).toChar else c
+  def toLowerCase(c: Char): Char = if (c >= 'A' && c <= 'Z') (c - 'A' + 'a').toChar else c
```
raboof: The `- 'A' + 'a'` is not optimized away in bytecode:
```
      13: bipush        65
      15: isub
      16: bipush        97
      18: iadd
```
`+ 0x20` is slightly less readable but since this is such a small method it might be fine to keep it?

jrudolph: Hah, good catch. I indeed assumed that it would optimize but didn't check. I guess the reason is that the Scala compiler is working on the expression level and there it looks like that:

```scala
(c.-('A'))
  .+('a')
```

where the opportunity for constant folding isn't as obvious. It would have to know that `a - b == a + (-b)` and `(a + b) + c == a + (b + c)` which it doesn't seem to do. Interesting enough,

```scala
- 'A' + 'a' + c
```

also doesn't optimize but


```scala
'a' - 'A' + c
```

again does. :)

PR #2452, opened 2019-03-06
Optimize RequestTimeout support for the common case

jrudolph:


akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -265,7 +265,7 @@ private[http] object HttpServerBluePrint {
           val access = new TimeoutAccessImpl(request, initialTimeout, requestEnd, callback,
             interpreter.materializer, log)
           openTimeouts = openTimeouts.enqueue(access)
-          push(requestOut, request.copy(headers = request.headers :+ `Timeout-Access`(access), entity = entity))
+          push(requestOut, request.copy(headers = `Timeout-Access`(access) +: request.headers, entity = entity))
```
jrudolph: No appending to a List.

akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
```diff
@@ -276,7 +276,7 @@ private[http] object StreamUtils {
    */
   @InternalApi
   object CaptureTerminationOp extends EntityStreamOp[Future[Unit]] {
-    def strictM: Future[Unit] = Future.successful(())
+    val strictM: Future[Unit] = Future.successful(())
```
jrudolph: Seems creation of this already completed Future was somewhat expensive.

PR #2453, opened 2019-03-06
=htc Use static scanning when searching for well-known ContentType for rendering

jrudolph:
Going through the HashMap is slow because calculating a hashCode is
slow compared to doing a series of quick reference equals checks.


jrudolph:
PLS BUILD

PR #2457, opened 2019-03-11
Allow (but ignore) RST and WindowUpdate frames on closed HTTP/2 streams

raboof:
This is explicitly allowed by the spec, and indeed happens in the wild when
the Netty client tries to reset a response stream.

Identified while diagnosing https://github.com/akka/akka-grpc/issues/357

### DIFF
--- a/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
+++ b/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
@@ -131,7 +131,14 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with StageLoggi
     }
   }
   case object Closed extends IncomingStreamState {
-    def handle(event: StreamFrameEvent): IncomingStreamState = receivedUnexpectedFrame(event)
+    def handle(event: StreamFrameEvent): IncomingStreamState = event match {
+      // https://http2.github.io/http2-spec/#StreamStates
+      // Endpoints MUST ignore WINDOW_UPDATE or RST_STREAM frames received in this state,
+      case _: RstStreamFrame | _: WindowUpdateFrame ⇒
+        Closed
+      case _ ⇒
+        receivedUnexpectedFrame(event)
+    }
   }
 
   class IncomingStreamBuffer(streamId: Int, outlet: SubSourceOutlet[ByteString]) extends OutHandler {
--- a/akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
+++ b/akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
@@ -245,6 +245,17 @@ class Http2ServerSpec extends AkkaSpec("""
         val error = entityDataIn.expectError()
         error.getMessage shouldBe "Stream with ID [1] was closed by peer with code INTERNAL_ERROR(0x02)"
       }
+      "not fail the whole connection when one stream is RST twice" in new WaitingForRequestData {
+        sendRST_STREAM(TheStreamId, ErrorCode.STREAM_CLOSED)
+        val error = entityDataIn.expectError()
+        error.getMessage shouldBe "Stream with ID [1] was closed by peer with code STREAM_CLOSED(0x05)"
+        expectNoBytes()
+
+        // https://http2.github.io/http2-spec/#StreamStates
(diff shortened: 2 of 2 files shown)


akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -131,7 +131,14 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with StageLoggi
     }
   }
   case object Closed extends IncomingStreamState {
-    def handle(event: StreamFrameEvent): IncomingStreamState = receivedUnexpectedFrame(event)
+    def handle(event: StreamFrameEvent): IncomingStreamState = event match {
+      // https://http2.github.io/http2-spec/#StreamStates
+      // Endpoints MUST ignore WINDOW_UPDATE or RST_STREAM frames received in this state,
+      case _: RstStreamFrame | _: WindowUpdateFrame ⇒
+        Closed
```
jrudolph:
```suggestion
        this
```

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Compliance.scala
```diff
@@ -44,9 +44,12 @@ private[http2] object Http2Compliance {
   final def requireZeroStreamId(id: Int): Unit =
     if (id != 0) throw new IllegalHttp2StreamIdException(id, "MUST BE == 0.")
 
+  final def requireNonZeroStreamId(id: Int): Unit =
+    if (id == 0) throw new Http2ProtocolException(ErrorCode.PROTOCOL_ERROR, "Stream ID MUST be > 0") // cause GOAWAY
+
   final def requirePositiveWindowUpdateIncrement(streamId: Int, increment: Int): Unit =
     if (increment <= 0)
-      if (streamId == 0) throw new Http2ProtocolException(ErrorCode.PROTOCOL_ERROR, "WINDOW_UPDATE MUST be > 0, was: " + increment) // cause GOAWAY
+      if (streamId == 0) throw new Http2ProtocolException(ErrorCode.PROTOCOL_ERROR, "Stream ID MUST be > 0") // cause GOAWAY
```
jrudolph: Oops, thanks for fixing.

jrudolph: Maybe just use `requireNonZeroStreamId` for that?

raboof: hm actually looking at https://tools.ietf.org/html/rfc7540#section-6.9 a 0 stream id here is fine, removing that check altogether.

jrudolph:
PLS BUILD

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Compliance.scala
```diff
@@ -48,9 +48,7 @@ private[http2] object Http2Compliance {
     if (id == 0) throw new Http2ProtocolException(ErrorCode.PROTOCOL_ERROR, "Stream ID MUST be > 0") // cause GOAWAY
 
   final def requirePositiveWindowUpdateIncrement(streamId: Int, increment: Int): Unit =
-    if (increment <= 0)
-      if (streamId == 0) throw new Http2ProtocolException(ErrorCode.PROTOCOL_ERROR, "Stream ID MUST be > 0") // cause GOAWAY
-      else throw new Http2ProtocolStreamException(streamId, ErrorCode.PROTOCOL_ERROR, "WINDOW_UPDATE MUST be > 0, was: " + increment) // cause RST_STREAM
+    if (increment <= 0) throw new Http2ProtocolStreamException(streamId, ErrorCode.PROTOCOL_ERROR, "WINDOW_UPDATE MUST be > 0, was: " + increment) // cause RST_STREAM
```
jrudolph: Did you intend to remove the check completely? Seems to fail a test now.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Compliance.scala
raboof: I did not look closely enough but the tests saved us: a 0 stream id is fine, but a 0 window increment with stream id 0 should be treated as a connection error where a 0 window increment on another stream should be treated as a stream error. Fixing.

PR #2460, opened 2019-03-12
[WIP] =htc consider request as complete when LastChunk has been seen, fixes #2455

### LINKED ISSUE #2455: Early response warning when ignoring the last chunk in a chunked request
Not sure if this should be considered an Akka HTTP bug or not, or if it's an intrinsic race condition that should instead be avoided by client code.

The following server code:

```scala
 Http().bindAndHandleAsync({ req =>
    req.entity match {
      case HttpEntity.Chunked(ct, chunks) =>
        chunks.takeWhile(!_.isLastChunk).runWith(Sink.fold(0)((count, _) => count + 1))
          .map(count => HttpResponse(entity = HttpEntity(s"Got chunked request with $count chunks")))
    }
  }, "localhost", 8080)
```

looks fair enough, it's taking all the chunks up to the last chunk, and then counting them, and telling the client how many chunks there were.

So, now in the client we invoke that with something like this:

```scala
  Http().singleRequest(
        HttpRequest(
          uri = s"http://localhost:$port",
          method = HttpMethods.POST,
          entity = HttpEntity.Chunked(ContentTypes.`text/plain(UTF-8)`, Source.single(ChunkStreamPart("a")))
        )
      )
```

And everything works just fine, except for this warning in the logs:

```
[WARN] [03/08/2019 16:04:17.480] [default-akka.actor.default-dispatcher-12] [akka.actor.ActorSystemImpl(default)] Sending an 2xx 'early' response before end of request was received... Note that the connection will be closed after this response. Also, many clients will not read early responses! Consider only issuing this response after the request data has been completely read!
```

It's a bit odd that the last chunk has been received, and so presumably the very next thing that the HTTP request body is going to do is terminate the stream, but somehow, my asynchronously provided result gets there first, and is considered to have been emitted before the request body stream has been consumed.

jrudolph:
Currently broken. It seems there are even more unfulfilled assumptions somewhere.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
@@ -34,6 +34,7 @@ import akka.http.javadsl.model
 import akka.http.scaladsl.model._
 import akka.http.scaladsl.model.ws.Message
 import akka.http.impl.util.LogByteStringTools._
+import akka.http.scaladsl.model.HttpEntity.LastChunk
 
 /**
  * INTERNAL API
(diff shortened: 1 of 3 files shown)


jrudolph:
Closing for now, there is a workable workaround available right now, see #2455.

PR #2464, opened 2019-03-13
Support HTTP2 in cleartext (h2c) via Upgrade from HTTP1

raboof:
Continuation of #2085, refs #1966

This allows HTTP/1.1 and HTTP/2 to coexist on the same insecure port.
This feature is not widely supported among clients: browsers
generally do not support h2c at all, curl and nghttp somewhat support
it, grpc-java only since [fairly recently](https://github.com/grpc/grpc-java/pull/4612)


raboof:
Rebased on top of master. Needs work: the test fails because it seems somehow the demand from the sink does not properly propagate to the protocol switch.

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -654,15 +654,7 @@ private[http] object HttpServerBluePrint {
           f()
       }
 
-      /*
-       * WebSocket support
-       */
-      def switchToWebSocket(handlerFlow: Either[Graph[FlowShape[FrameEvent, FrameEvent], Any], Graph[FlowShape[Message, Message], Any]]): Unit = {
```
jrudolph: Good cleanup here. It was a bit weird anyway that part of the WebSocket flow creation was done inside here.

akka-http-core/src/main/scala/akka/http/impl/engine/ws/Handshake.scala
```diff
@@ -144,15 +147,22 @@ private[http] object Handshake {
           concatenated value to obtain a 20-byte value and base64-
           encoding (see Section 4 of [RFC4648]) this 20-byte hash.
     */
-    def buildResponse(key: `Sec-WebSocket-Key`, handler: Either[Graph[FlowShape[FrameEvent, FrameEvent], Any], Graph[FlowShape[Message, Message], Any]], subprotocol: Option[String]): HttpResponse =
+    def buildResponse(key: `Sec-WebSocket-Key`, handler: Either[Graph[FlowShape[FrameEvent, FrameEvent], Any], Graph[FlowShape[Message, Message], Any]], subprotocol: Option[String], settings: WebSocketSettings, log: LoggingAdapter): HttpResponse = {
+      val frameHandler = handler match {
```
jrudolph: That's good for now. I guess we should move those flow creation things to `WebSocket` at some point.

akka-http-core/src/main/scala/akka/http/impl/model/parser/Base64Parsing.scala
```diff
@@ -81,6 +81,10 @@ private[http] object Base64Parsing {
   val rfc2045BlockDecoder: Decoder = decodeBlock(Base64.rfc2045())
   val customBlockDecoder: Decoder = decodeBlock(Base64.custom())
 
+  private val base64url = new Base64("ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-_=")
```
jrudolph: Great, yet another base64 variant ;)

akka-http-core/src/main/scala/akka/http/scaladsl/ConnectionContext.scala
```diff
@@ -12,7 +12,7 @@ import scala.collection.JavaConverters._
 import java.util.{ Optional, Collection => JCollection }
 
 import akka.http.javadsl
-import akka.http.scaladsl.UseHttp2.Negotiated
+import akka.http.scaladsl.UseHttp2.{ Negotiated, Never }
```
jrudolph: Extra import.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -198,6 +205,29 @@ private[http2] class Http2ServerDemux(http2Settings: Http2ServerSettings) extend
           multiplexer.registerSubStream(sub)
         }
       })
+
+      private def applySettings(settings: immutable.Seq[Setting]): Boolean = {
```
jrudolph: Ok :+1:

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/framing/Http2FrameParsing.scala
```diff
@@ -12,13 +12,33 @@ import akka.stream.stage.GraphStageLogic
...
-
 import FrameEvent._
 
+import scala.annotation.tailrec
+
+/** INTERNAL API */
+@InternalApi
+private[http] object Http2FrameParsing {
+
+  def readSettings(payload: ByteStringParser.ByteReader): immutable.Seq[Setting] = {
```
jrudolph: :+1:

akka-http2-support/src/test/scala/akka/http/scaladsl/Http2ServerTest.scala
```diff
@@ -22,11 +22,15 @@ import scala.concurrent.duration._
...
+/**
+ * App to manually test an HTTP2 server
+ */
 object Http2ServerTest extends App {
   val testConf: Config = ConfigFactory.parseString("""
-    akka.loglevel = INFO
+    akka.loglevel = debug
     akka.log-dead-letters = off
     akka.stream.materializer.debug.fuzzing-mode = off
+    akka.http.server.preview.enable-http2 = on
```
jrudolph: That's already defined a few lines below.

docs/src/main/paradox/server-side/http2.md
```diff
@@ -45,17 +45,41 @@ Note that `bindAndHandle` currently does not support HTTP/2, you must use `bindA
...
-We currently only support the approach requiring [Prior Knowledge](https://httpwg.org/specs/rfc7540.html#known-http). It is automatically enabled when HTTP/2 is enabled, on the same port:
+We support both approaches transparently on the same port. This feature is automatically enabled when HTTP/2 is enabled:
 
 Scala
 :   @@snip[Http2Spec.scala]($test$/scala/docs/http/scaladsl/Http2Spec.scala) { #bindAndHandlePlain }
 
 Java
 :   @@snip[Http2Test.java]($test$/java/docs/http/javadsl/Http2Test.java) { #bindAndHandlePlain }
 
+#### h2c Upgrade
```
jrudolph: :+1:

docs/src/test/scala/docs/http/scaladsl/Http2Spec.scala
```diff
@@ -24,6 +24,11 @@ import akka.http.scaladsl.HttpConnectionContext
 
 //#bindAndHandlePlain
 
+//#bindAndHandleNegotiateUpgrade
+import akka.http.scaladsl.UseHttp2.Negotiated
```
jrudolph: Still needed?

akka-http2-support/src/main/scala/akka/http/scaladsl/Http2.scala
```diff
@@ -84,6 +88,65 @@ final class Http2Ext(private val config: Config)(implicit val system: ActorSyste
...
+            log.warning("Failed to parse http2-settings header in upgrade [{}], continuing with HTTP/1.1", e.getMessage)
+            handler(req)
+          // A server MUST NOT upgrade the connection to HTTP/2 if this header field
+          // is not present or if more than one is present
+          case _ =>
+            log.debug("Invalid upgrade request (http2-settings header missing or repeated)")
+            handler(req)
+        }
+
+      case _ =>
```
jrudolph: Ok, so the extra overhead for HTTP/1.1 when HTTP/2 is enabled is now checking that header here. I guess that's ok.

akka-http2-support/src/main/scala/akka/http/scaladsl/Http2.scala
```diff
@@ -84,6 +88,65 @@ final class Http2Ext(private val config: Config)(implicit val system: ActorSyste
...
+          // Must be exactly one
+          case immutable.Seq(Success(settingsFromHeader)) =>
+            // inject the actual upgrade request with a stream identifier of 1
+            // https://http2.github.io/http2-spec/#rfc.section.3.2
+            val injectedRequest = Source.single(req.addHeader(Http2StreamIdHeader(1)))
+
+            val serverLayer: Flow[ByteString, ByteString, Future[Done]] = Flow.fromGraph(
+              Flow[HttpRequest]
+                .watchTermination()(Keep.right)
+                .merge(injectedRequest)
```
jrudolph: Should it be `prepend` instead? On the other hand, the order shouldn't matter, so `merge` should be fine as well.

akka-http-core/src/main/scala/akka/http/impl/engine/ws/UpgradeToWebSocketLowLevel.scala
```diff
@@ -5,6 +5,7 @@
 package akka.http.impl.engine.ws
 
 import akka.annotation.InternalApi
+import akka.http.impl.engine.server.InternalCustomHeader
```
jrudolph: Extra import?

raboof: This one is needed because `akka.http.impl.engine.ws.InternalCustomHeader` was promoted to `akka.http.impl.engine.server.InternalCustomHeader`

jrudolph: Ah, I see. Thanks.

akka-http2-support/src/main/scala/akka/http/scaladsl/Http2.scala
```diff
@@ -84,6 +88,65 @@ final class Http2Ext(private val config: Config)(implicit val system: ActorSyste
...
   }
 
+  private def handleUpgradeRequests(
+    handler:     HttpRequest => Future[HttpResponse],
+    settings:    ServerSettings,
+    parallelism: Int,
+    log:         LoggingAdapter
+  ): HttpRequest => Future[HttpResponse] = { req =>
+    req.header[Upgrade] match {
+      case Some(upgrade) if upgrade.protocols.exists(_.name equalsIgnoreCase "h2c") =>
```
johanandren: Feels like an if-else rather than a patmatch with catchall to me?

`if (req.header[Upgrade].exists(_.name.equalsIgnoreCase("h2x"))) {} else {}`

Maybe worth also caching that lambda since it is evaluated for each request?

raboof: it's a little more tricky because the `exists` is on the list of protocols, not on the option... good point about the lambda though, will try to rewrite

raboof: Hmm actually I struggle coming up with a solution that is meaningfully better, could you suggest something?

johanandren: Nah, leave it be in that case.

Issue #2462, opened 2019-03-13
Memory leak with Jetty ALPN support


jrudolph:
With Jetty ALPN, you need to put your ALPN handler inside a global map keyed by the SSLEngine. After our handler is called we remove it from that map again:

https://github.com/akka/akka-http/blob/392742107e3ba98dbeb0d9fca3f9c1a36ef85ebd/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2AlpnSupport.scala#L64-L67

When the ALPN handler is never called, though, e.g. because the TLS handshake fails or the connection is closed before the ALPN handshake takes place, we leak the SSLEngine and our handler by keeping it in the global map. Therefore, we somehow have to clean up the ALPN map when a connection is torn down.

Not sure, how we would do it, though, because it's hard to get a handle to the `SSLEngine` after a connection has been closed. Or maybe run a regular clean up task, that runs over all entries and removes the ones where the SSLEngine is already finished?

BrijJhala:
Hi @jrudolph . thank you for pointing this fix. is it same issue as shared below in screen shot ? We are using hive cortex https://github.com/TheHive-Project/Cortex and getting OOM and heap analysis points same class having leak
![image](https://user-images.githubusercontent.com/4096804/95140525-1ca66b00-073d-11eb-938b-0d2bf938a2d5.png)

![image](https://user-images.githubusercontent.com/4096804/95140408-d05b2b00-073c-11eb-83c7-c2ec36697a19.png)
![image](https://user-images.githubusercontent.com/4096804/95140420-dbae5680-073c-11eb-8071-256842f128bb.png)

jrudolph:
Yes, seems similar.

PR #2468, opened 2019-03-13
WIP Postpone ALPN registration until handshake is about to begin

ignasi35:
related #2462 
Produces a Wrapper class around the provided `SSLEngine` so that registering the instance on Jetty's ALPN `Map` only happens immediately before invoking `beginHandshake`. Also, if handshake fails (causing the ALPN negotiation to not complete) the instance is removed from the `Map`.

I am not familiar enough with the lifecycle of a connection/handshake/etc... required by  the `SSLEngine` so I suspect this approach only reduces the chances of a memory leak but doesn't fully prevents the leak. That is, I'm not sure if even with the changes from this PR, there's still a window where the leak could happen: the `beginHandshake` method completes successfully, but the actual handshake doesn't reach the point where `choose(protocol)` is invoked.

 - - - - - - - 

According to [the docs](https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/SSLEngine.html#beginHandshake--),  `beginHandshake` is not needed:

> This method [`beginHandshake`] is not needed for the initial handshake, as the `wrap()` and `unwrap()` methods will implicitly call this method if handshaking has not already begun.

so this solution is brittle for general use but it'll probably work because of how/when [`TLSActor`](https://github.com/akka/akka/blob/master/akka-stream/src/main/scala/akka/stream/impl/io/TLSActor.scala#L162) invokes `beginHandshake` for the initial handshake.

### DIFF
--- a/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2AlpnSupport.scala
+++ b/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2AlpnSupport.scala
@@ -4,10 +4,11 @@
 
 package akka.http.impl.engine.http2
 
+import java.nio.ByteBuffer
 import java.util.function.BiFunction
 import java.{ util ⇒ ju }
 
-import javax.net.ssl.{ SSLEngine, SSLParameters }
+import javax.net.ssl._
 import akka.annotation.InternalApi
 import akka.http.impl.util.JavaVersion
 import akka.stream.TLSClientAuth
@@ -52,7 +53,7 @@ private[http] object Http2AlpnSupport {
   }
 
   def jettyAlpnSupport(engine: SSLEngine, setChosenProtocol: String ⇒ Unit): SSLEngine = {
-    ALPN.put(engine, new ServerProvider {
+    val serverProvider: ServerProvider = new ServerProvider {
       override def select(protocols: ju.List[String]): String =
         choose {
           chooseProtocol(protocols)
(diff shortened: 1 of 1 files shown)


ignasi35:
I think this solution is incomplete and we should also track if the invocation to `beginHandshake` is for the initial handshake or a [renegotiation](https://github.com/akka/akka/blob/master/akka-stream/src/main/scala/akka/stream/impl/io/TLSActor.scala#L171).

jrudolph:
Interesting approach. I wonder if we should also try the alternative to add a stage next to the TLS stage that catches when the connection is closed and then removes it, might be even simpler, wdyt?

jrudolph:
> Are we sure there is no 'gap' between `SSLEngine.beginHandshake` and `ServerProvider.select` where failures may still cause leaks?

Yes, that's probably not going to work. We call `beginHandshake` in the constructor of `TLSActor`, but that just gives a signal and won't fail right there. Instead, it will usually only fail after a bit of back-and-forth with the peer.

raboof:
I tried that approach in #2471 - that indeed seems a bit safer, though also rather inelegant.

ignasi35:
> Interesting approach. I wonder if we should also try the alternative to add a stage next to the TLS stage that catches when the connection is closed and then removes it, might be even simpler, wdyt?

I'm not familiar enough with the composition of stages collaborating.

I tried this approach because it maintains the knowledge of what ALPN implementation is used inside `Http2AlpnSupport`. The idea to create a new stage in front of TLS stage would require the branching (`if (isAlpnSupportedByJDK) ... else ... `), right?

ignasi35:
#2471 looks cleaner. I'll close this

PR #2470, opened 2019-03-15
Akka-paradox theme 0.18

ennru:


raboof:
(change is just the CDATA thing so safe to merge)

PR #2471, opened 2019-03-15
Clean up ALPN when the stream is stopped

raboof:
Refs #2462

Manually tested by running Http2ServerTest, doing a
`curl http://localhost:9001` (speaking HTTP to the HTTPS port)
and inspecting the Jetty ALPN map with a debugger.

### DIFF
--- a/akka-http2-support/src/main/scala/akka/http/scaladsl/Http2.scala
+++ b/akka-http2-support/src/main/scala/akka/http/scaladsl/Http2.scala
@@ -16,7 +16,8 @@ import akka.http.scaladsl.model.{ HttpRequest, HttpResponse }
 import akka.http.scaladsl.settings.ServerSettings
 import akka.stream.TLSProtocol.{ SendBytes, SessionBytes, SslTlsInbound, SslTlsOutbound }
 import akka.stream.scaladsl.{ BidiFlow, Flow, Keep, Sink, TLS, Tcp }
-import akka.stream.{ IgnoreComplete, Materializer }
+import akka.stream.stage.{ GraphStage, GraphStageLogic, InHandler, OutHandler }
+import akka.stream.{ Attributes, BidiShape, IgnoreComplete, Inlet, Materializer, Outlet }
 import akka.util.ByteString
 import akka.{ Done, NotUsed }
 import com.typesafe.config.Config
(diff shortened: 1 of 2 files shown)


akka-http2-support/src/main/scala/akka/http/scaladsl/Http2.scala
```diff
@@ -139,16 +139,31 @@ final class Http2Ext(private val config: Config)(implicit val system: ActorSyste
...
         engine.setUseClientMode(false)
         Http2AlpnSupport.applySessionParameters(engine, httpsContext.firstSession)
         Http2AlpnSupport.enableForServer(engine, setChosenProtocol)
       }
       val tls = TLS(() ⇒ createEngine, _ ⇒ Success(()), IgnoreComplete)
 
+      val removeEngineOnTerminate: BidiFlow[ByteString, ByteString, ByteString, ByteString, NotUsed] = {
+        implicit val ec = fm.executionContext
+        BidiFlow.fromFlows(
+          Flow.fromFunction(identity),
```
jrudolph:
```suggestion
          Flow[ByteString]
```

:) Also below?

akka-http2-support/src/main/scala/akka/http/scaladsl/Http2.scala
```diff
@@ -139,16 +139,31 @@ final class Http2Ext(private val config: Config)(implicit val system: ActorSyste
...
+          Flow
+            .fromFunction[ByteString, ByteString](identity)
+            .watchTermination()((n, fd) ⇒ {
+              fd.onComplete(_ ⇒ eng.foreach(Http2AlpnSupport.cleanupForServer))
+              n
```
jrudolph: Is it even needed to return a new mat value? Or maybe `Unit` is just fine.

ignasi35:
If it's the `TLS` stage that's [responsible for creating the `SSLEngine`](https://github.com/akka/akka-http/blob/8f95ab0e4e7c50fdc43833b70b10eec840d98553/akka-http2-support/src/main/scala/akka/http/scaladsl/Http2.scala#L148) instance, shouldn't it be the one responsible for the cleanup?

This approach works but requires adding a very thin stage that seems a bit out of place.

Adding a `cleanup: () => Try[Unit]` argument in `TLS` and implementing the thin stage there would keep things in place.

ignasi35:
But, yeah, adding an argument to TLS isn't great (new release of `akka/akka`, more API,...).

I'm fine with this solution

PR #2475, opened 2019-03-19
Update ScalaTest to a binary compatible with 3.0.X series version

jlprat:
Update to a 3.0.x series binary compatible version.

From the release notes:
> This release has the same functionality as 3.0.6, and should be source compatible with 3.0.6, but unlike 3.0.6, is binary compatible with 3.0.5. Please see the release notes for 3.0.6 to find out the details of what the release contains. The goal with this release it to achieve binary compatibility with 3.0.5. Unfortunately although 3.0.6 should be source compatible with 3.0.5, it was not binary compatible.

https://github.com/scalatest/scalatest/releases/tag/release-3.0.7

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -27,7 +27,7 @@ object Dependencies {
     crossScalaVersions := Seq("2.12.8", "2.11.12", "2.13.0-M5"),
     scalaVersion := crossScalaVersions.value.head,
     scalaCheckVersion := System.getProperty("akka.build.scalaCheckVersion", "1.14.0"),
-    scalaTestVersion := "3.0.6",
+    scalaTestVersion := "3.0.7",
     specs2Version := "4.3.6"
   )
 


jlprat:
mmm... forget that is strange... It doesn't compile...
I would thought this is a compatible one...

raboof:
Perhaps we need to revert some/all of 11162fc533f7f077d87192e59e9a76362dcaead6?

jlprat:
Yes, I'm on it

jlprat:
Seems like something is fishy with the version...
`org.scalatest.junit.JUnitSuite` is deprecated in favor (in theory) of a class in `org.scalatestplus.junit` but doesn't seem to be accessible.

I'm asking in the Scalatest gitter chat: https://gitter.im/scalatest/scalatest?at=5c90c3468126720abc097ac4

jlprat:
Created an issue in ScalaTest, let's see: https://github.com/scalatest/scalatest/issues/1553

raboof:
I _suspect_ they just accidentally left the deprecation there while rolling back adding the new API.. but indeed let's see

jlprat:
That might well be, let's see :+1:

jlprat:
Reply from the ScalaTest people:
> it seems like our type alias approach for classes in the new package does not work for java source

They'll discuss how to solve it for next versions.

So, I guess we are good to go with the deprecated class

PR #2485, opened 2019-03-26
workaround for https://github.com/ronmamo/reflections/issues/169

raboof:
This problem is not typically visible, but might be when using a proxy that has
the problematic version of 'reflections' cached


raboof:
Nevermind

PR #2486, opened 2019-03-26
Fix link

ignasi35:


raboof:
@jlprat I'll edit the page manually

raboof:
(we typically treat the generated doc pages as immutable, but recently started tracking them in git so we can post-process old docs as well, so we can do some SEO improvements across the board)

Issue #2179, opened 2018-08-27
Switch to sbt-paradox-apidoc


raboof:
Instead of duplicating code between here akka-http and akka

PR #2490, opened 2019-03-27
Switch to sbt-paradox-apidoc

raboof:
Refs #2179


raboof:
:laughing: I actually generated the paradox docs both with the old and with the new version and verified there are no differences with 'meld' (except for a fix that I snuck in there :D )

PR #2492, opened 2019-03-27
Add step to update the Akka Http version on akka.io/docs

mikela:
Hi, I believe the step to update the Akka Http version on https://akka.io/docs/ is missing so I just added it.


raboof:
OK TO TEST

raboof:
Actually it seems to be already there on line 33, but indeed for 10.1.8 I missed it - maybe make that its own bullet?

mikela:
You are absolutely right! Sounds good :+1:

mikela:
Feel free to edit all you deem necessary of course! I just "denormalized" the bullet point :laughing:

PR #2494, opened 2019-03-28
Remove link to internal class

raboof:
Since it is not in the API docs anyway


jrudolph:
This wasn't merged. Was that intentional?

raboof:
Hmm, I don't remember - still looks valid to me ;)

raboof:
Ah looks like it was already changed in another PR

PR #2496, opened 2019-03-28
Upgrade caffeine to 2.7

### LINKED ISSUE #2443: Upgrade Caffeine to 2.7
[LfuCache](https://github.com/akka/akka-http/blob/8f95ab0e4e7c50fdc43833b70b10eec840d98553/akka-http-caching/src/main/scala/akka/http/caching/LfuCache.scala) currently uses a dummy loader to comply with the API and uses only the manual loading methods. Based on this and similar feedback, we now offer `AsyncCache` to avoid this workaround.

Another benefit is that the eviction policy is now adaptive. It still initially biases towards frequency, but will reconfigure itself based on the workload characteristics. I don't know the typical characteristics for an http cache, but it may be beneficial. I believe there is still some fine tuning needed for tiny caches (e.g. less than 100) that I'll look into for a patch release.

jiminhsieh:
## Purpose
Fix #2443


raboof:
OK TO TEST

PR #2502, opened 2019-04-03
use scalaVersion from System property akka.build.scalaVersion

patriknw:
* if it's defined, otherwise crossScalaVersions.head as before
* this is how the scalaVersion can be set for akka/akka and
  it's used from Jenkins/Travis jobs
* Scala 2.11 is not included in akka/akka crossScalaVersions any
  more because of trouble of mixing different such for different
  modules (excluding it only for -typed modules)
* need this for the akka-http-nightly build matrix

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -25,7 +25,7 @@ object Dependencies {
 
   val Versions = Seq(
     crossScalaVersions := Seq("2.12.8", "2.11.12", "2.13.0-M5"),
-    scalaVersion := crossScalaVersions.value.head,
+    scalaVersion := System.getProperty("akka.build.scalaVersion", crossScalaVersions.value.head),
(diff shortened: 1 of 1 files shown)


raboof:
I added `++${SCALA_VERSION}` to the nightly build config and that seems to have fixed it - are you sure we need this change?

patriknw:
Ok, that was what the job had when it failed for 2.11 in https://jenkins.akka.io:8498/job/akka-http-nightly/439/, but then it didn't have `-Dakka.build.scalaVersion=${SCALA_VERSION}`. Now it has both `-Dakka.build.scalaVersion=${SCALA_VERSION}` and ++${SCALA_VERSION}. One for selecting the version in akka-http, and the other for selecting in akka. Slightly confusing to have two ways but works for me.

PR #2503, opened 2019-04-05
Add support for unrecognized status codes

apavlovics:
## Purpose

Akka HTTP seems to be in violation of [HTTP/1.1 spec](https://tools.ietf.org/html/rfc7231#section-6) by throwing `ParsingException` on unrecognized HTTP status codes instead of treating them equivalent to the x00 status code of their class. This is how HTTP/1.1 spec describes it:

> HTTP status codes are extensible. HTTP clients are not required to understand the meaning of all registered status codes, though such understanding is obviously desirable. However, a client MUST understand the class of any status code, as indicated by the first digit, and treat an unrecognized status code as being equivalent to the x00 status code of that class, with the exception that a recipient MUST NOT cache a response with an unrecognized status code.
>
> For example, if an unrecognized status code of 471 is received by a client, the client can assume that there was something wrong with its request and treat the response as if it had received a 400 (Bad Request) status code. The response message will usually contain a representation that explains the status.

This PR aligns Akka HTTP with HTTP/1.1 spec and adds support for any HTTP status codes in the standard range of 100 to 599. Custom status codes outside of this range are still not supported (unless explicitly registered via [the existing feature](https://doc.akka.io/docs/akka-http/current/common/http-model.html#registering-custom-status-codes)), as there is no x00 status code equivalent for them. Backwards compatibility is kept, as any registered custom status code is used, if present, before failing over to an unrecognized one.

## References

Have not found any related references.

## Changes

* Adds support for unrecognized HTTP status codes in the standard range of 100 to 599.
* Adds corresponding unit test to verify the feature.

## Background Context

The main motivation for this change is to make Akka HTTP compatible with HTTP/1.1 spec in the area of unrecognized HTTP status codes and make life of Akka HTTP users easier by not having to register custom HTTP status codes in the standard range of 100 to 599, unless truly necessary.


raboof:
OK TO TEST

akka-http-core/src/test/scala/akka/http/impl/engine/parsing/ResponseParserSpec.scala
```diff
@@ -85,6 +86,17 @@ abstract class ResponseParserSpec(mode: String, newLine: String) extends FreeSpe
...
       }
 
+      "a response with an unrecognized status code" in new Test {
+        // A client must understand the class of any status code, as indicated by the first digit, and
+        // treat an unrecognized status code as being equivalent to the x00 status code of that class
+        // https://tools.ietf.org/html/rfc7231#section-6
+        """HTTP/1.1 456 Totally unrecognized
+          |Content-Length: 0
+          |
+          |""" should parseTo(HttpResponse(TotallyUnrecognized))
```
raboof: So it creates a custom status code 'on demand' when it encounters an unknown status code. Seems reasonable! (though not obviously the same as "treat an unrecognized status code as being equivalent to the x00 status code of that class")

apavlovics: I agree, the spec is not 100% clear on what "treat an unrecognized status code as being equivalent to the x00 status code of that class" means. But in Akka HTTP context it seems reasonable to think that this means that an unrecognized status code should have the same `isSuccess`, `isFailure`, `isRedirection` values as other statuses from its class. And that it should not prevent further processing of the response.

raboof: Agreed!

apavlovics: Are there any actions planned for this PR? One of the production projects in my company would benefit if these changes are approved and end up in Akka HTTP release.

akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpResponseParser.scala
```diff
@@ -70,7 +71,17 @@ private[http] class HttpResponseParser(protected val settings: ParserSettings, p
...
-          case None    ⇒ customStatusCodes(code) getOrElse badStatusCodeSpecific(code)
+          case None ⇒ customStatusCodes(code) getOrElse {
+            // A client must understand the class of any status code, as indicated by the first digit, and
+            // treat an unrecognized status code as being equivalent to the x00 status code of that class
+            // https://tools.ietf.org/html/rfc7231#section-6
+            try {
+              val reason = asciiString(input, reasonStartIdx, reasonEndIdx)
+              StatusCodes.custom(code, reason)
+            } catch {
+              case _: Exception ⇒ badStatusCodeSpecific(code)
```
jrudolph: What kind of exceptions do we expect here?

akka-http-core/src/test/scala/akka/http/impl/engine/parsing/ResponseParserSpec.scala
jrudolph: > x00 status code of that class

I interpret that as that status codes should be rounded to hundreds if they are not known. But I think this approach here is fine as well -- after all, it's only the parser, so you still need to interpret it.

akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpResponseParser.scala
apavlovics: `IllegalArgumentException` will be thrown by `StatusCodes.custom()` if the parsed `code` is _not_ between 100 and 599. Maybe the code is overly protective here and should be changed to expect only this particular exception, not sure about that.

jrudolph: In that case, maybe it would make sense to refactor it so that the `code` is checked explicitly before calling custom?

apavlovics: Sure, can be done. My motivation to implement it this way was not to duplicate logic inside `StatusCodes.custom()` to `HttpResponseParser` (e.g. if we do that, `code` value will be checked twice, in both places). If this is not a concern, I can adjust.

jrudolph: Ok, sounds good, thanks for the explanation!

jrudolph:
PLS BUILD

lightbend-cla-validator:

jrudolph:
> At least one pull request committer is not linked to a user. See [help.github.com/en/articles/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user](https://help.github.com/en/articles/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user)

@apavlovics could you configure github so that your email is recognized? That would placate our CLA checker (as you have signed the CLA already).

jrudolph:
Checked @apavlovics's CLA signature manually.

akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpResponseParser.scala
```diff
@@ -70,7 +71,17 @@ private[http] class HttpResponseParser(protected val settings: ParserSettings, p
...
-          case None    => customStatusCodes(code) getOrElse badStatusCodeSpecific(code)
+          case None => customStatusCodes(code) getOrElse {
+            // A client must understand the class of any status code, as indicated by the first digit, and
+            // treat an unrecognized status code as being equivalent to the x00 status code of that class
+            // https://tools.ietf.org/html/rfc7231#section-6
+            try {
+              val reason = asciiString(input, reasonStartIdx, reasonEndIdx)
+              StatusCodes.custom(code, reason)
+            } catch {
+              case _: Exception => badStatusCodeSpecific(code)
```
rahilb: should this not be `case NonFatal(_) =>`?

raboof: Thanks for the extra set of eyes! That indeed seems reasonable, would you be interested in doing a follow-up PR?

Issue #2487, opened 2019-03-26
MissingHeaderRejection construction differs when using `headerValueByType`


mfulgo:
Using the `headerValueByType` directive produces a `MissingHeaderRejection` that uses the name of a custom header class, rather than the header name. This behavior differs from when one uses `headerValueByName`.

While both directives reference the header differently, the `MissingHeaderRejection`'s field is `headerName`, which implies to me that it should be the _name_ of the header, not its class or companion object's name. It makes writing tests a bit awkward...

In the following spec, the second test fails.
```scala
import akka.http.scaladsl.model.headers.{ModeledCustomHeader, ModeledCustomHeaderCompanion}
import akka.http.scaladsl.server.Directives._
import akka.http.scaladsl.server.{Directive0, MissingHeaderRejection}
import akka.http.scaladsl.testkit.ScalatestRouteTest
import org.scalatest.{Matchers, OptionValues, WordSpec}

import scala.util.Try

class AkkaHttpMissingHeaderTest
  extends WordSpec
    with Matchers
    with OptionValues
    with ScalatestRouteTest
{
  private val reqByName: Directive0 = headerValueByName(CustomHeader.name)
    .flatMap[Unit](_ => pass)
  private val reqByType: Directive0 = headerValueByType[CustomHeader]()
    .flatMap[Unit](_ => pass)

  private val route =
    (path("byName") & reqByName) { complete("Success") } ~
    (path("byType") & reqByType) { complete("Success") }

  "MissingHeaderRejection" should {
    "contain the header name" when {
      "used by headerValueByName" in {
        Get("/byName") ~>
          route ~> check {
          rejection shouldBe MissingHeaderRejection(CustomHeader.name)
        }
      }
      "used by headerValueByType" in {
        Get("/byType") ~>
          route ~> check {
          rejection shouldBe MissingHeaderRejection(CustomHeader.name)
          // must use MissingHeaderRejection("CustomHeader") for this to work
        }
      }
    }
  }

  final class CustomHeader(content: String)
    extends ModeledCustomHeader[CustomHeader] {

    override def companion: ModeledCustomHeaderCompanion[CustomHeader] = CustomHeader
    override def value(): String = content
    override def renderInRequests(): Boolean = true
    override def renderInResponses(): Boolean = false
  }

  final object CustomHeader
    extends ModeledCustomHeaderCompanion[CustomHeader] {
    override def name: String = "x-custom"
    override def parse(value: String): Try[CustomHeader] =
[...]

PR #2504, opened 2019-04-05
Fix missing header rejection for custom headers #2487

mfulgo:
## Purpose
When the HeaderMagnet is formed for a user-defined custom header, it
used the class' name to populate the `headerName` field. However, the
user may have named their class differently than its name, by excluding
dashes, for example.

This could produce an inconsistency in behavior when routes would
generate a `MissingHeaderRejection` for a custom header because the
`headerValueByType` directive constructs a `MissingHeaderRejection` with
a `headerName` populated from the magnet, which in the case of a custom
header, comes from the header's class name rather than its name field.

## References
References #2487 

## Changes
* Updates the `HeaderMagnet` implicit for `ModeledCustomHeader`s to set the `headerName` from the companion's `name` field.

## Background Context
* See #2487 for background and an example of the undesired behavior.
* This approach was taken because it's a focused, one-line change that should have a fairly limited effect. See 1bcc63e for an alternative approach that was considered.


jrudolph:
OK TO TEST

mfulgo:
@jrudolph @johanandren - I'm not sure what the code review process or timeline is, but this should be ready for one.

johanandren:
Thanks @mfulgo, we have been traveling for team meetings so please bear with us waiting for reviews.

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/HeaderDirectivesSpec.scala
```diff
@@ -202,4 +206,30 @@ class HeaderDirectivesSpec extends RoutingSpec with Inside {
       }
     }
   }
+
+  "The HeaderMagnet" should {
```
jrudolph: HeaderMagnet is only an implementation detail, so not sure if we need exactly those tests. Could you additionally (or instead) create one or more tests, e.g. above for `headerValueForType` that asserts whether the rejection is returned similar to the test you changed in the ModeledCustomHeaderSpec?

akka-http/src/main/scala/akka/http/scaladsl/server/directives/HeaderDirectives.scala
```diff
@@ -199,7 +200,9 @@ trait LowPriorityHeaderMagnetImplicits {
     new HeaderMagnet[T] {
       override def classTag: ClassTag[T] = ClassTag(clazz)
       override def runtimeClass: Class[T] = clazz
-      override def extractPF: PartialFunction[HttpHeader, T] = { case x if runtimeClass.isAssignableFrom(x.getClass) ⇒ x.asInstanceOf[T] }
+      override def extractPF: PartialFunction[HttpHeader, T] = {
```
jrudolph: Let's keep the original formatting here and below.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/HeaderDirectives.scala
```diff
@@ -181,11 +181,12 @@ object HeaderMagnet extends LowPriorityHeaderMagnetImplicits {
...
     new HeaderMagnet[T] {
-      override def runtimeClass = tag.runtimeClass.asInstanceOf[Class[T]]
-      override def classTag = tag
-      override def extractPF = {
+      override def classTag: ClassTag[T] = tag
+      override def runtimeClass: Class[T] = tag.runtimeClass.asInstanceOf[Class[T]]
+      override def extractPF: PartialFunction[HttpHeader, T] = {
         case h if h.is(companion.lowercaseName) ⇒ companion.apply(h.value)
       }
+      override def headerName: String = companion.name
```
jrudolph: :+1:

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/HeaderDirectivesSpec.scala
mfulgo: Yeah. That should be easy enough. I'll do that when I get some free time, maybe later this week.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/HeaderDirectives.scala
mfulgo: Line 187 was on a second line, so the intention here was to give it some consistency across the class and hopefully make it a little more readable. Let me know if it's still not what you want.

mfulgo:
@jrudolph Added the test you asked for. Let me know about the other comment.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/HeaderDirectives.scala
jrudolph: Ah, I see. Sounds good, then.

PR #2505, opened 2019-04-10
Add String to UUID unmarshaller to the predefined Scala unmarshallers

maximskripnik:
## Purpose

Add ability to unmarshall Strings to UUIDs in Scala code via predefined implicit `Unmarshaller[String, UUID]`

## Background Context

So we have followings things already:
* [String -> UUID Java Unmarshaller](https://github.com/akka/akka-http/blob/master/akka-http/src/main/java/akka/http/javadsl/unmarshalling/StringUnmarshallers.java#L73)
* [UUID path matcher](https://github.com/akka/akka-http/blob/master/akka-http/src/main/scala/akka/http/scaladsl/server/PathMatcher.scala#L488)

But we don't have similar thing for Scala Unmarshaller, so one can't do stuff like this (that is actual use case which I needed):

```scala
parameter('id.as[UUID]) { id => ... }
```

While you can do this:
```scala
path(JavaUUID) { id => ... }
```

Seemed like a good thing, so I was surprised that it is not available out of the box for Scala, but is for Java, hence this PR. If I just looked in the wrong place and it is actually somewhere there, please guide me through it

### DIFF
--- a/akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/PredefinedFromStringUnmarshallers.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/PredefinedFromStringUnmarshallers.scala
@@ -4,6 +4,8 @@
 
 package akka.http.scaladsl.unmarshalling
 
+import java.util.UUID
+
 import scala.collection.immutable
 import akka.http.scaladsl.util.FastFuture
 import akka.util.ByteString
@@ -40,6 +42,15 @@ trait PredefinedFromStringUnmarshallers {
         case "false" | "no" | "off" | "0" ⇒ false
         case ""                           ⇒ throw Unmarshaller.NoContentException
         case x                            ⇒ throw new IllegalArgumentException(s"'$x' is not a valid Boolean value")
+      }
+    }
+
+  implicit val uuidFromStringUnmarshaller: Unmarshaller[String, UUID] =
+    Unmarshaller.strict { string ⇒
+      try UUID.fromString(string)
+      catch {
+        case e: IllegalArgumentException ⇒
+          throw new IllegalArgumentException(s"'$string' is not a valid UUID value", e)
       }
     }
 
--- a/akka-http-tests/src/test/scala/akka/http/scaladsl/unmarshalling/UnmarshallingSpec.scala
+++ b/akka-http-tests/src/test/scala/akka/http/scaladsl/unmarshalling/UnmarshallingSpec.scala
@@ -4,6 +4,8 @@
 
 package akka.http.scaladsl.unmarshalling
 
+import java.util.UUID
+
 import akka.http.scaladsl.unmarshalling.Unmarshaller.EitherUnmarshallingException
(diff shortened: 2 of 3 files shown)


johanandren:
OK TO TEST

maximskripnik:
@johanandren Does this really break binary compatibilty or is it safe to keep as is?

johanandren:
I think it does break bincomp and source compatibility because someone may have mixed in the `PredefinedFromStringUnmarshallers` trait and already implemented a method called `uuidFromStringUnmarshaller`.

I'm not sure if we have been historically strict with those. @jrudolph ?

akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/PredefinedFromStringUnmarshallers.scala
```diff
@@ -43,6 +45,15 @@ trait PredefinedFromStringUnmarshallers {
       }
     }
 
+  implicit val uuidFromStringUnmarshaller: Unmarshaller[String, UUID] =
+    Unmarshaller.strict { string ⇒
+      try UUID.fromString(string)
+      catch {
+        case e: IllegalArgumentException ⇒
+          throw new IllegalArgumentException(s"'$string' is not a valid UUID value", e)
```
jrudolph: Does that wrapping improve the existing error message? In that case :+1:.

jrudolph:
> I think it does break bincomp and source compatibility because someone may have mixed in the `PredefinedFromStringUnmarshallers` trait and already implemented a method called `uuidFromStringUnmarshaller`.
> 
> I'm not sure if we have been historically strict with those. @jrudolph ?

I don't think we need to be strict here. Maybe we should add `@DoNotExtend` to those classes for the future.

Can you add mima exclusions for these, @maximskripnik?

akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/PredefinedFromStringUnmarshallers.scala
maximskripnik: @jrudolph I think it improves it in a sense that it makes it more consistent with all other predefined unmarshallers' errors.

Java's `UUID.fromString` exception error message is implemented like this:

 `"Invalid UUID string: "+name`.

Now here are other predefined from string unmarshallers' errors in this file:

`s"'$x' is not a valid Boolean value"`
`s"'$value' is not a valid 8-bit hexadecimal integer value"`
`s"'$value' is not a valid 8-bit signed integer value"`

It's arguable whether it's worth wrapping exception for this, but I voted for consistency (and prettiness)

maximskripnik:
@jrudolph Sure, there those are

akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/PredefinedFromStringUnmarshallers.scala
jrudolph: I see, thanks for the explanation.

jrudolph:
Great, thanks @maximskripnik.

PR #2506, opened 2019-04-12
Success returns even if IO error occurs in FileUploadDirectives.

takayahilton:
## Purpose

Fail when IO error occurs in FileUploadDirectives

## Background Context

If an error occurs when writing to disk, it will be `Success(count, Failure(ex))` instead of `Failure(ex)`.
So I tried to throw an error.

`akka-stream FileSubscriber.scala`

```scala
  def receive = {
    case ActorSubscriberMessage.OnNext(bytes: ByteString) =>
      try {
        bytesWritten += chan.write(bytes.asByteBuffer)
      } catch {
        case ex: Exception =>
          closeAndComplete(Success(IOResult(bytesWritten, Failure(ex))))
          cancel()
      }
```
[code](https://github.com/akka/akka/blob/master/akka-stream/src/main/scala/akka/stream/impl/io/FileSubscriber.scala#L73)

### DIFF
--- a/akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileUploadDirectives.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileUploadDirectives.scala
@@ -7,18 +7,17 @@ package akka.http.scaladsl.server.directives
 import java.io.File
 
 import akka.annotation.ApiMayChange
-import akka.http.scaladsl.server.{Directive, Directive1, MissingFormFieldRejection}
-import akka.http.scaladsl.model.{ContentType, Multipart}
+import akka.http.scaladsl.server.{ Directive, Directive1, MissingFormFieldRejection }
+import akka.http.scaladsl.model.{ ContentType, Multipart }
 import akka.util.ByteString
 
 import scala.collection.immutable
 import scala.concurrent.Future
+import scala.util.{ Failure, Success }
 import akka.stream.scaladsl._
 import akka.http.javadsl
 import akka.http.scaladsl.server.RouteResult
 import akka.stream.IOResult
-
-import scala.util.{Failure, Success}
 
 /**
  * @groupname fileupload File upload directives
--- a/akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileUploadDirectives.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileUploadDirectives.scala
@@ -7,8 +7,8 @@ package akka.http.scaladsl.server.directives
 import java.io.File
 
 import akka.annotation.ApiMayChange
-import akka.http.scaladsl.server.{ Directive, Directive1, MissingFormFieldRejection }
-import akka.http.scaladsl.model.{ ContentType, Multipart }
+import akka.http.scaladsl.server.{Directive, Directive1, MissingFormFieldRejection}
+import akka.http.scaladsl.model.{ContentType, Multipart}
 import akka.util.ByteString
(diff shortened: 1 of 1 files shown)


akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileUploadDirectives.scala
```diff
@@ -60,7 +60,10 @@ trait FileUploadDirectives {
           val uploadedF: Future[(FileInfo, File)] =
             bytes
               .runWith(FileIO.toPath(dest.toPath))
-              .map(_ ⇒ (fileInfo, dest))
+              .map { ioResult ⇒
+                val _ = ioResult.status.get
+                (fileInfo, dest)
+              }
```
viktorklang: `.transform(_.status.map(_ => (fileInfo, dest)))`

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileUploadDirectives.scala
```diff
@@ -97,7 +100,8 @@ trait FileUploadDirectives {
             val fileInfo = FileInfo(part.name, part.filename.get, part.entity.contentType)
             val dest = destFn(fileInfo)
 
-            part.entity.dataBytes.runWith(FileIO.toPath(dest.toPath)).map { _ ⇒
+            part.entity.dataBytes.runWith(FileIO.toPath(dest.toPath)).map { ioResult ⇒
```
viktorklang: `part.entity.dataBytes.runWith(FileIO.toPath(dest.toPath)).transform(_.status.map(_ => (fileInfo, dest)))`

takayahilton: @viktorklang 
I tried it.
But that failed with ci.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileUploadDirectives.scala
takayahilton: @viktorklang 
I tried it.
But that failed with ci.

viktorklang: Hmm, 2.11 cross-compilation?

takayahilton: I guess so.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileUploadDirectives.scala
```diff
@@ -60,7 +60,10 @@ trait FileUploadDirectives {
           val uploadedF: Future[(FileInfo, File)] =
             bytes
               .runWith(FileIO.toPath(dest.toPath))
-              .map(_ ⇒ (fileInfo, dest))
+              .map { ioResult ⇒
+                val _ = ioResult.status.get
```
johanandren: I think a pattern match here would be more clear:
```
ioResult.status match {
  case Success(_) => (fileInfo, dest)
  case Failure(ex) => throw ex
}
```

jrudolph:
I thought, I had filed a ticket on this in akka/akka but cannot find it any more. This API is really weird :)

raboof:
OK TO TEST

takayahilton:
@raboof 
No, this fix is ​​no longer needed.
Thanks!

Issue #2126, opened 2018-07-26
Custom MediaType and MediaRange.matches are handled in a case sensitive way


pzeppezauer:
Hi,

if I understand [RFC 6838](https://tools.ietf.org/html/rfc6838) correctly, both maintype and subtype of a MediaType should be case-insensitive.
This does not seem to work for custom media types in AKKA HTTP 10.1.3 (all tested in the scaladsl version but might apply to the java version too):

`CommonActions.getMediaType` converts both maintype and subtype to lower case using `EnhancedString.toRootLowerCase`
 
1. In case we register the custom media type `application/CUSTOM`:
```
val `application/CUSTOM`: WithFixedCharset =
  MediaType.customWithFixedCharset("application", "CUSTOM", utf8)
val parserSettings = ParserSettings(system).withCustomMediaTypes(`application/CUSTOM`)
val serverSettings = ServerSettings(system).withParserSettings(parserSettings)

val routes = ...
val binding = Http().bindAndHandle(routes, host, port, settings = serverSettings)
```

`ParserSettings.withCustomMediaTypes` creates a case **sensitive** `MediaTypes.FindCustom` although the documentation of `MediaTypes.FindCustom` mentions that the input will be lower case.

The lookup for the custom media type in `CommonActions.getMediaType` (via `CommonActions.customMediaTypes` then does not work because it looks for `("application", "custom")` which does not match the registered `("application", "CUSTOM")`.

as a fallback `CommonActions` then creates `MediaType.customBinary("application", "custom", MediaType.Compressible, params = params, allowArbitrarySubtypes = true)` using the **lower** case main- and subtype.

2. In case we don't register it but only use it in a `Unmarshaller` or if the lookup above fails: 

If we convert the media type `application/CUSTOM` (from the example above) to a `ContentTypeRange` and then use it in a `Unmarshaller` for specific content types (via `Unmarshaller.EnhancedFromEntityUnmarshaller#forContentTypes`) we face then another issue.

The media type is converted to a ```MediaRange.One(`application/CUSTOM`, 0f)``` (from the example above) which then does not match the media type `application/custom` created by `CommonActions`, because `MediaRange.One.match` compares both main- and subtype in a case sensitive way.

IMO there are 2 issues:
 * `ParserSettings.withCustomMediaTypes` should register the custom media type with lower case main- and subtype to conform to the scaladoc of `MediaTypes.FindCustom`.
[...]

PR #2509, opened 2019-04-17
Make sure that MediaType is lowercase automatically

aafa:
## Purpose

This PR is a followup of #2136.
As discussed in #2136, in order to provide case insensitive comparison of the MediaType's mainType/subType we want to make sure that types are in a lower case at the moment we create an object.

This way we can be certain that both cases mentioned in #2126 are fixed since we're not preserving case information for custom media types.

Discarding case information has been done on multiple other occasions in the code so that seems like a preferable way to handle scenarios like this one.

## References

Original PR #2136
Issue #2126


raboof:
OK TO TEST

akka-http-core/src/main/scala/akka/http/scaladsl/model/MediaType.scala
```diff
@@ -136,10 +136,10 @@ object MediaType {
...
-    require(mainType != "multipart", "Cannot create a MediaType.Multipart here, use `customMultipart` instead!")
-    require(allowArbitrarySubtypes || subType != "*", "Cannot create a MediaRange here, use `MediaRange.custom` instead!")
-    val _params = params
-    new WithFixedCharset(renderValue(mainType, subType, params), mainType, subType, charset, fileExtensions) {
+    val (_params, _mainType, _subType) = (params, mainType.toRootLowerCase, subType.toRootLowerCase)
+    require(_mainType != "multipart", "Cannot create a MediaType.Multipart here, use `customMultipart` instead!")
+    require(allowArbitrarySubtypes || _subType != "*", "Cannot create a MediaRange here, use `MediaRange.custom` instead!")
+    new WithFixedCharset(renderValue(_mainType, _subType, params), _mainType, _subType, charset, fileExtensions) {
       override def params = _params
       override def isApplication = mainType == "application"
```
raboof: Should these be updated to refer to `_mainType` as well?

aafa: Here `mainType` references class field since we have it inside `new WithFixedCharset(... val mainType: String, val subType: String ... )`, the value is passed with `_mainType` as a constructor parameter.

raboof: Ah, you're right of course!

aafa:
> I think the general approach is good. Should this logic be applied to the other methods in `object MediaType` as well?

Fields `mainType`/`subType` are defined in a `trait MediaType`. As far as I can tell there are only 3 class constructors where they are accepted as a constructor params – `Binary`,  `WithFixedCharset`, `NonMultipartWithOpenCharset/WithOpenCharset`. They all end up using corresponding `custom*` method to handle type strings. 

Honestly I would prefer having them point to the one single method where string case is being handled, or at least have stronger guarantees that any further code changes will respect lower case handling but right now I don't have any good ideas. What do you think?

jrudolph:
> Honestly I would prefer having them point to the one single method where string case is being handled, or at least have stronger guarantees that any further code changes will respect lower case handling but right now I don't have any good ideas. What do you think?

Could we add `require` to all the subclasses or even to the `MediaType` trait body to check those?

aafa:
> Could we add `require` to all the subclasses or even to the `MediaType` trait body to check those?

- `require` checks are added to `MediaType` trait to provide stronger guarantees
-  `toRootLowerCase` called directly on constructor params instead of in `custom*` helpers

aafa:
@raboof @jrudolph Is there anything else that can be done here?

aafa:
@jrudolph Sure thing, suggestions applied 👌

akka-http-core/src/main/scala/akka/http/scaladsl/model/MediaType.scala
```diff
@@ -197,9 +197,11 @@ object MediaType {
     r.get
   }
 
-  sealed abstract class Binary(val value: String, val mainType: String, val subType: String, val comp: Compressibility,
+  sealed abstract class Binary(val value: String, val _mainType: String, val _subType: String, val comp: Compressibility,
```
jrudolph: `_mainType` and `_subType` don't have to be `val`.

```suggestion
  sealed abstract class Binary(val value: String, _mainType: String, _subType: String, val comp: Compressibility,
```

akka-http-core/src/main/scala/akka/http/scaladsl/model/MediaType.scala
```diff
@@ -197,9 +197,11 @@ object MediaType {
...
                                val fileExtensions: List[String]) extends MediaType with jm.MediaType.Binary {
     def binary = true
+    def mainType: String = _mainType.toRootLowerCase
```
jrudolph: These must be `val`s for performance reasons, also in all the other places.

aafa:
@jrudolph Here we stumbled because MediaType's `require` checks will kick in earlier than mainType/subType vals are initialized so I'd prefer to go back to `def`.
Other option could be to `lazy val` but that feels like a bit of a stretch.

raboof:
Looks reasonable, anyone else wants to verify? /cc @jrudolph @jlprat

akka-http-core/src/main/scala/akka/http/scaladsl/model/MediaType.scala
jrudolph: This is a hard requirement.

jrudolph:
> Looks reasonable, anyone else wants to verify? /cc @jrudolph @jlprat

We still need to find a solution where `mainType` and `subType` will still be simple vals.

jrudolph:
https://github.com/akka/akka-http/pull/2509/files/81e4a39c81b36e88105b3c234d912d7a373f7064..929052cb1a5964688d20aaec95e86a12d3b882b5 I removed the extra check and moved the definition of `mainType/subType` to `MediaType` directly so we control lower casing. WDYT, @raboof?

jrudolph:
(Unrelated failure, restarted build)

jrudolph:
Thanks a lot, @aafa!

Issue #2415, opened 2019-02-06
CachingSettings class hierarchy doesn't follow the usual public/impl classes split


jrudolph:
For consistency that should be cleaned up. See any of the other settings classes in other modules for a guideline how to do it.

heraklos:
Hello @jrudolph.
I would like to work on this issue if it is still concerned?

I'm going to reference `akka.http.scaladsl.settings.RoutingSettings` and `akka.http.impl.settings` as a guideline if I can take this issue.

raboof:
Go for it! Making sure binary compatibility is not broken is something to take into account here.

heraklos:
I created [#2510](https://github.com/akka/akka-http/pull/2510) for this issue 👍

PR #2510, opened 2019-04-21
Fix CachingSettings class public/impl hierarchy 

heraklos:
## Purpose
CacheSettings classes does not follow usual public/impl class split structure seen in other settings classes in `akka`, `akka-http-core` package.
This PR is intended to make CacheSettings classes hierarchy right. 

## References

References [#2415](https://github.com/akka/akka-http/issues/2415)

## Changes
- Separate CacheSettings class as public/impl structure
- Adding missing classes to mima-filters list
  - 1 unused method in CacheSettings class
  - 4 missing impl class that should not be used 

## Background Context
See [this Conversation](https://github.com/akka/akka-http/pull/2298/files#r252258481)


akka-http-caching/src/main/scala/akka/http/caching/javadsl/CachingSettings.scala
```diff
@@ -6,9 +6,8 @@ package akka.http.caching.javadsl
```
heraklos: Directory structure is not corresponding to that of impl classes (`impl.settings.CacheSettingsImpl`), because {javadsl, scaladsl}.CachingSettings object has already been used by other parts. 
Assuming changing package `javadsl` to `javadsl.settings` may cause application codes that instantiate CacheSettings to not function anymore, I let this structure remained.

jrudolph:
OK TO TEST

akka-http-caching/src/main/mima-filters/10.1.8.backwards.excludes
```diff
@@ -0,0 +1,9 @@
+# Scala 2.13.0-M5
+ProblemFilters.exclude[MissingTypesProblem]("akka.http.caching.scaladsl.CachingSettings$")
```
jrudolph: Probably because it's no companion any more? Good that we don't have to care for it, if it's only a problem for 2.13.0-M5.

Maybe we should disable MiMa for 2.13.0 for the time being, instead of having to had general excludes? WDYT, @raboof?

akka-http-caching/src/main/mima-filters/10.1.8.backwards.excludes
```diff
@@ -0,0 +1,9 @@
...
+
+# CachingSettings public/impl classes split
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.caching.scaladsl.CachingSettings.fromSubConfig")
```
jrudolph: Hmm, that's because that method was publicly implemented in `CachingSettings`. Strictly, it's a binary incompatibility but I guess we can risk it because it's unlikely to be used directly.

akka-http-caching/src/main/scala/akka/http/caching/javadsl/CachingSettings.scala
jrudolph: Yes, for this small module, it's fine.

akka-http-caching/src/main/mima-filters/10.1.8.backwards.excludes
jrudolph: Actually, it fails for me for all Scala versions with

```
[error]  * the type hierarchy of object akka.http.caching.scaladsl.CachingSettings is different in current version. Missing types {akka.http.impl.util.SettingsCompanionImpl}
```

which is strictly a binary incompatibility but actually the one we intended, so let's just replace the comment above.

akka-http-caching/src/main/mima-filters/10.1.8.backwards.excludes
```diff
@@ -0,0 +1,9 @@
+# Scala 2.13.0-M5
```
jrudolph:
```suggestion
# Type hierarchy of `CachingSettings` changed because `SettingCompanionImpl` is not a
# supertype any more. `SettingCompanionImpl` is `@InternalApi` so it's ok.
```

akka-http-caching/src/main/mima-filters/10.1.8.backwards.excludes
raboof: I agree it is probably better to disable MiMa for 2.13.0-M5 instead of creating filters (and potentially filtering things on 2.12 as well).

But not needed yet it looks like, since we want the filter for all versions in this case ;)

akka-http-caching/src/main/mima-filters/10.1.8.backwards.excludes
heraklos: Thanks for your suggestion!

heraklos:
Should I squash my commits into one, now?

jrudolph:
I used github for squashing the commits ;)

Great work. Thanks, @heraklos.

heraklos:
@jrudolph @raboof 
Thanks for your great support :)

PR #2519, opened 2019-05-06
use splitAfter in Websocket implementation

k-bespalov:
## Purpose
I found a FIXME in code and willing to fix it. This FIXME has reference to https://github.com/akka/akka/issues/16885

## References
https://github.com/akka/akka/issues/16885

## Problems
now I have 20 failed tests. 19 of them fail with error message that they expect strict message but get source (obviously why). Is it ok to fix those tests?
and one is complaining:
```
"for echoing user handler"
java.lang.AssertionError: assertion failed: expected 11 bytes, but got 11 bytes 
Expected: 
   ByteString(11 bytes)

     81 09 4D 65 73 73 61 67 65 20 31                 | ..Message 1
But got: 
   ByteString(11 bytes)

     01 09 4D 65 73 73 61 67 65 20 31                 | ..Message 1
	at scala.Predef$.assert(Predef.scala:223)
	at akka.http.impl.engine.ws.ByteStringSinkProbe$$anon$1.expectBytes(ByteStringSinkProbe.scala:83)
	at akka.http.impl.engine.ws.WebSocketServerSpec$TestSetup.expectBytes(WebSocketServerSpec.scala:183)
```
i have the same text messages but different bytes sequence. I suppose this is not ok, but I couldn't find the reason

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/ws/WebSocket.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/ws/WebSocket.scala
@@ -138,7 +138,7 @@ private[http] object WebSocket {
     val collectMessage: Flow[MessageDataPart, Message, NotUsed] =
       Flow[MessageDataPart]
         .prefixAndTail(1)
-        .collect {
+        .map {
           case (TextMessagePart(text, true) +: Nil, remaining) ⇒
             StreamUtils.cancelSource(remaining)(StreamUtils.OnlyRunInGraphInterpreterContext)
             TextMessage.Strict(text)
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/ws/WebSocket.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/ws/WebSocket.scala
@@ -139,19 +139,21 @@ private[http] object WebSocket {
       Flow[MessageDataPart]
         .prefixAndTail(1)
         .collect {
-          case ((first: TextMessagePart) +: Nil, remaining) ⇒
+          case (TextMessagePart(text, true) +: Nil, remaining) ⇒
+            StreamUtils.cancelSource(remaining)(StreamUtils.OnlyRunInGraphInterpreterContext)
+            TextMessage.Strict(text)
+          case ((first @ TextMessagePart(_, false)) +: Nil, remaining) ⇒
             TextMessage(
               (Source.single(first) ++ remaining)
-                .collect {
-                  case t: TextMessagePart if t.data.nonEmpty ⇒ t.data
-                }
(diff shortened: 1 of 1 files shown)


jrudolph:
OK TO TEST

akka-http-core/src/main/scala/akka/http/impl/engine/ws/WebSocket.scala
```diff
@@ -138,36 +138,27 @@ private[http] object WebSocket {
     val collectMessage: Flow[MessageDataPart, Message, NotUsed] =
       Flow[MessageDataPart]
         .prefixAndTail(1)
-        .mapConcat {
-          // happens if we get a MessageEnd first which creates a new substream but which is then
-          // filtered out by collect in `prepareMessages` below
-          case (Nil, _) ⇒ Nil
-          case (first +: Nil, remaining) ⇒ (first match {
-            case TextMessagePart(text, true) ⇒
```
jrudolph: I think we still want to have these cases and create strict messages there, don't we?

jrudolph:
Thanks @k-bespalov for tackling this FIXME. Basic approach looks good.

> 19 of them fail with error message that they expect strict message but get source (obviously why).

I think that can be fixed as commented in the code.

> i have the same text messages but different bytes sequence. I suppose this is not ok, but I couldn't find the reason

Seems like a flag in the first byte is now different, probably because it now produces an extra empty frame because no strict messages are created any more. So, that might just be fixed by fixing the above thing.

akka-http-core/src/main/scala/akka/http/impl/engine/ws/WebSocket.scala
```diff
@@ -138,36 +138,29 @@ private[http] object WebSocket {
...
-              StreamUtils.cancelSource(remaining)(StreamUtils.OnlyRunInGraphInterpreterContext)
-              BinaryMessage.Strict(data)
-            case first @ BinaryMessagePart(data, false) ⇒
-              BinaryMessage(
-                (Source.single(first) ++ remaining)
-                  .collect {
-                    case t: BinaryMessagePart if t.data.nonEmpty ⇒ t.data
-                  })
-          }) :: Nil
+        .collect {
```
jrudolph: Could that now be `map` or what we be missing?

k-bespalov: yes, it could be map. nothing is missed

jrudolph: :+1:

jrudolph:
Great, thanks @k-bespalov.

Issue #2520, opened 2019-05-06
ResponseRendering.scala doesn't respect server settings when logging suppression warnings


dmarticus:
As shown in the title, it looks like ResponseRendering.scala doesn't respect the server settings when calling the renderHeaders method (the suppressionWarning method doesn't accept the settings as an argument either, see https://github.com/akka/akka-http/blob/f57ec755a9107a8c1c25412ecde3149a91261e7c/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/ResponseRendering.scala#L86)

This seems like a bug to me because it seems like it would be great to suppress those warnings by turning down the error verbosity in the akka config, but since this method never references the server settings, changing the error verbosity in the akka config has no effect on these warnings.  

I think I know how to fix this, though, and I can submit a PR that addresses this change if we decide it's a worthy fix.

Thanks!

raboof:
Which server setting did you expect to suppress these warnings?

You might be thinking of `illegal-header-warnings`, but in that case I think it is intentional: that controls warnings "in case an incoming message (request or response) contains an HTTP header which cannot be parsed" (https://doc.akka.io/docs/akka-http/current/configuration.html). The line you are quoting, however, deals with the outgoing response message at the server side.

jrudolph:
I'd say the warning is warranted most of the times. It is only given if there's an mistake in your server code in that you add headers that you shouldn't be adding to a response.

PR #2521, opened 2019-05-06
fix for #2520

dmarticus:
## Purpose

This change is an attempt at fixing https://github.com/akka/akka-http/issues/2520 by giving users the option to suppress warning messages from RenderResponses.scala

## References

References #2520 

## Changes

* `ResponseRendering.scala`

## Background Context

I use akka-http at work and we had to put in a hack that strips certain headers from our server-side responses so that our logs don't blow up with superfluous.  I'd love to have the ability to just turn off these types of warnings at the config level so that we don't have to write any additional parsing layers in our service.


johanandren:
OK TO TEST

dmarticus:
Whoops, turns out I have a company policy about contributing to open-source libraries that hasn't been completely dialed in yet; as a result, I don't think I'm actually allowed to contribute here.  Closing this MR for now, but I'm leaving the issue open for now.

johanandren:
Sorry to hear that, I hope it will land at being allowed to contribute in the future!

Issue #2385, opened 2019-01-15
Cache refresh without remove


johanandren:
Would be good if we could allow a `put(key, future)` operation, that does not remove the cached value for `key` if there is one but replaces it when the future completes. This could allow a user to minimize the risk for a cache miss/refresh by having some scheduled operation to keep the cache hot for specific keys.

jrudolph:
Fixed by #2522.

PR #2522, opened 2019-05-07
Cache refresh without remove #2385

ashishkujoy:
## Purpose
Provided a put method in cache API and implemented the same in Lfu cache, this method takes a key and future value, if the key is not already cached it save key into the cache and in case key is already cached, it replaces the old cache value on once future value successfully completes

## References
https://github.com/akka/akka-http/issues/2385

References #xxxx

## Changes
Added put method in Cache


raboof:
OK TO TEST

akka-http-caching/src/main/scala/akka/http/caching/LfuCache.scala
```diff
@@ -111,7 +112,20 @@ private[caching] class LfuCache[K, V](val store: AsyncLoadingCache[K, V]) extend
 
   def apply(key: K, genValue: () ⇒ Future[V]): Future[V] = store.get(key, toJavaMappingFunction[K, V](genValue)).toScala
 
-  def getOrLoad(key: K, loadValue: K ⇒ Future[V]) = store.get(key, toJavaMappingFunction[K, V](loadValue)).toScala
+  def getOrLoad(key: K, loadValue: K ⇒ Future[V]): Future[V] =
+    store.get(key, toJavaMappingFunction[K, V](loadValue)).toScala
+
+  def put(key: K, mayBeValue: Future[V])(implicit ex: ExecutionContext): Future[Unit] = {
+    val cachedBefore = store.synchronous().asMap().containsKey(key)
```
raboof: I think `store.getIfPresent(key) != null` might be better as it allows for an optimized implementation

akka-http-caching/src/main/scala/akka/http/caching/LfuCache.scala
```diff
@@ -111,7 +112,20 @@ private[caching] class LfuCache[K, V](val store: AsyncLoadingCache[K, V]) extend
...
+    val promise = Promise[Unit]
+    if (cachedBefore) {
+      mayBeValue.onComplete {
+        case Success(value)      ⇒ promise.completeWith(Future(store.synchronous().put(key, value)))
```
raboof: How about `store.put(key, value); promise.success(())`?

akka-http-caching/src/main/scala/akka/http/caching/LfuCache.scala
```diff
@@ -111,7 +112,20 @@ private[caching] class LfuCache[K, V](val store: AsyncLoadingCache[K, V]) extend
...
+        case failure: Failure[_] ⇒ promise.failure(failure.exception)
+      }
+    } else promise.completeWith(Future(store.put(key, toJava(mayBeValue).toCompletableFuture)))
```
raboof: The logic of this 'else' branch seems OK, but perhaps it could be written as `store.put(key, toJava(mayBeValue).toCompletableFuture); Future.success(())`? That way we might not even need the `Promise`

What happens when `mayBeValue` fails here? Perhaps we should return `mayBeValue.map(_ => ())`?

ashishkujoy: How about this implementation.
```scala
def put(key: K, mayBeValue: Future[V])(implicit ex: ExecutionContext): Future[Unit] = {
    val previouslyCacheValue = Option(store.getIfPresent(key))

    previouslyCacheValue match {
      case None ⇒ store.put(key, toJava(mayBeValue).toCompletableFuture)
        Future.successful(())
      case _ ⇒ mayBeValue.map(value ⇒ store.put(key, toJava(Future.successful(value)).toCompletableFuture))
    }
  }
```
Now in this implementation if the given future fails, the caller of this method can take appropiate action.

akka-http-caching/src/main/scala/akka/http/caching/LfuCache.scala
ashishkujoy: Done

akka-http-caching/src/main/scala/akka/http/caching/LfuCache.scala
```diff
@@ -111,7 +112,19 @@ private[caching] class LfuCache[K, V](val store: AsyncLoadingCache[K, V]) extend
 
   def apply(key: K, genValue: () ⇒ Future[V]): Future[V] = store.get(key, toJavaMappingFunction[K, V](genValue)).toScala
 
-  def getOrLoad(key: K, loadValue: K ⇒ Future[V]) = store.get(key, toJavaMappingFunction[K, V](loadValue)).toScala
+  def getOrLoad(key: K, loadValue: K ⇒ Future[V]): Future[V] =
+    store.get(key, toJavaMappingFunction[K, V](loadValue)).toScala
+
+  def put(key: K, mayBeValue: Future[V])(implicit ex: ExecutionContext): Future[Unit] = {
```
raboof: Actually, since `put` doesn't return anything anyway, perhaps we should return `Unit` here and leave it up to the caller to decide whether or not to attach error handling to the `Future[V]`? Or just return the `Future[V]` to make it more explicit that it's the users' responsibility to do error handling?

ashishkujoy: I think returning Future[V] will better

ashishkujoy:
Why does Jenkins PR Validation fails saying 
```bash
sed: can't read ./akka-http2-support/target/test-reports/h2spec-junit.xml: No such file or directory
```
I haven't changed anything related to that

raboof:
PLS BUILD

akka-http-caching/src/main/scala/akka/http/caching/scaladsl/Cache.scala
```diff
@@ -12,8 +12,8 @@ import akka.japi.{ Creator, Procedure }
 
 import scala.collection.JavaConverters._
 import scala.collection.immutable
-import scala.compat.java8.FutureConverters.{ toJava => futureToJava, toScala => futureToScala }
-import scala.concurrent.{ Future, Promise }
+import scala.compat.java8.FutureConverters.{ toJava ⇒ futureToJava, toScala ⇒ futureToScala }
```
raboof:
```suggestion
import scala.compat.java8.FutureConverters.{ toJava => futureToJava, toScala => futureToScala }
```

akka-http-caching/src/main/scala/akka/http/caching/scaladsl/Cache.scala
```diff
@@ -23,7 +23,7 @@ import scala.concurrent.{ Future, Promise }
 @ApiMayChange
 @DoNotInherit
 abstract class Cache[K, V] extends akka.http.caching.javadsl.Cache[K, V] {
-  cache =>
+  cache ⇒
```
raboof:
```suggestion
  cache =>
```

akka-http-caching/src/test/scala/akka/http/caching/ExpiringLfuCacheSpec.scala
```diff
@@ -59,6 +59,50 @@ class ExpiringLfuCacheSpec extends WordSpec with Matchers with BeforeAndAfterAll
...
+      val done = cache.put(1, Future.successful(futureValue))
+
+      Await.result(done, 3.seconds)
+      Await.result(cache.get(1).get, 3.seconds) should be(futureValue)
+
+    }
+    "replace existing cache when evaluation of new value completes" in {
+      val cache = lfuCache[String]()
+      val latch = new CountDownLatch(1)
+      val future1 = cache.get(1, () ⇒ "A")
```
raboof:
```suggestion
      val future1 = cache.get(1, () => "A")
```

akka-http-caching/src/main/scala/akka/http/caching/scaladsl/Cache.scala
```diff
@@ -58,6 +58,12 @@ abstract class Cache[K, V] extends akka.http.caching.javadsl.Cache[K, V] {
   override def getOptional(key: K): Optional[CompletionStage[V]] =
     Optional.ofNullable(get(key).map(f => futureToJava(f)).orNull)
 
+  /**
+   * Cache the given future if not cache previously.
```
jrudolph:
```suggestion
   * Cache the given future if not cached previously.
```

akka-http-caching/src/main/scala/akka/http/caching/scaladsl/Cache.scala
```diff
@@ -58,6 +58,12 @@ abstract class Cache[K, V] extends akka.http.caching.javadsl.Cache[K, V] {
...
+   * Or Replace the old cache on successful completion of given future.
```
jrudolph:
```suggestion
   * Or replace the old cached value on successful completion of given future.
```

jrudolph: Maybe also add a comment here, that the old value will be kept if future fails.

akka-http-caching/src/main/scala/akka/http/caching/LfuCache.scala
```diff
@@ -105,7 +105,21 @@ private[caching] class LfuCache[K, V](val store: AsyncCache[K, V]) extends Cache
...
-  def getOrLoad(key: K, loadValue: K => Future[V]) = store.get(key, toJavaMappingFunction[K, V](loadValue)).toScala
+  def put(key: K, mayBeValue: Future[V])(implicit ex: ExecutionContext): Future[V] = {
+    val previouslyCacheValue = Option(store.getIfPresent(key))
+
+    previouslyCacheValue match {
+      case None =>
+        store.put(key, toJava(mayBeValue).toCompletableFuture)
+        mayBeValue
+      case _ => mayBeValue.map { value =>
+        store.put(key, toJava(Future.successful(value)).toCompletableFuture)
```
jrudolph: Isn't that prone to race conditions if multiple `put`s are ongoing for the same key? What may be surprising is that after a sequence of multiple `put`s to the same key the content of the cache might not be the result of the last call.

It's a cache so maybe we shouldn't care? Any opinion, @raboof?

akka-http-caching/src/main/scala/akka/http/caching/LfuCache.scala
```diff
@@ -105,7 +105,21 @@ private[caching] class LfuCache[K, V](val store: AsyncCache[K, V]) extends Cache
...
 
-  def getOrLoad(key: K, loadValue: K => Future[V]) = store.get(key, toJavaMappingFunction[K, V](loadValue)).toScala
+  def put(key: K, mayBeValue: Future[V])(implicit ex: ExecutionContext): Future[V] = {
+    val previouslyCacheValue = Option(store.getIfPresent(key))
+
+    previouslyCacheValue match {
+      case None =>
+        store.put(key, toJava(mayBeValue).toCompletableFuture)
+        mayBeValue
+      case _ => mayBeValue.map { value =>
```
jrudolph:
```suggestion
      case _ => mayBeValue.foreach { value =>
```

akka-http-caching/src/main/scala/akka/http/caching/LfuCache.scala
jrudolph: If we agree that the race conditions are ok then we should add a note about that to the docs as well.

akka-http-caching/src/main/scala/akka/http/caching/LfuCache.scala
raboof: (this does `map` so the `mayBeValue` is returned to the user after the `store.put` was invoked. I agree it'd be OK to do `foreach` here and just return `mayBeValue` directly, no preference)

akka-http-caching/src/main/scala/akka/http/caching/LfuCache.scala
raboof: I think they are fine for now indeed.

Keeping track of multiple parallel 'in-flight' `put`s might be an interesting optimization, but that's still possible to add later without breaking the API.

akka-http-caching/src/main/scala/akka/http/caching/LfuCache.scala
jrudolph: Ah, indeed, then `map` sounds fine to me as well.

raboof:
Great!

PR #2524, opened 2019-05-09
 htp: fix case of nested FormField if first formField usage is with single parameter

jrudolph:
So far, in #2283 the case for tuples was fixed which also worked for nested `formField` directives in most of the cases but not when the first usage was with a single argument.

Refs #73.

### DIFF
--- a/akka-http/src/main/scala/akka/http/scaladsl/server/directives/FormFieldDirectives.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/server/directives/FormFieldDirectives.scala
@@ -7,7 +7,6 @@ package directives
 
 import akka.annotation.DoNotInherit
 import akka.annotation.InternalApi
-
 import akka.http.impl.util._
 import akka.http.scaladsl.common._
 import akka.http.scaladsl.model.EntityStreamSizeException
@@ -20,12 +19,14 @@ import scala.collection.immutable
 import scala.concurrent.Future
 import scala.util.{ Failure, Success }
 import BasicDirectives._
+import akka.http.ccompat.pre213
+import akka.http.ccompat.since213
 
 /**
  * @groupname form Form field directives
  * @groupprio form 90
  */
-trait FormFieldDirectives extends ToNameReceptacleEnhancements with FormFieldDirectivesCompat {
+trait FormFieldDirectives extends ToNameReceptacleEnhancements {
   import FormFieldDirectives._
 
   /**
@@ -55,7 +56,17 @@ trait FormFieldDirectives extends ToNameReceptacleEnhancements with FormFieldDir
    *
    * @group form
    */
-  override def formField(pdm: FieldMagnet): Directive[pdm.U] = formFields(pdm)
+  @pre213
+  def formField(pdm: FieldMagnet): pdm.Out = formFields(pdm)
+
+  /**
+   * Extracts an HTTP form field from the request.
+   * Rejects the request if the defined form field matcher(s) don't match.
+   *
+   * @group form
+   */
+  @since213
+  def formField(pdm: FieldMagnet): Directive[pdm.U] = formFields(pdm)
 
   /**
    * Extracts a number of HTTP form field from the request.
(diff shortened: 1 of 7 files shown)


johanandren:
Tried to review this and looked at the previous PR but I don't understand how it works, could you give a brief explanation @jrudolph ?

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FormFieldDirectives.scala
```diff
@@ -52,22 +55,35 @@ trait FormFieldDirectives extends ToNameReceptacleEnhancements {
...
 
   /**
    * Extracts a number of HTTP form field from the request.
    * Rejects the request if the defined form field matcher(s) don't match.
    *
    * @group form
    */
-  def formFields(pdm: FieldMagnet): pdm.Out = pdm()
+  override def formFields(pdm: FieldMagnet): Directive[pdm.U] =
+    toStrictEntity(StrictForm.toStrictTimeout).wrap { pdm() }
```
jrudolph: That's basically the fix. We want to be sure that the complete inner part of a formField directive is wrapped in a `toStrictEntity` directive.

The problem in the old version was, that we couldn't prove that `pdm()` actually is a directive, so we couldn't use `wrap`.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FormFieldDirectives.scala
```diff
@@ -206,25 +223,29 @@ object FormFieldDirectives extends FormFieldDirectives {
...
+    implicit def forRepVDR[T]: FieldDefAux[RepeatedValueUnmarshallerReceptacle[T], Tuple1[Iterable[T]]] =
+      extractField[RepeatedValueUnmarshallerReceptacle[T], Iterable[T]] { rvr ⇒ repeatedFilter(rvr.name, StrictForm.Field.unmarshallerFromFSU(rvr.um)) }
 
     //////////////////// tuple support ////////////////////
 
     import akka.http.scaladsl.server.util.BinaryPolyFunc
     import akka.http.scaladsl.server.util.TupleOps._
 
-    implicit def forTuple[T](implicit fold: FoldLeft[Directive0, T, ConvertFieldDefAndConcatenate.type]): FieldDefAux[T, fold.Out] =
-      fieldDef[T, fold.Out](fold(toStrictEntity(StrictForm.toStrictTimeout), _))
```
jrudolph: That's the old fix which fixed it for `formField('a, 'b)`, i.e. if there are more than two form fields, than fold the directives starting with `toStrictEntity` on the outside.

It didn't fix it for the other cases of a single form field, though. One potential solution would be to add `toStrictEntity` to all the implicits for the single form field above, that would work but would mean that we would add `toStrictEntity` as many times as we have `formFields` which would add a bit of overhead.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FormFieldDirectives.scala
```diff
@@ -113,27 +129,28 @@ object FormFieldDirectives extends FormFieldDirectives {
   private def toMap(seq: Seq[(String, String)]): Map[String, String] = immutable.TreeMap(seq: _*)
 
   sealed trait FieldMagnet {
-    type Out
-    def apply(): Out
+    type U
+    def apply(): Directive[U]
```
jrudolph: So, the typesafe fix is to add enough types here (and everywhere) to declare that the result of `pdf()` above actually is a `Directive` and not just anything.

That is we add the `Directive` type here and remove the `Directive` wrapping in all types below. Note that `type Directive1[T] = Directive[Tuple1[T]]` etc.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FormFieldDirectives.scala
jrudolph: Thinking about it, there's already existing overhead: for every given field, we run the complete unmarshalling which seems wasteful. So, that could be another area of improvement.

raboof:
MiMa:
```
[error]  * method formFields(akka.http.scaladsl.server.directives.FormFieldDirectives#FieldMagnet)java.lang.Object in trait akka.http.scaladsl.server.directives.FormFieldDirectives does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.scaladsl.server.directives.FormFieldDirectives.formFields")
[error]  * method formField(akka.http.scaladsl.server.directives.FormFieldDirectives#FieldMagnet)java.lang.Object in trait akka.http.scaladsl.server.directives.FormFieldDirectives does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.scaladsl.server.directives.FormFieldDirectives.formField")
[error]  * method formFields(akka.http.scaladsl.server.directives.FormFieldDirectives#FieldMagnet)akka.http.scaladsl.server.Directive in trait akka.http.scaladsl.server.directives.FormFieldDirectives is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.scaladsl.server.directives.FormFieldDirectives.formFields")
[error]  * method formField(akka.http.scaladsl.server.directives.FormFieldDirectives#FieldMagnet)akka.http.scaladsl.server.Directive in trait akka.http.scaladsl.server.directives.FormFieldDirectives is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]
```

jrudolph:
> MiMa:

These mima failures happen only on 2.11. I tried to keep bin compat by adding this FormFieldDirectivesCompat subclass but that probably won't help for 2.11. I currently don't know how to solve that problem.

jrudolph:
> MiMa:

> These mima failures happen only on 2.11. I tried to keep bin compat by adding this FormFieldDirectivesCompat subclass but that probably won't help for 2.11. I currently don't know how to solve that problem.

Most easily by changing the return type to a dummy type member similar to before so that the erased result type is still `Object`/`AnyRef`.

raboof:
Hmm tricky.

Perhaps we should start planning dropping support for 2.11 (some time after Akka 2.6 is released?) and postponing this PR until then?

jrudolph:
I think I fixed the incompatibility now. The idea was to keep the `Out` type member alongside the new `U` type member. This way the erased result type is still `AnyRef`. I added different alternative using `@pre213` and `@since213` so that we can drop the extra structure once 2.12 support is gone.

PR #2516, opened 2019-05-02
2.13.0-RC1 support

raboof:
Perhaps not for merge, as RC2 is around the corner.

PR #2479, opened 2019-03-21
Tweak 2.13 support for RC1

adriaanm:

PR #2525, opened 2019-05-09
Cross-build to Scala 2.13.0-RC1

dwijnand:


dwijnand:
Brother of #2516.  Distant cousin to #2479.

jrudolph:
OK TO TEST

project/Dependencies.scala
```diff
@@ -38,10 +38,10 @@ object Dependencies {
   }
 
   object Compile {
-    val scalaXml      = "org.scala-lang.modules"      %% "scala-xml"                   % "1.1.1" // Scala License
+    val scalaXml      = "org.scala-lang.modules"      %% "scala-xml"                   % "1.2.0" // Scala License
 
     // For akka-http spray-json support
-    val sprayJson   = "io.spray"                     %% "spray-json"                   % "1.3.5"       // ApacheV2
+    val sprayJson   = "io.crashbox"                  %% "spray-json"                   % "1.3.5-3"       // ApacheV2
```
jrudolph: Not needed, we released spray-json for 2.13.0-RC1 already.

project/AkkaDependency.scala
```diff
@@ -14,7 +14,7 @@ object AkkaDependency {
   // else if akka version is "default", then the hard coded default will be used (jenkins doesn't allow empty values for config axis)
   // else if akka.version is anything else, then the given version will be used
 
-  val defaultAkkaVersion = "2.5.19"
+  val defaultAkkaVersion = "2.5.22"
```
jrudolph: We haven't updated the required version lightly in the past. Do we care now?

akka-http-tests/src/test/scala/akka/http/scaladsl/marshalling/ContentNegotiationSpec.scala
```diff
@@ -19,147 +19,147 @@ import HttpCharsets._
 class ContentNegotiationSpec extends FreeSpec with Matchers {
   "Content Negotiation should work properly for requests with header(s)" - {
     "(without headers)" test { accept ⇒
-      accept(`text/plain` withCharset `UTF-16`) should select(`text/plain` withCharset `UTF-16`)
-      accept(`text/plain`, `text/html`) should select(`text/plain` withCharset `UTF-8`)
-      accept(`text/html`, `text/plain`) should select(`text/html` withCharset `UTF-8`)
+      accept(Seq(`text/plain` withCharset `UTF-16`)) should select(`text/plain` withCharset `UTF-16`)
```
jrudolph: To avoid the extra verbosity, could we alternatively 
 * either, create a helper method right here on top which still takes a varargs param
 * or create a custom class instead of the plain Function1 which still has varargs params

project/Dependencies.scala
jrudolph:
```suggestion
    val sprayJson   = "io.spray"                     %% "spray-json"                   % "1.3.5"       // ApacheV2
```

jrudolph: Fixed.

akka-http-tests/src/test/scala/akka/http/scaladsl/marshalling/ContentNegotiationSpec.scala
jrudolph: Implemented the second version which works without any syntax changes in 2.12+ but doesn't for 2.11 because SAM-support wasn't so good back then. Solved it with a small compat cross source file.

project/AkkaDependency.scala
dwijnand: Because 2.5.22 is the first (and currently only) version that has a release for 2.13.0-RC1.

Could also do a cross-scala-variant thing with this (but that could get confusing if you forget)

jrudolph:
(mima still fails, fixing...)

project/AkkaDependency.scala
jrudolph: That's the question if we care enough to keep 2.5.19 for 2.11/2.12 but I guess it doesn't matter a lot.

dwijnand:
Green now.

jrudolph:
I'd be fine with merging but we might also just wait for RC2 like for Akka, wdyt @raboof ?

akka-http-tests/src/test/java/akka/http/javadsl/marshalling/sse/EventStreamMarshallingTest.java
```diff
@@ -44,7 +45,7 @@ public void testToEventStream() {
         final ByteString expectedEntity = events
                 .stream()
                 .map(e -> ((akka.http.scaladsl.model.sse.ServerSentEvent) e).encode())
-                .reduce(ByteString.empty(), ByteString::concat);
+                .reduce(ByteString$.MODULE$.empty(), ByteString::concat);
```
raboof: (can be improved once Akka is released with https://github.com/akka/akka/pull/26931)

docs/src/test/scala/docs/http/scaladsl/server/directives/PathDirectivesExamplesSpec.scala
```diff
@@ -418,7 +418,7 @@ class PathDirectivesExamplesSpec extends RoutingSpec with CompileOnlySpec {
         // Thanks to `ignoreTrailingSlash` it will serve both `/foo` and `/foo/`.
         complete("OK")
       } ~
-        path("bar" /) {
```
raboof: `postfixOps` might be OKish here... but this is OK to me as well

project/Doc.scala
```diff
@@ -48,14 +48,7 @@ object Scaladoc extends AutoPlugin {
...
       docs
-    }) ++ Seq(
-      // -Ymacro-annotations is not enabled for scaladoc in 2.13.0-M5 which fails the scaladoc build
-      // Therefore, we disable publishing of docs for the 2.13.0-M5 build for now.
-      // This is already fixed on 2.13.x and will be released with 2.13.0-M6/RC, so it
-      // can be removed once we move to 2.13.0-M6.
-      // See https://github.com/scala/bug/issues/11045 for more info on the underlying issue.
-      publishArtifact in (Compile, packageDoc) := scalaVersion.value != "2.13.0-M5"
-    )
+    })
```
raboof: :tada:

akka-http-tests/src/test/java/akka/http/javadsl/marshalling/sse/EventStreamMarshallingTest.java
jrudolph: Yep, filed https://github.com/akka/akka-http/issues/2530 so we don't forget.

adriaanm:
FYI I'm staging RC2 now

jrudolph:
> FYI I'm staging RC2 now

Great! Thanks for the heads up.

PR #2526, opened 2019-05-14
Add test cases reproducing testkit bug with Futures

kubukoz:
## Purpose

Add failing test cases reproducing an issue I found in the route testkit.

## The bug

Passing a request through some routes causes them to never complete nor reject the request.

In particular, I found that the problematic routes are the ones that are completed with a Future that doesn't have a value straight away (i.e. anything that's not `Future.successful(marshallable)`.

I will look for ways to fix this issue, as it's a massive blocker to implement https://github.com/softwaremill/sttp/pull/208 and for testing routes that have extra logic on top of futures returned from the service layer (essentially anything that looks like `complete(logic(input).map(toDTO))`).

It would be awesome if I could get some guidance as to where to look for the potential cause of the problem. Thanks!

Note that the test failure claims that the request was neither completed nor rejected in 5 seconds, but executing *all* the tests takes less than a second. Example:

image

### DIFF
--- a/akka-http-testkit/src/test/scala/akka/http/scaladsl/testkit/ScalatestRouteTestSpec.scala
+++ b/akka-http-testkit/src/test/scala/akka/http/scaladsl/testkit/ScalatestRouteTestSpec.scala
@@ -105,7 +105,7 @@ object Routes {
       pathPrefix("quick") {
         complete(Future.successful("done-quick"))
       } ~ pathPrefix("buggy") {
-        complete(Future.successful(()).map(_ => "done-buggy"))
+        complete(Future.successful(()).map(_ ⇒ "done-buggy"))
       } ~ pathPrefix("long") {
         complete(longFuture("done-long"))
       }
--- a/akka-http-testkit/src/test/scala/akka/http/scaladsl/testkit/ScalatestRouteTestSpec.scala
+++ b/akka-http-testkit/src/test/scala/akka/http/scaladsl/testkit/ScalatestRouteTestSpec.scala
@@ -16,8 +16,15 @@ import akka.http.scaladsl.model._
 import StatusCodes._
 import HttpMethods._
 import Directives._
+import scala.concurrent.Future
+import scala.concurrent.Await
+import scala.concurrent.Promise
+import akka.http.scaladsl.unmarshalling.Unmarshal
+import scala.concurrent.ExecutionContext
+import akka.actor.ActorSystem
 
 class ScalatestRouteTestSpec extends FreeSpec with Matchers with ScalatestRouteTest {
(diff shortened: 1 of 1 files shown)


kubukoz:
Further investigation in https://github.com/softwaremill/sttp/pull/208 brought new conclusions: even if I ran the routes in the expected way:

`request ~> route ~> check { responseAs[String] shouldBe ... }`

There was an interesting turn of events. Apparently, calling `request ~> route` *after* an asynchronous boundary also caused some issues. I'll try to provide a test case for this.

kubukoz:
Added more test cases (I realise there's a bit of a mess in the tests, but I'll make sure to make them cleaner once the issue is acknowledged).

akka-http-testkit/src/test/scala/akka/http/scaladsl/testkit/ScalatestRouteTestSpec.scala
```diff
@@ -38,6 +45,46 @@ class ScalatestRouteTestSpec extends FreeSpec with Matchers with ScalatestRouteT
...
+    "a route with a quick async boundary - without check {}" in {
+      stringResponse(Get("http://localhost/futures/buggy") ~> Route.seal(Routes.route)) shouldEqual "done-buggy"
+    }
+
+    "a long running future - without check {}" in {
+      stringResponse(Get("http://localhost/futures/long") ~> Route.seal(Routes.route)) shouldEqual "done-long"
+    }
+
+    def stringResponse(result: RouteTestResult): String = {
+      Await.result(Unmarshal(result.response).to[String], 5.seconds)
```
jrudolph: This is not a valid way to use `result.response`. You need to call `result.awaitResult` before accessing any of the other fields (which is done by `check`). Adding `result.awaitResult` before this line fixes one of the tests.

kubukoz: Yes. But the users can't do that (and I wanted these tests to show what the problem for users is). I'm fine with getting rid of this way to get the response, and getting the `check` examples to work :)

jrudolph:
Thanks for the report and the test, @kubukoz. I pushed a fix here: https://github.com/akka/akka-http/pull/2527/files. Can you see that works for you?

That said, the route testkit is currently blocking, so when you run tests you will always have to be careful not to run too many of those tests in parallel or you will end up with thread starvation and temporary deadlocks of the whole system. I haven't looked into the sttp use case, so it might be fine, because you will also just use it for testing. You just need to keep in mind that there are limits of what to expect.

kubukoz:
Thanks!

PR #2527, opened 2019-05-15
htk: manage blocking in RouteTestResult to make starvation less likely

jrudolph:
One particular starvation case is if a test is run on the ActorSystem dispatcher
which might use a BatchingExecutor. In that case thunks might be scheduled to
be run on the currently running thread but be starved by the unmanaged blocking
in RouteTestResultComponent.awaitResult.

Fixes #2526.

### DIFF
--- a/akka-http-testkit/src/test/scala/akka/http/scaladsl/testkit/ScalatestRouteTestSpec.scala
+++ b/akka-http-testkit/src/test/scala/akka/http/scaladsl/testkit/ScalatestRouteTestSpec.scala
@@ -17,6 +17,9 @@ import StatusCodes._
 import HttpMethods._
 import Directives._
 
+import scala.concurrent.Await
+import scala.concurrent.Future
+
 class ScalatestRouteTestSpec extends FreeSpec with Matchers with ScalatestRouteTest {
 
   "The ScalatestRouteTest should support" - {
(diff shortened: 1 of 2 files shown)


akka-http-testkit/src/test/scala/akka/http/scaladsl/testkit/ScalatestRouteTestSpec.scala
```diff
@@ -48,6 +51,15 @@ class ScalatestRouteTestSpec extends FreeSpec with Matchers with ScalatestRouteT
...
     }
 
+    "running on akka dispatcher threads" in Await.result(Future {
+      // https://github.com/akka/akka-http/pull/2526
+      // Check will block while waiting on the response, this might lead to starvation
+      // on the BatchingExecutor of akka's dispatcher if the blocking is not managed properly.
+      Get() ~> complete(Future(HttpResponse())) ~> check {
+        status shouldEqual OK
+      }
+    }, 5.seconds)
```
johanandren: Why would a user run their tests on the default dispatcher, isn't that a user error? Also: Isn't the test threads pretty much always managed by the test framework used?

jrudolph: Yeah, I also think the use case is somewhat bogus (see #2526 and the linked sttp PR). In any case, I guess it still makes sense to try to be a good citizen and manage blocking if possible to make starvation issues less likely.

jrudolph: Also, first time for me to see how easy it is to deadlock yourself with the BatchingExecutor if you don't manage blocking. `Await.result` etc do manage blocking so they already try to circumvent the problem to some degree. But of course, it's not a panacea: there are usually only so many extra threads to divert to until you might still get starved.

PR #2531, opened 2019-05-16
Remove unicode arrows

jrudolph:
With 2.13, we get so many warnings otherwise that it breaks Travis builds because of too much logging, see https://travis-ci.org/akka/akka-http/jobs/533278140.

As always, it's quite a nuisance to have to do that now, making history harder to read and introducing conflicts into every single open work.


jrudolph:
I [proposed a feature](https://github.com/scala-ide/scalariform/pull/280) to scalariform that it can automatically remove those pesky unicode arrows in the future. I'm really concerned about those people with the unicode keyboards.

dwijnand:
> I [proposed a feature](https://github.com/scala-ide/scalariform/pull/280) to scalariform that it can automatically remove those pesky unicode arrows in the future. I'm really concerned about those people with the unicode keyboards.

I've implemented these rewrites in https://github.com/scala/scala-rewrites/blob/da0221eccf3163973778d1aa3509944c52e15171/rewrites/src/main/scala/fix/Scala_2_13.scala#L98-L102.  And then I ran it on the akka/akka codebase to get https://github.com/akka/akka/pull/26944.  I was going to run it here too, but I got busy.  I think there's a way to check Scalafix rewrites are no-ops, for PR validation.

jrudolph:
> I've implemented these rewrites in [scala/scala-rewrites:rewrites/src/main/scala/fix/Scala_2_13.scala@`da0221e`#L98-L102](https://github.com/scala/scala-rewrites/blob/da0221eccf3163973778d1aa3509944c52e15171/rewrites/src/main/scala/fix/Scala_2_13.scala#L98-L102). And then I ran it on the akka/akka codebase to get [akka/akka#26944](https://github.com/akka/akka/pull/26944). I was going to run it here too, but I got busy. I think there's a way to check Scalafix rewrites are no-ops, for PR validation.

Sounds interesting. Would be interesting to see how quick it is.

PR #2538, opened 2019-05-20
Update MiMa

raboof:
To a version that understands more potential incompatibilities, hence
the extra filters.

To test https://github.com/lightbend/migration-manager/pull/299 - draft until that is actually released

### DIFF
--- a/akka-http/src/main/mima-filters/10.1.8.backwards.excludes
+++ b/akka-http/src/main/mima-filters/10.1.8.backwards.excludes
@@ -10,3 +10,19 @@ ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.scaladsl.server.
 ProblemFilters.exclude[IncompatibleResultTypeProblem]("akka.http.scaladsl.server.directives.FormFieldDirectives#FieldDef.apply")
 ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.scaladsl.server.directives.FormFieldDirectives#FieldDef.apply")
 ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.scaladsl.server.directives.FormFieldDirectives#FieldMagnet.convert")
+
+#### The following are against 10.1.5, but weren't caught by a previous version of MiMa
+
+# scaladsl:
+ProblemFilters.exclude[IncompatibleSignatureProblem]("akka.http.scaladsl.server.RouteResult#Rejected.getRejections")
+
+# Used to be public, became protected, which made a forwarder disappear
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.scaladsl.marshalling.Marshaller.randomBoundary")
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.scaladsl.marshalling.Marshaller.multipartBoundaryLength")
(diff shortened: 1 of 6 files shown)


jrudolph:
Interesting. Thanks for trying that, @raboof. Have you checked how many of those instances were also flagged for more serious changes (i.e. one that fails with `LinkageError`) and already excluded?

raboof:
I think none: if any 'more serious' signature discrepancies are found no `IncompatibleSignatureProblem` is emitted.

raboof:
Depends on https://github.com/lightbend/mima/issues/294

akka-http/src/main/mima-filters/10.1.8.backwards.excludes
```diff
@@ -9,4 +9,29 @@ ProblemFilters.exclude[IncompatibleResultTypeProblem]("akka.http.scaladsl.server
...
+
+ProblemFilters.exclude[IncompatibleResultTypeProblem]("akka.http.scaladsl.server.directives.FormFieldDirectives.formFields")
+ProblemFilters.exclude[IncompatibleResultTypeProblem]("akka.http.scaladsl.server.directives.FormFieldDirectives.formField")
+ProblemFilters.exclude[IncompatibleResultTypeProblem]("akka.http.scaladsl.server.Directives.formFields")
+ProblemFilters.exclude[IncompatibleResultTypeProblem]("akka.http.scaladsl.server.Directives.formField")
+ProblemFilters.exclude[IncompatibleResultTypeProblem]("akka.http.scaladsl.server.HttpApp.formFields")
+ProblemFilters.exclude[IncompatibleResultTypeProblem]("akka.http.scaladsl.server.HttpApp.formField")
+
+ProblemFilters.exclude[IncompatibleSignatureProblem]("akka.http.scaladsl.server.directives.FormFieldDirectives#FieldDef.fieldDef")
+ProblemFilters.exclude[IncompatibleSignatureProblem]("akka.http.scaladsl.server.directives.FormFieldDirectives#FieldDef.forTuple")
```
raboof: I'm not so sure the exclusions above, related to #2524, are safe. These only only apply to Scala 2.13.0, though.

jrudolph: I added a module to test this particular case in https://github.com/akka/akka-http/pull/2597. AFAICS usages of `formField` are binary compatible. If someone would be dealing with `FieldDefs` manually that would fail (because the signature of its apply method changed).

akka-http/src/main/scala/akka/http/scaladsl/server/Directive.scala
```diff
@@ -175,10 +175,14 @@ object Directive {
...
 
-  // previous, non-value class implementation kept around for binary compatibility
-  // TODO: remove with next binary incompatible release bump
-  private[server] def SingleValueModifiers[T](underlying: Directive1[T]): SingleValueModifiers[T] =
-    new SingleValueModifiers(underlying)
+  /**
+   * previous, non-value class implementation kept around for binary compatibility
+   * TODO: remove with next binary incompatible release bump
+   *
+   * INTERNAL API
```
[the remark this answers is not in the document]
raboof: hmm, that leads to `method SingleValueModifiers(akka.http.scaladsl.server.Directive)akka.http.scaladsl.server.Directive#SingleValueModifiers in class akka.http.scaladsl.server.Directive does not have a correspondent in current version` - I guess the static forwarder no longer gets generated when it is `private[...]` here on the companion object?

PR #2539, opened 2019-05-22
Support 2.13.0-RC2

raboof:


project/MiMa.scala
```diff
@@ -58,12 +58,13 @@ object MiMa extends AutoPlugin {
...
       )
       val post213Versions = Set(
-          // nothing released for 2.13.0-RC1 yet
+          // To be updated once we have released
+          // for 2.13.0-RC2
       )
 
       val scalaVersion = scalaBinaryVersion.value
       val versions =
-        if (scalaVersion.startsWith("2.13")) post213Versions
```
jrudolph: Isn't the old version more correct even in the future? Or I missed something why we have to put in a specific version here?

jrudolph:
Great, that no changes are necessary this time!

project/MiMa.scala
raboof: There is no specific necessity to put in a specific version here, I just liked referring to a constant rather than using `startsWith` string comparison.

There are guard rails against accidentally switching to a different 2.13 value in place: that produces a `Switch failed: no subprojects list "2.13.0-RC1" (or compatible version) in crossScalaVersions setting` sbt error.

I'm OK with switching back if you prefer though.

jrudolph: No, it's fine. I definitely don't care enough :)

PR #2540, opened 2019-05-22
Add modeled header for Content-Location

jypma:
As described in the HTTP/1.1 spec here
https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html

## Purpose

Add a modeled header for the `Content-Location` HTTP/1.1 header

## Changes

- Added Java and Scala types

## Background Context

We output that header in a service, and it'd be nice to do so in a type-safe manner.


raboof:
@jypma would you like to follow up on this one or should someone else pick up from here?

jypma:
Yes please, if someone can add a bit of tests. I don't know yet when I'll have time to come back to this :-)

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
```diff
@@ -436,11 +436,14 @@ final case class `Content-Length` private[http] (length: Long) extends jm.header
...
-// https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html
+// https://tools.ietf.org/html/rfc7231#section-3.1.4.2
 object `Content-Location` extends ModeledCompanion[`Content-Location`]
 final case class `Content-Location`(uri: Uri) extends jm.headers.ContentLocation with ResponseHeader {
+  require(uri.fragment.isEmpty, "Content-Location header URI must not contain a fragment")
+  require(uri.authority.userinfo.isEmpty, "Content-Location header URI must not contain a userinfo component")
+
   def renderValue[R <: Rendering](r: R): r.type = { import UriRendering.UriRenderer; r ~~ uri }
-  protected def companion = Location
+  protected def companion = `Content-Location`
```
jrudolph: Oops, caught by the tests and fixed ;)

Issue #2523, opened 2019-05-08
Automatically detect and use HTTP2 with prior knowledge clients


jroper:
I'm not sure if this is a good feature request or not, I did a quick search of Google and a quick skim of the RFC, and couldn't find any mention of servers supporting this, but that doesn't mean it doesn't exist.

When an HTTP/2 with prior knowledge connection is made, as required by [RFC 7540 Section 3.5](https://tools.ietf.org/html/rfc7540#section-3.5), the following preface is always sent:

```
PRI * HTTP/2.0

SM

```

If an HTTP/1.1 server received this preface, it could transparently switch to HTTP/2. If it did that, the same server could support HTTP/1.1, h2c (ie via upgrade mechanism, as implemented in #2464), and h2 with prior knowledge on the same port. One practical use of this would be that the same server could serve HTTP/1.1 and gRPC using HTTP/2 prior knowledge, which would mean that you could serve Kubernetes health checks (which are HTTP/1.1) from the same port as the gRPC app.

This is related, but different, to #1966.

raboof:
I agree this would be interesting, and the preface makes it kind of obvious, but I haven't seen it done anywhere else either. Not sure how well it would fit in code-wise, though ;)

jrudolph:
Fixed in #2543 by using `UseHttp2.Negotiated` on plain http servers.

PR #2543, opened 2019-05-24
=htc Support `UseHttp2.Negotiated` for http2 connections over plain http

viktorklang:
## Purpose

If we want to support things like serving unencrypted gRPC traffic from the same server instance as unencrypted REST traffic—we already support serving encrypted gRPC traffic and REST traffic from the same server as is with UseHttp2.ALWAYS—then the server logic needs to be changed.

This draft PR adds support for a new UseHttp2 setting called PriorKnowledge,
this allows for serving non-tls http requests for both HTTP1.1 and HTTP2.0 on the same server/port.
If an incoming request sends the HTTP2.0 PRI preamble we service that request through an HTTP2 pipeline, otherwise we serve it though the HTTP1.1 pipeline.

## References

References #2523

## Changes

Adds a PriorKnowledgeSwitch in the same vein as AlpnSwitch, and then sniffs the initial bytes of each request to match against the HTTP2 Prior Knowledge preamble.

## Background Context

@raboof and @jrudolph pointed me in this direction.

This feature, assuming approval by the maintainers, will be getting a test suite to verify the implementation. Current testing has been done with `curl` to validate that it works with existing, external, clients.

use the curl flag `--http2-prior-knowledge` to have curl send the HTTP2 Prior Knowledge preamble with the request.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/javadsl/UseHttp2.scala
+++ b/akka-http-core/src/main/scala/akka/http/javadsl/UseHttp2.scala
@@ -17,4 +17,5 @@ object UseHttp2 {
   def always: UseHttp2 = Always
   def negotiated: UseHttp2 = Negotiated
   def never: UseHttp2 = Never
+  def priorKnowledge: UseHttp2 = PriorKnowledge
 }
--- a/akka-http-core/src/main/scala/akka/http/javadsl/UseHttp2.scala
+++ b/akka-http-core/src/main/scala/akka/http/javadsl/UseHttp2.scala
@@ -17,5 +17,4 @@ object UseHttp2 {
   def always: UseHttp2 = Always
   def negotiated: UseHttp2 = Negotiated
   def never: UseHttp2 = Never
-  def priorKnowledge: UseHttp2 = PriorKnowledge
 }
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
@@ -21,7 +21,7 @@ import akka.http.impl.engine.server._
 import akka.http.impl.engine.ws.WebSocketClientBlueprint
 import akka.http.impl.settings.{ ConnectionPoolSetup, HostConnectionPoolSetup }
 import akka.http.impl.util.StreamUtils
-import akka.http.scaladsl.UseHttp2.{ Always, Never, PriorKnowledge }
+import akka.http.scaladsl.UseHttp2.{ Always, Negotiated, Never }
 import akka.http.scaladsl.model._
 import akka.http.scaladsl.model.headers.Host
 import akka.http.scaladsl.model.ws.{ Message, WebSocketRequest, WebSocketUpgradeResponse }
@@ -318,11 +318,11 @@ class HttpExt private[http] (private val config: Config)(implicit val system: Ex
     settings:          ServerSettings    = ServerSettings(system),
     parallelism:       Int               = 0,
     log:               LoggingAdapter    = system.log)(implicit fm: Materializer): Future[ServerBinding] = {
-    val http2Enabled = settings.previewServerSettings.enableHttp2 && connectionContext.http2 != Never
-    val http2Encouraged = connectionContext.http2 == Always || connectionContext.http2 == PriorKnowledge
-    if (http2Enabled && (connectionContext.isSecure || http2Encouraged)) {
+    val http2Enabled = settings.previewServerSettings.enableHttp2
+    val http2Forced = connectionContext.http2 == Always
+    if (http2Enabled && connectionContext.http2 != Never) {
       // We do not support HTTP/2 negotiation for insecure connections (h2c), https://github.com/akka/akka-http/issues/1966
-      log.debug("Binding server using HTTP/2{}", if (http2Encouraged) " (forced to be used without TLS)" else "")
+      log.debug("Binding server using HTTP/2{}", if (http2Forced && !connectionContext.isSecure) " (forced to be used without TLS)" else "")
 
       val definitiveSettings =
         if (parallelism > 0) settings.mapHttp2Settings(_.withMaxConcurrentStreams(parallelism))
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
@@ -21,7 +21,7 @@ import akka.http.impl.engine.server._
 import akka.http.impl.engine.ws.WebSocketClientBlueprint
 import akka.http.impl.settings.{ ConnectionPoolSetup, HostConnectionPoolSetup }
 import akka.http.impl.util.StreamUtils
-import akka.http.scaladsl.UseHttp2.{ Always, Never }
+import akka.http.scaladsl.UseHttp2.{ Always, Never, PriorKnowledge }
 import akka.http.scaladsl.model._
 import akka.http.scaladsl.model.headers.Host
 import akka.http.scaladsl.model.ws.{ Message, WebSocketRequest, WebSocketUpgradeResponse }
@@ -318,11 +318,11 @@ class HttpExt private[http] (private val config: Config)(implicit val system: Ex
     settings:          ServerSettings    = ServerSettings(system),
     parallelism:       Int               = 0,
     log:               LoggingAdapter    = system.log)(implicit fm: Materializer): Future[ServerBinding] = {
-    val http2Enabled = settings.previewServerSettings.enableHttp2 && connectionContext.http2 != Never
-    val http2Forced = connectionContext.http2 == Always
-    if (http2Enabled && (connectionContext.isSecure || http2Forced)) {
+    val http2Enabled = true || settings.previewServerSettings.enableHttp2 && connectionContext.http2 != Never
+    val http2Encouraged = connectionContext.http2 == Always || connectionContext.http2 == PriorKnowledge
+    if (http2Enabled && (connectionContext.isSecure || http2Encouraged)) {
       // We do not support HTTP/2 negotiation for insecure connections (h2c), https://github.com/akka/akka-http/issues/1966
-      log.debug("Binding server using HTTP/2{}", if (http2Forced) " (forced to be used without TLS)" else "")
+      log.debug("Binding server using HTTP/2{}", if (http2Encouraged) " (forced to be used without TLS)" else "")
 
       val definitiveSettings =
         if (parallelism > 0) settings.mapHttp2Settings(_.withMaxConcurrentStreams(parallelism))
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
@@ -318,7 +318,7 @@ class HttpExt private[http] (private val config: Config)(implicit val system: Ex
     settings:          ServerSettings    = ServerSettings(system),
     parallelism:       Int               = 0,
     log:               LoggingAdapter    = system.log)(implicit fm: Materializer): Future[ServerBinding] = {
-    val http2Enabled = true || settings.previewServerSettings.enableHttp2 && connectionContext.http2 != Never
+    val http2Enabled = settings.previewServerSettings.enableHttp2 && connectionContext.http2 != Never
     val http2Encouraged = connectionContext.http2 == Always || connectionContext.http2 == PriorKnowledge
     if (http2Enabled && (connectionContext.isSecure || http2Encouraged)) {
       // We do not support HTTP/2 negotiation for insecure connections (h2c), https://github.com/akka/akka-http/issues/1966
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/UseHttp2.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/UseHttp2.scala
@@ -14,5 +14,4 @@ object UseHttp2 {
   object Never extends UseHttp2
   object Negotiated extends UseHttp2
   object Always extends UseHttp2
-  object PriorKnowledge extends UseHttp2
 }
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/UseHttp2.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/UseHttp2.scala
@@ -14,4 +14,5 @@ object UseHttp2 {
   object Never extends UseHttp2
   object Negotiated extends UseHttp2
   object Always extends UseHttp2
+  object PriorKnowledge extends UseHttp2
 }
--- a/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/PriorKnowledgeSwitch.scala
+++ b/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/PriorKnowledgeSwitch.scala
@@ -12,34 +12,31 @@ import akka.annotation.InternalApi
 import akka.http.impl.engine.server.HttpAttributes
 import akka.http.scaladsl.model.{ HttpRequest, HttpResponse }
 import akka.stream.TLSProtocol.{ SessionBytes, SessionTruncated, SslTlsInbound, SslTlsOutbound }
-import akka.stream.scaladsl.{ BidiFlow, Flow }
+import akka.stream.scaladsl.Flow
 import akka.stream.stage.{ GraphStage, GraphStageLogic, InHandler, OutHandler }
 import akka.stream._
 
 /** INTERNAL API */
 @InternalApi
 private[http] object PriorKnowledgeSwitch {
-  type HttpServerBidiFlow = BidiFlow[HttpResponse, ByteString, ByteString, HttpRequest, NotUsed]
-  type HttpServerBidiShape = BidiShape[HttpResponse, ByteString, ByteString, HttpRequest]
+  type HttpServerFlow = Flow[ByteString, ByteString, NotUsed]
+  type HttpServerShape = FlowShape[ByteString, ByteString]
 
   private final val PRIOR_KNOWLEDGE_PREFACE = ByteString("PRI * HTTP/2.0\r\n\r\nSM\r\n\r\n")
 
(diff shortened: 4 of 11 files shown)


raboof:
Related to #2085 (a more common but arguably less elegant way to support both http 1 and 2 on the same unencrypted port)

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/PriorKnowledgeSwitch.scala
```diff
@@ -0,0 +1,144 @@
...
+import akka.stream.stage.{ GraphStage, GraphStageLogic, InHandler, OutHandler }
+import akka.stream._
+
+/** INTERNAL API */
+@InternalApi
+private[http] object PriorKnowledgeSwitch {
+  type HttpServerFlow = Flow[ByteString, ByteString, NotUsed]
+  type HttpServerShape = FlowShape[ByteString, ByteString]
+
+  private final val PRIOR_KNOWLEDGE_PREFACE = ByteString("PRI * HTTP/2.0\r\n\r\nSM\r\n\r\n")
```
raboof: perhaps just call it `HTTP2_CONNECTION_PREFACE`, since it is sent in each HTTP2 connection

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/PriorKnowledgeSwitch.scala
```diff
@@ -0,0 +1,144 @@
...
+            private[this] var grabbed = ByteString.empty
+            def onPush(): Unit = {
+              val data = grabbed ++ grab(netIn)
+              if (data.length >= PRIOR_KNOWLEDGE_PREFACE.length) { // We should know by now
+                if (data.startsWith(PRIOR_KNOWLEDGE_PREFACE, 0))
+                  install(http2Stack, data)
+                else
+                  install(http1Stack, data)
+              } else if (data.isEmpty || data.startsWith(PRIOR_KNOWLEDGE_PREFACE, 0)) { // Still unknown
+                grabbed = data
```
raboof: I think this should append rather than assign?

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/PriorKnowledgeSwitch.scala
```diff
@@ -0,0 +1,144 @@
...
+          setHandler(netIn, new InHandler {
+            private[this] var grabbed = ByteString.empty
+            def onPush(): Unit = {
+              val data = grabbed ++ grab(netIn)
+              if (data.length >= PRIOR_KNOWLEDGE_PREFACE.length) { // We should know by now
+                if (data.startsWith(PRIOR_KNOWLEDGE_PREFACE, 0))
+                  install(http2Stack, data)
+                else
+                  install(http1Stack, data)
+              } else if (data.isEmpty || data.startsWith(PRIOR_KNOWLEDGE_PREFACE, 0)) { // Still unknown
```
raboof: `data.startsWith(PRIOR_KNOWLEDGE_PREFACE, 0)` will always be false here, since `data.length >= PRIOR_KNOWLEDGE_PREFACE.length` was already `false`, I think?

raboof: (looks like the intent was to check whether the data received so far matches the preface, so we can switch to http1 as soon as we see a mismatch - which makes sense)

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/PriorKnowledgeSwitch.scala
```diff
@@ -0,0 +1,144 @@
...
+import akka.http.impl.engine.server.HttpAttributes
+import akka.http.scaladsl.model.{ HttpRequest, HttpResponse }
+import akka.stream.TLSProtocol.{ SessionBytes, SessionTruncated, SslTlsInbound, SslTlsOutbound }
+import akka.stream.scaladsl.Flow
+import akka.stream.stage.{ GraphStage, GraphStageLogic, InHandler, OutHandler }
+import akka.stream._
+
+/** INTERNAL API */
+@InternalApi
+private[http] object PriorKnowledgeSwitch {
```
raboof: Seems pretty verbose for what it does, but I can't really think of a more straightforward way either.

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
```diff
@@ -318,11 +318,11 @@ class HttpExt private[http] (private val config: Config)(implicit val system: Ex
     settings:          ServerSettings    = ServerSettings(system),
     parallelism:       Int               = 0,
     log:               LoggingAdapter    = system.log)(implicit fm: Materializer): Future[ServerBinding] = {
-    val http2Enabled = settings.previewServerSettings.enableHttp2 && connectionContext.http2 != Never
-    val http2Forced = connectionContext.http2 == Always
-    if (http2Enabled && (connectionContext.isSecure || http2Forced)) {
+    val http2Enabled = true || settings.previewServerSettings.enableHttp2 && connectionContext.http2 != Never
```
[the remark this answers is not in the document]
viktorklang: haha, good catch! :D

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/PriorKnowledgeSwitch.scala
viktorklang: :+1:

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/PriorKnowledgeSwitch.scala
viktorklang: I'd be more than happy to cut down the noise here, open to all suggestions!

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/PriorKnowledgeSwitch.scala
viktorklang: @raboof Indeed!

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/PriorKnowledgeSwitch.scala
viktorklang: It's already appended at that point: https://github.com/akka/akka-http/pull/2543/files/97dc3d2970b3e2db8d91d9f26060d6527fef2168#diff-a4285c89230780862b0dd25f0676f59eR54

akka-http-core/src/main/scala/akka/http/javadsl/UseHttp2.scala
```diff
@@ -17,4 +17,5 @@ object UseHttp2 {
   def always: UseHttp2 = Always
   def negotiated: UseHttp2 = Negotiated
   def never: UseHttp2 = Never
+  def priorKnowledge: UseHttp2 = PriorKnowledge
```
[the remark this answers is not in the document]
viktorklang: Perhaps consolidate that behavior once both PRs are merged? (to make them independent)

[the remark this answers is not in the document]
raboof: sure, just throwing this out there since finding a better name now is nicer than changing it (or not ;) ) later ;)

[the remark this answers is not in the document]
viktorklang: @raboof I didn't find any preexisting good place to write tests for this, is there somewhere you'd recommend?

viktorklang:
@raboof @jrudolph This is ready for final review!

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/WithPriorKnowledgeSpec.scala
```diff
@@ -0,0 +1,83 @@
...
+      source.offer("UFJJICogSFRUUC8yLjANCg0KU00NCg0KAAASBAAAAAAAAAMAAABkAARAAAAAAAIAAAAAAAAECAAAAAAAP/8AAQAAHgEFAAAAAYKEhkGKCJ1cC4Fw3HwAf3qIJbZQw6u20uBTAyovKg==").futureValue
+
+      // read settings frame
+      Http2Protocol.FrameType.byId(sink.pull().futureValue.get(3)) should be(Http2Protocol.FrameType.SETTINGS)
+      // read settings frame
+      Http2Protocol.FrameType.byId(sink.pull().futureValue.get(3)) should be(Http2Protocol.FrameType.SETTINGS)
+      // ack settings
+      source.offer("AAAABAEAAAAA")
+
+      // source.complete()
```
raboof:
```suggestion
```
(my bad ;) )

viktorklang:
@raboof Feel free to merge as long as the latest test suite sweep is green

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/WithPriorKnowledgeSpec.scala
```diff
@@ -0,0 +1,82 @@
...
+import java.util.Base64
+import akka.stream.scaladsl.{ Sink, Source, Tcp }
+import akka.testkit.AkkaSpec
+import akka.util.ByteString
+
+import scala.concurrent.Future
+import akka.stream.OverflowStrategy
+import akka.stream.scaladsl.{ SinkQueue, Keep }
+
+class WithPriorKnowledgeSpec extends AkkaSpec("""
```
jrudolph: The HTTP/2 internals are somewhat well tested in `Http2ServerSpec` (without adding the transport layers). I think we are mostly missing any kind of end-to-end tests for HTTP/2. I guess we could keep the test like this for now but create a ticket to evolve it into more end-to-end like tests that could reuse all the high-level testing tools from `Http2ServerSpec`.

akka-http-core/src/main/scala/akka/http/scaladsl/UseHttp2.scala
```diff
@@ -14,4 +14,5 @@ object UseHttp2 {
   object Never extends UseHttp2
   object Negotiated extends UseHttp2
   object Always extends UseHttp2
+  object PriorKnowledge extends UseHttp2
```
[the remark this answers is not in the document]
viktorklang: @jrudolph OTOH, is there any reason why you wouldn't want to run both HTTP/1.1 and HTTP/2 at all time?

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/PriorKnowledgeSwitch.scala
viktorklang: @jrudolph I thought about that, but then again, the AlpnSwitch deals with Bidi and this works with Flow, so I think the correct move in that case would be to abstract each separately so that you'd be reusable for other stages. That does seem orthogonal to this PR though?

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/WithPriorKnowledgeSpec.scala
viktorklang: @jrudolph Agreed.

viktorklang:
@jrudolph @raboof 

How about merging this as is, and then create an Issue to collapse the UseHttp2 flags (ideally deprecating all of them and having things just work by inspecting the isSecure option).

jrudolph:
> How about merging this as is, and then create an Issue to collapse the UseHttp2 flags (ideally deprecating all of them and having things just work by inspecting the isSecure option).

It doesn't seem pressing to merge it right now, so I'd rather improve it first and then merge it, instead of adding a rarely-used feature to our pile of code we have no capacity to maintain. Especially, the public API part needs to be figured out, since `UseHttp2` doesn't seem to be marked `@ApiMayChange`.

akka-http-core/src/main/scala/akka/http/scaladsl/UseHttp2.scala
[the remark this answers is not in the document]
raboof: I suspect there might be a performance penalty to that?

viktorklang:
@jrudolph Well, I need it for StateServ: https://github.com/lightbend/stateful-serverless/commit/a6b6b706557f67d011ca49cdd8d733db99205343

akka-http-core/src/main/scala/akka/http/scaladsl/UseHttp2.scala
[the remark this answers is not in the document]
viktorklang: @jrudolph @raboof If we collapse this feature into Negotiated, what would that look like in practice?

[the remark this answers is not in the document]
raboof: I'll give it a go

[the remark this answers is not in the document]
viktorklang: @raboof :+1: let me know when/if you want me to review :)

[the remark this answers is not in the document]
raboof: Tried that in https://github.com/akka/akka-http/pull/2553

I think a fourth option makes sense. That would make the matrix:
* `Never`: no HTTP/2 support, no overhead relative to just HTTP/1.1
* `Negotiated`: ALPN when secure, support for h2c via 'Upgrade' when unsecure, no/low overhead compared to just HTTP/1.1
* `PriorKnowledge`: n/a when secure, support for both h2c mechanism when unsecure, some overhead (from `PriorKnowledgeSwitch`)
* `Always`: no HTTP/1.1 support

Am I overestimating the performance penalty of `PriorKnowledgeSwitch`?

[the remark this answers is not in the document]
viktorklang: I suspect that this might get even more complicated once HTTP/3 is released.
How about we just add so that the user lists what protocols they want to use? If the concern is purely a performance one, then perhaps work instead should be placed into minimizing overhead of the implementation of the features orthogonally to the configuration?

protocols = [HTTP1] == Never
protocols = [HTTP1, HTTP2] == Negotiated
protocols = [HTTP2] == Always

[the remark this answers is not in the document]
jrudolph: I would even simplify things even more: either you disable HTTP/2 (default), so there's no performance penalty for HTTP/1.1, or you enable HTTP/2 in addition to HTTP/1.1, and you get a small performance penalty for the switches. I don't see the reason why we necessarily would have to support `Always` at all, since the only added feature over `Negotiated` is failing when someone tries to connect with HTTP/1.1 (+ the performance benefit which alone might not be worth adding a feature for it).

The benefit would be that we remove UseHttp2 completely and just use the `enable-http2` flag.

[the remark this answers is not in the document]
viktorklang: @jrudolph Sounds good to me! Less code to maintain and less permutations to worry about.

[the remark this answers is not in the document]
raboof: > The benefit would be that we remove UseHttp2 completely and just use the enable-http2 flag

Yes that sounds good, since `enable-http2` can still be tweaked via `ServerSettings` if really needed.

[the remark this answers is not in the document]
viktorklang: @raboof @jrudolph Let's do it!

viktorklang:
@raboof @jrudolph Let me know if you want me to do the honors off merging!

jrudolph:
I'd like to have a try at removing the code duplication in the switches.

viktorklang:
@jrudolph Ok, cool, let me know if you want me to review it!

jrudolph:
Thanks, @viktorklang, that was some quick turnaround on that feature.

viktorklang:
Thanks @jrudolph, I couldn't have completed it in such short amount of time without the great advice and contributions from the Akka team! ^^

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/PriorKnowledgeSwitch.scala
```diff
@@ -0,0 +1,141 @@
...
+          setHandler(netIn, new InHandler {
+            private[this] var grabbed = ByteString.empty
+            def onPush(): Unit = {
+              val data = grabbed ++ grab(netIn)
+              if (data.length >= HTTP2_CONNECTION_PREFACE.length) { // We should know by now
+                if (data.startsWith(HTTP2_CONNECTION_PREFACE, 0))
+                  install(http2Stack, data)
+                else
+                  install(http1Stack, data)
+              } else if (data.isEmpty || data.startsWith(HTTP2_CONNECTION_PREFACE, 0)) { // Still unknown
```
jroper: Shouldn't this be `HTTP2_CONNECTION_PREFACE.startsWith(data, 0)`? `data.startsWith(HTTP2_CONNECTION_PREFACE, 0)` will always evaluate to false, since we know that `data.length < HTTP2_CONNECTION_PREFACE.length`.

raboof: Jup, I'll PR

viktorklang:
Wow—nicely spotted, James!
-- 
Cheers,
√

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/PriorKnowledgeSwitch.scala
jrudolph: We can fix that in my PR directly.

raboof: https://github.com/akka/akka-http/pull/2558

Issue #2548, opened 2019-05-29
Use status code 408 if extractStrictEntity fails with timeout


jrudolph:
See https://discuss.lightbend.com/t/consider-http-status-code-408-for-timeout-in-extractstrictentity/4267

psisoyev:
I would like to take this one

PR #2549, opened 2019-05-29
Return 408 RequestTimeout status when `toStrict` times out #2548

psisoyev:
## Purpose

Now if extractStrictEntity fails with a timeout it will return 408 Request Timeout status instead of 500 Internal Service Error

## References

https://github.com/akka/akka-http/issues/2548

## Changes

Added TimeoutException handler for toStrict method

### DIFF
--- a/akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/BasicDirectivesSpec.scala
+++ b/akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/BasicDirectivesSpec.scala
@@ -182,5 +182,19 @@ class BasicDirectivesSpec extends RoutingSpec {
         }
       } ~> check { responseAs[String] shouldEqual "1" }
     }
+
+    "return 408 Request Timeout status on timeout" in {
+      val neverEndingEntity = HttpEntity(ContentTypes.`text/plain(UTF-8)`, Source.maybe[ByteString])
+      val timeout = 10.milliseconds
+
+      Post("/abc", neverEndingEntity) ~> {
+        extractStrictEntity(timeout) { _ =>
+          complete("content")
+        }
+      } ~> check {
+        entityAs[String] shouldEqual s"Request timed out after $timeout while waiting for entity data"
+        status shouldEqual StatusCodes.RequestTimeout
+      }
+    }
   }
 }
--- a/akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/BasicDirectivesSpec.scala
+++ b/akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/BasicDirectivesSpec.scala
@@ -5,7 +5,7 @@
 package akka.http.scaladsl.server
 package directives
 
-import java.util.concurrent.ThreadLocalRandom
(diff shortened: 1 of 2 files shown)


raboof:
OK TO TEST

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/BasicDirectivesSpec.scala
```diff
@@ -182,5 +182,19 @@ class BasicDirectivesSpec extends RoutingSpec {
...
+    "return 408 Request Timeout status on timeout" in {
+      val neverEndingEntity = HttpEntity(ContentTypes.`text/plain(UTF-8)`, Source.maybe[ByteString])
+      val timeout = 10.milliseconds
+
+      Post("/abc", neverEndingEntity) ~> {
+        extractStrictEntity(timeout) { _ =>
+          complete("content")
+        }
+      } ~> check {
+        entityAs[String] shouldEqual s"Request timed out after $timeout while waiting for entity data"
```
raboof: I think I would prefer to make the timeout explicit in this check rather than invoking the `toString` on the `timeout` - but this is OK as well.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/BasicDirectives.scala
```diff
@@ -398,6 +400,11 @@ trait BasicDirectives {
       ctx.request.entity.toStrict(timeout, maxBytes).flatMap { strictEntity =>
         val newCtx = ctx.mapRequest(_.copy(entity = strictEntity))
         inner(())(newCtx)
+      }.recover {
+        case _: TimeoutException =>
+          throw IllegalRequestException(
```
raboof: Is `TimeoutException` specifically thrown when strictify-ing the request times out, or might it also occur when it's the generation of the response that times out? For the former indeed an 4xx Client Error `IllegalRequestException` is approprate, for the latter arguably a 5xx Server Error might be better.

Do we have a test for that scenario? If not that might be useful to add.

jrudolph: We discussed a bit whether throwing an `IllegalRequestException` would be good enough vs. introducing a specific exception or even a rejection so that user code would have more control over how to handle that situation.

One argument for not making it a rejection is that a request that failed `toStrict` cannot really be handled in an alternative way and shouldn't run further through the routing tree in my opinion.

@raboof wdyt?

jrudolph: That's a good point. The solution then would be to swap the flatMap and the recover so that the recover only covers the `toStrict` and not the handling further down.

raboof: I agree a rejection isn't helpful in this case. A specific exception is not strictly needed but might be nice, that's up to the contributor AFAIC.

psisoyev: I will cover this

Issue #2547, opened 2019-05-28
Failure: 2.13.0-RC2 nightlies fail on JDK 8


jrudolph:
E.g. https://jenkins.akka.io:8498/job/akka-http-nightly/AKKA_VERSION=default,Node=akka-http-nightly-node,SCALA_VERSION=2.13.0-RC2,jdk=AdoptOpenJDK%208/500/console

That's because spray-json 1.3.5 was released building against a JDK 11 API. Will need a new spray-json release build against JDK8.

```
[error] java.lang.NoSuchMethodError: java.nio.ByteBuffer.flip()Ljava/nio/ByteBuffer;
[error] 	at spray.json.ParserInput$IndexedBytesParserInput.decode$1(JsonParser.scala:313)
[error] 	at spray.json.ParserInput$IndexedBytesParserInput.nextUtf8Char(JsonParser.scala:332)
[error] 	at spray.json.JsonParser.string(JsonParser.scala:167)
[error] 	at spray.json.JsonParser.value(JsonParser.scala:78)
[error] 	at spray.json.JsonParser.members$1(JsonParser.scala:96)
[error] 	at spray.json.JsonParser.object(JsonParser.scala:100)
[error] 	at spray.json.JsonParser.value(JsonParser.scala:75)
[error] 	at spray.json.JsonParser.parseJsValue(JsonParser.scala:51)
[error] 	at spray.json.JsonParser.parseJsValue(JsonParser.scala:47)
[error] 	at spray.json.JsonParser$.apply(JsonParser.scala:30)
[error] 	at akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport.$anonfun$sprayJsValueByteStringUnmarshaller$3(SprayJsonSupport.scala:42)
```

raboof:
(travis still tests 2.10 and 2.11)

jrudolph:
Should be fixed by #2550 (and a new release of spray-json built with JDK 8).

PR #2550, opened 2019-05-31
Update to Scala 2.13.0-RC3

raboof:
(might interfere with playframework nightlies, let's check before merging)


raboof:
(play is fine depending on released snapshots, so this is good to go)

raboof:
closes #2547

PR #2553, opened 2019-06-05
Roll together 'prior knowledge' and 'negotiated'

raboof:
Just for discussion, I'm not sure we should do this.
https://github.com/akka/akka-http/pull/2543#discussion_r290340917

I'm hesitant because `Negotiated` is the default, and adding the
`PriorKnowledgeSwitch` might have a performance penalty on top of 'plain'
HTTP/1.

For the same reason I'm not sure we should roll it into `Always`.

## References

References #xxxx

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/javadsl/UseHttp2.scala
+++ b/akka-http-core/src/main/scala/akka/http/javadsl/UseHttp2.scala
@@ -17,5 +17,4 @@ object UseHttp2 {
   def always: UseHttp2 = Always
   def negotiated: UseHttp2 = Negotiated
   def never: UseHttp2 = Never
-  def priorKnowledge: UseHttp2 = PriorKnowledge
 }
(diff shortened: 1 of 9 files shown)


jrudolph:
> I'm hesitant because `Negotiated` is the default, and adding the
> `PriorKnowledgeSwitch` might have a performance penalty on top of 'plain'
> HTTP/1.

You would still have to enable http2 and depend on the akka-http2-support module, wouldn't you? And in that case, you have already decided about the consequences. This would only make it more consistent with the https side of things which has the same sort of consequences if you enable http2.

jrudolph:
If we want to support `Always`, we should then also make sure that it works consistently for http and https. So, either, we should remove it (when we merge PriorKnowledge) or make sure that it fails properly and consistently for both http and https.

viktorklang:
@raboof @jrudolph I think it's fine to roll this into Negotiated for the reasons @jrudolph outlined.

raboof:
> You would still have to enable http2 and depend on the akka-http2-support module, wouldn't you? And in that case, you have already decided about the consequences.

Good point, agreed. And if you really want to have HTTP/2 support but also avoid the penalty of supporting both on the same port, you can do that with `Never` and `Always`.

> make sure that contacting an Http/2 Always server fails properly and consistently for both http and https

Agreed, I think it makes sense to support that, added #2555

PR #2556, opened 2019-06-05
Retire UseHttp2 in favour of enable-http2 boolean

raboof:


raboof:
:+1: then I'll migrate #2558 after that :smile:

PR #2557, opened 2019-06-05
Consolidate http2 switches and instantiation code in Http2

jrudolph:
A slight downside is that the plaintext layers will use some extra wrapping and unwrapping into TLS types but that's consistent with what the Http/1.1 side also does.


akka-http2-support/src/main/scala/akka/http/scaladsl/Http2.scala
```diff
@@ -94,125 +42,105 @@ final class Http2Ext(private val config: Config)(implicit val system: ActorSyste
...
+  type HttpPlusSwitching = (HttpImplementation, HttpImplementation) => Flow[ByteString, ByteString, NotUsed]
 
-    val connections = Tcp().bind(interface, port, settings.backlog, settings.socketOptions, halfClose = false, settings.timeouts.idleTimeout)
+  def onlyHttp2(http1: HttpImplementation, http2: HttpImplementation): Flow[ByteString, ByteString, NotUsed] =
+    TLSPlacebo().reversed join http2
 
-    val masterTerminator = new MasterServerTerminator(log)
+  def priorKnowledge(http1: HttpImplementation, http2: HttpImplementation): Flow[ByteString, ByteString, NotUsed] = {
+    def chooseProtocol(sessionBytes: SessionBytes): String =
+      if (sessionBytes.bytes.startsWith(Http2Protocol.ClientConnectionPreface)) "h2" else "http/1.1"
```
jrudolph: We only support the case here where the full preface is received in one go (in contrast to the `PriorKnowledgeSwitch`). I'd say it's sensible enough to just require that it's received in one go. Alternatively, we can change `AlpnSwitch` to gather a fixed or configurable number of bytes before making a protocol decision.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/AlpnSwitch.scala
```diff
@@ -18,35 +18,29 @@ import akka.stream._
 /** INTERNAL API */
 @InternalApi
 private[http] object AlpnSwitch {
```
raboof: Since it's now more generic, we should probably give it a more generic name - `ProtocolSwitch`?

jrudolph:
The failure was a legitimate failure that was previously masked by a timeout in H2SpecIntegrationSpec. I added a few commits fixing and adding to the tests.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/AlpnSwitch.scala
```diff
@@ -18,35 +18,29 @@ import akka.stream._
...
-  type HttpServerBidiFlow = BidiFlow[HttpResponse, SslTlsOutbound, SslTlsInbound, HttpRequest, NotUsed]
+  type HttpServerFlow = Flow[SslTlsInbound, SslTlsOutbound, NotUsed]
```
raboof: Cool that it even got a bit simpler!

akka-http2-support/src/main/scala/akka/http/scaladsl/Http2.scala
raboof: I think this would be reasonable to support in the future, but that can be a future improvement.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -815,12 +815,9 @@ class Http2ServerSpec extends AkkaSpec("""
     "respect the substream state machine" should {
       abstract class SimpleRequestResponseRoundtripSetup extends TestSetup with RequestResponseProbes
 
-      "reject other frame than HEADERS/PUSH_PROMISE in idle state with connection-level PROTOCOL_ERROR (5.1)" inPendingUntilFixed new SimpleRequestResponseRoundtripSetup {
+      "reject other frame than HEADERS/PUSH_PROMISE in idle state with connection-level PROTOCOL_ERROR (5.1)" in new SimpleRequestResponseRoundtripSetup {
```
raboof: nice

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/AlpnSwitch.scala
jrudolph: Indeed!

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/AlpnSwitch.scala
jrudolph: Good point. Done.

PR #2558, opened 2019-06-06
Add some Prior Knowledge switch tests

raboof:
https://github.com/akka/akka-http/pull/2543#discussion_r287314992
https://github.com/akka/akka-http/pull/2543#discussion_r290989734

### DIFF
--- a/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/ProtocolSwitch.scala
+++ b/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/ProtocolSwitch.scala
@@ -9,9 +9,8 @@ import javax.net.ssl.SSLException
 import akka.NotUsed
 import akka.annotation.InternalApi
 import akka.http.impl.engine.server.HttpAttributes
-import akka.http.scaladsl.model.{ HttpRequest, HttpResponse }
 import akka.stream.TLSProtocol.{ SessionBytes, SessionTruncated, SslTlsInbound, SslTlsOutbound }
-import akka.stream.scaladsl.{ BidiFlow, Flow }
+import akka.stream.scaladsl.Flow
 import akka.stream.stage.{ GraphStage, GraphStageLogic, InHandler, OutHandler }
 import akka.stream._
 
@@ -135,4 +134,10 @@ private[http] object ProtocolSwitch {
         }
       }
     )
+
+  def byPreface(http1Stack: HttpServerFlow, http2Stack: HttpServerFlow): HttpServerFlow = {
+    def chooseProtocol(sessionBytes: SessionBytes): String =
+      if (sessionBytes.bytes.startsWith(Http2Protocol.ClientConnectionPreface)) "h2" else "http/1.1"
+    ProtocolSwitch(chooseProtocol, http1Stack, http2Stack)
+  }
 }
--- a/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Protocol.scala
+++ b/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Protocol.scala
@@ -353,12 +353,11 @@ private[http] object Http2Protocol {
    *  which in hex notation is:
    *
    *     0x505249202a20485454502f322e300d0a0d0a534d0d0a0d0a
+   *
+   * That is, the connection preface starts with the string "PRI * HTTP/2.0\r\n\r\nSM\r\n\r\n".
    */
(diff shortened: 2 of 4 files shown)


raboof:
> Maybe you could change the PR to only add tests to those, then it doesn't conflict with #2557?

I think we can simply close this one when merging #2557, and add similar tests when adding this to the consolidated switch

jrudolph:
> I think we can simply close this one when merging #2557, and add similar tests when adding this to the consolidated switch

Sounds good, that would basically be solving the conflict when we merge #2557.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/PriorKnowledgeSwitchSpec.scala
```diff
@@ -0,0 +1,106 @@
...
+import akka.stream.scaladsl.Keep
+import akka.stream.scaladsl.Source
+import akka.stream.scaladsl.Sink
+import akka.util.ByteString
+
+import akka.testkit.AkkaSpec
+import org.scalatest.exceptions.TestFailedException
+import org.scalatest.time.{ Span, Second, Milliseconds }
+
+class PriorKnowledgeSwitchSpec extends AkkaSpec {
```
jroper: One good test to add here for regression testing purposes is an HTTP/1.0 request with no headers (HTTP/1.1 requires a Host header, so won't work). Eg:

```
GET / HTTP/1.0

```

The reason that test would be good is because that request is valid, and is shorter than the HTTP/2 preface. There's a possible optimisation that would break this request, that is to wait until you have enough bytes before comparing to the HTTP/2 preface, rather than doing the startsWith comparison on each set of bytes grabbed before the preface length is reached. With that optimisation in place, the above request will hang . A test case that tests this will ensure no one makes this optimisation in future.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/PriorKnowledgeSwitch.scala
```diff
@@ -52,13 +44,14 @@ private[http] object PriorKnowledgeSwitch {
...
               val data = grabbed ++ grab(netIn)
-              if (data.length >= HTTP2_CONNECTION_PREFACE.length) { // We should know by now
-                if (data.startsWith(HTTP2_CONNECTION_PREFACE, 0))
+              if (data.length >= Http2Protocol.ClientConnectionPreface.length) { // We should know by now
+                if (data.startsWith(Http2Protocol.ClientConnectionPreface, 0))
                   install(http2Stack, data)
                 else
                   install(http1Stack, data)
-              } else if (data.isEmpty || data.startsWith(HTTP2_CONNECTION_PREFACE, 0)) { // Still unknown
+              } else if (data.isEmpty || Http2Protocol.ClientConnectionPreface.startsWith(data)) { // Still unknown
```
viktorklang:
```suggestion
              } else if (Http2Protocol.ClientConnectionPreface.startsWith(data)) { // Still unknown
```

Not needed to check for data.isEmpty here since startsWith("") returns true.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/ProtocolSwitch.scala
```diff
@@ -135,4 +134,10 @@ private[http] object ProtocolSwitch {
         }
       }
     )
+
+  def byPreface(http1Stack: HttpServerFlow, http2Stack: HttpServerFlow): HttpServerFlow = {
+    def chooseProtocol(sessionBytes: SessionBytes): String =
+      if (sessionBytes.bytes.startsWith(Http2Protocol.ClientConnectionPreface)) "h2" else "http/1.1"
```
viktorklang: @raboof Do we know that we have enough bytes here?

raboof: Currently we don't - this PR only adds some tests (including an ignored one for that very scenario), but doesn't fix it.

viktorklang: Ok!

viktorklang:
@raboof Ready to merge this?

raboof:
@viktorklang it's still waiting for its second LGTM, but I suppose it's uncontroversial enough that we can merge it now

viktorklang:
@raboof Awesome, thanks!

PR #2559, opened 2019-06-06
 WIP +htp add ifUnauthenticated variant of authentication directives 

### LINKED ISSUE #2554: Provide a `unauthenticatedOnly` directive
It's somewhat hard to provide a directive that is the inverse of the authentication directives (e.g. to show a login page only when unauthenticated).

See here for one attempt: https://gist.github.com/bmc/3980a551abcc5675778f8413d9a640a8

How could we provide that out of the box? E.g. by providing a wrapper directive `ifUnauthenticated(authenticationDirective)`?

jrudolph:
Fixes #2554

Missing:
 * examples and documentation
 * release note entry about behavior change of optional for `CredentialsRejected` case

### DIFF
--- a/akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/SecurityDirectivesSpec.scala
+++ b/akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/SecurityDirectivesSpec.scala
@@ -12,133 +12,80 @@ import akka.http.scaladsl.server.AuthenticationFailedRejection.{ CredentialsReje
 import akka.testkit.EventFilter
 
 class SecurityDirectivesSpec extends RoutingSpec {
-  val dontBasicAuth = authenticateBasicAsync[String]("MyRealm", _ => Future.successful(None))
-  val dontOAuth2Auth = authenticateOAuth2Async[String]("MyRealm", _ => Future.successful(None))
   val doBasicAuth = authenticateBasicPF("MyRealm", { case Credentials.Provided(identifier) => identifier })
-  val doOAuth2Auth = authenticateOAuth2PF("MyRealm", { case Credentials.Provided(identifier) => identifier })
-  val authWithAnonymous = doBasicAuth.withAnonymousUser("We are Legion")
-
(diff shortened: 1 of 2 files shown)


akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/SecurityDirectivesSpec.scala
```diff
@@ -12,133 +12,110 @@ import akka.http.scaladsl.server.AuthenticationFailedRejection.{ CredentialsReje
 import akka.testkit.EventFilter
 
 class SecurityDirectivesSpec extends RoutingSpec {
-  val dontBasicAuth = authenticateBasicAsync[String]("MyRealm", _ => Future.successful(None))
-  val dontOAuth2Auth = authenticateOAuth2Async[String]("MyRealm", _ => Future.successful(None))
   val doBasicAuth = authenticateBasicPF("MyRealm", { case Credentials.Provided(identifier) => identifier })
-  val doOAuth2Auth = authenticateOAuth2PF("MyRealm", { case Credentials.Provided(identifier) => identifier })
-  val authWithAnonymous = doBasicAuth.withAnonymousUser("We are Legion")
-
+  val dontBasicAuth = authenticateBasicAsync[String]("MyRealm", _ => Future.successful(None))
```
raboof: I know this is a Draft, but I'm not sure this does what we are looking for: when an unauthenticated request comes in, this will still reject the request with a challenge, right? I think the desired behavior would be that it passes through in that case.

jrudolph: Not sure? There are tests below, which one do you think doesn't do what it should? Do you mean "pass ifUnauthenticated when authentication data is missing"?

raboof: Ha, indeed that didn't do what I assumed :D

Issue #2409, opened 2019-01-30
Don't include sensitive user data in `toString` output


pfcoperez:
There are certain environments where it is easy to either print, write or log `HttpRequest` or `HttpResponse` instances  (for example when side-effecting in the passed function when using directives such as [`mapRequest`](https://doc.akka.io/docs/akka-http/current/routing-dsl/directives/basic-directives/mapRequest.html) or [`mapRequestContext`](https://doc.akka.io/docs/akka-http/current/routing-dsl/directives/basic-directives/mapRequestContext.html)).

Both of these classes override `toString` as:

https://github.com/akka/akka-http/blob/6cac8db36f25cfe2408b0517100d534324eed704/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala#L352

and 

https://github.com/akka/akka-http/blob/6cac8db36f25cfe2408b0517100d534324eed704/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala#L499

respectively. This means that the contents of the headers, for example, could inadvertently leak into log files. For some industrial environments this poses legal risks.

This issue aims to suggest adding a template method to `HttpMessage` to allow customizing how `toString` behaves in each case. Surely, defaulting to the above listed implementations.

For example:

```scala
/**
 * Common base class of HttpRequest and HttpResponse.
 */
sealed trait HttpMessage extends jm.HttpMessage {
  type Self <: HttpMessage { type Self = HttpMessage.this.Self }
  def self: Self

  def isRequest: Boolean
  def isResponse: Boolean

  def headers: immutable.Seq[HttpHeader]
  def entity: ResponseEntity
  def protocol: HttpProtocol
  def stringRepresentation: Self ⇒ String // <---------------------  NEW

  override def toString: String = stringRepresentation(self) // <---------------------  NEW

  /** Returns a copy of this message where `f` is used to generate its string representation. **/
  def withStringRepresentation(f: Self ⇒ String): Self  // <---------------------  NEW

...

```

And 

```scala
/**
 * The immutable model HTTP request model.
 */
final class HttpRequest(
  val method:   HttpMethod,
  val uri:      Uri,
  val headers:  immutable.Seq[HttpHeader],
  val entity:   RequestEntity,
  val protocol: HttpProtocol,
  val stringRepresentation: HttpRequest ⇒ String = { request ⇒ // <---------------------  NEW
    import request._
[...]

PR #2560, opened 2019-06-06
Hide body and headers by default in `HttpRequest#toString` and `HttpResponse#toString`

pfcoperez:
## Purpose

Aimed to avoid PII and SPI leakage, closes https://github.com/akka/akka-http/issues/2409

## References

References https://github.com/akka/akka-http/pull/2412
References https://github.com/akka/akka-http/issues/2409#issuecomment-496417344
References https://github.com/akka/akka-http/pull/2412#pullrequestreview-207325113

## Changes

-  Makes `HttpResponse#toString` and `HttpRequest#toString` omit bodies and headers from the default string representation thus avoiding potential _PII_ and _SPI_ in logs and outputs. 
- Adds documentation and examples explaining how to use type classes to expand the trimmed `toString` behaviour.

## Background Context

It tackles the same problem as https://github.com/akka/akka-http/pull/2412 but with a different approach:  Default to not showing and show via additional behaviour (e.g using type classes). Also, this PR includes the documentation to add this extra behaviour as well as some examples.


raboof:
OK TO TEST

docs/src/main/paradox/common/http-model.md
```diff
@@ -67,6 +67,19 @@ simplify creating requests for common cases. Also, to aid readability, there are
 named after HTTP methods to create a request with a given method and URI directly.
 @@@
 
+@@@ note { title='String representation' }
+
+There are certain environments where it is easy to inadvertently print, write or log entries built out of string representations of @apidoc[HttpRequest] instances. On the other hand, it is not uncommon for HTTP headers and entities to contain _Personal Identifying Information (PII)_ or _Sensitive Personal Information (SPI)_ . 
+
+For all this, if @apidoc[HttpRequest] `toString` method included body and/or headers, it would be easy to accidentally violate privacy laws such as _GDPR_. These fields are, therefore, omitted from the string representation.
```
raboof: I'm not sure we should be making such legal-ish statements in our documentation - perhaps leave it at "To avoid accidentally leaking such information into the logging these fields are omitted from the `toString` output."?

project/Dependencies.scala
```diff
@@ -125,7 +127,7 @@ object Dependencies {
 
   lazy val httpJackson = l ++= Seq(jackson)
 
-  lazy val docs = l ++= Seq(Docs.sprayJson, Docs.gson, Docs.jacksonXml, Docs.reflections)
+  lazy val docs = l ++= Seq(Docs.sprayJson, Docs.gson, Docs.jacksonXml, Docs.reflections, Docs.cats, Docs.scalaz)
```
raboof: I'm not sure we want to show cats/scalaz in our documentation - people who are using that probably already know how to create such typeclasses?

docs/src/main/paradox/common/http-model.md
pfcoperez: > erhaps leave it at "To avoid accidentally leaking such information into the logging these fields are omitted from the toString output."

:100: Changing to that.

project/Dependencies.scala
pfcoperez: Should I remove Scalaz and Cats examples from `HttpRequestShow` and `HttpResponseShow`. I kind of like the idea of having both options explained:

Some Cats and Scalaz users are unaware of the `Show[T]` typeclass.

Non-users might be completely unaware of the fact that these libraries exist or that they provide typeclasses and syntax to add to-string methods on top of existing classes and think: _Oh, well now I have to write a trait, and object, implicit instances and implicit conversions to have what I used to have_.

raboof: Not sure - I can see how it might give the reader an interesting pointer, on the other hand it increases the scope of the docs making them harder to understand and to maintain. Perhaps another reviewer can weigh in?

> now I have to write a trait, and object, implicit instances and implicit conversions to have what I used to have

I'm not sure they'd have to go that full mile - they could just write a function?

pfcoperez: I've no problem with removing the Cats and Scalaz examples and adding simpler ones (using vanilla Scala) + keeping the  from-scratch type class.

pfcoperez: This commit https://github.com/akka/akka-http/pull/2560/commits/409bbc98044e0bcbaa84f6b7cb61a2eb6389958b 

Removes Cats and Scalaz dependencies and the examples using them. I initially included these because I liked @jypma suggestion in https://github.com/akka/akka-http/pull/2412#issuecomment-466986505 . 

But I also understand that you don't want to introduce additional dependencies.

In any case, this is done in the last commit so if you finally decide that you want to include the examples, let me know and I can drop that commit.

Also there is a commit before that one adding examples using extension classes.

docs/src/main/paradox/common/http-model.md
pfcoperez: Done:

```
To avoid accidentally leaking such information, these fields are omitted from @apidoc[HttpRequest] `toString` output. 
```

pfcoperez:
@raboof Thanks for reviewing this :bowing_man: Do you think it might hit next Akka-HTTP version?

raboof:
@pfcoperez happy to, thanks for the contribution! the PR is waiting for its second reviewer, perhaps @jrudolph or @jlprat?

pfcoperez:
Nice! Thank you @raboof

docs/src/main/paradox/common/http-model.md
```diff
@@ -108,6 +121,19 @@ In addition to the simple @scala[@apidoc[HttpEntity] constructors]@java[`HttpEnt
...
+There are certain environments where it is easy to inadvertently print, write or log entries built out of string representations of @apidoc[HttpResponse] instances. On the other hand, it is not uncommon for HTTP headers and entities to contain _Personal Identifying Information (PII)_ or _Sensitive Personal Information (SPI)_ . 
+
+To avoid accidentally leaking such information, these fields are omitted from @apidoc[HttpResponse] `toString` output. 
+
+If needed, it is possible to alter this behavior using type classes as shown in the following examples: 
+
+Scala
+:   @@snip [HttpResponseShow.scala]($test$/scala/docs/http/scaladsl/HttpResponseShow.scala) 
+
+@@@
```
johanandren: Missing corresponding sample how to handle it for Java. I think for Java it might be less probable that type class is a known concept, so perhaps a a separate phrasing there as well.

Actually, to be honest, I think just having a method (like `def toDetailedString(res): String`) would suffice for Scala as well, people who wants to use type classes will know how to do that anyway.

pfcoperez: Thanks @johanandren ! Are you suggesting adding this behaviour:

```scala
s"""HttpRequest(${_1},${_2},${_3},${_4},${_5})"""
```

as a separated `def toDetailedString(res): String` method, living along with `toString` ?

pfcoperez: That's perfectly fine for me, I can add it and update the docs if that approach is OK.

pfcoperez: Or are you suggesting that the doc should show just an example of such a method taking a response or a request as its input?

johanandren: Just a regular method or function going from HTTPRequest to the desired string format rather than a typeclass and implicit decoration. Would also make the Java and Scala samples pretty much identical.

pfcoperez: Done, please check 02212f21b8df759d465f94235f15f5b6d07c8ef7 .

johanandren: Great, thanks!

pfcoperez:
Hi @jrudolph , thanks for taking a look at this!

Following the minimal approach you suggest, and regarding _filter out a small set of headers known to carry sensitive information_. I worry about blacklisting as it is easy to all off the list, for example, implementing `CustomHeader`.

In this context, what do you think of my original proposal of making `HttpMessage`'s `toString` customizable https://github.com/akka/akka-http/pull/2412 ?

If a small set of headers known to carry sensitive information is still your solution of choice, I suggest compiling a list of Akka-HTTP defined headers classes to whitelist and blacklist anything else falling out that list, wdyt?

johanandren:
Just a heads up so you know, Johannes is on vacation and then team meetings next week so a response may be a while.

pfcoperez:
@johanandren Thank you for the heads-up! That was very considerated from your part :bowing_man: 
I missed his initial comment for the same reason (being on vacation) :smile:

pfcoperez:
@jrudolph I've open a separate PR for hiding bodies of `HttpEntity.Strict` instances https://github.com/akka/akka-http/pull/2737

This way I can center this one into working of secrets redaction of headers.

pfcoperez:
@jrudolph 

- Removing body total o partial string representation is addressed in https://github.com/akka/akka-http/pull/2737
- On the other hand, I've just committed a change to this branch (dfb647488c6c1834738b6ff7750e041461a7b3b6) which adds a `safeToString` which can be invoked by potentially PII/Sensitive information leaking code, such `HttpMessage#toString`.
    - It defaults to not showing more than the header name. No modeled headers don't show their details in this method.
    - Except for modeled headers which are considered safe by default and, therefore, use `toString` implementation.
    - The ten modeled headers that have been identified as containers of sensitive information (https://github.com/akka/akka-http/pull/2560#issuecomment-537103538) override modeled headers default behaviour, showing just their names.

I didn't want to change the renderer nor override `toString` to make these changes orthogonal to the rest of the codebase.

ennru:
Sorry, for leaving this unhandled for so long. I took the liberty to rebase it on current master.

docs/src/test/java/docs/http/javadsl/HttpRequestDetailedStringExampleTest.java
```diff
@@ -0,0 +1,52 @@
...
+                .withEntity(piiBody)
+                .withHeaders(Arrays.asList(piiHeader));
+
+        // Our custom string representation includes body and headers string representations...
+        assertTrue(toDetailedString(httpRequestWithHeadersAndBody).contains(piiHeader.toString()));
+        assertTrue(toDetailedString(httpRequestWithHeadersAndBody).contains(piiBody.toString()));
+
+        // ... while default `toString` doesn't.
+        assertFalse(httpRequestWithHeadersAndBody.toString().contains(piiHeader.toString()));
+        // TODO: Uncomment once https://github.com/akka/akka-http/pull/2737 gets merged
```
ennru, line 47: This is now merged.

docs/src/test/java/docs/http/javadsl/HttpResponseDetailedStringExampleTest.java
```diff
@@ -0,0 +1,54 @@
...
+                .withEntity(piiBody)
+                .withHeaders(Arrays.asList(piiHeader));
+
+        // Our custom string representation includes body and headers string representations...
+        assertTrue(toDetailedString(httpResponseWithHeadersAndBody).contains(piiHeader.toString()));
+        assertTrue(toDetailedString(httpResponseWithHeadersAndBody).contains(piiBody.toString()));
+
+        // ... while default `toString` doesn't.
+        assertFalse(httpResponseWithHeadersAndBody.toString().contains(piiHeader.toString()));
+        // TODO: Uncomment once https://github.com/akka/akka-http/pull/2737 gets merged
```
ennru, line 49: Same here.

docs/src/test/scala/docs/http/scaladsl/HttpRequestDetailedStringExampleSpec.scala
```diff
@@ -0,0 +1,48 @@
...
+      toDetailedString(httpRequestWithHeadersAndBody)
+        .contains(piiHeader.toString)
+    )
+    assert(
+      toDetailedString(httpRequestWithHeadersAndBody).contains(piiBody.toString)
+    )
+
+    // ... while default `toString` doesn't.
+    assert(!s"$httpRequestWithHeadersAndBody".contains(piiHeader.toString))
+    // TODO: Uncomment once https://github.com/akka/akka-http/pull/2737 gets merged
```
ennru, line 44: And here.

docs/src/test/scala/docs/http/scaladsl/HttpResponseDetailedStringExampleSpec.scala
```diff
@@ -0,0 +1,49 @@
...
+        .contains(piiHeader.toString)
+    )
+    assert(
+      toDetailedString(httpResponseWithHeadersAndBody)
+        .contains(piiBody.toString)
+    )
+
+    // ... while default `toString` doesn't.
+    assert(!s"$httpResponseWithHeadersAndBody".contains(piiHeader.toString))
+    // TODO: Uncomment once https://github.com/akka/akka-http/pull/2737 gets merged
```
ennru, line 45: And here.

pfcoperez:
> Sorry, for leaving this unhandled for so long. I took the liberty to rebase it on current master.

@ennru Thank you so much for coming back at this PR :bow: 
These changes are really important in the environment where we are using Akka-HTTP.

> This looks good to me, just enable the negative tests and we should be good to go.

This is awesome! I'll commit the changes you've indicated shortly.

docs/src/test/java/docs/http/javadsl/HttpRequestDetailedStringExampleTest.java
pfcoperez: Fixed in dcecca6595fbc27f9b9a46ae75b27b9b0db15feb

docs/src/test/scala/docs/http/scaladsl/HttpResponseDetailedStringExampleSpec.scala
pfcoperez: Fixed in dcecca6595fbc27f9b9a46ae75b27b9b0db15feb

docs/src/test/scala/docs/http/scaladsl/HttpRequestDetailedStringExampleSpec.scala
pfcoperez: Fixed in dcecca6595fbc27f9b9a46ae75b27b9b0db15feb

docs/src/test/java/docs/http/javadsl/HttpResponseDetailedStringExampleTest.java
pfcoperez: Fixed in dcecca6595fbc27f9b9a46ae75b27b9b0db15feb

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpHeader.scala
```diff
@@ -26,6 +26,7 @@ abstract class HttpHeader extends jm.HttpHeader with ToStringRenderable {
   def lowercaseName: String
   def is(nameInLowerCase: String): Boolean = lowercaseName == nameInLowerCase
   def isNot(nameInLowerCase: String): Boolean = lowercaseName != nameInLowerCase
+  def safeToString: String = name
```
raboof, line 29: I would've been OK with overriding `toString` as well, but this is :+1: to me too - perhaps it is good to be explicit about it.

raboof:
@jrudolph you requested changes before, could you re-review?

ennru:
@pfcoperez Did you notice the suggestions @jrudolph had to improve this PR?

pfcoperez:
Thanks for the ping @ennru and thank you for the upgrade @jrudolph :bow: I've been absent these last two weeks, completely overwhelmed by other issues at work. 

- :100: to `unsafeToString` .
- I will also add more tests, so far I just relied on the documentation ones.
-  _for sensitive headers, could we introduce a marker trai_ Sure :100: 

I will address these points and add the doc by the end of the week and hopefully we can finish this issue up, I am looking forward to having this in Akka-HTTP :pray: 

Thanks for pushing this forward.

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpHeader.scala
```diff
@@ -16,6 +16,19 @@ import akka.util.OptionVal
...
+ * Marker trait for headers which contain portentially secret / sensitive information.
+ *
+ * Mixing this trait will make `toString` to return the name of the header thus avoiding any
+ * detail leak.
+ */
+trait SensitiveHttpHeader {
+  this: HttpHeader =>
+
+  // This header is tagged as potentially containing personal sensitive information
+  override def toString: String = name
```
jrudolph, line 29: That this works is somewhat subtle given that below `def unsafeToString` is implemented in terms of `super.toString`. Should we mark this method as `final` to ensure it is the definitive implementation of `toString` in cases where it is used?

pfcoperez: Indeed, I was about to open a PR to include your suggestion just to find that you already did (https://github.com/jrudolph/akka-http/commit/626485284a0b661f487418dee8284eb2c789ff90). Thank you :bow:

pfcoperez: I've also opened another follow-up PR cleaning the outdated exclusion added here: https://github.com/akka/akka-http/pull/3181

PR #2561, opened 2019-06-06
+ht2 add akka.http.server.http2.log-frames for verbose frame logging for debugging purposes

jrudolph:
+ enable it for some test suites which use WithLogCapturing

### DIFF
--- a/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/FrameLogger.scala
+++ b/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/FrameLogger.scala
@@ -20,6 +20,10 @@ import FrameEvent._
 private[http2] object FrameLogger {
   final val maxBytes = 16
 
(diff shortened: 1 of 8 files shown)


jrudolph:
Missing mima ignore for 2.11, now fixed.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/FrameLogger.scala
```diff
@@ -20,6 +20,10 @@ import FrameEvent._
 private[http2] object FrameLogger {
   final val maxBytes = 16
 
+  def logFramesIfEnabled(shouldLog: Boolean): BidiFlow[FrameEvent, FrameEvent, FrameEvent, FrameEvent, NotUsed] =
+    if (shouldLog) bidi
+    else BidiFlow.identity
```
raboof: Does this 'disappear' on materialization or does it have a run-time cost?

jrudolph: It should, if it doesn't it would be a bug (though, for HTTP/2 I'm not too concerned about runtime costs currently as we haven't really optimized anything so far).

Issue #857, opened 2017-02-10
java cachedHostConnectionPool API could not setting pool by programming


XuefengWu:
In safety internal network environment,  some system using http and have host-level limitation.
There is lack API for setting HTTP host-level connection pool. 

The exists create pool with settings API is only for HTTPS.
[javadsl/Http.scala](https://github.com/akka/akka-http/blob/master/akka-http-core/src/main/scala/akka/http/javadsl/Http.scala)
```
  /**
   * Same as [[cachedHostConnectionPool]] but with HTTPS encryption.
   *
   * The given [[ConnectionContext]] will be used for encryption on the connection.
   */
  def cachedHostConnectionPool[T](
    to:       ConnectHttp,
    settings: ConnectionPoolSettings,
    log:      LoggingAdapter, materializer: Materializer): Flow[Pair[HttpRequest, T], Pair[Try[HttpResponse], T], HostConnectionPool] =
    adaptTupleFlow(delegate.cachedHostConnectionPoolHttps[T](to.host, to.port, to.effectiveHttpsConnectionContext(defaultClientHttpsContext).asScala, settings.asScala, log)(materializer)
      .mapMaterializedValue(_.toJava))
```

t:http:core

jrudolph:
This is the most general API entry-point which should allow both for HTTP and HTTPS connection pools. It seems that just the javadoc is wrong.

jrudolph:
On the other hand there's also this one:

```scala
  def cachedHostConnectionPool[T](to: ConnectHttp, materializer: Materializer): Flow[Pair[HttpRequest, T], Pair[Try[HttpResponse], T], HostConnectionPool] =
    adaptTupleFlow(delegate.cachedHostConnectionPool[T](to.host, to.port)(materializer).mapMaterializedValue(_.toJava))
```

which looks similarly as if an https connection pool could be created with that (using `ConnectHttp`) but which isn't possible. I think these might need to be revised.

laszlovandenhoek:
In the current 10.1.3, using the Java DSL, it is impossible to use custom `ConnectionPoolSettings` with a plain (non-HTTPS) `cachedHostConnectionPool`. The only overload that allows specifying it, is the deprecated 4-arg version, and using it doesn't yield the desired result; since it delegates internally to `cachedHostConnectionPoolHttps`, it forces all connections going through it to be treated as HTTPS, which of course causes requests to fail because of a missing SSL handshake.

[...]

PR #2562, opened 2019-06-06
Fixes a bug that caused the library to use a https connection pool for http connections (#857)

lhotari:
## Purpose

Fixes a bug that caused the library to use a https connection pool for http connections.
* besides the misbehavior, it was not possible to configure a
  plain http connection pool. That is now possible too.
This bug has been reported as #857 a long time ago. The issue report has more details.

## References

References #857


raboof:
OK TO TEST

lhotari:
@raboof Could you please review this change and provide feedback? This should be a quick one...

PR #2566, opened 2019-06-10
remove unused.

takayahilton:


raboof:
OK TO TEST

akka-http/src/main/scala/akka/http/scaladsl/server/Route.scala
```diff
@@ -34,7 +34,6 @@ object Route {
    */
   def seal(route: Route)(implicit
     routingSettings: RoutingSettings = null,
-                         parserSettings:   ParserSettings   = null,
```
raboof: I'm afraid we can't remove this parameter because we want to be binary compatible between releases... I think it can be marked `@deprecated` though.

PR #2569, opened 2019-06-11
Improve validation in uuid unmarshaller

Philippus:
## Purpose

Improve the validation in the uuid unmarshaller.

## References

Follows #2505 which introduced the unmarshaller.

## Changes

Uses a regex that accepts uuid's as defined in [RFC 4122](https://tools.ietf.org/html/rfc4122). A special case is added for the nil uuid. Also avoids a try/catch block.

## Background Context

The validation in `UUID.fromString`-method in java 8 is pretty poor and accepts a lot of invalid uuid strings, java 9 improves on it somewhat but is still far too accepting.

### DIFF
--- a/akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/PredefinedFromStringUnmarshallers.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/PredefinedFromStringUnmarshallers.scala
@@ -45,14 +45,17 @@ trait PredefinedFromStringUnmarshallers {
       }
     }
 
-  implicit val uuidFromStringUnmarshaller: Unmarshaller[String, UUID] =
-    Unmarshaller.strict { string =>
-      try UUID.fromString(string)
-      catch {
-        case e: IllegalArgumentException =>
-          throw new IllegalArgumentException(s"'$string' is not a valid UUID value", e)
-      }
+  implicit val uuidFromStringUnmarshaller: Unmarshaller[String, UUID] = {
+    val validUuidPattern =
+      "[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[1-5][0-9a-fA-F]{3}-[89abAB][0-9a-fA-F]{3}-[0-9a-fA-F]{12}|00000000-0000-0000-0000-000000000000".r.pattern
+
+    Unmarshaller.strict[String, UUID] { string =>
+      if (validUuidPattern.matcher(string).matches)
+        UUID.fromString(string)
+      else
+        throw new IllegalArgumentException(s"'$string' is not a valid UUID value")
     }
+  }
 
   implicit def CsvSeq[T](implicit unmarshaller: Unmarshaller[String, T]): Unmarshaller[String, immutable.Seq[T]] =
     Unmarshaller.strict[String, immutable.Seq[String]] { string =>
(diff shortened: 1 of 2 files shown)


akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/PredefinedFromStringUnmarshallers.scala
```diff
@@ -45,14 +45,17 @@ trait PredefinedFromStringUnmarshallers {
...
-      catch {
-        case e: IllegalArgumentException =>
-          throw new IllegalArgumentException(s"'$string' is not a valid UUID value", e)
-      }
+  implicit val uuidFromStringUnmarshaller: Unmarshaller[String, UUID] = {
+    val validUuidRegex = "[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[1-5][0-9a-fA-F]{3}-[89abAB][0-9a-fA-F]{3}-[0-9a-fA-F]{12}"
+    val nilUuidString = "00000000-0000-0000-0000-000000000000"
+
+    Unmarshaller.strict {
+      case string if string.matches(validUuidRegex) | string == nilUuidString =>
```
raboof:
```suggestion
      case string if string == nilUuidString || string.matches(validUuidRegex) =>
```
Or since we're not actually pattern-matching, perhaps a 'regular' function might be fine here?

Philippus: done

akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/PredefinedFromStringUnmarshallers.scala
```diff
@@ -45,14 +45,17 @@ trait PredefinedFromStringUnmarshallers {
...
-      catch {
-        case e: IllegalArgumentException =>
-          throw new IllegalArgumentException(s"'$string' is not a valid UUID value", e)
-      }
+  implicit val uuidFromStringUnmarshaller: Unmarshaller[String, UUID] = {
+    val nilUuidString = "00000000-0000-0000-0000-000000000000"
+    val validUuidRegex = "[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[1-5][0-9a-fA-F]{3}-[89abAB][0-9a-fA-F]{3}-[0-9a-fA-F]{12}"
+
+    Unmarshaller.strict[String, UUID] { string =>
+      if (string == nilUuidString || string.matches(validUuidRegex))
```
raboof: AFAICS this will compile the regex for each unmarshalling action, which I is expensive. Can we precompile the regex and hold on to the compiled pattern?

Philippus: You mean something like this?:
```
val validUuidPattern =
  "[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[1-5][0-9a-fA-F]{3}-[89abAB][0-9a-fA-F]{3}-[0-9a-fA-F]{12}|00000000-0000-0000-0000-000000000000".r.pattern

  Unmarshaller.strict[String, UUID] { string =>
    if (validUuidPattern.matcher(string).matches)
      UUID.fromString(string)
    else
      throw new IllegalArgumentException(s"'$string' is not a valid UUID value")
  }
```

raboof: Yes that looks like it'd avoid compiling the pattern for every unmarshalling!

PR #2574, opened 2019-06-18
core: allow client setting overrides for target hosts in config

raboof:
Rebased version of #1926

Refs #996

The default connection pool object can lookup host overrides for a
particular host, but client supplied connection pools will not be
overridden

### DIFF
--- a/akka-http-core/src/main/resources/reference.conf
+++ b/akka-http-core/src/main/resources/reference.conf
@@ -490,7 +490,8 @@ akka.http {
     #   }
     #]
     #
-    # When multiple overrides match, the first matching override is selected.
+    # wildcards are allowed in the patterns, but should not overlap, because in that case it is
+    # not defined which set of overrides will take precedence.
     per-host-override = []
 
   }
--- a/akka-http-core/src/main/resources/reference.conf
+++ b/akka-http-core/src/main/resources/reference.conf
@@ -489,6 +489,8 @@ akka.http {
     #   "*.example.com" = { # allow `*` to apply overrides for all subdomains, including `example.com`
     #   }
     #]
+    #
+    # When multiple overrides match, the first matching override is selected.
     per-host-override = []
 
   }
--- a/akka-http-core/src/main/resources/reference.conf
+++ b/akka-http-core/src/main/resources/reference.conf
@@ -478,6 +478,19 @@ akka.http {
     client = {
       # no overrides by default, see `akka.http.client` for default values
     }
+
+    # Add client specific overrides to host-connection-pool. Allows exact and prefixed urls
+    # eg:
+    # per-host-override = [
+    #   "akka.io" = { # can use same things as in global `host-connection-pool` section
+    #     max-connections = 10
+    #   },
+
+    #   "*.example.com" = { # allow `*` to apply overrides for all subdomains, including `example.com`
+    #   }
+    #]
+    per-host-override = []
+
   }
 
   # Modify to tweak default parsing settings.
--- a/akka-http-core/src/main/resources/reference.conf
+++ b/akka-http-core/src/main/resources/reference.conf
@@ -479,23 +479,39 @@ akka.http {
       # no overrides by default, see `akka.http.client` for default values
     }
 
-    # Add client specific overrides to host-connection-pool. Allows exact and prefixed urls
-    # eg:
+    # Allows overriding settings per host. The setting must be a list in which each entry
+    # is an object with a `host-pattern` entry that specifies for which hosts the overrides
+    # should take effect. All other entries have the same syntax as entries in the 
+    # `host-connection-pool` section.
+    #
+    # The `host-pattern` can have these forms:
+    #  * `regex:<pattern>`: the host name is matched against the regular expression pattern
+    #  * `glob:<glob-pattern>` or just `<glob-pattern>`: the host is matched against the given
+    #    pattern. In the pattern the wildcard `*` stands for zero or more characters and `?`
+    #    for any single character
+    #
+    # In both cases, a pattern that matches `*.` at the beginning, i.e. every subdomain,
+    # is expanded to also cover the domain itself (without the leading dot).
+    #
+    # If patterns from multiple entries in the list are matched, only settings from the
+    # first entry found are applied.
+    #
+    # Example:
+    #
     # per-host-override = [
     # {
-    #   host-pattern = "akka.io"
-    #   #can use same things as in global `host-connection-pool` section
+    #   host-pattern = "doc.akka.io"
+    #   # Use the same entries as in the `host-connection-pool` section
     #   max-connections = 10
     # },
     # {
-    #   # allows `*` as a wildcard to apply overrides for all subdomains, including `example.com`
-    #   host-pattern = "*.example.com"
+    #   # `*.akka.io` matches all subdomains like `repo.akka.io` but also `akka.io` itself.
+    #   # `doc.akka.io` is already covered by a previous entry, so these settings here
+    #   # will not apply to `doc.akka.io`.
+    #   host-pattern = "*.akka.io"
     #   max-connections = 11
     # }
     # ]
-    #
-    # wildcards are allowed in the patterns. When wildcards overlap, the first matching set of overrides
-    # is selected.
     per-host-override = []
 
   }
--- a/akka-http-core/src/main/resources/reference.conf
+++ b/akka-http-core/src/main/resources/reference.conf
@@ -482,16 +482,20 @@ akka.http {
     # Add client specific overrides to host-connection-pool. Allows exact and prefixed urls
     # eg:
     # per-host-override = [
+    # {
     #   "akka.io" = { # can use same things as in global `host-connection-pool` section
     #     max-connections = 10
-    #   },
-
-    #   "*.example.com" = { # allow `*` to apply overrides for all subdomains, including `example.com`
     #   }
-    #]
+    # },
+    # {
+    #   {
+    #     "*.example.com" = { # allow `*` to apply overrides for all subdomains, including `example.com`
+    #   }
+    # }
+    # ]
     #
-    # wildcards are allowed in the patterns, but should not overlap, because in that case it is
-    # not defined which set of overrides will take precedence.
+    # wildcards are allowed in the patterns. When wildcards overlap, the first matching set of overrides
+    # is selected.
     per-host-override = []
 
   }
--- a/akka-http-core/src/main/resources/reference.conf
+++ b/akka-http-core/src/main/resources/reference.conf
@@ -483,14 +483,14 @@ akka.http {
     # eg:
     # per-host-override = [
     # {
-    #   "akka.io" = { # can use same things as in global `host-connection-pool` section
-    #     max-connections = 10
-    #   }
+    #   host-pattern = "akka.io"
+    #   #can use same things as in global `host-connection-pool` section
+    #   max-connections = 10
     # },
     # {
-    #   {
-    #     "*.example.com" = { # allow `*` to apply overrides for all subdomains, including `example.com`
-    #   }
+    #   # allows `*` as a wildcard to apply overrides for all subdomains, including `example.com`
+    #   host-pattern = "*.example.com"
+    #   max-connections = 11
     # }
     # ]
     #
--- a/akka-http-core/src/main/scala/akka/http/impl/settings/ConnectionPoolSettingsImpl.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/settings/ConnectionPoolSettingsImpl.scala
@@ -28,8 +28,7 @@ private[akka] final case class ConnectionPoolSettingsImpl(
   idleTimeout:                       Duration,
   connectionSettings:                ClientConnectionSettings,
   responseEntitySubscriptionTimeout: Duration,
-  hostOverrides:                     immutable.Seq[(Regex, ConnectionPoolSettings)],
-  hostOverrideImpl:                  HostOverride)
+  hostOverrides:                     immutable.Seq[(Regex, ConnectionPoolSettings)])
   extends ConnectionPoolSettings {
 
   require(maxConnections > 0, "max-connections must be > 0")
@@ -73,7 +72,6 @@ private[akka] final case class ConnectionPoolSettingsImpl(
     maxConnectionBackoff:              FiniteDuration                                   = maxConnectionBackoff,
     idleTimeout:                       Duration                                         = idleTimeout,
     connectionSettings:                ClientConnectionSettings                         = connectionSettings,
-    poolImplementation:                PoolImplementation                               = poolImplementation,
     responseEntitySubscriptionTimeout: Duration                                         = responseEntitySubscriptionTimeout): ConnectionPoolSettings =
     copy(
       maxConnections,
@@ -86,7 +84,6 @@ private[akka] final case class ConnectionPoolSettingsImpl(
       maxConnectionBackoff,
       idleTimeout,
       connectionSettings,
-      poolImplementation,
       responseEntitySubscriptionTimeout,
       hostOverrides = hostOverrides.map { case (k, v) => k -> mapHostOverrides(v) })
 
@@ -134,9 +131,8 @@ private[akka] object ConnectionPoolSettingsImpl extends SettingsCompanionImpl[Co
       c.getFiniteDuration("max-connection-backoff"),
       c.getPotentiallyInfiniteDuration("idle-timeout"),
       ClientConnectionSettingsImpl.fromSubConfig(root, c.getConfig("client")),
-      c.getPotentiallyInfiniteDuration("response-entity-subscription-timeout"),
-      List.empty,
-      NoOpHostOverride
+      c getPotentiallyInfiniteDuration "response-entity-subscription-timeout",
+      List.empty
     )
   }
 
--- a/akka-http-core/src/main/scala/akka/http/impl/settings/ConnectionPoolSettingsImpl.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/settings/ConnectionPoolSettingsImpl.scala
@@ -44,6 +44,7 @@ private[akka] final case class ConnectionPoolSettingsImpl(
     minConnections == 0 || (baseConnectionBackoff.toMillis > 0 && maxConnectionBackoff.toMillis > 10),
     "If min-connections > 0, you need to set a base-connection-backoff must be > 0 and max-connection-backoff must be > 10 millis " +
       "to avoid client pools excessively trying to open up new connections.")
+  require(hostOverrides.isEmpty || hostOverrides.forall(_._2.hostOverrides.isEmpty), "host-overrides should not be nested")
 
   override def productPrefix = "ConnectionPoolSettings"
 
@@ -93,6 +94,23 @@ private[akka] final case class ConnectionPoolSettingsImpl(
 @InternalApi
 private[akka] object ConnectionPoolSettingsImpl extends SettingsCompanionImpl[ConnectionPoolSettingsImpl]("akka.http.host-connection-pool") {
 
+  def fromSubConfig(root: Config, c: Config): ConnectionPoolSettingsImpl = {
+    new ConnectionPoolSettingsImpl(
+      c.getInt("max-connections"),
+      c.getInt("min-connections"),
+      c.getInt("max-retries"),
+      c.getInt("max-open-requests"),
+      c.getInt("pipelining-limit"),
+      c.getPotentiallyInfiniteDuration("max-connection-lifetime"),
+      c.getFiniteDuration("base-connection-backoff"),
+      c.getFiniteDuration("max-connection-backoff"),
+      c.getPotentiallyInfiniteDuration("idle-timeout"),
+      ClientConnectionSettingsImpl.fromSubConfig(root, c.getConfig("client")),
+      c getPotentiallyInfiniteDuration "response-entity-subscription-timeout",
+      List.empty
+    )
+  }
+
   private[akka] def hostRegex(pattern: String): Regex = {
     val regexPattern = if (pattern.startsWith("regex:")) {
       pattern.stripPrefix("regex:")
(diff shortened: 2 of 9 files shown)


akka-http-core/src/main/scala/akka/http/scaladsl/settings/ConnectionPoolSettings.scala
```diff
@@ -52,38 +85,81 @@ abstract class ConnectionPoolSettings extends js.ConnectionPoolSettings { self:
...
+   */
+  @ApiMayChange
+  def withOverrides(system: ActorSystem): ConnectionPoolSettings = {
+    withOverrides(system.settings.config)
+  }
+
+  /**
+   * Builds a ConnectionPoolSettings that has the host specific overrides populated from config
+   *
+   * This is the ONLY place that a connection pool object can be created that has the abililty to be configured per
```
jrudolph:
```suggestion
   * This is the ONLY place that a connection pool object can be created that has the ability to be configured per
```

akka-http-core/src/main/scala/akka/http/scaladsl/settings/ConnectionPoolSettings.scala
```diff
@@ -52,38 +85,81 @@ abstract class ConnectionPoolSettings extends js.ConnectionPoolSettings { self:
...
-  override def withMaxConnectionLifetime(newValue: Duration): ConnectionPoolSettings = self.copy(maxConnectionLifetime = newValue)
+  @ApiMayChange
+  def withHostOverrides(hostOverrides: immutable.Seq[(String, ConnectionPoolSettings)]): ConnectionPoolSettings = self.copy(hostOverrides = hostOverrides.map { case (h, s) => ConnectionPoolSettingsImpl.hostRegex(h) -> s })
 
-  // overloads for idiomatic Scala use
-  def withConnectionSettings(newValue: ClientConnectionSettings): ConnectionPoolSettings = self.copy(connectionSettings = newValue)
+  @ApiMayChange
+  def appendHostOverride(hostPattern: String, settings: ConnectionPoolSettings): ConnectionPoolSettings = self.copy(hostOverrides = hostOverrides :+ (ConnectionPoolSettingsImpl.hostRegex(hostPattern) -> settings))
+
+  override def withMaxConnections(n: Int): ConnectionPoolSettings = self.copy(maxConnections = n, hostOverrides = hostOverrides.map { case (k, v) => k -> v.withMaxConnections(n) })
```
jrudolph: I think we should create an alternative copy method that provides part of that functionality to keep this code DRYer. E.g.

```scala
def copyAndAdaptOverrides(/* same parameters as copy here but manually provided */, updateOverrides: ConnectionPoolSettings => ConnectionPoolSettings)
```

Then this would be

```scala
self.copyAndAdaptOverrides(maxConnections = n, _.withMaxConnections(n))
```

akka-http-core/src/main/scala/akka/http/scaladsl/settings/ConnectionPoolSettings.scala
```diff
@@ -52,38 +85,81 @@ abstract class ConnectionPoolSettings extends js.ConnectionPoolSettings { self:
...
 }
 
 object ConnectionPoolSettings extends SettingsCompanion[ConnectionPoolSettings] {
+
   override def apply(config: Config) = ConnectionPoolSettingsImpl(config)
+
+  /**
+   * Builds a ConnectionPoolSettings that has the host specific overrides populated from config
+   *
+   * This is the ONLY place that a connection pool object can be created that has the abililty to be configured per
```
jrudolph:
```suggestion
   * This is the ONLY place that a connection pool object can be created that has the ability to be configured per
```

akka-http-core/src/main/scala/akka/http/scaladsl/settings/ConnectionPoolSettings.scala
```diff
@@ -21,6 +24,26 @@ object PoolImplementation {
...
+private[akka] object NoOpHostOverride extends HostOverride {
+  override def apply(host: String, settings: ConnectionPoolSettings): ConnectionPoolSettings = settings
+}
+
+@InternalApi
+private[akka] object DefaultHostOverride extends HostOverride {
+  override def apply(host: String, settings: ConnectionPoolSettings): ConnectionPoolSettings = {
+    if (settings.hostOverrides.isEmpty) settings
+    else {
+      settings.hostOverrides.find(_._1.pattern.matcher(host).matches()).map(_._2).getOrElse(settings)
```
jrudolph: `find` means we take the first override which only makes sense if the ordering is specified.

So, I think we still need to resolve the issue about what should happen if multiple clauses match a target host (the original author tried a solution for "cascading overrides" but found it too complex in the end). As a good-enough first solution we could either ignore overrides if multiple overrides are matching, or make sure that clauses are really in a specified order and then define that the first clause will be taken.

akka-http-core/src/main/scala/akka/http/scaladsl/settings/ConnectionPoolSettings.scala
```diff
@@ -21,6 +24,26 @@ object PoolImplementation {
   case object New extends PoolImplementation
 }
 
+@ApiMayChange
+trait HostOverride {
```
jrudolph: I think by now we have resolved the issue about confusion when loading configuration from a file first and then adopting it with code because we defined and implemented that hostOverrides will get all the updates made through code (because we now do `copy(hostOverrides = hostOverrides.map(...)` everywhere).

I wonder if the additional complexity of this extra indirection here (and also the `withOverrides` API) is still necessary at all, or we can now just live with the full API being available to users both from config files and code at all times?

akka-http-core/src/main/scala/akka/http/scaladsl/settings/ConnectionPoolSettings.scala
raboof: agreed, adding a (private) `copyDeep` makes this a bit nicer

raboof:
failure was #2108

raboof:
PLS BUILD

jrudolph:
> failure was #2108

That has been fixed in master by now. I guess the branch needs a rebase / merge then?

raboof:
PLS BUILD

akka-http-core/src/test/scala/akka/http/scaladsl/GracefulTerminationSpec.scala
```diff
@@ -55,7 +55,7 @@ class GracefulTerminationSpec
       // immediately trying a new connection should cause `Connection refused` since we unbind immediately:
       val r3 = makeRequest(ensureNewConnection = true)
       val ex = intercept[StreamTcpException] {
-        Await.result(r3, 2.seconds)
+        Await.result(r3, 5.seconds)
```
raboof, line 58: Hold on, this really shouldn't be necessary, seems the config is different after all. looking into it.

akka-http-core/src/main/scala/akka/http/scaladsl/settings/ConnectionPoolSettings.scala
```diff
@@ -52,38 +85,81 @@ abstract class ConnectionPoolSettings extends js.ConnectionPoolSettings { self:
...
 
   def withUpdatedConnectionSettings(f: ClientConnectionSettings => ClientConnectionSettings): ConnectionPoolSettings
 }
 
 object ConnectionPoolSettings extends SettingsCompanion[ConnectionPoolSettings] {
+
   override def apply(config: Config) = ConnectionPoolSettingsImpl(config)
+
+  /**
+   * Builds a ConnectionPoolSettings that has the host specific overrides populated from config
```
ennru, line 126:
```suggestion
   * Builds `ConnectionPoolSettings` that have the host-specific overrides populated from config.
```

akka-http-core/src/main/scala/akka/http/scaladsl/settings/ConnectionPoolSettings.scala
```diff
@@ -52,38 +85,81 @@ abstract class ConnectionPoolSettings extends js.ConnectionPoolSettings { self:
...
+   *
+   * This is the ONLY place that a connection pool object can be created that has the ability to be configured per
+   * host, and it's ONLY used by `HttpExt.defaultConnectionPoolSettings`. The intent is NOT to let users define or use
+   * this, or even for akka internally to use this. Think of this more like a placeholder for the
+   * `defaultConnectionPoolSettings` provided in the HttpExt.singleRequest and other client methods instead of breaking
```
ennru, line 131:
```suggestion
   * `defaultConnectionPoolSettings` provided in the `HttpExt.singleRequest` and other client methods instead of breaking
```

akka-http-core/src/main/scala/akka/http/scaladsl/settings/ConnectionPoolSettings.scala
```diff
@@ -52,38 +85,81 @@ abstract class ConnectionPoolSettings extends js.ConnectionPoolSettings { self:
...
+   * binary compatibility by making those calls take an `Option[_]`or using `null`
```
ennru, line 132:
```suggestion
   * binary compatibility by making those calls take an `Option[_]` or using `null`.
```

akka-http-core/src/main/scala/akka/http/scaladsl/settings/ConnectionPoolSettings.scala
```diff
@@ -52,38 +85,81 @@ abstract class ConnectionPoolSettings extends js.ConnectionPoolSettings { self:
...
+   *
+   */
+  @ApiMayChange
+  def withOverrides(system: ActorSystem): ConnectionPoolSettings = {
+    withOverrides(system.settings.config)
+  }
+
+  /**
+   * Builds a ConnectionPoolSettings that has the host specific overrides populated from config
```
ennru, line 141: As above.

akka-http-core/src/main/scala/akka/http/scaladsl/settings/ConnectionPoolSettings.scala
raboof: hostOverrides are a list in the configuration and parsed to a `Seq`, so perhaps it would be sufficient to document that we select the first matching override?

akka-http-core/src/main/scala/akka/http/scaladsl/settings/ConnectionPoolSettings.scala
raboof: Agreed, updated

akka-http-core/src/main/scala/akka/http/impl/settings/ConnectionPoolSettingsImpl.scala
```diff
@@ -26,7 +28,8 @@ private[akka] final case class ConnectionPoolSettingsImpl(
   idleTimeout:                       Duration,
   connectionSettings:                ClientConnectionSettings,
   poolImplementation:                PoolImplementation,
-  responseEntitySubscriptionTimeout: Duration)
+  responseEntitySubscriptionTimeout: Duration,
+  hostOverrides:                     immutable.Seq[(Regex, ConnectionPoolSettings)])
```
jrudolph, line 32: Do we allow nested `hostOverrides`? If no, we should add a require clause here to check that. Something like

```scala
require(hostOverrides.isEmpty || hostOverrides.forall(_._2.isEmpty))
```

akka-http-core/src/main/resources/reference.conf
```diff
@@ -436,6 +436,19 @@ akka.http {
     client = {
       # no overrides by default, see `akka.http.client` for default values
     }
+
+    # Add client specific overrides to host-connection-pool. Allows exact and prefixed urls
```
jrudolph, line 440: Needs an explanation about how overlapping entries are resolved (i.e. currently take first matching).

akka-http-core/src/main/scala/akka/http/javadsl/settings/ConnectionPoolSettings.scala
```diff
@@ -51,23 +56,32 @@ abstract class ConnectionPoolSettings private[akka] () { self: ConnectionPoolSet
...
+    import scala.collection.JavaConverters._
+    self.copy(hostOverrides = hostOverrides.asScala.toList.map { case (h, s) => ConnectionPoolSettingsImpl.hostRegex(h) -> s.asScala })
+  }
+
+  @ApiMayChange
+  def appendHostOverride(hostPattern: String, settings: ConnectionPoolSettings): ConnectionPoolSettings = self.copy(hostOverrides = hostOverrides :+ (ConnectionPoolSettingsImpl.hostRegex(hostPattern) -> settings.asScala))
+
+  def withMaxConnections(n: Int): ConnectionPoolSettings = self.copyDeep(_.withMaxConnections(n), maxConnections = n)
+  def withMinConnections(n: Int): ConnectionPoolSettings = self.copyDeep(_.withMinConnections(n), minConnections = n)
+  def withMaxRetries(n: Int): ConnectionPoolSettings = self.copyDeep(_.withMaxRetries(n), maxRetries = n)
```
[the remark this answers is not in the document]
raboof: Hmm, you'd say so. MiMa doesn't like it, but that looks like it could be a false positive:
```
[error]  * abstract method withMaxRetries(Int)akka.http.javadsl.settings.ConnectionPoolSettings in class akka.http.javadsl.settings.ConnectionPoolSettings does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectAbstractMethodProblem]("akka.http.javadsl.settings.ConnectionPoolSettings.withMaxRetries")
[error]  * in current version there is abstract method withMaxRetries(Int)akka.http.javadsl.settings.ConnectionPoolSettings in class akka.http.javadsl.settings.ConnectionPoolSettings, which does not have a correspondent
[error]    filter with: ProblemFilters.exclude[ReversedAbstractMethodProblem]("akka.http.javadsl.settings.ConnectionPoolSettings.withMaxRetries")
```
Needs a closer look

akka-http-core/src/test/scala/akka/http/scaladsl/settings/ConnectionPoolSettingsSpec.scala
```diff
@@ -39,6 +40,123 @@ class ConnectionPoolSettingsSpec extends AkkaSpec {
...
+          .withFallback(ConfigFactory.defaultReference(getClass.getClassLoader))).forHost("scala-lang.com").maxConnections shouldEqual 36
+
+      // Make sure calls to .apply instead of .forDefault don't get overridden by config
+      ConnectionPoolSettings(
+        ConfigFactory.parseString(settingsString)
+          .withFallback(ConfigFactory.defaultReference(getClass.getClassLoader))).forHost("akka.io").maxConnections shouldEqual 7
+
+    }
+
+    "disregard per host overrides when not created via forDefault method" in {
```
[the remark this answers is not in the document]
raboof: I agree now that we 'cascade' updates in the `withXxx` methods this distinction is no longer needed. Will update.

akka-http-core/src/main/scala/akka/http/scaladsl/settings/ConnectionPoolSettings.scala
raboof: Ah, but unfortunately typesafe-config doesn't keep the order of maps (mainly because of what the semantics should be when merging those, https://github.com/lightbend/config/issues/365). Perhaps we should consider changing to a list after all...

akka-http-core/src/main/scala/akka/http/javadsl/settings/ConnectionPoolSettings.scala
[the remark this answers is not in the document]
raboof: I'd say these can be ignored since this class is not for user extension, so there's no risk of missing implementations.

akka-http-core/src/main/resources/reference.conf
```diff
@@ -460,6 +460,26 @@ akka.http {
...
+    # eg:
+    # per-host-override = [
+    # {
+    #   "akka.io" = { # can use same things as in global `host-connection-pool` section
+    #     max-connections = 10
+    #   }
+    # },
+    # {
+    #   {
```
jrudolph, line 473: Nesting/indentation is wrong (confusing?) here.

akka-http-core/src/main/resources/reference.conf
```diff
@@ -460,6 +460,26 @@ akka.http {
...
+    #     "*.example.com" = { # allow `*` to apply overrides for all subdomains, including `example.com`
```
jrudolph, line 474:
```suggestion
    #     "*.example.com" = { # use `*.` to apply overrides for all subdomains, including `example.com`
```

akka-http-core/src/main/resources/reference.conf
```diff
@@ -460,6 +460,26 @@ akka.http {
...
+    #   }
+    # }
+    # ]
+    #
+    # wildcards are allowed in the patterns. When wildcards overlap, the first matching set of overrides
+    # is selected.
```
jrudolph, line 480:
```suggestion
    # Wildcards are allowed in the patterns. When more than one matching entry for a target
    # host is found, only the first one is used.
```

akka-http-core/src/main/scala/akka/http/impl/settings/ConnectionPoolSettingsImpl.scala
```diff
@@ -80,7 +116,35 @@ private[akka] object ConnectionPoolSettingsImpl extends SettingsCompanionImpl[Co
...
         case "new"    => PoolImplementation.New
       },
-      c.getPotentiallyInfiniteDuration("response-entity-subscription-timeout")
+      c getPotentiallyInfiniteDuration "response-entity-subscription-timeout",
+      List.empty
     )
   }
+
+  private[akka] def hostRegex(pattern: String): Regex = {
+    val regexPattern = if (pattern.startsWith("regex:")) {
```
jrudolph, line 125: This syntax needs to be documented in the config.

akka-http-core/src/main/resources/reference.conf
raboof: I agree it's confusing, but I think it's actually right/required like this, to make sure the list remains ordered...

akka-http-core/src/main/scala/akka/http/scaladsl/settings/ConnectionPoolSettings.scala
```diff
@@ -52,38 +62,55 @@ abstract class ConnectionPoolSettings extends js.ConnectionPoolSettings { self:
...
 object ConnectionPoolSettings extends SettingsCompanion[ConnectionPoolSettings] {
-  override def apply(config: Config) = ConnectionPoolSettingsImpl(config)
-  override def apply(configOverrides: String) = ConnectionPoolSettingsImpl(configOverrides)
+
+  override def apply(config: Config): ConnectionPoolSettingsImpl = {
+    import scala.collection.JavaConverters._
+
+    val hostOverrides = config.getConfigList("akka.http.host-connection-pool.per-host-override").asScala.toList.flatMap { cfg =>
+      val entries = cfg.root.entrySet().asScala
+      require(entries.size < 2, "Please specify each per-host-override in its own object. This is needed to preserve the order.")
```
jrudolph, line 105: How does that work?

akka-http-core/src/main/resources/reference.conf
jrudolph: I think here are three opening braces but only two closing?

akka-http-core/src/main/resources/reference.conf
```diff
@@ -460,6 +460,26 @@ akka.http {
...
+    # {
+    #   "akka.io" = { # can use same things as in global `host-connection-pool` section
+    #     max-connections = 10
+    #   }
+    # },
+    # {
+    #   {
+    #     "*.example.com" = { # allow `*` to apply overrides for all subdomains, including `example.com`
+    #   }
+    # }
```
jrudolph, line 476:
```suggestion
    # {
    #   "*.example.com" = { # use `*.` to apply overrides for all subdomains, including `example.com`
    #   }
    # }
```

akka-http-core/src/main/resources/reference.conf
jrudolph: See https://github.com/akka/akka-http/pull/2574/files#r429813471

akka-http-core/src/main/resources/reference.conf
```diff
@@ -460,6 +460,26 @@ akka.http {
...
     }
+
+    # Add client specific overrides to host-connection-pool. Allows exact and prefixed urls
+    # eg:
+    # per-host-override = [
+    # {
+    #   "akka.io" = { # can use same things as in global `host-connection-pool` section
+    #     max-connections = 10
+    #   }
+    # },
```
jrudolph, line 471: An alternative encoding could be to put the pattern into a field to remove one level of nesting.

```suggestion
    # {
    #   host-pattern = "akka.io"
    #   max-connections = 10
    # },
```

raboof: Yes, I think that's better. Changing.

akka-http-core/src/main/resources/reference.conf
```diff
@@ -478,6 +478,26 @@ akka.http {
     client = {
       # no overrides by default, see `akka.http.client` for default values
     }
+
+    # Add client specific overrides to host-connection-pool. Allows exact and prefixed urls
+    # eg:
+    # per-host-override = [
+    # {
+    #   host-pattern = "akka.io"
```
jrudolph, line 486: :+1:

akka-http-core/src/main/resources/reference.conf
```diff
@@ -478,6 +478,26 @@ akka.http {
...
+    # },
+    # {
+    #   # allows `*` as a wildcard to apply overrides for all subdomains, including `example.com`
+    #   host-pattern = "*.example.com"
+    #   max-connections = 11
+    # }
+    # ]
+    #
+    # wildcards are allowed in the patterns. When wildcards overlap, the first matching set of overrides
+    # is selected.
```
jrudolph, line 498:
```suggestion
    # Allows overriding settings per host. The setting must be a list in which each entry
    # is an object with a `host-pattern` entry that specifies for which hosts the overrides
    # should take effect. All other entries have the same syntax as entries in the 
    # `host-connection-pool` section.
    #
    # The `host-pattern` can have these forms:
    #  * `regex:<pattern>`: the host name is matched against the regular expression pattern
    #  * `glob:<glob-pattern>` or just `<glob-pattern>`: the host is matched against the given
    #    pattern. In the pattern the wildcard `*` stands for zero or more characters and `?`
    #    for any single character
    #
    # In both cases, a pattern that matches `*.` at the beginning, i.e. every subdomain,
    # is expanded to also cover the domain itself (without the leading dot).
    #
    # If patterns from multiple entries in the list are matched, only settings from the
    # first entry found are applied.
    #
    # Example:
    #
    # per-host-override = [
    # {
    #   host-pattern = "doc.akka.io"
    #   # Use the same entries as in the `host-connection-pool` section
    #   max-connections = 10
    # },
    # {
    #   # `*.akka.io` matches all subdomains like `repo.akka.io` but also `akka.io` itself.
    #   # `doc.akka.io` is already covered by a previous entry, so these settings here
    #   # will not apply to `doc.akka.io`.
    #   host-pattern = "*.akka.io"
    #   max-connections = 11
    # }
    # ]
```

jrudolph:
Thanks again, @Falmarri for the original PR almost two years ago and following up on it, @raboof!

Issue #2580, opened 2019-06-27
fileUpload delays responding until the file has been entirely consumed


raboof:
Currently, the `fileUpload` directive doesn't complete before the entire file has been consumed by the inner route.

This means inner routes where the consumption of the file and the production of the response are independently slower than needed. Also it makes it impossible to write a body where the consumption of the file depends on the consumption of the response (like echo-ing the file back to the user).

We can adapt this to have the directive complete as soon as the inner route completes, regardless of whether the incoming file has been completely consumed, which will make the former faster and the latter possible.

PR #2582, opened 2019-06-28
Support 'streaming back' an uploaded file

raboof:
Refs #2580

### DIFF
--- a/akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileUploadDirectives.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileUploadDirectives.scala
@@ -125,6 +125,9 @@ trait FileUploadDirectives {
         import ctx.materializer
         import ctx.executionContext
 
+        // We complete the directive through this promise as soon as we encounter the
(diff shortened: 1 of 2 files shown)


akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileUploadDirectives.scala
```diff
@@ -122,23 +125,37 @@ trait FileUploadDirectives {
         import ctx.materializer
         import ctx.executionContext
 
+        val done = Promise[RouteResult]()
+
         // Streamed multipart data must be processed in a certain way, that is, before you can expect the next part you
         // must have fully read the entity of the current part.
         // That means, we cannot just do `formData.parts.runWith(Sink.seq)` and then look for the part we are interested in
         // but instead, we must actively process all the parts, regardless of whether we are interested in the data or not.
```
jrudolph: Maybe you can add a new comment here that the promise is needed to continue response processing even when parts of the request are still coming in.

jrudolph:
Thanks, @raboof!

Issue #2551, opened 2019-05-31
Long lived clients from grpc eventual oom an akka grpc/http2 server


Marcus-Rosti:
Please refer to https://github.com/akka/akka-grpc/issues/591

In brief, grpc encourages creating a long lived client to complete all http2 requests. However, these long lived clients don't allow GC to happen on the OutStreams from https://github.com/akka/akka-http/blob/master/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Multiplexer.scala#158

Since the client is never closed, it keeps all of the objects active, and thus we eventually run out of heap space.

jrudolph:
Thanks for the report, @Marcus-Rosti. It would be good if you could provide some reproducer, so we could debug the issue. Otherwise, DEBUG log output for the HTTP/2 parts would be helpful as well.

It seems, in the file you linked, `streamFor` can create (or recreate) OutStreams whenever requested, even if the stream has been closed before. It would be good to find out which frame would have recreated a stream that was closed before.

Some background information:

There's some inherent difficulty in implementing HTTP/2 that is to keep track of the state of streams, especially when they are inactive, i.e. still idle or already closed. In both cases, you cannot just throw away all state because you might want to keep prioritization information even for closed streams (currently, we don't implement much of prioritization, so we should not keep prioritization information for closed streams).

One important thing, we don't seem to implement yet, is to keep a running counter of what the highest stream id is that we have seen yet. Because of https://tools.ietf.org/html/rfc7540#section-5.1.1:

```
   The first use of a new stream identifier implicitly closes all
   streams in the "idle" state that might have been initiated by that
   peer with a lower-valued stream identifier.  For example, if a client
   sends a HEADERS frame on stream 7 without ever sending a frame on
   stream 5, then stream 5 transitions to the "closed" state when the
   first frame for stream 7 is sent or received.
```

we can then maintain that any stream that's not in the `outStreams` map is either idle or closed and distinguish those by comparing to the highest established stream id. We can then make sure not to put those streams back in the map if they have already been closed.

Marcus-Rosti:
@jrudolph I have a repro case here and a screenshot of the effect. https://github.com/Marcus-Rosti/akka-grpc-sample-kubernetes-scala/pull/1

Marcus-Rosti:
also that's very informative! Thank you.

[...]

PR #2585, opened 2019-07-01
Close outgoing HTTP/2 streams earlier

raboof:
Before they would be closed when the connection closes, which caused
unnecessary memory use for long-lived HTTP/2 connections with many streams
(as is common with gRPC).

Refs #2551

### DIFF
--- a/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Multiplexer.scala
+++ b/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Multiplexer.scala
@@ -42,7 +42,7 @@ private[http2] trait Http2Multiplexer {
  */
 @InternalApi
 private[http2] trait Http2MultiplexerSupport { logic: GraphStageLogic with StageLogging =>
-  def createMultiplexer(outlet: GenericOutlet[FrameEvent], prioritizer: StreamPrioritizer): Http2Multiplexer =
+  def createMultiplexer(outlet: GenericOutlet[FrameEvent], prioritizer: StreamPrioritizer, http2StreamHandling: Http2StreamHandling): Http2Multiplexer =
     new Http2Multiplexer with OutHandler with StateTimingSupport with LogSupport { self =>
       outlet.setHandler(this)
 
@@ -112,6 +112,7 @@ private[http2] trait Http2MultiplexerSupport { logic: GraphStageLogic with Stage
           trailer = None
           maybeInlet.foreach(_.cancel())
           self.closeStream(this)
+          http2StreamHandling.handleOutgoingEnded(streamId)
 
           if (maybeInlet.isDefined) {
             maybeInlet = None
--- a/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Multiplexer.scala
+++ b/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Multiplexer.scala
@@ -38,11 +38,19 @@ private[http2] trait Http2Multiplexer {
 /**
  * INTERNAL API
  *
- * The current default multiplexer.
+ * Multiplexes the outgoing side of the streams on a HTTP/2 connection.
+ * Accepts the streams from the Akka HTTP side and turns them into `FrameEvent`s
+ * to be passed to the network side.
+ *
+ * Mixed into the Http2ServerDemux graph logic.
  */
 @InternalApi
 private[http2] trait Http2MultiplexerSupport { logic: GraphStageLogic with StageLogging =>
-  def createMultiplexer(outlet: GenericOutlet[FrameEvent], prioritizer: StreamPrioritizer, http2StreamHandling: Http2StreamHandling): Http2Multiplexer =
+  // Signal an outgoing stream has ended, so when
+  // the incoming side is also finished it can be cleaned up.
+  def handleOutgoingEnded(streamId: Int): Unit
+
+  def createMultiplexer(outlet: GenericOutlet[FrameEvent], prioritizer: StreamPrioritizer): Http2Multiplexer =
     new Http2Multiplexer with OutHandler with StateTimingSupport with LogSupport { self =>
       outlet.setHandler(this)
 
@@ -112,7 +120,7 @@ private[http2] trait Http2MultiplexerSupport { logic: GraphStageLogic with Stage
           trailer = None
           maybeInlet.foreach(_.cancel())
           self.closeStream(this)
-          http2StreamHandling.handleOutgoingEnded(streamId)
+          handleOutgoingEnded(streamId)
 
           if (maybeInlet.isDefined) {
             maybeInlet = None
--- a/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
+++ b/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
@@ -52,15 +52,17 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with StageLoggi
         }
     }
   def handleStreamEvent(e: StreamFrameEvent): Unit = {
-    updateState(e.streamId, streamFor(e.streamId).handle(e))
+    updateState(e.streamId, _.handle(e))
   }
   // Called by the outgoing stream multiplexer when that side of the stream is ended.
   def handleOutgoingEnded(streamId: Int): Unit = {
-    updateState(streamId, streamFor(streamId).handleOutgoingEnded())
+    updateState(streamId, _.handleOutgoingEnded())
   }
-  def updateState(streamId: Int, newState: IncomingStreamState): Unit = {
-    if (newState == Closed) incomingStreams -= streamId
-    else incomingStreams += streamId -> newState
+  private def updateState(streamId: Int, handle: IncomingStreamState => IncomingStreamState): Unit = {
+    handle(streamFor(streamId)) match {
+      case Closed   => incomingStreams -= streamId
+      case newState => incomingStreams += streamId -> newState
+    }
   }
   def resetStream(streamId: Int, errorCode: ErrorCode): Unit = {
     incomingStreams -= streamId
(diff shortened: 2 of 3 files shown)


akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Multiplexer.scala
```diff
@@ -42,7 +42,7 @@ private[http2] trait Http2Multiplexer {
  */
 @InternalApi
 private[http2] trait Http2MultiplexerSupport { logic: GraphStageLogic with StageLogging =>
-  def createMultiplexer(outlet: GenericOutlet[FrameEvent], prioritizer: StreamPrioritizer): Http2Multiplexer =
+  def createMultiplexer(outlet: GenericOutlet[FrameEvent], prioritizer: StreamPrioritizer, http2StreamHandling: Http2StreamHandling): Http2Multiplexer =
```
raboof: (Information of closing 'outgoing' streams must somehow be passed from the multiplexer (which deals with 'outgoing' streams) to the stream handling (which deals with 'incoming' streams), but passing in `Http2StreamHandling` in wholesale is not very elegant)

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -40,38 +39,53 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with StageLoggi
...
+    def handleOutgoingEnded(): IncomingStreamState
 
     def stateName: String = productPrefix
     def receivedUnexpectedFrame(e: StreamFrameEvent): IncomingStreamState = {
       pushGOAWAY(ErrorCode.PROTOCOL_ERROR, s"Received unexpected frame of type ${e.frameTypeName} for stream ${e.streamId} in state $stateName")
       Closed
     }
   }
-  case object Idle extends IncomingStreamState {
+  case class Idle(outgoingStreamEnded: Boolean) extends IncomingStreamState {
```
raboof: (perhaps we should make this 2 objects to reduce garbage - might not be worth it though?)

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -121,17 +142,26 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with StageLoggi
...
   // on the incoming side there's (almost) no difference between Open and HalfClosedLocal
-  case class Open(buffer: IncomingStreamBuffer) extends ReceivingData(HalfClosedRemote)
-  // currently unused: we never close a stream before the peer does
-  // case class HalfClosedLocal(buffer: Buffer) extends ReceivingData(Closed)
+  case class Open(buffer: IncomingStreamBuffer) extends ReceivingData(HalfClosedRemote) {
+    override def handleOutgoingEnded(): IncomingStreamState = HalfClosedLocal(buffer)
+  }
+  case class HalfClosedLocal(buffer: IncomingStreamBuffer) extends ReceivingData(Closed) {
+    override def handleOutgoingEnded(): IncomingStreamState = {
+      log.error("handleOutgoingEnded called twice")
```
raboof: Are such errors serious enough to warrant resetting the stream?

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
jrudolph: These states are meant to correspond to the states in the [state diagram](https://httpwg.org/specs/rfc7540.html#StreamStates) of the RFC (we should document that).

Wrt that it's not obvious to me what the new state is and how we would end up there? I'd say we should treat it as an error.

jrudolph: I guess the HTTP/2 terminology might be a bit misleading in that regard. "Idle" means "so far unused" instead of "used before but currently unused".

Another confusing thing might be, that - so far - the logic is only for the server side. On the server side, the only way (leaving `PUSH_PROMISE` aside for now) to open a stream is receiving a `HEADERS` frames, so there has been no traffic in the outgoing direction at all so far.

On the client side there will be another edge in the state graph. In that case, we will move from `Idle` to `Open` when we, the client, send out our original `HEADERS` frame. Then we might (and will in a standard request/response case) indeed see an incoming `HEADERS` frame (= start of response) after the outgoing stream (= request stream) has been closed.  So, we can still keep the simple Idle state but will have to introduce another handler for moving to `Open` when we create a new stream actively.

raboof: I agree on closer examination that this was overly defensive.

In the state diagram, there is no way we could send `EndStream` in the `Idle` state: after all, we haven't received headers so the outgoing stream can't have started. Since in our implementation we have separated the incoming and outgoing streams, it wasn't entirely obvious to me there couldn't be a race condition where we'd see the `sending EndStream` before we saw the `incoming Headers`. Looking at it a bit better, though, since we do `dispatchSubstream` inside `Idle` message handling, this is actually safe and we don't need the separate flag to remember we've sent the `EndStream` even though we were still in the `Idle` state.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
jrudolph: Hard to say because a bug like this might have unknown consequences. We should in any case log that that's a bug and should be filed.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Multiplexer.scala
jrudolph: It's a mixture of cake pattern and composition, so we could also add abstract internal SPI methods to `Http2MultiplexerSupport` and implement them in `Http2StreamHandling`.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -45,18 +52,27 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with StageLoggi
...
-    if (newState == Closed) incomingStreams -= e.streamId
-    else incomingStreams += e.streamId -> newState
+    updateState(e.streamId, streamFor(e.streamId).handle(e))
+  }
+  // Called by the outgoing stream multiplexer when that side of the stream is ended.
+  def handleOutgoingEnded(streamId: Int): Unit = {
+    updateState(streamId, streamFor(streamId).handleOutgoingEnded())
+  }
+  def updateState(streamId: Int, newState: IncomingStreamState): Unit = {
+    if (newState == Closed) incomingStreams -= streamId
```
Marcus-Rosti: This like looks like the overall fix. I can confirm from my tests that it's able to GC all of the Outstreams and incoming calls.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -121,17 +144,26 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with StageLoggi
...
+
   // on the incoming side there's (almost) no difference between Open and HalfClosedLocal
-  case class Open(buffer: IncomingStreamBuffer) extends ReceivingData(HalfClosedRemote)
-  // currently unused: we never close a stream before the peer does
-  // case class HalfClosedLocal(buffer: Buffer) extends ReceivingData(Closed)
+  case class Open(buffer: IncomingStreamBuffer) extends ReceivingData(HalfClosedRemote) {
+    override def handleOutgoingEnded(): IncomingStreamState = HalfClosedLocal(buffer)
+  }
+  case class HalfClosedLocal(buffer: IncomingStreamBuffer) extends ReceivingData(Closed) {
+    override def handleOutgoingEnded(): IncomingStreamState = {
```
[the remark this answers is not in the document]
jrudolph: Maybe it could be the default implementation in `IncomingStreamState` with a message "handleOutgoingEnded received unexpectedly in state $productPrefix. This indicates a bug in Akka HTTP, please report it to the issue tracker."?

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -17,7 +16,15 @@ import scala.collection.immutable
...
 
-/** INTERNAL API */
+/**
+ * INTERNAL API
+ *
+ * Handles the 'incoming' side of HTTP/2 streams.
+ * Accepts `FrameEvent`s from the network side and emits `ByteHttp2SubStream`s for streams
+ * to be handled by the Akka HTTP layer.
+ *
+ * Mixed into the Http2ServerDemux graph logic.
```
jrudolph: :+1: thanks

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -45,18 +52,27 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with StageLoggi
...
-    val newState = streamFor(e.streamId).handle(e)
-    if (newState == Closed) incomingStreams -= e.streamId
-    else incomingStreams += e.streamId -> newState
+    updateState(e.streamId, streamFor(e.streamId).handle(e))
+  }
+  // Called by the outgoing stream multiplexer when that side of the stream is ended.
+  def handleOutgoingEnded(streamId: Int): Unit = {
+    updateState(streamId, streamFor(streamId).handleOutgoingEnded())
+  }
+  def updateState(streamId: Int, newState: IncomingStreamState): Unit = {
```
jrudolph: You could also make that a 

```
def updateState(streamId: Int, handle: IncomingStreamState => IncomingStreamState): Unit
```

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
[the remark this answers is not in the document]
raboof: It's a trade-off: while this is more verbose, it also forces us to consider whether the signal that the outgoing side was closed should have an effect for each state separately. Given the state diagram should be mostly static I agree we can use a default implementation, though - will update.

jrudolph:
> Tested 'manually', didn't find a nice way to capture this in a unit test :/

Created https://github.com/akka/akka-http/issues/2586 to provide a way of looking at current stream statistics or states to use in tests.

jrudolph:
Thanks, @raboof!

PR #2591, opened 2019-07-04
Add script to create release train issue

raboof:


raboof:
I used it to create #2590 , but then made some fixes after that run.

PR #2593, opened 2019-07-05
 Fix too wider IllegalArgumentException catch

bigwheel:
* [x] Have you updated the documentation?
    * no need to fix doc
* [x] Have you added tests for any changed functionality?
-->
## Purpose

Fix inconsistent behavior.

## References

Nothing. This PR includes issue report, proof test and fix patch all.

## Changes

If there are both `Case Class Extraction` and `throw new IllegalArgumentException` in single routing, response becomes ~500 Internal Server Error from 400 Bad Request~ 400 Bad Request from 500 Internal Server Error.

## Background Context

When throw IllegalArgumentException in routing, normally it results 500 internal server error.
However, if there is [Case Class Extraction](https://doc.akka.io/docs/akka-http/current/routing-dsl/case-class-extraction.html#case-class-extraction) it results 400 bad request because of [too wider catch](https://github.com/akka/akka-http/blob/38b49a87/akka-http/src/main/scala/akka/http/scaladsl/server/Directive.scala#L48-L51).


raboof:
OK TO TEST

akka-http/src/main/scala/akka/http/scaladsl/server/Directive.scala
```diff
@@ -45,9 +46,10 @@ abstract class Directive[L](implicit val ev: Tuple[L]) {
     def validatedMap[R](f: L => R)(implicit tupler: Tupler[R]): Directive[tupler.Out] =
       Directive[tupler.Out] { inner =>
         tapply { values => ctx =>
-          try inner(tupler(f(values)))(ctx)
-          catch {
-            case e: IllegalArgumentException => ctx.reject(ValidationRejection(e.getMessage.nullAsEmpty, Some(e)))
+          Try(f(values)) match {
```
raboof: perhaps try/catch is a little more efficient since it saves some allocations?

bigwheel: I tried first try/catch way. But I couldn't work out code with try/catch because try/catch have not "not throwed case" (like ruby's else of "begin - rescue - else").

If you have good idea, I'm very happy :)

raboof: Hmm, you can break this out to a separate function and use 'return' in the 'catch' block... but I agree that's a bit weird perhaps this is OK then :)

bigwheel:
```scala
try {
    val r = f(values)
    inner(tupler(r))(ctx) // ☆
} catch {
    case e: IllegalArgumentException => ctx.reject(ValidationRejection(e.getMessage.nullAsEmpty, Some(e)))
}
```

This code doesn't solve the problem because ☆ code is stiil inside of try/catch.

```scala
val rOrRejected = try {
    f(values)
} catch {
    case e: IllegalArgumentException => ctx.reject(ValidationRejection(e.getMessage.nullAsEmpty, Some(e))) // △
}

inner(tupler(r))(ctx) // XD  r is isntace of R or Future[RouteResult], result of △
```

But if I put ☆ outside of try-catch, I have to handle concrete r case and Future[RouteResult].

bigwheel: > Hmm, you can break this out to a separate function and use 'return' in the 'catch' block... but I agree that's a bit weird perhaps this is OK then :)

Great Idea! Wait a moment ~

lightbend-cla-validator:

lightbend-cla-validator:

jrudolph:
(CLA check failed only because of autoformatter).

jrudolph:
Thanks a lot, @bigwheel.

PR #2596, opened 2019-07-16
Improve JavaUUID path matcher

Philippus:
## Purpose

Improve the JavaUUID path matcher

## Changes

This PR tightens the existing regex and adds the nil uuid, making it accept uuid's as defined in [RFC 4122](https://tools.ietf.org/html/rfc4122). Also avoids a try/catch block.

### DIFF
--- a/akka-http/src/main/scala/akka/http/scaladsl/server/PathMatcher.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/server/PathMatcher.scala
@@ -486,10 +486,8 @@ trait PathMatchers {
    * @group pathmatcher
    */
   val JavaUUID: PathMatcher1[UUID] =
-    PathMatcher("""[\da-fA-F]{8}-[\da-fA-F]{4}-[\da-fA-F]{4}-[\da-fA-F]{4}-[\da-fA-F]{12}""".r) flatMap { string =>
-      try Some(UUID.fromString(string))
-      catch { case _: IllegalArgumentException => None }
-    }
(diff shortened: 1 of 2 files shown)


akka-http/src/main/scala/akka/http/scaladsl/server/PathMatcher.scala
```diff
@@ -486,9 +486,8 @@ trait PathMatchers {
    * @group pathmatcher
    */
   val JavaUUID: PathMatcher1[UUID] =
-    PathMatcher("""[\da-fA-F]{8}-[\da-fA-F]{4}-[\da-fA-F]{4}-[\da-fA-F]{4}-[\da-fA-F]{12}""".r) flatMap { string =>
-      try Some(UUID.fromString(string))
```
jrudolph: You removed the try/catch because you now know that all paths matching will not throw in `UUID.fromString`? It seemed a slight bit safer before because we might not know for which input `UUID.fromString` might throw in all JDKs (and also in the future).

Philippus: Yes, I think we're safe here, we know what a UUID should look like. The `IllegalArgumentException` is thrown when the string is too large or when there are no 4 `-`s in the string.

jrudolph: Seems reasonable, thanks for the explanation.

Could it then be changed to `.map(UUID.fromString)`?

Issue #2606, opened 2019-07-19
Multiple subprotocols are not supported in websockets handshake


RomanBrodetski:
According to the [Websockets RFC](https://tools.ietf.org/html/rfc6455#section-11.3.4), a websocket client can send multiple values in `Sec-WebSocket-Protocol` header. 
However, the [`WebSocketRequest`](https://github.com/akka/akka-http/blob/master/akka-http-core/src/main/scala/akka/http/scaladsl/model/ws/WebSocketRequest.scala#L22) case class has a `subprotocol` field of type `Option[String]`. Passing subprotocols as comma separated values there doesn't work either, as in this case the handshake fails when the server responds with one of the subprotocols (see https://github.com/akka/akka-http/blob/master/akka-http-core/src/main/scala/akka/http/impl/engine/ws/Handshake.scala#L271).

It seems like this feature is only missing on the `WebSocketRequest` level - the `Option[String]` is then converted to `Seq[String]` with `.toList()` (see [here](https://github.com/akka/akka-http/blob/master/akka-http-core/src/main/scala/akka/http/impl/engine/ws/WebSocketClientBlueprint.scala#L57)). Instead, the `WebSocketRequest` should accept the subprotocols as a `Seq[String]`.

jrudolph:
Fixed in #2607.

Issue #2349, opened 2018-12-21
GracefulTerminationSpec: Expected exception akka.stream.StreamTcpException to be thrown, but akka.http.impl.engine.client.pool.SlotState$BusyState$$anon$1 was thrown


raboof:
https://jenkins.akka.io:8498/job/pr-validator-akka-http/3092/testReport/junit/akka.http.scaladsl/GracefulTerminationSpec/Graceful_termination_should_in_flight_request_responses_should_include_Connection__close_and_connection_should_be_closed/
```
org.scalatest.exceptions.TestFailedException: Expected exception akka.stream.StreamTcpException to be thrown, but akka.http.impl.engine.client.pool.SlotState$BusyState$$anon$1 was thrown
	at org.scalatest.MatchersHelper$.newTestFailedException(MatchersHelper.scala:138)
	at org.scalatest.MatchersHelper$.checkExpectedException(MatchersHelper.scala:312)
	at org.scalatest.words.ResultOfTheTypeInvocation.thrownBy(ResultOfTheTypeInvocation.scala:39)
	at akka.http.scaladsl.GracefulTerminationSpec.akka$http$scaladsl$GracefulTerminationSpec$$ensureConnectionIsClosed(GracefulTerminationSpec.scala:193)
	at akka.http.scaladsl.GracefulTerminationSpec$$anon$6.<init>(GracefulTerminationSpec.scala:137)
	at akka.http.scaladsl.GracefulTerminationSpec.$anonfun$new$10(GracefulTerminationSpec.scala:123)
```

raboof:
These tests are testing that after graceful shutdown, requests are not handled.

The test currently expects a `StreamTcpException` from Akka Streams, but when this test failure occurs it gets the `IllegalStateException("Connection was shutdown.") with NoStackTrace` from `SlotState` that is thrown when the connection is completed when in 'busy' state instead. It's not obvious to me that that is wrong, but I'm not 100% sure. The comment for that exception is:

```
      // There's no good reason why the connection stream (i.e. the user-facing client Flow[HttpRequest, HttpResponse])
      // would complete during processing of a request.
      // One reason might be that failures on the TCP layer don't necessarily propagate through the stack as failures
      // because of the notorious cancel/failure propagation which can convert failures into completion.
```

johanandren:
https://jenkins.akka.io:8498/job/pr-validator-akka-http/3208/

jrudolph:
[...]

PR #2607, opened 2019-07-19
Support multiple subprotocols in Websocket handshake for client side

RomanBrodetski:
Currently passing multiple websockets subprotocols in `Sec-WebSocket-Protocol` header is not supported (see https://github.com/akka/akka-http/issues/2606). This PR makes it possible to do so, passing them separated by comma in `subprotocol: Option[String]` field of the [`WebSocketRequest`](https://github.com/akka/akka-http/blob/master/akka-http-core/src/main/scala/akka/http/scaladsl/model/ws/WebSocketRequest.scala#L22) model. This corresponds to the `Sec-WebSocket-Protocol` header definition in [RFC 6455](https://tools.ietf.org/html/rfc6455#section-11.3.4)

A better solution would be to change the `WebSocketRequest` model to accept a `subprotocols: Seq[String]` field, but that would break compatibility with current client code. If it, nevertheless, makes sense to do this change, please let us know - we are happy to submit a corresponding PR.

References https://github.com/akka/akka-http/issues/2606

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/ws/WebSocketClientBlueprint.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/ws/WebSocketClientBlueprint.scala
@@ -27,6 +27,8 @@ import akka.http.impl.engine.ws.Handshake.Client.NegotiatedWebSocketSettings
 import akka.http.impl.util.{ SingletonException, StreamUtils }
 import akka.stream.impl.fusing.GraphStages.SimpleLinearGraphStage
 
+import scala.collection.immutable
+
 /** INTERNAL API */
 @InternalApi
 private[http] object WebSocketClientBlueprint {
@@ -54,7 +56,8 @@ private[http] object WebSocketClientBlueprint {
 
     val valve = StreamUtils.OneTimeValve()
 
-    val (initialRequest, key) = Handshake.Client.buildRequest(uri, extraHeaders, subprotocol.toList, settings.websocketRandomFactory())
+    val subprotocols: immutable.Seq[String] = subprotocol.toList.flatMap(_.split(",")).map(_.trim)
+    val (initialRequest, key) = Handshake.Client.buildRequest(uri, extraHeaders, subprotocols, settings.websocketRandomFactory())
     val hostHeader = Host(uri.authority.normalizedFor(uri.scheme))
     val renderedInitialRequest =
       HttpRequestRendererFactory.renderStrict(RequestRenderingContext(initialRequest, hostHeader), settings, log)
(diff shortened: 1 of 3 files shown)


lightbend-cla-validator:
Hi @RomanBrodetski,

raboof:
OK TO TEST

akka-http-core/src/main/scala/akka/http/impl/engine/ws/WebSocketClientBlueprint.scala
```diff
@@ -54,7 +56,8 @@ private[http] object WebSocketClientBlueprint {
 
     val valve = StreamUtils.OneTimeValve()
 
-    val (initialRequest, key) = Handshake.Client.buildRequest(uri, extraHeaders, subprotocol.toList, settings.websocketRandomFactory())
+    val subprotocols: immutable.Seq[String] = subprotocol.toList.flatMap(_.split(",")).map(_.trim)
```
raboof: From the RFC, "If present, this value indicates one
        or more comma-separated subprotocol the client wishes to speak,
        ordered by preference.  The elements that comprise this value
        MUST be non-empty strings with characters in the range U+0021 to
        U+007E not including separator characters as defined in
        [RFC2616] and MUST all be unique strings" - that suggests comma's are allowed inside the protocol names (U+002C), but how does that work when the values are also comma-separated?

raboof:
(failure was unrelated, #2349)

raboof:
PLS BUILD

akka-http-core/src/main/scala/akka/http/impl/engine/ws/WebSocketClientBlueprint.scala
RomanBrodetski: it says `not including separator characters as defined in
[RFC2616]`, and I think a comma is a part of this definition:

> separators     = "(" | ")" | "<" | ">" | "@"
                      | "," | ";" | ":" | "\" | <">
                      | "/" | "[" | "]" | "?" | "="
                      | "{" | "}" | SP | HT

so, seems like subprotocol cannot include commas

raboof: Oops :D - that's clear then! :+1:

raboof: Should we ignore all separator characters or just trim the whitespace?

RomanBrodetski: not sure. If we want to enforce the RFC, we can validate the `subprotocols` field and throw an exception if it contradicts the specification.
Although, this might break client code. Current  implementation is the closest to current behavior (not enforcing any rules).

raboof: yeah, also the specification itself uses spaces in the examples, so it'd be silly to disallow that. I agree this seems fine.

Issue #2605, opened 2019-07-19
ConnectionContext.https now (10.1.9) is deprecated, but in docs it is still only one described


abdolence:
As you can see at https://doc.akka.io/docs/akka-http/current/server-side/server-https-support.html the only one available official example shows how to use the deprecated version of ConnectionContext.https.

Probably the docs should be updated or was it somehow wrongly deprecated?

raboof:
Hmm, looks like I didn't correctly move the default parameters over to the new recommended method in 455a226051, indeed.

jrudolph:
Fixed in #2610.

PR #2610, opened 2019-07-22
Add default parameters to the right ConnectionContext.https

raboof:
Refs #2605


akka-http-core/src/main/mima-filters/10.1.9.backwards.excludes
```diff
@@ -0,0 +1,4 @@
+# https://github.com/akka/akka-http/issues/2605
```
jrudolph: Let's start creating excludes files per issue to avoid conflicts.

So, it should go into 

```
mima-filters/10.1.9.backwards.excludes/issue-2605-deprecate-right-method.excludes
```

(or similar).


See https://github.com/lightbend/mima/commit/2573ceeb4a222efb8405ab0793b33e3ba930ca60#diff-eec91a6d045cc72eff0f61fed09b4ac3 for more info about that feature.

Issue #2603, opened 2019-07-18
feature request: HttpEntity.withDataBytes


dsilvasc:
The `HttpEntity` trait has a `dataBytes` property -- can it also have a `withDataBytes` method to replace the `Source[ByteString, Any]`?

The stream passed into a response can fail.  When it does, akka-http responds with status 200, sends the header, maybe sends some of the bytes from the stream, then closes the connection without any explanation.  The client just gets a connection reset, and the server logs don't say anything.  This kind of connection reset is difficult to debug.

I'd like to wrap the stream in my server's responses with some logging:

```scala
response.mapEntity(e => e.withDataBytes(e.dataBytes.recover {
  case NonFatal(e) =>
    logger.error("response stream failed:", e)
    throw e
}))
```

dsilvasc:
Related to https://github.com/akka/akka-http/issues/894 and https://github.com/akka/akka-http/issues/848

raboof:
Would [transformDataBytes](https://doc.akka.io/api/akka-http/current/akka/http/scaladsl/model/HttpEntity.html#transformDataBytes(transformer:akka.stream.scaladsl.Flow[akka.util.ByteString,akka.util.ByteString,Any]):akka.http.scaladsl.model.HttpEntity) work for you here?

dsilvasc:
Hmm, yes, this seems to work:

```scala
response.mapEntity(_.transformDataBytes(Flow.fromFunction((x: ByteString) => x).recover {
  case e =>
    logger.error("response stream failed: ", e)
    throw e
}))
```

But it seems odd to recover from errors in the transformation flow instead of errors in the stream.

dsilvasc:
Unfortunate consequence: `ResponseEntity.transformDataBytes` turns a strict or default http entity into a chunked one.

`UniversalEntity` has a transformDataBytes method that takes a content length and doesn't turn strict entities into chunked ones, but `mapEntity` takes a `ResponseEntity => ResponseEntity` instead of `UniversalEntity => UniversalEntity`.

Matching on the different entity cases preserves it though:

```scala
response.entity match {
  case HttpEntity.Strict(_, _) => response
  case HttpEntity.Chunked(contentType, chunks) =>
    response.withEntity(HttpEntity.Chunked(contentType, logStreamFailure(requestContext.requestID, chunks)))
  case HttpEntity.CloseDelimited(contentType, data) =>
[...]

PR #2621, opened 2019-07-30
core: pull transformDataBytes(newContentLength, transformer) up to HttpEntity

jrudolph:
It can work on any kind of entity and will always return a Universal entity.

Refs #2603.


akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
```diff
@@ -334,17 +335,14 @@ object HttpEntity {
     override def isKnownEmpty: Boolean = data.isEmpty
     override def isStrict: Boolean = true
 
-    override def dataBytes: Source[ByteString, NotUsed] = Source(data :: Nil)
+    override def dataBytes: Source[ByteString, NotUsed] = Source.single(data)
```
raboof: :+1:

PR #2624, opened 2019-07-31
Widen JavaUUID regexp's

Philippus:
## Purpose

Widen JavaUUID regex's (again) to allow for old and future variants of uuid's (described in chapter 4.1.1 of [RFC 4122](https://tools.ietf.org/html/rfc4122).

## References

Following-up after #2569 and #2596. Thanks to @plokhotnyuk for pointing this out in https://github.com/spray/spray-json/pull/313.


jrudolph:
Could you add a few tests for those newly allowed UUIDs so that we can see that `UUID.fromString` will also accept them? Otherwise, we might want to add the try/catch again.

Issue #2614, opened 2019-07-24
Plaintext connection to HTTPS endpoint hangs on JDK 11


jrudolph:
On JDK 8, the connection closed, on JDK 11 the request just hangs until the idle-timeout kicks in.

Thanks for the report, @makkarpov.

See https://discuss.lightbend.com/t/akka-http-tls-server-hangs-when-it-receives-plaintext-connection/4636 and the reproducer at https://github.com/makkarpov/akka-tls-hangs-example.

raboof:
On failure, the TLSActor closes the inputs (https://github.com/akka/akka/blob/master/akka-stream/src/main/scala/akka/stream/impl/io/TLSActor.scala#L451). Perhaps a cancellation cause (https://github.com/akka/akka/pull/27266) could make the error handling here easier, though it seems TLSActor is using the reactivestreams API directly so that might not be straightforward...

makkarpov:
By the way, this no longer reproduces for me on 10.1.12 and JDK 11.

jrudolph:
This is the reverse situation as in #2614.

Should be fixed by akka/akka#29121, released in Akka 2.6.6 and by akka/akka#29148 in not-yet-released 2.5.32.

To close this issue, we should add a test, once we update to 2.5.32.

PR #2626, opened 2019-07-31
Plaintext connection to https endpoint hangs on jdk11: reproducer

raboof:
Refs #2614

## References

References #xxxx

### DIFF
--- a/akka-http-core/src/test/scala/akka/http/scaladsl/ClientServerSpec.scala
+++ b/akka-http-core/src/test/scala/akka/http/scaladsl/ClientServerSpec.scala
@@ -1005,12 +1005,11 @@ Host: example.com
       val (hostname, port) = SocketUtil.temporaryServerHostnameAndPort()
       val request = HttpRequest(uri = s"http://$hostname:$port", headers = headers.Connection("close") :: Nil)
 
-      val serverConnectionContext = ExampleHttpContexts.exampleServerContext
-
-      val routes: Flow[HttpRequest, HttpResponse, Any] = Flow[HttpRequest].map { _ => ??? }
       val serverBinding =
         Http()
-          .bindAndHandle(routes, hostname, port, connectionContext = serverConnectionContext)
+          .newServerAt(hostname, port)
+          .enableHttps(ExampleHttpContexts.exampleServerContext)
+          .bind(_ => ???)
           .futureValue
 
       val failure = Http()
@@ -1020,7 +1019,8 @@ Host: example.com
 
       failure.getMessage should include("Connection reset by peer")
 
-      serverBinding.unbind()
+      serverBinding.unbind().futureValue
+      Http().shutdownAllConnectionPools()
     }
 
   }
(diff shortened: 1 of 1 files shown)


jrudolph:
@raboof, anything we could to about that?

raboof:
I don't think I've looked into this before apart from recording the reproducer - I bet we can once we look closer ;)

akka-http-core/src/test/scala/akka/http/scaladsl/ClientServerSpec.scala
```diff
@@ -751,6 +752,30 @@ Host: example.com
 
       Await.result(binding.unbind(), 10.seconds)
     }
+
+    "produce a useful error message when connecting to a HTTPS endpoint over HTTP" in Utils.assertAllStagesStopped {
+      // akka/akka-http#1219
```
ennru, line 757: Isn't this about https://github.com/akka/akka-http/issues/2614 ?

raboof:
> Why doesn't PR validation fail on this test? Aren't the tests run on JDK 11?

I think PR validation is on jdk8 (and we have jdk11 nightlies)

akka-http-core/src/test/scala/akka/http/scaladsl/ClientServerSpec.scala
raboof: Not sure what I was thinking
```suggestion
      // akka/akka-http#2614
```

jrudolph:
Tried that with the fixes in https://github.com/akka/akka/pull/29121. It seems while it just hangs in the current version, with the fix, it will close the connection properly. The request will then fail with "The http server closed the connection unexpectedly before delivering responses for 1 outstanding requests" which is already better than hanging.

It seems for these kinds of errors TLSActor currently calls `fail(ex, closeTransport = false)` which means only cancellation is sent but no error. Not sure why that's the case.

jrudolph:
> It seems for these kinds of errors TLSActor currently calls `fail(ex, closeTransport = false)` which means only cancellation is sent but no error. Not sure why that's the case.

Removing that makes `TlsSpec` "emit an error if the TLS handshake fails certificate checks" because it seems server and client might fail at the same time and there might be a race about which error is received first?

raboof:
This seems to succeed now

jrudolph:
Did you test on JDK 11? Would we expect it to succeed with the fix not yet released on 2.5.x?

raboof:
Oops, indeed I was still on JDK8 here :confused: sorry about the noise.

jrudolph:
Can now be restested with an update to 2.5.32.

ignasi35:
Cant' test this locally due to #3556

jrudolph:
Rebased on master where on JDK 11 it doesn't fail anymore with a timeout but now with the wrong exception because it seems the connection reset is received as a regular connection closure for some reason.

jrudolph:
> Rebased on master where on JDK 11 it doesn't fail anymore with a timeout but now with the wrong exception because it seems the connection reset is received as a regular connection closure for some reason.

Seems to happen on both 2.5.32 and 2.6.10. Verified with wireshark that the server closes the connection regularly.

The culprit seems to still be the TLSActor which sends failures user-side but maybe not towards the network?

raboof:
reported on the Akka side as https://github.com/akka/akka/issues/29957

Issue #894, opened 2017-02-23
Deal more gracefully with response entity stream failures


jrudolph:
It seems we are currently pretty silent about response entity stream errors: http://stackoverflow.com/questions/42283433/how-to-get-error-from-akka-stream-propagate-to-akka-http-to-both-be-logged-and-n

It seems the response renderer will just fail if a response entity failure occurs and this failure will trickle down the pipeline.

We cannot give a better error to the client but we should log it in a consistent way.

jrudolph:
Related to #848

jrudolph:
Finally fixed in #2627.

Issue #848, opened 2017-02-06
Failure to materialize response entity from server goes up in smoke


johanandren:
This sample: https://github.com/johanandren/akka-stream-samples/blob/8ff72a06a11a78a66fa7d58d7f42328f8b277193/src/main/java/com/lightbend/akka/johan/stream/samples/java/Sample3.java

Throws an exception when materializing (because the range is converted to a Seq and they cannot contain that many elements), the exception is nowhere to be seen, client get connection error:
```
$ curl http://127.0.0.1:8080/numbers
curl: (56) Recv failure: Connection reset by peer
```

Would be nice if it resulted in Http 500, or at least the exception got logged on the server.

jrudolph:
I can reproduce this issue. It is logged on DEBUG level as 

```
[info] [DEBUG] [07/31/2019 10:50:44.766] [default-akka.actor.default-dispatcher-2] [akka://default/system/StreamSupervisor-0/flow-1-0-detacher] Aborting tcp connection to /127.0.0.1:56120 because of upstream failure: java.lang.IllegalArgumentException: 0 to 2147483647 by 1: seqs cannot contain more than Int.MaxValue elements.
[info] scala.collection.immutable.Range.fail(Range.scala:143)
[info] scala.collection.immutable.Range.length(Range.scala:140)
[info] scala.collection.IndexedSeqLike$class.iterator(IndexedSeqLike.scala:90)
[info] scala.collection.immutable.Range.iterator(Range.scala:61)
[info] akka.stream.javadsl.Source$$anon$2.create(Source.scala:173)
[info] akka.stream.javadsl.Source$$anon$2.create(Source.scala:171)
[info] akka.stream.javadsl.Source$$anonfun$fromIterator$1.apply(Source.scala:103)
[info] akka.stream.javadsl.Source$$anonfun$fromIterator$1.apply(Source.scala:103)
[info] akka.stream.scaladsl.Source$$anon$1.iterator(Source.scala:261)
[info] akka.stream.impl.fusing.StatefulMapConcat$$anon$30.onPush(Ops.scala:2035)
[info] akka.stream.impl.fusing.GraphInterpreter.processPush(GraphInterpreter.scala:523)
[info] akka.stream.impl.fusing.GraphInterpreter.processEvent(GraphInterpreter.scala:510)
[info] akka.stream.impl.fusing.GraphInterpreter.execute(GraphInterpreter.scala:376)
```

[...]

PR #2627, opened 2019-07-31
Better error handling on server when response entity stream fails

jrudolph:
I.e.

 * when materializing the response entity stream fails: return a 500 response
 * when streaming itself fails: abort the connection 

Refs #848, #894.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/rendering/HttpResponseRendererFactory.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/rendering/HttpResponseRendererFactory.scala
@@ -23,6 +23,8 @@ import akka.annotation.InternalApi
 import ResponseRenderingContext.CloseRequested
 import headers._
 
+import scala.util.control.NonFatal
+
 /**
  * INTERNAL API
  */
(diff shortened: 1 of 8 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/rendering/HttpResponseRendererFactory.scala
```diff
@@ -79,29 +81,39 @@ private[http] class HttpResponseRendererFactory(
...
                 push(out, outElement)
                 if (close) completeStage()
-              case Streamed(outStream) => transfer(outStream)
+              case Streamed(outStream) =>
+                try transfer(outStream)
+                catch {
+                  case NonFatal(e) =>
+                    transferring = false
+                    log.error(e, s"Rendering of response failed because response entity stream materialization failed with '${e.getMessage}'. Sending out 500 response instead.")
+                    push(out, render(ResponseRenderingContext(HttpResponse(500, entity = "OOPS!"))).asInstanceOf[Strict].bytes)
```
raboof: :smile:

jrudolph: Let's use the default message there as well ;)

Updated.

PR #2628, opened 2019-07-31
Fix a few warnings

jrudolph:
The change to use `attributes.get` instead of `getFirst` is probably not strictly backwards compatible if you set multiple `SizeLimit` attributes before, on the other hand the previous behavior probably wasn't documented like that anyway.


akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
```diff
@@ -623,7 +623,7 @@ object HttpEntity {
       private var bytesLeft = Long.MaxValue
 
       override def preStart(): Unit = {
-        attributes.getFirst[SizeLimit] match {
+        attributes.get[SizeLimit] match {
```
raboof: Hmm, some test failures around this features - is `get` really a good replacement for `getFirst`?

jrudolph: Maybe not? Just followed the deprecation instructions to see where that leads us. Removing that commit for now...

Issue #2617, opened 2019-07-25
Remove reliance on deprecated akka.stream.actor.ActorPublisher


raboof:
as it is going away entirely in Akka 2.6

It is used in `PoollInterfaceActor`

jrudolph:
We can try first if it is simple and safe to convert to GraphStage directly. Otherwise, we'll have to include a private copy of ActorPublisher for the time being.

jrudolph:
The funny thing about `PoolInterfaceActor` is that it uses its publisher/subscriber-me only to connect itself to the actual pool implementation. I guess it was written when `Source/Sink.queue` wasn't there. So, that might actually be somewhat easy to fix.

jrudolph:
Fixed by reimplementing `PoolInterfaceActor` as a GraphStage in #2630.

PR #2629, opened 2019-07-31
Embed copy of ActorPublisher / ActorSubscriber

jrudolph:
So, we rely less on deprecated APIs of akka-stream that will be removed
in 2.6.

Refs #2617


jrudolph:
Maybe not needed, a replacement seems in reach. Closing for now.

PR #2631, opened 2019-08-01
Force connection closure when pool is stopped

jrudolph:
Otherwise, those connections might hang around for a while when a request is ongoing and the client flow isn't expecting or reacting to completion or cancellation.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/NewHostConnectionPool.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/NewHostConnectionPool.scala
@@ -25,6 +25,7 @@ import akka.util.OptionVal
 
 import scala.concurrent.Future
 import scala.concurrent.duration._
+import scala.util.control.NoStackTrace
 import scala.util.control.NonFatal
 import scala.util.{ Failure, Random, Success, Try }
 
@@ -194,7 +195,7 @@ private[client] object NewHostConnectionPool {
           def isConnected: Boolean = state.isConnected
           def shutdown(): Unit = {
             // TODO: should we offer errors to the connection?
-            closeConnection()
+            closeConnection(force = true)
 
             state.onShutdown(this)
           }
@@ -269,7 +270,7 @@ private[client] object NewHostConnectionPool {
 
                 if (!state.isConnected && connection != null) {
                   debug(s"State change from [${previousState.name}] to [Unconnected]. Closing the existing connection.")
-                  closeConnection()
+                  closeConnection(force = false)
                 }
 
                 if (!previousState.isIdle && state.isIdle && !(state == Unconnected && currentEmbargo != Duration.Zero)) {
@@ -315,7 +316,7 @@ private[client] object NewHostConnectionPool {
 
                   try {
                     cancelCurrentTimeout()
-                    closeConnection()
+                    closeConnection(force = true)
                     state.onShutdown(this)
                     logic.slotsWaitingForDispatch.remove(this)
                     OptionVal.None
@@ -387,9 +388,9 @@ private[client] object NewHostConnectionPool {
             }
           }
 
-          def closeConnection(): Unit =
+          def closeConnection(force: Boolean): Unit =
             if (connection ne null) {
-              connection.close()
+              connection.close(force)
               connection = null
             }
           def isCurrentConnection(conn: SlotConnection): Boolean = connection eq conn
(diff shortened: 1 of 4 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/NewHostConnectionPool.scala
```diff
@@ -560,8 +563,8 @@ private[client] object NewHostConnectionPool {
           super.onDownstreamFinish()
         }
         override def postStop(): Unit = {
-          log.debug("Pool stopped")
           slots.foreach(_.shutdown())
+          log.warning(s"Pool stopped")
```
raboof: Does this have to be a warning? Isn't it 'normal' for a pool to stop?

akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/NewHostConnectionPool.scala
```diff
@@ -439,8 +440,10 @@ private[client] object NewHostConnectionPool {
 
             emitRequest(newRequest)
           }
-          def close(): Unit = {
-            requestOut.complete()
+          def close(force: Boolean): Unit = {
+            if (force) requestOut.fail(new IllegalStateException("pool slot shutdown") with NoStackTrace)
+            else requestOut.complete()
```
raboof: :+1: seems premature to just 'complete' it if the stream has not ended 'naturally'

akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/NewHostConnectionPool.scala
```diff
@@ -269,7 +270,7 @@ private[client] object NewHostConnectionPool {
 
                 if (!state.isConnected && connection != null) {
                   debug(s"State change from [${previousState.name}] to [Unconnected]. Closing the existing connection.")
-                  closeConnection()
+                  closeConnection(force = false)
```
raboof: Should we disambiguate between 'natural' and 'abrupt' closures here?

akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/NewHostConnectionPool.scala
```diff
@@ -387,9 +388,9 @@ private[client] object NewHostConnectionPool {
             }
           }
 
-          def closeConnection(): Unit =
+          def closeConnection(force: Boolean): Unit =
```
raboof: Should we add some scaladoc to explain when 'force' may be 'false' here?

akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/NewHostConnectionPool.scala
jrudolph: Ah, yes that's pretty interesting. Looking at the SlotStates it seems that there's only a single case where closing would be normal and all the others would be a forced connection closure after a failure.

akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/NewHostConnectionPool.scala
jrudolph: Right, after the idle timeout e.g. Let's make it an info or even debug (should also be logged at the PoolInterfaceActor as well).

akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/NewHostConnectionPool.scala
```diff
@@ -439,8 +441,17 @@ private[client] object NewHostConnectionPool {
...
+
+          /**
+           * If regular is true connection is closed, otherwise, it is aborted.
+           *
+           * A connection should be closed regularly after a request/response with `Connection: close` has been completed.
+           * A connection should be aborted after failures.
+           */
+          def close(regular: Boolean): Unit = {
+            if (regular) requestOut.complete()
+            else requestOut.fail(new IllegalStateException("pool slot shutdown") with NoStackTrace)
```
raboof: I wonder if `regular: Boolean` should be `error: Option[String]` so we can add a 'cause' to the stacktrace in the message (since we don't have a stack trace here)?

jrudolph: Good idea. Done.

Issue #618, opened 2016-12-06
Substream Source cannot be materialized more than once failure in either unmarshaller


leachbj:
Using a `Either[Type1, Type2]` unmarshaller where the two types have spray-json unmarshallers available failed with;
 
`Failed to unmarshal Either[Type1, Type2] (attempted Type2 first). Right failure: Object is missing required member 'foo', Left failure: Substream Source cannot be materialized more than once`

jrudolph:
Yes, I agree, it seems the logic for Either unmarshallers does not work with non-strict entities.

ktoso:
Thanks for reporting! Bug indeed.

jrudolph:
Fixed by making it explicit that the marshaller only works for strict entities in #2632.

PR #2632, opened 2019-08-06
http: fail eitherUnmarshaller when entity is not strict

jrudolph:
Alternatives could be:
  * deprecate the eitherMarshaller (since trying out several unmarshallers in
    sequence, might not be what you want in any case)
  * auto `toStrict`, that would require to pull out a toStrict timeout out
    of thin air

Refs #618

### DIFF
--- a/akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/GenericUnmarshallers.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/GenericUnmarshallers.scala
@@ -4,9 +4,9 @@
 
 package akka.http.scaladsl.unmarshalling
 
+import akka.annotation.InternalApi
 import akka.http.scaladsl.unmarshalling.Unmarshaller.EitherUnmarshallingException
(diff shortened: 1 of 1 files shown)


akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/GenericUnmarshallers.scala
```diff
@@ -37,28 +37,47 @@ sealed trait LowerPriorityGenericUnmarshallers {
    * Attempt unmarshalling the entity as as `R` first (yielding `R`),
    * and if it fails attempt unmarshalling as `L` (yielding `Left`).
    *
+   * The either unmarshaller only works with strict entities, so make sure to wrap routes that want to use it with
+   * `toStrictEntity`.
```
johanandren: Maybe explicitly point out that it will fail with an exception if you miss out on this.

jrudolph: Done.

PR #2634, opened 2019-08-06
[WIP] Remove all side effects from pool state machine

jrudolph:
On top of #2631. Not sure yet, I like it like this.


jrudolph:
Let's stash that one for now.

PR #2635, opened 2019-08-07
build: pull sbt-paradox to 0.6.2

jrudolph:
That's to fix the nightly build against akka master which now also uses
paradox 0.6.1.


project/plugins.sbt
```diff
@@ -18,8 +18,9 @@ addSbtPlugin("net.virtual-void" % "sbt-dependency-graph" % "0.9.0") // for advan
 addSbtPlugin("io.spray" % "sbt-boilerplate" % "0.6.1")
 addSbtPlugin("org.foundweekends" % "sbt-bintray" % "0.5.2")
 addSbtPlugin("com.lightbend.sbt" % "sbt-javaagent" % "0.1.4")
-addSbtPlugin("com.lightbend.akka" % "sbt-paradox-akka" % "0.18")
+addSbtPlugin("com.lightbend.paradox" % "sbt-paradox" % "0.6.1")
 addSbtPlugin("com.lightbend.paradox" % "sbt-paradox-apidoc" % "0.1+9-d846d815")
```
jrudolph: @raboof do we still need to use that snapshot version?

raboof: No, looks like we can depend on 0.2 now

project/plugins.sbt
```diff
@@ -18,8 +18,9 @@ addSbtPlugin("net.virtual-void" % "sbt-dependency-graph" % "0.9.0") // for advan
 addSbtPlugin("io.spray" % "sbt-boilerplate" % "0.6.1")
 addSbtPlugin("org.foundweekends" % "sbt-bintray" % "0.5.2")
 addSbtPlugin("com.lightbend.sbt" % "sbt-javaagent" % "0.1.4")
+addSbtPlugin("com.lightbend.paradox" % "sbt-paradox" % "0.6.1")
+addSbtPlugin("com.lightbend.paradox" % "sbt-paradox-apidoc" % "0.2")
 addSbtPlugin("com.lightbend.akka" % "sbt-paradox-akka" % "0.18")
```
helena: This one is up to 0.20 now but side note: https://github.com/akka/akka/pull/27473#issuecomment-519795226

johanandren:
```
[info]   _: java.lang.IllegalStateException: Duplicate anchor [#improvements-in-docs] on [release-notes/10.1.x.html]
```

raboof:
Fixed in the next release, https://github.com/lightbend/paradox/pull/341

jrudolph:
Woops, seems we did the same work here

project/plugins.sbt
jrudolph: Yep, we should do that in another step and visually check the change.

PR #2637, opened 2019-08-08
=doc Passing `Route` to `bindAndHandle` requires `Materializer`

raboof:


akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
```diff
@@ -215,6 +215,10 @@ class HttpExt private[http] (private val config: Config)(implicit val system: Ex
    *
    * To configure additional settings for a server started using this method,
    * use the `akka.http.server` config section or pass in a [[akka.http.scaladsl.settings.ServerSettings]] explicitly.
+   *
+   * You can pass a [[akka.http.scaladsl.server.Route]] into this function if you
```
jrudolph: Wouldn't you get an error about no materializer anyway, because this method also needs an implicit Materializer?

jrudolph: No, indeed, that error seems to be masked by the fact that you don't even get a handler.

raboof:
(if we merge this we should update it again in #2893 - perhaps not worth it?)

ennru:
So you'd want to close this one? Go ahead.

raboof:
Let's keep it open until we either merge or close #2893

ennru:
#2893 is now merged. Close this?

PR #2643, opened 2019-08-14
Fix performance regression introduced in #2627

jrudolph:


jrudolph:
> Did you benchmark with non-strict requests?

No we currently don't do benchmarks with non-strict requests/responses but they are probably much slower anyway.

jrudolph:
Created #2644 to start tracking performance of streamed entity responses as well.

PR #2644, opened 2019-08-14
Add benchmark to compare strict vs streamed response types

jrudolph:
Currently, it seems streamed response are ~1 order of magnitude slower at least. I guess some of that is somewhat expected as streaming is a memory optimization and not a performance optimization. Might still make sense to optimize if there are any low-hanging fruit. Also the benchmark will be used to look out for regressions.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
@@ -373,7 +373,7 @@ private[http] object HttpServerBluePrint {
     val shape = new BidiShape(requestParsingIn, requestPrepOut, httpResponseIn, responseCtxOut)
 
     def createLogic(effectiveAttributes: Attributes) = new GraphStageLogic(shape) {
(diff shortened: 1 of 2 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -373,7 +373,7 @@ private[http] object HttpServerBluePrint {
     val shape = new BidiShape(requestParsingIn, requestPrepOut, httpResponseIn, responseCtxOut)
 
     def createLogic(effectiveAttributes: Attributes) = new GraphStageLogic(shape) {
-      val pullHttpResponseIn = () => pull(httpResponseIn)
+      val pullHttpResponseIn = () => tryPull(httpResponseIn)
```
raboof: Was this intended/needed?

jrudolph: It's needed. It failed for the benchmark which doesn't use the full pipeline and probably triggers different race conditions on closing.

PR #2645, opened 2019-08-14
Streamed response processing performance improvements

jrudolph:
Culprits were `Future.failed` and `Source.concat`. This improved speed for streamed responses from ~20x slower to only ~10x slower than strict responses (which is still a lot).

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/rendering/HttpResponseRendererFactory.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/rendering/HttpResponseRendererFactory.scala
@@ -82,9 +82,8 @@ private[http] class HttpResponseRendererFactory(
                 push(out, outElement)
                 if (close) completeStage()
               case HeadersAndStreamedEntity(headerData, outStream) =>
-                try {
-                  transfer(headerData, outStream)
-                } catch {
+                try transfer(headerData, outStream)
+                catch {
                   case NonFatal(e) =>
                     transferring = false
                     log.error(e, s"Rendering of response failed because response entity stream materialization failed with '${e.getMessage}'. Sending out 500 response instead.")
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/rendering/HttpResponseRendererFactory.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/rendering/HttpResponseRendererFactory.scala
@@ -83,8 +83,7 @@ private[http] class HttpResponseRendererFactory(
                 if (close) completeStage()
               case HeadersAndStreamedEntity(headerData, outStream) =>
                 try {
-                  push(out, ResponseRenderingOutput.HttpData(headerData))
-                  transfer(outStream)
+                  transfer(headerData, outStream)
                 } catch {
                   case NonFatal(e) =>
                     transferring = false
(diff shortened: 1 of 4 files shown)


raboof:
Jenkins:
```
java.lang.AssertionError: expected OnNext, found OnError(java.lang.IllegalArgumentException: Cannot push port (HttpResponseRenderer.out(1347198232)) twice, or before it being pulled
	at akka.stream.stage.GraphStageLogic.push(GraphStage.scala:623)
	at akka.http.impl.engine.rendering.HttpResponseRendererFactory$HttpResponseRenderer$$anon$1$$anon$2.onPush(HttpResponseRendererFactory.scala:93)
```

jrudolph:
> Jenkins:
> 
> ```
> java.lang.AssertionError: expected OnNext, found OnError(java.lang.IllegalArgumentException: Cannot push port (HttpResponseRenderer.out(1347198232)) twice, or before it being pulled
> 	at akka.stream.stage.GraphStageLogic.push(GraphStage.scala:623)
> 	at akka.http.impl.engine.rendering.HttpResponseRendererFactory$HttpResponseRenderer$$anon$1$$anon$2.onPush(HttpResponseRendererFactory.scala:93)
> ```

That was interesting. The problem here was that headers are sent out before materialization but that didn't take into account that I might want to also handle errors during materialization with an error response. I now moved sending out headers into the substream.

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -422,9 +425,11 @@ private[http] object HttpServerBluePrint {
...
             else response0.mapEntity { e =>
               val (newEntity, fut) = HttpEntity.captureTermination(e)
-              fut.failed.foreach { ex =>
-                log.error(ex, s"Response stream for [${requestStart.debugString}] failed with '${ex.getMessage}'. Aborting connection.")
-              }(materializer.executionContext)
+              fut.onComplete {
+                case Failure(ex) =>
+                  log.error(ex, s"Response stream for [${requestStart.debugString}] failed with '${ex.getMessage}'. Aborting connection.")
+                case _ => // ignore
+              }(ExecutionContexts.sameThreadExecutionContext)
```
raboof: :+1:

akka-http-core/src/main/scala/akka/http/impl/engine/rendering/RenderSupport.scala
```diff
@@ -70,7 +70,7 @@ private[rendering] object RenderSupport {
                         skipEntity: Boolean = false): Source[ByteString, Any] = {
     val messageStart = Source.single(header)
     val messageBytes =
-      if (!skipEntity) (messageStart ++ entityBytes).mapMaterializedValue(_ => ())
+      if (!skipEntity) messageStart ++ entityBytes
```
raboof: nice

akka-http-core/src/main/scala/akka/http/impl/engine/rendering/HttpResponseRendererFactory.scala
```diff
@@ -240,8 +249,16 @@ private[http] class HttpResponseRendererFactory(
           def renderContentLengthHeader(contentLength: Long) =
             if (status.allowsEntity) r ~~ `Content-Length` ~~ contentLength ~~ CrLf else r
 
-          def byteStrings(entityBytes: => Source[ByteString, Any]): Source[ResponseRenderingOutput, Any] =
-            renderByteStrings(r.asByteString, entityBytes, skipEntity = noEntity).map(ResponseRenderingOutput.HttpData(_))
```
raboof: Ah, so the concat was in here, and is replaced with simply separate 'push' calls here. Makes sense.

It seems `renderByteStrings` is now unused entirely, so perhaps we should remove it?

PR #2630, opened 2019-07-31
Rewrite pool interface without actors

jrudolph:
* adding `PoolInterface` as yet another abstraction layer to get the sprawling layers of the client implementation under control.
 * reimplement `PoolInterface` as a `GraphStage` instead of using `ActorPublisher`/`ActorSubscriber`

Refs #2617

PR #2653, opened 2019-08-20
Reduce number of stages needed to implement CaptureMaterializationAndTerminationOp

jrudolph:
Also report underlying reason for response entity failure when materialization times out.

Follow up to #2630.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
@@ -25,6 +25,7 @@ import scala.util.Try
 /**
  * INTERNAL API
  */
+@InternalApi
 private[http] object StreamUtils {
 
   /**
@@ -83,9 +84,7 @@ private[http] object StreamUtils {
           failStage(ex)
         }
 
-        override def postStop(): Unit = {
-          terminationPromise.trySuccess(())
-        }
+        override def postStop(): Unit = terminationPromise.trySuccess(())
 
         setHandlers(in, out, this)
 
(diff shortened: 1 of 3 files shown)


akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
```diff
@@ -49,26 +53,52 @@ private[http] object StreamUtils {
...
+    val materializationPromise = Promise[Unit]()
+    val killResult = Promise[Unit]()
+    val killSwitch = new KillSwitch {
+      override def shutdown(): Unit = killResult.trySuccess(())
+      override def abort(ex: Throwable): Unit = killResult.tryFailure(ex)
+    }
     val transformer = new SimpleLinearGraphStage[T] {
       override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new GraphStageLogic(shape) with InHandler with OutHandler {
+        override def preStart(): Unit = {
+          materializationPromise.trySuccess(())
```
raboof: Hard to say if this might be 'too early'?

akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
```diff
@@ -262,20 +292,19 @@ private[http] object StreamUtils {
...
-      val (newSource, completion) =
-        StreamUtils.captureTermination(newSource0.mapMaterializedValue { mat =>
-          materializationPromise.trySuccess(())
-          mat
-        })
-
-      (newSource, (materializationPromise.future, completion, killSwitchPromise.future))
+  object CaptureMaterializationAndTerminationOp extends EntityStreamOp[(Future[Any], Future[Unit], Option[KillSwitch])] {
+    def strictM: (Future[Any], Future[Unit], Option[KillSwitch]) = (Future.successful(()), Future.successful(()), None)
+    def apply[T, Mat](source: Source[T, Mat]): (Source[T, Mat], (Future[Any], Future[Unit], Option[KillSwitch])) = {
```
raboof: Why not:
```suggestion
    def apply[T, Mat](source: Source[T, Mat]): (Source[T, Mat], (Future[Mat], Future[Unit], Option[KillSwitch])) = {
```
? Might make it clearer what it is even if we're not using the value ;)

akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
jrudolph: Do you have some concrete issue in mind with that? What we use that for is for checking whether the stream has been materialized in time. So, for the pool it's enough to have any indication that any stage of the stream has been materialized.

akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
jrudolph: It won't work like this because it doesn't fit the `EntityStreamOp` supertype and also it doesn't actually return the materialized value.

But you are right that this result type is a bit opaque. I'll introduce a small case class for it for better readability.

akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
raboof: Agreed it's OK for the current use case and that's good enough since it's an internal API anyway :+1:

Issue #2651, opened 2019-08-20
java.lang.AbstractMethodError when using Akka HTTP with Akka 2.6 M6


fvyaba:
I'm using akka-http(java:oracle jdk1.8.0_202, os: macos), just found 'java.lang.AbstractMethodError' under benchmark(using wrk).

- pom.xml(just fragement),listed below:
```
        <properties>
		<scala.version>2.13</scala.version>
		<akka-core.version>2.6.0-M6</akka-core.version>
		<akka-http.version>10.1.9</akka-http.version>
	</properties>
	<dependencies>
		<dependency>
			<groupId>com.typesafe.akka</groupId>
			<artifactId>akka-actor_${scala.version}</artifactId>
			<version>${akka-core.version}</version>
		</dependency>
		<dependency>
			<groupId>com.typesafe.akka</groupId>
			<artifactId>akka-stream_${scala.version}</artifactId>
			<version>${akka-core.version}</version>
		</dependency>
		<dependency>
			<groupId>com.typesafe.akka</groupId>
			<artifactId>akka-http_${scala.version}</artifactId>
			<version>${akka-http.version}</version>
		</dependency>
	</dependencies>
```

- server side code:
```
public class ServerEngine {
  public static void start(int port) {
    ActorSystem system = ActorSystem.create("akka-http");
    Http http = Http.get(system);
    ActorMaterializer materializer = ActorMaterializer.create(system);
    HttpFacade app = new HttpFacade();
    Flow<HttpRequest, HttpResponse, NotUsed> routeFlow = app.createRoute().flow(system, materializer);

    final CompletionStage<ServerBinding> binding = http.bindAndHandle(routeFlow, ConnectHttp.toHost("127.0.01", port),
        materializer);
    binding.thenRun(() -> {
      System.out.println("akka http server listening on port " + port);
    });
  }

  public static void main(String[] args) {
    start(8080);
  }
}
```

```
public class HttpFacade extends AllDirectives {
  public Route createRoute() {
    return concat(path("", () -> get(() -> {
      return complete("ok");
    })));
  }
}
```

- benchmark script(need install [wrk](https://github.com/wg/wrk.git) first)
`wrk -t16 -c100 -d30s --timeout=15s http://127.0.0.1:8080`

- err stack trace
```
[...]

PR #2654, opened 2019-08-21
build: add command line flag that allows running akka-http-core tests against akka master branch

jrudolph:
The idea is to add another build step in nightlies that exercises those tests, so we'll catch binary incompatibilities (i.e. ones we might otherwise miss because they are fixed by recompiling against akka master) like #2651 early on.

Currently, only akka-http-core tests would be included in those tests but just that would already add a lot of coverage to that case.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -148,6 +148,13 @@ lazy val httpCore = project("akka-http-core")
   .dependsOn(parsing)
   .addAkkaModuleDependency("akka-stream", "provided")
   .addAkkaModuleDependency("akka-stream-testkit", "test")
+  .addAkkaModuleDependency(
+    "akka-stream",
+    "test",
+    shouldUseSourceDependency = true,
+    uri("git://github.com/akka/akka.git#master"),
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -148,6 +148,13 @@ lazy val httpCore = project("akka-http-core")
   .dependsOn(parsing)
   .addAkkaModuleDependency("akka-stream", "provided")
   .addAkkaModuleDependency("akka-stream-testkit", "test")
+  .addAkkaModuleDependency(
+    "akka-stream",
+    "test",
```
jrudolph: So, we compile against a published artifact but then override it here for test compilation / execution.

jrudolph:
Running a test build with the new build step here:

https://jenkins.akka.io:8498/job/akka-http-nightly/600/AKKA_VERSION=default,Node=akka-http-nightly-node,SCALA_VERSION=2.12.9,jdk=AdoptOpenJDK%2011/console

PR #2656, opened 2019-08-21
Update mima

raboof:
Needs release of mima

Some exclusions needed to be adapted due to the change in https://github.com/lightbend/mima/pull/362

### DIFF
--- a/akka-http-core/src/main/mima-filters/10.1.8.backwards.excludes
+++ b/akka-http-core/src/main/mima-filters/10.1.8.backwards.excludes
@@ -44,16 +44,16 @@ ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.impl.engine.render
 
 #### The following are against 10.1.5, but weren't caught by a previous version of MiMa
 
-ProblemFilters.exclude[IncompatibleResultTypeProblem]("akka.http.javadsl.Http.get")
-ProblemFilters.exclude[IncompatibleResultTypeProblem]("akka.http.scaladsl.Http.apply")
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.javadsl.Http.get")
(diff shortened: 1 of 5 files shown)


raboof:
preferably also with https://github.com/lightbend/mima/pull/378

akka-http-core/src/main/mima-filters/10.1.8.backwards.excludes
```diff
@@ -44,16 +44,16 @@ ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.impl.engine.render
 
 #### The following are against 10.1.5, but weren't caught by a previous version of MiMa
 
-ProblemFilters.exclude[IncompatibleResultTypeProblem]("akka.http.javadsl.Http.get")
-ProblemFilters.exclude[IncompatibleResultTypeProblem]("akka.http.scaladsl.Http.apply")
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.javadsl.Http.get")
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.scaladsl.Http.apply")
```
jrudolph: That's quite a heavy exclusion. Somehow the forwarder signatures changed since then? I filed https://github.com/lightbend/mima/issues/381 to allow for more fine-grained exclusions.

raboof: looking at this again I wonder if we could move these exclusions to `10.1.5.backwards.excludes`, that'd make them a lot less painful

jrudolph: Yeah, might make sense. But let's merge anyway, we can still fix that afterwards.

raboof: (yes, we can. will do in a separate PR)

PR #2662, opened 2019-08-27
Scaladoc: link to source and canonical

ennru:
Add canonical URLs in Scaladoc and use the more accurate links to source code.


jrudolph:
That broke doc generation on 2.11: https://travis-ci.org/akka/akka-http/jobs/577758724

ennru:
Oh, I didn't think of that being done. I'll try to come up with a work-around.

jrudolph:
> Oh, I didn't think of that being done. I'll try to come up with a work-around.

Already on it.

jrudolph:
See #2667

PR #2667, opened 2019-08-28
docs: disable unsupported `-doc-canonical-base-url` for Scala 2.11

jrudolph:
Follow up to #2662.

### DIFF
--- a/project/Doc.scala
+++ b/project/Doc.scala
@@ -52,7 +52,10 @@ object Scaladoc extends AutoPlugin {
     })
 
   def scaladocOptions(scalaBinaryVersion: String, ver: String, base: File, plugins: Seq[File]): List[String] = {
-    val urlString = GitHub.url(ver) + "/€{FILE_PATH_EXT}#L€{FILE_LINE}"
+    val urlString = GitHub.url(ver) +
+                    // supported from Scala 2.12.9
+                    (if (scalaBinaryVersion != "2.11") "€{FILE_PATH_EXT}#L€{FILE_LINE}" else "€{FILE_PATH}.scala")
+
     val opts = List(
       "-implicits",
       "-groups",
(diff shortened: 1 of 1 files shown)


project/Doc.scala
```diff
@@ -50,7 +51,7 @@ object Scaladoc extends AutoPlugin {
       docs
     })
 
-  def scaladocOptions(ver: String, base: File, plugins: Seq[File]): List[String] = {
+  def scaladocOptions(scalaBinaryVersion: String, ver: String, base: File, plugins: Seq[File]): List[String] = {
     val urlString = GitHub.url(ver) + "/€{FILE_PATH_EXT}#L€{FILE_LINE}"
```
ennru: This needs to be reverted as well for pre-Scala 2.12.9.

jrudolph: Ah, indeed, will do.

Does it work with 2.13.0 and was backported to 2.12.9?

ennru: Exactly. It is 2.12.9+

project/Doc.scala
```diff
@@ -50,19 +51,22 @@ object Scaladoc extends AutoPlugin {
       docs
     })
 
-  def scaladocOptions(ver: String, base: File, plugins: Seq[File]): List[String] = {
-    val urlString = GitHub.url(ver) + "/€{FILE_PATH_EXT}#L€{FILE_LINE}"
+  def scaladocOptions(scalaBinaryVersion: String, ver: String, base: File, plugins: Seq[File]): List[String] = {
+    val urlString = GitHub.url(ver) +
+                    // supported from Scala 2.12.9
+                    (if (scalaBinaryVersion != "2.11") "€{FILE_PATH_EXT}#L€{FILE_LINE}" else "€{FILE_PATH}.scala")
```
jrudolph: I removed an extra `/` at the beginning I noticed in the generated docs. Will save 1 byte per documented class.

project/Doc.scala
jrudolph: :+1:

jrudolph: Great, thanks.

PR #2668, opened 2019-08-28
fix: akka-stream-testkit version

vitaliis:
The old one (19) results in:
sbt.librarymanagement.ResolveException: unresolved dependency: com.typesafe.akka#akka-stream-testkit_2.13;2.5.19: not found

## References

References #xxxx

### DIFF
--- a/docs/src/main/paradox/routing-dsl/testkit.md
+++ b/docs/src/main/paradox/routing-dsl/testkit.md
@@ -9,7 +9,7 @@ route logic easy and convenient. This "route test DSL" is made available with th
 To use Akka HTTP TestKit, add the module to your project:
 
 @@dependency [sbt,Gradle,Maven] {
-  group="com.typesafe.akka" artifact="akka-stream-testkit_$scala.binary.version$" version="2.5.23"
+  group="com.typesafe.akka" artifact="akka-stream-testkit_$scala.binary.version$" version="$akka.version$"
   group2="com.typesafe.akka" artifact2="akka-http-testkit_$scala.binary.version$" version2="$project.version$"
 }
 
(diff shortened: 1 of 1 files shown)


docs/src/main/paradox/routing-dsl/testkit.md
```diff
@@ -9,7 +9,7 @@ route logic easy and convenient. This "route test DSL" is made available with th
 To use Akka HTTP TestKit, add the module to your project:
 
 @@dependency [sbt,Gradle,Maven] {
-  group="com.typesafe.akka" artifact="akka-stream-testkit_$scala.binary.version$" version="2.5.19"
+  group="com.typesafe.akka" artifact="akka-stream-testkit_$scala.binary.version$" version="2.5.23"
```
jrudolph:
```suggestion
  group="com.typesafe.akka" artifact="akka-stream-testkit_$scala.binary.version$" version="$akka.version$"
```

jrudolph:
OK TO TEST

docs/src/main/paradox/routing-dsl/testkit.md
raboof: Hmm, the version we recommend in the docs might still be higher than the version we really require though... perhaps we should introduce yet another variable for this?

jrudolph: Wouldn't we always recommend the latest? Right now, we seem to use `akka.version` throughout, so maybe let's keep it consistent for now?

I guess we can think about always recommending the latest version of akka?

raboof: Yeah, not necessary for this PR indeed ;)

PR #2669, opened 2019-08-29
Update authenticateOAuth2.md

PiotrJander:
# Pull Request Checklist

* [x] Have you updated the documentation?

## Purpose

It is misleading that the documentation does not mention that the token can be coming from places other than the `Authorization` header, even though these are supported by the existing implementation. There are legitimate use cases where the token should be sent in the URI as a query parameter, for example JavaScript WebSocket API does not support adding the Authorization header to the request, so a query parameter is arguably the next best option.


jrudolph:
OK TO TEST

jrudolph:
Thanks @PiotrJander.

Issue #2577, opened 2019-06-25
Uri.Path() is ambiguous about accepting % encoded vs not encoded


jsweet-datto:
If you put characters that should be percent-encoded in a string passed to Uri.Path(), it will automatically percent-encode them (e.g. turn a space into %20).  Unless the character itself is a percent, in which case it assumes you are trying to percent encode this.  I guess this is sort of by design, but the ambiguity definitely can create bugs.  Ideally there'd be two variants you could pass to, one that expected unencoded input and encoded everything (including %), and another that expects encoded input and rejects characters that need to be percent-encoded.

jrudolph:
Thanks for the report, @jsweet-datto. I agree it can be confusing. I think we would need some concrete proposals about what could be changed.

vasiliybondarenko:
Can anyone take a look at my approach in linked PR?

PR #2670, opened 2019-08-29
Uri.Path() should accept raw input with %

vasiliybondarenko:
## Purpose

The main purpose of the PR is to make Uri.Path.apply() method non ambiguous regarding accepting % character. 
For instance 
```scala
Path("/example/%12%", encodingMode = EncodingMode.NonEncoded)
```  
accepts raw unencoded string with % characters 
while 
```scala
Path("/example/%12%", encodingMode = EncodingMode.Encoded)
``` 
throws error because it assumes input is already encoded.

to keep backward compatibility 
```scala
Path("/example/%12%")
```  

is the same as 

```scala
Path("/example/%12%", encodingMode = EncodingMode.Deafult)
```  

`EncodingMode.Deafult` means apply method decides whether the input ic encoded or not according to % symbol in input

## References
#2577 issue

References #2577

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/Uri.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/Uri.scala
@@ -18,6 +18,7 @@ import akka.http.impl.model.parser.UriParser
 import akka.http.impl.model.parser.CharacterClasses._
 import akka.http.impl.util._
 import Uri._
+import akka.http.scaladsl.model.Uri.EncodingMode.Default
 
 /**
  * An immutable model of an internet URI as defined by http://tools.ietf.org/html/rfc3986.
@@ -41,7 +42,7 @@ sealed abstract case class Uri(scheme: String, authority: Authority, path: Path,
   /**
    * Returns the query part of the Uri in its decoded form.
    */
-  def queryString(charset: Charset = UTF8): Option[String] = rawQueryString.map(s => decode(s, charset))
+  def queryString(charset: Charset = UTF8): Option[String] = rawQueryString.map(s => decode(s, charset, EncodingMode.Default))
 
   /**
    * The effective port of this Uri given the currently set authority and scheme values.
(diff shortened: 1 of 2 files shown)


lightbend-cla-validator:
Hi @vasiliybondarenko,

raboof:
OK TO TEST

vasiliybondarenko:
There are binary incompatibility issues due to changing public api. Is it acceptable in our case?

ennru:
This has now been revisited for Akka HTTP 10.2 in https://github.com/akka/akka-http/pull/3003

jrudolph:
#3003 is about the query part of the URI while this one is about the path component. If we come back to this one we should see if can learn something from our solution in #3003.

raboof:
> There are binary incompatibility issues due to changing public api. Is it acceptable in our case?

Since those really are public API's I'm afraid we'll have to maintain binary compatibility here.

raboof:
> The main purpose of the PR is to make Uri.Path.apply() method non ambiguous regarding accepting % character

`Uri.Path.apply()` currently expects a percent-encoded string.

This PR proposes adding a variation that expects a plain Java string. I'm not entirely sure that makes sense, though: for example in

    Uri.Path("/foo/bar /baz")

Should the result be `Path("/foo/bar%20/baz")`? or `Path("/foo/bar%20%2Fbaz")`?

With that in mind I think we shouldn't have an API that turns an unencoded string into a path, but leave it up to the user to pass in individual segments like this: `Path("")./("foo")./("bar /baz")` which unambiguously produces `Path("/foo/bar%20%2Fbaz")`. If the user want to take a `/`-separated string and turn that into segments that *otherwise* need to be encoded, that's then up to them.

I do think it would make sense to document this better, though. And perhaps provide a `Path()` for the empty starting point?

raboof:
> `Path("")./("foo")./("bar /baz")` (...) And perhaps provide a `Path()` for the empty starting point?

Ah, there's `Path./("foo")./("bar /baz")` - never mind.

Let's close this one, a PR for clarifying the docs in this regard would be great!

PR #2671, opened 2019-08-30
Bump up dependencies.

takayahilton:

### DIFF
--- a/project/AkkaDependency.scala
+++ b/project/AkkaDependency.scala
@@ -14,7 +14,8 @@ object AkkaDependency {
   // else if akka version is "default", then the hard coded default will be used (jenkins doesn't allow empty values for config axis)
   // else if akka.version is anything else, then the given version will be used
 
-  val defaultAkkaVersion = "2.5.25"
+  // Updated only when needed, https://github.com/akka/akka/issues/26985
+  val defaultAkkaVersion = "2.5.23"
   val akkaVersion =
     System.getProperty("akka.build.version", defaultAkkaVersion) match {
       case "default" => defaultAkkaVersion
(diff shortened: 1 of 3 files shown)


project/AkkaDependency.scala
```diff
@@ -14,7 +14,7 @@ object AkkaDependency {
   // else if akka version is "default", then the hard coded default will be used (jenkins doesn't allow empty values for config axis)
   // else if akka.version is anything else, then the given version will be used
 
-  val defaultAkkaVersion = "2.5.23"
+  val defaultAkkaVersion = "2.5.25"
```
jrudolph: We usually bump the Akka version only if strictly needed.

jrudolph:
OK TO TEST

project/AkkaDependency.scala
raboof:
```suggestion
  // Updated only when needed, https://github.com/akka/akka/issues/26985
  val defaultAkkaVersion = "2.5.25"
```

project/AkkaDependency.scala
```diff
@@ -14,6 +14,7 @@ object AkkaDependency {
   // else if akka version is "default", then the hard coded default will be used (jenkins doesn't allow empty values for config axis)
   // else if akka.version is anything else, then the given version will be used
 
+  // Updated only when needed, https://github.com/akka/akka/issues/26985
```
jrudolph: Great, thanks :+1:

jrudolph:
Thanks, @takayahilton.

laszlovandenhoek:
Jackson Databind 2.9.9 is still vulnerable to CVE-2019-14439. While we're upgrading, we should go for 2.9.9.3. I'll prepare a separate PR.

PR #2678, opened 2019-09-02
Reproducer and potential fix for #2349

jrudolph:
Refs #2349

~The fix builds on top of akka/akka#27594 (Akka 2.6) and akka/akka#27608 (Akka 2.5.x) so it will not build on Jenkins but it can be verified locally using `sbt -Dakka.sources=`.~

Updated to run on Akka 2.5.26 which has the fix.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/NewHostConnectionPool.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/NewHostConnectionPool.scala
@@ -505,11 +505,14 @@ private[client] object NewHostConnectionPool {
               slot.onConnectionCompleted()
             }
           override def onUpstreamFailure(ex: Throwable): Unit =
-            if (connectionEstablished)
-              withSlot { slot =>
+            withSlot { slot =>
+              if (connectionEstablished) {
                 slot.debug("Connection failed")
                 slot.onConnectionFailed(ex)
               }
+              // otherwise, rely on connection.onComplete to fail below
+              // (connection error is sent through matValue future and through the stream)
+            }
 
           def onPull(): Unit = () // emitRequests makes sure not to push too early
 
@@ -539,10 +542,13 @@ private[client] object NewHostConnectionPool {
           responseIn.pull()
 
           slot.debug("Establishing connection")
+          import Attributes.CancellationStrategy
+          import CancellationStrategy._
           val connection =
             Source.fromGraph(requestOut.source)
               .viaMat(connectionFlow)(Keep.right)
               .to(responseIn.sink)
+              .addAttributes(Attributes(CancellationStrategy(AfterDelay(1.second, FailStage))))
               .run()(subFusingMaterializer)
 
           val slotCon = new SlotConnection(slot, requestOut, responseIn)
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/NewHostConnectionPool.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/NewHostConnectionPool.scala
@@ -19,7 +19,7 @@ import akka.http.scaladsl.Http
 import akka.http.scaladsl.model.{ HttpEntity, HttpRequest, HttpResponse, headers }
 import akka.http.scaladsl.settings.ConnectionPoolSettings
 import akka.stream._
-import akka.stream.scaladsl.{ Flow, Keep, Sink, Source }
+import akka.stream.scaladsl.{ Flow, Sink }
 import akka.stream.stage.{ GraphStage, GraphStageLogic, InHandler, OutHandler }
 import akka.util.OptionVal
 
@@ -542,13 +542,9 @@ private[client] object NewHostConnectionPool {
           responseIn.pull()
 
           slot.debug("Establishing connection")
-          import Attributes.CancellationStrategy
-          import CancellationStrategy._
           val connection =
-            Source.fromGraph(requestOut.source)
-              .viaMat(connectionFlow)(Keep.right)
-              .to(responseIn.sink)
-              .addAttributes(Attributes(CancellationStrategy(AfterDelay(1.second, FailStage))))
+            connectionFlow
+              .join(Flow.fromSinkAndSource(responseIn.sink, requestOut.source))
               .run()(subFusingMaterializer)
 
           val slotCon = new SlotConnection(slot, requestOut, responseIn)
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
@@ -26,6 +26,9 @@ import akka.http.scaladsl.model.headers.Host
 import akka.http.scaladsl.model.ws.{ Message, WebSocketRequest, WebSocketUpgradeResponse }
 import akka.http.scaladsl.settings.{ ClientConnectionSettings, ConnectionPoolSettings, ServerSettings }
 import akka.http.scaladsl.util.FastFuture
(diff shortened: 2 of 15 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/NewHostConnectionPool.scala
```diff
@@ -539,10 +542,13 @@ private[client] object NewHostConnectionPool {
           responseIn.pull()
 
           slot.debug("Establishing connection")
+          import Attributes.CancellationBehavior
+          import CancellationBehavior._
           val connection =
             Source.fromGraph(requestOut.source)
               .viaMat(connectionFlow)(Keep.right)
               .to(responseIn.sink)
+              .addAttributes(Attributes(CancellationBehavior(AfterDelay(1.second, FailStage))))
```
jrudolph: This delays cancellation in the connection flow between all stages. We should add this attribute wherever we run or build streams that contain BidiStages that might exhibit the problematic behavior.

jrudolph:
I mostly finished this:
 * upgrade to Akka 2.5.26
 * added attribute to our main stream setups, client/server, ws, http/2
 * setting for client and server to set the timeout (I don't see reasons why it would need to be changed but you never know)

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
```diff
@@ -245,7 +248,8 @@ class HttpExt private[http] (private val config: Config)(implicit val system: Ex
                 masterTerminator.registerConnection(connectionTerminator)(fm.executionContext)
                 future // drop the terminator matValue, we already registered is which is all we need to do here
             }
-
+            // attribute is also added to serverLayer but we also want it to cover the full stack
+            .addAttributes(cancellationStrategyAttributeForDelay(settings.streamCancellationDelay))
```
johanandren: Hmm, do we really, what does it help with in the context of the stream of incoming connections, there's no bidi-stuff going on here?

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
```diff
@@ -360,7 +364,9 @@ class HttpExt private[http] (private val config: Config)(implicit val system: Ex
     val server = HttpServerBluePrint(settings, log, isSecureConnection)
       .addAttributes(HttpAttributes.remoteAddress(remoteAddress))
 
-    server atop delayCancellationStage(settings)
+    server
+      .atop(delayCancellationStage(settings)) // TODO: figure out if we still need that at all
+      .addAttributes(cancellationStrategyAttributeForDelay(settings.streamCancellationDelay))
```
johanandren: Yeah, I would have guessed the delay cancellation stage becomes no longer needed with the delay built in

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
```diff
@@ -462,6 +468,8 @@ class HttpExt private[http] (private val config: Config)(implicit val system: Ex
     val hostHeader = if (port == connectionContext.defaultPort) Host(host) else Host(host, port)
     val layer = clientLayer(hostHeader, settings, log)
     layer.joinMat(_outgoingTlsConnectionLayer(host, port, settings, connectionContext, log))(Keep.right)
+      // already added in clientLayer but needed here again to also include transport layer
+      .addAttributes(cancellationStrategyAttributeForDelay(settings.streamCancellationDelay))
```
johanandren: since this is `(clientLayer.join(tlsLayer)).addAttributes(delay)` that also covers the inner one, should it only be here then?

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
```diff
@@ -762,6 +772,8 @@ class HttpExt private[http] (private val config: Config)(implicit val system: Ex
 
     webSocketClientLayer(request, settings, log)
       .join(_outgoingTlsConnectionLayer(host, port, settings.withLocalAddressOverride(localAddress), ctx, log))
+      // also added webSocketClientLayer but we want to make sure it covers the whole stack
+      .addAttributes(cancellationStrategyAttributeForDelay(settings.streamCancellationDelay))
```
johanandren: Same here, `(blueprint.join(tls)).addAttr(delay)` should be enough, or is the clientLayer used separately somewhere?

akka-http2-support/src/main/scala/akka/http/scaladsl/Http2.scala
```diff
@@ -67,6 +67,7 @@ final class Http2Ext(private val config: Config)(implicit val system: ActorSyste
             httpPlusSwitching(http1, http2).addAttributes(Http.prepareAttributes(settings, incoming))
               .watchTermination()(Keep.right)
               .join(incoming.flow)
+              .addAttributes(Http.cancellationStrategyAttributeForDelay(settings.streamCancellationDelay))
```
johanandren: Could it make sense to only add attributes in one place, combining `Http.prepareAttributes` with the stream cancelation delay one and only do a single `addAttributes`?

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
jrudolph: I think it's the per-connection stream here, isn't it? I thought, `fullLayer` above would add TCP and TLS which are bidi-like, so adding the streams cancellation delay here as well could be interesting.

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
jrudolph: That's right, it also covers the inner one but the idea is, that if you use `clientLayer` alone, it should still be applied. Here TCP and TLS are added which would otherwise be missing it.

akka-http2-support/src/main/scala/akka/http/scaladsl/Http2.scala
jrudolph: If we would have a solution for the needed duplication above where we want to enable it for all cuts of the stack that we provide APIs for, then maybe yes. When we cannot solve that, we probably cannot clean it up easily.

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
johanandren: Ah, that is public API so users could be using it, missed that.

akka-http2-support/src/main/scala/akka/http/scaladsl/Http2.scala
johanandren: I think I didn't think that all the way through, we wouldn't want the other `Http.prepareAttributes` to apply to the incoming, so NVM

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
johanandren: Yep, NVM, I thought it was on the outer layer of stream but it isn't.

akka-http2-support/src/main/scala/akka/http/scaladsl/Http2.scala
johanandren: I think I didn't think that all the way through, we wouldn't want the other `Http.prepareAttributes` to apply to the incoming, so NVM

jrudolph:
There's quite a bit broken still. I have a branch with already a few updates that I didn't yet push.

jrudolph:
I'm going to close this one and set up a new, clean branch that with reduced scope.

Issue #2673, opened 2019-08-30
Fail fast when building with jdk9+


raboof:
The Akka HTTP build specifies `-target:jvm-1.8` for the scalac options. However, this is not sufficient to build an Akka HTTP library that works on JDK8: because even with `-target:jvm-1.9`, the code is still linked to the newer standard library. Because of this, this might cause problems like:
```
java.lang.NoSuchMethodError: java.nio.ByteBuffer.clear()Ljava/nio/ByteBuffer
```
when a downstream project later tries to run with the jdk8 standard library.

In Akka we have fixed the build to produce correct JDK8 classes even when building with a later JDK, because there we need that later JDK anyway for the reactive streams API's. For Akka HTTP, I would suggest just failing fast when accidentally compiling with a newer JDK.

When we drop support for jdk8 we can start using the `-release` flags, but that'll be a while :)

jrudolph:
Maybe we could use the same change as here in here: 

https://github.com/spray/spray-json/commit/5bf165745bc50027f1b6239ad4c7a98dfde25bb5

i.e. guard the addition of the `-release 8` flag looking at Scala and JDK versions.

PR #2680, opened 2019-09-02
Always target java 8 bootclasspath

raboof:
Refs #2673

Tested with javap (`javap -p -c akka-http-core/target/scala-2.12/classes/akka/http/impl/engine/parsing/HttpHeaderParser\$.class | grep clear` should be `Method java/nio/ByteBuffer.clear:()Ljava/nio/Buffer;` and not `Method java/nio/ByteBuffer.clear:()Ljava/nio/ByteBuffer`).

Also tested with cinnamon, also looks good.

### DIFF
--- a/project/Common.scala
+++ b/project/Common.scala
@@ -1,0 +1,53 @@
+/*
+ * Copyright (C) 2017-2019 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka
+
+import sbt._
+import Keys._
+
+object Common extends AutoPlugin {
+  override def trigger: PluginTrigger = allRequirements
+  override lazy val projectSettings = Seq(
+    scalacOptions ++= Seq(
+      "-deprecation",
+      "-encoding", "UTF-8", // yes, this is 2 args
+      "-target:jvm-1.8",
+      "-unchecked",
+      "-Xlint",
+      // "-Yno-adapted-args", //akka-http heavily depends on adapted args and => Unit implicits break otherwise
+      "-Ywarn-dead-code"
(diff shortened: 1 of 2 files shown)


project/Common.scala
```diff
@@ -0,0 +1,53 @@
...
+      if (scalaMinorVersion.value >= 12 && !sys.props("java.version").startsWith("1.") /* i.e. Java version >= 9 */)
+        Seq("-release", "8")
+      else
+        Nil
+    },
+    javacOptions ++= Seq(
+      "-encoding", "UTF-8",
+    ),
+    javacOptions ++= {
+      if (!sys.props("java.version").startsWith("1.") /* i.e. Java version >= 9 */)
```
jrudolph: Maybe extract that method now that it is used so much (also it's somewhat subtle).

project/Common.scala
```diff
@@ -0,0 +1,53 @@
...
+      "-target:jvm-1.8",
+      "-unchecked",
+      "-Xlint",
+      // "-Yno-adapted-args", //akka-http heavily depends on adapted args and => Unit implicits break otherwise
+      "-Ywarn-dead-code"
+      // "-Xfuture" // breaks => Unit implicits
+    ),
+    // '-release' parameter is restricted to 'Compile, compile' scope because
+    // otherwise `sbt akka-http-xml/compile:doc` fails with it on Scala 2.12.9
+    scalacOptions in (Compile, compile) ++= {
```
jrudolph: I guess the sharing between scalac and scaladoc options is a bit weird anyway.

Issue #2685, opened 2019-09-06
Content of response with 205 HTTP status code should be empty


sunghyunzz:
IMHO, response with 205 HTTP status code should have empty content. I checked following two documents:
1. http://en.wikipedia.org/wiki/List_of_HTTP_status_codes
1. https://tools.ietf.org/html/rfc7231#section-6.3.6

However current version of Akka HTTP sends contents while its HTTP status code is 205. 
```
HTTP/1.1 205 Reset Content
Content-Length: 80
Content-Type: text/plain; charset=UTF-8
Date: Fri, 06 Sep 2019 06:07:31 GMT
Server: akka-http/10.1.7

The server successfully processed the request, but is not returning any content.
```
It could make trouble when other HTTP client libraries do not expect contents to be sent. (for example, iOS app uses [Alamofire](https://github.com/Alamofire/Alamofire/blob/538edbfb891ab154f638a598e657b75c73d7e9ec/Source/ResponseSerialization.swift#L104) would be in a trouble)

jrudolph:
I agree, thanks for the PR.

jrudolph:
Fixed in #2686.

PR #2686, opened 2019-09-06
Disallow entity for 205 status code

sunghyunzz:
## Purpose

Disallow entity for 205 status code as response with this status code [should contain empty content](https://tools.ietf.org/html/rfc7231#section-6.3.6).

## References

References #2685

## Changes

- Now you cannot pass HTTP entity to response with 205 status code

## Background Context

Set `allowsEntity` as `false` would be the simplest way of following RFC spec.

### DIFF
--- a/docs/src/main/paradox/release-notes/10.1.x.md
+++ b/docs/src/main/paradox/release-notes/10.1.x.md
@@ -1,5 +1,13 @@
 # 10.1.x Release Notes
 
+## 10.1.10
+
+### Changes since 10.1.9
+
+##### akka-http
+
+* Fix 205 HTTP status not to contain any HTTP entity [#2686](https://github.com/akka/akka-http/pull/2686)
+
 ## 10.1.9
 
 10.1.9 is the tenth release in the 10.1.x series of Akka HTTP. See the [announcement](https://akka.io/blog/news/2019/07/17/akka-http-10.1.9-released.html) and the full list of changes below.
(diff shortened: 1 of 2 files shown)


sunghyunzz:
@raboof Thanks for the review. To mention breaking changes, should I create `10.2.x` release notes or create `10.1.10` section in existing `10.1.x` release notes?

jrudolph:
Add it to existing 10.1.x release notes. It's a breaking change but since it fixes a (currently) uncommonly used status code, we think it would be fine to be included in 10.1.x.

raboof:
OK TO TEST

sunghyunzz:
@raboof @jrudolph I mentioned about the change on release note, but I'm not sure if it is written correctly since english is not my first language 😅please suggest me any changes. Thanks :)

docs/src/main/paradox/release-notes/10.1.x.md
```diff
@@ -1,5 +1,13 @@
 # 10.1.x Release Notes
 
+## 10.1.10
+
```
raboof:
```suggestion

### Migration notes

[RFC 7231](https://tools.ietf.org/html/rfc7231#section-6.3.6) dictates an HTTP response with status code 205 ('Reset Content') is not allowed to have an entity body. Since [#2686](https://github.com/akka/akka-http/pull/2686) we enforce this restriction, so if you (incorrectly) produced such responses you will have to either remove the entity body or select a different status code.

```

sunghyunzz: Thanks for the suggestion :)

PR #2687, opened 2019-09-06
Temporary fix for Typed doc sample code

patriknw:
Quick fix to fix the build against akka master.

Maybe we should backport some of the toClassic methods to release-2.5 branch since it's causing these kind of problems?


patriknw:
I have created ticket https://github.com/akka/akka/issues/27649

PR #2689, opened 2019-09-06
Add scala-steward config to exclude akka

raboof:
So that if we enable it it won't send PR's to upgrade Akka


jrudolph:
Maybe let's enable scala steward for a while, and see what it does?

PR #2701, opened 2019-09-11
http2: fix receiving HEADERS with more than one CONTINUATION frame

jrudolph:
I guess no one has noticed so far because it only shows starting from HEADER blocks with more than 2 * frame size.


akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -134,15 +134,17 @@ class Http2ServerSpec extends AkkaSpecWithMaterializer("""
           expectedResponseHeaderBlock = HPackSpecExamples.C63ThirdResponseWithHuffman
         )
       }
-      "GET request in one HEADERS and one CONTINUATION frame" in new TestSetup with RequestResponseProbes {
+      "GET request in one HEADERS and two CONTINUATION frames" in new TestSetup with RequestResponseProbes {
```
johanandren: Wasn't the old test case also useful to keep?

PR #2707, opened 2019-09-12
build: collapse PR validation report for less clutter

jrudolph:
I just turned on the PR validation reports I developed a long time ago (which needed a fix in the jenkins github pr validator plugin which was released only a long time afterwards). To prevent too much clutter this PR collapses the report by default.

### DIFF
--- a/project/ValidatePullRequest.scala
+++ b/project/ValidatePullRequest.scala
@@ -366,7 +366,7 @@ object AggregatePRValidation extends AutoPlugin {
       log.info("")
       if (failed.nonEmpty || mimaFailures.nonEmpty || failedTasks.nonEmpty) {
         write("")
-        write("<details>\n<summary>Pull request validation report</summary>")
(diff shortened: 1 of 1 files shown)


project/ValidatePullRequest.scala
```diff
@@ -363,98 +364,102 @@ object AggregatePRValidation extends AutoPlugin {
       LogExchange.bindLoggerAppenders("testLogger", appender -> Level.Info :: Nil)
 
       log.info("")
-      write("")
-      write("# Pull request validation report")
-      write("")
+      if (failed.nonEmpty || mimaFailures.nonEmpty || failedTasks.nonEmpty) {
```
raboof: Perhaps we should write an empty file otherwise?

jrudolph:
Applied your suggestion, @raboof. Now the individual sections are
collapsed. An empty file was written already before by just opening and
closing the `FileWriter`.

PR #2708, opened 2019-09-12
[WIP] Auto-create mima exclusions

jrudolph:
This adds a new task `mimaCreateExclusionTemplate` which is called by the `pr-formatter-akka-http`.

(Removed a few previous exclusion to show it in action)


akka-http-core/src/main/mima-filters/10.1.9.backwards.excludes/pr-2708-auto-generated.excludes
```diff
@@ -0,0 +1,21 @@
+# Autogenerated MiMa filters, please pull branch and check validity, and add comments why they are needed etc.
+# Don't merge like this.
+
+# Incompatibilities against 10.1.6, 10.1.7, 10.1.8, 10.1.9
```
raboof: I really like this. I wonder if it wouldn't make more sense to have it in the validation report rather than appended to the PR, though.

jrudolph:
Closing for now. I created an upstream version of this at https://github.com/lightbend/mima/pull/432 but there are valid doubts whether the functionality makes sense or whether we should improve mima instead to report less false positives (i.e. when classes are marked `@Internal`).

PR #2737, opened 2019-10-01
core: hide `HttpEntity.Strict` body data by default

pfcoperez:
## Purpose

Aimed to avoid PII and SPI leakage, partially tackles https://github.com/akka/akka-http/issues/2409

## References

References https://github.com/akka/akka-http/pull/2560
References https://github.com/akka/akka-http/pull/2412
References https://github.com/akka/akka-http/issues/2409#issuecomment-496417344
References https://github.com/akka/akka-http/pull/2412#pullrequestreview-207325113

## Changes

This PR makes `HttpEntity.Strict#toString` method to print just the entity content type and its contents size. Replacing the previous behaviour which included the whole contents (or their beginning if they were excessively long).

## Background Context

https://github.com/akka/akka-http/pull/2560#pullrequestreview-285451541

> avoid including body data from strict entities (not a good idea anyways)

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
@@ -356,26 +356,9 @@ object HttpEntity {
     override def productPrefix = "HttpEntity.Strict"
 
     override def toString = {
-      val dataAsString = contentType match {
-        case _: Binary =>
-          data.toString()
-        case _: WithMissingCharset =>
-          data.toString()
-        case nb: NonBinary =>
-          try {
-            val maxBytes = 4096
-            if (data.length > maxBytes) {
-              val truncatedString = data.take(maxBytes).decodeString(nb.charset.value).dropRight(1)
-              s"$truncatedString ... (${data.length} bytes total)"
-            } else
-              data.decodeString(nb.charset.value)
-          } catch {
-            case NonFatal(e) =>
-              data.toString()
-          }
-      }
+      val dataSizeStr = s"${data.length} bytes total"
 
-      s"$productPrefix($contentType,$dataAsString)"
+      s"$productPrefix($contentType,$dataSizeStr)"
     }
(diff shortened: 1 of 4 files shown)


raboof:
OK TO TEST

lloydmeta:
@raboof it looks like you approved this ~ a month ago. Is there anything else needed to get this merged and hopefully released? Having this would make it much easier to use AkkaHttp in secrets-sensitive environments that also happen to need auditing 🙂

johanandren:
This has the potential to break existing applications, since calling toString currently guarantees the entity stream is consumed but with this PR it changes that. @jrudolph wdyt?

raboof:
> calling toString currently guarantees the entity stream is consumed

This PR is only about `Strict`, so that doesn't seem like a concern here I think? Or am I just not seeing it yet?

johanandren:
Ah, I missed that it is only about `Strict`, my bad!

pfcoperez:
As this approach was suggested in https://github.com/akka/akka-http/pull/2560#pullrequestreview-285451541 by @jrudolph , I wonder what might be holding this part back from being merged.

May I add something else to it?

Thanks!

luksow:
Hi @jrudolph @pfcoperez , FYI this change affected `logResult` (and thus dependant directives) - entity (when strict) is no longer showing up in the logs which may break some production code (it did for me). Not sure what should be done about this (if anything).

raboof:
@luksow thanks for the feedback - indeed there can be legitimate use cases for logging the entity, and I see how this change could've caught you off-guard (even though we tried to list is somewhat prominently in the release notes, https://akka.io/blog/news/2019/12/02/akka-http-10.1.11-released). That said, I do think this is the most reasonable default behavior here. It might be interesting to add some documentation and perhaps even API's to make it easier to log the full entity if you really want to.

luksow:
@raboof thanks for your reply! Sure, I agree with the decision, just wanted to let you know that it might be that this change affected some users.

Issue #1869, opened 2018-02-19
Parsing trailing headers in HTTP/2 


raboof:
When we create a HTTP/2 stream from an `HttpEntity.Chunked` we correctly convert any trailing headers from the `Source[HttpEntity.ChunkStreamPart]` into HTTP/2 trailing headers (#1843), but when accepting a HTTP/2 stream we ignore any headers received after we started receiving data.

We should decide what the API should look like to retrieve these headers, and make sure they are correctly passed along.

PR #2748, opened 2019-10-10
core: support chunked trailing headers in transformDataBytes

pasikon:
## Purpose

To support trailing headers in `akka.http.scaladsl.model.HttpEntity.Chunked#transformDataBytes` to improve compatibility with akka-grpc which is setting trailing header (gRPC status) in `akka.http.scaladsl.model.HttpEntity.Chunked`.

When data of such an entity is about to be processed with `akka.http.scaladsl.model.HttpEntity.Chunked#transformDataBytes`  we get `IllegalArgumentException("Chunked.transformDataBytes not allowed for chunks with metadata") `

## References

References #2218 #45 #1869

## Changes

If the subject of data transormation - `akka.http.scaladsl.model.HttpEntity.chunks` is containing the chunk with trailing headers concatenate it to the chunks which are the result of data transformation with `akka.http.scaladsl.model.HttpEntity.Chunked#transformDataBytes`

## Background Context

I'm instrumenting akka-grpc calls with kamon-akka which performing such http entity transformations, they are not supported at the moment, I would like to fix that

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
@@ -510,20 +510,20 @@ object HttpEntity {
       val transformChunks = GraphDSL.create() { implicit builder: GraphDSL.Builder[NotUsed] =>
         import akka.stream.scaladsl.GraphDSL.Implicits._
 
-        val partition = builder.add(Partition[ChunkStreamPart](2, {
+        val partition = builder.add(Partition[HttpEntity.ChunkStreamPart](2, {
           case c: Chunk     => 0
           case c: LastChunk => 1
         }))
-        val concat = builder.add(Concat[ChunkStreamPart](2))
+        val concat = builder.add(Concat[HttpEntity.ChunkStreamPart](2))
 
-        val chunkTransformer: Flow[ChunkStreamPart, ChunkStreamPart, Any] =
-          Flow[ChunkStreamPart]
+        val chunkTransformer: Flow[HttpEntity.ChunkStreamPart, HttpEntity.ChunkStreamPart, Any] =
+          Flow[HttpEntity.ChunkStreamPart]
             .map(_.data)
             .via(transformer)
             .map(b => Chunk(b))
 
-        val trailerBypass: Flow[ChunkStreamPart, ChunkStreamPart, Any] =
-          Flow[ChunkStreamPart]
+        val trailerBypass: Flow[HttpEntity.ChunkStreamPart, HttpEntity.ChunkStreamPart, Any] =
+          Flow[HttpEntity.ChunkStreamPart]
             // make sure to filter out any errors here, otherwise they don't go through the user transformer
             .recover { case NonFatal(ex) => Chunk(ByteString(0), "") }
             .collect { case lc @ LastChunk(_, s) if s.nonEmpty => lc }
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
@@ -507,43 +507,33 @@ object HttpEntity {
       withSizeLimit(SizeLimit.Disabled)
 
     override def transformDataBytes(transformer: Flow[ByteString, ByteString, Any]): HttpEntity.Chunked = {
-      import akka.stream.scaladsl.GraphDSL.Implicits._
+      val transformChunks = GraphDSL.create() { implicit builder: GraphDSL.Builder[NotUsed] =>
+        import akka.stream.scaladsl.GraphDSL.Implicits._
 
-      val buf = scala.collection.mutable.ArrayBuffer.empty[Throwable]
-
-      val transformChunks: Flow[ChunkStreamPart, ChunkStreamPart, NotUsed] = Flow.fromGraph(GraphDSL.create() { implicit builder: GraphDSL.Builder[NotUsed] =>
-
-        val bcast = builder.add(Broadcast[ChunkStreamPart](2))
+        val partition = builder.add(Partition[ChunkStreamPart](2, {
+          case c: Chunk     => 0
+          case c: LastChunk => 1
+        }))
         val concat = builder.add(Concat[ChunkStreamPart](2))
 
-        val newDataF: Flow[ChunkStreamPart, ByteString, Any] = Flow[ChunkStreamPart].map {
-          case Chunk(data, "")  => data
-          case LastChunk("", _) => ByteString.empty
-        }
-          .via(transformer).recover {
-            case t: Throwable =>
-              buf += t
-              ByteString.empty
-          }
+        val chunkTransformer: Flow[ChunkStreamPart, ChunkStreamPart, Any] =
+          Flow[ChunkStreamPart]
+            .map(_.data)
+            .via(transformer)
+            .map(b => Chunk(b))
 
-        val toChunksF: Flow[ByteString, ChunkStreamPart, Any] = Flow[ByteString].collect {
-          case b: ByteString if b.nonEmpty => Chunk(b)
-        }
+        val trailerBypass: Flow[ChunkStreamPart, ChunkStreamPart, Any] =
+          Flow[ChunkStreamPart]
+            // make sure to filter out any errors here, otherwise they don't go through the user transformer
+            .recover { case NonFatal(ex) => Chunk(ByteString(0), "") }
+            .collect { case lc @ LastChunk(_, s) if s.nonEmpty => lc }
 
-        val trailHeaderF: Flow[ChunkStreamPart, ChunkStreamPart, Any] = Flow[ChunkStreamPart].collect {
-          case lc @ LastChunk(_, s) if s.nonEmpty => lc
-        }.recover {
-          case _: Throwable =>
-            // just needs to complete
-            LastChunk
-        }
+        partition ~> chunkTransformer ~> concat
+        partition ~> trailerBypass ~> concat
+        FlowShape(partition.in, concat.out)
+      }
 
-        bcast ~> newDataF ~> toChunksF ~> concat
-        bcast ~> trailHeaderF ~> concat
-        FlowShape(bcast.in, concat.out)
-      })
-
-      HttpEntity.Chunked(contentType, chunks.via(transformChunks).map(c => if (buf.nonEmpty) throw buf.head else c))
+      HttpEntity.Chunked(contentType, chunks.via(transformChunks))
     }
 
     def withContentType(contentType: ContentType): HttpEntity.Chunked =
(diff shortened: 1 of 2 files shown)


lightbend-cla-validator:
Hi @pasikon,

pasikon:
Can somebody please look at it, I need it in my project and I wonder if this is the correct way of fixing the issue... :smirk_cat:

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
```diff
@@ -507,15 +507,43 @@ object HttpEntity {
...
-          case Chunk(data, "")    => data
-          case LastChunk("", Nil) => ByteString.empty
-          case _ =>
-            throw new IllegalArgumentException("Chunked.transformDataBytes not allowed for chunks with metadata")
-        } via transformer
-
-      HttpEntity.Chunked.fromData(contentType, newData)
+      import akka.stream.scaladsl.GraphDSL.Implicits._
+
+      val buf = scala.collection.mutable.ArrayBuffer.empty[Throwable]
```
jrudolph: Could you explain why this is needed? The problem with introducing mutable state here, is that an otherwise reusable entity becomes a single use entity that will fail in weird ways. So, any solution cannot depend on mutable state like this.

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
```diff
@@ -507,15 +507,43 @@ object HttpEntity {
...
+
+        val bcast = builder.add(Broadcast[ChunkStreamPart](2))
+        val concat = builder.add(Concat[ChunkStreamPart](2))
+
+        val newDataF: Flow[ChunkStreamPart, ByteString, Any] = Flow[ChunkStreamPart].map {
+          case Chunk(data, "")  => data
+          case LastChunk("", _) => ByteString.empty
+        }
+          .via(transformer).recover {
+            case t: Throwable =>
```
jrudolph: Not quite clear to me why errors should be discarded here?

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
```diff
@@ -507,15 +507,43 @@ object HttpEntity {
...
+          }
+
+        val toChunksF: Flow[ByteString, ChunkStreamPart, Any] = Flow[ByteString].collect {
+          case b: ByteString if b.nonEmpty => Chunk(b)
+        }
+
+        val trailHeaderF: Flow[ChunkStreamPart, ChunkStreamPart, Any] = Flow[ChunkStreamPart].collect {
+          case lc @ LastChunk(_, s) if s.nonEmpty => lc
+        }.recover {
+          case _: Throwable =>
```
jrudolph: Same here: Not quite clear to me why errors should be discarded here?

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
```diff
@@ -507,15 +507,43 @@ object HttpEntity {
...
+            // just needs to complete
+            LastChunk
+        }
+
+        bcast ~> newDataF ~> toChunksF ~> concat
+        bcast ~> trailHeaderF ~> concat
+        FlowShape(bcast.in, concat.out)
+      })
+
+      HttpEntity.Chunked(contentType, chunks.via(transformChunks).map(c => if (buf.nonEmpty) throw buf.head else c))
```
jrudolph: Ah, I see, errors should be thrown here. That's a somewhat uncommon approach that might not be thread-safe depending on how streams are executed.

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
```diff
@@ -507,15 +507,43 @@ object HttpEntity {
...
+          case lc @ LastChunk(_, s) if s.nonEmpty => lc
+        }.recover {
+          case _: Throwable =>
+            // just needs to complete
+            LastChunk
+        }
+
+        bcast ~> newDataF ~> toChunksF ~> concat
+        bcast ~> trailHeaderF ~> concat
+        FlowShape(bcast.in, concat.out)
```
jrudolph: That's an interesting solution but I wonder if it really works. It seems prone to deadlocks because the second part of the stream shouldn't generate any demand until the first part is complete.I wonder why it works in the tests.

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
pasikon: It will hold at most one throwable, using that is the leftover of some experimenting, will remove that

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
pasikon: I haven't thought about dedalocks, just treated that as simple bcast concat flow, can you please explain?

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
pasikon: Well LastChunk doesn't carry any data so its not the subject of processing with the `transformer` Flow. 

`bcast ~> trailHeaderF ~> concat` will fail only when theres bad encoding (for example: akka/http/scaladsl/server/directives/CodingDirectivesSpec.scala:117) but we don't care because in such case `bcast ~> newDataF ~> toChunksF ~> concat` will fail as well with same exception.

What we care about is the potential failure of `.via(transformer).` and that is just being written to the buffer.

Maybe you can elaborate more on the thread safety issue because I don't see it (yet :smile:  )

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
jrudolph: It seems to work indeed in this case because `concat` adds a buffer in front of the stage that makes sure there is demand for the second branch even if a potential element will not be consumed until the first branch is done. This works here because the second branch will generate a single element at most.

jrudolph:
Here's an alternative slightly simpler version:

```scala
    override def transformDataBytes(transformer: Flow[ByteString, ByteString, Any]): HttpEntity.Chunked = {
      val transformChunks = GraphDSL.create() { implicit builder: GraphDSL.Builder[NotUsed] =>
        import akka.stream.scaladsl.GraphDSL.Implicits._

        val bcast = builder.add(Broadcast[ChunkStreamPart](2))
        val concat = builder.add(Concat[ChunkStreamPart](2))

        val chunkTransformer: Flow[ChunkStreamPart, ChunkStreamPart, Any] =
          Flow[ChunkStreamPart]
            .collect {
              case Chunk(b, "") => b
            }
            .via(transformer)
            .map(b => Chunk(b))

        val trailerBypass: Flow[ChunkStreamPart, ChunkStreamPart, Any] =
          Flow[ChunkStreamPart]
            // make sure to filter out any errors here, otherwise they don't go through the user transformer
            .recover { case NonFatal(ex) => Chunk(ByteString(0), "") }
            .collect { case lc @ LastChunk(_, s) if s.nonEmpty => lc }

        bcast ~> chunkTransformer ~> concat
        bcast ~> trailerBypass ~> concat
        FlowShape(bcast.in, concat.out)
      }

      HttpEntity.Chunked(contentType, chunks.via(transformChunks))
    }
```

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
jrudolph: I see what the problem was. Without that solution, potential errors might skip the user-supplied transformer because they race through the second branch of the bcast/concat. I've posted a simple solution in the alternative snippet.

jrudolph:
I agree it would be good to support trailers in the light of grpc using trailing headers. On the other hand, this change might be somewhat expensive because the processing path for each chunk is now somewhat more heavy with control flow. On the other hand, you can always amortize the processing costs by providing bigger chunks.

Here's another iteration that uses `Partition` instead of `Broadcast` to prevent the second branch from being active until the last element:

```scala
override def transformDataBytes(transformer: Flow[ByteString, ByteString, Any]): HttpEntity.Chunked = {
      val transformChunks = GraphDSL.create() { implicit builder: GraphDSL.Builder[NotUsed] =>
        import akka.stream.scaladsl.GraphDSL.Implicits._

        val partition = builder.add(Partition[ChunkStreamPart](2, {
          case c: Chunk     => 0
          case c: LastChunk => 1
        }))
        val concat = builder.add(Concat[ChunkStreamPart](2))

        val chunkTransformer: Flow[ChunkStreamPart, ChunkStreamPart, Any] =
          Flow[ChunkStreamPart]
            .map(_.data)
            .via(transformer)
            .map(b => Chunk(b))

        val trailerBypass: Flow[ChunkStreamPart, ChunkStreamPart, Any] =
          Flow[ChunkStreamPart]
            // make sure to filter out any errors here, otherwise they don't go through the user transformer
            .recover { case NonFatal(ex) => Chunk(ByteString(0), "") }
            .collect { case lc @ LastChunk(_, s) if s.nonEmpty => lc }

        partition ~> chunkTransformer ~> concat
        partition ~> trailerBypass ~> concat
        FlowShape(partition.in, concat.out)
      }

      HttpEntity.Chunked(contentType, chunks.via(transformChunks))
    }
```

WDYT? I might find this acceptable. (And anyway, thanks for raising this issue and proposing a fix, @pasikon).

pasikon:
@jrudolph yeah I will rely a lot on such byte processing, difficult to judge perf impact now but widening the grpc support is critical to my project, thanks for improving my proposal, its great to gain some experience with Akka Streams! 
I have committed your improvement also the issue is that its not compiling with Scala 2.11, maybe you can help with that?

```
[error] /home/michal/dev/sources/akka-http/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala:536:50: type mismatch;
[error]  found   : akka.stream.Graph[akka.stream.FlowShape[Chunked.this.ChunkStreamPart,Chunked.this.ChunkStreamPart],akka.NotUsed]
[error]     (which expands to)  akka.stream.Graph[akka.stream.FlowShape[akka.http.javadsl.model.HttpEntity.ChunkStreamPart,akka.http.javadsl.model.HttpEntity.ChunkStreamPart],akka.NotUsed]
[error]  required: akka.stream.Graph[akka.stream.FlowShape[akka.http.scaladsl.model.HttpEntity.ChunkStreamPart,akka.http.scaladsl.model.HttpEntity.ChunkStreamPart],?]
[error]       HttpEntity.Chunked(contentType, chunks.via(transformChunks))

```

jrudolph:
>  I have committed your improvement also the issue is that its not compiling with Scala 2.11, maybe you can help with that?

Try replacing `ChunkStreamPart` with `HttpEntity.ChunkStreamPart`. Otherwise, Scala 2.11 seems to confuse it with the nested interface in `javadsl.model.Chunked.ChunkStreamPart`.

jrudolph:
Thanks a lot, @pasikon.

PR #2754, opened 2019-10-14
Add dependency override when overriding akka dependency to source dependency

### LINKED ISSUE #2747: sslconfig version 0.4.0 in Akka master
Tests are failing when running with latest Akka master because sslconfig was updated to 0.4.0.

```
Uncaught error from thread [ClientServerSpec-akka.actor.default-dispatcher-11]: com/typesafe/sslconfig/ssl/NoopHostnameVerifier, shutting down JVM since 'akka.jvm-exit-on-fatal-error' is enabled for ActorSystem[ClientServerSpec]
java.lang.NoClassDefFoundError: com/typesafe/sslconfig/ssl/NoopHostnameVerifier
	at com.typesafe.sslconfig.akka.AkkaSSLConfig.useJvmHostnameVerification(AkkaSSLConfig.scala:78)
	at akka.stream.scaladsl.TLS$.$anonfun$apply$2(TLS.scala:110)
	at akka.stream.impl.io.TLSActor.liftedTree1$1(TLSActor.scala:159)
	at akka.stream.impl.io.TLSActor.<init>(TLSActor.scala:159)
```

https://jenkins.akka.io:8498/job/akka-http-nightly/647/AKKA_VERSION=default,Node=akka-http-nightly-node,SCALA_VERSION=2.12.9,jdk=AdoptOpenJDK%208/consoleFull

I can't find that Akka HTTP is defining an explicit dependency to ssl-config so why doesn't it run with the transitive dependency from akka-stream?

chbatey:
Without this the released version overrides the source version so the
test runs with the released version of Akka but typically with the
newer transative dependencies.

Fixes #2747

### DIFF
--- a/project/AkkaDependency.scala
+++ b/project/AkkaDependency.scala
@@ -28,7 +28,7 @@ object AkkaDependency {
       akkaVersion match {
         case "master" => "git://github.com/akka/akka.git#master"
         case "release-2.5" => "git://github.com/akka/akka.git#release-2.5"
-        case x => ""
+        case _ => ""
       }
 
     System.getProperty("akka.sources", fallback)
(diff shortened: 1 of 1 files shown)


jrudolph:
I wonder how it ever did the right thing before? Is it only about the transitive dependencies being wrong?

jrudolph:
Fixed by #2761 instead.

PR #2758, opened 2019-10-16
build: make sure to lift all akka dependencies to source when test-against-akka-master is used

### LINKED ISSUE #2747: sslconfig version 0.4.0 in Akka master

jrudolph:
Fixes #2747. Might obsolete #2754.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -149,7 +149,7 @@ lazy val httpCore = project("akka-http-core")
   .addAkkaModuleDependency("akka-stream", "provided")
   .addAkkaModuleDependency("akka-stream-testkit", "test")
   .addAkkaModuleDependency(
-    "akka-stream",
+    "akka-stream-testkit",
     "test",
     shouldUseSourceDependency = true,
     uri("git://github.com/akka/akka.git#master"),


jrudolph:
Going ahead with merging to check if it fixes the nightlies.

build.sbt
```diff
@@ -149,7 +149,7 @@ lazy val httpCore = project("akka-http-core")
   .addAkkaModuleDependency("akka-stream", "provided")
   .addAkkaModuleDependency("akka-stream-testkit", "test")
   .addAkkaModuleDependency(
-    "akka-stream",
+    "akka-stream-testkit",
```
patriknw: ok, but should the above dependency to `akka-stream-testkit` still be there?

jrudolph: Hmm good point, maybe it needs to be guarded with an opposite condition than the one below.

jrudolph:
Anyway on Jenkins, the problem is still the same. I wonder why it worked on my machine.

build.sbt
jrudolph: Seems it is ok like that because the other line will be evicted once https://github.com/akka/akka-http/pull/2761 is merged.

PR #2761, opened 2019-10-16
build: rename `akka.build.version` env var not to clash with the one in the akka build

jrudolph:
Follow up to #2758, obsoletes #2754.


jrudolph:
Merging right away to test on jenkins.

PR #2776, opened 2019-10-21
Potential fix for idle timeouts in http2

viktorklang:
## Purpose

Fixes idle timeouts for HTTP/2 connections by removing the TCP idle timeout just as for HTTP/1 and instead attaching the timeouts at the HTTP layer.

## References

References [akka-grpc/664](https://github.com/akka/akka-grpc/issues/664) where it was discovered.

## Changes

Idle timeout management for HTTP/2

## Background Context

Because it is near-identical as the HTTP/1 solution

### DIFF
--- a/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
+++ b/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
@@ -7,6 +7,7 @@ package akka.http.impl.engine.http2
 import akka.NotUsed
 import akka.annotation.InternalApi
 import akka.event.LoggingAdapter
+import akka.http.impl.engine.HttpConnectionIdleTimeoutBidi
 import akka.http.impl.engine.http2.FrameEvent._
 import akka.http.impl.engine.http2.framing.{ Http2FrameParsing, Http2FrameRendering }
 import akka.http.impl.engine.http2.hpack.{ HeaderCompression, HeaderDecompression }
@@ -19,6 +20,8 @@ import akka.http.scaladsl.settings.{ Http2ServerSettings, ParserSettings, Server
 import akka.stream.TLSProtocol._
 import akka.stream.scaladsl.{ BidiFlow, Flow, Source }
 import akka.util.ByteString
+
+import scala.concurrent.duration.{ FiniteDuration }
 
 import scala.concurrent.{ ExecutionContext, Future }
 import scala.collection.immutable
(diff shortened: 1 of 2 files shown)


jrudolph:
If you turn on `akka.http.server.http2.log-frames = true` then you can see the difference before and after the fix:

Before:

```
[info] [ERROR] [10/21/2019 15:27:43.750] [ServerTest-akka.actor.default-dispatcher-4] [akka.stream.Log(akka://ServerTest/system/StreamSupervisor-0)] [ UP ] Upstream failed. (akka.stream.scaladsl.TcpIdleTimeoutException: TCP idle-timeout encountered on connection to [/127.0.0.1:48696], no bytes passed in the last 5 seconds)
[info] [ERROR] [10/21/2019 15:27:43.751] [ServerTest-akka.actor.default-dispatcher-4] [akka.stream.Log(akka://ServerTest/system/StreamSupervisor-0)] [DOWN] Upstream failed. (akka.stream.scaladsl.TcpIdleTimeoutException: TCP idle-timeout encountered on connection to [/127.0.0.1:48696], no bytes passed in the last 5 seconds)
```

After:

```
system/StreamSupervisor-0)] [DOWN] Downstream finished.
[info] [ERROR] [10/21/2019 15:29:16.518] [ServerTest-akka.actor.default-dispatcher-6] [akka.stream.Log(akka://ServerTest/system/StreamSupervisor-0)] [ UP ] Upstream failed. (akka.http.impl.engine.HttpIdleTimeoutException: HTTP idle-timeout encountered, no bytes passed in the last 5 seconds. This is configurable by akka.http.[server|client].idle-timeout.)
```

The idle timeout sends both a failure through `UP` and cancellation through `DOWN`. In the first case, the failure is delivered first, in the second case, the cancellation is delivered first. This is a race condition that might also play out differently under different circumstances.

The bug here is that an upstream (= connection) failure does not shut down outgoing user streams correctly.

viktorklang:
Great discovery, @jrudolph!

Do you have an idea on how to address the issue?

jrudolph:
> Do you have an idea on how to address the issue?

Yep, should be quite simple. We just need to give the outgoing side of the multiplexer a chance to clean up when the stage is closed down.

viktorklang:
@jrudolph Awesome, ping me in the PR—I'd love to have a look! :)

viktorklang:
@jrudolph Did you fix this?

jrudolph:
I fixed a related thing but #2856 seems to suggest we might want to have this fix as well. I'll have a look soon.

viktorklang:
Thanks @jrudolph, I want to be able to ugrade Cloudstate to run a non-daily dep for akka-http.

jrudolph:
> Thanks @jrudolph, I want to be able to ugrade Cloudstate to run a non-daily dep for akka-http.

It should already work with the other fix, shouldn't it? That was released in 10.1.11.

PR #2779, opened 2019-10-22
h2: make sure all substreams are closed when connection is torn…

### LINKED ISSUE #2777: Http2: outgoing request streams not closed when upstream finished
In the Http2 stream multiplexer only `onDownstreamFinish` from the connection is closing down all the `outStreams` but not `onUpstreamFinish` / `onUpstreamFailure`.

That means that for any connection error or completion or the idle-timeout might leave user streams dangling. It depends on how the race between finishing and cancellation plays out. If cancellation is first, then everything is fine, otherwise, the outstreams are left dangling.

This is the root cause of https://github.com/akka/akka-grpc/issues/664.

jrudolph:
Fixes #2777.


viktorklang:
LGTM, @jrudolph! Thank you!

Issue #2781, opened 2019-10-23
Try out akka-http with Akka 2.6 and add documentation and improvements where needed


jrudolph:
We already have this section in the documentation: https://doc.akka.io/docs/akka-http/current/routing-dsl/index.html#interaction-with-akka-typed. 

To review: 

 * How to use Akka with the typed actor system? Is the information still valid?
 * Enable the compilation of the docs example using latest 2.5 toClassic support
 * Look into simplifying interaction by adding helper to `akka.http.scaladsl.Http`
 * Link to the section from the Akka main docs?

johanandren:
I guess we still don't want to require 2.5.26, as it is only needed to get the same API for 2.6 and 2.6 used in one place in the docs?

jrudolph:
We need it for the cancellation strategies anyway so we will update.

PR #2784, opened 2019-10-23
docs: typed sample polish

johanandren:
Hold this PR until the quickstart update is done. I think we may want to link to that instead, or align this sample with that when it is done for 2.6

Refs #2781

 * mostly works with both 2.5.26 and 2.6.0-RC (requires a few small changes)
 * updated to follow styleguide closer
 * made it a runnable instead of a compile-only test
 * a bit closer to what a real(tm) server wrapping actor would have to do

There are a few source incompatibilities for typed making crossbuilding impossible.
Not sure if we should really try to address those or just live with that
something marked @MayChange are not 100% compatible. Not actually sure that we even can make it source compatible because of the implicits becoming ambigous.


johanandren:
I'll do the Java part as well, but would want this LGTM:ed first since that will be as close to 1:1 with this sample as possible.

johanandren:
Created #2790 to discuss some of the problems I found here

docs/src/test/scala-2.12+/docs/http/scaladsl/HttpServerWithTypedSample.scala
```diff
@@ -0,0 +1,228 @@
...
+            rejectEmptyResponse {
+              complete(maybeJob)
+            }
+          }
+        )
+      }
+  }
+  //#akka-typed-route
+
+  /* there are still differences making this impossible to compile both for 2.5 and 2.6 at the same time
```
jrudolph: So, for now, this would compile correctly in the regular build, but would fail building against 2.6 master.

jrudolph:
@johanandren I guess we can merge that now? It won't be published until 10.1.11 anyway.

johanandren:
Yeah I think so, not perfect but better than before!

Issue #2786, opened 2019-10-24
EmptyHeader terminating the stream with "Entity stream truncation"


supercharger:
When used for HTTP transfer chunked encoding (server is Akka as well), We are seeing following exception intermittently which is very difficult to reproduce it.  It looks like this exception is originating at this  [HttpMessageParser.scala#L143 ](https://github.com/akka/akka-http/blob/master/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpMessageParser.scala#L143).  What does EmptyHeader mean, Is there a reason to terminate the stream/throw exception instead of ignoring it.

One thing that needs to be fixed for certain is that this case **doesn't have any logging to indicate what went wrong** before throwing the exception, so it needs to be added. We've enabled Debug logging, but the logging doesn't indicate what went wrong.

```
Caused by: 
akka.http.scaladsl.model.EntityStreamException: Entity stream truncation
        at akka.http.impl.engine.parsing.HttpMessageParser$$anonfun$2.applyOrElse(HttpMessageParser.scala:348)
        at akka.http.impl.engine.parsing.HttpMessageParser$$anonfun$2.applyOrElse(HttpMessageParser.scala:346)
        at akka.stream.impl.fusing.Collect$$anon$6.$anonfun$wrappedPf$1(Ops.scala:227)
        at akka.stream.impl.fusing.SupervisedGraphStageLogic.withSupervision(Ops.scala:186)
        at akka.stream.impl.fusing.Collect$$anon$6.onPush(Ops.scala:229)
        at akka.stream.impl.fusing.GraphInterpreter.processPush(GraphInterpreter.scala:523)
        at akka.stream.impl.fusing.GraphInterpreter.processEvent(GraphInterpreter.scala:480)
```
Thanks for your time in looking in to this, appreciate it. I can raise a PR if more details on EmptyHeader can be provided.

Debug logs ( which doesn't help ) :
```
DEBUG] [10/20/2019 16:30:20.425] [main] [AkkaSSLConfig(akka://ImportUserPreferences)] Initializing AkkaSSLConfig extension...
[DEBUG] [10/20/2019 16:30:20.428] [main] [AkkaSSLConfig(akka://ImportUserPreferences)] buildHostnameVerifier: created hostname verifier: com.typesafe.sslconfig.ssl.NoopHostnameVerifier@4e3
01795
[...]

PR #2789, opened 2019-10-28
core: improve "Entity stream truncation" error message

jrudolph:
Refs #2786


jrudolph:
Fixed test assertions.

PR #2780, opened 2019-10-23
build: update to sbt 1.3.3, mima to 0.6.1, sbt-dependency-graph…

jrudolph:
Coursier needs to be turned off for sbt-dependency-graph to work correctly
until https://github.com/sbt/sbt/issues/4706 is fixed.

Issue #2788, opened 2019-10-28
Akka Websocket 4x slower when updating Scala 2.12 to 2.13


galbarm:
Hello,
Akka version is 2.5.26 and Play is 2.7.3.
Following is a sample server that demonstrates the issue:

```
object Test extends App{

  val app = GuiceApplicationBuilder(configuration = Configuration.apply("play.server.websocket.frame.maxLength" -> 11 * 1024 * 1024)).build

  val routes: PartialFunction[RequestHeader, Handler] = {
    case GET(p"/test_pressure") => WebSocket.accept[play.api.http.websocket.Message, play.api.http.websocket.Message] { request => {
      Flow.fromSinkAndSource(Sink.foreach(in => {
        println("got message")
      }), Source.maybe)
    }
    }
  }

  val config = ServerConfig(
    rootDir = new File(""),
    port = Some(0),
    sslPort = None,
    address = "127.0.0.1",
    mode = Mode.Test,
    properties = System.getProperties,
    configuration = app.configuration
  )

  val server = AkkaHttpServer.fromRouterWithComponents(config)(_ => routes)

  println("port = " + server.httpPort.get)
}
```

A client connects to this server and sends 100 websocket messages, 10MB in size each (1GB total) as fast as possible.

I’m starting the server in 2 configurations:

scalaVersion := “2.12.10” in build.sbt
scalaVersion := “2.13.1” in build.sbt
This is the only difference between the two tests.

The issue is that in Scala 2.12.10 the test takes 6 seconds to complete, whereas in Scala 2.13.1 the same test takes 22 seconds.

galbarm:
Likely to be related or caused by:
https://github.com/akka/akka/issues/28114

galbarm:
As I mentioned in https://github.com/akka/akka/issues/28114
The issue only reproduces when using AkkaHttpServer via it's Play wrapper (play-akka-http-server).

When using the stand alone version `Http().bindAndHandle...` , or when using Netty as backend `NettyServer.fromRouterWithComponents` the problem does not reproduce.

jrudolph:
Thanks for the information, @galbarm. Play reimplements part of the WS stack so it might interact differently. If you have an example play project to share that shows the issue that would be great.

We identified (and are fixing) at least one issue in akka/akka#28116. I also improved WS masking performance in #2801. Once we have backported the fix to 2.5.x, would you be able to try with a nightly snapshot?

I've filed https://github.com/playframework/playframework/issues/9799.

[...]

PR #2801, opened 2019-11-05
Improve WS masking performance

jrudolph:
On top of #2780 

It now masks 4 bytes together and also avoids an extra buffer copy at the end by using `ByteString.fromArrayUnsafe`.

(It cannot fix the 2.13 performance problem because ByteString.toArray is still used)


jrudolph:
According to `https://jenkins.akka.io:8498/job/akka-http-bench-adhoc/139/jmh-run-report/` that's a 95% improvement.

jrudolph:
Refs #2788

Issue #2804, opened 2019-11-06
UnsupportedRequestContentTypeRejection formats error messages not in a user-friendly way


satorg:
If a request is rejected with `UnsupportedRequestContentTypeRejection` and actual `contentType` is provided to the rejection, then the error message for this rejection will look like the following:

> The request's Content-Type [Some(Some(image/gif))] is not supported. Expected:\nimage/jpeg or image/png

which is not user-friendly.

The reason is in these two lines:
https://github.com/akka/akka-http/blob/v10.1.10/akka-http/src/main/scala/akka/http/scaladsl/server/RejectionHandler.scala#L265-L266

The `unsupportedContentType` value has got type `Option[Option[ContentType]]`.

raboof:
That indeed could be better. Would you be interested in improving this?

satorg:
Yes, for sure. I have shot a tiny PR for this issue (#2807), could you take a look, please?

jrudolph:
Fixed in #2807.

PR #2807, opened 2019-11-08
http: fix default rejection message for unsupported content type (#2804)

satorg:
# Pull Request Checklist

* [ ] Have you updated the documentation?
* [x] Have you added tests for any changed functionality?

## Purpose

This PR fixes a message which the default rejection handler produces for `UnsupportedRequestContentTypeRejection`. With this PR the message looks something like
```
The request's Content-Type [image/bmp] is not supported. Expected:
image/jpeg or image/png
```
if the corresponding rejection has the `contentType` option defined or
```
The request's Content-Type is not supported. Expected:
image/jpeg or image/png
```
if the `contentType` option is set to `None`.
In case if `UnsupportedRequestContentTypeRejection` doesn't provide supported content type ranges,
the rejection message will look like:
```
The request's Content-Type [image/bmp] is not supported.
```

## References

References #2804

## Changes

* The default `RejectionHandler`'s error message for `UnsupportedRequestContentTypeRejection` is changed.

### DIFF
--- a/akka-http/src/main/scala/akka/http/scaladsl/server/RejectionHandler.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/server/RejectionHandler.scala
@@ -261,9 +261,13 @@ object RejectionHandler {
           supported.map(_.value).mkString("\n")))
       }
       .handleAll[UnsupportedRequestContentTypeRejection] { rejections =>
+        val unsupported = rejections.find(_.contentType.isDefined).flatMap(_.contentType).fold("")(" [" + _ + "]")
         val supported = rejections.flatMap(_.supported).mkString(" or ")
-        val unsupportedContentType = rejections.find(_.contentType.isDefined).map(_.contentType)
-        rejectRequestEntityAndComplete((UnsupportedMediaType, s"The request's Content-Type [$unsupportedContentType] is not supported. Expected:\n" + supported))
+        val expected =
+          if (supported.isEmpty) ""
+          else " Expected:\n" + supported
+
+        rejectRequestEntityAndComplete((UnsupportedMediaType, s"The request's Content-Type$unsupported is not supported.$expected"))
       }
       .handleAll[UnsupportedRequestEncodingRejection] { rejections =>
         val supported = rejections.map(_.supported.value).mkString(" or ")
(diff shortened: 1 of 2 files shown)


satorg:
@raboof, I could you suggest please, where can I update tests and docs for this change?
It seems the default handler for this rejection is not covered by any test...

johanandren:
OK TO TEST

johanandren:
I think the `akka.http.scaladsl.server.RejectionSpec` would be a reasonable place to put a test, can't see that any docs are needed.

akka-http/src/main/scala/akka/http/scaladsl/server/RejectionHandler.scala
```diff
@@ -262,8 +262,9 @@ object RejectionHandler {
       }
       .handleAll[UnsupportedRequestContentTypeRejection] { rejections =>
         val supported = rejections.flatMap(_.supported).mkString(" or ")
-        val unsupportedContentType = rejections.find(_.contentType.isDefined).map(_.contentType)
-        rejectRequestEntityAndComplete((UnsupportedMediaType, s"The request's Content-Type [$unsupportedContentType] is not supported. Expected:\n" + supported))
+        val unsupportedContentType =
+          rejections.find(_.contentType.isDefined).flatMap(_.contentType).fold("")(" [" + _ + "]")
```
raboof: Perhaps this could be written shorter, something like `rejections.flatMap(_.contentType).mkString(" [", ", ", "]")` perhaps?

johanandren: It's an `Option` so that wouldn't give empty string for `None` though

raboof: aah right I got confused

satorg:
Ok, got it, thanks. I'll add unit tests shortly.

satorg:
I added the unit tests, please take a look.
Also I slightly changed formatting to handle a case when the `UnsupportedRequestContentTypeRejection#supported` set is empty.
I described the change in the PR's head commit.

jrudolph:
Thanks, @satorg.

PR #2808, opened 2019-11-09
Update naming for status codes 413, 414, 416 and 505

Philippus:
For status codes 413, 414 and 416 older names from older RFCs are currently used, I updated them to follow the naming in RFC 7231 and RFC 7233. 
For status code 505 the Scala naming convention was not followed. 
I deprecated the older names.
I decided to leave `OK` (which also doesn't follow the naming convention) as is, because that probably would mean a lot of changes downstream.

### DIFF
--- a/akka-http-core/src/main/java/akka/http/javadsl/model/StatusCodes.java
+++ b/akka-http-core/src/main/java/akka/http/javadsl/model/StatusCodes.java
@@ -57,9 +57,27 @@ public final class StatusCodes {
     public static final StatusCode GONE = akka.http.scaladsl.model.StatusCodes.Gone();
     public static final StatusCode LENGTH_REQUIRED = akka.http.scaladsl.model.StatusCodes.LengthRequired();
     public static final StatusCode PRECONDITION_FAILED = akka.http.scaladsl.model.StatusCodes.PreconditionFailed();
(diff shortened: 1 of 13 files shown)


akka-http-core/src/main/java/akka/http/javadsl/model/StatusCodes.java
```diff
@@ -57,9 +57,27 @@ private StatusCodes() {}
     public static final StatusCode GONE = akka.http.scaladsl.model.StatusCodes.Gone();
     public static final StatusCode LENGTH_REQUIRED = akka.http.scaladsl.model.StatusCodes.LengthRequired();
     public static final StatusCode PRECONDITION_FAILED = akka.http.scaladsl.model.StatusCodes.PreconditionFailed();
+    public static final StatusCode PAYLOAD_TOO_LARGE = akka.http.scaladsl.model.StatusCodes.PayloadTooLarge();
+
+    /**
+     * @deprecated deprecated in favor of Payload Too Large
```
jrudolph: Can you put the exact new symbol name into all of the deprecation messages? I think that would help people following up on that change. So, here it would be `PAYLOAD_TOO_LARGE`:

```suggestion
     * @deprecated deprecated in favor of PAYLOAD_TOO_LARGE
```

PR #2811, opened 2019-11-11
Implement reverseProxy directive (#240)

### LINKED ISSUE #240: Add `reverseProxy` directive

nrktkt:
## Purpose

Simple reverse proxy directive that will send the current request to a specified destination and return the response back to the caller.

## References

Resolves #240

## Changes

* Added reverse proxy directive
* Added Via header

## Background Context

Would love to get this in. Seems like there's a big demand.

### DIFF
--- a/akka-http/src/main/scala/akka/http/scaladsl/server/directives/ReverseProxyDirectives.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/server/directives/ReverseProxyDirectives.scala
@@ -89,18 +89,20 @@ trait ReverseProxyDirectives {
       .orElse(intermediary.map(i => Via(List(i))))
 
     val outgoingHeaders = incomingHeaders.flatMap {
-      case _: `X-Real-Ip`           => updatedXRealIpHeaderOption
-      case _: `X-Forwarded-For`     => updatedXForwardedForHeaderOption
-      case _: Via                   => Nil // added back at the end
-      case _: `Timeout-Access`      => Nil
+      // region hop-by-hop headers https://tools.ietf.org/html/rfc2616#section-13.5.1
       case _: Connection            => Nil
       // keep alive header is not included in modeled headers
       case _: `Proxy-Authenticate`  => Nil
       case _: `Proxy-Authorization` => Nil
-      case _: `Transfer-Encoding`   => Nil
       case h if h.is("te")          => Nil
       case h if h.is("trailers")    => Nil
+      case _: `Transfer-Encoding`   => Nil
       case _: Upgrade               => Nil
+      // endregion
+      case _: `X-Real-Ip`           => updatedXRealIpHeaderOption
+      case _: `X-Forwarded-For`     => updatedXForwardedForHeaderOption
+      case _: `Timeout-Access`      => Nil
+      case _: Via                   => Nil // added back at the end
       case h                        => Some(h)
     } ++ maybeVia
 
--- a/akka-http/src/main/scala/akka/http/scaladsl/server/directives/ReverseProxyDirectives.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/server/directives/ReverseProxyDirectives.scala
@@ -14,7 +14,7 @@ import akka.http.scaladsl.model._
 import akka.http.scaladsl.model.headers._
 import akka.http.scaladsl.server.Directives._
 import akka.http.scaladsl.server.Route
-import akka.http.scaladsl.server.directives.ReverseProxyDirectives.{ReverseProxyTargetConfig, ReverseProxyTargetMagnet}
+import akka.http.scaladsl.server.directives.ReverseProxyDirectives.{ ReverseProxyTargetConfig, ReverseProxyTargetMagnet }
 import akka.util.ByteString
 
 trait ReverseProxyDirectives {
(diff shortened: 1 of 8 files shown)


kag0:
I _think_ we established in #2020 that the MIMA issue is acceptable.

jrudolph:
OK TO TEST

jrudolph:
Thanks for that contribution, @kag0. The last time the addition of a reverseProxy directive was discussed, it got stuck on the question how safe it would be to use it.

To merge this we will need some documentation page that will document the behavior and also discusses potential security issues.

The API should be be marked as `@ApiMayChange` for now so we are still flexible to change it if needed later.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/ReverseProxyDirectives.scala
```diff
@@ -0,0 +1,147 @@
...
+      }
+
+      config.viaId.map(ViaIntermediary(protocol, version, _))
+    }
+
+    val maybeVia = request.header[Via]
+      .map(via => via.copy(intermediaries = via.intermediaries ++ intermediary))
+      .orElse(intermediary.map(i => Via(List(i))))
+
+    val outgoingHeaders = incomingHeaders.flatMap {
```
jrudolph: How was this list compiled? Is there an RFC that gives some recommendations? Should the user be able to configure it? E.g. for the ones that are just filtered-out we could give the list in the configuration where we could also document why they are filtered out.

What about headers that might carry authentication information? Should they be forwarded automatically? Or should it be opt-in?

akka-http/src/main/scala/akka/http/scaladsl/server/directives/ReverseProxyDirectives.scala
```diff
@@ -0,0 +1,147 @@
...
+import akka.http.scaladsl.model._
+import akka.http.scaladsl.model.headers._
+import akka.http.scaladsl.server.Directives._
+import akka.http.scaladsl.server.Route
+import akka.http.scaladsl.server.directives.ReverseProxyDirectives.{ ReverseProxyTargetConfig, ReverseProxyTargetMagnet }
+import akka.util.ByteString
+
+trait ReverseProxyDirectives {
+
+  def reverseProxy(target: ReverseProxyTargetMagnet): Route =
```
jrudolph: By now, we'd rather not use magnets if it can be avoided. It seems here it is required to access the `system` to get the client. Maybe with the Akka 2.6 changes to use system more than materializer in many places, we could try to add a system to the `RequestContext` in the first place? That would be a follow up issue to this one.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/ReverseProxyDirectives.scala
kag0: Hop-by-hop headers are defined in [http 1.1](https://tools.ietf.org/html/rfc2616#section-13.5.1). These don't seem to be something the user would configure. 
[7230](https://tools.ietf.org/html/rfc7230#section-5.7) talks about the via header. To me that seems more for public and/or forward proxies rather than reverse proxies. So it is something the user wants/needs to configure.
As far as a literal list, https://www.mnot.net/blog/2011/07/11/what_proxies_must_do was educational.

I think we should establish that aside from headers that are specifically relevant to proxying, we should establish the default as "the request is forwarded as-is" (which is something that should go in that documentation page you mentioned :)). This way the user doesn't form an expectation that any part of the request will be treated specially, and that they need to modify it in an outer directive if they want to change what is forwarded. For authentication specifically this prevents us from having to define what qualifies as a "default" authentication header/cookie.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/ReverseProxyDirectives.scala
kag0: Are you suggesting we keep the magnet for now and remove it in a follow up issue, 
or remove it now and use overloaded functions that take an implicit system (removing the implicit parameter in the follow up issue)?

I support either course of action.

kag0:
I'll start on the documentation if the api looks generally fine.
Two questions on the docs:
1. What are the main safety concerns you want to cover? The two big directions I see are security (authentication mostly), and public use (ie. people pointing the reverse proxy at servers they don't control and the response ending up back in a browser).
2. I realize that this isn't actually a directive, it's a route. I should probably rename some classes and stuff. In that case, is https://doc.akka.io/docs/akka-http/current/routing-dsl/routes.html a good place for the reverse proxy documentation to go?

PS. as far as safety concerns blocking a merge my stance is that given enough people are writing their own reverse-proxies (sometimes poorly) or using the code from the last PR, it is safer to have this in the main library than tell people to implement it themselves.

Issue #2216, opened 2018-09-19
Add `aroundRequest` directive which captures complete request / response processing duration


jrudolph:
Many people have attempted to write an `aroundRequest` directive that captures the full request / response cycle to precisely measure and log processing times.

It needs to take into regard:

 * sending request and response entity out
 * timeouts
 * exceptions / rejections

It would be good if we would provide this directive out-of-box so there's only one implementation to maintain.

Here are some third-party attempts:

https://medium.com/@krzysztof.ciesielski/i-extended-this-implementation-to-make-it-handle-timeouts-27fefd5b2e23
https://blog.softwaremill.com/measuring-response-time-in-akka-http-7b6312ec70cf

kstokoz:
Honestly i think this boils down to provide out of the box directives to allow people create proper access logs (see #2181) which would allow connecting request/response, properly handling rejections and timeouts, having callbacks for measuring response times for SLAs (last byte in-fist byte out, last byte in - last byte out).
Existing `logRequestResult` directive is not flexible enough.

In my current project we have similar to linked custom directives to write access logs, it would be quite convenient to have it out of the box. I think for any enterprise proper access logs is a must.

patriknw:
We can go ahead with this.

kstokoz:
For example server idle timeout which is happening outside of routes are not even reported correctly right now.

For put/post requests when entity idle timeout is triggered original `HttpIdleTimeoutException` is lost in debug log and in handleExceptions  receives `akka.http.scaladsl.model.EntityStreamException: Entity stream truncation`

kstokoz:
It would be also great to be able to log requests rejected by `GracefulTerminatorStage` during shutdown.

kstokoz:
@jrudolph i would agree that with current implementation addressing all of cases can be problematic.
In current project we have something similar to the suggested `around` directive + combination of exception handlers + timeout handlers - i can cover most of the mentioned cases and manage to write access log entry. Only which is fully out of our control is requests rejected during termination.

It would be really awesome if akka would either provided out of the box directive for all of them or ability to register a callback to write access log, if not as a part of this issue though.

ssgao:
The examples provided here won't really capture the full response time. It only captures the time right before the response starts streaming. That's more of a [*first byte time*](https://en.wikipedia.org/wiki/Time_to_first_byte). A full response time should include the time it takes for the client to receive it. So far I haven't found a way to capture that. Would be interested to know if anyone knows.

[...]

PR #2814, opened 2019-11-17
aroundrequest: Initial commit - handles rejections, timeouts, exceptions

mantisMolloy:
# Pull Request Checklist

* [y ] Have you read through the [contributor guidelines](./CONTRIBUTING.md)?
* [y] Have you signed the [Lightbend CLA](https://www.lightbend.com/contribute/cla)?
* [n] Have you updated the documentation? // Will update if approach is deemed acceptable
* [ y] Have you added tests for any changed functionality?

## Purpose

Adds new `aroundRequest` directive which takes into consideration the following:

- sending request and response entity out
- `aroundRequest` directives can be composed on top of each other
- handles timeouts
- handles exceptions / rejections

## References

References #2216 

## Changes
- Adds new synthetic header `OnComplete`
- `OnComplete` header has `OnCompleteAccess` field which holds functions to be applied when request completes
- `aroundRequest` directive intercepts `OnComplete` header in DSL routes and add user supplied functions to `OnComplete` header
- In `HttpServerBluePrint.scala` we add an empty `OnComplete` header on incoming requests
- In `HttpServerBluePrint.scala` we extract `OnComplete` header when request completes and apply any functions added to header

## Background Context

A user can supply a function of type `RequestContext => HttpResponse => HttpResponse` using `aroundRequest` directive.

Timeouts, exceptions, rejections will already have been handled at the point where we extract the header and apply the functions.

Usage: 

```
import io.prometheus.client.{Gauge, Histogram}

val timed: RequestContext => HttpResponse => HttpResponse = ctx => {
  requestsInProgress.inc()
  val requestTimer = requestLatencySeconds.startTimer()

   (response: HttpResponse) => requestTimer.observeDuration(); response
}

val route: Route = ???

val sealed = Route.seal(route)

val http = Http().bindAndHandle(handler = aroundRequest(timed)(sealed), interface = hostname, port = httpPort)
```


mantisMolloy:
Need to address the following issues:

```[error]  * abstract method update(scala.concurrent.duration.Duration,scala.Function1)Unit in interface akka.http.scaladsl.TimeoutAccess does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.scaladsl.TimeoutAccess.update")
[error]  * abstract method updateOnDone(scala.Function1)Unit in interface akka.http.scaladsl.TimeoutAccess is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.scaladsl.TimeoutAccess.updateOnDone")
[error]  * abstract method update(scala.concurrent.duration.Duration,scala.Function1,scala.Function1)Unit in interface akka.http.scaladsl.TimeoutAccess is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.scaladsl.TimeoutAccess.update")```

File: akka-http/src/main/scala/akka/http/scaladsl/server/directives/AroundRequestDirectives.scala [hunk 1]
```diff
@@ -0,0 +1,65 @@
...
+        case r: Rejected                                   => onDone(Success(r)); r
+      }.recoverWith({
+        case e: Throwable =>
+          onDone(Failure(e))
+          Future.failed(e)
+      })
+    }
+
+  private def onEntityTermination(onDone: Try[RouteResult] => Unit, result: RouteResult)(implicit ec: ExecutionContextExecutor): Graph[FlowShape[ByteString, ByteString], NotUsed] =
+    Flow[ByteString].watchTermination() {
```
ssgao: `watchTermination` is built on a GraphStage, which only supports `onUpstreamFinish()`, and it's `onDownstreamFinish()` is only called when there's an exception. See https://github.com/akka/akka/blob/master/akka-stream/src/main/scala/akka/stream/impl/fusing/GraphStages.scala#L128.

In case of the chunk response, it'll stop before the last chunk is streamed to the client, and in case of a large response, it'll be the first byte time.

See https://github.com/ssgao/akka-http-quickstart-scala for simple demo.

mantisMolloy: I'm not seeing this issue. Hopefully I'm not missing something. 

I tested this 2 ways.

1. This unit test: https://github.com/akka/akka-http/pull/2814/files#diff-5a8c270ba581eaaa8c1ad33be1e3202eR134

2. I cloned the repo you linked and tested with a large file - `1.3GB`

**Procedure:**
Configure curl to measure transfer times. Create a file

mantisMolloy: I tried the second route named `fooc` in the example app and got similar results

```
tom.molloy@...:~/repos/public/akka-http-quickstart-scala (master)$ curl -w "@curl-format.txt" -o /dev/null -s "http://localhost:8080/fooc"
time_namelookup:  0.005077
time_connect:  0.005419
time_appconnect:  0.000000
time_pretransfer:  0.005470
time_redirect:  0.000000
time_starttransfer:  0.294510
----------
time_total:  18.760435
```

```
root [helloAkkaHttpServer-akka.actor.default-dispatcher-3] INFO  c.e.QuickstartServer$ - [200] GET http://localhost:8080/fooc took: 18647ms
```

shestero: > `watchTermination` is built on a GraphStage, which only supports `onUpstreamFinish()`, and it's `onDownstreamFinish()` is only called when there's an exception. 
Can some transparent Flow in the bottom send Cancel/Done/NonFailureCancellation etc up to signal when it receive its onUpstreamFinish? So we can ensure that streams are drain?

mantisMolloy: Interesting idea. This might be a simple solution if it's doable.

Another idea I was thinking of is adding a function in `OutgoingConnectionBlueprint.scala`. The function will be applied when `MessageEnd` is received.

We can add a synthetic header to the request which allows access and modification of the function similar to how `Timeout-Access` is implemented.

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/AroundRequestDirectivesSpec.scala
```diff
@@ -0,0 +1,220 @@
...
+      }
+    }
+  }
+
+  "the aroundRequest directive when used on a chunked response" should {
+    "call the onDone function when last bytes are sent" in {
+
+      val timed: RequestContext => Try[RouteResult] => Unit = ctx => {
+        val testKey = ctx.request.uri.toRelative.toString
+        val start = System.currentTimeMillis()
```
Crack, line 138: Shouldn't this use `System.nanoTime()`? Unlike `currentTimeMillis` it's guaranteed to be monotonic.

raboof:
OK TO TEST

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -397,12 +397,13 @@ private[http] object HttpServerBluePrint {
         def onPush(): Unit =
           grab(requestParsingIn) match {
             case r: RequestStart =>
-              openRequests = openRequests.enqueue(r)
+              val rs0 = r.copy(headers = r.headers :+ OnComplete(OnCompleteAccess()))
```
raboof, line 400: This would probably be a little nicer when implemented as an attribute (https://doc.akka.io/docs/akka-http/snapshot/common/http-model.html#attributes)

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -430,19 +431,24 @@ private[http] object HttpServerBluePrint {
...
-            else response0.mapEntity { e =>
-              val (newEntity, fut) = HttpEntity.captureTermination(e)
-              fut.onComplete {
-                case Failure(ex) =>
-                  log.error(ex, s"Response stream for [${requestStart.debugString}] failed with '${ex.getMessage}'. Aborting connection.")
-                case _ => // ignore
-              }(ExecutionContexts.sameThreadExecutionContext)
-              newEntity
-            }
+          val response = response0 mapEntity { e =>
```
raboof, line 438: This looks like it would fairly seriously impact performance - perhaps it should be behind a configuration flag?

PR #2820, opened 2019-11-19
build: use more resilient checking for Scala 2.13 in Mima

### LINKED ISSUE #2819: Failed: Mima broken with Scala 2.13
https://jenkins.akka.io:8498/job/akka-http-nightly/702/AKKA_VERSION=default,Node=akka-http-nightly-node,SCALA_VERSION=2.13.0,jdk=AdoptOpenJDK%208/console

```
[error] (akka-http-caching / mimaPreviousClassfiles) sbt.librarymanagement.ResolveException: unresolved dependency: com.typesafe.akka#akka-http-caching_2.13;10.1.1: not found
```

jrudolph:
Fix #2819

The real issue is that nightlies were using `++2.13.0` last night.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -24,10 +24,8 @@ object Dependencies {
   lazy val specs2Version = settingKey[String]("The version of Specs2 to use")
   lazy val scalaCheckVersion = settingKey[String]("The version of ScalaCheck to use.")
 
-  val Scala213 = "2.13.1"
-
   val Versions = Seq(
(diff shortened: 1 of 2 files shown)


raboof:
I thought there was something in sbt that'd prevent choosing `++2.13.0` when `crossScalaVersions` does not contain that version... apparently not?

LGTM, but if we can instead prevent switching to versions not in the list I'd prefer that

jrudolph:
> LGTM, but if we can instead prevent switching to versions not in the list I'd prefer that

That's a good point, I was also surprised that the check doesn't seem to be there any more.

Issue #2830, opened 2019-11-27
Request level client API connection failures should not be a warning


huntc:
Failed HTTP connections such as the following shouldn't be treated as a warning IMHO. Failure is normal.

```
2019-11-27T05:58:30.541Z [05:58:30.538UTC] 127.0.0.1 com.cisco.streambed.controlcenter WARN  PoolGateway - Connection attempt failed. Backing off new connection attempts for at least 100 milliseconds.
2019-11-27T05:58:34.988Z [05:58:34.988UTC] 127.0.0.1 com.cisco.streambed.controlcenter WARN  PoolGateway - Connection attempt failed. Backing off new connection attempts for at least 200 milliseconds.
2019-11-27T05:58:35.440Z [05:58:35.440UTC] 127.0.0.1 com.cisco.streambed.controlcenter WARN  PoolGateway - Connection attempt failed. Backing off new connection attempts for at least 400 milliseconds.
2019-11-27T05:58:43.230Z [05:58:43.229UTC] 127.0.0.1 com.cisco.streambed.controlcenter WARN  PoolGateway - Connection attempt failed. Backing off new connection attempts for at least 800 milliseconds.
2019-11-27T05:58:44.295Z [05:58:44.295UTC] 127.0.0.1 com.cisco.streambed.controlcenter WARN  PoolGateway - Connection attempt failed. Backing off new connection attempts for at least 1600 milliseconds.
```

These messages pollute logs where it is quite normal for there to be intermittent network connectivity.

I recommend that connection failures are instead logged at the debug level. Here's the particular line that I recommend being changed: https://github.com/akka/akka-http/blob/095949f13acb802f533df8413f8ce31438ac84fc/akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/NewHostConnectionPool.scala#L136

I'm happy to raise a PR if there's agreement.

jrudolph:
Warning might indeed be to strong.

So, you would say, if the response would finally fail and carry a good enough error message, that should suffice for most cases. If an error is intermittent on the other hand and requests succeed finally then the user shouldn't be bothered?

Sounds reasonable.

huntc:
Yeah, the stream/future will fail with `StreamTcpException` as proven with the following code:

```
val _ = RestartSource
  .withBackoff(minBackoff, maxBackoff, backoffRandomFactor)(
    () =>
      someHttpConnectedSource
        .recover {
          case _: StreamTcpException => Source.empty
      })
[...]

PR #2837, opened 2019-11-27
Tone down the connection failure warning

huntc:
Failure is normal.

Fixes https://github.com/akka/akka-http/issues/2830


raboof:
OK TO TEST

PR #2853, opened 2019-12-09
= only catch NonFatal Exceptions

rahilb:
## Purpose

 I spotted this while reviewing some code https://github.com/akka/akka-http/pull/2503/files/fa1e4241b004a2c92d0513a1a9a55ab00b811f0c#diff-934eaac30edce2e74265c0760314da50 

No longer catch exceptions that are fatal to the JVM. 

## References

References  https://github.com/akka/akka-http/pull/2503/files/fa1e4241b004a2c92d0513a1a9a55ab00b811f0c#diff-934eaac30edce2e74265c0760314da50

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpResponseParser.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpResponseParser.scala
@@ -6,7 +6,7 @@ package akka.http.impl.engine.parsing
 
 import scala.annotation.tailrec
 import scala.concurrent.Promise
-import scala.util.control.NoStackTrace
+import scala.util.control.{ NoStackTrace, NonFatal }
 import akka.http.scaladsl.settings.ParserSettings
 import akka.http.impl.model.parser.CharacterClasses
 import akka.util.ByteString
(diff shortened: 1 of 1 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpResponseParser.scala
```diff
@@ -79,7 +79,7 @@ private[http] class HttpResponseParser(protected val settings: ParserSettings, p
               val reason = asciiString(input, reasonStartIdx, reasonEndIdx)
               StatusCodes.custom(code, reason)
             } catch {
-              case _: Exception => badStatusCodeSpecific(code)
+              case NonFatal(_) => badStatusCodeSpecific(code)
```
johanandren: There are no fatals that are `Exception` they are all `Throwable` or `Error`.

jrudolph: I agree, though, I'm also fine with using the common style of using `NonFatal` here. It's probably slightly slower (since it checks for more cases) but that shouldn't matter since it is an error case anyway.

jrudolph:
OK TO TEST

PR #2856, opened 2019-12-11
Disable idle-timeout in Http2

pribor:
Idle timeout handling logic was moved to http layer in this commit: akka/akka-http@81ae350

idle timeout parameter was disabled in http 1.x https://github.com/akka/akka-http/blob/master/akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala#L138

however it wasn't disabled in http2 https://github.com/akka/akka-http/blob/master/akka-http2-support/src/main/scala/akka/http/scaladsl/Http2.scala#L63 which results in duplicate idle timeout logic on diffrent layers http/tcp that doesn't work correctly together. As a result terminated connection doesn't propagate `onDownstreamFinish` and the stream is stopped forever.


raboof:
OK TO TEST

pribor:
#2776

PR #2857, opened 2019-12-11
Fix project-info links to API docs

raboof:
project-info.jar does not work (https://github.com/lightbend/sbt-paradox-project-info/issues/15),
but linking to `current` is probably better for SEO reasons anyway.


raboof:
(this was only in the 10.1.11 and snapshot versions of the deployed docs, fixed the 10.1.11 manually on gustav)

PR #2858, opened 2019-12-12
Docs: set the correct version for project info

ennru:
## Purpose

Make sure the project info gets hold of the correct version number.

## References

Fixes https://github.com/lightbend/sbt-paradox-project-info/issues/15

### DIFF
--- a/project/Common.scala
+++ b/project/Common.scala
@@ -11,7 +11,7 @@ import com.lightbend.paradox.projectinfo.ParadoxProjectInfoPluginKeys._
 object Common extends AutoPlugin {
   override def trigger: PluginTrigger = allRequirements
   override lazy val projectSettings = Seq(
-  projectInfoVersion := (if (isSnapshot.value) "snapshot" else version.value),
+    projectInfoVersion := (if (isSnapshot.value) "snapshot" else version.value),
     scalacOptions ++= Seq(
       "-deprecation",
       "-encoding", "UTF-8", // yes, this is 2 args
--- a/project/Common.scala
+++ b/project/Common.scala
@@ -6,10 +6,12 @@ package akka
 
 import sbt._
 import Keys._
+import com.lightbend.paradox.projectinfo.ParadoxProjectInfoPluginKeys._
 
 object Common extends AutoPlugin {
   override def trigger: PluginTrigger = allRequirements
   override lazy val projectSettings = Seq(
+  projectInfoVersion := (if (isSnapshot.value) "snapshot" else version.value),
     scalacOptions ++= Seq(
       "-deprecation",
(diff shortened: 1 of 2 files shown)


ennru:
I find it important to let the users discover the versioned API docs here. That is part of my idea behind the project info section.
If we don't want to show patch version API docs to anybody, we should always just keep the latest version per minor instead.

project/Common.scala
```diff
@@ -6,10 +6,12 @@ package akka
 
 import sbt._
 import Keys._
+import com.lightbend.paradox.projectinfo.ParadoxProjectInfoPluginKeys._
 
 object Common extends AutoPlugin {
   override def trigger: PluginTrigger = allRequirements
   override lazy val projectSettings = Seq(
+  projectInfoVersion := (if (isSnapshot.value) "snapshot" else version.value),
```
jrudolph: Indentation:

```suggestion
    projectInfoVersion := (if (isSnapshot.value) "snapshot" else version.value),
```

ennru:
This value is used to create the API links in the Project Info sections of the docs (it reverts #2857).
I know @raboof and I might have different opinions on what to link to, without this change Akka HTTP 10.1.x documentation will link to 10.2.x API docs once they are considered "current", which I find wrong. With this change they will point to the exact same version of API docs as the docs they belong to.

jrudolph:
> This value is used to create the API links in the Project Info sections of the docs (it reverts #2857).
> I know @raboof and I might have different opinions on what to link to, without this change Akka HTTP 10.1.x documentation will link to 10.2.x API docs once they are considered "current", which I find wrong. With this change they will point to the exact same version of API docs as the docs they belong to.

Hmm, for me it only shows the project info link for `current` right now, so it probably wouldn't help.

That said, for API docs, I agree, that putting the right version there makes sense (but I'm more concerned with usability of the site vs. SEO).

We also have https://doc.akka.io/docs/akka-http/10.1.11/reference.html which also links to the right version of the API docs (and is more likely being used than the collapsed project info box).

jrudolph:
> Hmm, for me it only shows the project info link for `current` right now, so it probably wouldn't help.

Do we have an explanation for that?

ennru:
With these changes, you should not get `current`, either `snapshot` or the fixed version during releasing.

jrudolph:
Hmm, locally built the links now go to `https://doc.akka.io/api/akka-http/0.1.0-SNAPSHOT/akka/http/scaladsl/index.html` which doesn't seem right.

ennru:
Are you sure you reloaded sbt?

jrudolph:
> Are you sure you reloaded sbt?

Hmm, seems to work for me today :)

Issue #2854, opened 2019-12-09
Problem with query parameter separator


rtfpessoa:
After moving Play Framework from version 2.4 to 2.7 and starting to use akka-http as the server we encountered one problem with some weird URLs that we were mistakenly generating.

This url `http://localhost:9000/issue/1300068591272/_lines?projectId=22877&commitId=15565520`  reached the server with query parameter called `projectId` and `commitId` while this url `http://localhost:9000/issue/1300068591272/_lines?projectId=22877&amp;commitId=15565520` does not get the second.

Seems like this might be related to the way akka-http parses URLs.
Apparently the second URL is also valid because at least `;` is a supported separator.

I would gladly fix this if you point me in the right direction.

References:
* https://www.ietf.org/rfc/rfc3986.txt
* https://www.w3.org/TR/html401/appendix/notes.html#h-B.2.2

raboof:
Interesting, I wasn't aware of that recommendation but indeed it seems somewhat widespread: from your message I understand Netty does this, it is mentioned on https://en.wikipedia.org/wiki/Query_string#Web_forms and for example the Apache http client code seems to do it as well:

```
@ import $ivy.`org.apache.httpcomponents:httpclient:4.5.9`
@ import java.net.URI
@ import java.nio.charset.Charset
@ val params = URLEncodedUtils.parse(new URI("https://akka.io?x=1&amp;y=2"), Charset.forName("UTF-8"))
params: java.util.List[org.apache.http.NameValuePair] = [x=1, amp, y=2]
```

I think it might be reasonable to support this in Akka HTTP as well. There are 2 possible consequences:

* people who really wanted to use `;` characters in their query parameters have a problem
* it might be tricky to support both `&` and `;` as separators with the same performance as supporting `&` specifically

I think the first consequence might be OK: they should have been encoding those as `%3B` anyway, so this behavior change might be acceptable for the upcoming 10.2.0 release.

The second consequence is trickier: it might be good to make sure to benchmark the parsing performance before and after the change. You can have a look at the `akka-http-bench-jmh` subproject.

For the actual code change `akka.http.impl.model.parser.UrlParser` is probably where you should look.

Thanks for bringing this up, I'm looking forward to seeing what you come up with!

rtfpessoa:
Just as a note, the class seems to be `akka.http.impl.model.parser.UriParser`

rtfpessoa:
@raboof, imagine that I want to run the benchmarks to check the code, how would be the command?
[...]

PR #2861, opened 2019-12-26
bench: add UriParser benchmarks

rtfpessoa:
Initial work for https://github.com/akka/akka-http/issues/2854

### DIFF
--- a/akka-http-bench-jmh/src/main/scala/akka/http/impl/model/parser/UriParserBenchmark.scala
+++ b/akka-http-bench-jmh/src/main/scala/akka/http/impl/model/parser/UriParserBenchmark.scala
@@ -1,0 +1,34 @@
+package akka.http.impl.model.parser
+
+import java.util.concurrent.TimeUnit
+
+import akka.actor.ActorSystem
+import akka.http.scaladsl.model.Uri
+import akka.parboiled2.UTF8
+import org.openjdk.jmh.annotations._
+
+import scala.concurrent.Await
+import scala.concurrent.duration._
+
+@State(Scope.Benchmark)
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+@BenchmarkMode(Array(Mode.Throughput))
+class UriParserBenchmark {
+  implicit val system: ActorSystem = ActorSystem("uri-parser-benchmark")
+
+  val url = "http://any.hostname?param1=111&amp;param2=222"
+
+  @Setup
+  def setup(): Unit = ()
+
+  @TearDown
+  def tearDown(): Unit = {
+    Await.result(system.terminate(), 5.seconds)
+  }
+
+  @Benchmark
+  def bench_parse_uri(): Unit = {
+    Uri(url).query(UTF8, Uri.ParsingMode.Relaxed)
+  }
+
+}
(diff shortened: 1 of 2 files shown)


akka-http-bench-jmh/src/main/scala/akka/http/impl/model/parser/UriParserBenchmark.scala
```diff
@@ -0,0 +1,34 @@
...
+  def setup(): Unit = ()
+
+  @TearDown
+  def tearDown(): Unit = {
+    Await.result(system.terminate(), 5.seconds)
+  }
+
+  @Benchmark
+  def bench_parse_uri(): Unit = {
+    Uri(url).query(UTF8, Uri.ParsingMode.Relaxed)
```
raboof: I think you have to return (or explicitly `Blackhole`) the result to make double-sure the optimizer doesn't remove the thing we're trying to benchmark - though I see we haven't done that in `LineParserBenchmark` either it seems...

raboof:
OK TO TEST

akka-http-bench-jmh/src/main/scala/akka/http/impl/model/parser/UriParserBenchmark.scala
```diff
@@ -0,0 +1,35 @@
...
+import scala.concurrent.Await
+import scala.concurrent.duration._
+
+@State(Scope.Benchmark)
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+@BenchmarkMode(Array(Mode.Throughput))
+class UriParserBenchmark {
+  implicit val system: ActorSystem = ActorSystem("uri-parser-benchmark")
+
+  val url = "http://any.hostname?param1=111&amp;param2=222"
```
jrudolph: Could you make this a `@Param` instead so that we can test more that this one uri? There might be performance implications not just for the immediate case to be changed.

akka-http-bench-jmh/src/main/scala/akka/http/impl/model/parser/UriParserBenchmark.scala
```diff
@@ -0,0 +1,35 @@
...
+import org.openjdk.jmh.infra.Blackhole
+
+import scala.concurrent.Await
+import scala.concurrent.duration._
+
+@State(Scope.Benchmark)
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+@BenchmarkMode(Array(Mode.Throughput))
+class UriParserBenchmark {
+  implicit val system: ActorSystem = ActorSystem("uri-parser-benchmark")
```
jrudolph: Is the system really needed?

akka-http-bench-jmh/src/main/scala/akka/http/impl/model/parser/UriParserBenchmark.scala
rtfpessoa: Ok, I will also add a bigger url.

akka-http-bench-jmh/src/main/scala/akka/http/impl/model/parser/UriParserBenchmark.scala
rtfpessoa: That is right, not sure why I left it there

Issue #1500, opened 2017-10-31
HTTP message attributes


gmethvin:
It would be nice if we could attach arbitrary typed values to Akka HTTP requests and responses. This could be used for things like tracking request ids/metadata, and perhaps for other things people are currently using custom headers for.

For example:

```scala
// This is a type-safe key to retrieve the attribute
// The String ID is a unique identifier for this attribute
val RequestId: Attr[UUID] = Attr("com.example.foo.bar.RequestId")

// get a new request with the RequestId attribute added
val requestWithId = request.withAttr(RequestId, UUID.randomUUID)

// get the attribute from the request
val requestId: UUID = requestWithId.getAttr(RequestId)
```
Ideally the API should be type safe so I can only assign an `Attr[A]` to a value of type `A`, but if it were possible to write a type-safe wrapper that would probably be sufficient.

My future use case for this would be allowing us to completely represent a Play request as an Akka HTTP request, attaching various Play metadata as attributes.

ktoso:
Agreed, we currently work around this by having synthetic headers which isn't the most hygienic way to go around it and leaks sometimes. 

FYI @bantonsson @jrudolph

sarahgerweck:
This would be fantastically useful. 👍

johanandren:
For me that sounds an awful lot like the request context of Servlets/JSP (or thread local for that part). Wouldn't it be better to use functional composition and explicit state passing for these kind of things than an arbitrary map of things in context?

jrudolph:
@scr-oath, the whole routing layer uses immutable data structures (aside from the streaming aspects). So, it will not be possible to mutate things in inner directive and have access to it from outer directives.

What you can do already now is to add custom headers to a request or a response:

 * Add it to the request to pass info from the outside to inside routes
 * Add it to the response to pass info from the inside to outside routes

This already works now but is a bit cumbersome because you need to define custom headers (though, note, that this is how HTTP works in general. E.g. in proxied setups, the proxies also add headers to requests and may read responses to communicate with backend servers).

This ticket describes that we want to introduce another structure, similar to custom headers that simplifies passing around information inside of Akka HTTP but the concept will exactly be the same as with custom headers.

scr-oath:
Ok I understand the thirst for immutable rather than the Von Neumann style.  Then, I suppose this change/feature-request would be to add arbitrary objects to the request/response (not just strings) and not to the headers (because they're not destined for output)?  That would really be great!

PR #2872, opened 2020-01-06
First draft of message attributes

raboof:
Refs  #1500

This approach follows the Play API (https://www.playframework.com/documentation/2.8.x/Highlights26)
in that you can add attributes of any user type. Otherwise it follows the
existing conventions from Akka HTTP Headers, so you can have multiple
attributes of the same type. If you want to distinguish different attributes of
the same type, like you could in Play with different keys, you would have to
introduce a wrapper type (either holding the key or creating a separate wrapper
for each key).

The main downside of this approach is that it increases the memory usage of a
message with one field.

### DIFF
--- a/akka-http-core/src/main/java/akka/http/javadsl/model/HttpMessage.java
+++ b/akka-http-core/src/main/java/akka/http/javadsl/model/HttpMessage.java
@@ -49,13 +49,13 @@ public interface HttpMessage {
 
     /**
      * Try to find the first header with the given name (case-insensitive) and return
-     * Some(header), otherwise this method returns None.
+     * Optional.of(header), otherwise this method returns an empty Optional.
      */
     Optional<HttpHeader> getHeader(String headerName);
 
     /**
      * Try to find the first header of the given class and return
-     * Some(header), otherwise this method returns None.
+     * Optional.of(header), otherwise this method returns an empty Optional.
      */
     <T extends HttpHeader> Optional<T> getHeader(Class<T> headerClass);
 
@@ -64,6 +64,22 @@ public interface HttpMessage {
      * of this message.
      */
     <T extends HttpHeader> Iterable<T> getHeaders(Class<T> headerClass);
+
+    /**
+     * An iterable containing the attributes for this message.
+     */
+    Iterable<Object> getAttributes();
+
+    /**
+     * Try to find the first attribute of the given class and return
+     * Optional.of(attribute), otherwise this method returns an empty Optional
+     */
+    <T> Optional<T> getAttribute(Class<T> attributeClass);
+
+    /**
+     * An iterable containing all attributes of the given class of this message
+     */
+     <T> Iterable<T> getAttributes(Class<T> attributeClass);
 
     /**
      * The entity of this message.
--- a/akka-http-core/src/main/mima-filters/10.1.11.backwards.excludes/issue-1500-attributes.excludes
+++ b/akka-http-core/src/main/mima-filters/10.1.11.backwards.excludes/issue-1500-attributes.excludes
@@ -1,0 +1,12 @@
+# Adding a method to an interface marked DoNotInherit is OK
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.javadsl.model.HttpMessage.getAttributes")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.javadsl.model.HttpMessage.getAttribute")
+
+# Adding a method to a sealed trait is safe
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.scaladsl.model.HttpMessage.attributes")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.scaladsl.model.HttpMessage.withAttributes")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.scaladsl.model.HttpMessage.mapAttributes")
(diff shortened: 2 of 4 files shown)


raboof:
failure was #2349

raboof:
PLS BUILD

akka-http-core/src/main/java/akka/http/javadsl/model/HttpMessage.java
```diff
@@ -65,6 +65,22 @@
...
+    /**
+     * An iterable containing the attributes for this message.
+     */
+    Iterable<Object> getAttributes();
+
+    /**
+     * Try to find the first attribute of the given class and return
+     * Optional.of(header), otherwise this method returns an empty Optional
+     */
+    <T> Optional<T> getAttribute(Class<T> attributeClass);
```
johanandren: If we leave it allowing multiple maybe it should rather be `firstAttributeOf` or something that conveys what the docs now say (that'd also align it with the collection api somewhat on the Scala side)?

johanandren: Ah, but header can be multiple and getHeader returns first. Got it. NVM then.

raboof: yeah, it's somewhat confusing but being consistent with `getHeader` might be worth it anyway

raboof:
I picked it mainly for its API familiarity, also given attributes and headers are similar contexts (being attached to the requests/responses and not the requestcontext). We can't predict what kinds of objects people would want to store in the attributes, and adding typed keys/values like in play (https://www.playframework.com/documentation/2.8.x/Highlights26#Request-attributes) seemed somewhat boilerplatey from a user perspective. Do you have another pattern in mind that might fit well?

marcospereira:
> We can't predict what kinds of objects people would want to store in the attributes, and adding typed keys/values like in play (https://www.playframework.com/documentation/2.8.x/Highlights26#Request-attributes) seemed somewhat boilerplatey from a user perspective.

Yeah, but I wonder what would be the use case of adding attributes without a way to clearly identify them:

```scala
req.addAttributes("something", "something else")
```

It feels to me that users will then need to define custom types to wrap values so that they can be identified by that type.

Given that, should we use `@ApiMayChange` here since we aren't sure about how this API should be? (sorry if this is obvious and the annotations aren't there because it is still a draft).

raboof:
> should we use `@ApiMayChange` here since we aren't sure about how this API should be

I think we should pick how this should look like before merging. Perhaps still add `@ApiMayChange` to allow small tweaks, but iterate on the general approach in PR's.

> I wonder what would be the use case of adding attributes without a way to clearly identify them

I agree that doesn't make sense

> It feels to me that users will then need to define custom types to wrap values so that they can be identified by that type.

Do you think that is a good or a bad thing? ;).

One of the use cases for message attributes is the TLS metadata. With the proposal in this PR getting that would be something like `request.attributes[TlsSessionInfo]`, producing a possibly-empty list of `TlsSessionInfo` objects. Especially since in this case there's always one (or zero) that doesn't make too much sense. Perhaps indeed modelling this after the headers API's wasn't such a good fit.

We could also have the convention to put the key on the companion object, so you'd get something like `val tlsSessionInfo: Option[TlsSessionInfo] = request.attribute(TlsSessionInfo.key)`? So introducing a `case class AttributeKey[Type](name: String)` and getting the value through `def attribute[T](key: AttributeKey[T]): Option[T]`? (or a directive that does the same)

jrudolph:
> We could also have the convention to put the key on the companion object, so you'd get something like `val tlsSessionInfo: Option[TlsSessionInfo] = request.attribute(TlsSessionInfo.key)`? So introducing a `case class AttributeKey[Type](name: String)` and getting the value through `def attribute[T](key: AttributeKey[T]): Option[T]`? (or a directive that does the same)

I would like that.

Another use case would be the `Remote-Address` feature so it couldn't be mistaken as a client-provided header as suggested in #2924.

jrudolph:
> One of the use cases for message attributes is the TLS metadata. With the proposal in this PR getting that would be something like `request.attributes[TlsSessionInfo]`, producing a possibly-empty list of `TlsSessionInfo` objects. Especially since in this case there's always one (or zero) that doesn't make too much sense. Perhaps indeed modelling this after the headers API's wasn't such a good fit.

I would suggest to allow only a single value per key then (or none).

raboof:
> I would suggest to allow only a single value per key then (or none).

Yes, exactly - that's another motivation for not trying to follow the headers API for the attributes, then. I'll try to prototype the approach with `AttributeKey[Type]` today.

marcospereira:
> Do you think that is a good or a bad thing? ;).

Sorry for not being clear: I think it is bad, mainly if you are adding your own types into the attributes (for example `User`). As a user, I don't want to wrap my own types in order to better identify them.

> We could also have the convention to put the key on the companion object, so you'd get something like `val tlsSessionInfo: Option[TlsSessionInfo] = request.attribute(TlsSessionInfo.key)`? So introducing a `case class AttributeKey[Type](name: String)` and getting the value through `def attribute[T](key: AttributeKey[T]): Option[T]`? (or a directive that does the same)

That sounds great to me. In Play, there is an inner companion object called `Attrs`, so that it is clear what the `key` is about and also if for some reason, it is necessary to have multiple keys, they are better grouped. The access would be like `request.attribute(TlsSessionInfo.Attrs.key)`. I prefer that since it is clear the `key` is about attributes.

raboof:
(other proposal is at #2938)

raboof:
Closing in favor of #2938

PR #2877, opened 2020-01-08
Add detail to onBadProtocol

mdedetrich:
## Purpose

This PR adds detail to `onBadProtocol` error messages to better help diagnose when this error is thrown

## Changes

- This PR simply allows the `detail: String` as well as the original `summary: String` to be passed into the `ErrorInfo` class.
- Note that one side effect of this change is I had to remove default parameters from `ParsingException` due to conflicts with overloaded definitions (you cannot have default parameters with overloaded functions)

## Background Context

`ErrorInfo` already contains a `detail: String` field designed to provide extra information about the issue, this PR simply uses that field to provide more details about the error when `onBadProtocol` is called.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpMessageParser.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpMessageParser.scala
@@ -48,7 +48,7 @@ private[http] trait HttpMessageParser[Output >: MessageOutput <: ParserOutput] {
   protected def headerParser: HttpHeaderParser
   protected def isResponseParser: Boolean
   /** invoked if the specified protocol is unknown */
-  protected def onBadProtocol(): Nothing
+  protected def onBadProtocol(input: ByteString): Nothing
   protected def parseMessage(input: ByteString, offset: Int): HttpMessageParser.StateResult
   protected def parseEntity(headers: List[HttpHeader], protocol: HttpProtocol, input: ByteString, bodyStart: Int,
                             clh: Option[`Content-Length`], cth: Option[`Content-Type`], isChunked: Boolean,
@@ -120,10 +120,10 @@ private[http] trait HttpMessageParser[Output >: MessageOutput <: ParserOutput] {
       protocol = c(7) match {
         case '0' => `HTTP/1.0`
         case '1' => `HTTP/1.1`
-        case _   => onBadProtocol()
+        case _   => onBadProtocol(input.drop(cursor))
       }
       cursor + 8
-    } else onBadProtocol()
+    } else onBadProtocol(input.drop(cursor))
   }
 
   /**
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpRequestParser.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpRequestParser.scala
@@ -82,7 +82,7 @@ private[http] final class HttpRequestParser(
           parseHeaderLines(input, cursor + 2)
         else if (byteChar(input, cursor) == '\n')
           parseHeaderLines(input, cursor + 1)
-        else onBadProtocol()
+        else onBadProtocol(input.drop(cursor))
       } else
         // Without HTTP pipelining it's likely that buffer is exhausted after reading one message,
         // so we check above explicitly if we are done and stop work here without running into NotEnoughDataException
@@ -167,7 +167,7 @@ private[http] final class HttpRequestParser(
       uriEnd + 1
     }
 
-    override def onBadProtocol(): Nothing = throw new ParsingException(HttpVersionNotSupported)
+    override def onBadProtocol(input: ByteString): Nothing = throw new ParsingException(HttpVersionNotSupported, "")
 
     // http://tools.ietf.org/html/rfc7230#section-3.3
     override def parseEntity(headers: List[HttpHeader], protocol: HttpProtocol, input: ByteString, bodyStart: Int,
--- a/akka-http-core/src/main/mima-filters/10.2.4.backwards.excludes/2877-better-LogByteStringTools-printByteString.excludes
+++ b/akka-http-core/src/main/mima-filters/10.2.4.backwards.excludes/2877-better-LogByteStringTools-printByteString.excludes
@@ -1,0 +1,2 @@
+# Internals
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.impl.util.LogByteStringTools.printByteString")
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpResponseParser.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpResponseParser.scala
@@ -5,7 +5,6 @@
 package akka.http.impl.engine.parsing
 
 import javax.net.ssl.SSLSession
-
 import scala.annotation.tailrec
 import scala.concurrent.Promise
 import scala.util.control.{ NoStackTrace, NonFatal }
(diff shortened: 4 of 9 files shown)


lightbend-cla-validator:

jrudolph:
OK TO TEST

lightbend-cla-validator:

mdedetrich:
@jrudolph Are this  bincompat errors an issue (I was under the impression these API's are internal/private, they are marked as such)?

jrudolph:
> @jrudolph Are this bincompat errors an issue (I was under the impression these API's are internal/private, they are marked as such)?

No, they are not an issue but we don't have anything in place which would tone them down automatically so we need to create a filter manually.

mdedetrich:
> No, they are not an issue but we don't have anything in place which would tone them down automatically so we need to create a filter manually.

Okay, let me know if I need to do anything additional here.

Also I am fixing up the tests now

mdedetrich:
Okay so tests are now passing, the only issues are erroneous MIMA ones. Let me know if anything else is needed for the PR.

akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpRequestParser.scala
```diff
@@ -158,7 +158,7 @@ private[http] final class HttpRequestParser(
       uriEnd + 1
     }
 
-    override def onBadProtocol(): Nothing = throw new ParsingException(HttpVersionNotSupported)
+    override def onBadProtocol(detail: String): Nothing = throw new ParsingException(HttpVersionNotSupported, "")
```
jrudolph: Why is the detail ignored here?

akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpMessageParser.scala
```diff
@@ -117,10 +117,10 @@ private[http] trait HttpMessageParser[Output >: MessageOutput <: ParserOutput] {
       protocol = c(7) match {
         case '0' => `HTTP/1.0`
         case '1' => `HTTP/1.1`
-        case _   => onBadProtocol
+        case _   => onBadProtocol(input.utf8String)
```
jrudolph: What happens if `utf8String` fails?

jrudolph:
> Okay so tests are now passing, the only issues are erroneous MIMA ones. Let me know if anything else is needed for the PR.

For the mima failures you need to add exclusions to a new file in `akka-http-core/src/main/mima-filters/10.1.11.excludes`.

akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpRequestParser.scala
mdedetrich: I am keeping the original behavior for this specific case because at least personally I was only interested in the `The server-side HTTP version is not supported`.

Generally speaking if the protocol is wrong its very easy to figure out the cause because there are only 3 protocols, but the `The server-side HTTP version is not supported` gets thrown generically in many places.

I have no objection to adding the detail here if requested.

akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpMessageParser.scala
mdedetrich: I couldn't find any documentation on `utf8String` throwing any `Exception`, either directly or indirectly by calling `decodeString`.

mdedetrich:
> For the mima failures you need to add exclusions to a new file in akka-http-core/src/main/mima-filters/10.1.11.excludes.

Sure thing, I will have a look at this tonight (first time doing MIMA related things).

mdedetrich:
@jrudolph Okay so I tried to eliminate as many erroneous bincompat issues however there is still one left, namely

```
Problems for akka-http-core:
akka-http-core: found 1 potential binary incompatibilities while checking against com.typesafe.akka:akka-http-core_2.12:10.1.11  (filtered 4)
 * abstract method onBadProtocol()scala.runtime.Nothing# in interface akka.http.impl.engine.parsing.HttpMessageParser does not have a correspondent in current version
   filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.impl.engine.parsing.HttpMessageParser.onBadProtocol")

akka-http-core: found 1 potential binary incompatibilities while checking against com.typesafe.akka:akka-http-core_2.12:10.1.10  (filtered 4)
 * abstract method onBadProtocol()scala.runtime.Nothing# in interface akka.http.impl.engine.parsing.HttpMessageParser does not have a correspondent in current version
   filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.impl.engine.parsing.HttpMessageParser.onBadProtocol")
```

I am not sure how to get rid of this because I have the same filter here https://github.com/akka/akka-http/blob/93cabed52c239d9639553988a8106233b64c896b/akka-http-core/src/main/mima-filters/10.1.11.backwards.excludes#L2

mdedetrich:
Also note that the context of this PR is that I am experiencing https://stackoverflow.com/questions/46407564/akka-http-post-results-in-the-server-side-http-version-is-not-supported and I am having issues reliably replicating it

jrudolph:
> Also note that the context of this PR is that I am experiencing [stackoverflow.com/questions/46407564/akka-http-post-results-in-the-server-side-http-version-is-not-supported](https://stackoverflow.com/questions/46407564/akka-http-post-results-in-the-server-side-http-version-is-not-supported) and I am having issues reliably replicating it

That's interesting. Since the HTTP protocol version will be the first thing parsed on a connection, it could mean that the other side just isn't an HTTP server. Is the request supposed to be using TLS?

mdedetrich:
Well I have 2 theories as to  why this could be happening

1. Its  an issue related to stackoverflow.com/questions/46407564/akka-http-post-results-in-the-server-side-http-version-is-not-supported
2. Its an issue related to malform HEAD responses from the http server

I don't know which one it is, hence the reason for this PR so I can diagnose it better.

> That's interesting. Since the HTTP protocol version will be the first thing parsed on a connection, it could mean that the other side just isn't an HTTP server. Is the request supposed to be using TLS?

Yes, the request is done against a host with https, specifically `ECDHE-RSA-AES128-GCM-SHA256 TLSv1.2`

jrudolph:
> Its an issue related to stackoverflow.com/questions/46407564/akka-http-post-results-in-the-server-side-http-version-is-not-supported

Not sure that issue can be taken seriously, it has too little information about what actually happened.

> Yes, the request is done against a host with https, specifically `ECDHE-RSA-AES128-GCM-SHA256 TLSv1.2`

I guess trying to connect to an HTTPS/TLS endpoint with http instead of https could lead to that error message. That's also what the SO question seems to imply. The question is why that would happen.

mdedetrich:
> I guess trying to connect to an HTTPS/TLS endpoint with http instead of https could lead to that error message. That's also what the SO question seems to imply. The question is why that would happen.

Well we aren't doing this, instead we are directly connection to the `https://` endpoint so it seems unlikely then.

akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpMessageParser.scala
jrudolph: The relevant input only starts at offset `cursor` so that needs to be taken into account. Maybe we should use `LogByteStringTools.printByteString` instead with a sensible limit (32?) to print the invalid data? That might make it easier to find out if someone is attempting to connect to a server that speaks a binary protocol instead of text-y HTTP.

Btw. this is more important at the other usages. At this place in particular we have already seen `HTTP/1.` so we could be a bit more cleverer here?

jrudolph (approved):
LGTM

I picked this up because someone else recently stumbled over the error messa ge (https://discuss.lightbend.com/t/use-client-side-http2-in-akka-http-2-13-v10-1-12/8473/2). I slightly changed error format to hexdump as I suggested before and also added and adapted more tests.

jrudolph:
Thanks again, @mdedetrich, for the suggestion and the original implementation.

mdedetrich:
@jrudolph No worries, I actually completely forgot about this PR and didn't notice activity due to my old github notification settings being way too eager in spamming me.

Glad that it worked out.

PR #2878, opened 2020-01-08
build: cache coursier directory in travis

jrudolph:


johanandren:
@jrudolph just noticed that may be the wrong directory: https://github.com/akka/akka-grpc/pull/782#issuecomment-572124849

jrudolph:
> @jrudolph just noticed that may be the wrong directory: [akka/akka-grpc#782 (comment)](https://github.com/akka/akka-grpc/pull/782#issuecomment-572124849)

The funny thing is that probably because I added the entry there it actually also seems to use the directory (as hinted in https://get-coursier.io/docs/cache.html#compatibility-with-former-versions).

https://travis-ci.org/akka/akka-http/builds/634246764 found some files in there.

Might still make sense to change it to the right directory.

Issue #2886, opened 2020-01-13
Promote Akka 2.6


raboof:
Since the recent Akka 2.5 updates it becomes smoother to use Akka HTTP with Akka 2.6. 

 * [x] Update the docs to show 2.6 usage (#2996)
 * [ ] Streamline usage where Materializer was required before
 * [ ] Review bootstrapping code wrt to Akka 2.6
 * [x] Go through documentation and remove mentions of Materializer

raboof:
What kinds of things should we be looking for before closing this?

ennru:
Another part of this is https://github.com/akka/akka-http/pull/3242

ennru:
I believe we've covered this now. Closing.

PR #2893, opened 2020-01-17
Simplify route2HandlerFlow

raboof:
Refs #70

~Before merging should be updated not to take a `Materializer` implicit but instead use either #2886 or https://github.com/akka/akka/pull/28494~

~Before merging we need either a release of Akka 2.5.30 (that contains https://github.com/akka/akka/pull/28545) or #2886~

This is ready to go!

### DIFF
--- a/akka-http/src/main/scala/akka/http/scaladsl/server/Route.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/server/Route.scala
@@ -5,6 +5,7 @@
 package akka.http.scaladsl.server
 
 import akka.NotUsed
+import akka.actor.ActorSystem
 import akka.http.scaladsl.model.{ HttpRequest, HttpResponse }
 import akka.http.scaladsl.server.directives.BasicDirectives
 import akka.http.scaladsl.settings.{ ParserSettings, RoutingSettings }
@@ -53,8 +54,16 @@ object Route {
   /**
    * Turns a `Route` into a server flow.
    *
-   * This conversion is also implicitly available through [[RouteResult#route2HandlerFlow]].
+   * This conversion is also implicitly available whereever a `Route`` is used through [[RouteResult#routeToFlow]].
    */
+  // FIXME https://github.com/akka/akka-http/issues/2886 or https://github.com/akka/akka/pull/28494
+  def toFlow(route: Route)(implicit system: ActorSystem, materializer: Materializer): Flow[HttpRequest, HttpResponse, NotUsed] =
+    Flow[HttpRequest].mapAsync(1)(asyncHandler(route, RoutingSettings(system), ParserSettings(system)))
+
+  /**
+   * Turns a `Route` into a server flow.
+   */
+  @deprecated("replaced by handlerFlow that takes a system", "10.2.0")
   def handlerFlow(route: Route)(implicit
     routingSettings: RoutingSettings,
                                 parserSettings:   ParserSettings,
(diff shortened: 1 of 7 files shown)


akka-http/src/main/scala/akka/http/scaladsl/server/Route.scala
```diff
@@ -65,9 +73,29 @@ object Route {
                                 exceptionHandler: ExceptionHandler         = null): Flow[HttpRequest, HttpResponse, NotUsed] =
     Flow[HttpRequest].mapAsync(1)(asyncHandler(route))
 
+  def asyncHandler(route: Route, routingSettings: RoutingSettings, parserSettings: ParserSettings)(implicit system: ActorSystem): HttpRequest => Future[HttpResponse] = {
+    val routingLog = RoutingLog(system.log)
+    val executionContext = system.dispatcher
+    val materializer = SystemMaterializer(system).materializer
```
johanandren: One thing I thought about is that this may now be a different serializer than the http flow is materialized with, wouldn't it make more sense to use that? (The `fromMaterializer(materializer => ...)` operator, or is that only in 2.6?)

raboof:
Depending on where you defined the implicit instances and your compiler parameters, it'll warn about the customer exception/rejectionhandler being unused. Not sure if we can do more...

akka-http/src/main/scala/akka/http/scaladsl/server/Route.scala
```diff
@@ -65,9 +73,31 @@ object Route {
...
+  def asyncHandler(route: Route, routingSettings: RoutingSettings, parserSettings: ParserSettings)(implicit system: ActorSystem): HttpRequest => Future[HttpResponse] = {
+    val routingLog = RoutingLog(system.log)
+    val executionContext = system.dispatcher
+    val materializer = Materializer.matFromSystem(new ClassicActorSystemProvider {
+      override val classicSystem = system
+    })
+    // We can seal more efficiently here because we know we can have no inherited settings:
+    val sealedRoute = {
+      import directives.ExecutionDirectives._
+      (handleExceptions(ExceptionHandler.seal(ExceptionHandler.default(routingSettings))(routingSettings)) & handleRejections(RejectionHandler.default.seal))
```
jrudolph: The default handlers should already be sealed, so maybe this could just be

```scala
handleExceptions(ExceptionHandler.default(routingSettings)) & 
  handleRejections(RejectionHandler.default))
```

?

akka-http/src/main/scala/akka/http/scaladsl/server/Route.scala
```diff
@@ -65,9 +73,31 @@ object Route {
                                 exceptionHandler: ExceptionHandler         = null): Flow[HttpRequest, HttpResponse, NotUsed] =
     Flow[HttpRequest].mapAsync(1)(asyncHandler(route))
 
+  def asyncHandler(route: Route, routingSettings: RoutingSettings, parserSettings: ParserSettings)(implicit system: ActorSystem): HttpRequest => Future[HttpResponse] = {
+    val routingLog = RoutingLog(system.log)
+    val executionContext = system.dispatcher
```
jrudolph: could be implicit

akka-http/src/main/scala/akka/http/scaladsl/server/Route.scala
```diff
@@ -65,9 +73,31 @@ object Route {
                                 exceptionHandler: ExceptionHandler         = null): Flow[HttpRequest, HttpResponse, NotUsed] =
     Flow[HttpRequest].mapAsync(1)(asyncHandler(route))
 
+  def asyncHandler(route: Route, routingSettings: RoutingSettings, parserSettings: ParserSettings)(implicit system: ActorSystem): HttpRequest => Future[HttpResponse] = {
+    val routingLog = RoutingLog(system.log)
+    implicit val executionContext = system.dispatcher
+    implicit val materializer = Materializer.matFromSystem(new ClassicActorSystemProvider {
+      override val classicSystem = system
+    })
+    // We can seal more efficiently here because we know we can have no inherited settings:
```
jrudolph: Not sure what that comment refers to?

akka-http/src/main/scala/akka/http/scaladsl/server/Route.scala
```diff
@@ -65,9 +73,31 @@ object Route {
...
+    val sealedRoute = {
+      import directives.ExecutionDirectives._
+      (handleExceptions(ExceptionHandler.default(routingSettings)) & handleRejections(RejectionHandler.default))
+        .tapply(_ => route)
+    }
+    request => {
```
jrudolph: If that part is exactly the same as below in the old impl, could you extract it?

akka-http/src/main/scala/akka/http/scaladsl/server/Route.scala
```diff
@@ -65,9 +73,25 @@ object Route {
                                 exceptionHandler: ExceptionHandler         = null): Flow[HttpRequest, HttpResponse, NotUsed] =
     Flow[HttpRequest].mapAsync(1)(asyncHandler(route))
 
+  def asyncHandler(route: Route, routingSettings: RoutingSettings, parserSettings: ParserSettings)(implicit system: ActorSystem): HttpRequest => Future[HttpResponse] = {
+    val routingLog = RoutingLog(system.log)
+    implicit val executionContext = system.dispatcher
```
jrudolph: Maybe let's pass those explicitly (if only used for `createAsyncHandler`).

akka-http/src/main/scala/akka/http/scaladsl/server/Route.scala
```diff
@@ -65,9 +73,25 @@ object Route {
...
+    implicit val materializer = Materializer.matFromSystem(new ClassicActorSystemProvider {
```
jrudolph: Isn't this an elaborate way of saying `SystemMaterializer(system).materializer`?

PR #2894, opened 2020-01-17
Introduce scala silencer

raboof:


akka-http-core/src/main/scala/akka/http/javadsl/Http.scala
```diff
@@ -40,6 +42,7 @@ class Http(system: ExtendedActorSystem) extends akka.actor.Extension {
 
   import language.implicitConversions
   private implicit def completionStageCovariant[T, U >: T](in: CompletionStage[T]): CompletionStage[U] = in.asInstanceOf[CompletionStage[U]]
+  @silent("never used")
```
jrudolph: Why is it needed?

raboof: The implicit 'evidence' parameter `ev` is not actually used, so the compiler emits a warning for this line.

Of course introducing the silencer compiler plugin just for this one warning wouldn't make sense, but I think it's a nice tool to be able to suppress spurious warnings going forward in general.

akka-http-core/src/main/scala/akka/http/javadsl/Http.scala
```diff
@@ -40,6 +42,7 @@ class Http(system: ExtendedActorSystem) extends akka.actor.Extension {
...
   private implicit def javaModelIsScalaModel[J <: AnyRef, S <: J](in: Future[J])(implicit ev: JavaMapping.Inherited[J, S]): Future[S] = in.asInstanceOf[Future[S]]
```
jrudolph: We could create a method upcast in `Inherited` instead:

```scala
  class Inherited[J <: AnyRef, S <: J](implicit classTag: ClassTag[S]) extends JavaMapping[J, S] {
    def toJava(scalaObject: S): J = scalaObject
    def toScala(javaObject: J): S = cast[S](javaObject)

    import scala.language.higherKinds
    def upcast[F[+_]](f: F[J]): F[S] = f.asInstanceOf[F[S]]
  }
```

jrudolph: proposed here: https://github.com/akka/akka-http/pull/2963

raboof: Nice! That indeed removes this warning entirely, which is of course much nicer than silencing ;)

PR #2899, opened 2020-01-17
Add the remote address in parse errors when possible

johanandren:
Not entirely sure this is the right place to do it but couldn't figure out where else it would go.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpRequestParser.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpRequestParser.scala
@@ -16,6 +16,7 @@ import headers._
 import StatusCodes._
 import ParserOutput._
 import akka.annotation.InternalApi
+import akka.http.impl.engine.server.HttpAttributes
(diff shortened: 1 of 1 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpRequestParser.scala
```diff
@@ -220,6 +223,11 @@ private[http] final class HttpRequestParser(
         }
       } else failMessageStart("Request is missing required `Host` header")
 
+    private def remoteAddressStr: String =
+      inheritedAttributes.get[HttpAttributes.RemoteAddress].map(_.address) match {
+        case Some(addr) => s" from ${addr.getHostString}"
```
jrudolph: Let's also add the port if we are at it. Might be random most of the times but might still help associating it with client logs.

johanandren: Ah, started out with that in, but then thought it wouldn't be useful. I'll re-add

PR #2922, opened 2020-02-03
docs: add warning on usage on extractClientIP

winguse:
update the document as this can be a common error developers may make.


docs/src/main/paradox/routing-dsl/directives/misc-directives/extractClientIP.md
```diff
@@ -15,6 +15,10 @@ Provides the value of `X-Forwarded-For`, `Remote-Address`, or `X-Real-IP` header
 The akka-http server engine adds the `Remote-Address` header to every request automatically if the respective
 setting `akka.http.server.remote-address-header` is set to `on`. Per default it is set to `off`.
 
+@@@ warning
+Clients can send any values in these headers. When the value is an invalid IP address, then this extractor will return `RemoteAddress.Unknown`. If the client is not a tructed upstream, the IP address can be malicious and by pass your security rules.
```
raboof:
```suggestion
Clients can send any values in these headers. When the value is an invalid IP address, then this extractor will return `RemoteAddress.Unknown`. If the client is not a trusted upstream, the IP address can be malicious and by pass your security rules.
```

winguse:
@raboof thanks

PR #2926, opened 2020-02-04
core: identify Content-Type `charset` parameter even if not lower case

### LINKED ISSUE #2910: Incorrect Parsing of Content-Type Header
Parsing a `Content-Type` header seems to incorrectly detect the media type if the word "charset" is capitalized.

When parsing `text/html; charset=utf-8` everything works fine but for `text/html; Charset=utf-8` (capital C) I get the media type `text/html; Charset=utf-8` instead of just `text/html`; the resulting content type is then also `ContentTypeWithMissingCharset` instead of `ContentTypeWithCharset`

### How to Reproduce
```
scala> import akka.http.scaladsl.model.HttpHeader, akka.http.scaladsl.model.headers.`Content-Type`
import akka.http.scaladsl.model.HttpHeader
import akka.http.scaladsl.model.headers.Content$minusType

scala> val goodHeader = HttpHeader.parse("content-type", "text/html; charset=utf-8")
goodHeader: akka.http.scaladsl.model.HttpHeader.ParsingResult = Ok(Content-Type: text/html; charset=UTF-8,List())

scala> goodHeader match { case HttpHeader.ParsingResult.Ok(`Content-Type`(c), _) => c.mediaType }
res0: akka.http.scaladsl.model.MediaType = text/html

scala> val badHeader = HttpHeader.parse("content-type", "text/html; Charset=utf-8")
badHeader: akka.http.scaladsl.model.HttpHeader.ParsingResult = Ok(Content-Type: text/html; Charset=utf-8,List())

scala> badHeader match { case HttpHeader.ParsingResult.Ok(`Content-Type`(c), _) => c.mediaType }
res1: akka.http.scaladsl.model.MediaType = text/html; Charset=utf-8
```

Looking at the  [RFC](https://tools.ietf.org/html/rfc7231#section-3.1.1.1) it seems that `text/html; Charset=utf-8` should be a valid content type:

> For example, the following examples are all equivalent, but the first is preferred for consistency:
> text/html;charset=utf-8
> text/html;charset=UTF-8
> Text/HTML;Charset="utf-8"
> text/html; charset="utf-8"

Hope the issue explanation is clear enough, let me know if I can answer additional questions. I tried looking through the current issues but couldn't find a similar one.

jrudolph:
Fixes #2910

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/model/parser/ContentTypeHeader.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/model/parser/ContentTypeHeader.scala
@@ -9,6 +9,7 @@ import scala.collection.immutable.TreeMap
 
 import akka.parboiled2.Parser
 import akka.http.scaladsl.model._
+import akka.http.impl.util._
 
 private[parser] trait ContentTypeHeader { this: Parser with CommonRules with CommonActions =>
 
@@ -33,7 +34,7 @@ private[parser] trait ContentTypeHeader { this: Parser with CommonRules with Com
           case x: MediaType.WithOpenCharset if charset.isEmpty   => ContentType.WithMissingCharset(x)
         }
 
-      case Seq(("charset", value), tail @ _*) =>
+      case Seq((key, value), tail @ _*) if key.toRootLowerCase == "charset" =>
         contentType(main, sub, tail, Some(getCharset(value)), builder)
 
       case Seq(kvp, tail @ _*) =>
--- a/akka-http-core/src/main/scala/akka/http/impl/model/parser/ContentTypeHeader.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/model/parser/ContentTypeHeader.scala
@@ -9,7 +9,6 @@ import scala.collection.immutable.TreeMap
 
 import akka.parboiled2.Parser
 import akka.http.scaladsl.model._
-import akka.http.impl.util._
 
 private[parser] trait ContentTypeHeader { this: Parser with CommonRules with CommonActions =>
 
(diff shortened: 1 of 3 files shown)


akka-http-core/src/main/scala/akka/http/impl/model/parser/ContentTypeHeader.scala
```diff
@@ -33,7 +34,7 @@ private[parser] trait ContentTypeHeader { this: Parser with CommonRules with Com
           case x: MediaType.WithOpenCharset if charset.isEmpty   => ContentType.WithMissingCharset(x)
         }
 
-      case Seq(("charset", value), tail @ _*) =>
+      case Seq((key, value), tail @ _*) if key.toRootLowerCase == "charset" =>
```
raboof: `toRootLowerCase` takes code paths that take the locale into account (even though it's`Locale.ROOT`), so since we only want to know if it's equal to `"charset"` or not there's an opportunity for optimization here.

Hard to say if it's worth it here, but it feels like a relatively 'hot' part of the code? OTOH perhaps not compared to other handshaking/headerparsing.

akka-http-core/src/test/scala/akka/http/impl/model/parser/HttpHeaderSpec.scala
```diff
@@ -224,6 +224,8 @@ class HttpHeaderSpec extends AnyFreeSpec with Matchers {
         `Content-Type`(`application/json`)
       "Content-Type: text/plain; charset=utf8" =!=
         `Content-Type`(ContentType(`text/plain`, `UTF-8`)).renderedTo("text/plain; charset=UTF-8")
+      "Content-Type: text/plain; Charset=utf8" =!=
+        `Content-Type`(ContentType(`text/plain`, `UTF-8`)).renderedTo("text/plain; charset=UTF-8")
```
raboof: :+1:

akka-http-core/src/main/scala/akka/http/impl/model/parser/ContentTypeHeader.scala
jrudolph: Great point. I didn't think too much about it. The reason I used it, is that there might be some potential for ambiguities where an attacker might use weird UTF-8 characters that might lowercase to ascii chars (but only in some servers). I guess we might want to have a fast case-insensitive ASCII char comparison.

jrudolph: Now replaced with a custom matching method for ASCII chars.

Issue #1354, opened 2017-08-15
Support Cookie SameSite attribute and maybe allow specifying custom cookie attributes


jrudolph:
It's currently impossible to model a `Set-Cookie` header with attributes unknown to akka-http like this:

```
Set-Cookie: key=value; HttpOnly; SameSite=strict
```

We should allow specifying custom attributes and/or add popular ones missing from the current model.

The current workaround is to use a `RawHeader` in such a case.

Prompted by https://groups.google.com/d/topic/akka-user/XbAoF_EQAHI/discussion.

myDisconnect:
As Chrome 80 is inevitably coming, https://www.chromestatus.com/feature/5088147346030592, any plans for this to be updated?

alexglasser:
It's my understanding that the only workaround for this today is to use `RawHeader` in place of `Set-Cookie`.

Should I go ahead with this change in my codebase, or will this ticket be prioritized before Chrome begins rejecting cookies without the `SameSite` attribute? Note that this also impacts Safari running on macOS 10.14.

ktulinger:
This is gettin real in few months. Any plan?

jrudolph:
Thanks for the reminder, @ktulinger. I'm putting it tentatively on the roadmap for the upcoming 10.2.0.

PR #2928, opened 2020-02-04
Add SameSite attribute to Cookies

marcospereira:
References #1354.

## Status

This is a work in progress since I didn't handle any bin-compat issues yet.


akka-http-core/src/main/mima-filters/10.1.11.backwards.excludes/1354-add-HttpCookie-SameSite-attribute.backwards.excludes
```diff
@@ -0,0 +1,11 @@
+# Add SameSite attribute to HttpCookie
+
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.javadsl.model.headers.HttpCookie.getSameSite")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.javadsl.model.headers.HttpCookie.withSameSite")
+
+ProblemFilters.exclude[IncompatibleSignatureProblem]("akka.http.scaladsl.model.headers.HttpCookie.fromPair$default$8")
```
jrudolph: The problem is that we cannot change these because it's a genuine binary incompatibility. We need to convert the `case class` to a regular class and keep providing those methods as deprecated until the next release cycle.

akka-http-core/src/test/scala/akka/http/javadsl/model/headers/HttpCookieSpec.scala
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+package akka.http.javadsl.model.headers
+
+import java.util.{ Optional, OptionalLong }
+
+import org.scalatest.freespec.AnyFreeSpec
+import org.scalatest.matchers.must.Matchers
+
+class HttpCookieSpec extends AnyFreeSpec with Matchers {
```
jrudolph: Makes sense to create those here. Test coverage of Java API is not too great so far.

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/HttpCookie.scala
```diff
@@ -125,8 +134,9 @@ object HttpCookie {
     path:      Option[String]   = None,
     secure:    Boolean          = false,
     httpOnly:  Boolean          = false,
+    sameSite:  Option[SameSite] = None,
```
jrudolph: If we move it to the last position, it will probably keep binary compat for the default value. We'll still need a deprecated version of the old one. Or we just deprecate the version with the default arguments, saying that people should use `fromPair(pair).withABC.withDEF`.

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/HttpCookie.scala
```diff
@@ -8,10 +8,13 @@ import akka.http.impl.model.parser.CharacterClasses
 import akka.http.javadsl.model.headers
 import akka.parboiled2.CharPredicate
 import java.util.{ Optional, OptionalLong }
+
 import akka.http.scaladsl.model.DateTime
 import akka.http.impl.util._
 import akka.http.javadsl.{ model => jm }
 import akka.http.impl.util.JavaMapping.Implicits._
+import akka.http.javadsl.model.headers.SameSite
```
jrudolph: So far, we created separate models for javadsl and scaladsl. It feels weird that you would have to import classes from `javadsl` to use it from Scala. That will probably mean duplicating the enums for Scala and Java and converting between them.

akka-http-core/src/main/java/akka/http/javadsl/model/headers/SameSite.java
```diff
@@ -0,0 +1,12 @@
...
+ * Copyright (C) 2019 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.http.javadsl.model.headers;
+
+/**
+ * The Cookie SameSite attribute as defined by <a href="https://tools.ietf.org/html/draft-ietf-httpbis-cookie-same-site-00">RFC6265bis</a>.
+ */
+public enum SameSite {
+    Strict, Lax, None
```
jrudolph: Having an enum is nice. Do we need to provide an escape hatch for the future when other things might be in there?

Do we need `None` since it will be packed into an Option? What would `Some(None)` mean as a value opposed to `None`?

akka-http-core/src/test/scala/akka/http/scaladsl/model/headers/HttpCookieTest.scala
```diff
@@ -0,0 +1,79 @@
...
+      }
+      "not include HttpOnly if cookie is not http only" in {
+        val cookie = HttpCookie(name = "cookieName", value = "cookieValue", httpOnly = false)
+        val rendering = new StringRendering()
+        cookie.render(rendering).get must not include ("; HttpOnly")
+      }
+      "not include SameSite if not specified" in {
+        val cookie = HttpCookie(name = "cookieName", value = "cookieValue")
+        val rendering = new StringRendering()
+        cookie.render(rendering).get must not include ("; SameSite=None")
```
jrudolph: Since we had it at another place: we need to verify if parsing should be case-insensitive and if yes add tests that make sure it works for whatever casing.

akka-http-core/src/main/java/akka/http/javadsl/model/headers/SameSite.java
jrudolph: Looking into the spec, the [grammar](https://tools.ietf.org/html/draft-ietf-httpbis-cookie-same-site-00#section-3.1) only seems to allow `Strict` and `Lax`.

IIUC, the part about a `None` value is about how a browser should store such an attribute which needs to store the fact that the attribute was missing. We already have the `Option` type for that, so I guess `None` wouldn't be needed as part of the enum here. Or am I reading things wrong?

akka-http-core/src/main/java/akka/http/javadsl/model/headers/HttpCookie.java
```diff
@@ -91,6 +118,11 @@ public static HttpCookie create(
      */
     public abstract HttpCookie withHttpOnly(boolean httpOnly);
 
+    /**
+     * Returns a copy of this HttpCookie instance with the given {@link SameSite} set.
+     */
+    public abstract HttpCookie withSameSite(SameSite sameSite);
```
jrudolph: What about if you want to unset the attribute?

We also need a getter that converts back to `Optional`.

raboof:
Great comments! Agreed on all points I think...

akka-http-core/src/main/java/akka/http/javadsl/model/headers/HttpCookie.java
marcospereira: This is now overloaded to also accept an `Optional<SameSite>`.

akka-http-core/src/main/mima-filters/10.1.11.backwards.excludes/1354-add-HttpCookie-SameSite-attribute.backwards.excludes
```diff
@@ -0,0 +1,5 @@
+# Add SameSite attribute to HttpCookie
+
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.javadsl.model.headers.HttpCookie.getSameSite")
```
marcospereira: @jrudolph these are the only places where bin-compat is broken now.

akka-http-core/src/test/scala/akka/http/scaladsl/model/headers/HttpCookieTest.scala
```diff
@@ -0,0 +1,79 @@
...
+
+package akka.http.scaladsl.model.headers
+
+import akka.http.impl.util.StringRendering
+import akka.http.javadsl.model.headers.SameSite
+import akka.http.scaladsl.model.DateTime
+import org.scalatest.freespec.AnyFreeSpec
+import org.scalatest.matchers.must.Matchers
+
+class HttpCookieTest extends AnyFreeSpec with Matchers {
```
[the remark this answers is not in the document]
marcospereira: Thanks for pointing that out. Those tests revealed that a lot was missing in this PR. I'm keeping this anyway, but focused on unit tests. Let me know if you think we should move these somewhere else too.

jrudolph:
> The reason there is a `SameSite=None` is that it is a new value supported by [Chrome](https://blog.chromium.org/2019/10/developers-get-ready-for-new.html), Firefox and others. Not adding `SameSite` attribute will then default to `Lax`. In other words, `withSameSite(None)` is different than`withSameSite(SameSite.None)`.

Thanks for that pointer. I agree we'll need it then.

> Edit: It is a little unclear to me if I should deprecate the methods added for the sake of binary compatibility, or if I should make them `private[http]`.

Deprecate on the Java side, maybe both `deprecate` and `private[http]` on the Scala side to make sure we get a warning for unintended internal usages.

akka-http-core/src/main/mima-filters/10.1.11.backwards.excludes/1354-add-HttpCookie-SameSite-attribute.backwards.excludes
raboof: Right. So the 'excuse' here is that we don't expect users to extend those classes, even though we didn't explicitly mark them `@DoNotInherit`. That seems acceptable to me, but perhaps we should add those `@DoNotInherit` marks now?

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/HttpCookie.scala
```diff
@@ -113,10 +188,54 @@ final case class HttpCookie(
...
+  /** Java API */
+  def withSameSite(sameSite: Optional[jm.headers.SameSite]): headers.HttpCookie = new HttpCookie(name, value, expires, maxAge, domain, path, secure, httpOnly, extension, sameSite.asScala.map(_.asScala()))
+  /** Java API */
   def withExtension(extension: String): headers.HttpCookie = copy(extension = Some(extension))
 }
 
 object HttpCookie {
+
+  @deprecated("for binary compatibility", since = "10.2.0")
+  def apply(
```
raboof: To reduce the source-incompatibility (now requiring 'new'), I think we should duplicate this apply method into 2: one deprecated one without default values and without sameSite, and one with default values and with sameSite. Would that work?

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/HttpCookie.scala
```diff
@@ -140,3 +259,29 @@ object HttpCookie {
...
+object SameSite {
+  def apply(s: String): Option[SameSite] = {
+    if ("Lax".equalsIgnoreCase(s)) Some(Lax)
+    else if ("Strict".equalsIgnoreCase(s)) Some(Strict)
+    else if ("None".equalsIgnoreCase(s)) Some(None)
+    else Option.empty
+  }
+  case object Strict extends SameSite
+  case object Lax extends SameSite
+  case object None extends SameSite
```
raboof: Should we document here why `None` is useful and different from not having a SameSite parameter at all?

marcospereira:
@raboof addressed your comments here. :-)

akka-http-core/src/main/java/akka/http/javadsl/model/headers/SameSite.java
```diff
@@ -0,0 +1,25 @@
...
+package akka.http.javadsl.model.headers;
+
+/**
+ * The Cookie SameSite attribute as defined by <a href="https://tools.ietf.org/html/draft-ietf-httpbis-cookie-same-site-00">RFC6265bis</a>
+ * and <a href="https://tools.ietf.org/html/draft-west-cookie-incrementalism-00">Incrementally Better Cookies</a>.
+ */
+public enum SameSite {
+    Strict,
+    Lax,
+    // SameSite.None is different from not adding the SameSite attribute in a cookie.
```
jrudolph: Should we make this a proper javadoc comment? Could you also add the link to the Chrome post here if that's the best reference we currently have?

akka-http-core/src/main/java/akka/http/javadsl/model/headers/SameSite.java
```diff
@@ -0,0 +1,25 @@
...
+    // - Cookies without a SameSite attribute will be treated as SameSite=Lax.
+    // - Cookies for cross-site usage must specify `SameSite=None; Secure` to enable inclusion in third party
+    //   context. We are not enforcing `; Secure` when `SameSite=None`, but users should.
```
jrudolph: What we have done before is to add a `require()` to a model class if there are certain constraints. Not sure if we would do it here, though, at least not as long there's no proper spec.

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/HttpCookie.scala
```diff
@@ -55,7 +57,7 @@ object HttpCookiePair {
  * for a full definition of the http cookie header fields, see
  * http://tools.ietf.org/html/rfc6265
  */
-final case class HttpCookie(
+final class HttpCookie(
```
jrudolph: Let's make the new constructor `private[http]` as well.

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/HttpCookie.scala
```diff
@@ -64,7 +66,65 @@ final case class HttpCookie(
...
+    maxAge:    Option[Long],
+    domain:    Option[String],
+    path:      Option[String],
+    secure:    Boolean,
+    httpOnly:  Boolean,
+    extension: Option[String]) = this(name, value, expires, maxAge, domain, path, secure, httpOnly, extension, None)
+
+  private[http] def copy(sameSite: Option[SameSite]) = new HttpCookie(name, value, expires, maxAge, domain, path, secure, httpOnly, extension, sameSite)
+
+  def copy(
```
jrudolph: Going forward, we would probably rather provide `withXYZ` methods instead of a copy method that is hard to evolve. Maybe we should do this right now and deprecate the copy as well?

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/HttpCookie.scala
```diff
@@ -87,9 +147,24 @@ final case class HttpCookie(
...
+    if (sameSite.isDefined) r ~~ "; SameSite=" ~~ sameSite.get
     r
   }
 
+  override def name(): String = this.name
+  override def value(): String = this.value
+  override def secure(): Boolean = this.secure
+  override def httpOnly(): Boolean = this.httpOnly
+
+  private[http] def expires(): Option[DateTime] = this.expires
```
jrudolph: Why are these `private[http]` now? Also we should remove the extra parentheses (future Scala versions might enforce them for usages).

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/HttpCookie.scala
```diff
@@ -140,3 +272,40 @@ object HttpCookie {
...
+    if ("Lax".equalsIgnoreCase(s)) Some(Lax)
+    else if ("Strict".equalsIgnoreCase(s)) Some(Strict)
+    else if ("None".equalsIgnoreCase(s)) Some(None)
+    else Option.empty
+  }
+
+  case object Strict extends SameSite
+  case object Lax extends SameSite
+
+  // SameSite.None is different from not adding the SameSite attribute in a cookie.
```
jrudolph: Can be scaladoc and add link as mentioned above for the java version.

akka-http-core/src/test/scala/akka/http/scaladsl/model/headers/HttpCookieTest.scala
```diff
@@ -0,0 +1,44 @@
...
+      "return Option empty when value is invalid" in {
+        SameSite("Wrong") must be(None)
+      }
+      "be case insensitive" in {
+        SameSite("sTRIct") must be(Some(SameSite.Strict))
+        SameSite("lAX") must be(Some(SameSite.Lax))
+        SameSite("NONE") must be(Some(SameSite.None))
+      }
+    }
+    "SameSite conversation to Java" - {
```
jrudolph:
```suggestion
    "SameSite conversion to Java" - {
```

akka-http-core/src/main/java/akka/http/javadsl/model/headers/SameSite.java
marcospereira: I'm not sure we should be the ones explaining how `SameSite=None` works, and adding javadoc to `None` implies that we should do the same for `Lax` and `Strict`(?). That is why it is a code comment instead. I briefly discussed that with @raboof as well and we decided that a code comment for maintainers would suffice. 

Regarding the reference, I've added this: https://github.com/akka/akka-http/pull/2928/files#diff-f4b4205d613cfb57115bc38c5c78060eR9

akka-http-core/src/main/java/akka/http/javadsl/model/headers/SameSite.java
marcospereira: There is a draft spec though: https://tools.ietf.org/html/draft-west-cookie-incrementalism-00

It says _cookies that explicitly assert "SameSite=None" in order to enable cross-site delivery should also be marked as "Secure"_.

With the current state of this PR, it is up to users to correctly call `withSameSite(SameSite.None)` *and* `withSecure(true)` for that cookie. Using a `require` seems like the correct approach to me.

WDYT?

marcospereira: I also wonder if we should set `secure=true` automatically when calling `withSameSite(SameSite.None)`. Maybe a combination of that and the `require` to handle the case where users (accidentally, I hope) call `withSameSite(SameSite.None).withSecure(false)`.

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/HttpCookie.scala
marcospereira: Hum, yeah, they don't need to be private. I guess I was following too much on the Java API side. I guess I can declare them as `val`s and not have those methods at all.

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/HttpCookie.scala
marcospereira: I will mark this as resolved and, when we finish the conversation there, I will do the same for both APIs. ;-)

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/HttpCookie.scala
marcospereira: Making it `private[http]` here increases the number of binary compatibility issues (since the constructor deprecated constructor does not have the default values). Source compatibility will be good though.

WDYT?

akka-http-core/src/main/java/akka/http/javadsl/model/headers/SameSite.java
jrudolph: Ah, missed that. I agree both suggestions makes sense. We should have a test that makes sure that this `require` fails the response in a reasonable way.

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/HttpCookie.scala
jrudolph: I think it should be private. Couldn't we just move the default params to the deprecated constructor? In any case, we want people to use the apply method in the future.

akka-http-core/src/main/java/akka/http/javadsl/model/headers/SameSite.java
marcospereira: I realized that it won't be possible to do this check. Both cookies below are valid:

```
Set-Cookie: name=123; Secure; SameSite=None
```

```
Set-Cookie: name=123; SameSite=None; Secure
```

But the second one sets `SameSite` first, and the parser will then try to create a `HttpCookie` with that, but fails because `secure=false` since it was not parsed yet. It feels to me that using a `require` is good when you are checking an attribute that does not relate to another one.

WDYT?

raboof: Sounds fine to me to skip it then.

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/HttpCookie.scala
```diff
@@ -55,16 +57,75 @@ object HttpCookiePair {
...
+  val maxAge:    Option[Long],
+  val domain:    Option[String],
+  val path:      Option[String],
+  secure:        Boolean,
+  httpOnly:      Boolean,
+  val extension: Option[String],
+  val sameSite:  Option[SameSite]) extends jm.headers.HttpCookie with ToStringRenderable with Product with Serializable with Equals {
+
+  @deprecated("for binary compatibility", since = "10.2.0")
+  def this(
```
raboof: we can make this private[headers]

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/HttpCookie.scala
```diff
@@ -113,10 +182,67 @@ final case class HttpCookie(
...
+    domain:    Option[String]   = None,
+    path:      Option[String]   = None,
+    secure:    Boolean          = false,
+    httpOnly:  Boolean          = false,
+    extension: Option[String]   = None,
+    sameSite:  Option[SameSite] = None
+  ) = new HttpCookie(name, value, expires, maxAge, domain, path, secure, httpOnly, extension, sameSite)
+
+  @deprecated("for binary compatibility", since = "10.2.0")
+  def apply(
```
raboof: (here too)

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/HttpCookie.scala
```diff
@@ -113,10 +182,67 @@ final case class HttpCookie(
...
+    maxAge:    Option[Long],
+    domain:    Option[String],
+    path:      Option[String],
+    secure:    Boolean,
+    httpOnly:  Boolean,
+    extension: Option[String]
+  ) = new HttpCookie(name, value, expires, maxAge, domain, path, secure, httpOnly, extension, None)
+
+  @deprecated("for binary compatibility", since = "10.2.0")
+  def unapply(cookie: HttpCookie) = Option((
```
raboof: (makes sense to me to deprecate this one, but let's indeed not make this one private to keep things source-compatible)

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/HttpCookie.scala
```diff
@@ -113,10 +182,67 @@ final case class HttpCookie(
...
+    maxAge:    Option[Long]     = None,
+    domain:    Option[String]   = None,
+    path:      Option[String]   = None,
+    secure:    Boolean          = false,
+    httpOnly:  Boolean          = false,
+    extension: Option[String]   = None,
+    sameSite:  Option[SameSite] = None): HttpCookie =
+    new HttpCookie(pair.name, pair.value, expires, maxAge, domain, path, secure, httpOnly, extension, sameSite)
+
+  @deprecated("Use createFromPair instead", "10.2.0")
```
raboof: (this could become `private[headers]` as well again)

marcospereira: This was part of `HttpCookie` object public API. Not sure if users should be using it like that, but I would prefer to only have it deprecated.

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/HttpCookie.scala
marcospereira: Making this constructor `private[headers]` (and also the deprecated `apply` method) will both have an impact on binary compatibility. I can do that and configure the filters, but want to check with your first before doing that.

raboof: Ah, you're right, because of of the default parameters and the change from `case class` to `class`... In that case I agree it makes sense to keep the constructor with default parameters separate like this, but then perhaps it should be public and not deprecated?

marcospereira: I wonder if we can instead target source compatibility here instead of binary compatibility. If so, we can have a single constructor with default params and accepting `Option[SameSite]` (defaulting to `None`).

The alternative is to either keep this constructor without the deprecation (which does not add value to the API but creates some confusion if I want to create a cookie with `SameSite`) or keep it deprecated and go through the cycle of removing it later and adding the default value in the main constructor.

WDYT?

marcospereira: By the way, going for source compatibility would require the following new filters:

```
ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.scaladsl.model.headers.HttpCookie.<init>$default$9")
ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.scaladsl.model.headers.HttpCookie.<init>$default$8")
ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.scaladsl.model.headers.HttpCookie.<init>$default$7")
ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.scaladsl.model.headers.HttpCookie.<init>$default$6")
ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.scaladsl.model.headers.HttpCookie.<init>$default$5")
ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.scaladsl.model.headers.HttpCookie.<init>$default$4")
ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.scaladsl.model.headers.HttpCookie.<init>$default$3")
ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.scaladsl.model.headers.HttpCookie.this")
```

Just to make the impact clearer.

raboof: I don't think we can afford to skip binary compatibility here - any library that uses the current HttpCookie constructor is using those methods for the default values, so if we'd break binary compatibility all such libraries would start throwing exceptions.

> The alternative is to either keep this constructor without the deprecation (which does not add value to the API but creates some confusion if I want to create a cookie with `SameSite`)

I agree it is a bit confusing that you'd have to use `withSameSite` for sameSite instead of using this constructor, but for all other uses it adds value by providing a way to create a HttpCookie without providing values for all fields.

> keep it deprecated and go through the cycle of removing it later and adding the default value in the main constructor

Unforunately you can't have 2 constructors that both have default values

akka-http-core/src/main/java/akka/http/javadsl/model/headers/SameSite.java
jrudolph: Ah, yes, if it's built incrementally, then `require` won't work, so the only thing we could do is set `Secure` automatically when we see `SameSite=None` and only fail when trying to remove `Secure` when `SameSite=None` is set as well. But not sure if we should go for that right now?

akka-http-core/src/main/java/akka/http/javadsl/model/headers/SameSite.java
jrudolph: I see, fine with me.

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/HttpCookie.scala
jrudolph: Do we have a decision on that one? @raboof, wdyt?

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/HttpCookie.scala
jrudolph: I think we have two options here:
 * remove the default parameter values here, add `private[headers`, and rename the new one to `fromPair` so it's also source compatible
 * keep this one just deprecated and create a completely different method that only takes a `HttpCookiePair` and suggest to use the `withXYZ` methods to change other fields

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/HttpCookie.scala
raboof: In #2938 I kept one but made it `private[package]` - that way it's nice for implementing those `withXyz` methods but we don't need to keep it binary compatible going forward.

raboof: though I don't mind having it public much either, even if it leaves behind some cruft each time we add a field

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/HttpCookie.scala
marcospereira: > remove the default parameter values here, add private[headers, and rename the new one to fromPair so it's also source compatible

Maybe this is slightly different from what @raboof explained here: https://github.com/akka/akka-http/pull/2928#discussion_r380949103

(where source compatibility is not tolerated)

> keep this one just deprecated and create a completely different method that only takes a `HttpCookiePair` and suggest to use the `withXYZ` methods to change other fields

Yes, I think we can create `def apply(pair: HttpCookiePair)` and add the missing `with*`  methods. I'm just unsure how this would be a better API, though. As a user, this:

```scala
val cookie = HttpCookie(pair, domain = Some("akka.io"), secure = true, httpOnly = true)
```

Looks more convenient than:

```scala
val cookie = HttpCookie(pair).withDomain("akka.io").withSecure(true).withHttpOnly(true)
```

I consider the former easier to read as well, and it makes fewer allocations.

But, anyhow, _overall_ API consistent wins here, therefore, if Akka HTTP is adopting the later style, I agree it would be better to have the same for `HttpCookie`.

WDYT?

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/HttpCookie.scala
marcospereira: > In that case I agree it makes sense to keep the constructor with default parameters separate like this, but then perhaps it should be public and not deprecated?

@raboof I went this route. This has a little conflict with what @jrudolph suggested before (make the constructor `private[headers]`).

marcospereira:
@jrudolph @raboof I make some small changes, but there are some open conversations that we need to resolve before moving forward:

1. https://github.com/akka/akka-http/pull/2928#discussion_r378130335
2. https://github.com/akka/akka-http/pull/2928#discussion_r379487265
3. https://github.com/akka/akka-http/pull/2928#discussion_r379488623

Linking them to make it easier to navigate since there are many other threads here.

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/HttpCookie.scala
jrudolph: > In #2938 I kept one but made it `private[package]` - that way it's nice for implementing those `withXyz` methods but we don't need to keep it binary compatible going forward.

Let's go with this way.

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/HttpCookie.scala
raboof: (cannot make this private[...] as that changes the static forwarder creation)

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/HttpCookie.scala
```diff
@@ -100,23 +178,71 @@ final case class HttpCookie(
...
-  /** Java API */
-  def withHttpOnly(httpOnly: Boolean): headers.HttpCookie = copy(httpOnly = httpOnly)
+  def withExpires(dateTime: jm.DateTime): HttpCookie = copy(expires = Some(dateTime.asScala))
+
+  def withDomain(domain: String): HttpCookie = copy(domain = Some(domain))
+  def withPath(path: String): HttpCookie = copy(path = Some(path))
+  def withMaxAge(maxAge: Long): HttpCookie = copy(maxAge = Some(maxAge))
+  def withSecure(secure: Boolean): HttpCookie = copy(secure = secure)
+  def withHttpOnly(httpOnly: Boolean): HttpCookie = copy(httpOnly = httpOnly)
+  def withSameSite(sameSite: jm.headers.SameSite): HttpCookie = copy(sameSite = Option(sameSite.asScala()))
```
jrudolph: Ok, this one should also work for the Scala one, I guess.

jrudolph: Ah, no it doesn't, so should we add another overload for the Scala API (without the Option)?

akka-http-core/src/test/scala/akka/http/scaladsl/model/headers/HttpCookieTest.scala
```diff
@@ -0,0 +1,44 @@
...
+        SameSite("Lax") must be(Some(SameSite.Lax))
+      }
+      "None" in {
+        SameSite("None") must be(Some(SameSite.None))
+      }
+      "return Option empty when value is invalid" in {
+        SameSite("Wrong") must be(None)
+      }
+      "be case insensitive" in {
+        SameSite("sTRIct") must be(Some(SameSite.Strict))
```
jrudolph: Can we add these to HttpHeaderSpec and then remove the file?

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/HttpCookie.scala
```diff
@@ -100,23 +178,71 @@ final case class HttpCookie(
...
+    maxAge:    Option[Long]     = None,
+    domain:    Option[String]   = None,
+    path:      Option[String]   = None,
+    secure:    Boolean          = false,
+    httpOnly:  Boolean          = false,
+    extension: Option[String]   = None,
+    sameSite:  Option[SameSite] = None
+  ) = new HttpCookie(name, value, expires, maxAge, domain, path, secure, httpOnly, extension, sameSite)
+
+  @deprecated("for binary compatibility", since = "10.2.0")
```
jrudolph: I guess we don't expect this method to be selected? Otherwise, we should improve the deprecation hint.

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/HttpCookie.scala
```diff
@@ -100,23 +178,71 @@ final case class HttpCookie(
...
+    expires:   Option[DateTime],
+    maxAge:    Option[Long],
+    domain:    Option[String],
+    path:      Option[String],
+    secure:    Boolean,
+    httpOnly:  Boolean,
+    extension: Option[String]
+  ) = new HttpCookie(name, value, expires, maxAge, domain, path, secure, httpOnly, extension, None)
+
+  @deprecated("for binary compatibility", since = "10.2.0")
```
jrudolph: Let's add better deprecation hints where the methods are still public and the hints likely to be shown.

```suggestion
  @deprecated("Pattern matching on HttpCookie is deprecated because of the big number of fields and potential future compatibility hazards. Please use other means to check the fields.", since = "10.2.0")
```

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/HttpCookie.scala
```diff
@@ -100,23 +178,71 @@ final case class HttpCookie(
...
+    cookie.expires,
+    cookie.maxAge,
+    cookie.domain,
+    cookie.path,
+    cookie.secure(),
+    cookie.httpOnly(),
+    cookie.extension
+  ))
+
+  @deprecated("Use HttpCookiePair.toCookie and withXxx methods instead", "10.2.0")
```
jrudolph: :+1:

jrudolph:
Would be nice to put this over the finish line.

akka-http-core/src/main/scala/akka/http/impl/model/parser/CommonRules.scala
```diff
@@ -288,20 +288,28 @@ private[parser] trait CommonRules { this: Parser with StringBuilding =>
...
   def `path-value` = rule {
     capture(zeroOrMore(`av-octet`)) ~ OWS
   }
 
+  def `same-site-av` = rule {
+    ignoreCase("samesite=") ~ OWS ~ `same-site-value` ~> { (c: HttpCookie, sameSiteValue: String) => c.withSameSite(sameSite = SameSite(sameSiteValue)) }
+  }
+
+  def `same-site-value` = rule {
+    capture(ignoreCase("lax") | ignoreCase("strict") | ignoreCase("none")) ~ OWS
```
raboof: This means `samesite=wrong` will fail to parse. Do we want that?

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/HttpCookie.scala
jrudolph: Do we still want to keep this public and un-deprecated? I think we should at least deprecate it.

```suggestion
  @deprecated("Use apply method, HttpCookie(...), instead of new HttpCookie(...)", since = "10.2.0")
  def this(
```

akka-http-core/src/main/scala/akka/http/impl/model/parser/CommonRules.scala
jrudolph: Would be ok I think for the first version. If people run into that problem often enough, we can think about providing alternatives.

jrudolph: Also, since it is the server-side sending those, only our client-side would be affected where cookie usage currently isn't as important.

raboof: Ah, you're right. Let's keep it like this then. I'll update the tests - do you think we should call this out in the release/migration notes?

jrudolph: Wouldn't hurt to mention it. It's not really worse than before, though.

raboof: Good point, let's leave it like this.

raboof:
> Anything you want to add still, @raboof?

OK AFAICS now!

akka-http-core/src/main/scala/akka/http/scaladsl/model/ErrorInfo.scala
```diff
@@ -47,6 +47,8 @@ final class ErrorInfo(
 
   /** INTERNAL API */
   @InternalApi private[akka] def this(summary: String, detail: String) = this(summary, detail, "")
+
+  override def toString(): String = s"ErrorInfo($summary, $detail, $errorHeaderName)"
```
jrudolph: Do we want the detail in the `toString`? I'd say yes, but people might think otherwise, given, that the feature is there to prevent accidental leakage of information. Of course, leakage to the client is still prevented but details might leak to logs.

raboof:
> When the deprecation for this is in it's good to be merged from my POV.

Good one, done. Also removed the new `apply` method keeping the old one and pointing people to 'withSameSite' instead.

jrudolph:
:fireworks:

jrudolph:
We will have to decide if that needs backporting but maybe let's wait until asks for it.

PR #2938, opened 2020-02-05
Support for request/response message attributes

raboof:
In this draft, attributes are identified by AttributeKey instance. Explicitly
naming the keys doesn't seem necessary, as they are 'internal' and never
serialized.

Refs #1500, contrast with #2872

If we decide to go with this approach we should also add directives to extract
attributes by key.

### DIFF
--- a/akka-http-core/src/main/java/akka/http/javadsl/model/HttpMessage.java
+++ b/akka-http-core/src/main/java/akka/http/javadsl/model/HttpMessage.java
@@ -140,6 +140,11 @@ public interface HttpMessage {
         Self removeHeader(String headerName);
 
         /**
+         * Returns a copy of this message with the attribute with this key (if any) removed.
+         */
+        Self removeAttribute(AttributeKey<?> key);
+
+        /**
          * Returns a copy of this message with a new entity.
          */
         Self withEntity(String string);
--- a/akka-http-core/src/main/java/akka/http/javadsl/model/HttpMessage.java
+++ b/akka-http-core/src/main/java/akka/http/javadsl/model/HttpMessage.java
@@ -66,6 +66,12 @@ public interface HttpMessage {
     <T extends HttpHeader> Iterable<T> getHeaders(Class<T> headerClass);
 
     /**
+     * Try to find the attribute for the given key and return
+     * Optional.of(attribute), otherwise this method returns an empty Optional.
+     */
+    <T> Optional<T> getAttribute(AttributeKey<T> key);
+
+    /**
      * The entity of this message.
      */
     ResponseEntity entity();
@@ -120,6 +126,8 @@ public interface HttpMessage {
          * Returns a copy of this message with new headers.
          */
         Self withHeaders(Iterable<HttpHeader> headers);
+
+        <T> Self addAttribute(AttributeKey<T> key, T value);
 
         /**
          * Returns a copy of this message with the given http credential header added to the list of headers.
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
@@ -42,6 +42,7 @@ sealed trait HttpMessage extends jm.HttpMessage {
   def isResponse: Boolean
 
   def headers: immutable.Seq[HttpHeader]
+  def attributes: Map[AttributeKey[Any], Any]
   def entity: ResponseEntity
   def protocol: HttpProtocol
 
@@ -100,6 +101,9 @@ sealed trait HttpMessage extends jm.HttpMessage {
   def withDefaultHeaders(firstHeader: HttpHeader, otherHeaders: HttpHeader*): Self =
     withDefaultHeaders(firstHeader +: otherHeaders.toList)
 
+  /** Returns a copy of this message with the attributes set to the given ones. */
+  def withAttributes(headers: Map[AttributeKey[Any], Any]): Self
+
   /** Returns a copy of this message with the entity set to the given one. */
   def withEntity(entity: MessageEntity): Self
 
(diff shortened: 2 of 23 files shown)


akka-http-core/src/main/java/akka/http/javadsl/model/HttpMessage.java
```diff
@@ -121,6 +127,8 @@
          */
         Self withHeaders(Iterable<HttpHeader> headers);
 
+        <T> Self addAttribute(AttributeKey<T> key, T value);
```
marcospereira: Should we also have a way to remove attributes?

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
```diff
@@ -335,12 +360,21 @@ final class HttpRequest(
 
   /* Manual Case Class things, to easen bin-compat */
 
+  @deprecated("Kept for binary compatibility", "10.0.2")
```
marcospereira:
```suggestion
  @deprecated("Kept for binary compatibility", "10.2.0")
```

raboof:
> what's still missing? I guess a bit of documentation?

Yes, and I think we'd also want directives to add/extract attributes like we have for headers - though perhaps that could be in an separate PR

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
```diff
@@ -284,6 +302,10 @@ final class HttpRequest(
   override def isRequest = true
   override def isResponse = false
 
+  @deprecated("for backwards compatibility", "10.2.0")
```
ennru:
```suggestion
  @deprecated("use constructor with `attributes` instead", "10.2.0")
```
or should it even be `private[akka]`?

docs/src/main/paradox/common/http-model.md
```diff
@@ -356,6 +356,18 @@ Implement @apidoc[ModeledCustomHeader] and @java[@javadoc[ModeledCustomHeaderFac
 able to use the convenience methods that allow parsing the custom user-defined header from @apidoc[HttpHeader].
 @@@
 
+## Attributes
+
+Sometimes it can be useful to keep track of some information associated with a request without
```
ennru: Maybe be clear about that attributes are not part of the HTTP communication?

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
```diff
@@ -353,10 +387,11 @@ final class HttpRequest(
   }
 
   override def equals(obj: scala.Any): Boolean = obj match {
-    case HttpRequest(_method, _uri, _headers, _entity, _protocol) =>
+    case request @ HttpRequest(_method, _uri, _headers, _entity, _protocol) =>
```
raboof: We cannot add 'attributes' to the unapply in a binary-compatible (neither source-compatible) way, so we have no choice but to have it only as a field. Matching on so many attributes becomes unwieldy anyway, so not too bad.

jrudolph: I totally agree.

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
```diff
@@ -42,6 +42,7 @@ sealed trait HttpMessage extends jm.HttpMessage {
   def isResponse: Boolean
 
   def headers: immutable.Seq[HttpHeader]
+  def attributes: Map[AttributeKey[Any], Any]
```
jrudolph: Why not `AttributeKey[_]` (not that we could properly model that in a map but it would at least be a bit more accurate)?

jrudolph: I.e. we could probably remove part of the `instanceOf`s below.

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
```diff
@@ -196,6 +210,11 @@ sealed trait HttpMessage extends jm.HttpMessage {
     import JavaMapping.Implicits._
     withHeaders(headers.asScala.toVector.map(_.asScala))
   }
+  /** Java API */
+  def getAttribute[T](attributeKey: jm.AttributeKey[T]): Optional[T] = {
```
jrudolph: Extra braces.

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
```diff
@@ -284,6 +304,10 @@ final class HttpRequest(
   override def isRequest = true
   override def isResponse = false
 
+  @deprecated("use the constructor that includes an attributes parameter instead", "10.2.0")
```
jrudolph: I guess, we should propose to start with a simple request/response and use the `withX` more. That will also be the more compatible way going forward.

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
```diff
@@ -335,12 +362,21 @@ final class HttpRequest(
...
 
+  @deprecated("use the method that includes an attributes parameter instead", "10.2.0")
+  private[model] def copy(
+    method:   HttpMethod,
+    uri:      Uri,
+    headers:  immutable.Seq[HttpHeader],
+    entity:   RequestEntity,
+    protocol: HttpProtocol) = new HttpRequest(method, uri, headers, attributes, entity, protocol)
+
   def copy(
```
jrudolph: We should make copy private and people should use the `withXYZ` methods which are better to read (imo) and easier to stay compatible with.

jrudolph:
:fireworks:

PR #2942, opened 2020-02-06
Fix tests failing against akka master / 2.6

jrudolph:
Turns out those have been pure testing issues

### DIFF
--- a/akka-http-tests/src/test/scala/akka/http/javadsl/server/HttpAppSpec.scala
+++ b/akka-http-tests/src/test/scala/akka/http/javadsl/server/HttpAppSpec.scala
@@ -10,19 +10,20 @@ import java.util.concurrent.TimeUnit
 import java.net.SocketException
 
 import akka.Done
+import akka.http.impl.util.AkkaSpecWithMaterializer
 import akka.http.javadsl.ServerBinding
 import akka.http.javadsl.settings.ServerSettings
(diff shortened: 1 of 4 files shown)


akka-http-tests/src/test/scala/akka/http/javadsl/server/HttpAppSpec.scala
```diff
@@ -195,7 +196,7 @@ class HttpAppSpec extends AkkaSpec with RequestBuilding with Eventually {
         val port = serverSocket.getLocalPort
 
         try {
-          EventFilter[SocketException](message = "Address already in use", occurrences = 1) intercept {
+          EventFilter[SocketException](pattern = ".*Address already in use.*", occurrences = 1) intercept {
```
jrudolph: Seems the message is now wrapped.

docs/src/test/resources/application.conf
```diff
@@ -1,4 +1,4 @@
 akka {
-  # needed for ErrorFilter in RouteDirectivesExampleSpec.scala
-  loggers = ["akka.testkit.TestEventListener"]
+  # needed because we add akka-actor-typed here but don't include an slf4j implementation
+  use-slf4j = off
```
jrudolph: Basically, because of this logging was completely turned off in the docs tests.

Issue #2818, opened 2019-11-19
Don't handle TestFailedExceptions in RouteTest ExceptionHandlers


jrudolph:
That will allow assertions in test routes to fail the test instead of being converted to 500 response.

Seen here: https://guardrail.dev/scala/akka-http/generating-a-server#akka-http

raboof:
Seems reasonable.

Since we don't want to add a dependency on scalatest, I guess we'd recognize them by class name?

jrudolph:
Either that or try to put it into ScalaTestRouteTest where scalatest is available.

jrudolph:
Fixed in #2949.

PR #2949, opened 2020-02-09
Handle test exceptions separately from regular exceptions

edubrovski:
If we use fail() or an assertion fails inside a route the test won't be
marked as failed. Instead 500 response will be returned. Custom
ExceptionHandler rethrowing TestFailedException fixes this problem.

Refs #2818

### DIFF
--- a/akka-http-testkit/src/main/scala/akka/http/scaladsl/testkit/RouteTest.scala
+++ b/akka-http-testkit/src/main/scala/akka/http/scaladsl/testkit/RouteTest.scala
@@ -155,7 +155,13 @@ trait RouteTest extends RequestBuilding with WSTestRequestBuilding with RouteTes
               securedConnection = defaultHostInfo.securedConnection,
               defaultHostHeader = defaultHostInfo.host)
           val ctx = new RequestContextImpl(effectiveRequest, routingLog.requestLog(effectiveRequest), routingSettings)
-          val sealedExceptionHandler = ExceptionHandler.seal(exceptionHandler)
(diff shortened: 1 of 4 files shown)


lightbend-cla-validator:
Hi @edubrovski,

raboof:
Those implicits are planned to change in #2893 - would it be possible to do this via a directive instead of via an implicit?

edubrovski:
Ok, I'll look into it.

edubrovski:
> Those implicits are planned to change in #2893 - would it be possible to do this via a directive instead of via an implicit?
@raboof, sorry, didn't understand. Exception handling is already done explicitly via a directive in RouteTest:
`val semiSealedRoute = // sealed for exceptions but not for rejections
            Directives.handleExceptions(sealedExceptionHandler)(route)`

huntc:
Unsure where to document this, but since upgrading to 2.6, and given that I use `utest`, I had to do the following for my tests:

```scala
  override def testExceptionHandler: ExceptionHandler = ExceptionHandler {
    case e: AssertionError => throw e
  }
```

jrudolph:
> Unsure where to document this, but since upgrading to 2.6, and given that I use `utest`, I had to do the following for my tests:

Thanks, good point. I created https://github.com/akka/akka-http/issues/3399 to maybe support utest out of the box in the testkit. I guess you wrote a small implementation of `TestFrameworkInterface` for utest? Maybe you can share it in the ticket or even open up a PR?

Issue #1240, opened 2017-06-27
Incorrect handling of `filename`, `filename*` in Content-Disposition header


richdougherty:
This issue is visible in Play when we're sending a result. Play creates a header value like the following:

```
Content-Disposition: inline; filename="测 试.tmp"; filename*=utf-8''%E6%B5%8B%20%E8%AF%95.tmp
```

Akka HTTP parses this value in Play's `AkkaModelConversion` class, then renders the result as:

```
Content-Disposition: inline; filename="utf-8''%E6%B5%8B%20%E8%AF%95.tmp"
```

I believe this shows that Akka HTTP has bugs in both parsing and rendering the header.

* When parsing the header, the `filename*` parameter is preferred over the `filename`, since it comes second in the parameter list. This is correctly handled. However, the `filename*` has a special prefix (`utf8''`) that specifies the character encoding. This prefix should be used by the parser to set the character encoding, but not included in the filename itself. Akka HTTP appears to be including this value in the filename itself, rather than using it to specify a character set when decoding.

* When rendering the header, the `filename` parameter should probably only include "safe" ISO-8859-1, characters. Unsafe characters that need escaping should probably be stripped out, rather than encoded with `%`, like Akka HTTP does. This is because RFC2616 says that `\` should be used for quoting strings here. To be safe, complex characters could be excluded from the `filename` parameter and encoded properly in the `filename*` parameter instead. Modern clients will recognise and correctly handle the `filename*` parameter, ignoring the `filename` parameter. The `filename*` parameter can include the true filename, encoded and percent-escaped as per RFC6266. The suggested encoding is UTF-8, since it is mandatory for clients to support this encoding.

See:
* Related Play issue: https://github.com/playframework/playframework/issues/7501
* RFC about the filename parameter: https://tools.ietf.org/html/rfc6266#section-4.3
* Example Content-Disposition: https://tools.ietf.org/html/rfc6266#section-5
* RFC about mandatory client encodings: https://tools.ietf.org/html/rfc5987#section-3.2
* How to encode `*` parameters with non-standard characters: https://tools.ietf.org/html/rfc2231#section-4
* How to encode non-`*` parameters: https://tools.ietf.org/html/rfc2616#section-2.2

richdougherty:
Note: more discussion of how clients handle `filename` and `filename*` parameters. May be outdated.

* https://stackoverflow.com/questions/93551/how-to-encode-the-filename-parameter-of-content-disposition-header-in-http
[...]

PR #2962, opened 2020-02-12
Fix HeaderParser Unicode issue

jrudolph:
Seems to be a byte->char conversion problem. Since someone finally provided a reproducible test case, I had a quick look. Still needs proper tests.

Refs #1484

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpHeaderParser.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpHeaderParser.scala
@@ -211,8 +211,8 @@ private[engine] final class HttpHeaderParser private (
   @tailrec
   private def insert(input: ByteString, value: AnyRef)(cursor: Int = 0, endIx: Int = input.length, nodeIx: Int = 0, colonIx: Int = 0): Unit = {
     val char =
-      if (cursor < colonIx) CharUtils.toLowerCase(input(cursor).toChar)
-      else if (cursor < endIx) input(cursor).toChar
+      if (cursor < colonIx) CharUtils.toLowerCase((input(cursor) & 0xff).toChar)
+      else if (cursor < endIx) (input(cursor) & 0xff).toChar
       else '\u0000'
     val node = nodes(nodeIx)
     if (char == node) insert(input, value)(cursor + 1, endIx, nodeIx + 1, colonIx) // fast match, descend into only subnode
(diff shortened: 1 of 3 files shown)


jrudolph:
Now added a proper (but very simple) test.

The reason it was so elusive is that it was related to caching. So, to trigger it, there have to be two subsequent headers lines where the second one is a prefix of the first one up to the non-ascii char in the first line. Since caching is done on a per-connection base, those headers could be requests / responses apart.

In the example that @shkhln provided it was a single multipart file upload request with two files that both started with a non-ascii character. In that case the resulting `Content-Disposition` lines of the two body parts shared a prefix up to the filename where it crashed on the occurrence of the second one...:

```
0000: POST /upload HTTP/1.1
0017: Host: 127.0.0.1:8080
002d: User-Agent: curl/7.65.3
0046: Accept: */*
0053: Content-Length: 342
0068: Content-Type: multipart/form-data; boundary=--------------------
00a8: ----5bee37ebfde51c72
00be: 
=> Send data, 342 bytes (0x156)
0000: --------------------------5bee37ebfde51c72

// The next line puts the broken entry into the cache (".." is actually a utf-8 char that curl doesn't display correctly)

002c: Content-Disposition: form-data; name="test"; filename=".."
0068: Content-Type: application/octet-stream
0090: 
0092: 
0094: --------------------------5bee37ebfde51c72

// This line is a prefix up to the broken character and would crash the parser

00c0: Content-Disposition: form-data; name="test"; filename=".."
00fc: Content-Type: application/octet-stream
0124: 
0126: 
0128: --------------------------5bee37ebfde51c72--
```

Thanks a lot again, @shkhln for providing a reproducer. I guess, it's a common enough case for multi file uploads of files with non-ascii characters, which are uncommon in English-speaking countries but less so any other countries.

(There's still related issue #1240 open. In theory, you shouldn't put non-ascii chars into the `filename` parameter but use `filename*` instead, but it seems support for that is something wild in reality).

/cc @sirthias this took us a few years to track down :)

sirthias:
Ahh, damn.
I'm truly sorry about that!!
I've had a few of these "more bits than expected" issues myself recently and they are true beasts to find. Unfortunately even the `-Ywarn-numeric-widen` compiler settings (that I always have enabled) is no help here...

sirthias:
One thought: Have you done a quick check of all `input(cursor)` and `byteAt(...)` call sites to look for possible bugs of the same kind that are unrelated to this issue?

jrudolph:
> Ahh, damn.
> I'm truly sorry about that!!

No worries, tricky but fast code :)

> One thought: Have you done a quick check of all `input(cursor)` and `byteAt(...)` call sites to look for possible bugs of the same kind that are unrelated to this issue?

Yeah, I think I looked at more than the ones needed for this particular problem. At least in the HttpHeaderParser.

jrudolph:
> Unfortunately even the `-Ywarn-numeric-widen` compiler settings (that I always have enabled) is no help here...

Yep, this one here is probably also about signed byte vs. unsigned char. So probably something like this happens:

```
scala> 0x82.toByte
res0: Byte = -126

scala> res0.toChar.toInt
res3: Int = 65410

scala> res0.toInt
res4: Int = -126

scala> (res0 & 0xff).toInt
res6: Int = 130

scala> (res0 & 0xff).toChar.toInt
res7: Int = 130
```

sirthias:
Exactly.
It truly is a bummer that the JVM has no support for unsigned integer types.
It would make sooo many bugs simply disappear.

YikSanChan:
@jrudolph Can you tag here when the fix is released?

jrudolph:
> Can you tag here when the fix is released?

10.2.0-M1 already has the fix and we'll soon release 10.1.12 which will have the fix as well.

raboof:
10.1.12 is now released

PR #2963, opened 2020-02-12
core: introduce Inherited.upcast to fix warning in javadsl.Http

jrudolph:
An alternative to #2894 for this particular case.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/util/JavaMapping.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/util/JavaMapping.scala
@@ -180,6 +180,9 @@ private[http] object JavaMapping {
   class Inherited[J <: AnyRef, S <: J](implicit classTag: ClassTag[S]) extends JavaMapping[J, S] {
     def toJava(scalaObject: S): J = scalaObject
     def toScala(javaObject: J): S = cast[S](javaObject)
+
+    import scala.language.higherKinds
+    def downcast[F[+_]](f: F[J]): F[S] = f.asInstanceOf[F[S]]
   }
 
   implicit object ConnectionContext extends Inherited[ConnectionContext, akka.http.scaladsl.ConnectionContext]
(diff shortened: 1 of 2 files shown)


akka-http-core/src/main/scala/akka/http/impl/util/JavaMapping.scala
```diff
@@ -180,6 +180,9 @@ private[http] object JavaMapping {
   class Inherited[J <: AnyRef, S <: J](implicit classTag: ClassTag[S]) extends JavaMapping[J, S] {
     def toJava(scalaObject: S): J = scalaObject
     def toScala(javaObject: J): S = cast[S](javaObject)
+
+    import scala.language.higherKinds
+    def upcast[F[+_]](f: F[J]): F[S] = f.asInstanceOf[F[S]]
```
jrudolph: or should it be `downcast`?

raboof: I guess it's a (safe) downcast https://en.wikipedia.org/wiki/Downcasting

akka-http-core/src/main/scala/akka/http/javadsl/Http.scala
```diff
@@ -40,7 +40,7 @@ class Http(system: ExtendedActorSystem) extends akka.actor.Extension {
 
   import language.implicitConversions
   private implicit def completionStageCovariant[T, U >: T](in: CompletionStage[T]): CompletionStage[U] = in.asInstanceOf[CompletionStage[U]]
-  private implicit def javaModelIsScalaModel[J <: AnyRef, S <: J](in: Future[J])(implicit ev: JavaMapping.Inherited[J, S]): Future[S] = in.asInstanceOf[Future[S]]
+  private implicit def javaModelIsScalaModel[J <: AnyRef, S <: J](in: Future[J])(implicit ev: JavaMapping.Inherited[J, S]): Future[S] = ev.upcast(in)
```
raboof: nice, more explicit

akka-http-core/src/main/scala/akka/http/impl/util/JavaMapping.scala
jrudolph: Woops, thanks, fixed.

PR #2964, opened 2020-02-12
core: nest correctly in NewHostConnectionPool

jrudolph:
To avoid unsafe access of `connectionEstablished`

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/NewHostConnectionPool.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/NewHostConnectionPool.scala
@@ -511,11 +511,14 @@ private[client] object NewHostConnectionPool {
               slot.onConnectionCompleted()
             }
           override def onUpstreamFailure(ex: Throwable): Unit =
-            if (connectionEstablished)
-              withSlot { slot =>
+            withSlot { slot =>
+              if (connectionEstablished) {
                 slot.debug("Connection failed")
                 slot.onConnectionFailed(ex)
               }
+              // otherwise, rely on connection.onComplete to fail below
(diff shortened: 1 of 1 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/NewHostConnectionPool.scala
```diff
@@ -511,11 +511,14 @@ private[client] object NewHostConnectionPool {
               slot.onConnectionCompleted()
             }
           override def onUpstreamFailure(ex: Throwable): Unit =
-            if (connectionEstablished)
-              withSlot { slot =>
+            withSlot { slot =>
+              if (connectionEstablished) {
```
raboof: :+1:

akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/NewHostConnectionPool.scala
```diff
@@ -546,9 +549,8 @@ private[client] object NewHostConnectionPool {
 
           slot.debug("Establishing connection")
           val connection =
-            Source.fromGraph(requestOut.source)
-              .viaMat(connectionFlow)(Keep.right)
-              .to(responseIn.sink)
+            connectionFlow
+              .join(Flow.fromSinkAndSource(responseIn.sink, requestOut.source))
```
raboof: How/why is this better?

jrudolph: Good point, no good reason aside from two imports less :).

I just found that whole change in my big batch and didn't look closely enough. I find the previous version simpler to read so reverted.

Issue #2956, opened 2020-02-10
Type safety for path directives


kag0:
Today, it's possible to build routing DSLs with the path directives that result in unreachable code. 
For example
```scala
path("users" / Segment) { userId =>
  get {
    complete(userService.getUser(userId)
  } ~
  path("friends") { 
    get {
      complete(userService.listFriends(userId)
    }
  }
}
```

Noticing this requires an attention to detail that it would be great if the compiler could exert for us.
The general idea would be to create a trait which indicates if a route or directive is path-sensitive and if so, don't allow it to be passed to an inner directive of `path` or `pathEnd`. 
I could see the scope of this going a little crazy, so a first step could be to just to try to cover use cases with the routing DSL, and say it's fair to lose type safety when flat mapping and such.

jrudolph:
I ran into that trap often enough myself, so I can feel the pain and see how that would be useful.

That said, a compile-time solution is very unlikely. The routing DSL is by design very dynamic. This has various drawbacks but the main benefit is that the DSL is consistent. There are basically only two types, `Directive` and `Route` that can be used to implement any kind of behavior you need. This is probably not going to change. If you think about it, it would have to be a massively invasive change just to support the type-safety for path. What if we would like to introduce type-safe variants of other directives like `get`/`post` etc?

I can see different ways of improvement:

 * a scalafix style tooling that tries to analyze routes and gives hints
 * Right now the routing DSL is more dynamic that needed in most cases (because of monadic extraction). This prevents a runtime analysis of problems like the one here because most of the routing tree is only constructed when a request comes in. We might be able to pull off a change to the API that would remove dynamic execution for most routes so that we could inspect the route tree. In that case we could offer runtime hints at the start of the problem when we find invalid routes.
 * This problem does not seem to be a runtime problem but more of a debugging problem. We could just bail and say that this a simple thing to find in tests so it's inconvenient but not too serious. Still when it happens, it keeps you scratching your head so it would be good if we could offer more guidance especially to new users so they can trace route execution and understand why a route didn't match.

kag0:
> it would have to be a massively invasive change just to support the type-safety for path

This is what I meant by "the scope of this going a little crazy". I was hopeful that a minimally invasive solution could be brainstormed. One that doesn't make things completely type-safe, but covers the common gotchas. But,

Being able to inspect the route tree would be almost as good, for me. Especially the ability to write a test to check the structure itself without having to send a request to hit every end route.
[...]

Issue #201, opened 2016-09-09
Support REST API doc generation


ktoso:
<a href="https://github.com/guersam"><img src="https://avatars.githubusercontent.com/u/969120?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [guersam](https://github.com/guersam)**
_Saturday Dec 20, 2014 at 23:05 GMT_
_Originally opened as https://github.com/akka/akka/issues/16591_

----

Related ML thread: https://groups.google.com/forum/#!topic/akka-user/d18_smRGeoE

> Hi,
>
>Although this topic has been discussed several times in spray-user list [1], I'd like to refresh this issue because I think it's an important feature for wider adoption as well as the mostly asked ones like WebSocket.
>
>Spray/akka-http's routing DSL is the best among the existing framework/libraries IMO, however, it's cascaded and mixed routing structure makes automated documentation and code generation somewhat difficult. Although there's already good looking spray-swagger, it's annotation based approach seems too verbose sometimes.
>
>A random thoughts for alternatives are:
>1. Macro based introspection which extracts routing structure from the DSL
>2. Integration with the testkit (either way of doc <-> test code generation) instead of with routing code directly
>
>I know it isn't urgent now, just hoping to gather some wisdom to bring it to the next step for the long term goal. 
>
>
>Regards,
>Jisoo
>
>
>[1] https://groups.google.com/forum/#!searchin/spray-user/swagger
>[2] https://github.com/gettyimages/spray-swagger
>[3] http://www.scalatra.org/2.3/guides/swagger.html

ktoso:
<a href="https://github.com/ktoso"><img src="https://avatars.githubusercontent.com/u/120979?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [ktoso](https://github.com/ktoso)**
_Monday Dec 22, 2014 at 10:23 GMT_

----

Hi there, I agree this would be a really really good feature. Thanks for making it an issue so we don't forget about it!

Question being if it isn't better to be taken up as an community effort – as we currently do not have enough man-power to tackle this one.

// cc @jrudolph @sirthias

ktoso:
<a href="https://github.com/drewhk"><img src="https://avatars.githubusercontent.com/u/932285?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [drewhk](https://github.com/drewhk)**
_Monday Dec 22, 2014 at 10:29 GMT_

----

[...]

PR #2966, opened 2020-02-13
[WIP] provide inspectable routes with minimal changes

jrudolph:
Refs #201, #2956

In a previous try, I tried to create a full AST of possible route structures. However, I don't think that's necessary. What we should strive for is

 * a generic structure that is good enough to provide information for debugging and introspection purposes
 * only minimal API changes to help with migration
 * an incremental way of adding metadata to existing directives without having to create a complete fork of the API

### DIFF
--- a/akka-http-tests/src/test/scala/akka/http/scaladsl/server/TestServer.scala
+++ b/akka-http-tests/src/test/scala/akka/http/scaladsl/server/TestServer.scala
@@ -63,7 +63,7 @@ object SwaggerRoute {
 
   private def apiDescForRoute(route: Route): OpenApi = {
     case class RoutePath(segments: Seq[DirectiveRoute], last: Route) {
-      override def toString: String = segments.map(_.directiveName).mkString(" -> ") + " -> " + last.getClass.toString
+      override def toString: String = segments.map(s => s"${s.directiveName}${s.directiveInfo.fold("")(i => s"($i)")}").mkString(" -> ") + " -> " + last.getClass.toString
     }
     def leaves(route: Route, prefix: Seq[DirectiveRoute]): Seq[RoutePath] = route match {
       case AlternativeRoutes(alternatives) =>
--- a/akka-http-tests/src/test/scala/akka/http/scaladsl/server/TestServer.scala
+++ b/akka-http-tests/src/test/scala/akka/http/scaladsl/server/TestServer.scala
@@ -14,10 +14,89 @@ import akka.stream._
 import akka.stream.scaladsl._
 import akka.http.scaladsl.Http
 import akka.http.scaladsl.common.EntityStreamingSupport
+import akka.http.scaladsl.model.headers.{ HttpOrigin, `Access-Control-Allow-Methods`, `Access-Control-Allow-Origin` }
+import akka.http.scaladsl.server.util.ApplyConverter
 
 import scala.concurrent.ExecutionContext
 import scala.concurrent.duration._
 import scala.io.StdIn
+
+object SwaggerRoute {
+  import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport._
+
+  import spray.json.DefaultJsonProtocol._
+
+  case class ParameterSpec(
+    name: String,
+    in:   String
+  )
+  case class ResponseSpec(
+    summary: String
+  )
+  case class OperationSpec(
+    summary:    String,
+    parameters: Seq[ParameterSpec]        = Vector.empty,
+    responses:  Map[String, ResponseSpec] = Map.empty
+  )
+  case class PathSpec(
+    get:  Option[OperationSpec] = None,
+    post: Option[OperationSpec] = None,
+    put:  Option[OperationSpec] = None
+  )
+
+  case class OpenApi(
+    title:   String,
+    openapi: String,
+    paths:   Map[String, PathSpec]
+  )
+  object OpenApi {
+    implicit val paramaterSpecFormat = jsonFormat2(ParameterSpec.apply _)
+    implicit val responseSpecFormat = jsonFormat1(ResponseSpec.apply _)
+    implicit val operationSpecFormat = jsonFormat3(OperationSpec.apply _)
+    implicit val pathSpecFormat = jsonFormat3(PathSpec.apply _)
+    implicit val openApiFormat = jsonFormat3(OpenApi.apply _)
+  }
+
+  import Directives._
+  def route(forRoute: Route): Route =
+    complete(apiDescForRoute(forRoute))
+
+  private def apiDescForRoute(route: Route): OpenApi = {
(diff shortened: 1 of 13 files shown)


akka-http/src/main/scala/akka/http/scaladsl/server/RequestContextImpl.scala
```diff
@@ -23,17 +23,18 @@ import scala.concurrent.{ ExecutionContextExecutor, Future }
 private[http] class RequestContextImpl(
   val request:          HttpRequest,
   val unmatchedPath:    Uri.Path,
+  val tokenValues:      Map[ExtractionToken[_], Any],
```
jrudolph: This could be a request attribute as well once they are managed but since it is specific to routing maybe it fits here well.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/ParameterDirectives.scala
```diff
@@ -84,10 +84,10 @@ object ParameterDirectives extends ParameterDirectives {
     def apply(): Out
   }
   object ParamMagnet {
-    implicit def apply[T](value: T)(implicit pdef: ParamDef[T]): ParamMagnet { type Out = pdef.Out } =
+    implicit def apply[T, O](value: T)(implicit pdef: ParamDef[T] { type Out = Directive[O] }): ParamMagnet { type Out = pdef.Out } =
       new ParamMagnet {
         type Out = pdef.Out
-        def apply() = pdef(value)
+        def apply() = pdef(value) named "parameters"
```
jrudolph: Ultimately, that's not enough and we want to add separate information about each of the parameters.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/MethodDirectives.scala
```diff
@@ -114,12 +114,12 @@ object MethodDirectives extends MethodDirectives {
...
 
   // format: OFF
-  private val _delete : Directive0 = method(DELETE)
-  private val _get    : Directive0 = method(GET)
-  private val _head   : Directive0 = method(HEAD)
-  private val _options: Directive0 = method(OPTIONS)
-  private val _patch  : Directive0 = method(PATCH)
-  private val _post   : Directive0 = method(POST)
-  private val _put    : Directive0 = method(PUT)
+  private val _delete : Directive0 = method(DELETE) named "delete"
```
jrudolph: These ones override the name from the `def method(HttpMethod)` one above.

akka-http-tests/src/test/scala/akka/http/scaladsl/server/TestServer.scala
```diff
@@ -14,11 +14,90 @@ import akka.stream._
...
+    case class RoutePath(segments: Seq[DirectiveRoute], last: Route) {
+      override def toString: String = segments.map(_.directiveName).mkString(" -> ") + " -> " + last.getClass.toString
+    }
+    def leaves(route: Route, prefix: Seq[DirectiveRoute]): Seq[RoutePath] = route match {
+      case AlternativeRoutes(alternatives) =>
+        alternatives.flatMap(leaves(_, prefix))
+      case dr: DirectiveRoute => leaves(dr.child, prefix :+ dr)
+      case last               => Vector(RoutePath(prefix, last))
+    }
+    leaves(route, Vector.empty).foreach(println)
```
jrudolph, line 74: What this prints currently is this:

```
[info] pathPrefix(<function1>) -> pathEnd -> post -> class akka.http.scaladsl.server.StandardRoute$$anon$1
[info] pathPrefix(<function1>) -> pathEnd -> put -> class akka.http.scaladsl.server.StandardRoute$$anon$1
[info] pathPrefix(<function1>) -> path(<function1>) -> get -> class akka.http.scaladsl.server.DynamicDirective$$$Lambda$310/1221027335
[info] pathPrefix(<function1>) -> path(<function1>) -> delete -> class akka.http.scaladsl.server.DynamicDirective$$$Lambda$310/1221027335
```

That means, we can already somewhat enumerate the leaf routes but don't have enough metadata available to look into the actual path matchers. That should be the next step.

akka-http/src/main/scala/akka/http/scaladsl/server/InspectableRoute.scala
```diff
@@ -0,0 +1,111 @@
...
+}
+object DynamicDirective {
+  def dynamic: (ExtractionContext => Route) => Route =
+    inner => ctx => inner {
+      new ExtractionContext {
+        override def extract[T](token: ExtractionToken[T]): T = ctx.tokenValue(token)
+      }
+    }(ctx)
+
+  implicit class AddStatic[T](d: Directive1[T]) {
```
jrudolph, line 100: It seems, this one isn't necessary at all. Adding implicits like the above `addByNameNullaryApply` and `addDirective1Apply` means that we can change the semantics of route building with a single import that brings those implicits into scope.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/MethodDirectives.scala
```diff
@@ -114,12 +114,12 @@ object MethodDirectives extends MethodDirectives {
...
+  private val _get    : Directive0 = method(GET) named "get"
+  private val _head   : Directive0 = method(HEAD) named "head"
+  private val _options: Directive0 = method(OPTIONS) named "options"
+  private val _patch  : Directive0 = method(PATCH) named "patch"
+  private val _post   : Directive0 = method(POST) named "post"
+  private val _put    : Directive0 = method(PUT) named "put"
```
jrudolph, line 123: Ultimately, these `named` decorators are just a toy. What we should provide instead is something like `meta` which allows to decorate a Directive with typed meta information.

akka-http-tests/src/test/scala/akka/http/scaladsl/server/TestServer.scala
```diff
@@ -14,11 +14,90 @@ import akka.stream._
...
+    implicit val pathSpecFormat = jsonFormat3(PathSpec.apply _)
+    implicit val openApiFormat = jsonFormat3(OpenApi.apply _)
+  }
+
+  import Directives._
+  def route(forRoute: Route): Route =
+    complete(apiDescForRoute(forRoute))
+
+  private def apiDescForRoute(route: Route): OpenApi = {
+    case class RoutePath(segments: Seq[DirectiveRoute], last: Route) {
```
jrudolph, line 65: So, what we get here per leaf route is an chain ordered chain of all (the metadata of) the directives that are applied for that leaf route. Once we expose enough metadata, we should be able to generate openapi specifications but also do "static" (= at route building time) analytics of the routing tree. Even with this simple prototype we could prevent that people would be accidentally nesting `path` or method directives.

PR #2967, opened 2020-02-18
Clean up CustomHttpMethodSpec

raboof:


docs/src/test/scala/docs/http/scaladsl/server/directives/CustomHttpMethodSpec.scala
```diff
@@ -45,10 +43,15 @@ class CustomHttpMethodSpec extends AkkaSpec with ScalaFutures
 
       val request = HttpRequest(BOLT, s"http://$host:$port/", protocol = `HTTP/1.1`)
       //#application-custom
-      val response = Http().singleRequest(request).futureValue
 
+      // Make sure we're bound
+      binding.futureValue
```
jrudolph: Good catch here.

jrudolph:
> LGTM (though, this test should probably just go into `ClientServerSpec` or `HttpServerSpec` where all the necessary infrastructure is already in place.

Not sure what I though here? This is in docs, so everything as expected. Sorry for that.

Issue #2738, opened 2019-10-02
Prepare routing API for deprecation / removal of argument adaption


jrudolph:
In akka-http routing we make heavy use of adapting arguments into tuples (see http://spray.io/blog/2012-12-13-the-magnet-pattern/#dryness-for-many-similar-overloads for an explanation of why that's needed). The required feature from the Scala compiler is argument adaptation which automatically convert a sequence of arguments of a method to a tuple if the target method takes only a single parameter. This feature is going away with upcoming Scala versions and the consequences can already be experienced by using the `-Yno-adapted-args` compiler flags.

Most importantly the `parameters` and `formFields` directives are affected.

Potential alternatives are:
 * require users to explicitly supply tuples to directives that make use of the magnet pattern (somewhat verbose, all users need to change code, but we might be able to provide scalafix migrations)
 * use a combinator approach to join a set of arguments, i.e. convert `parameters('name, 'age)` to  `parameters('name ~~ 'age)` or similar (still requires users to change code, type inference will be probably be more complex leading to longer compile times and (even) worse error messages)
 * use sbt-boilerplate to generate stubs for all arities and remove / deprecate old variant (same syntax, probably better error messages, straight-forward typing, probably some simplification of the old magnet pattern code, some considerations needed to deal with binary compatibility)

So far, the last option sounds most appealing to me.

raboof:
Let's try the sbt-boilerplate approach first and reconsider if this turns out to blow up

jrudolph:
>     * use sbt-boilerplate to generate stubs for all arities and remove / deprecate old variant (same syntax, probably better error messages, straight-forward typing, probably some simplification of the old magnet pattern code, some considerations needed to deal with binary compatibility)

I had a try at this option. This is hard (impossible?) to pull off because the magnet pattern is inevitable in this place to be able to join the results of to parameter extractions. E.g. to implement the binary

Each param usually will either be equivalent to a Directive1 or a Directive0. Directive1 is the usual case. while using `"name".!(defaultName)" will lead to a `Directive0`. To be able to produce a combined directive here we need to join the result types of `param1` and `param2`. However, that can only be done at the call-site because only there all the parameters and types. The solution with regular implicits (in contrast to magnets) would be to use:

```
def parameters(param1: ParamSpec, param2 ParamSpec)(implicit join: Join[param1.Out, param2.Out]): Directive[join.Out]
```

[...]

PR #2971, opened 2020-02-20
[WIP] Suggestion about how to replace magnets by simpler implicits for variadic directives

jrudolph:
Refs #2738


akka-http/src/main/scala/akka/http/scaladsl/server/directives/ParameterDirectives.scala
```diff
@@ -79,12 +87,29 @@ object ParameterDirectives extends ParameterDirectives {
   private val _parameterSeq: Directive1[immutable.Seq[(String, String)]] =
     extract(_.request.uri.query().toSeq)
 
+  trait ParamSpec {
+    type Out
+    def apply(): Directive1[Out]
```
jrudolph: This is the important difference to make it work: we require here that each parameter argument must provide exactly one value. This almost works out of the box, only for required parameters there's a slight syntax change because they used to result in a `Directive0` where they are now a somewhat weird `Directive1[Unit]`.

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/ParameterDirectivesSpec.scala
```diff
@@ -212,22 +212,22 @@ class ParameterDirectivesSpec extends AnyFreeSpec with GenericRoutingSpec with I
   "The 'parameter' requirement directive should" - {
     "reject the request with a MissingQueryParamRejection if request do not contain the required parameter" in {
       Get("/person?age=19") ~> {
-        parameter('nose ! "large") { completeOk }
+        parameter('nose ! "large") { _ => completeOk }
```
jrudolph: This is the require syntax change.

docs/src/test/scala/docs/http/scaladsl/HttpServerExampleSpec.scala
```diff
@@ -416,7 +416,7 @@ class HttpServerExampleSpec extends AnyWordSpec with Matchers
         concat(
           pathEnd {
             concat(
-              (put | parameter('method ! "put")) {
+              (put | parameter('method ! "put").tmap(_ => ())) {
```
jrudolph: Here the syntax change is a bit more awkward to be able to continue supporting `put | parameter(...)` which requires `Directive0` as both operands.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/ParameterDirectives.scala
```diff
@@ -79,12 +87,29 @@ object ParameterDirectives extends ParameterDirectives {
...
+      }
+  }
+
   sealed trait ParamMagnet {
     type Out
     def apply(): Out
   }
   object ParamMagnet {
-    implicit def apply[T](value: T)(implicit pdef: ParamDef[T]): ParamMagnet { type Out = pdef.Out } =
+    /*implicit*/ def apply[T](value: T)(implicit pdef: ParamDef[T]): ParamMagnet { type Out = pdef.Out } =
```
jrudolph: Needed to disambiguate the new and the old overloads. Fortunately, that makes it binary compatible.

docs/src/test/scala/docs/http/scaladsl/HttpServerExampleSpec.scala
ennru: Would adding a special method make it less weird?
(Either instead of `.tmap(...)` or even instead of `parameter`.)
But how would users discover what to do when it doesn't compose...

jrudolph: Instead of `.tmap(_ => ()))` we could have something like `.dropExtractions` or `.dropAll`.

raboof:
This direction seems compelling to me! Might even be slightly simpler, and indeed I think we can expect the `!` notation to be rare.

jrudolph:
I pushed this now a bit further:

 * variadic methods now generated by sbt-boilerplate
 * added direct implicits for the new `ParamSpec` implicit API. Before, the new implicit API was just a shim over the old one which would mean we couldn't deprecate the old API because it was still implicitly referenced from  user code. Now, it's separate and recompiled code will not require the old implementation any more. We still need to keep it around for bincompat.

docs/src/test/scala/docs/http/scaladsl/HttpServerExampleSpec.scala
```diff
@@ -394,7 +394,7 @@ class HttpServerExampleSpec extends AnyWordSpec with Matchers
         concat(
           pathEnd {
             concat(
-              (put | parameter("method".requiredValue("put"))) {
+              (put | parameter("method".requiredValue("put")).tmap(_ => ())) {
```
ennru, line 397: I think the DSL becomes more regular when you always get a value for a parameter definition. The `tmap` is more of a distraction.
```suggestion
              (put | parameter("method".requiredValue("put"))) { _ => 
```
In the case of `requiredValue` it could even be that fixed value, but that would break the API too much.

jrudolph: In that case, the `tmap` (or some equivalent) is required, though, to make the `|`-operator work. That only works when both sides extract the same number of objects.

jrudolph: We could rather say, to show this particular example is a distraction for the example, since it is such a rare case.

raboof:
@jrudolph is there anything left to do here before we un-draft?

jrudolph:
> @jrudolph is there anything left to do here before we un-draft?

For parameters, it seems everything is done. We need to decide if we are fine with the source incompatibility for the `requiredValue` case, where an extra value is now extracted. Also, it should be mentioned in the release notes / migration guide.

jrudolph:
Then we need the same for `formFields` which I have already started and need to finish.

ennru:
> We need to decide if we are fine with the source incompatibility for the requiredValue case

I think it is fine. A note in the migration docs will do for this presumably rare case.

raboof:
> A note in the migration docs will do for this presumably rare case.

I'll add it

jrudolph:
Thanks for adding the notes. Could be merged from my side!

jrudolph:
Rebased and squashed again.

Issue #2957, opened 2020-02-11
Http2StreamHandling - handleOutgoingEnded received unexpectedly in state Closed


steinybot:
I had just started an Akka gRPC service and was trying it out with grpcurl. I noticed a log message which said that it might be a bug and to report it so here I go.

The individual message is:
```
[info] [WARN] [02/11/2020 22:10:46.808] [ConsoleAPIServer-akka.actor.default-dispatcher-4] [Http2ServerDemux(akka://ConsoleAPIServer)] handleOutgoingEnded received unexpectedly in state Closed. This indicates a bug in Akka HTTP, please report it to the issue tracker.
```

The mostly full log is:
```
[info] [DEBUG] [02/11/2020 21:55:05.585] [main] [akka.actor.ActorSystemImpl(ConsoleAPIServer)] Binding server using HTTP/2
[info] [DEBUG] [02/11/2020 21:55:05.729] [ConsoleAPIServer-akka.actor.internal-dispatcher-6] [akka://ConsoleAPIServer/system/IO-TCP/selectors/$a/0] Successfully bound to /127.0.0.1:8080
[info] 21:55:05.735 INFO  - Console API Server is bound to: /127.0.0.1:8080 (ConsoleServer.scala:37)
[info] [DEBUG] [02/11/2020 21:57:17.874] [ConsoleAPIServer-akka.actor.internal-dispatcher-2] [akka://ConsoleAPIServer/system/IO-TCP/selectors/$a/0] New connection accepted
[info] [DEBUG] [02/11/2020 21:58:18.095] [ConsoleAPIServer-akka.actor.default-dispatcher-9] [akka://ConsoleAPIServer/system/Materializers/StreamSupervisor-0/flow-1-0-detacher] Aborting tcp connection to /127.0.0.1:55991 because of upstream failure: akka.stream.scaladsl.TcpIdleTimeoutException: TCP idle-timeout encountered on connection to [/127.0.0.1:55991], no bytes passed in the last 1 minute
[info] [DEBUG] [02/11/2020 21:58:18.096] [ConsoleAPIServer-akka.actor.default-dispatcher-9] [TcpConnectionStage$TcpStreamLogic(akka://ConsoleAPIServer)] Aborting connection from 127.0.0.1:55991 because of downstream failure: java.util.concurrent.TimeoutException: No elements passed in the last 1 minute.
[...]

PR #2976, opened 2020-02-21
Accept RST on an already-half-closed HTTP/2 stream

raboof:
Refs #2957

When grpccurl is used on a Akka gRPC server that does not support server
reflection, we send a 404, and grpcurl sends a RST with error code
PROTOCOL_ERROR.

We definitely didn't handle the RST correctly in that scenario. This PR
fixes handling the RST.

Separate from this PR, we should also investigate why grpcurl is sending
a PROTOCOL_ERROR: this indicates a HTTP/2 protocol-level error, while us
sending a 404 is an application-level error. It might be that grpcurl is
just incorrectly sending a PROTOCOL_ERROR here, or that we're in fact doing
something else wrong that triggers it.

The test is not very nice, but I don't see another way.

### DIFF
--- a/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
+++ b/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
@@ -141,7 +141,7 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with StageLoggi
         }
       case r: RstStreamFrame =>
         buffer.onRstStreamFrame(r)
-        multiplexer.cancelSubStream(r.streamId)
+        onReset(r.streamId)
         Closed
 
       case h: ParsedHeadersFrame =>
@@ -156,6 +156,7 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with StageLoggi
 
       case _ => throw new IllegalStateException(s"Unexpected frame type ${event.frameTypeName} in state ${this.getClass.getName}.")
     }
+    protected def onReset(streamId: Int): Unit
 
     protected def maybeFinishStream(endStream: Boolean): IncomingStreamState =
       if (endStream) afterEndStreamReceived else this
@@ -169,11 +170,18 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with StageLoggi
   // on the incoming side there's (almost) no difference between Open and HalfClosedLocal
   case class Open(buffer: IncomingStreamBuffer) extends ReceivingData(HalfClosedRemote) {
     override def handleOutgoingEnded(): IncomingStreamState = HalfClosedLocal(buffer)
+
+    override protected def onReset(streamId: Int): Unit =
+      multiplexer.cancelSubStream(streamId)
   }
   /**
    * We have closed the outgoing stream, but the incoming stream is still going.
    */
-  case class HalfClosedLocal(buffer: IncomingStreamBuffer) extends ReceivingData(Closed)
+  case class HalfClosedLocal(buffer: IncomingStreamBuffer) extends ReceivingData(Closed) {
+    override protected def onReset(streamId: Int): Unit = {
+      // nothing further to do as we're already half-closed
+    }
+  }
 
   /**
    * They have closed the incoming stream, but the outgoing stream is still going.
--- a/akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
+++ b/akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
@@ -302,6 +302,28 @@ class Http2ServerSpec extends AkkaSpecWithMaterializer("""
         val error = entityDataIn.expectError()
         error.getMessage shouldBe "Stream with ID [1] was closed by peer with code INTERNAL_ERROR(0x02)"
       }
+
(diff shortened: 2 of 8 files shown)


akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -141,7 +141,7 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with StageLoggi
         }
       case r: RstStreamFrame =>
         buffer.onRstStreamFrame(r)
-        multiplexer.cancelSubStream(r.streamId)
+        onReset(r.streamId)
```
jrudolph: So, the problem is here that the side-effects are executed in the wrong order? I.e. we want to go to the `Closed` state but `multiplexer.cancelSubStream` accesses the state before it has changed.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -302,6 +302,28 @@ class Http2ServerSpec extends AkkaSpecWithMaterializer("""
...
+
+        // We emit a 404 response, half-closing the stream.
+        emitResponse(streamId = 1, HttpResponse(StatusCodes.NotFound))
+
+        // The client closes the stream with a protocol error. We should look into whether this is legit - but
+        // in any case we should handle it correctly.
+        sendRST_STREAM(streamId = 1, ErrorCode.PROTOCOL_ERROR)
+        entityDataIn.expectError()
+
+        // Unfortunately we can't check the problem in
```
jrudolph: So, before the fix the connection is just aborted? Could we check that?

jrudolph: Maybe the warning should have been a crash instead?

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
raboof: No, the problem was that this resetting logic was shared between the `Open` and `HalfClosedLocal` states. As far as I understand:

In `Open`, the old behavior was correct: `buffer.onRstStreamFrame(r)` made sure we closed the incoming request stream and `multiplexer.cancelSubStream(r.streamId)` made sure we closed the outgoing response.

However, in the scenario that we have already closed the outgoing response we are in `HalfClosedLocal` rather than `Open`. If we are in `HalfClosedLocal` and we receive a RST, we should still close the incoming request stream, but since the outgoing response stream was already closed we should not call `multiplexer.cancelSubStream(r.streamId)`.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
raboof: The behavior before and after is the same: before it'd abort the incoming request stream (OK), and try to close the outgoing response stream, but the latter would fail since it was already closed. Crashing everything just because one stream is in an unexpected state seems a bit dangerous...

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
jrudolph: What I meant was, that the next line says, we should go to `Close` but that only applies after `cancelSubStream` has already run (and triggered the warning in `HalfClosedLocal`.

jrudolph: I.e. `updateState` was called in a nested fashion which is basically not supported. The fix is fine, though :)

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
jrudolph: Any reason not to use the regular `EventFilter` here?

raboof: hmm, can you use EventFilter to check we do _not_ get a log message?

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
raboof: Ah - that problem might still be there for the `Open` state, then? For a different PR though.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
jrudolph: Good point, maybe with `occurrences=0` and a bit of sleep in the intercept?

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
jrudolph: It's not cleanly solved but what happens is that in the open state it is able to handle the `handleOutgoingEnded` without a warning (temporarily shifting the state to half closed).

Issue #2845, opened 2019-12-02
deprecate Http().createClientHttpsContext and provide alternative


robstoll:
AkkaSSLConfig was deprecated with 2.6.0, the migration guide points to https://doc.akka.io/docs/akka/current/stream/stream-io.html#tls but without showing how it has to be used in conjunction with akka-http. 

I am currently using `Http().createClientHttpsContext(...)` and look for an alternative (the new way to go) but cannot come up with something senseful as also `HttpsConnectionContext` is still using `val sslConfig: Option[AkkaSSLConfig]` meaning I would still stick with AkkaSSLConfig.

raboof:
For the scaladsl, `HttpsConnectionContext` might be an alternative, passing `None` for `AkkaSSLConfig`.

For the javadsl I agree we need something nicer.

As for deprecating the methods that take an AkkaSSLConfig, we added this in https://github.com/akka/akka/issues/20214 to be able to disable SNI on a per-connection basis (instead of globally for the whole JVM). Let's double-check we indeed have a new way to do this first.

jrudolph:
Done for now in #2983.

PR #2983, opened 2020-02-26
core: allow creating HttpsConnectionContext from SSLContext/Engine

raboof:
WiP, only scaladsl so far.

This seems rather verbose, and still depending on ssl-config (though
no longer AkkaSSLConfig). Is that what we want?

References #2845


raboof:
There seems to be something wrong with closing the TLSActor, for example running `TlsEndpointVerificationSpec` the test succeeds but logs:
```
Failed to stop [AkkaSpecWithMaterializer] within [10 seconds]
-> / LocalActorRefProvider$$anon$2 class akka.actor.LocalActorRefProvider$Guardian status=0 1 children
   ⌊-> system LocalActorRef class akka.actor.LocalActorRefProvider$SystemGuardian status=4 Terminating(Termination)
       |    toDie: Actor[akka://AkkaSpecWithMaterializer/system/StreamSupervisor-0#-1000633408]
       ⌊-> StreamSupervisor-0 RepointableActorRef class akka.stream.impl.StreamSupervisor status=4 Terminating(Termination)
           |    toDie: Actor[akka://AkkaSpecWithMaterializer/system/StreamSupervisor-0/flow-0-2#468195066]
           ⌊-> flow-0-2 RepointableActorRef class akka.stream.impl.io.TLSActor status=2 no children
```

akka-http-core/src/test/scala/akka/http/scaladsl/ClientServerSpec.scala
```diff
@@ -632,10 +633,28 @@ Host: example.com
       val clientSettings = ConnectionPoolSettings(system).withConnectionSettings(ClientConnectionSettings(system).withSocketOptions(SO.ReceiveBufferSize(serverToClientNetworkBufferSize) :: Nil))
 
       val serverConnectionContext = ExampleHttpContexts.exampleServerContext
-      // Disable hostname verification as ExampleHttpContexts.exampleClientContext sets hostname as akka.example.org
-      val sslConfigSettings = SSLConfigSettings().withLoose(SSLLooseConfig().withDisableHostnameVerification(true))
-      val sslConfig = AkkaSSLConfig().withSettings(sslConfigSettings)
-      val clientConnectionContext = ConnectionContext.https(ExampleHttpContexts.exampleClientContext.sslContext, Some(sslConfig))
+
+      // Create SSLEngine without hostname verification as ExampleHttpContexts.exampleClientContext sets hostname as akka.example.org
+      // Create SSLEngine without hostname verification as ExampleHttpContexts.exampleClientContext sets hostname as akka.example.org
```
jrudolph:
```suggestion
```

akka-http-core/src/test/scala/akka/http/scaladsl/ClientServerSpec.scala
```diff
@@ -632,10 +633,28 @@ Host: example.com
...
+      val context = {
```
jrudolph: Can we update `ExampleHttpContexts.exampleClientContext` instead to make that configurable (or provide a different way to make it look as if we were connecting to akka.example.org, e.g. using a custom ClientTransport)?

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
```diff
@@ -879,10 +879,32 @@ class HttpExt private[http] (private val config: Config)(implicit val system: Ex
   }
 
   /** Creates real or placebo SslTls stage based on if ConnectionContext is HTTPS or not. */
-  private[http] def sslTlsStage(connectionContext: ConnectionContext, role: TLSRole, hostInfo: Option[(String, Int)] = None) =
+  private[http] def sslTlsClientStage(connectionContext: ConnectionContext, host: String, port: Int) =
```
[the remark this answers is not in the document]
raboof: Splitting up made things easier to understand (at least to me) because it was clearer that host/port would be always provided for clients and never for servers. I agree the duplication doesn't make sense, I'll move that back to the original function but make it private.

raboof:
> There seems to be something wrong with closing the `TLSActor`, for example running `TlsEndpointVerificationSpec` the test succeeds but logs:

This is an upstream issue, https://github.com/akka/akka/issues/28691

akka-http-core/src/main/scala/akka/http/scaladsl/ConnectionContext.scala
```diff
@@ -40,20 +90,38 @@ object ConnectionContext {
...
+ */
+@InternalApi
 final class HttpsConnectionContext(
-  val sslContext:          SSLContext,
-  val sslConfig:           Option[AkkaSSLConfig]         = None,
-  val enabledCipherSuites: Option[immutable.Seq[String]] = None,
-  val enabledProtocols:    Option[immutable.Seq[String]] = None,
-  val clientAuth:          Option[TLSClientAuth]         = None,
-  val sslParameters:       Option[SSLParameters]         = None)
+  val sslContextData:      Either[SSLContext, Option[(String, Int)] => Engine],
```
jrudolph, line 107: Do we need that distinction here? Couldn't we change it to just use the engine factory and provide that directly from the context in the constructor?

jrudolph: I see, it might be for backwards compat?

akka-http-core/src/test/scala/akka/http/scaladsl/GracefulTerminationSpec.scala
```diff
@@ -205,13 +206,29 @@ class GracefulTerminationSpec
...
 
-    def nextRequest = HttpRequest(uri = s"https://$hostname:$port/${counter.incrementAndGet()}", entity = "hello-from-client")
+    def nextRequest() = HttpRequest(uri = s"https://$hostname:$port/${counter.incrementAndGet()}", entity = "hello-from-client")
 
     val serverConnectionContext = ExampleHttpContexts.exampleServerContext
-    // Disable hostname verification as ExampleHttpContexts.exampleClientContext sets hostname as akka.example.org
-    val sslConfigSettings = SSLConfigSettings().withLoose(SSLLooseConfig().withDisableHostnameVerification(true))
-    val sslConfig = AkkaSSLConfig().withSettings(sslConfigSettings)
-    val clientConnectionContext = ConnectionContext.https(ExampleHttpContexts.exampleClientContext.sslContext, Some(sslConfig))
+    // Create SSLEngine without hostname verification as ExampleHttpContexts.exampleClientContext sets hostname as akka.example.org
```
jrudolph, line 212: Maybe fix instead by using the custom ClientTransport from ExampleHttpContexts.

akka-http-core/src/main/scala/akka/http/scaladsl/ConnectionContext.scala
raboof: Yes - all the non-deprecated methods provide `Option[(String, Int)] => Engine`, but the `SSLContext` is here to keep the old behavior for when the deprecated API's have been called.

akka-http-core/src/test/scala/akka/http/scaladsl/ClientServerSpec.scala
```diff
@@ -632,10 +633,15 @@ Host: example.com
...
-      // Disable hostname verification as ExampleHttpContexts.exampleClientContext sets hostname as akka.example.org
-      val sslConfigSettings = SSLConfigSettings().withLoose(SSLLooseConfig().withDisableHostnameVerification(true))
-      val sslConfig = AkkaSSLConfig().withSettings(sslConfigSettings)
-      val clientConnectionContext = ConnectionContext.https(ExampleHttpContexts.exampleClientContext.sslContext, Some(sslConfig))
+
+      // Create SSLEngine without hostname verification as ExampleHttpContexts.exampleServerContext has a certificate
+      // that is valid for the 'akka.example.org' but we're connecting to 'localhost' - so we need to disable hostname verification.
+      val insecureSslEngineFactory = () => {
+        val engine = ExampleHttpContexts.exampleClientSSLContext.createSSLEngine()
+        engine.setUseClientMode(true)
```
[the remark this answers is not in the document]
raboof: We could, though that would obscure an error when the user tries to use a client-mode SSLEngine for a server-side connection or vice-versa.

Also, because SSLEngine is mutable, this could lead to confusing problems when people are reusing an SSLEngine. Since the SSLEngine is connection-specific anyway, though, that wouldn't work in the first place so that's probably fine.

I think I agree it's helpful to set this on 'our side'.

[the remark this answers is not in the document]
raboof: Then we should probably also do `engine.getSSLParameters.setEndpointIdentificationAlgorithm("https")` for them, and make using a custom ClientTransport the only way to disable hostname verification.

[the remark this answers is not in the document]
jrudolph: I think we should do as much as possible. E.g. I found that it matters when you call `engine.setUseClientMode`. If you do it to late, former parameter settings might be silently discarded (I noticed while testing client-side ALPN for HTTP/2).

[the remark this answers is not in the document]
raboof: > it matters when you call engine.setUseClientMode. If you do it to late, former parameter settings might be silently discarded (I noticed while testing client-side ALPN for HTTP/2)

Ouch - doesn't that mean we shouldn't set it as late as `sslTlsServerStage`, since that might parameters set by users?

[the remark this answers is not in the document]
raboof: ... and of course this is also different between jdk8 and jdk11. after out-of-band discussion: to be on the safe side, let's update the client mode in `sslTlsServerStage`, but only when needed.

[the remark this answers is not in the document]
raboof: Unfortunately 'only when needed' is not possible on JDK11: `getUseClientMode` does return a value when the mode isn't set yet, but if you then try to use the engine, it throws because the mode wasn't set yet :/.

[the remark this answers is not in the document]
jrudolph: Ok, this was thought to be a usability improvement, but if it's not possible, what can you do?

akka-http-core/src/main/scala/akka/http/scaladsl/ConnectionContext.scala
```diff
@@ -23,15 +28,60 @@ trait ConnectionContext extends akka.http.javadsl.ConnectionContext {
...
+          engine.getSSLParameters.setEndpointIdentificationAlgorithm("https")
+          engine.setUseClientMode(true)
+          engine
+        })
+      case Some((host, port)) =>
+        Engine(
+          () => {
+            val engine = context.createSSLEngine(host, port)
+            engine.getSSLParameters.setEndpointIdentificationAlgorithm("https")
+            engine.getSSLParameters.setServerNames(Collections.singletonList(new SNIHostName(host)))
```
jroper, line 64: As mentioned in #3255, making SNI enabled by default here and above is an important feature, just commenting here for posterity so there's a record of this decision being very intentional in the PR.

project/AkkaDependency.scala
```diff
@@ -42,9 +42,10 @@ object AkkaDependency {
   }
 
   // Default version updated only when needed, https://doc.akka.io//docs/akka/current/project/downstream-upgrade-strategy.html
-  val minimumExpectedAkkaVersion = "2.5.31"
+  // Needs https://github.com/akka/akka/pull/29148
+  val minimumExpectedAkkaVersion = "2.5-20200529-210210"
   val default = akkaDependency(defaultVersion = minimumExpectedAkkaVersion)
-  val minimumExpectedAkka26Version = "2.6.4"
+  val minimumExpectedAkka26Version = "2.6.5+106-33d64f9d"
```
jrudolph, line 48: Are those still needed?

akka-http-core/src/main/scala/akka/http/scaladsl/ConnectionContext.scala
```diff
@@ -24,31 +29,98 @@ trait ConnectionContext extends akka.http.javadsl.ConnectionContext {
...
+      })
+      case Some((host, port)) => Engine(() => {
+        val engine = sslContext.createSSLEngine(host, port)
+        engine.setUseClientMode(false)
+        engine
+      })
+    }))
+
+  def httpsClient(context: SSLContext)(implicit system: ClassicActorSystemProvider): HttpsConnectionContext = {
+    val verifier = new DefaultHostnameVerifier(new AkkaLoggerFactory(system.classicSystem))
```
jrudolph, line 50: Is there a specific reason we need this one?

akka-http-core/src/main/scala/akka/http/scaladsl/ConnectionContext.scala
jrudolph: Do you mean we should enable it explicitly? Usually, it isn't necessary to set it explicitly when using `context.createSSLEngine(host, port)`.

akka-http-core/src/main/scala/akka/http/scaladsl/ConnectionContext.scala
jrudolph: Revisting that part: is the only reason we need to keep the SSLContext to be able to call 

```
TLS(sslContext, connectionContext.sslConfig, ...
```

later on?

This TLS constructor also eventually calls the other one, so in theory, we could copy over the sslconfig related bits from there, right? That would allow us to decouple our deprecation from the one in Akka. On the other hand, 10.3.0 might be before Akka 2.7.0, so maybe it's not worth it.

jrudolph: Could it make sense otherwise, to include the `sslConfig` param in the left side of the either as well?

raboof: I agree this would be neater, but I think this is not possible because there would be one less default `Option` in the constructor, and while for normal methods you could add the previously-automatically-generated method for the missing generic, because this is a constructor the name is `<init>$default$6` which AFAICS you can't recreate...

akka-http-core/src/main/scala/akka/http/scaladsl/ConnectionContext.scala
raboof: I'm not sure what you mean - without this hostname verification would not work (demonstrated by TlsEndpointVerificationSpec). Is there another implementation that might be more suitable?

akka-http-core/src/main/scala/akka/http/scaladsl/ConnectionContext.scala
raboof: Hmm, it does seem so (in TlsEndpointVerificationSpec)

raboof: I just tried this both with jdk8 and jdk11, and indeed the SNI seems to be set even when this parameter is not explicitly set.

It seems to have been very deliberately set in https://github.com/akka/akka/pull/19858/files#diff-d2926806d32db424bb9b3e288fc1fb0cR159 though - which was also already using `createSSLEngine(hostname, port)`.

akka-http-core/src/main/scala/akka/http/scaladsl/ConnectionContext.scala
jrudolph: Hostname verification should work out of the box when `setEndpointVerificationMethod("https")` is used.

akka-http-core/src/main/scala/akka/http/scaladsl/ConnectionContext.scala
```diff
@@ -4,55 +4,158 @@
...
+            val engine = context.createSSLEngine()
+            engine.getSSLParameters.setEndpointIdentificationAlgorithm("https")
+            engine.setUseClientMode(true)
+            engine
+          })
+        case Some((host, port)) =>
+          Engine(
+            () => {
+              val engine = context.createSSLEngine(host, port)
+              engine.getSSLParameters.setEndpointIdentificationAlgorithm("https")
```
jrudolph, line 79: This doesn't work. You have to set the ssl parameters to apply them...

akka-http-core/src/main/scala/akka/http/scaladsl/ConnectionContext.scala
```diff
@@ -4,55 +4,158 @@
...
+   *  Use this sslContext to create a HttpsConnectionContext for client-side use.
+   */
+  def httpsClient(context: SSLContext)(implicit system: ClassicActorSystemProvider): HttpsConnectionContext = // ...
+    //#https-context-creation
+    {
+      val verifier = new DefaultHostnameVerifier(new AkkaLoggerFactory(system.classicSystem))
+
+      new HttpsConnectionContext(Right {
+        case None =>
+          Engine(() => {
```
jrudolph, line 69: What is this case? We shouldn't support it. If the host info is missing, we shouldn't create a client-side TLS stage. `setEndpointIdentificationAlgorithm` will not work without a host info.

akka-http-core/src/main/scala/akka/http/scaladsl/ConnectionContext.scala
jrudolph: If you do, the `DefaultHostnameVerifier` can be removed above.

akka-http-core/src/main/scala/akka/http/scaladsl/ConnectionContext.scala
```diff
@@ -4,55 +4,158 @@
...
+            engine.getSSLParameters.setEndpointIdentificationAlgorithm("https")
+            engine.setUseClientMode(true)
+            engine
+          })
+        case Some((host, port)) =>
+          Engine(
+            () => {
+              val engine = context.createSSLEngine(host, port)
+              engine.getSSLParameters.setEndpointIdentificationAlgorithm("https")
+              engine.getSSLParameters.setServerNames(Collections.singletonList(new SNIHostName(host)))
```
jrudolph, line 80: This is not needed. The SSLEngine will always add the hostname as given to `createSSLEngine(host, ...)` to the list of server names. Is it documented? You wish...

akka-http-core/src/main/scala/akka/http/scaladsl/ConnectionContext.scala
raboof: Verified it is set correctly automatically

akka-http-core/src/main/scala/akka/http/scaladsl/ConnectionContext.scala
raboof: Indeed :+1:

akka-http-core/src/main/scala/akka/http/scaladsl/ConnectionContext.scala
raboof: You're right. I've replaced it with throwing an exception.

That does beg the question whether the (private) type of `Engine` is correct, but it seems making that type more specific just moves the problem elsewhere: I guess the root of the problem is that a `HttpsConnectionContext` is server- or client-specific but that isn't visible in the type, so at some point you'll have to fail when using a 'server' `HttpsConnectionContext` as a client or a 'client' `HttpsConnectionContext` as a server.

akka-http-core/src/main/scala/akka/http/javadsl/ConnectionContext.scala
```diff
@@ -79,14 +93,19 @@ abstract class HttpsConnectionContext extends akka.http.javadsl.ConnectionContex
   override final def isSecure = true
 
   /** Java API */
+  @Deprecated @deprecated("here for binary compatibility", since = "10.2.0")
```
jrudolph, line 96: Do we have replacements for those methods? I guess we say, you have to provide a ready configured SSLContext / SSLEngine factory from now on, right? Might make sense to add that to the deprecation methods (only needed for the constructors).

akka-http-core/src/main/scala/akka/http/javadsl/ConnectionContext.scala
```diff
@@ -5,26 +5,38 @@
...
 import akka.japi.Util
 import akka.stream.TLSClientAuth
-import com.typesafe.sslconfig.akka.AkkaSSLConfig
-import javax.net.ssl.{ SSLContext, SSLParameters }
 
 import scala.compat.java8.OptionConverters
 
 object ConnectionContext {
   //#https-context-creation
+  def httpsClient(sslContext: SSLContext): HttpsConnectionContext = // ...
```
raboof, line 22: Should we create a variant without parameters that just takes `SSLContext.getDefault()`?

akka-http-core/src/main/scala/akka/http/javadsl/ConnectionContext.scala
raboof: I think https://github.com/akka/akka-http/pull/2983/files#diff-f7c7e05e650a31f0e256ceaa3bca12a3R127 is the only one, made it a bit more complete

akka-http-core/src/main/scala/akka/http/javadsl/ConnectionContext.scala
jrudolph: Would there be a reason to use it? Wouldn't that be used if you don't configure anything else?

raboof: Ah, good point - I think the default currently still uses ssl-config.

I agree we should probably switch that to `ConnectionContext.https(Client|Server)(SSLContext.getDefault())` and call that out explicitly in the migration docs.

Or do you think that's a too invasive change in behaviour? I guess we could do a "if there is config, warn and use ssl-config, otherwise use SSLContext.getDefault()"?

jrudolph: > Or do you think that's a too invasive change in behaviour? I guess we could do a "if there is config, warn and use ssl-config, otherwise use SSLContext.getDefault()"?

How would we figure out "if there is config"? If that's possible, it would be nice to make that distinction.

akka-http-core/src/main/scala/akka/http/scaladsl/ConnectionContext.scala
```diff
@@ -4,55 +4,151 @@
...
 object ConnectionContext {
-  // ConnectionContext
+  /**
+   *  If you want complete control over how to create the SSLEngine you can use this method:
+   *
+   *  Note that this means it is up to you to make sure features like SNI and Hostname Verification
+   *  are enabled as needed.
+   */
+  @ApiMayChange
+  def https(createSSLEngine: Option[(String, Int)] => () => SSLEngine): HttpsConnectionContext = // ...
```
jrudolph, line 36: Needs docs about the function argument. Or should we instead provide overloads for` httpsServer` and `httpsClient` so we can get rid of the `Option`?

project/AkkaDependency.scala
```diff
@@ -42,9 +42,11 @@ object AkkaDependency {
   }
 
   // Default version updated only when needed, https://doc.akka.io//docs/akka/current/project/downstream-upgrade-strategy.html
-  val minimumExpectedAkkaVersion = "2.5.31"
+  // Needs https://github.com/akka/akka/pull/29148
```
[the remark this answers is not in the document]
raboof: I think a test failed, but we can probably make that select TLS1.2. OK for RC, but would be good to update before 10.2.0 GA

docs/src/main/paradox/migration-guide/migration-guide-10.2.x.md
```diff
@@ -167,3 +167,27 @@ Only parts of `Encoder`, `Decoder`, and `Coder` are still public. Predefined cod
...
+by default. Instead, it will use the JRE defaults for client connections. For
+server connections you should create a @apidoc[HttpsConnectionContext] with
+the relevant configuration.
+
+Previously, the SSL configuration was constructed based on the parameters
+you pass to @apidoc[HttpsConnectionContext$] and the logic and configuration
+used by the [ssl-config](https://lightbend.github.io/ssl-config/) library. This was
+because back then the default JDK SSL configuration was a bad match for HTTP
+connections. Nowadays, however, the JDK defaults are much better and the
+introduction of `ssl-config` makes the configuration hard to reason about.
```
ignasi35, line 183:
```suggestion
use of `ssl-config` makes the configuration hard to reason about.
```

docs/src/main/paradox/server-side/server-https-support.md
```diff
@@ -46,23 +46,10 @@ Scala
...
 :  @@snip [SimpleServerApp.java](/akka-http-tests/src/main/java/akka/http/javadsl/server/examples/simple/SimpleServerApp.java) { #https-http-config }
 
-@scala[Once you configured the HTTPS context, you can set it as default:]
-@java[Then pass it to the `akka.http.javadsl.Http` class's `setDefaultServerHttpContext` method, like in the below `main` method.]
-
-Scala
-:  @@snip [HttpsServerExampleSpec.scala]($test$/scala/docs/http/scaladsl/server/HttpsServerExampleSpec.scala) { #set-low-level-context-default }
-
-Java
-: @@snip [SimpleServerApp.java](/akka-http-tests/src/main/java/akka/http/javadsl/server/examples/simple/SimpleServerApp.java) { #https-http-app }
```
ignasi35, line 56: There's code somewhere with a `// #https-http-app` comment that we can probably remove (or at least remove the comment ;-) )

ignasi35: Looks like https://github.com/akka/akka-http/blob/ea534cca2cb0a99e7121df92adfd2302291c224e/akka-http-tests/src/main/java/akka/http/javadsl/server/examples/simple/SimpleServerApp.java#L44-L129

docs/src/main/paradox/compatibility-guidelines.md
```diff
@@ -61,6 +61,8 @@ Java
 Scala
 :   ```scala
     akka.http.scaladsl.ClientTransport
+    akka.http.scaladsl.ConnectionContext#httpsClient
+    akka.http.scaladsl.ConnectionContext#httpsServer
```
ignasi35, line 65: Are the javadsl equivalents missing on the list or is it some scaladsl-specific API change?

jrudolph: Now added, thanks.

PR #2984, opened 2020-02-28
Require Akka 2.6

### LINKED ISSUE #2886: Promote Akka 2.6

raboof:
Refs #2886

This PR only does the mostly mechanical changes: updating,
removing ActorMaterializer references where possible,
and propagating the onDownstreamFinish cancellation reason.

Before we close #2886 we should updating the Java API to work
without an explicit Materializer, but that could be a separate
PR

### DIFF
--- a/akka-http-bench-jmh/src/main/scala/akka/http/impl/engine/HeaderParserBenchmark.scala
+++ b/akka-http-bench-jmh/src/main/scala/akka/http/impl/engine/HeaderParserBenchmark.scala
@@ -16,7 +16,7 @@ import scala.concurrent.duration._
 @State(Scope.Benchmark)
 @OutputTimeUnit(TimeUnit.MILLISECONDS)
 @BenchmarkMode(Array(Mode.Throughput))
-class HeaderParserBenchmark {
+final class HeaderParserBenchmark {
   implicit val system: ActorSystem = ActorSystem("header-parser-benchmark")
 
   @Param(Array("no", "yes"))
--- a/akka-http-bench-jmh/src/main/scala/akka/http/impl/engine/ServerProcessingBenchmark.scala
+++ b/akka-http-bench-jmh/src/main/scala/akka/http/impl/engine/ServerProcessingBenchmark.scala
@@ -10,7 +10,6 @@ import akka.http.scaladsl.model.HttpRequest
 import akka.http.scaladsl.model.HttpResponse
 import akka.http.scaladsl.model.headers
 import akka.http.scaladsl.settings.ServerSettings
-import akka.stream.ActorMaterializer
 import akka.stream.scaladsl.Flow
 import akka.stream.scaladsl.Source
 import akka.stream.scaladsl.TLSPlacebo
@@ -24,7 +23,6 @@ class ServerProcessingBenchmark extends CommonBenchmark {
 
   var httpFlow: Flow[ByteString, ByteString, Any] = _
   implicit var system: ActorSystem = _
-  implicit var mat: ActorMaterializer = _
 
   @Benchmark
   @OperationsPerInvocation(10000)
@@ -48,7 +46,6 @@ class ServerProcessingBenchmark extends CommonBenchmark {
         """)
         .withFallback(ConfigFactory.load())
     system = ActorSystem("AkkaHttpBenchmarkSystem", config)
-    mat = ActorMaterializer()
     httpFlow =
       Flow[HttpRequest].map(_ => response) join
         (HttpServerBluePrint(ServerSettings(system), NoLogging, false) atop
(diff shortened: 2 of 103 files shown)


jrudolph:
Could we do this in multiple steps, so we could have a more complete picture about what it would mean if we would stay on 2.5.29 which now also has the SystemMaterializer? If we postpone the propagation of cancellation causes, it seems the rest of the changes would mostly still work (though, you'd still need an implicit Materializer in scope for some usages).

raboof:
> If we postpone the propagation of cancellation causes, it seems the rest of the changes would mostly still work (though, you'd still need an implicit Materializer in scope for some usages).

Since we didn't backport the implicit conversion from ClassicActorSystemProvider to Materializer to 2.5 (I think we figured it'd be too much of a risk of a behavior change, possibly source-incompatible in some cases), actually I'm not sure any of the changes in 368727e would apply to 2.5.29...

jrudolph:
> Since we didn't backport the implicit conversion from ClassicActorSystemProvider to Materializer to 2.5 (I think we figured it'd be too much of a risk of a behavior change, possibly source-incompatible in some cases)

That's right, the implicit is there but we didn't backport akka.actor.ActorSystem implementing `ClassicActorSystemProvider` so you still need the ActorMaterializer in 2.5.

But there are only very few changes to the core classes so far, so we might still be able to stay on 2.5 while making sure the usages are idomatic with 2.6 without requiring it. We could pull up the dependency to 2.6 in our docs if we decided to go that way.

You see I'm still hesitant to make the final call :) Let's think a bit more about what would be enabled by dropping 2.5 support:

 * Code base cleanup and less complexity (next chance to drop it will only be in 10.3.0 which might be > 1 year out)
 * Propagate cancellation causes
 * Enabling internal dispatchers (this is also interesting from a performance point of view)
 * Easier adoption of future 2.6.x features without putting pressure on the 2.5.x line to backport them

What else?

raboof:
Yes, for me the main advantages are:
* Propagate cancellation causes
* Not needing to show the Materializer in docs (as long as we support 2.5 perhaps we should show providing the Materializer, otherwise it'd be confusing?)
* More safely showing the final Akka Typed API's in docs
* Less user confusion ('is akka-http released for akka 2.6 yet?')

> You see I'm still hesitant to make the final call :)

If we're not comfortable I'm fine with staying on 2.5 for 10.2.x. 

> We could pull up the dependency to 2.6 in our docs if we decided to go that way.

That's an interesting idea, though wouldn't that be confusing, as then the examples wouldn't work unless users also explicitly pull in Akka 2.6?

jrudolph:
Good point, I guess we should talk also about goals and not just about solutions

 * We want to make sure the migration process to 10.2 is straight-forward. That requires:
   * binary compatibility (+ source compatibility if possible)
   * only reasonable extra constraints (e.g. like having to update to 2.6)
 * We will have to support the 10.1.x line for a while, but we'd like to be able to keep work minimal
 * We want to give a strong signal (especially to new users) that you should use Akka 2.6

Akka-Http has functionality more orthogonal than how other features relate inside of akka core. Enabling more incremental updates for such a component seems to be a good practice (if possible with reasonable effort).

Addressing your points:

> * Propagate cancellation causes

That's somewhat obviously a good thing in itself. But do we require it for any concrete feature or fix right now?

> * Not needing to show the Materializer in docs (as long as we support 2.5 perhaps we should show providing the Materializer, otherwise it'd be confusing?)
> * More safely showing the final Akka Typed API's in docs
> 
> * Less user confusion ('is akka-http released for akka 2.6 yet?')
> > We could pull up the dependency to 2.6 in our docs if we decided to go that way.
> 
> That's an interesting idea, though wouldn't that be confusing, as then the examples wouldn't work unless users also explicitly pull in Akka 2.6?

All these are more about our documentation and messaging. And I think we should be very clear there that Akka 2.6 is the way to go. I think we cannot and shouldn't go as far providing separate documentation for 2.5 and 2.6. Instead, we could migrate everything to 2.6 usages and add a single new page that explains what changes are necessary when you are stuck on 2.5 (maybe linking to old 10.1 documentation but I'm even wary about that). We could then add notes pointing to that page for content or code snippets that would need those code changes.

That would mean new users would be on the right track right away. Existing code bases could continue to compile and run (source and binary compatibility) and could be easily migrated following  deprecation hints or scalafix rules we could provide. The only disadvantaged developers would be ones newly coming into a 2.5 project and having to write new code that still uses the old syntax.

raboof:
>> Propagate cancellation causes
>
> But do we require it for any concrete feature or fix right now?

I don't have a clear picture of it, but I was kind of hoping it would allow us to fix some of the failure/cancellation races in a more reliable way.

> I think we cannot and shouldn't go as far providing separate documentation for 2.5 and 2.6.

Agreed

> we could migrate everything to 2.6 usages and add a single new page that explains what changes are necessary when you are stuck on 2.5

I guess that could work, especially when we make sure we show the Akka 2.6 dependency alongside the Akka HTTP dependencies in boxes like at https://doc.akka.io/docs/akka-http/current/introduction.html#using-akka-http

raboof:
Closing this since #3036 means we can play nice with 2.6 users without forcing it, and #2996 allows showing that in docs.

jrudolph:
Great :+1:

PR #2996, opened 2020-03-04
Update docs to Akka 2.6, de-emphasize Materializer

raboof:
This contains part of #2984, focusing on the Scala API's for now. It currently fails because Akka 2.5.29 doesn't have https://github.com/akka/akka/pull/28545 yet.

### DIFF
--- a/akka-http-core/src/test/scala/akka/http/impl/engine/client/ClientCancellationSpec.scala
+++ b/akka-http-core/src/test/scala/akka/http/impl/engine/client/ClientCancellationSpec.scala
@@ -12,13 +12,14 @@ import scala.concurrent.duration._
 import akka.http.impl.util.AkkaSpecWithMaterializer
 import akka.http.scaladsl.{ ConnectionContext, Http }
 import akka.http.scaladsl.model.{ HttpRequest, HttpResponse }
-import akka.stream.ActorMaterializer
+import akka.stream.SystemMaterializer
 import akka.stream.scaladsl.{ Flow, Sink, Source }
 import akka.stream.testkit.{ TestPublisher, TestSubscriber, Utils }
 import akka.http.scaladsl.model.headers
 
 class ClientCancellationSpec extends AkkaSpecWithMaterializer {
-  val noncheckedMaterializer = ActorMaterializer()
+  // TODO document why this explicit materializer is needed here?
+  val noncheckedMaterializer = SystemMaterializer(system).materializer
 
   "Http client connections" must {
     val address = Await.result(
--- a/akka-http-marshallers-scala/akka-http-spray-json/src/test/scala/akka/http/scaladsl/marshallers/sprayjson/SprayJsonSupportSpec.scala
+++ b/akka-http-marshallers-scala/akka-http-spray-json/src/test/scala/akka/http/scaladsl/marshallers/sprayjson/SprayJsonSupportSpec.scala
@@ -4,6 +4,8 @@
 
 package akka.http.scaladsl.marshallers.sprayjson
 
+import scala.concurrent.ExecutionContext
+
 import akka.actor.ActorSystem
 import akka.http.scaladsl.marshalling.Marshal
 import akka.http.scaladsl.model.MessageEntity
@@ -24,7 +26,7 @@ class SprayJsonSupportSpec extends AnyWordSpec with Matchers with ScalaFutures {
   implicit val exampleFormat = jsonFormat1(Example.apply)
   implicit val sys = ActorSystem("SprayJsonSupportSpec")
   implicit val mat = ActorMaterializer()
-  import sys.dispatcher
+  implicit val ec: ExecutionContext = sys.dispatcher
 
   val TestString = "Contains all UTF-8 characters: 2-byte: £, 3-byte: ﾖ, 4-byte: 😁, 4-byte as a literal surrogate pair: \uD83D\uDE01"
 
--- a/akka-http-testkit/src/main/scala/akka/http/javadsl/testkit/JUnitRouteTest.scala
+++ b/akka-http-testkit/src/main/scala/akka/http/javadsl/testkit/JUnitRouteTest.scala
@@ -8,7 +8,8 @@ import akka.actor.ActorSystem
 import akka.event.Logging
 import akka.http.javadsl.model.HttpRequest
 import akka.http.javadsl.server._
-import akka.stream.{ Materializer, ActorMaterializer }
+import akka.stream.Materializer
+import akka.stream.SystemMaterializer
 import com.typesafe.config.{ ConfigFactory, Config }
 import org.junit.rules.ExternalResource
 import org.junit.{ Assert, Rule }
@@ -60,22 +61,18 @@ abstract class JUnitRouteTest extends JUnitRouteTestBase {
 class ActorSystemResource(name: String, additionalConfig: Config) extends ExternalResource {
   protected def config = additionalConfig.withFallback(ConfigFactory.load())
   protected def createSystem(): ActorSystem = ActorSystem(name, config)
-  protected def createMaterializer(system: ActorSystem): ActorMaterializer = ActorMaterializer()(system)
 
   implicit def system: ActorSystem = _system
-  implicit def materializer: ActorMaterializer = _materializer
+  implicit def materializer: Materializer = SystemMaterializer.get(system).materializer
 
   private[this] var _system: ActorSystem = null
-  private[this] var _materializer: ActorMaterializer = null
 
   override def before(): Unit = {
-    require((_system eq null) && (_materializer eq null))
+    require(_system eq null)
     _system = createSystem()
-    _materializer = createMaterializer(_system)
   }
   override def after(): Unit = {
     Await.result(_system.terminate(), 5.seconds)
     _system = null
-    _materializer = null
   }
 }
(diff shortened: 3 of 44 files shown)


akka-http-core/src/test/scala/akka/http/impl/engine/client/ClientCancellationSpec.scala
```diff
@@ -12,13 +12,14 @@ import scala.concurrent.duration._
...
 import akka.http.scaladsl.model.{ HttpRequest, HttpResponse }
-import akka.stream.ActorMaterializer
+import akka.stream.SystemMaterializer
 import akka.stream.scaladsl.{ Flow, Sink, Source }
 import akka.stream.testkit.{ TestPublisher, TestSubscriber, Utils }
 import akka.http.scaladsl.model.headers
 
 class ClientCancellationSpec extends AkkaSpecWithMaterializer {
-  val noncheckedMaterializer = ActorMaterializer()
+  // TODO document why this explicit materializer is needed here?
```
jrudolph: Seems to be pretty old code, maybe just remove it and the usage and see if something breaks?

akka-http-marshallers-scala/akka-http-spray-json/src/test/scala/akka/http/scaladsl/marshallers/sprayjson/SprayJsonSupportSpec.scala
```diff
@@ -24,7 +26,7 @@ class SprayJsonSupportSpec extends AnyWordSpec with Matchers with ScalaFutures {
   implicit val exampleFormat = jsonFormat1(Example.apply)
   implicit val sys = ActorSystem("SprayJsonSupportSpec")
   implicit val mat = ActorMaterializer()
```
jrudolph:
```suggestion
```

project/AkkaDependency.scala
```diff
@@ -8,64 +8,60 @@ import sbt._
...
-
-  val akkaRepository = {
-    // as a little hacky side effect also disable aggregation of samples
-    System.setProperty("akka.build.aggregateSamples", "false")
+  // Default version updated only when needed, https://doc.akka.io//docs/akka/current/project/downstream-upgrade-strategy.html
+  val default = akkaDependency(defaultVersion = "2.5.30")
+  val docs = akkaDependency(defaultVersion = "2.6.4")
 
-    uri(akkaSourceDependencyUri)
+  val akkaVersion: String = default match {
```
jrudolph: Has the value changed to before? Have you checked where this is used? It probably doesn't matter but would make sense to check.

project/AkkaDependency.scala
```diff
@@ -8,64 +8,60 @@ import sbt._
...
-            val withConfig =
-              if (config == "") dep
-              else dep % config
-            withConfig
-          })
+            project.dependsOn(withConfig)
+          case Artifact(akkaVersion) =>
+            project.settings(libraryDependencies += {
+              val dep = "com.typesafe.akka" %% module % akkaVersion
+                val withConfig =
```
jrudolph: Indentation a bit messed up.

project/AkkaDependency.scala
```diff
@@ -8,64 +8,60 @@ import sbt._
...
-
-  // Needs to be a URI like git://github.com/akka/akka.git#master or file:///xyz/akka
-  val akkaSourceDependencyUri = {
-    val fallback =
-      akkaVersion match {
-        case "master" => "git://github.com/akka/akka.git#master"
-        case "release-2.5" => "git://github.com/akka/akka.git#release-2.5"
-        case x => ""
-      }
+  sealed trait Akka
```
jrudolph: :+1: Thanks for that cleanup. It was long due!

jrudolph: We need to make sure nightlies still work after that.

docs/src/main/paradox/routing-dsl/index.md
```diff
@@ -105,7 +105,7 @@ Scala
 
 
 Note that the `akka.actor.typed.ActorSystem` is converted with `toClassic`, which comes from
-`import akka.actor.typed.scaladsl.adapter._`. If you are using an earlier version than Akka 2.5.26 this conversion method is named `toUntyped`.
+`import akka.actor.typed.scaladsl.adapter._`.
```
jrudolph: Ok, so we'll touch that again with 2.5.31 where even that won't be needed any more.

akka-http2-support/src/test/scala/akka/http/scaladsl/Http2ServerTest.scala
```diff
@@ -37,13 +38,8 @@ object Http2ServerTest extends App {
...
   implicit val system = ActorSystem("ServerTest", testConf)
-  import system.dispatcher
-
-  val settings = ActorMaterializerSettings(system)
-    .withFuzzing(false)
-    //    .withSyncProcessingLimit(Int.MaxValue)
-    .withInputBuffer(128, 128)
-  implicit val fm = ActorMaterializer(settings)
+  implicit val ec: ExecutionContext = system.dispatcher
+  implicit val fm = ActorMaterializer()
```
jrudolph: I guess here it's still needed because we materialize streams which needs an explicit implicit materializer until 2.5.31. :+1:

ennru:
I suggest we wait with merging this and make first use of the `ClassicActorSystemProvider` changes from Akka 2.5.31, so that they get run against 2.5.

ennru:
Added #3036 to show the tests work with Akka 2.5.

akka-http2-support/src/test/scala/akka/http/scaladsl/Http2ServerTest.scala
ignasi35: Hmm, I just tested locally replacing `ActorMaterializer()` with `SystemMaterializer(system).materializer` and works just fine. Do we really need the `ActorMaterializer()`?

akka-http-testkit/src/main/scala/akka/http/javadsl/testkit/JUnitRouteTest.scala
```diff
@@ -60,22 +61,18 @@ abstract class JUnitRouteTest extends JUnitRouteTestBase {
...
   implicit def system: ActorSystem = _system
-  implicit def materializer: ActorMaterializer = _materializer
+  implicit def materializer: Materializer = SystemMaterializer.get(system).materializer
 
   private[this] var _system: ActorSystem = null
-  private[this] var _materializer: ActorMaterializer = null
 
   override def before(): Unit = {
-    require((_system eq null) && (_materializer eq null))
+    require(_system eq null)
```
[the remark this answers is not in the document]
raboof: Indeed this should normally never happen, but perhaps when someone makes a mistake and assigns to it from async code or something? Seems like a pretty hard-to-hit safeguard indeed.

akka-http2-support/src/test/scala/akka/http/scaladsl/Http2ServerTest.scala
raboof: Indeed an intermediate solution might be to use SystemMaterializer here, but we'd really want to remove the line - let's see if with 2.5.31 we can.

akka-http/src/main/scala/akka/http/scaladsl/server/Route.scala
```diff
@@ -56,8 +56,8 @@ object Route {
    *
    * This conversion is also implicitly available whereever a `Route` is used through [[RouteResult#routeToFlow]].
    */
-  def toFlow(route: Route)(implicit system: ActorSystem): Flow[HttpRequest, HttpResponse, NotUsed] =
-    Flow[HttpRequest].mapAsync(1)(asyncHandler(route, RoutingSettings(system), ParserSettings.forServer(system)))
+  def toFlow(route: Route)(implicit system: ClassicActorSystemProvider): Flow[HttpRequest, HttpResponse, NotUsed] =
+    Flow[HttpRequest].mapAsync(1)(asyncHandler(route, RoutingSettings(system.classicSystem), ParserSettings.forServer(system.classicSystem)))
```
raboof, line 60: I wonder if we also want `ClassicActorSystemProvider`-based implicits for `SettingsCompanion` /cc @ennru

ennru: The `apply(ActorSystem)` should get an overload with the provider.
Only the `ActorRefFactory` is accepted implicitly, I'm not sure what we should do there.

jrudolph: > I wonder if we also want `ClassicActorSystemProvider`-based implicits for `SettingsCompanion` /cc @ennru

Yes, that seems to be a commonly used API that should work seemlessly with 2.6.

jrudolph: >  Only the `ActorRefFactory` is accepted implicitly, I'm not sure what we should do there.

Not sure what that is about?

ennru: In `SettingsCompanion` the `ActorRefFactory` is the implicit. I assume we need to change it to take `ClassicActorContextProvider` implicitly instead.

raboof: hmm that would mean you really need the actor system and the actor context would not be sufficient anymore - since the actor context never has a different configuration than the actor system that seems fine to me, though. I'll update this PR.

Issue #2961, opened 2020-02-12
Default entity size limit for client


johanandren:
The current client size limit is inherited from the default parser settings, where it is set to 8 Mb. Isn't that an unreasonably low limit for a client? (I can see how it makes sense on the server side)

I propose we increase it quite a bit or perhaps even disable it by default.

jrudolph:
I agree it is probably more important on the server. In the past, we debated whether those limits were useful in a streaming setting at all. In the end, you will probably be also be guarded by the `toStrict` limits before loading stuff into memory. The original reasoning was to have limits for everything for more defense in depth but I can see how people would be confused about that on the client side. Maybe we should just disable it there? WDYT, @raboof?

raboof:
Sounds reasonable to me!

seglo:
I would vote to remove it on the client side.  This issue has come and and been worked around several times in Alpakka by disabling the default limit.

References (GCS and S3)
* https://github.com/akka/alpakka/issues/2084
* https://github.com/akka/alpakka/issues/303

PR #2998, opened 2020-03-06
Drop client-side parsing.max-content-length by default

raboof:
Refs #2961

### DIFF
--- a/akka-http-core/src/main/resources/reference.conf
+++ b/akka-http-core/src/main/resources/reference.conf
@@ -463,7 +463,9 @@ akka.http {
     #
     # Set to `infinite` to completely disable entity length checks. (Even then you can still apply one
     # programmatically via `withSizeLimit`.)
-    max-content-length = 8m
+    #
+    # A value of "default" translates to "infinite" for clients and "8m" for servers
+    max-content-length = default
 
     # The maximum number of bytes to allow when reading the entire entity into memory with `toStrict`
     # (which is used by the `toStrictEntity` and `extractStrictEntity` directives)
--- a/akka-http-core/src/main/scala/akka/http/impl/settings/ParserSettingsImpl.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/settings/ParserSettingsImpl.scala
@@ -82,7 +82,7 @@ object ParserSettingsImpl extends SettingsCompanionImpl[ParserSettingsImpl]("akk
       c.getIntBytes("max-header-name-length"),
       c.getIntBytes("max-header-value-length"),
       c.getIntBytes("max-header-count"),
-      c.getPossiblyInfiniteBytes("max-content-length", defaultMaxContentLength),
+      c.getPossiblyInfiniteBytesOr("max-content-length", defaultMaxContentLength),
       c.getPossiblyInfiniteBytes("max-to-strict-bytes"),
       c.getIntBytes("max-chunk-ext-length"),
       c.getIntBytes("max-chunk-size"),
--- a/akka-http-core/src/main/scala/akka/http/impl/settings/ParserSettingsImpl.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/settings/ParserSettingsImpl.scala
@@ -64,12 +64,15 @@ object ParserSettingsImpl extends SettingsCompanionImpl[ParserSettingsImpl]("akk
   private[this] val noCustomStatusCodes: Int => Option[StatusCode] = ConstantFun.scalaAnyToNone
   private[ParserSettingsImpl] val noCustomMediaTypes: (String, String) => Option[MediaType] = ConstantFun.scalaAnyTwoToNone
 
+  val defaultMaxContentLengthServer = 8 * 1024 * 1024
+  val defaultMaxContentLengthClient = Long.MaxValue
+
   def forServer(root: Config): ParserSettingsImpl =
-    fromSubConfig(root, root.getConfig("akka.http.server.parsing"))
+    fromSubConfig(root, root.getConfig("akka.http.server.parsing"), defaultMaxContentLengthServer)
 
   @deprecated("please supply a default value for max-content-length")
   def fromSubConfig(root: Config, inner: Config): ParserSettingsImpl =
-    fromSubConfig(root, inner, 8 * 1000 * 1000)
+    fromSubConfig(root, inner, defaultMaxContentLengthServer)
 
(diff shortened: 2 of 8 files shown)


akka-http-core/src/main/resources/reference.conf
```diff
@@ -463,7 +463,9 @@ akka.http {
     #
     # Set to `infinite` to completely disable entity length checks. (Even then you can still apply one
     # programmatically via `withSizeLimit`.)
-    max-content-length = 8m
+    #
+    # A value of "default" translates to "infinite" for clients and "8m" for servers
```
ennru: Is "8m" correct? https://github.com/lightbend/config/blob/master/HOCON.md#size-in-bytes-format

akka-http-core/src/main/scala/akka/http/impl/settings/ServerSettingsImpl.scala
```diff
@@ -98,7 +98,7 @@ private[http] object ServerSettingsImpl extends SettingsCompanionImpl[ServerSett
           throw new ConfigurationException(info.formatPretty)
       },
     WebSocketSettingsImpl.server(c.getConfig("websocket")),
-    ParserSettingsImpl.fromSubConfig(root, c.getConfig("parsing")),
+    ParserSettingsImpl.fromSubConfig(root, c.getConfig("parsing"), 8 * 1000 * 1000),
```
ennru: Should that magical number live in a constant?

akka-http-core/src/main/resources/reference.conf
raboof: yes, 8 mebibytes...

ennru: Hah, you changed to "1024 * 1024" when you made it a constant.

akka-http-core/src/test/scala/akka/http/impl/engine/client/HttpConfigurationSpec.scala
```diff
@@ -16,10 +16,12 @@ class HttpConfigurationSpec extends AkkaSpec {
 
   "Reference configurations" should {
     "have default client and server `parsing` settings" in {
-      ServerSettings(system).parserSettings.toString should ===(ClientConnectionSettings(system).parserSettings.toString)
+      ServerSettings(system).parserSettings.withMaxContentLength(Long.MaxValue).toString should ===(ClientConnectionSettings(system).parserSettings.toString)
+      ServerSettings(system).parserSettings.toString should ===(ClientConnectionSettings(system).parserSettings.withMaxContentLength(8 * 1000 * 1000).toString)
```
ennru: The tests fail as you changed to the "8MiB" instead of "8MB".

akka-http-core/src/main/scala/akka/http/impl/settings/ParserSettingsImpl.scala
```diff
@@ -64,7 +64,14 @@ object ParserSettingsImpl extends SettingsCompanionImpl[ParserSettingsImpl]("akk
   private[this] val noCustomStatusCodes: Int => Option[StatusCode] = ConstantFun.scalaAnyToNone
   private[ParserSettingsImpl] val noCustomMediaTypes: (String, String) => Option[MediaType] = ConstantFun.scalaAnyTwoToNone
 
-  def fromSubConfig(root: Config, inner: Config): ParserSettingsImpl = {
+  val defaultMaxContentLengthServer = 8 * 1024 * 1024
+  val defaultMaxContentLengthClient = Long.MaxValue
+
+  @deprecated("please supply a default value for max-content-length")
```
ennru: This needs a `since` marker.

docs/src/main/paradox/release-notes/10.2.x.md
```diff
@@ -11,4 +11,14 @@ versions. This is relevant for user code if it uses methods from @scaladoc[Scala
 (which are in scope if your test extends from @scaladoc[ScalaTestRouteTest](akka.http.scaladsl.testkit.ScalaTestRouteTest)).
 In this case, the project itself needs to update to use Scalatest >= 3.1.0.
 
+#### Default parsing.max-content-length
+
+Previously, the default maximum size for incoming entities was 8m. This default has now changed to
+'infinite' for the client side (response parsing), while it remains at 8m for the server side
+(request parsing).
+
+Changing `akka.http.parsing.max-content-length`affects both the client and the server, while
```
ennru:
```suggestion
Changing `akka.http.parsing.max-content-length` affects both the client and the server, while
```

akka-http-core/src/main/resources/reference.conf
jrudolph: Hmm, wouldn't it be more straightforward to change it in the included sections? E.g. we could either comment it out here and set it separately for client and server. Or we keep it as it was here and override it by default in the client section?

raboof:
> I'd prefer to do it in the configuration directly, i.e. to have different default values for `akka.http.server.parsing.max-content-length` and `akka.http.client.parsing.max-content-length`. Is that impossible for some reason?

I would also prefer that, but I'm not sure how.

The question is what to do with `akka.http.parsing.max-content-length`. The 'logical' thing to do would be to remove it, and only have `akka.http.client.parsing.max-content-length` and `akka.http.server.parsing.max-content-length`. That is tricky, however, since we have quite a few places where we use the 'bare' `akka.http.parsing` settings (somewhat related to #2999 - not sure if max-content-length is ever used in the remaining places, but at least it's confusing).

The alternative is to keep `akka.http.parsing.max-content-length`, but then it's confusing for end users that changing `akka.http.parsing.max-content-length` does not work in most cases, since `akka.http.server.parsing.max-content-length`/`akka.http.client.parsing.max-content-length` overrides it.

Neither seem particularly nice, so that's why I ended up with the 'default' value (which I agree is not particularly nice either...)

raboof:
> The question is what to do with `akka.http.parsing.max-content-length`.

The third option is to make it an `Option`, moving the defaults to `client`/`server`/etc and providing a run-time error when it is missing where it is expected. @jrudolph I think you started on that? It sounded like that was working out nicely - should we PR that and close this one?

jrudolph:
> The third option is to make it an `Option`, moving the defaults to `client`/`server`/etc and providing a run-time error when it is missing where it is expected. @jrudolph I think you started on that? It sounded like that was working out nicely - should we PR that and close this one?

I'll have another look at that. Let's keep this open until we have decided that the alternative way would really work.

ennru:
This can be closed in favour of #3098, right?

raboof:
Yes, I think so

Issue #3002, opened 2020-03-09
Inconsistency in the escaping of the query string in the Uri API


raboof:
A double quote character is invalid both in the query string and in the fragment, however we treat them differently:

```
scala> Uri(queryString = Some("foo\"bar"), fragment = Some("foo\"bar")).toString()
res1: String = ?foo"bar#foo%22bar

scala> Uri(queryString = Some("foo%22bar"), fragment = Some("foo%22bar")).toString()
res2: String = ?foo%22bar#foo%2522bar

scala> Uri("http://foo.example/?foo\"bar#foo\"bar").toString()
res5: String = http://foo.example/?foo"bar#foo%22bar

scala> Uri("http://foo.example/?foo%22bar#foo%22bar").toString()
res4: String = http://foo.example/?foo%22bar#foo%22bar
```

What's going on here is that for query strings we store the 'raw' query string (with any encodings 'intact'), while for fragments we store the 'evaluated' string (and re-apply any encodings when rendering the URI again). This explains why fragments that contain a double quote get converted into a %22 while the same does not happen for query strings.

## Uri parsing from string

The behavior of the `Uri` constructor that takes a string is surprising even though it is at least consistent for correct input. We used to parse the query string into 'evaluated' key-value pairs, but found situations where we'd like to keep query strings that are not key-value pairs intact in https://github.com/akka/akka/issues/18479 - however later in #1120 we have actually added support for values containing =. This prompted us to keep the 'raw' query string and only parse it into key-value pairs on request. It looks like the intent there was to provide access to the 'raw' query string for rare use cases, but not affect 'regular' use. Even though we cannot store the 'evaluated' value as we do for the fragment (if we do that we can't distinguish between data and delimiters anymore), I think we should give that 'illusion' more strongly. This means we should throw on any disallowed characters when parsing in 'strict' mode, and otherwise present them percent-encoded. (We should not touch "allowed but special" characters (such as `%`, `=`) regardless of mode though)

## Uri construction by parts

The `Uri` constructor that takes individual `queryString` and `fragment` parameters is confusing: for queryString it is up to the user to escape the input, while for the fragment this is taken care of by Akka HTTP. We should use same strategy as when parsing the Uri here (or make it configurable).

PR #3003, opened 2020-03-11
core: percent encode illegal chars when creating URI query

raboof:
This makes sure 'strict' parsing of the Uri query string only allows characters that are actually allowed here by the RFC.

It does not change the fact that Uri's `rawQueryString` is the percent-encoded representation of the query string, since this information is needed to parse it into a key-value structure.

When parsing the Uri in 'relaxed' mode or constructing programmatically, we do accept characters that are not allowed, but percent-encode any would-be invalid characters.

API's that can be used to programmatically add or modify the query string are adapted to not accept invalid percent-encoded values (e.g. `%x2`) and convert illegal characters. `withRawQueryString` is made available both in a 'strict' and a 'relaxed' variation.

References #3002

### DIFF
--- a/akka-http-core/src/main/mima-filters/10.1.x.backwards.excludes/query-string-encoding.backwards.excludes
+++ b/akka-http-core/src/main/mima-filters/10.1.x.backwards.excludes/query-string-encoding.backwards.excludes
@@ -1,1 +1,2 @@
 ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.impl.model.parser.CharacterClasses.raw-query-char")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.javadsl.model.Uri.rawQueryString")
--- a/akka-http-core/src/main/mima-filters/10.1.x.backwards.excludes/query-string-encoding.backwards.excludes
+++ b/akka-http-core/src/main/mima-filters/10.1.x.backwards.excludes/query-string-encoding.backwards.excludes
@@ -1,0 +1,1 @@
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.impl.model.parser.CharacterClasses.raw-query-char")
--- a/akka-http-core/src/main/mima-filters/10.1.x.backwards.excludes/query-string-encoding.backwards.excludes
+++ b/akka-http-core/src/main/mima-filters/10.1.x.backwards.excludes/query-string-encoding.backwards.excludes
@@ -1,2 +1,0 @@
-ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.impl.model.parser.CharacterClasses.raw-query-char")
-ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.javadsl.model.Uri.rawQueryString")
--- a/akka-http-core/src/main/scala/akka/http/impl/model/parser/UriParser.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/model/parser/UriParser.scala
@@ -85,6 +85,10 @@ private[http] final class UriParser(
     case Uri.ParsingMode.Strict => `pchar-base`
     case _                      => `relaxed-path-segment-char`
   }
+  private[this] val `query-char` = uriParsingMode match {
+    case Uri.ParsingMode.Strict => `query-fragment-char`
+    case _                      => `relaxed-query-char`
+  }
   private[this] val `query-key-char` = uriParsingMode match {
     case Uri.ParsingMode.Strict  => `strict-query-key-char`
     case Uri.ParsingMode.Relaxed => `relaxed-query-key-char`
@@ -194,7 +198,7 @@ private[http] final class UriParser(
   def pchar = rule { `path-segment-char` ~ appendSB() | `pct-encoded` }
 
   def rawQueryString = rule {
-    clearSB() ~ oneOrMore(`raw-query-char` ~ appendSB()) ~ run(setRawQueryString(sb.toString)) | run(setRawQueryString(""))
+    clearSB() ~ oneOrMore(`query-char` ~ appendSB() | `pct-encoded`) ~ run(setRawQueryString(sb.toString)) | run(setRawQueryString(""))
   }
 
   // http://www.w3.org/TR/html401/interact/forms.html#h-17.13.4.1
--- a/akka-http-core/src/main/scala/akka/http/impl/model/parser/UriParser.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/model/parser/UriParser.scala
@@ -348,4 +348,3 @@ private[http] final class UriParser(
     createSafe(_scheme, Authority(_host, normalizePort(_port, _scheme), _userinfo), path, _rawQueryString, _fragment)
   }
 }
-
--- a/akka-http-core/src/main/scala/akka/http/impl/model/parser/UriParser.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/model/parser/UriParser.scala
@@ -36,7 +36,7 @@ private[http] final class UriParser(
 
   def parseAbsoluteUri(): Uri =
     rule(`absolute-URI` ~ EOI).run() match {
-      case Right(_)    => createSafe(_scheme, Authority(_host, _port, _userinfo), collapseDotSegments(_path), _rawQueryString, _fragment)
+      case Right(_)    => createUnsafe(_scheme, Authority(_host, _port, _userinfo), collapseDotSegments(_path), _rawQueryString, _fragment)
       case Left(error) => fail(error, "absolute URI")
     }
 
@@ -48,7 +48,7 @@ private[http] final class UriParser(
 
   def parseAndResolveUriReference(base: Uri): Uri =
     rule(`URI-reference` ~ EOI).run() match {
-      case Right(_)    => resolveSafe(_scheme, _userinfo, _host, _port, _path, _rawQueryString, _fragment, base)
+      case Right(_)    => resolveUnsafe(_scheme, _userinfo, _host, _port, _path, _rawQueryString, _fragment, base)
       case Left(error) => fail(error, "URI reference")
     }
 
@@ -294,7 +294,7 @@ private[http] final class UriParser(
     rule(`request-target` ~ EOI).run() match {
       case Right(_) =>
         val path = if (_scheme.isEmpty) _path else collapseDotSegments(_path)
-        createSafe(_scheme, Authority(_host, _port, _userinfo), path, _rawQueryString, _fragment)
+        createUnsafe(_scheme, Authority(_host, _port, _userinfo), path, _rawQueryString, _fragment)
       case Left(error) => fail(error, "request-target")
     }
 
@@ -345,6 +345,6 @@ private[http] final class UriParser(
 
   private def createUriReference(): Uri = {
     val path = if (_scheme.isEmpty) _path else collapseDotSegments(_path)
-    createSafe(_scheme, Authority(_host, normalizePort(_port, _scheme), _userinfo), path, _rawQueryString, _fragment)
+    createUnsafe(_scheme, Authority(_host, normalizePort(_port, _scheme), _userinfo), path, _rawQueryString, _fragment)
   }
 }
--- a/akka-http-core/src/main/scala/akka/http/impl/model/parser/UriParser.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/model/parser/UriParser.scala
@@ -69,22 +69,16 @@ private[http] final class UriParser(
    */
   def parseRawQueryString(): String = {
     rule(rawQueryString ~ EOI).run() match {
-      case Right(())   => encodeRawQueryString(sb.toString)
+      case Right(()) =>
+        uriParsingMode match {
+          case Uri.ParsingMode.Strict =>
+            // Cannot contain invalid characters in strict mode
+            sb.toString
+          case Uri.ParsingMode.Relaxed =>
+            // Percent-encode invalid characters
+            UriRendering.encode(new StringRendering, sb.toString, uriParsingCharset, `query-fragment-char` ++ '%', false).get
+        }
       case Left(error) => fail(error, "rawQueryString")
-    }
-  }
-
-  /**
-   * @param rawQueryString raw, percent-encoded query string, possibly containing invalid unescaped characters
-   *                       when in 'relaxed' mode, but never invalid escaped characters (e.g. '%x2')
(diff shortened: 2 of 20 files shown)


raboof:
(I'm confused why Jenkins is reporting mima failures, AFAICS those should be filtered out now..)

akka-http-core/src/main/mima-filters/10.1.x.backwards.excludes/query-string-encoding.backwards.excludes
```diff
@@ -0,0 +1 @@
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.impl.model.parser.CharacterClasses.raw-query-char")
```
jrudolph: Maybe it needs name mangling and mima is reporting the unmangled names? I don't think `-` is allowed in symbol names.

akka-http-core/src/test/scala/akka/http/scaladsl/model/UriSpec.scala
```diff
@@ -593,6 +594,8 @@ class UriSpec extends AnyWordSpec with Matchers {
       normalize("?a=1&b=2") shouldEqual "?a=1&b=2"
       normalize("?a+b=c%2Bd") shouldEqual "?a+b=c%2Bd"
       normalize("?a&a") shouldEqual "?a&a"
+      normalize("?foo\"bar") shouldEqual "?foo%22bar"
+      a[IllegalUriException] should be thrownBy normalize("?foo\"bar", mode = Uri.ParsingMode.Strict)
```
jrudolph: :+1:

akka-http-core/src/test/scala/akka/http/scaladsl/model/UriSpec.scala
```diff
@@ -754,6 +772,10 @@ class UriSpec extends AnyWordSpec with Matchers {
       uri.withQuery(Query(Map("param1" -> "value1"))) shouldEqual Uri("http://host/path?param1=value1#fragment")
       uri.withRawQueryString("param1=value1") shouldEqual Uri("http://host/path?param1=value1#fragment")
 
+      uri.withQuery(Query("param1" -> "val\"ue1")) shouldEqual Uri("http://host/path?param1=val%22ue1#fragment")
```
jrudolph: Yep, this API should definitely encode when converting to raw :+1:

akka-http-core/src/test/scala/akka/http/scaladsl/model/UriSpec.scala
```diff
@@ -754,6 +772,10 @@ class UriSpec extends AnyWordSpec with Matchers {
...
+      uri.withRawQueryString("param1=val%22ue1") shouldEqual Uri("http://host/path?param1=val%22ue1#fragment")
+      uri.withRawQueryString("param1=val\"ue1") shouldEqual Uri("http://host/path?param1=val\"ue1#fragment")
```
jrudolph: Didn't we want to prevent illegal URIs? Is that allowed?

jrudolph: Or maybe the question is more about how is it rendered?

akka-http-tests/src/test/java/akka/http/javadsl/server/directives/ParameterDirectivesTest.java
```diff
@@ -31,11 +31,6 @@ public void testStringParameterExtraction() {
       .assertStatusCode(200)
       .assertEntity("john");
 
-    route
-        .run(HttpRequest.create().withUri("/abc?stringParam=a%b"))
```
jrudolph: Could we use the `Raw-Request-URI` header instead, here?

akka-http-core/src/test/scala/akka/http/scaladsl/model/UriSpec.scala
jrudolph: Maybe add a test about how it rendered for clarity.

akka-http-core/src/main/mima-filters/10.1.x.backwards.excludes/query-string-encoding.backwards.excludes
raboof: the strange thing is it succeeds for me locally...

akka-http-tests/src/test/java/akka/http/javadsl/server/directives/ParameterDirectivesTest.java
raboof: invalid percent-encodings aren't allowed even in 'relaxed' mode, so I'm not sure what to test in that case?

Or do we think it's too restrictive to disallow invalid percent-encodings in 'relaxed' mode? (it's what we do for other parts of the URI as well...)

akka-http-core/src/main/mima-filters/10.1.x.backwards.excludes/query-string-encoding.backwards.excludes
raboof: figured it out: we're duplicating some of the mima code in the pullrequestValidation task, and the '10.1.x' support wasn't in there yet.

jrudolph: Good catch

akka-http-tests/src/test/java/akka/http/javadsl/server/directives/ParameterDirectivesTest.java
jrudolph: Just to keep this test (and the one on the Scala side) which explicitly tests the response to invalid URIs.

akka-http-core/src/test/scala/akka/http/scaladsl/model/UriSpec.scala
```diff
@@ -798,6 +820,13 @@ class UriSpec extends AnyWordSpec with Matchers {
       Uri("http://example.com//foo/bar?query=1#frag").toHttpRequestTargetOriginForm.toString shouldEqual "//foo/bar?query=1"
     }
 
+    "properly render query strings with invalid values" in {
+      val uri = Uri("http://host/path?query#fragment")
+      uri.withQuery(Query("param1" -> "val\"ue1")).toString shouldEqual "http://host/path?param1=val%22ue1#fragment"
+      uri.withRawQueryString("param1=val%22ue1").toString shouldEqual "http://host/path?param1=val%22ue1#fragment"
+      uri.withRawQueryString("param1=val\"ue1").toString shouldEqual "http://host/path?param1=val%22ue1#fragment"
```
jrudolph: :+1:

akka-http-core/src/test/scala/akka/http/scaladsl/model/UriSpec.scala
```diff
@@ -754,6 +772,10 @@ class UriSpec extends AnyWordSpec with Matchers {
       uri.withQuery(Query(Map("param1" -> "value1"))) shouldEqual Uri("http://host/path?param1=value1#fragment")
       uri.withRawQueryString("param1=value1") shouldEqual Uri("http://host/path?param1=value1#fragment")
 
+      uri.withQuery(Query("param1" -> "val\"ue1")) shouldEqual Uri("http://host/path?param1=val%22ue1#fragment")
+      uri.withRawQueryString("param1=val%22ue1") shouldEqual Uri("http://host/path?param1=val%22ue1#fragment")
+      uri.withRawQueryString("param1=val\"ue1") shouldEqual Uri("http://host/path?param1=val%22ue1#fragment")
```
jrudolph: Good, so both toString and equality is now tested for those cases.

akka-http-tests/src/test/java/akka/http/javadsl/server/directives/ParameterDirectivesTest.java
raboof: The problem here is we can no longer create a `Uri` object with that error: it will throw an InvalidUriException before we have a chance to send it to the `Route`... so the 'corresponding' test is now on the URI parsing side, https://github.com/akka/akka-http/pull/3003/files#diff-197042ca975154fb454f1aea34cd34d7R649

docs/src/main/paradox/migration-guide/migration-guide-10.2.x.md
```diff
@@ -0,0 +1,14 @@
...
+See the general @ref[compatibility guidelines](../compatibility-guidelines.md).
+
+## Akka HTTP 10.1.11 - > 10.2.0
+
+### Strict query strings
+
+In 10.1.x, parsing the query string of URL's accepted characters that are
+strictly speaking not allowed, even when `parsing.uri-parsing-mode` was
+set to the default value of `strict`. If you want to allow such characters
+in incoming URI's, set `parsing.uri-parsing-mode` to `relaxed`
```
jrudolph:
```suggestion
In 10.1.x, while parsing the query string of a URI, characters were accepted that are
not allowed according to RFC 3986, even when `parsing.uri-parsing-mode` was
set to the default value of `strict`. Parsing such URIs will now fail in `strict` mode.
If you want to allow such characters in incoming URIs, set `parsing.uri-parsing-mode` to `relaxed`, in which case these characters will be percent-encoded automatically.
```

akka-http-core/src/main/scala/akka/http/impl/model/parser/UriParser.scala
```diff
@@ -194,7 +220,7 @@ private[http] final class UriParser(
   def pchar = rule { `path-segment-char` ~ appendSB() | `pct-encoded` }
 
   def rawQueryString = rule {
-    clearSB() ~ oneOrMore(`raw-query-char` ~ appendSB()) ~ run(setRawQueryString(sb.toString)) | run(setRawQueryString(""))
+    clearSB() ~ oneOrMore(`query-char` ~ appendSB() | `pct-encoded`) ~ run(setRawQueryString(sb.toString)) | run(setRawQueryString(""))
```
jrudolph: So, when parsing a full URI, we first parse the raw query string into the var and then later `Uri.create` is called which will have another parse run using `parseRawQueryString` to fix potentially broken characters. That seems a bit wasteful. Is there a way we could do it in one step?

I guess it's not performance critical if you use `strict` mode but I guess we might need to be careful because URI parsing is one of the hotspots of the whole request processing.

akka-http-tests/src/test/java/akka/http/javadsl/server/directives/ParameterDirectivesTest.java
jrudolph: With `Raw-Request-URI` you don't have to create a URI object on the client side. But maybe it still wouldn't work because `RouteTest` doesn't support `Raw-Request-URI`?

raboof: Indeed it doesn't support it - I guess we could add that but I'm not sure how useful that really is.

akka-http-core/src/main/scala/akka/http/scaladsl/model/Uri.scala
```diff
@@ -830,12 +883,31 @@ object Uri {
...
   private[http] def create(scheme: String, authority: Authority, path: Path, queryString: Option[String],
                            fragment: Option[String]): Uri =
+    createSafe(scheme, authority, path, queryString.map(new UriParser(_).parseRawQueryString()), fragment)
+
+  /**
+   * @param queryString percent-encoded query string that must be guaranteed
+   *                    not to contain invalid percent-encodings or characters not allowed by
+   *                    the RFC.
+   */
+  private[http] def createSafe(scheme: String, authority: Authority, path: Path, queryString: Option[String],
```
jrudolph: I found that a bit confusing. Shouldn't this one be called `createUnsafe`?

akka-http-core/src/main/scala/akka/http/scaladsl/model/Uri.scala
```diff
@@ -278,8 +321,11 @@ object Uri {
    * Parses the given string as if it were the value of an HTTP/2 ":path" pseudo-header.
    * The result is a path and a query string as defined in
    * https://tools.ietf.org/html/rfc7540#section-8.1.2.3
-   * If strict is `false`, accepts unencoded visible 7-bit ASCII characters in addition to the RFC.
    * If the given string is not a valid path or query string the method throws an `IllegalUriException`.
+   *
+   * @param mode if `Relaxed`, accepts unencoded visible 7-bit ASCII characters in addition to the RFC.
+   * @return path and percent-encoded query string. When in in 'relaxed' mode, characters not permitted by https://tools.ietf.org/html/rfc3986#section-3.4
+   *         *  are already automatically percent-encoded here
```
ennru:
```suggestion
   * are already automatically percent-encoded here
```

akka-http-core/src/main/scala/akka/http/scaladsl/model/Uri.scala
```diff
@@ -52,6 +57,10 @@ sealed abstract case class Uri(scheme: String, authority: Authority, path: Path,
 
   /**
    * Returns a copy of this Uri with the given components.
+   *
+   *  If you want to use the copy constructor to update the rawQueryString, it is up to you to
+   *  make sure the query string does not contain invalid characters. For this reason we
+   *  recommend you to use [[Uri#withRawQueryString()]] instead.
```
ennru:
```suggestion
   * If you want to use the copy constructor to update the 'rawQueryString', it is up to you to
   * make sure the query string does not contain invalid characters. For this reason we
   * recommend using [[Uri#withRawQueryString()]] instead.
```

akka-http-core/src/main/scala/akka/http/scaladsl/model/Uri.scala
ignasi35: @ennru  I think it should be:

```suggestion
   *         are already automatically percent-encoded here
```

To align with the previous line.

akka-http-tests/src/test/java/akka/http/javadsl/server/directives/ParameterDirectivesTest.java
jrudolph: Ok, let's just remove it for now. Thanks for checking.

raboof:
> Do you think you have now closed all loopholes to create Uris with an invalid `rawQueryString`?

Yes, I think so - of course there's a couple of situations where we parse the string again to make sure there are no invalid percent-encodings in there, but no longer in the 'hot path'.

jrudolph:
> Yes, I think so - of course there's a couple of situations where we parse the string again to make sure there are no invalid percent-encodings in there, but no longer in the 'hot path'.

Yeah, good point. If someone cares about those paths they can still be optimized later on.

PR #3004, opened 2020-03-11
Update the year in the copyright header

ignasi35:
Happy New Year


ignasi35:
> LGTM, any reason this is still marked [WIP]?

None.

jrudolph:
@ignasi35, we usually squash commits before or during merge if they are not standalone changes. No worries, though about this time ;)

jrudolph:
Btw. I tried using `headerCreate` and `test:headerCreate` to do the upgrade and it seems to work for the vast majority of cases. It seems it currently doesn't work for boilerplate and build files and a few files where it is unclear.

PR #3005, opened 2020-03-12
Update to mima 0.7.0 to prevent compilation issues with Akka build

jrudolph:


jrudolph:
Good point, thanks @raboof. Added. Also added a mima exclusion for a previous backport that only happens on 2.11 and was only now detected with the change to mima 0.7.0.

Issue #2952, opened 2020-02-10
Failed: docs.http.javadsl.server.directives.FutureDirectivesExamplesTest.testOnCompleteWithBreaker


jrudolph:
https://jenkins.akka.io:8498/job/akka-http-nightly/6/testReport/
https://jenkins.akka.io:8498/job/akka-http-nightly/16/testReport/

```
java.lang.AssertionError: Unexpected entity! expected:<[The server is currently unavailable (because it is overloaded or down for maintenance).]> but was:<[An error occurred: java.lang.ArithmeticException: / by zero]>
  Request was:      HttpRequest(HttpMethod(GET),/divide/10/0,List(),HttpEntity.Strict(none/none,0 bytes total),HttpProtocol(HTTP/1.1))
  Route result was: Future(Success(Complete(HttpResponse(500 Internal Server Error,List(),HttpEntity.Strict(text/plain; charset=UTF-8,59 bytes total),HttpProtocol(HTTP/1.1)))))

	at akka.http.javadsl.testkit.JUnitRouteTestBase$$anon$1.reportDetails(JUnitRouteTest.scala:47)
	at akka.http.javadsl.testkit.JUnitRouteTestBase$$anon$1.assertEquals(JUnitRouteTest.scala:30)
	at akka.http.javadsl.testkit.TestRouteResult.assertEqualsKind(TestRouteResult.scala:227)
	at akka.http.javadsl.testkit.TestRouteResult.assertEntity(TestRouteResult.scala:166)
	at docs.http.javadsl.server.directives.FutureDirectivesExamplesTest.testOnCompleteWithBreaker(FutureDirectivesExamplesTest.java:162)
```

PR #3007, opened 2020-03-15
Increase the resetTimeout to assert the circuit breaker is open

ignasi35:
References #2952 

A pause (GC, noisy neighbor) before:

https://github.com/akka/akka-http/blob/7683d53d97ed6878ac132c6575a714b199db14ba/docs/src/test/java/docs/http/javadsl/server/directives/FutureDirectivesExamplesTest.java#L161

allows the circuit breaker to close again. Increasing the `resetTimeout` should reduce the odds.

### DIFF
--- a/docs/src/test/java/docs/http/javadsl/server/directives/FutureDirectivesExamplesTest.java
+++ b/docs/src/test/java/docs/http/javadsl/server/directives/FutureDirectivesExamplesTest.java
@@ -4,6 +4,7 @@
 
 package docs.http.javadsl.server.directives;
 
+import java.time.Duration;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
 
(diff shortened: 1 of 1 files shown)


jrudolph:
I think this comment is still valid:

https://github.com/akka/akka-http/issues/1063#issuecomment-354817510

Something was fixed afterwards but not the thing I commented on.

The tl;dr is that there's no happens-before chain between the time a failed invocation can be observed and the circuit breaker tripping. I.e. you can see the failure but the circuit breaker has not yet tripped.

jrudolph:
This could probably also be fixed in the CircuitBreaker implementation. I filed https://github.com/akka/akka/issues/24228 back them.

docs/src/test/java/docs/http/javadsl/server/directives/FutureDirectivesExamplesTest.java
```diff
@@ -169,7 +179,7 @@ public void testOnCompleteWithBreaker() throws InterruptedException {
     new TestKit(system()) {
       {
         awaitAssert(
-            FiniteDuration.create(500, TimeUnit.MILLISECONDS),
+            Duration.ofSeconds(500),
```
ignasi35: Prefer java types

ignasi35:
Thanks @jrudolph for the comments. I've rollbacked my original approach and adopted eventuality of circuit breakers in the test.

docs/src/test/java/docs/http/javadsl/server/directives/FutureDirectivesExamplesTest.java
```diff
@@ -156,11 +157,20 @@ public void testOnCompleteWithBreaker() throws InterruptedException {
...
       .assertEntity("An error occurred: java.lang.ArithmeticException: / by zero");
-    // opened the circuit-breaker 
 
-    testRoute(route).run(HttpRequest.GET("/divide/10/0"))
-          .assertEntity("The server is currently unavailable (because it is overloaded or down for maintenance).")
-          .assertStatusCode(StatusCodes.ServiceUnavailable());
+    // The circuit-breaker will eventually be opened
+    new TestKit(system()) {
+      {
+        awaitAssert(
```
jrudolph: Can you move the whole testing code from there on into that initializer block?

jrudolph:
Great, thanks, @ignasi35.

Issue #2930, opened 2020-02-05
Failed: FileUploadDirectivesExamplesSpec


chbatey:
https://jenkins.akka.io:8498/job/akka-http-nightly-release-10.1/AKKA_VERSION=default,Node=akka-http-nightly-node,SCALA_VERSION=2.11.12,jdk=AdoptOpenJDK%2011/788/testReport/junit/docs.http.scaladsl.server.directives/FileUploadDirectivesExamplesSpec/storeUploadedFile/

```
org.scalatest.exceptions.TestFailedException: Request was neither completed nor rejected within 3 seconds
	at akka.http.scaladsl.testkit.TestFrameworkInterface$Scalatest$class.failTest(TestFrameworkInterface.scala:24)
	at akka.http.scaladsl.server.RoutingSpec.failTest(RoutingSpec.scala:19)
	at akka.http.scaladsl.testkit.RouteTestResultComponent$RouteTestResult.failNeitherCompletedNorRejected(RouteTestResultComponent.scala:93)
	at akka.http.scaladsl.testkit.RouteTestResultComponent$RouteTestResult.rawResponse(RouteTestResultComponent.scala:58)
	at akka.http.scaladsl.testkit.RouteTestResultComponent$RouteTestResult.response(RouteTestResultComponent.scala:39)
	at akka.http.scaladsl.testkit.RouteTest$class.response(RouteTest.scala:61)
	at akka.http.scaladsl.server.RoutingSpec.response(RoutingSpec.scala:19)
	at akka.http.scaladsl.testkit.RouteTest$class.status(RouteTest.scala:79)
	at akka.http.scaladsl.server.RoutingSpec.status(RoutingSpec.scala:19)
	at docs.http.scaladsl.server.directives.FileUploadDirectivesExamplesSpec$$anonfun$5$$anonfun$apply$5.apply(FileUploadDirectivesExamplesSpec.scala:77)
	at docs.http.scaladsl.server.directives.FileUploadDirectivesExamplesSpec$$anonfun$5$$anonfun$apply$5.apply(FileUploadDirectivesExamplesSpec.scala:77)
```

ignasi35:
Related to #2733 (and partially to #2691)

ignasi35:
Duplicate of #2733

Issue #2733, opened 2019-10-01
Failed: akka.http.scaladsl.server.directives.FileUploadDirectivesSpec.the fileUploadAll directive should stream all file uploads if multiple with the same name are posted as streamed entities


jrudolph:
https://jenkins.akka.io:8498/job/akka-http-nightly/639/AKKA_VERSION=release-2.5,Node=akka-http-nightly-node,SCALA_VERSION=2.12.9,jdk=AdoptOpenJDK%208/testReport/junit/akka.http.scaladsl.server.directives/FileUploadDirectivesSpec/the_fileUploadAll_directive_should_stream_all_file_uploads_if_multiple_with_the_same_name_are_posted_as_streamed_entities/

```
Stacktrace

org.scalatest.exceptions.TestFailedException: Request was neither completed nor rejected within 3 seconds
	at akka.http.scaladsl.testkit.TestFrameworkInterface$Scalatest.failTest(TestFrameworkInterface.scala:24)
	at akka.http.scaladsl.testkit.TestFrameworkInterface$Scalatest.failTest$(TestFrameworkInterface.scala:24)
	at akka.http.scaladsl.server.RoutingSpec.failTest(RoutingSpec.scala:19)
	at akka.http.scaladsl.testkit.RouteTestResultComponent$RouteTestResult.failNeitherCompletedNorRejected(RouteTestResultComponent.scala:93)
	at akka.http.scaladsl.testkit.RouteTestResultComponent$RouteTestResult.rawResponse(RouteTestResultComponent.scala:58)
	at akka.http.scaladsl.testkit.RouteTestResultComponent$RouteTestResult.response(RouteTestResultComponent.scala:39)
	at akka.http.scaladsl.testkit.RouteTest.response(RouteTest.scala:61)
	at akka.http.scaladsl.testkit.RouteTest.response$(RouteTest.scala:61)
	at akka.http.scaladsl.server.RoutingSpec.response(RoutingSpec.scala:19)
	at akka.http.scaladsl.testkit.RouteTest.status(RouteTest.scala:79)
	at akka.http.scaladsl.testkit.RouteTest.status$(RouteTest.scala:79)
	at akka.http.scaladsl.server.RoutingSpec.status(RoutingSpec.scala:19)
	at akka.http.scaladsl.server.directives.FileUploadDirectivesSpec.$anonfun$new$70(FileUploadDirectivesSpec.scala:407)
[...]

Issue #2691, opened 2019-09-09
Failed: docs.http.javadsl.server.directives.FileUploadDirectivesExamplesTest.testUploadedFile


jrudolph:
https://jenkins.akka.io:8498/job/akka-http-nightly/617/AKKA_VERSION=default,Node=akka-http-nightly-node,SCALA_VERSION=2.13.0,jdk=AdoptOpenJDK%208/testReport/junit/docs.http.javadsl.server.directives/FileUploadDirectivesExamplesTest/testUploadedFile/

```
java.util.concurrent.TimeoutException: Future timed out after [3 seconds]
	at scala.concurrent.impl.Promise$DefaultPromise.tryAwait0(Promise.scala:212)
	at scala.concurrent.impl.Promise$DefaultPromise.ready(Promise.scala:219)
	at scala.concurrent.impl.Promise$DefaultPromise.ready(Promise.scala:104)
	at scala.concurrent.Await$.$anonfun$ready$1(package.scala:174)
	at scala.concurrent.BlockContext$DefaultBlockContext$.blockOn(BlockContext.scala:62)
	at scala.concurrent.Await$.ready(package.scala:124)
	at akka.http.impl.util.package$AddFutureAwaitResult.awaitResult(package.scala:61)
	at akka.http.javadsl.testkit.TestRouteResult._response(TestRouteResult.scala:35)
	at akka.http.javadsl.testkit.TestRouteResult.response$lzycompute(TestRouteResult.scala:53)
	at akka.http.javadsl.testkit.TestRouteResult.response(TestRouteResult.scala:53)
	at akka.http.javadsl.testkit.TestRouteResult.status(TestRouteResult.scala:96)
	at akka.http.javadsl.testkit.TestRouteResult.assertStatusCode(TestRouteResult.scala:136)
	at docs.http.javadsl.server.directives.FileUploadDirectivesExamplesTest.testUploadedFile(FileUploadDirectivesExamplesTest.java:87)
```

johanandren:
https://jenkins.akka.io:8498/job/akka-http-nightly/AKKA_VERSION=default,Node=akka-http-nightly-node,SCALA_VERSION=2.12.10,jdk=AdoptOpenJDK%2011/746/

chbatey:
[...]

PR #3008, opened 2020-03-15
Reduce flakiness of FileUploadDirectivesExamples (Test and Spec)

ignasi35:
References #2733 (#2930 is a duplicate of it) and #2691

As the code comment says:

`  // test touches disk, so give it some time`

so let's give it a little more time.


akka-http-testkit/src/main/scala/akka/http/javadsl/testkit/RouteTest.scala
```diff
@@ -34,7 +34,8 @@ abstract class RouteTest extends AllDirectives with WSTestRequestBuilding {
   implicit def materializer: Materializer
   implicit def executionContext: ExecutionContextExecutor = system.dispatcher
 
-  protected def awaitDuration: FiniteDuration = 3.seconds.dilated
+  protected def defaultAwaitDuration = 3.seconds
```
ignasi35: Introduced `defaultAwaitDuration` so it's simpler to override the `awaitDuration` from java code (where invoking `dilated` is not available without too much machinery)

PR #3012, opened 2020-03-19
[backport] core: percent encode illegal chars when creating URI query

### LINKED ISSUE #3002: Inconsistency in the escaping of the query string in the Uri API

jrudolph:
release 10.1 backport of #3003

This makes sure 'strict' parsing of the Uri query string only allows characters that are actually allowed here by the RFC.

It does not change the fact that Uri's rawQueryString is the percent-encoded representation of the query string, since this information is needed to parse it into a key-value structure.

When parsing the Uri in 'relaxed' mode or constructing programmatically, we do accept characters that are not allowed, but percent-encode any would-be invalid characters.

API's that can be used to programmatically add or modify the query string are adapted to not accept invalid percent-encoded values (e.g. %x2) and convert illegal characters. withRawQueryString is made available both in a 'strict' and a 'relaxed' variation.

Fixes #3002

(cherry picked from commit e284816f667294022510172e9d06e1e597264b40)

### DIFF
--- a/docs/src/main/paradox/migration-guide/migration-guide-10.2.x.md
+++ b/docs/src/main/paradox/migration-guide/migration-guide-10.2.x.md
@@ -1,0 +1,14 @@
+# Migration Guide to and within Akka HTTP 10.2.x
+
+## General Notes
+
+See the general @ref[compatibility guidelines](../compatibility-guidelines.md).
+
+## Akka HTTP 10.1.11 - > 10.2.0
+
+### Strict query strings
+
+In 10.1.x, while parsing the query string of a URI, characters were accepted that are
+not allowed according to RFC 3986, even when `parsing.uri-parsing-mode` was
+set to the default value of `strict`. Parsing such URIs will now fail in `strict` mode.
+If you want to allow such characters in incoming URIs, set `parsing.uri-parsing-mode` to `relaxed`, in which case these characters will be percent-encoded automatically.
--- a/akka-http-core/src/main/java/akka/http/javadsl/model/Uri.java
+++ b/akka-http-core/src/main/java/akka/http/javadsl/model/Uri.java
@@ -123,8 +123,32 @@ public abstract class Uri {
 
   /**
    * Returns a copy of this instance with a new query.
+   *
+   * Characters that are not within the range described at https://tools.ietf.org/html/rfc3986#section-3.4
+   * should be percent-encoded. Characters that are in that range may or may not be percent-encoded,
+   * and depending on how the query string is parsed this might be relevant: for example, when interpreting
+   * the query string as 'key=value' pairs you could use the percent-encoded '=' ('%22) to include a '=' in the
+   * key or value.
+   *
+   * When characters are encountered that are outside of the RFC3986 range they are automatically
+   * percent-encoded, but be aware that relying on this is usually a programming error.
    */
   public abstract Uri rawQueryString(String rawQuery);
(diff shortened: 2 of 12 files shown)


docs/src/main/paradox/migration-guide/migration-guide-10.2.x.md
```diff
@@ -0,0 +1,14 @@
+# Migration Guide to and within Akka HTTP 10.2.x
+
+## General Notes
+
+See the general @ref[compatibility guidelines](../compatibility-guidelines.md).
+
+## Akka HTTP 10.1.11 - > 10.2.0
```
jrudolph: If we backport this fully, we need to adapt this. Depending on whether we release 10.1.12 first or 10.2.0, it probably suffices to add this section only once.

raboof:
Hmm. I know I added the `needs-backport` label to #3003, but considering now if it's needed: it is a behavior change (being more strict by default and a little more strict even in 'relaxed' mode) and it's not _necessarily_ a problem (just a risk). What do you think?

jrudolph:
> Hmm. I know I added the `needs-backport` label to #3003, but considering now if it's needed: it is a behavior change (being more strict by default and a little more strict even in 'relaxed' mode) and it's not _necessarily_ a problem (just a risk). What do you think?

Yeah, maybe you are right, let's not do it for now for compatibility reasons.

PR #3014, opened 2020-03-19
Route integration testing

jrudolph:
Follow up to #1125

Refs #39, #952


akka-http-testkit/src/test/scala/akka/http/scaladsl/testkit/ScalatestRouteTestSpec.scala
```diff
@@ -42,6 +43,20 @@ class ScalatestRouteTestSpec extends AnyFreeSpec with Matchers with ScalatestRou
       }
     }
 
+    "a test using ~!> and some checks" in {
+      // raw here, should have been parsed into modelled header when going through an actual server when using `~!>`
+      val extraHeader = RawHeader("X-Forwarded-Proto", "abc")
+      Get() ~> addHeader(extraHeader) ~!> {
+        respondWithHeader(extraHeader) {
```
raboof: This test might be clearer if we respond with a completely different header?

jrudolph:
>  You might want to create an issue and link to it from the `@ApiMayChange` markers.

I removed the `@ApiMayChange` markers completely as we don't offer compatibility for testkit APIs.

akka-http-testkit/src/test/scala/akka/http/scaladsl/testkit/ScalatestRouteTestSpec.scala
jrudolph: The `addHeader` a line above seemed to be useless so I just removed that.

jrudolph:
Squashed from the command line to keep attribution.

jrudolph:
After long last! 3 years later, still thanks a lot @jonas for the initial push ;)

docs/src/main/paradox/routing-dsl/testkit.md
```diff
@@ -222,6 +222,23 @@ Java
...
+  ASSERTIONS
+}
+```
+
+Certain routes can only be tested with `~!>`, for example routes that use
+the `withRequestTimeout` directive.
+
+@@@ note
+Using `~!>` adds considerable extra overhead since each test will start
+a server and bind to a port so use it only when necessary.
```
chetanmeh, line 239: >  Using `~!>` adds considerable extra overhead since each test will start a server and bind to a port so use it only when necessary.

Would it be possible to let the request be routed to a already running server and not start/stop a new server for each test. Objective being that my server is already running in stage environment and we would like to implement some integration test using the test routing DSL. From my quick read of this PR `~!>` looks just the right solution ... just that instead of the test framework starting a new server it can connect to an existing server

raboof: I think using 'regular' requests might be easier in that case.

What part of the test routing DSL is hard to recreate with checks on the 'regular' response?

Issue #2093, opened 2018-06-28
Responding with a stream sometimes result in "Connection reset by peer" error


ilyai:
In our application we have routes that are streaming JSON documents. Here is an example:

    /** GET api/1/tenant/(tenantId)/ads/ */
    def getAllAdsByOwner(advertiserId: AdvertiserId): Route =
      get {
        httpRequiredSession { username =>
          getAllTenantAds(username, advertiserId) { (adSource: Source[AdView, Any]) =>
            complete(adSource)
          }
        }
      }

Most of the time it works as expected, but sometimes, especially when there are many simultaneous requests, the server starts resetting connection just after the headers have been sent.
I tested with a script that requests this route with curl in a loop and aborting if the request failed. It was running for about 2 minutes before stopping. Trace when request fails is the following:

    <= Recv header, 17 bytes (0x11)
    0000: HTTP/1.1 200 OK
    <= Recv header, 54 bytes (0x36)
    0000: Access-Control-Allow-Origin: https://<...>
    <= Recv header, 135 bytes (0x87)
    0000: Access-Control-Expose-Headers: Content-Type, Authorization, Refr
    0040: esh-Token, Set-Authorization, Set-Refresh-Token, asset-content-l
    0080: ength
    <= Recv header, 40 bytes (0x28)
    0000: Access-Control-Allow-Credentials: true
    <= Recv header, 24 bytes (0x18)
    0000: Content-Encoding: gzip
    <= Recv header, 23 bytes (0x17)
    0000: X-Frame-Options: DENY
    <= Recv header, 33 bytes (0x21)
    0000: X-Content-Type-Options: nosniff
    <= Recv header, 26 bytes (0x1a)
    0000: Content-Security-Policy: .
    <= Recv header, 20 bytes (0x14)
    0000: default-src 'self';.
    <= Recv header, 63 bytes (0x3f)
    0000: style-src 'self' 'unsafe-inline' https://fonts.googleapis.com;.
    <= Recv header, 59 bytes (0x3b)
    0000: font-src 'self' 'unsafe-inline' https://fonts.gstatic.com;.
    <= Recv header, 99 bytes (0x63)
    0000: script-src 'self' 'unsafe-inline' 'unsafe-eval' https://*.google
    0040: apis.com https://maps.gstatic.com;.
    <= Recv header, 69 bytes (0x45)
    0000: img-src 'self' data: https://*.googleapis.com https://*.gstatic.
    0040: com;.
[...]

Issue #2116, opened 2018-07-14
Request entity completes normally on HttpIdleTimeoutException


kstokoz:
On POST/PUT request when processing entity stream elements takes longer than server idle timeout caller receives `connection reset error` (expected), but entity flow completes normally (not expected) processing only part of the stream.

Reproducer:
```
object CompleteStreamReproducer extends App {

  val actorSystemConfig = ConfigFactory.parseString(
    """
      akka {
        loglevel = DEBUG
        loggers = ["akka.event.slf4j.Slf4jLogger"]
      }
      akka.http.server.parsing.max-content-length = infinite

      akka.http.server.idle-timeout = 5s
    """
  )

  implicit val system = ActorSystem("actor-system", actorSystemConfig)
  implicit val materializer = ActorMaterializer(ActorMaterializerSettings.create(system).withDebugLogging(true))(system)
  import system.dispatcher

  val route: Route =
    (path("putData") & put) {
      extractRequestEntity { entity =>
        complete {
          entity.dataBytes.map { data =>
            val start = System.currentTimeMillis()

            // simulate CPU intensive task like encryption, encoding, etc.
            while ((System.currentTimeMillis() - start) < 10000) {}

            data
          }.runWith(Sink.fold(0L){ (acc, s) => acc + s.length }).map { length =>
            println(s"Total length: $length")
            "OK"
          }.andThen {
            case result => println(s"Completed with $result")
          }
        }
      }
    }

  val bindingFuture = Http().bindAndHandle(route, "localhost", 8081)
  System.in.read()
}
```

Trying to send data:
```
$ curl -vv -X PUT -H "Content-Type: text/plain" "http://localhost:8081/putData" --data-binary @output-1G.dat
* Connected to localhost (127.0.0.1) port 8081 (#0)
> PUT /putData HTTP/1.1
> Host: localhost:8081
> User-Agent: Mozilla/5.0 (compatible; MSIE 9.0; Windows NT 6.1; Trident/5.0)
> Accept: */*
> Referer:
> Content-Type: text/plain
> Content-Length: 1073741824
> Expect: 100-continue
>
< HTTP/1.1 100 Continue
< Server: akka-http/10.1.1
< Date: Sat, 14 Jul 2018 19:08:25 GMT
* Recv failure: Connection reset by peer
* stopped the pause stream!
* Closing connection 0
curl: (56) Recv failure: Connection reset by peer
```

On Server side:
```
[...]

PR #3022, opened 2020-03-23
Fix server-side cancellation race

jrudolph:
So far we have only fixed the client-side cancellation race. This will be a fix for the server-side.

So far, I have reproduced the problem as described in #2116 which is suspected to be due to the race condition.

### DIFF
--- a/akka-http-core/src/test/scala/akka/http/impl/engine/server/HttpServerSpec.scala
+++ b/akka-http-core/src/test/scala/akka/http/impl/engine/server/HttpServerSpec.scala
@@ -46,7 +46,6 @@ class HttpServerSpec extends AkkaSpec(
      akka.loglevel = DEBUG
      akka.http.server.log-unencrypted-network-bytes = 100
      akka.http.server.request-timeout = infinite
-     akka.scheduler.implementation = "akka.testkit.ExplicitlyTriggeredScheduler"
   """) with Inside with WithLogCapturing { spec =>
   implicit val materializer = ActorMaterializer()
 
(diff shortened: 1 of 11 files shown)


jrudolph:
I now moved the tests requiring the explicit scheduler to a new file (divide and conquer). I "fixed" the few remaining tests in HttpServerSpec that are failing, though, the fixes needs more scrutiny.

akka-http-core/src/test/scala/akka/http/impl/engine/server/HttpServerSpec.scala
```diff
@@ -1009,7 +906,7 @@ class HttpServerSpec extends AkkaSpec(
         netIn.sendComplete()
         netOut.cancel()
 
-        requests.expectComplete()
+        requests.expectError()
```
jrudolph: What happens here is that these scenarios depended on cancellation eventually completing the controller stage which would propagate to the user handler. The new cancellation strategy is "fail after delay". So now the user handler is failed in these situations.

This is somewhat questionable. In most cases, the user handler flow does not really care about stream completion / failures and will just propagate it, but if it does it will now see a somewhat opaque NoMoreElementsNeeded failure. This could be fixed by using the 2.6-only cancellation strategy which will convert those cancellations into completions. Or, we need to see how to change the server-side not to depend on cancellation in these scenarios.

jrudolph:
Might also fix #2093.

akka-http-core/src/test/scala/akka/http/impl/engine/server/HttpServerSpec.scala
ignasi35: Hmmmm, changing this to change it again when we use "the 2.6-only cancellation strategy" sounds _not great_. Should we hold this PR until we're 2.6+ only?

jrudolph:
Moved some unrelated changes to an extra PR and finally fixed all the tests. Now ready for review.

akka-http-core/src/test/scala/akka/http/impl/engine/server/HttpServerSpec.scala
jrudolph: No, then we'd keep the bugs until another year when we drop 2.6 support.

jrudolph:
Rebased and squashed to be merged after validation.

PR #3029, opened 2020-03-30
core: prepend headers in `HttpMessage.addHeader`

### LINKED ISSUE #3028: Investigate ~ 10% performance slow-down of message attributes
Since https://jenkins.akka.io:8498/view/Akka-http/job/akka-http-bench-nightly/106/

we see a ~ 10% performance slowdown. That's when message attributes have been merged.

jrudolph:
Since by default headers are stored in a list, we would slowly append to the
list. It's questionable whether List is the right datastructure here but
that can be changed separately if needed.

Fixes #3028

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
@@ -171,7 +171,7 @@ sealed trait HttpMessage extends jm.HttpMessage {
   /** Removes the header with the given name (case-insensitive) */
   def removeHeader(headerName: String): Self = {
     val lowerHeaderName = headerName.toRootLowerCase
-    mapHeaders(_.filterNot(_.is(lowerHeaderName)))
+    withHeaders(headers.filterNot(_.is(lowerHeaderName)))
   }
 
   def removeAttribute(key: jm.AttributeKey[_]): Self = {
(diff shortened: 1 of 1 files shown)


jrudolph:
Performance check running here: https://jenkins.akka.io:8498/job/akka-http-bench-adhoc/

 * 148: current master
 * 149: only first commit
 * 150: both commits

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
```diff
@@ -158,7 +158,8 @@ sealed trait HttpMessage extends jm.HttpMessage {
    */
   def connectionCloseExpected: Boolean = HttpMessage.connectionCloseExpected(protocol, header[Connection])
 
-  def addHeader(header: jm.HttpHeader): Self = mapHeaders(_ :+ header.asInstanceOf[HttpHeader])
+  /** Return a new instance with the given header added to the headers sequence. It's undefined where the header is added to the sequence */
```
raboof: :+1: seems reasonable to me. probably not worth mentioning in migration notes

jrudolph:
> * 148: current master
> * 151: only first commit
> * 150: both commits
> * 152: before attributes were merged

Seems we are still missing quite a bit of performance compared to before the attributes. Also the second commit seems to make things slower which seems unlikely. I started another series, this time hopefully in more useful order:

 * 153: before attributes were merged
 * 154: current master
 * 155: only first commit
 * 156: both commits

jrudolph:
From the new set of runs, I'd say there's too much noise to conclude that the second commit would have a negative impact (also why should it?). In any case, we haven't completely made up what was lost by giving up `copy`. One reason I found is that `verifyUri` will be called again and again when you chain `withX` methods. That won't be a problem in general. Since it only turned up in relation to the RequestTimeouts which are known to be a significant performance impediment, maybe it's no reason to look into that right now. I'd say if you need ultimate performance you need to disable those timeouts.

raboof:
> One reason I found is that verifyUri will be called again and again when you chain withX methods

Hm, I see how that's wasteful. Perhaps we should add an internal `copyUnsafe` that shifts the responsibility for checks to the `withXxx` method? Of course at the cost of violating DRY... I could try it out in a branch and if it makes enough of a difference we could consider it.

jrudolph:
> > One reason I found is that verifyUri will be called again and again when you chain withX methods
> 
> Hm, I see how that's wasteful. Perhaps we should add an internal `copyUnsafe` that shifts the responsibility for checks to the `withXxx` method? Of course at the cost of violating DRY... I could try it out in a branch and if it makes enough of a difference we could consider it.

Maybe not now but we can keep it in mind. One alternative way of solving it would cache the result of that call in URI instead.

raboof:
> Maybe not now but we can keep it in mind

Works for me!

PR #3031, opened 2020-03-31
2020 performance improvements

jrudolph:
Some stuff needs to go to akka/akka while other stuff needs more validation and cleanup. AFAICS ~ 15-20% performance improvement in the ServerProcessingBenchmark (i.e. only processing without IO layers).


ennru:
Some nifty things in here.

jrudolph:
Part of those now broken out into #3106, #3107, and #3108.

PR #3036, opened 2020-04-01
Akka 2.5.31; accept ClassicActorSystemProvider in extensions

ennru:
Upgrade to Akka 2.5.31.
So that `ClassicActorSystemProvider` can be used to accept both classic and new actor systems to create the Akka HTTP extensions.

Methods with `akka.actor.ActorSystem` are kept for bin-compatiblity.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/javadsl/Http.scala
+++ b/akka-http-core/src/main/scala/akka/http/javadsl/Http.scala
@@ -6,18 +6,20 @@ package akka.http.javadsl
 
 import java.net.InetSocketAddress
 import java.util.Optional
+
 import akka.http.impl.util.JavaMapping
 import akka.http.impl.util.JavaMapping.HttpsConnectionContext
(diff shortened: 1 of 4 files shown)


jrudolph:
Do you want this to be merged, or should it be integrated into @raboof's more comprehensive improvements?

raboof:
(I'm fine with merging independently)

ennru:
I wanted to do it independently so that we know it works on Akka 2.5.
So I guess we can merge this first.

jrudolph:
Hmm, why did the Travis checks not run?

raboof:
Perhaps we should merge this, then #2996 (which already exercises some `Http()` accepting a ClassicActorSystemProvider by no longer providing an implicit classic system in the typed example anymore), then follow up with more specific tests?

jrudolph:
Yep, I just triggered a new set of validations to make sure we haven't missed any Mima failures in Scala 2.13.

PR #3042, opened 2020-04-06
core: parse empty query `?` to Query.empty

### LINKED ISSUE #2769: akka.http.scaladsl.model.Uri("?").query not parsed as expected.
I think `Uri("?").query` should be parsed the same as `Uri("").query`, but instead they are different:

```scala
akka.http.scaladsl.model.Uri("").query().toString // String = ""
akka.http.scaladsl.model.Uri("?").query().toString // String = ""
akka.http.scaladsl.model.Uri("?foo=bar").query().toString // String = "foo=bar"

akka.http.scaladsl.model.Uri("").query().nonEmpty // Boolean = false
akka.http.scaladsl.model.Uri("?").query().nonEmpty // Boolean = true <- expected false
akka.http.scaladsl.model.Uri("?foo=bar").query().nonEmpty // Boolean = true

akka.http.scaladsl.model.Uri("").query().toMap // Map[String,String] = Map()
akka.http.scaladsl.model.Uri("?").query().toMap // Map[String,String] = Map("" -> "") <- expected Map()
akka.http.scaladsl.model.Uri("?foo=bar").query().toMap // Map[String,String] = Map(foo -> bar)
```

jrudolph:
This still doesn't mean `Uri("") == Uri("?")`, because we'd like to keep
the distinction between "no query" and "empty query" (at least for historical
reasons).

Fixes #2769

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/model/parser/UriParser.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/model/parser/UriParser.scala
@@ -263,7 +263,7 @@ private[http] final class UriParser(
       if (valueStack.size + 5 <= maxValueStackSize) keyValuePairsWithReversalAvoidance
       else keyValuePairsWithLimitedStackUse
 
-    rule { keyValuePairs }
+    rule { (EOI ~ push(Query.Empty)) | keyValuePairs }
   }
 
   def fragment = rule(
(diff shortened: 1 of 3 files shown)


jrudolph:
> Could we have a test which asserts that `"http://foo.com"` is parsed differently from `"http://foo.com/?"`? I mean, there's `Uri("?").query() shouldBe empty` but what would `Uri("").query() shouldBe ` be?

Good point, added such a check. It should also be empty.

PR #3044, opened 2020-04-06
build: configure travis git depth to fix sbt-dynver

### LINKED ISSUE #3043: Snapshot version numbering is broken
These are the latest versions according to our snapshot repo at https://bintray.com/akka/snapshots/akka-http in reverse chronological order. As you can see they are all over the place:

```
0.0.0+104-4387f72e
0.0.0+105-76525121
0.0.0+107-b005ca48
0.0.0+110-3a2db8c8
0.0.0+115-61ca6bbc
0.0.0+122-e8be2027
0.0.0+129-0edbcafa
0.0.0+128-9e8fb91e
0.0.0+142-62355867
0.0.0+150-272d3d15
0.0.0+154-e284816f
10.1.10+158-d7ea94c7
10.1.10+162-bee4cd63
10.1.10+160-39b04ca6
10.1.10+163-a8da008a
10.1.10+161-2c8cd3bd
10.1.10+162-909781a8
10.1.10+159-bad9165e
10.1.10+160-69bd7bf3
10.1.10+154-9f99363a
10.1.10+155-a811227c
10.1.10+140-eaca3e98
10.1.10+143-fc69abca
10.1.10+144-cc2fab77
10.1.10+145-85efd85f
10.1.10+139-b38edc59
10.1.10+140-cc5529ce
10.1.10+142-99028e11
10.1.10+142-6ce12ca6
10.1.10+142-16cb71ce
10.1.10+142-1ef209fd
10.1.10+141-a4daaee8
10.1.10+140-5950c3e0
10.1.10+145-9925fbe4
10.1.10+147-52a80233
10.1.10+149-4e5f8f5b
10.1.10+152-75d954b6
10.1.10+154-2f877845
10.1.10+157-e066319b
10.1.10+156-39eb96a5
10.1.11+111-2f88d20d
```

The first one to go wrong seems to be 39eb96a5. Do we know what might have happened there?

jrudolph:
Otherwise, git in travis will not know about tags sufficiently far in the
past and version numbers will be weird.

Fixes #3043

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,9 @@ scala:
 
 sudo: false
 
+git:
+  depth: 500 # to make sure we have enough depth to power git describe for sbt-dynver (used for snapshot version numbers)
+
 before_install:
   # make comparing to origin/master work
   - git remote set-branches --add origin master && git fetch


.travis.yml
```diff
@@ -5,6 +5,9 @@ scala:
 
 sudo: false
 
+git:
+  depth: 500 # to make sure we have enough depth to power git describe for sbt-dynver (used for snapshot version numbers)
```
ennru, line 9: Not sure which is better, Alpakka does 

https://github.com/akka/alpakka/blob/b0e99507ffce4e4cef6d4e64cc5222455a178653/.travis.yml#L7-L9

jrudolph: Yeah, maybe we don't need the full history. It sounds like 500 commits are plenty.

jrudolph:
> Hmm, this is a good improvement, but the versions in #3043 are still suspicious - let's keep an eye on this.

Yep, a bit weird that 10.1.11 went out of depth before 10.1.10. The more recent 0.0.0 prefix then makes more sense again.

I guess the 10.1.10 might have surviced longer because there might have been a merge than spanned a long interval?

jrudolph:
> I guess the 10.1.10 might have survived longer because there might have been a merge than spanned a long interval?

6bfe23d0229237c6ec56b524c54aef380dd2da86 is a likely candidate for that. It started only a few commits after 10.1.10 and was merged only after 10.1.11.

jrudolph:
So, while `git describe` has a reasonable way of counting "commits since tag", depth might still have included the commit for 10.1.10 because the shortest path from master to 10.1.10 was still shorter than 50 commits because of that merge commit spanning a long time.

PR #3046, opened 2020-04-07
core: make sure to cancel response entity on failure

jrudolph:
Otherwise, the response entity stream can leak when the response
rendering stage is completed by a failure instead of by the
associated cancellation.

Currently, it seems failures like this seem to be rare to reach this point
and cancellation usually wins the race (or failures are converted into
completions).

Discovered while working on #3022 and extracted here.

Should be backported to 10.1.x.


akka-http-core/src/main/scala/akka/http/impl/engine/rendering/HttpResponseRendererFactory.scala
```diff
@@ -94,20 +94,26 @@ private[http] class HttpResponseRendererFactory(
           override def onUpstreamFinish(): Unit =
             if (transferring) closeMode = CloseConnection
             else completeStage()
+
+          override def onUpstreamFailure(ex: Throwable): Unit = {
+            stopTransfer()
```
jrudolph, line 99: This is the fix. The rest is refactoring to make `stopTransfer` accessible here.

ennru:
Created backport in #3064

PR #3048, opened 2020-04-09
docs: link to major/minor Akka docs

raboof:
Refs #2886


ennru:
Wouldn't it be better to link to the binary version, 2.6?

PR #3049, opened 2020-04-09
Allow customizing parsing errors

raboof:
Refs #210

A bit unfortunate that ServerSettings includes the class name and
not the object, but to instantiate the class it'd be good to use
system.dynamicAccess and that is not available when creating the
settings...

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/ParsingErrorHandler.scala
+++ b/akka-http-core/src/main/scala/akka/http/ParsingErrorHandler.scala
@@ -1,0 +1,22 @@
+package akka.http
+
+import akka.event.LoggingAdapter
+import akka.http.javadsl.{ model => jm }
+import akka.http.scaladsl.model.{ ErrorInfo, HttpResponse, StatusCode }
+import akka.http.scaladsl.settings.ServerSettings
+
+abstract class ParsingErrorHandler {
+  def handle(status: StatusCode, error: ErrorInfo, log: LoggingAdapter, settings: ServerSettings): jm.HttpResponse
+}
+
+object DefaultParsingErrorHandler extends ParsingErrorHandler {
+  import akka.http.impl.engine.parsing.logParsingError
+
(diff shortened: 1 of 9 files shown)


akka-http-core/src/main/scala/akka/http/ParsingErrorHandler.scala
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+package akka.http
+
+import akka.event.LoggingAdapter
+import akka.http.javadsl.{ model => jm }
+import akka.http.scaladsl.model.{ ErrorInfo, HttpResponse, StatusCode }
+import akka.http.scaladsl.settings.ServerSettings
+
+abstract class ParsingErrorHandler {
```
jrudolph, line 12: Why not a trait? It probably doesn't matter a lot.

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
```diff
@@ -360,7 +361,8 @@ class HttpExt private[http] (private val config: Config)(implicit val system: Ex
     remoteAddress:      Option[InetSocketAddress] = None,
     log:                LoggingAdapter            = system.log,
     isSecureConnection: Boolean                   = false): ServerLayer = {
-    val server = HttpServerBluePrint(settings, log, isSecureConnection)
+    val parsingErrorHandler = system.dynamicAccess.createInstanceFor[ParsingErrorHandler](settings.parsingErrorHandler, Nil).get
```
jrudolph, line 364: I guess conceptionally, this could also go into the settings class but if it's only used here, I guess it's fine to keep it here.

jrudolph:
> A bit unfortunate that ServerSettings includes the class name and
> not the object, but to instantiate the class it'd be good to use
> system.dynamicAccess and that is not available when creating the
> settings...

Yeah, indeed, but I guess it's good enough.

Issue #2924, opened 2020-02-04
Allow ignoring incoming headers that would be reported with extractClientIP


jrudolph:
Depending from whom you receive these kinds of headers, they might not be trustworthy (you might trust your reverse proxy but you cannot trust headers from any kind of client). So, you cannot distinguish headers from the client from headers by the akka-http infrastructure.

An alternative could be to put remote address information into a message attribute in 10.2.x.

When done, review documentation (also wrt to changes added in #2922).

TODO list:

 * [x] add new `remote-address-attribute` feature (#3051)
 * [x] adapt documentation for `extractClientIp`
 * [x] deprecate `remote-address-header` feature

raboof:
[extractClientIP](https://doc.akka.io/docs/akka-http/current/routing-dsl/directives/misc-directives/extractClientIP.html) gets its data from the possibly-untrustworthy `X-Forwarded-For`, `Remote-Address`, or `X-Real-IP` headers. When `akka.http.server.remote-address-header` is on, `Remote-Address` is filled with a trustworthy address. That is indeed confusing.

What we could do is:
* deprecate `akka.http.server.remote-address-header`
* introduce a `akka.http.server.remote-address-attribute` that works like `akka.http.server.remote-address-header` but sets an attribute (perhaps of type `RemoteAddress`?) instead of a header
* document how to get that attribute, and how to use specific `X-Forwarded-For`/`Remote-Address`/`X-Real-IP` headers only if you are sure you are behind a trustworthy proxy that reliably sets those headers
* update the `extractClientIP` docs to clarify that it may be used for debugging/diagnostics/convenience but not for security, and link to the new docs to reliably get the client IP

jrudolph:
New remote-address-attribute feature added in #3051.

jrudolph:
https://github.com/akka/akka-http/pull/3083#pullrequestreview-398029478:

> When deprecating the `remote-address-header` setting we should touch extractClientIp again and change the priority to take the attribute with highest priority. Then we should update the documentation de-emphasizing the previous setting.

raboof:
> we should touch extractClientIp again and change the priority to take the attribute with highest priority

I'm not sure I agree. What is the purpose of `extractClientIP`?

AFAICS it's not to get a safe/reliable/secure client IP: if you want to get the client IP reliably, you should use the attribute, or (if you have proxies in between) you should first make sure the proxies correctly set a certain header (which might be one of the `extractClientIP` ones) and use that header specifically.

[...]

PR #3051, opened 2020-04-10
core: add akka.http.server.remote-address-attribute #2924

changvvb:
References #2924


lightbend-cla-validator:

lightbend-cla-validator:

lightbend-cla-validator:

akka-http-core/src/main/scala/akka/http/impl/settings/ServerSettingsImpl.scala
```diff
@@ -83,6 +84,12 @@ private[http] object ServerSettingsImpl extends SettingsCompanionImpl[ServerSett
     c.getInt("max-connections"),
     c.getInt("pipelining-limit"),
     c.getBoolean("remote-address-header"),
+    c.getString("remote-address-attribute").toOption.map { attribute =>
```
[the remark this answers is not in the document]
changvvb: Thanks for your comments.
I will redesign this PR according to your advice.

lightbend-cla-validator:

lightbend-cla-validator:

lightbend-cla-validator:

akka-http-core/src/main/mima-filters/10.1.11.backwards.excludes/2924-add-remote-address-attribute.backwards.excludes
```diff
@@ -0,0 +1,6 @@
+# add remoteAddressAttribute for akka.http.scaladsl.settings.ServerSettings
```
raboof, line 1: These filters can be added to `10.1.x.backwards.excludes` - that way we avoid errors when this feature is released in 10.2.0, and we later release a 10.1.12.

akka-http-core/src/test/scala/akka/http/impl/engine/server/HttpServerSpec.scala
```diff
@@ -989,6 +989,28 @@ class HttpServerSpec extends AkkaSpec(
...
+      }
+
+      send(
+        """GET / HTTP/1.1
+            |Host: example.com
+            |
+            |""".stripMarginWithNewline("\r\n"))
+
+      val request = expectRequest()
+      request.attributes(HttpRequest.AttributeKeys.remoteAddress) should equal(RemoteAddress(theAddress, Some(8080)))
```
raboof, line 1010: It might be nice to add those other headers with other addresses, to show that these headers don't change the attribute

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
```diff
@@ -485,6 +485,10 @@ object HttpRequest {
     protocol: HttpProtocol              = HttpProtocols.`HTTP/1.1`) = new HttpRequest(method, uri, headers, Map.empty, entity, protocol)
 
   def unapply(any: HttpRequest) = new OptHttpRequest(any)
+
+  object AttributeKeys {
+    def remoteAddress: AttributeKey[RemoteAddress] = AttributeKey[RemoteAddress]("remote-address")
```
raboof, line 490: How do we make this key available for Java users?

akka-http-core/src/test/scala/akka/http/impl/engine/server/HttpServerSpec.scala
changvvb: Good idea!

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
changvvb: It's a problem. I will try to resolve it later :).

lightbend-cla-validator:

lightbend-cla-validator:

lightbend-cla-validator:

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -123,14 +122,23 @@ private[http] object HttpServerBluePrint {
       override def onPush(): Unit = grab(in) match {
         case RequestStart(method, uri, protocol, hdrs, entityCreator, _, _) =>
           val effectiveMethod = if (method == HttpMethods.HEAD && settings.transparentHeadRequests) HttpMethods.GET else method
+          lazy val remoteAddressOpt = inheritedAttributes.get[HttpAttributes.RemoteAddress].map(_.address)
```
raboof, line 125: It's nicer to keep this where it was, because then it's only evaluated once per connection, instead of once per request on that connection.

akka-http-core/src/main/java/akka/http/javadsl/model/AttributeKeys.java
```diff
@@ -0,0 +1,9 @@
+/*
+ * Copyright (C) 2019-2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.http.javadsl.model;
+
+public final class AttributeKeys {
+    public static final AttributeKey<RemoteAddress> remoteAddress = (AttributeKey<RemoteAddress>)(Object)akka.http.scaladsl.model.AttributeKeys.remoteAddress();
```
raboof, line 8: I think if you import `akka.http.impl.util.JavaMapping._` you can use `akka.http.scaladsl.model.AttributeKeys.remoteAddress().asJava` instead of having to do the cast

changvvb: This is a java class and I cannot find  where `JavaMaping` used in java.

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
changvvb: Oh, I know why it was there.

akka-http-core/src/main/java/akka/http/javadsl/model/AttributeKeys.java
changvvb: On the other hand, `AttributeKey[T]` has a type parameter, I'm not sure `JavaMapping` can convert this kind of type.

raboof: Yeah, I see. `JavaMapping` could handle that, but I agree it's probably not worth moving this class from Java to Scala just for that.

The cast is safe (since the scala AttributeKey subclasses the java AttributeKey and the scala RemoteAddress subclasses the Java RemoteAddress), and I don't really see a neater way to write it, so this might be fine.

raboof:
Nice, thanks!

akka-http-core/src/main/mima-filters/10.1.x.backwards.excludes
```diff
@@ -0,0 +1,6 @@
+# add remoteAddressAttribute for akka.http.scaladsl.settings.ServerSettings
```
jrudolph, line 1:
```suggestion
# add remoteAddressAttribute to @DoNotInherit akka.http.scaladsl.settings.ServerSettings
```

jrudolph, line 1: This needs to go into a different file. Can you move it to `akka-http-core/src/main/mima-filters/10.1.x.backwards.excludes/3051-remote-address-attribute.excludes`?

akka-http-core/src/main/resources/reference.conf
```diff
@@ -101,6 +101,10 @@ akka.http {
     # holding the clients (remote) IP address.
     remote-address-header = off
 
+    # Enables/disables the addition of a remote-address attribute in HttpRequest
+    # holding the clients (remote) IP address.
```
jrudolph, line 105:
```suggestion
    # holding the clients (remote) IP address. This is preferred over `remove-address-header`
    # because it cannot be confused with a real header.
```

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -125,12 +125,20 @@ private[http] object HttpServerBluePrint {
...
+              headers.`Remote-Address`(RemoteAddress(remoteAddressOpt.get)) +: hdrs
             else hdrs
 
           val entity = createEntity(entityCreator) withSizeLimit settings.parserSettings.maxContentLength
-          push(out, HttpRequest(effectiveMethod, uri, effectiveHeaders, entity, protocol))
+          val httpRequest = HttpRequest(effectiveMethod, uri, effectiveHeaders, entity, protocol)
+
+          val effectiveHttpRequest = if (settings.remoteAddressAttribute) {
+            remoteAddressOpt.fold(httpRequest) { remoteAddress =>
+              httpRequest.addAttribute[RemoteAddress](AttributeKeys.remoteAddress, RemoteAddress(remoteAddress))
```
jrudolph, line 137:
```suggestion
              httpRequest.addAttribute(AttributeKeys.remoteAddress, RemoteAddress(remoteAddress))
```

akka-http-core/src/main/scala/akka/http/scaladsl/model/AttributeKeys.scala
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright (C) 2019-2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.http.scaladsl.model
+
+object AttributeKeys {
```
jrudolph, line 7: :+1: Seems like the right place to put these.

changvvb:
> A few small things below. The `extractClientIp` directive documentation needs to be adapted.
> 
> I wonder if we really want two separate config options, though?
> 
> We still have the potential confusion in `extractClientIp` about which should have precedence. Is that something we'd like to tackle in a follow-up?

Yes, we should document how extractClientIp obtain the ip clearly. I will make new PR to update the documentation later.

jrudolph:
> I wonder if we really want two separate config options, though?

@raboof what do you think about offering these now in two config settings? Alternatively, we could make `remote-address-header` a three-way `on | off | attribute`. Or is our plan to deprecate `remote-address-header` and offer only `remote-address-attribute` going forward?

raboof:
Yes, I think it would be nice to eventually deprecate the synthetic remote-address header - keeping them separate configuration options (and eventually removing `remote-address-header`) seems the most natural way to get there.

jrudolph:
Great, thanks a lot, @changvvb.

Issue #2994, opened 2020-03-04
Is building with Akka as a source dependency still useful?


raboof:
There is some machinery to make developing Akka HTTP against Akka as a source dependency possible.

Now that the split between Akka and Akka HTTP has stabilized, and Akka is publishing snapshot artifacts to test against (which is also simple, though somewhat slow, locally), perhaps we should remove this machinery?

jrudolph:
If we can find a way to consume the published snapshots, then I'm for it. If we just publish them locally, we still occur a big part of the cost (though maybe not all of the complexity).

PR #3055, opened 2020-04-14
build: test against published snapshots instead of source deps

raboof:
Refs #2994

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -36,6 +36,10 @@ inThisBuild(Def.settings(
   Formatting.formatSettings,
   shellPrompt := { s => Project.extract(s).currentProject.id + " > " },
   concurrentRestrictions in Global += Tags.limit(Tags.Test, 1),
+  onLoad in Global := {
+    sLog.value.info(s"Building Akka HTTP ${version.value} against Akka ${AkkaDependency.akkaVersion}")
+    (onLoad in Global).value
+  }
 ))
 
 lazy val root = Project(
--- a/project/AkkaDependency.scala
+++ b/project/AkkaDependency.scala
@@ -14,7 +14,7 @@ object AkkaDependency {
     // so 'x.y', 'x.y.z', 'current' or 'snapshot'
     def link: String
   }
-  case class Artifact(version: String) extends Akka {
+  case class Artifact(version: String, isSnapshot: Boolean = false) extends Akka {
     override def link = VersionNumber(version) match { case VersionNumber(Seq(x, y, _*), _, _) => s"$x.$y" }
   }
   case class Sources(uri: String, link: String = "current") extends Akka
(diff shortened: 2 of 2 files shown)


project/AkkaDependency.scala
```diff
@@ -54,15 +55,28 @@ object AkkaDependency {
...
   }
+
+  private def determineLatestSnapshot(prefix: String = ""): String = {
+    import sbt.librarymanagement.Http.http
+    import gigahorse.GigahorseSupport.url
+    import scala.concurrent.Await
+    import scala.concurrent.duration._
+
+    val body = Await.result(http.run(url("https://repo.akka.io/snapshots/com/typesafe/akka/akka-actor_2.12/")), 10.seconds).bodyAsString
+    """href="([^?/].*?)/"""".r.findAllMatchIn(body).map(_.group(1)).filter(_.startsWith(prefix)).toList.last
```
jrudolph, line 80: Seems a bit brittle but might be good enough. How sure are we that the listing is ordered by version? I.e. the dir listing is ordered lexically and lexical order == version order.

I think we should at least log the result here or somewhere else so it's easier to figure out if somethings goes wrong.

raboof: It's ordered by file timestamp right now. That's obviously brittle, though most approaches are... I'd propose using this until we find out we need something better ;).

jrudolph: Ok, timestamp sounds good, maybe better than lexical.

build.sbt
```diff
@@ -36,6 +36,10 @@ inThisBuild(Def.settings(
   Formatting.formatSettings,
   shellPrompt := { s => Project.extract(s).currentProject.id + " > " },
   concurrentRestrictions in Global += Tags.limit(Tags.Test, 1),
+  onLoad in Global := {
+    sLog.value.info(s"Building Akka HTTP ${version.value} against Akka ${AkkaDependency.akkaVersion}")
```
jrudolph, line 40: :+1:

jrudolph:
Running a subset of nightlies here: https://jenkins.akka.io:8498/job/akka-http-nightly/

jrudolph:
Looking good (https://jenkins.akka.io:8498/job/akka-http-nightly/AKKA_VERSION=master,Node=akka-http-nightly-node,SCALA_VERSION=2.12.10,jdk=AdoptOpenJDK%208/89/consoleFull):

```
Building Akka HTTP 10.1.11+166-b13ea0be against Akka 2.6.4+117-d611bb0d
```

jrudolph:
I tried reenabling also the builds for `release-2.5` but it failed with a compilation error in the docs. I guess that might be fixed easily. Could you try locally if compiling docs works with `release-2.5`?

raboof:
That's not expected to work since #2996... I guess we should either:

* not build the docs on `release-2.5`
* always build the docs with 2.6.4, even if there's a different version specified for the rest
* Also build the docs with the specified Akka version, unless that's a 2.5 version, in that case fall back to 2.6.4 again

I guess the last is the nicest, even if it's also a bit tricky logic. I'll PR

PR #3070, opened 2020-04-16
Introduce discovery-based client transport

raboof:
References #92

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/DiscoveryClientTransport.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/DiscoveryClientTransport.scala
@@ -1,3 +1,7 @@
+/*
+ * Copyright (C) 2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
 package akka.http.scaladsl
 
 import java.net.InetSocketAddress
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/DiscoveryClientTransport.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/DiscoveryClientTransport.scala
@@ -103,6 +103,11 @@ class DiscoveryClientTransport(discovery: ServiceDiscovery) extends ClientTransp
           //will later be propagated to the materialized flow (by examining isAvailable(out))
           override def onPull(): Unit = {}
 
+          var downstreamFinished: Boolean = false
+          override def onDownstreamFinish(): Unit = {
+            downstreamFinished = true
+          }
+
           def onFuture(futureRes: Try[Flow[In, Out, M]]) = futureRes match {
             case Failure(exception) =>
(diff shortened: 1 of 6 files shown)


jrudolph:
Interesting! Why would we use discovery instead of DNS? Is it because we have it and it is more general, so we could e.g. make use of SRV records? Wouldn't it be potentially confusing that the host from the URI is resolved using something else than DNS?

akka-http-core/src/main/scala/akka/http/scaladsl/DiscoveryClientTransport.scala
```diff
@@ -0,0 +1,175 @@
...
+                  settings.socketOptions, halfClose = true, settings.connectingTimeout, settings.idleTimeout)
+                  .mapMaterializedValue(_.map(tcpConn => OutgoingConnection(tcpConn.localAddress, tcpConn.remoteAddress))(system.dispatcher))
+            }
+        }
+      }
+    ).mapMaterializedValue(_.flatten)
+
+  }
+
+  private def futureFlow[I, O, M](flow: Future[Flow[I, O, M]]): Flow[I, O, Future[M]] =
```
jrudolph, line 55: Could we use the 2.5 `Flow.lazyInitAsync` instead?

raboof: I tried that initially. That materializes a future that is resolved when the first element arrives. By flatmapping that future I can keep the `ClientTransport` signature intact, but then a `singleRequest` call stalls. I haven't dug deeper, but perhaps we don't send any elements before the materialized future completes?

raboof:
> Why would we use discovery instead of DNS? Is it because we have it and it is more general, so we could e.g. make use of SRV records?

Yes, that was my thinking.

> Wouldn't it be potentially confusing that the host from the URI is resolved using something else than DNS?

As long as we keep DNS as the default and only use Discovery when explicitly configured I think that might be fine?

jrudolph:
Sounds reasonable.

akka-http-core/src/main/scala/akka/http/scaladsl/DiscoveryClientTransport.scala
raboof: `NewHostConnectionPool` uses this future to trigger `onConnectionAttemptSucceeded`, after which the slot becomes `Idle` for accepting requests.

jrudolph: Makes sense, forgot that `lazyInitAsync` waits for the first element.

ennru:
I added #3202 as a more open variant of this PR.

raboof:
Not sure if this still makes sense now that we have #3202

Issue #2980, opened 2020-02-24
Add style guide for Akka HTTP route structure


jrudolph:
And add section how that compares to play route files + extra logic in handlers. More info about that here: https://www.google.com/url?q=https://www.playframework.com/documentation/2.8.x/ScalaRouting%23The-routes-file-syntax&sa=D&ust=1582537250637000&usg=AFQjCNFOSZ17O-__XqjVJHiyHGkf15FoBA

ennru:
Fixed with #3079 and #3099. Can be improved ofc...

PR #3079, opened 2020-04-17
Docs: routing DSL compared with Play routes

ennru:
Add a page in the routing docs which shows the examples from the Play docs in Akka HTTP notation with accompanying tests.

References #2980

### DIFF
--- a/docs/src/main/paradox/routing-dsl/directives/parameter-directives/parameters.md
+++ b/docs/src/main/paradox/routing-dsl/directives/parameter-directives/parameters.md
@@ -117,7 +117,7 @@ Java
 
 ### Parameter with required value
 
-@scala[The `!` decorator makes the route match only if the parameter contains the specified value.]
+@scala[The `requiredValue` decorator makes the route match only if the parameter contains the specified value.]
 @java[The directive `parameterRequiredValue` makes the route match only if the parameter contains the specified value.]
 
 Scala
--- a/docs/src/main/paradox/routing-dsl/directives/parameter-directives/parameters.md
+++ b/docs/src/main/paradox/routing-dsl/directives/parameter-directives/parameters.md
@@ -117,7 +117,7 @@ Java
 
 ### Parameter with required value
 
-@java[The `!` decorator makes the route match only if the parameter contains the specified value.]
+@scala[The `!` decorator makes the route match only if the parameter contains the specified value.]
 @java[The directive `parameterRequiredValue` makes the route match only if the parameter contains the specified value.]
 
 Scala
--- a/docs/src/main/paradox/routing-dsl/directives/parameter-directives/parameters.md
+++ b/docs/src/main/paradox/routing-dsl/directives/parameter-directives/parameters.md
@@ -117,6 +117,9 @@ Java
 
 ### Parameter with required value
 
+@java[The `!` decorator makes the route match only if the parameter contains the specified value.]
+@java[The directive `parameterRequiredValue` makes the route match only if the parameter contains the specified value.]
+
 Scala
 :   @@snip [ParameterDirectivesExamplesSpec.scala]($test$/scala/docs/http/scaladsl/server/directives/ParameterDirectivesExamplesSpec.scala) { #required-value }
 
--- a/docs/src/main/paradox/routing-dsl/index.md
+++ b/docs/src/main/paradox/routing-dsl/index.md
@@ -72,7 +72,7 @@ the Routing DSL will look like:
 
 ## Compared with Play framework routes
 
-If you have been using Play framework's routes file notation before this @ref[Play! comparison](play-comparison.md) may help you to get started with Akka HTTP routing.
+If you have been using Play framework's routes file notation before this @ref[Play comparison](play-comparison.md) may help you to get started with Akka HTTP routing.
 
 <a name="interaction-with-akka-typed">
 ## Interaction with Actors
--- a/docs/src/main/paradox/routing-dsl/index.md
+++ b/docs/src/main/paradox/routing-dsl/index.md
@@ -16,6 +16,7 @@ from a background with non-"streaming first" HTTP Servers.
 @@@ index
 
 * [overview](overview.md)
+* [play](play-comparison.md)
 * [routes](routes.md)
 * [directives/index](directives/index.md)
 * [rejections](rejections.md)
@@ -68,6 +69,10 @@ the Routing DSL will look like:
 @@snip [HttpServerExampleSpec.scala]($test$/scala/docs/http/scaladsl/HttpServerExampleSpec.scala) { #long-routing-example }
 
 @@@
+
+## Compared with Play framework routes
+
+If you have been using Play framework's routes file notation before this @ref[Play! comparison](play-comparison.md) may help you to get started with Akka HTTP routing.
 
 <a name="interaction-with-akka-typed">
 ## Interaction with Actors
--- a/docs/src/main/paradox/routing-dsl/play-comparison.md
+++ b/docs/src/main/paradox/routing-dsl/play-comparison.md
@@ -1,0 +1,162 @@
+# Compared with Play routes
+
+If you have been using @scala[[Play's routes file syntax](https://www.playframework.com/documentation/2.8.x/ScalaRouting#The-routes-file-syntax)]@java[[Play's routes file syntax](https://www.playframework.com/documentation/2.8.x/JavaRouting#The-routes-file-syntax)] earlier, this page may help you to use the Akka HTTP routing DSL.
+
+Where Play routes require to specify the HTTP method for every entry, in Akka HTTP it may be specified on different levels of the `Route`.
+
+Both Play and Akka HTTP choose the first matching route within the routes file/routes definition. In Play routes are listed with one route per line, in Akka HTTP multiple routes must be concatenated with the `concat` method.
+
+### Static path
+
+For example, to exactly match incoming `GET /clients/all` requests, you can define this route in Play!.
+
+```
+GET   /clients/all          controllers.Clients.list()
+```
+
+In Akka HTTP every path segment is specified as a separate `String` @scala[concatenated with the `/` method]@java[concatenated by the `slash` method on `segment`].
+
+Scala
+:   @@snip [snip](/docs/src/test/scala/docs/http/scaladsl/server/PlayRoutesComparisonSpec.scala) { #fixed }
+
+Scala test
+:   @@snip [snip](/docs/src/test/scala/docs/http/scaladsl/server/PlayRoutesComparisonSpec.scala) { #fixed-test }
+
+Java
+:   @@snip [snip](/docs/src/test/java/docs/http/javadsl/server/testkit/PlayRoutesComparisonTest.java) { #fixed }
+
+Java test
+:   @@snip [snip](/docs/src/test/java/docs/http/javadsl/server/testkit/PlayRoutesComparisonTest.java) { #fixed-test }
+
+
+### Dynamic parts
+
+If you want to define a route that retrieves a client by ID, you’ll need to add a dynamic part.
+
+```
+GET   /clients/:id          controllers.Clients.show(id: Long)
+```
+
+Akka HTTP uses @ref[path matchers](path-matchers.md#basic-pathmatchers) which match certain data types and pass their data on.
+
+Scala
+:   @@snip [snip](/docs/src/test/scala/docs/http/scaladsl/server/PlayRoutesComparisonSpec.scala) { #long }
+
+Scala test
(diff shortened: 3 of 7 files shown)


docs/src/test/scala/docs/http/scaladsl/server/PlayRoutesComparisonSpec.scala
```diff
@@ -0,0 +1,137 @@
...
+      val itemParameterList: Route =
+        // #parameter-list
+        // TODO:
+        // Note that implicit conversions are not applicable because they are ambiguous:
+        //  both method augmentString in object Predef of type (x: String)scala.collection.immutable.StringOps
+        //  and method _string2NR in trait ToNameReceptacleEnhancements of type (string: String)akka.http.scaladsl.common.NameReceptacle[String]
+        //  are possible conversion functions from String("item") to ?{def *: ?}
+        //         path("api" / "list-items")(parameters("item".*) { items =>
+        get {
+          path("api" / "list-items")(parameters(ToNameReceptacleEnhancements._string2NR("item").*) { items =>
```
raboof, line 124: If we want to keep PR's independent, we could keep showing the symbol syntax here for now until we introduce syntax for this that works with strings?

docs/src/main/paradox/routing-dsl/index.md
```diff
@@ -69,6 +70,10 @@ the Routing DSL will look like:
 
 @@@
 
+## Compared with Play framework routes
+
+If you have been using Play framework's routes file notation before this @ref[Play! comparison](play-comparison.md) may help you to get started with Akka HTTP routing.
```
ignasi35, line 75:
```suggestion
If you have been using Play framework's routes file notation before this @ref[Play comparison](play-comparison.md) may help you to get started with Akka HTTP routing.
```

I don't think we use the `!` anymore.

docs/src/main/paradox/routing-dsl/play-comparison.md
```diff
@@ -0,0 +1,162 @@
...
+
+If you have been using @scala[[Play's routes file syntax](https://www.playframework.com/documentation/2.8.x/ScalaRouting#The-routes-file-syntax)]@java[[Play's routes file syntax](https://www.playframework.com/documentation/2.8.x/JavaRouting#The-routes-file-syntax)] earlier, this page may help you to use the Akka HTTP routing DSL.
+
+Where Play routes require to specify the HTTP method for every entry, in Akka HTTP it may be specified on different levels of the `Route`.
+
+Both Play and Akka HTTP choose the first matching route within the routes file/routes definition. In Play routes are listed with one route per line, in Akka HTTP multiple routes must be concatenated with the `concat` method.
+
+### Static path
+
+For example, to exactly match incoming `GET /clients/all` requests, you can define this route in Play!.
```
ignasi35, line 11:
```suggestion
For example, to exactly match incoming `GET /clients/all` requests, you can define this route in Play.
```

etc...

docs/src/test/java/docs/http/javadsl/server/testkit/PlayRoutesComparisonTest.java
```diff
@@ -0,0 +1,162 @@
...
+    static String apiItems(List<String> items) {
+        return items.stream().collect(Collectors.joining(","));
+    }
+
+    static class Routes extends AllDirectives {
+
+        Route clientsAll() {
+            return
+                    // #fixed
+                    get(() -> path(segment("clients").slash("all"), () -> complete(Clients.list())));
```
ignasi35, line 55: Screenshot 2020-04-17 at 16 43 59


I found all these one-liners (in both tests and main code for Java) hard to read. I think that, even if it's not java idiomatic and feels for scala-esque, somthing like the following is more readable:

```suggestion
  get(() -> 
    path(segment("clients").slash("all"), 
      () -> complete(Clients.list())
    )
  );
```

I mean, if we're writing these docs for someone coming from Play, I think it's visually easier to parse this format then the one-liners and gently introduces the reader to a slightly different format.

docs/src/main/paradox/routing-dsl/routes.md
```diff
@@ -88,8 +88,8 @@ of either the incoming request, the outgoing response or both
  * Route filtering, which only lets requests satisfying a given filter condition pass and rejects all others
  * Route chaining, which tries a second route if a given first one was rejected
 
-The last point is achieved with the concatenation operator `~`, which is an extension method that becomes available
-when you `import akka.http.scaladsl.server.Directives._`.
+@scala[The last point is achieved with the concatenation operator `~` (or the `concat` method), which is an extension method that becomes available
```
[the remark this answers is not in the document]
ennru: I removed the `~` altogether from here.

docs/src/test/scala/docs/http/scaladsl/server/PlayRoutesComparisonSpec.scala
ennru: Did that now. Other PR incoming...

docs/src/test/java/docs/http/javadsl/server/testkit/PlayRoutesComparisonTest.java
ennru: Reformatted.

docs/src/main/paradox/routing-dsl/directives/parameter-directives/parameters.md
```diff
@@ -115,6 +115,9 @@ Java
 
 ### Parameter with required value
 
+@scala[The `!` decorator makes the route match only if the parameter contains the specified value.]
```
jrudolph, line 118: Does that need to be updated after #3082?

docs/src/main/paradox/routing-dsl/play-comparison.md
```diff
@@ -0,0 +1,162 @@
+# Compared with Play routes
+
+If you have been using @scala[[Play's routes file syntax](https://www.playframework.com/documentation/2.8.x/ScalaRouting#The-routes-file-syntax)]@java[[Play's routes file syntax](https://www.playframework.com/documentation/2.8.x/JavaRouting#The-routes-file-syntax)] earlier, this page may help you to use the Akka HTTP routing DSL.
+
+Where Play routes require to specify the HTTP method for every entry, in Akka HTTP it may be specified on different levels of the `Route`.
```
jrudolph, line 5: Not quite sure what that means and why it is the first thing to say here :)

ennru:
This will require some updated if we agree on #3082

docs/src/main/paradox/routing-dsl/routes.md
```diff
@@ -173,7 +173,7 @@ A sealed route has these properties:
 As described in @ref[Rejections](rejections.md) and @ref[Exception Handling](exception-handling.md),
 there are generally two ways to handle rejections and exceptions.
 
- * Bring rejection/exception handlers @scala[`into implicit scope at the top-level`]@java[`seal()` method of the @scala[@scaladoc[Route](akka.http.scaladsl.server.index#Route=akka.http.scaladsl.server.RequestContext=%3Escala.concurrent.Future[akka.http.scaladsl.server.RouteResult])]@java[@apidoc[Route]]]
+ * Bring rejection/exception handlers @scala[into `implicit` scope at the top-level]@java[`seal()` method of the @apidoc[Route]]
```
jrudolph, line 176: That's not correct even for Scala any more.

docs/src/main/paradox/routing-dsl/play-comparison.md
```diff
@@ -0,0 +1,162 @@
+# Compared with Play routes
+
+If you have been using @scala[[Play's routes file syntax](https://www.playframework.com/documentation/2.8.x/ScalaRouting#The-routes-file-syntax)]@java[[Play's routes file syntax](https://www.playframework.com/documentation/2.8.x/JavaRouting#The-routes-file-syntax)] earlier, this page may help you to use the Akka HTTP routing DSL.
+
```
jrudolph, line 4: I think we need an overview section about the more general differences between play and akka-http routes.

 * internal vs. external DSL
 * what is covered (and what not) by play route files vs. what is covered by Akka HTTP routing

docs/src/test/scala/docs/http/scaladsl/server/PlayRoutesComparisonSpec.scala
```diff
@@ -0,0 +1,131 @@
...
+
+    def get(id: Long): String = "clientB"
+  }
+
+  "Play examples" should {
+
+    "show a fixed URL" in {
+      val clientsAll: Route =
+        // #fixed
+        (get & path("clients" / "all")) {
```
jrudolph, line 28: While you could write it that way, I think we should recommend a different style in our new best practices guide. There, we would prefer matching on path first and then on method. But in general, those might not be in the same line of code. We could still show them in the same line of code but would then point back to that guide about how you would do it more generally.

docs/src/test/scala/docs/http/scaladsl/server/PlayRoutesComparisonSpec.scala
```diff
@@ -0,0 +1,131 @@
...
+      Get("/clients/all") ~> clientsAll ~> check {
+        responseAs[String] shouldEqual "clientA,clientB,clientC"
+      }
+      // #fixed-test
+    }
+
+    "show reading a long" in {
+      val clientById: Route =
+        // #long
+        (get & pathPrefix("client" / LongNumber)) { id =>
```
jrudolph, line 43: Probably `path` here but the same comment as above applies.

docs/src/test/scala/docs/http/scaladsl/server/PlayRoutesComparisonSpec.scala
```diff
@@ -0,0 +1,131 @@
...
+      }
+      // #long-test
+    }
+
+    "use multiple segments" in {
+      def download(name: String) = s"$name: file contents"
+
+      val files: Route =
+        // #segments
+        (get & pathPrefix("files" / Segments)) { names =>
```
jrudolph, line 60: You would probably use `Remaining` or `RemainingPath` instead here.

docs/src/test/scala/docs/http/scaladsl/server/PlayRoutesComparisonSpec.scala
```diff
@@ -0,0 +1,131 @@
...
+      }
+      // #segments-test
+    }
+
+    "require a parameter" in {
+      def getPage(name: String) = s"The requested [$name]."
+
+      val pageParameter: Route =
+        // #mandatory-parameter
+        (get & parameter("page")) { page =>
```
jrudolph, line 78: We should add a path here as well to make it consistent with Play.

docs/src/test/scala/docs/http/scaladsl/server/PlayRoutesComparisonSpec.scala
```diff
@@ -0,0 +1,131 @@
...
+      // #mandatory-parameter-test
+    }
+
+    "accept a parameter" in {
+      def listAll(version: Option[String]) = version.fold("ff")(_ => "aa,bb,cc")
+
+      val optionalPageParameter: Route =
+        // #optional-parameter
+        get {
+          path("api" / "list-all")(parameter("version".?) { version =>
```
jrudolph, line 96: Let's either use proper nesting with one directive per line, or use `&`. I.e.

```suggestion
          (path("api" / "list-all") & parameter("version".?)) { version =>
```

docs/src/test/scala/docs/http/scaladsl/server/PlayRoutesComparisonSpec.scala
ignasi35: Hmm, that's a good idea. If, in the very fisrt example we introduced:

```
// Play routes file
GET				/clients/all    controllers.getAll()

// Akka-HTTP
(get & path("clients" / "all")) {
  ...
}

// idiomatic Akka-HTTP
path("clients" / "all"){
  get {
    ...
  }
}
```

So users can see it's possible to do a more direct conversion but we specify, early on in the page, that the idiomatic Akka-HTTP approach is to match on the path before matching on the method.

ennru: I'd like to keep the focus on the direct comparison and link to the style guide.

ignasi35: Ah, good idea! Introducing the style guide early is a better approach.

docs/src/main/paradox/routing-dsl/play-comparison.md
ennru: I added a more general introduction, but I don't want to go too deep as I think the main purpose of this is the side-by-side illustration.

jrudolph:
Woops, that should have been squashed, @ignasi35. ;)

PR #3080, opened 2020-04-17
Introduce sbt-java-formatter

raboof:
This is mostly a 'straw man' PR, I don't think we want to do this

### DIFF
--- a/docs/src/test/java/docs/http/javadsl/server/testkit/MyAppFragment.java
+++ b/docs/src/test/java/docs/http/javadsl/server/testkit/MyAppFragment.java
@@ -3,22 +3,17 @@
  */
 
 package docs.http.javadsl.server.testkit;
-//#source-quote
+// #source-quote
 import akka.http.javadsl.server.AllDirectives;
 import akka.http.javadsl.server.Route;
 
 public class MyAppFragment extends AllDirectives {
 
-    public Route createRoute() {
-        return
-                //#fragment
-                pathEnd(() ->
-                        get(() ->
-                                complete("Fragments of imagination")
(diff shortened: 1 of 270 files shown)


docs/src/test/java/docs/http/javadsl/server/testkit/MyAppFragment.java
```diff
@@ -3,22 +3,17 @@
...
-                                complete("Fragments of imagination")
-                        )
-                );
-                //#fragment
-    }
-
+  public Route createRoute() {
+    return
+    // #fragment
+    pathEnd(() -> get(() -> complete("Fragments of imagination")));
```
raboof, line 15: Cramming routes on one line like this really isn't nice, so that alone is a reason not to use sbt-java-formatter I think (though some of the other proposed changes are nice).

ignasi35: Time to fork https://github.com/google/google-java-format (joking)

ignasi35: The collapse of routes DSL into a single line is unfortunate and won't pay off for the rest of improvements. :-(

ennru: We could try to use it for `compile` scope only, assuming that most of the DSL use is in `test` scope.

raboof: Not sure if it's worth it then...

Issue #3081, opened 2020-04-17
Routing DSL: Parameter enhancement `.*` clashes with `predef.StringOps`


ennru:
To enable the `"parameter".*` notation to capture a list of values for an URL parameter, the `ToNameReceptacleEnhancements` contains an implicit conversion of `String` to `NameReceptacle[String]`.

This implicit is ambiguous with `Predef`s `augmentString` which adds the `*` (times) method on Strings:

```
 found   : String("city")
 required: ?{def *: ?}
Note that implicit conversions are not applicable because they are ambiguous:
 both method augmentString in object Predef of type (x: String)scala.collection.immutable.StringOps
 and method _string2NR in trait ToNameReceptacleEnhancements of type (string: String)akka.http.scaladsl.common.NameReceptacle[String]
 are possible conversion functions from String("city") to ?{def *: ?}
      parameters(("color", "city".*)) { (color, cities) =>
```

The conversion works fine as long as Scala symbols are used, which is what the docs show.

References #2738, #3079

jrudolph:
Good catch, interesting that no one noticed yet. I guess using `.*` is somewhat obscure anyway.

PR #3082, opened 2020-04-17
Routing DSL: Names for to symbolic enhancements

ennru:
As Scala symbols lose their language support in Scala 2.13 users should switch to using regular strings in the parameters and form fields directives.

The `String.*` enhancement conflicts with the `StringOps.*` enhancement from `Predef` (see #3081), so with Strings users must switch to the new `.repeated` enhancer.

* Complement the symbolic methods with regular method names in `NameReceptacle` and `NameUnmarshallerReceptacle`
* The names mimic the names used in the Java DSL
* Remove the use of Scala symbols in the docs.

### DIFF
--- a/akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/FormFieldDirectivesSpec.scala
+++ b/akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/FormFieldDirectivesSpec.scala
@@ -41,21 +41,21 @@ class FormFieldDirectivesSpec extends RoutingSpec {
   "The 'formFields' extraction directive" should {
     "properly extract the value of www-urlencoded form fields" in {
       Post("/", urlEncodedForm) ~> {
-        formFields('firstName, "age".as[Int], 'sex.?, "VIP" ? false) { (firstName, age, sex, vip) =>
+        formFields("firstName", "age".as[Int], "sex".optional, "VIP".withDefault(false)) { (firstName, age, sex, vip) =>
           complete(firstName + age + sex + vip)
         }
       } ~> check { responseAs[String] shouldEqual "Mike42Nonefalse" }
     }
     "properly extract the value of www-urlencoded form fields when an explicit unmarshaller is given" in {
       Post("/", urlEncodedForm) ~> {
-        formFields('firstName, "age".as(HexInt), 'sex.?, "VIP" ? false) { (firstName, age, sex, vip) =>
+        formFields("firstName", "age".as(HexInt), "sex".optional, "VIP".withDefault(false)) { (firstName, age, sex, vip) =>
           complete(firstName + age + sex + vip)
         }
       } ~> check { responseAs[String] shouldEqual "Mike66Nonefalse" }
     }
     "properly extract the value of multipart form fields" in {
       Post("/", multipartForm) ~> {
-        formFields('firstName, "age", 'sex.?, "VIP" ? nodeSeq) { (firstName, age, sex, vip) =>
+        formFields("firstName", "age", "sex".optional, "VIP".withDefault(nodeSeq)) { (firstName, age, sex, vip) =>
           complete(firstName + age + sex + vip)
         }
       } ~> check { responseAs[String] shouldEqual "Mike<int>42</int>None<b>yes</b>" }
@@ -70,7 +70,7 @@ class FormFieldDirectivesSpec extends RoutingSpec {
     }
     "reject the request with a MissingFormFieldRejection if a required form field is missing" in {
       Post("/", urlEncodedForm) ~> {
-        formFields('firstName, "age", 'sex, "VIP" ? false) { (firstName, age, sex, vip) =>
+        formFields("firstName", "age", 'sex, "VIP".withDefault(false)) { (firstName, age, sex, vip) =>
           complete(firstName + age + sex + vip)
         }
       } ~> check { rejection shouldEqual MissingFormFieldRejection("sex") }
@@ -78,7 +78,7 @@ class FormFieldDirectivesSpec extends RoutingSpec {
     "properly extract the value if only a urlencoded deserializer is available for a multipart field that comes without a" +
       "Content-Type (or text/plain)" in {
         Post("/", multipartForm) ~> {
-          formFields('firstName, "age", 'sex.?, "VIPBoolean" ? false) { (firstName, age, sex, vip) =>
+          formFields("firstName", "age", "sex".optional, "VIPBoolean".withDefault(false)) { (firstName, age, sex, vip) =>
             complete(firstName + age + sex + vip)
           }
         } ~> check {
@@ -87,7 +87,7 @@ class FormFieldDirectivesSpec extends RoutingSpec {
       }
     "work even if only a FromStringUnmarshaller is available for a multipart field with custom Content-Type" in {
       Post("/", multipartFormWithTextHtml) ~> {
-        formFields(('firstName, "age", 'super ? false)) { (firstName, age, vip) =>
+        formFields(("firstName", "age", "super".withDefault(false))) { (firstName, age, vip) =>
           complete(firstName + age + vip)
         }
       } ~> check {
@@ -96,7 +96,7 @@ class FormFieldDirectivesSpec extends RoutingSpec {
     }
     "work even if only a FromEntityUnmarshaller is available for a www-urlencoded field" in {
       Post("/", urlEncodedFormWithVip) ~> {
-        formFields('firstName, "age", 'sex.?, "super" ? nodeSeq) { (firstName, age, sex, vip) =>
+        formFields("firstName", "age", "sex".optional, "super".withDefault(nodeSeq)) { (firstName, age, sex, vip) =>
           complete(firstName + age + sex + vip)
         }
       } ~> check {
@@ -113,7 +113,7 @@ class FormFieldDirectivesSpec extends RoutingSpec {
       request.entity shouldNot be('strict)
 
       request ~> {
-        formFields('firstName, "age".as[Int], 'sex.?, "VIP" ? false) { (firstName, age, sex, vip) =>
+        formFields("firstName", "age".as[Int], "sex".optional, "VIP".withDefault(false)) { (firstName, age, sex, vip) =>
           complete(firstName + age + sex + vip)
         }
       } ~> check { responseAs[String] shouldEqual "Mike42Nonefalse" }
@@ -129,9 +129,9 @@ class FormFieldDirectivesSpec extends RoutingSpec {
       request.entity shouldNot be('strict)
 
       request ~> {
-        formFields('firstName) { firstName =>
-          formFields("age".as[Int], 'sex.?) { (age, sex) =>
-            formFields("VIP" ? false) { vip =>
+        formFields("firstName") { firstName =>
+          formFields("age".as[Int], "sex".optional) { (age, sex) =>
+            formFields("VIP".withDefault(false)) { vip =>
               complete(firstName + age + sex + vip)
             }
           }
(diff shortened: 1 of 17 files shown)


akka-http/src/main/scala/akka/http/scaladsl/server/directives/ParameterDirectives.scala
```diff
@@ -43,11 +43,6 @@ trait ParameterDirectives extends ToNameReceptacleEnhancements {
    * Extracts a query parameter value from the request.
    * Rejects the request if the defined query parameter matcher(s) don't match.
    *
-   * Due to a bug in Scala 2.10, invocations of this method sometimes fail to compile with an
```
jrudolph, line 46: :+1:

akka-http/src/main/scala/akka/http/scaladsl/common/NameReceptacle.scala
```diff
@@ -16,19 +17,35 @@ class NameReceptacle[T](val name: String) {
...
 
 class NameUnmarshallerReceptacle[T](val name: String, val um: FSU[T]) {
   def as[B](implicit unmarshaller: Unmarshaller[T, B]) =
     new NameUnmarshallerReceptacle(name, um.transform[B](implicit ec => implicit mat => { _ flatMap unmarshaller.apply }))
+  /** Symbolic alias for [[optional]]. */
   def ? = new NameOptionUnmarshallerReceptacle[T](name, um)
+  def optional = new NameOptionUnmarshallerReceptacle[T](name, um)
+  /** Symbolic alias for [[withDefault]]. */
   def ?(default: T) = new NameDefaultUnmarshallerReceptacle(name, default, um)
+  def withDefault(default: T) = new NameDefaultUnmarshallerReceptacle(name, default, um)
```
jrudolph, line 42: Good to have names as well here.

I find the naming a bit inconsistent with `optional`, `withDefault`, and `requiredValue` being different grammatical forms but if we have the same names on the Java side, let's keep them for consistency with those. If we have better ideas, we can still introduce different aliases later on.

docs/src/main/paradox/routing-dsl/directives/form-field-directives/formFields.md
```diff
@@ -31,14 +31,14 @@ as required, optional, or repeated, or to filter requests where a form field has
 `"color"`
 : extract value of field "color" as `String`
 
-`"color".?`
-: extract optional value of field "color" as `Option[String]`
+`"color".optional`
```
jrudolph, line 34: I'm a bit conflicted about deemphasizing the symbolic operators because they seem quite useful in this restricted setting where they are much more concise than the textual counterparts.

On the other hand, after reading the changes in the examples below, maybe it's just fine to use the names in most cases. The operators are still there if people would like to use them.

docs/src/test/scala/docs/http/scaladsl/HttpServerExampleSpec.scala
```diff
@@ -396,9 +394,9 @@ class HttpServerExampleSpec extends AnyWordSpec with Matchers
         concat(
           pathEnd {
             concat(
-              (put | parameter('method ! "put")) {
+              (put | parameter("method" ! "put")) {
```
jrudolph, line 397: :+1:

docs/src/main/paradox/routing-dsl/directives/parameter-directives/parameters.md
```diff
@@ -81,8 +77,8 @@ if the unmarshaller threw an `Unmarshaller.NoContentException` or a @apidoc[Malf
 (see also @ref[Rejections](../../../routing-dsl/rejections.md))
 
 @@@ div { .group-scala }
-There's also a singular version, @ref[parameter](parameter.md). Form fields can be handled in a similar way, see `formFields`. If
-you want unified handling for both query parameters and form fields, see `anyParams`.
+There's also a singular version, @ref[parameter](parameter.md). Form fields can be handled in a similar way, see @ref[`formFields`](../form-field-directives/index.md). If
+you want unified handling for both query parameters and form fields, see `anyParams`<!-- TODO: Does it exist? -->.
```
jrudolph, line 81: Good point, can be removed, `AnyParamDirectives` never made it from spray to akka-http because of the excessive complexity involved.

ennru:
I now pushed to use the names everywhere in our code. Gives a good overview how it will "feel".

docs/src/main/paradox/routing-dsl/directives/parameter-directives/parameters.md
ennru: I removed that sentence now.

akka-http/src/main/scala/akka/http/scaladsl/common/NameReceptacle.scala
```diff
@@ -16,19 +17,35 @@ class NameReceptacle[T](val name: String) {
   def as[B] = new NameReceptacle[B](name)
   def as[B](unmarshaller: Unmarshaller[T, B])(implicit fsu: FSU[T]) =
     new NameUnmarshallerReceptacle(name, fsu.transform[B](implicit ec => implicit mat => { _ flatMap unmarshaller.apply }))
+  /** Symbolic alias for [[optional]]. */
```
raboof, line 20: Too bad we don't have https://dotty.epfl.ch/api/scala/annotation/alpha.html yet

akka-http/src/main/scala/akka/http/scaladsl/server/directives/ParameterDirectives.scala
```diff
@@ -56,11 +51,6 @@ trait ParameterDirectives extends ToNameReceptacleEnhancements {
    * Extracts a number of query parameter values from the request.
    * Rejects the request if the defined query parameter matcher(s) don't match.
    *
-   * Due to a bug in Scala 2.10, invocations of this method sometimes fail to compile with an
-   * "too many arguments for method parameters" or "type mismatch" error.
-   *
-   * As a workaround add an `import ParameterDirectives.ParamMagnet` or use Scala 2.11.x.
-   *
```
raboof, line 63: :+1:

ennru:
> I like adding the non-symbolic alternatives for all enhancements, but I'd like to keep showing the symbolic names in the docs

Yes, I wasn't sure how far we should go here. So I went all the way to see what it there...

akka-http/src/main/scala/akka/http/scaladsl/common/NameReceptacle.scala
ennru: There it was, I knew something had been done to improve this.

ennru:
The Java DSL notation for `*` is `parameterList`, should the new Scala DSL name align with that `"items".list` instead of `"items".repeated`?

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/FormFieldDirectivesSpec.scala
```diff
@@ -87,7 +87,7 @@ class FormFieldDirectivesSpec extends RoutingSpec {
       }
     "work even if only a FromStringUnmarshaller is available for a multipart field with custom Content-Type" in {
       Post("/", multipartFormWithTextHtml) ~> {
-        formFields(('firstName, "age", 'super ? false)) { (firstName, age, vip) =>
+        formFields(("firstName", "age", "super".withDefault(false))) { (firstName, age, vip) =>
```
ignasi35, line 90:
```suggestion
        formFields(("firstName", "age", "super".withDefault(false))) { (firstName, age, vip) =>
```

(not really a suggestion, but a comment)

I find this syntax (no Symbol + name-based method) a lot cleaner. 👍🏼

akka-http/src/main/scala/akka/http/scaladsl/common/NameReceptacle.scala
```diff
@@ -16,19 +17,35 @@ class NameReceptacle[T](val name: String) {
...
   def ? = new NameOptionReceptacle[T](name)
+  def optional = new NameOptionReceptacle[T](name)
+  /** Symbolic alias for [[withDefault]]. */
   def ?[B](default: B) = new NameDefaultReceptacle(name, default)
+  def withDefault[B](default: B) = new NameDefaultReceptacle(name, default)
```
ignasi35, line 25:
```suggestion
  def orElse[B](default: B) = new NameDefaultReceptacle(name, default)
```
How about `orElse`? I'm not entirely sure I like it more than `withDefault`, TBH.

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/FormFieldDirectivesSpec.scala
```diff
@@ -168,58 +168,58 @@ class FormFieldDirectivesSpec extends RoutingSpec {
   "The 'formField' requirement directive" should {
     "block requests that do not contain the required formField" in {
       Post("/", urlEncodedForm) ~> {
-        formField('name ! "Mr. Mike") { completeOk }
+        formField("name".requiredValue("Mr. Mike")) { completeOk }
```
ignasi35, line 171: I found `requiredValue` a bit confusing. How about something like:

```scala
formField("name".shouldBe("Mr. Mike"))
formField("name".whenValued("Mr. Mike"))
formField("name".ifValued("Mr. Mike"))
formField("name".valued("Mr. Mike"))
formField("name".is("Mr. Mike")) // I think this is my preferred
```
so it looks like:
```scala
Post("/", urlEncodedForm) ~> { 
   formField("name".is("Mr. Mike")){ 
      ...
   }
}
```

ennru:
Great suggestions, @ignasi35 

`orElse` aligns nicely with Java `Optional`'s `orElse` (but Scala uses `getOrElse`). 

`is` makes a lot of sense.

If we agree on those, I can update this PR and do the Java alignment in another PR.

ignasi35:
> orElse aligns nicely with Java Optional's orElse (but Scala uses getOrElse).

Good call. I always confuse one with the other.

But now I doubt. I mean, even if scala's `Option` method is called `getOrElse` we're not _getting_ but _matching_. Hmmm, how would that look?

```
formFields(("firstName", "age", "super".withDefault(false))) { (firstName, age, vip) =>
formFields(("firstName", "age", "super".getOrElse(false))) { (firstName, age, vip) =>
// in java it'd be something like:
// formFields(("firstName", "age", "super".orElse(false))) { (firstName, age, vip) =>
```

Pff, I'm sure what's the alternative I dislike the least. 

PS: If we stick to `orElse`/`getOrElse`, I'm fine if each implementation (java vs scala) has a more idiomatic variant.

jrudolph:
> The Java DSL notation for `*` is `parameterList`, should the new Scala DSL name align with that `"items".list` instead of `"items".repeated`?

I like "repeated" because that's what it is parsing, repeated occurrences of a parameter even if the result is a list. It's somewhat questionable that the Java DSL overloads "parameterList" for both `repated` and the `parameterList` directive in Scala.

jrudolph:
> But now I doubt. I mean, even if scala's `Option` method is called `getOrElse` we're not _getting_ but _matching_. Hmmm, how would that look?

`orElse` is usually used to denote an alternative so that both sides are of the same type.

It seems that it's hard choose names that would be consistent, short, and intuitive. The danger is always that you choose names that are somehow smart (or appealing to the monad brain) but you would still have to look them up similar as to what the symbolic names do.

So, maybe the original names @ennru suggested are just right for the purpose: to have names that are explicit about what they do so you don't have to look it up even if they are a bit clunky.

docs/src/main/paradox/routing-dsl/directives/parameter-directives/parameters.md
```diff
@@ -29,47 +25,47 @@ The signature shown is simplified and written in pseudo-syntax, the real signatu
...
-can be supplied either as a String or as a Symbol. Parameter extraction can be modified to mark a query parameter
+is supplied as a String. Parameter extraction can be modified to mark a query parameter
 as required, optional, or repeated, or to filter requests where a parameter has a certain value:
 
 `"color"`
 : extract value of parameter "color" as `String`
 
-`"color".?`
-: extract optional value of parameter "color" as `Option[String]`
+`"color".optional`
```
jrudolph, line 34: Here we should still show the symbolic names as well (if possible with the definition list syntax).

ennru: It is shown in the text below. Adding it to the same line looked very ugly/

ennru:
I updated the descriptions (document rejections) and added scaladoc comments.

PR #3089, opened 2020-04-20
Routing DSL: Allow names to match paths

ennru:
Introduce `slash`, `or`, `and`, `append`, `optional` and `not` for path matching.

References #3082, #3085

I don't think we should promote these, just offer the alternative.

`~` becomes `concat` for routes, here I picked `append` as `concat` is a varargs method to combine multiple routes.

### DIFF
--- a/akka-http/src/main/scala/akka/http/scaladsl/server/PathMatcher.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/server/PathMatcher.scala
@@ -20,25 +20,35 @@ import akka.http.impl.util._
 abstract class PathMatcher[L](implicit val ev: Tuple[L]) extends (Path => PathMatcher.Matching[L]) { self =>
   import PathMatcher._
 
-  def / : PathMatcher[L] = this ~ PathMatchers.Slash
+  /** Alias for [[slash]]. */
+  def / : PathMatcher[L] = slash
 
-  def /[R](other: PathMatcher[R])(implicit join: Join[L, R]): PathMatcher[join.Out] =
+  def slash: PathMatcher[L] = this ~ PathMatchers.Slash
+
+  /** Alias for [[slash]]. */
+  def /[R](other: PathMatcher[R])(implicit join: Join[L, R]): PathMatcher[join.Out] = slash(other)
+
+  def slash[R](other: PathMatcher[R])(implicit join: Join[L, R]): PathMatcher[join.Out] =
     this ~ PathMatchers.Slash ~ other
 
-  def |[R >: L: Tuple](other: PathMatcher[_ <: R]): PathMatcher[R] =
+  /** Alias for [[or]]. */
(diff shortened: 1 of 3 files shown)


akka-http/src/main/scala/akka/http/scaladsl/server/PathMatcher.scala
```diff
@@ -321,6 +332,11 @@ trait ImplicitPathMatcherConstruction {
 trait PathMatchers {
   import PathMatcher._
 
+  def not(self: PathMatcher[_]): PathMatcher0 =
```
jrudolph, line 335: So, we prefer `not("test")` over `"test".not`. Seems to make sense for this one.

jrudolph:
> Test FAILed.

Sorry, enabled checking compilation of bench-jmh submodule before fixing it on master...

docs/src/main/paradox/routing-dsl/path-matchers.md
```diff
@@ -216,16 +216,16 @@ Path matchers can be combined with these combinators to form higher-level constr
...
 Tilde Operator (`~`)
 : The tilde is the most basic combinator. It simply concatenates two matchers into one, i.e if the first one matched
-(and consumed) the second one is tried. The extractions of both matchers are combined type-safely.
+(and consumed) the second one is tried. The extractions of both matchers are combined type-safely. (This is an alias for the `append` method.)
 For example: `"foo" ~ "bar"` yields a matcher that is identical to `"foobar"`.
 
 Slash Operator (`/`)
-: This operator concatenates two matchers and inserts a `Slash` matcher in between them.
+: This operator concatenates two matchers and inserts a `Slash` matcher in between them. (This is an alias for the `slash` method.)
 For example: `"foo" / "bar"` is identical to `"foo" ~ Slash ~ "bar"`.
```
ignasi35, line 224:
```suggestion
For example: `"foo" / "bar"` is identical to `"foo" slash "bar"`. You can also use the composition of `~` and path mathcers so `"foo" ~ Slash ~ "bar"` is also equivalent.
```

ignasi35: Yeah, listing all the alternatives may be too much. Let's keep it as you originally posted.

docs/src/main/paradox/routing-dsl/path-matchers.md
```diff
@@ -216,16 +216,16 @@ Path matchers can be combined with these combinators to form higher-level constr
 
 Tilde Operator (`~`)
 : The tilde is the most basic combinator. It simply concatenates two matchers into one, i.e if the first one matched
-(and consumed) the second one is tried. The extractions of both matchers are combined type-safely.
+(and consumed) the second one is tried. The extractions of both matchers are combined type-safely. (This is an alias for the `append` method.)
```
jrudolph, line 219: Maybe just remove the parentheses here and below to remove a bit of clutter.

PR #3098, opened 2020-04-23
core: move max-content-length definition from top to server/client

jrudolph:
Refs #2961

Alternative to #2998.


akka-http-core/src/main/scala/akka/http/impl/settings/ParserSettingsImpl.scala
```diff
@@ -55,6 +56,12 @@ private[akka] final case class ParserSettingsImpl(
   override def headerValueCacheLimit(headerName: String): Int =
     headerValueCacheLimits.getOrElse(headerName, defaultHeaderValueCacheLimit)
 
+  override def maxContentLength: Long =
+    maxContentLengthSetting.getOrElse(
```
jrudolph, line 60: This is quite unfortunate but a consequence of keeping maxContentLength in ParserSettings. I'll have another quick try of just moving the value out of there and directly into `ServerSettings` / `ClientConnectionSettings`.

jrudolph: I tried to push the setting to ServerSettings / ClientConnectionSettings but that really doesn't give us a lot: people still have to use `ParserSettings.forServer` / `ParserSettings.forClient` to get the right set of settings.

I guess that's the way it is. If you programmatically use e.g. `ServerSettings(system)` and then modify those settings, you will have kind of a complex stack of settings.

E.g. if we merge the per-host settings you will have this stack:

host-specific programmatic parsing settings -> host-specific application.conf parsing settings -> host-connection-pool parsing settings -> client parsing settings -> general parsing settings

which is already quite complex.

Maybe we need to provide the old setting here hard-coded ? At least for a deprecation period?

akka-http-core/src/main/scala/akka/http/javadsl/settings/ParserSettings.scala
```diff
@@ -94,7 +94,22 @@ object ParserSettings extends SettingsCompanion[ParserSettings] {
   trait ErrorLoggingVerbosity
   trait IllegalResponseHeaderValueProcessingMode
 
+  /**
+   * @deprecated Use forServer or forClient instead.
```
ennru, line 98: No since version?

akka-http-core/src/main/scala/akka/http/javadsl/settings/ParserSettings.scala
```diff
@@ -56,7 +56,7 @@ abstract class ParserSettings private[akka] () extends BodyPartParser.Settings {
   def withMaxHeaderNameLength(newValue: Int): ParserSettings = self.copy(maxHeaderNameLength = newValue)
   def withMaxHeaderValueLength(newValue: Int): ParserSettings = self.copy(maxHeaderValueLength = newValue)
   def withMaxHeaderCount(newValue: Int): ParserSettings = self.copy(maxHeaderCount = newValue)
-  def withMaxContentLength(newValue: Long): ParserSettings = self.copy(maxContentLength = newValue)
+  def withMaxContentLength(newValue: Long): ParserSettings = self.copy(maxContentLengthSetting = Some(newValue))
```
ennru, line 59:
```suggestion
  def withMaxContentLength(newValue: Long): ParserSettings = self.copy(maxContentLengthSetting = Option(newValue))
```

jrudolph: Isn't that just doing an extra unnecessary null check?

ennru: Yes, my thinking is that this is the javadsl and it doesn't contain a variant to unset the value.

jrudolph: Ah, I see. That still won't work as `scala.Long` is the primitive type. But we can introduce an explicitly method for unsetting.

akka-http-core/src/main/scala/akka/http/impl/settings/ParserSettingsImpl.scala
jrudolph: I had not addressed this point.

IIRC, it means that code patterns where users used `ClientConnectionSettings(system).withParsingSettings(ParsingSettings(system).withXYZ())` will now throw a runtime exception when used.

jrudolph: Created #3179 so we don't forget about this.

PR #3099, opened 2020-04-23
Routing DSL: style guide

ennru:
Add some guidance how we recommend users organise their routing DSL code.

References #2980

### DIFF
--- a/docs/src/main/paradox/routing-dsl/style-guide.md
+++ b/docs/src/main/paradox/routing-dsl/style-guide.md
@@ -1,0 +1,45 @@
+# Routing DSL style guide
+
+Akka HTTP's routing DSL is at the center of most Akka HTTP-based servers. It's where the incoming requests diverge into the different parts of the implemented services.
+
+Keeping all routing in one big structure will easily become hard to grasp and maintain. This page is tries to give a few hints for how you may want to break down the routing logic.
+
+Main recommendations
+
+1. Most `Route`s consist of multiple `Route`s in themselves, isolate them into values or methods.
+1. Directives combine into other directives, isolate repeated combination into values.
+1. Keep the most static part of a route outermost (eg. the fixed path segments), end with the HTTP methods.
+1. Encapsulate patterns you want to establish into helpers.
+
+## Structure
+
+### Routes are built out of directives
+
+Think of a route as a function describing how an incoming request maps to a reply (technically `RequestContext => Future[RouteResult]`) (see @ref[Routes](routes.md)). A route is expressed in directives. Directives compose into new directives (see @ref[Composing directives](directives/index.md#composing-directives)).
+
+## Paths
+
+Keep the most static part of a route outermost (eg. the fixed path segments), end with the HTTP methods.
+
+Scala
+:   @@snip[snip](/docs/src/test/scala/docs/http/scaladsl/server/directives/StyleGuideExamplesSpec.scala) { #path-outermost }
+
+Group routes with a `pathPrefix` where possible, use `path` for the last bit.
+
+Scala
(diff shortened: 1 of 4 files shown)


docs/src/main/paradox/routing-dsl/style-guide.md
```diff
@@ -0,0 +1,45 @@
+# Routing DSL style guide
+
+Akka HTTP's routing DSL is at the center of most Akka HTTP-based servers. It's where the incoming requests diverge into the different parts of the implemented services.
+
+Keeping all routing in one big structure will easily become hard to grasp and maintain. This page is tries to give a few hints for how you may want to break down the routing logic.
```
zwift-mark, line 5: `is tries` --> `tries` ?

ignasi35: I would even remove the `tries` altogether:

```suggestion
Keeping all routing in one big structure will easily become hard to grasp and maintain. This page gives a few hints for how you may want to break down the routing logic.
```

docs/src/test/scala/docs/http/scaladsl/server/directives/StyleGuideExamplesSpec.scala
```diff
@@ -0,0 +1,194 @@
...
+
+      val customerRoutes: Route =
+        concat(
+          path("customer" / IntNumber) { customerId =>
+            complete("")
+          }
+        // ...
+        )
+
+      val prefer: Route =
```
ignasi35, line 82: Add a comment here highlightng the composition.

```suggestion
      // 2. Then compose the relative routes under their corresponding path prefix.
      val prefer: Route =
```

docs/src/test/scala/docs/http/scaladsl/server/directives/StyleGuideExamplesSpec.scala
```diff
@@ -0,0 +1,194 @@
...
+          }
+        }
+      )
+      // #path-prefix
+    }
+
+    "split" in {
+      // #path-compose
+      // prefer
+      val itemRoutes: Route =
```
ignasi35, line 60: Add a comment here that first we cerate a couple of partial, relative path matchers:

```suggestion
    // 1. First, create partial matchers (with a relative path)
    val itemRoutes: Route =
```

ennru:
Thank you for hopping on this. Do you think it is useful in this form? Does the structure make sense?

ennru:
Thank you for the feedback. I'll work more on this to find a better scope.
The least I need to do is to complement this PR with Java examples.

docs/src/test/scala/docs/http/scaladsl/server/directives/StyleGuideExamplesSpec.scala
```diff
@@ -0,0 +1,192 @@
...
+          path("show" / Segment) { itemId =>
+            get {
+              complete("")
+            }
+          }
+        )
+
+      val customerRoutes: Route =
+        concat(
+          path("customer" / IntNumber) { customerId =>
```
mfulgo, line 78: This should have been
```scala
path(IntNumber) { customerId =>
```
to match the Java and avoid a confusing `/customer/customer/123` route.

ennru: Thank you. By coincidence we got a fix for this the other day in https://github.com/akka/akka-http/pull/3448

mfulgo: Perfect! Thanks.

PR #3053, opened 2020-04-13
gracefully discard unsupported h2 SETTINGS

bjaglin:
References https://github.com/akka/akka-http/pull/2039

I ran into that when bumping Envoy, running in front of my `akka-grpc` service, from 1.13.0 to 1.14.1. 
```
15:40:39.584 [debug] a.i.TcpListener - New connection accepted
15:40:39.602 [debug] a.h.i.e.h.Http2ServerDemux - Changing state from Idle to WaitingForNetworkToSendControlFrames
15:40:39.603 [debug] a.h.i.e.h.Http2ServerDemux - Changing state from WaitingForNetworkToSendControlFrames to Idle
15:40:39.603 [debug] a.h.i.e.h.Http2ServerDemux - Changing state from Idle to WaitingForData
15:40:39.608 [debug] a.a.RepointableActorRef - Aborting tcp connection to /127.0.0.1:35878 because of upstream failure: akka.stream.impl.io.ByteStringParser$ParsingException: Parsing failed in step akka.http.impl.engine.http2.framing.Http2FrameParsing$$anon$1$ReadFrame$@1266e312
akka.stream.impl.io.ByteStringParser$ParsingLogic.doParseInner(ByteStringParser.scala:83)
akka.stream.impl.io.ByteStringParser$ParsingLogic.doParse(ByteStringParser.scala:106)
akka.stream.impl.io.ByteStringParser$ParsingLogic.onPush(ByteStringParser.scala:127)
akka.stream.impl.fusing.GraphInterpreter.processPush(GraphInterpreter.scala:541)
akka.stream.impl.fusing.GraphInterpreter.execute(GraphInterpreter.scala:423)
akka.stream.impl.fusing.GraphInterpreterShell.runBatch(ActorGraphInterpreter.scala:624)
akka.stream.impl.fusing.ActorGraphInterpreter$SimpleBoundaryEvent.execute(ActorGraphInterpreter.scala:55)
akka.stream.impl.fusing.ActorGraphInterpreter$SimpleBoundaryEvent.execute$(ActorGraphInterpreter.scala:51)
akka.stream.impl.fusing.ActorGraphInterpreter$BatchingActorInputBoundary$OnNext.execute(ActorGraphInterpreter.scala:94)
akka.stream.impl.fusing.GraphInterpreterShell.processEvent(ActorGraphInterpreter.scala:599)
[...]

PR #3101, opened 2020-04-24
h2 frame parsing: improve handling of unknown values

bjaglin:
Follow-up of https://github.com/akka/akka-http/pull/3053#discussion_r408714500

The first commit tries to simplify the existence/lookup flow for spec elements, forcing a more graceful handling of unknown frame types & error codes (with a slightly different approach for error codes where I guess it's interesting to capture the code, even if it's not strongly typed).


raboof:
OK TO TEST

bjaglin:
I just added 68bf345 to leverage `OptionVal`.

PR #3106, opened 2020-04-27
core: header rendering with less indirection

jrudolph:
Slight performance improvements:

 * precompute some often used string combinations
 * in `ModeledHeader` avoid `~~` syntax and use direct call instead


akka-http-core/src/main/mima-filters/10.1.11.backwards.excludes/internal-RenderSupport-changes.excludes
```diff
@@ -0,0 +1,2 @@
+# Changes to internal methods
```
raboof, line 1: should this be `10.1.x.backwards.excludes` or do we intend to backport?

jrudolph: No, I don't think we should backport there. I'll fix it.

PR #3107, opened 2020-04-27
parsing: don't extend from Char => Boolean for CharPredicate

jrudolph:
Slight performance improvement.

That's because Function1 is not specialized for Char so there's boxing involved
(if not elided by the JVM). The solution is provide an explicit
specialized apply method instead.

### DIFF
--- a/akka-parsing/src/main/scala/akka/parboiled2/CharPredicate.scala
+++ b/akka-parsing/src/main/scala/akka/parboiled2/CharPredicate.scala
@@ -16,10 +16,15 @@
 
 package akka.parboiled2
 
+import akka.annotation.InternalApi
+
 import scala.annotation.tailrec
 import scala.collection.immutable.NumericRange
 
-sealed abstract class CharPredicate extends (Char => Boolean) {
+@InternalApi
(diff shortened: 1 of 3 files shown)


akka-parsing/src/main/scala/akka/parboiled2/CharPredicate.scala
```diff
@@ -88,15 +93,16 @@ sealed abstract class CharPredicate extends (Char => Boolean) {
...
+  protected def and(that: CharPredicate): CharPredicate =
     if (this == Empty) Empty else from(c => this(c) && that(c))
-  protected def andNot(that: Char => Boolean): CharPredicate =
-    from(if (this == Empty) c => !that(c) else c => this(c) && !that(c))
+  protected def andNot(that: CharPredicate): CharPredicate =
+    if (this == Empty) from(c => !that(c)) else from(c => this(c) && !that(c))
 }
 
-object CharPredicate {
+@InternalApi
```
jrudolph, line 104: Added that for completeness. The whole akka-parsing module does not provide binary compatibility, though.

PR #3108, opened 2020-04-27
Reuse UriParser In HttpRequestParser

jrudolph:
Seems to be a big performance improvement because the ValueStack in the UriParser can so be reused and doesn't have to be reallocated for each request.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpRequestParser.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpRequestParser.scala
@@ -10,7 +10,7 @@ import scala.annotation.{ switch, tailrec }
 import akka.http.scaladsl.settings.{ ParserSettings, WebSocketSettings }
 import akka.util.{ ByteString, OptionVal }
 import akka.http.impl.engine.ws.Handshake
-import akka.http.impl.model.parser.CharacterClasses
+import akka.http.impl.model.parser.{ CharacterClasses, UriParser }
 import akka.http.scaladsl.model.{ ParsingException => _, _ }
 import headers._
 import StatusCodes._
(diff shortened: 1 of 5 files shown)


jrudolph:
Bench run here: https://jenkins.akka.io:8498/view/Akka-http/job/akka-http-bench-adhoc/170/jmh-run-report/

Was accidentally run in reverse order, so the 10% decline is actually a 10% increase.

akka-parsing/src/main/scala/akka/parboiled2/Parser.scala
```diff
@@ -114,15 +114,18 @@ abstract class Parser(
   private var _cursor: Int = _
 
   // the value stack instance we operate on
-  private var _valueStack: ValueStack = _
+  private var _valueStack: ValueStack = new ValueStack(initialValueStackSize, maxValueStackSize)
```
raboof, line 117: So we're deviating from the 'upstream' implementation - seems fine, we can find the differences in git anyway

akka-parsing/src/main/scala/akka/parboiled2/RuleDSLBasics.scala
```diff
@@ -92,7 +92,7 @@ trait RuleDSLBasics {
   /**
    * Matches the EOI (end-of-input) character.
    */
-  def EOI: Char = akka.parboiled2.EOI
+  final def EOI: Char = akka.parboiled2.EOI
```
raboof, line 95: Why not `val`?

akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpRequestParser.scala
```diff
@@ -154,7 +157,8 @@ private[http] final class HttpRequestParser(
       val uriEnd = findUriEnd()
       try {
         uriBytes = input.slice(uriStart, uriEnd)
-        uri = Uri.parseHttpRequestTarget(new ByteStringParserInput(uriBytes), mode = uriParsingMode)
+        uriParser.input = new ByteStringParserInput(uriBytes)
+        uri = uriParser.parseHttpRequestTarget()
```
raboof, line 161: So we can be sure this never happens concurrently because we're inside a stream stage. Cool :+1:

akka-parsing/src/main/scala/akka/parboiled2/Parser.scala
jrudolph: Yeah, not sure it's still the same anyway.

akka-parsing/src/main/scala/akka/parboiled2/RuleDSLBasics.scala
jrudolph: I can remember I checked something about that, maybe I can try again.

PR #3091, opened 2020-04-21
Build: silence warnings for argument adaptation

ennru:
Introduce the scalac silencer to silence warnings for argument adaptation (auto tupleing) as they risk to hide other (for the time being) more relevant warnings.

Depending on the solutions to #2738 we would remove this again.

PR #3114, opened 2020-04-29
Enable some fatal warnings for docs

raboof:
To reduce the risk that we're still documenting API's that we for example 
have already deprecated

On top of #3091


docs/src/main/paradox/routing-dsl/directives/alphabetically.md
```diff
@@ -150,7 +150,6 @@
...
 |@scala[@ref[textract](basic-directives/textract.md)]                                                                         | @scala[Extracts a number of values using a `RequestContext => Tuple` function]                                                                                                                                     |
 |@scala[@ref[tprovide](basic-directives/tprovide.md)]                                                                         | @scala[Injects a given tuple of values into a directive]                                                                                                                                                          |
-|@ref[uploadedFile](file-upload-directives/uploadedFile.md)                                                           | Streams one uploaded file from a multipart request to a file on disk                                                                                                                                      |
```
jrudolph, line 153: What happened to this? Was it renamed and the page duplicated?

docs/src/test/scala/docs/http/scaladsl/server/ExceptionHandlerExamplesSpec.scala
```diff
@@ -77,6 +78,7 @@ object MyImplicitExceptionHandler {
   //#implicit-handler-example
 }
 
+@silent("Evaluation of a constant expression results in an arithmetic error")
```
jrudolph, line 81: Could we put these directly on the expression?

build.sbt
```diff
@@ -325,9 +325,21 @@ lazy val docs = project("docs")
...
-    // In docs adding an unused variable can be helpful, for example
-    // to show its type
-    scalacOptions += "-Xlint:-unused",
+    scalacOptions ++= Seq(
+      // Make sure we don't accidentally keep documenting deprecated calls
+      "-Xfatal-warnings",
+      // In docs adding an unused variable can be helpful, for example
+      // to show its type
+      "-Xlint:-unused",
+      // We use this for `complete`
```
jrudolph, line 334: Good point, added a comment about that to https://github.com/akka/akka-http/issues/2738#issuecomment-621109993.

docs/src/test/scala/docs/http/scaladsl/server/directives/JsonStreamingFullExamples.scala
```diff
@@ -4,8 +4,10 @@
 
 package docs.http.scaladsl.server.directives
 
+import com.github.ghik.silencer.silent
 import org.scalatest.wordspec.AnyWordSpec
 
+@silent("The outer reference in this type test cannot be checked at run time")
```
jrudolph, line 10: I never ran into a bug because of this. Maybe we could just silence it globally as well?

docs/src/main/paradox/routing-dsl/directives/alphabetically.md
raboof: It has been deprecated since 10.0.11, so I replaced the documentation page with a redirect to the replacement (https://github.com/akka/akka-http/pull/3114/files#diff-3cd40b5f95731f5d8b37e8cc9a01fc63R17).

I guess it's time to remove the code itself as well?

docs/src/test/scala/docs/http/scaladsl/server/ExceptionHandlerExamplesSpec.scala
raboof: I wanted to avoid `@silent` annotations showing up in the documentation

jrudolph: Ah, yes, makes sense of course. Alternatively we could change the code to use a variable from the request to trigger the exception which would be more realistic anyway.

docs/src/main/paradox/routing-dsl/directives/alphabetically.md
jrudolph: Yes, removing sounds good.

raboof:
travis seems happy (https://travis-ci.org/github/akka/akka-http/builds/680932072) but not reporting back

docs/src/test/scala/docs/http/scaladsl/server/directives/JsonStreamingFullExamples.scala
```diff
@@ -4,6 +4,7 @@
 
 package docs.http.scaladsl.server.directives
 
+import com.github.ghik.silencer.silent
```
jrudolph, line 7: Can we have fatal warnings but not for all warning types? It's not that `-Xlint:unused` is useless but it just shouldn't fail the build. ;)

PR #3115, opened 2020-04-29
Provide Route.toFunction as alternative to Route.asyncHandler

raboof:
On top of #3114

### DIFF
--- a/akka-http/src/main/scala/akka/http/scaladsl/server/Route.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/server/Route.scala
@@ -57,7 +57,7 @@ object Route {
    * This conversion is also implicitly available whereever a `Route` is used through [[RouteResult#routeToFlow]].
    */
   def toFlow(route: Route)(implicit system: ClassicActorSystemProvider): Flow[HttpRequest, HttpResponse, NotUsed] =
-    Flow[HttpRequest].mapAsync(1)(asyncHandler(route, RoutingSettings(system), ParserSettings.forServer))
+    Flow[HttpRequest].mapAsync(1)(toFunction(route))
 
   /**
    * Turns a `Route` into a server flow.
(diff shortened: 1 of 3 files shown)


akka-http/src/main/scala/akka/http/scaladsl/server/Route.scala
```diff
@@ -73,8 +73,10 @@ object Route {
                                 exceptionHandler: ExceptionHandler         = null): Flow[HttpRequest, HttpResponse, NotUsed] =
     Flow[HttpRequest].mapAsync(1)(asyncHandler(route))
 
-  def asyncHandler(route: Route, routingSettings: RoutingSettings, parserSettings: ParserSettings)(implicit system: ClassicActorSystemProvider): HttpRequest => Future[HttpResponse] = {
+  def toFunction(route: Route)(implicit system: ClassicActorSystemProvider): HttpRequest => Future[HttpResponse] = {
```
jrudolph, line 76: Is this just about naming?

raboof: No: akka-grpc was using the `asyncHandler` that takes a `Route` and a bunch of implicits, and had to replace it with this method that required explicit RoutingSettings and ParserSettings. Removing those, however, would clash with the existing asyncHandler.

Renaming seemed elegant (with `toFunction` being similar to `toFlow`).

Alternatively we could change the deprecated `asyncHandler` and turn some implicit parameters into explicit ones, but that does look a bit strange. I have no strong preference.

jrudolph: Ah, I see, thanks for the explanation.

I find `toFunction` a bit unexpected as `Route` already is an alias for a function. But maybe people don't even care about that (and most people wouldn't use it anyway). So, fine for me!

docs/src/test/scala/docs/http/scaladsl/server/directives/JsonStreamingFullExamples.scala
```diff
@@ -7,6 +7,7 @@ package docs.http.scaladsl.server.directives
 import com.github.ghik.silencer.silent
 import org.scalatest.wordspec.AnyWordSpec
 
+@silent("The outer reference in this type test cannot be checked at run time")
```
jrudolph, line 10: Still needed?

Issue #3100, opened 2020-04-23
Checkout ALPN support in Java 8 > u251 and update documentation build etc.


jrudolph:
See https://github.com/netty/netty/pull/10196

We should probably keep the jetty alpn agent stuff around for a while? Would be nice if we could get rid of the extra interface sooner or later.

jrudolph:
Done in #3117 and #3125 (10.1 backport).

PR #3117, opened 2020-04-29
Support native ALPN in JDK >=8u251

jrudolph:
Refs #3100

Not quite done, we need to decide if we want to 

 * drop HTTP2 support for JDK < 8u251 (and jetty-alpn agent)
 * still try to use jetty-alpn on JDK >= 8u251 if it's available (seems to be dangerous as jetty-alpn will replace SSL classes for which it is not clear whether they really support those newer JDKs)

Also needs appropriate changes in the documentation.

I tend to say that we should drop jetty-alpn support for 10.2.0 and just require a recent JDK 8 if you want to use HTTP/2. This requirement will transitively propagate to akka-grpc as well so that should be taken into account as well.

### DIFF
--- a/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2AlpnSupport.scala
+++ b/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2AlpnSupport.scala
@@ -16,6 +16,7 @@ import org.eclipse.jetty.alpn.ALPN
 import org.eclipse.jetty.alpn.ALPN.ServerProvider
 
 import scala.language.reflectiveCalls
+import scala.util.Try
 
 /**
  * INTERNAL API
@@ -36,7 +37,16 @@ private[http] object Http2AlpnSupport {
 
   def isAlpnSupportedByJDK: Boolean =
     // ALPN is supported starting with JDK 9
-    JavaVersion.majorVersion >= 9
+    JavaVersion.majorVersion >= 9 ||
+      (classOf[SSLEngine].getMethods.exists(_.getName == "setHandshakeApplicationProtocolSelector")
+        && {
+          // This method only exists in the jetty-alpn provided implementation. If it exists an old version of the jetty-alpn-agent is active which is not supported
+          // on JDK>= 8u252. When running on such a JVM, you can either just remove the agent or (if you want to support older JVMs with the same command line),
+          // use jetty-alpn-agent >= 2.0.10
+          val jettyAlpnClassesAvailable = Try(Class.forName("sun.security.ssl.ALPNExtension")).toOption.exists(_.getDeclaredMethods.exists(_.getName == "init"))
+          if (jettyAlpnClassesAvailable) throw new RuntimeException("On JDK >= 8u252 you need to either remove jetty-alpn-agent or use version 2.0.10 (which is a noop)")
+          else true
+        })
 
   private type JDK9SSLEngine = {
     def setHandshakeApplicationProtocolSelector(selector: BiFunction[SSLEngine, ju.List[String], String]): Unit
--- a/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2AlpnSupport.scala
+++ b/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2AlpnSupport.scala
@@ -4,18 +4,14 @@
 
 package akka.http.impl.engine.http2
 
-import java.util.function.BiFunction
 import java.{ util => ju }
 
-import javax.net.ssl.{ SSLEngine, SSLException }
+import javax.net.ssl.SSLEngine
 import akka.annotation.InternalApi
 import akka.http.impl.util.JavaVersion
 import akka.stream.TLSProtocol.NegotiateNewSession
 import akka.stream.impl.io.TlsUtils
-import org.eclipse.jetty.alpn.ALPN
-import org.eclipse.jetty.alpn.ALPN.ServerProvider
 
-import scala.language.reflectiveCalls
 import scala.util.Try
 
 /**
@@ -33,7 +29,7 @@ private[http] object Http2AlpnSupport {
    */
   def enableForServer(engine: SSLEngine, setChosenProtocol: String => Unit): SSLEngine =
     if (isAlpnSupportedByJDK) jdkAlpnSupport(engine, setChosenProtocol)
-    else jettyAlpnSupport(engine, setChosenProtocol)
+    else throw new RuntimeException(s"Need to run on a JVM >= 8u252 for ALPN support needed for HTTP/2. Running on ${sys.props("java.version")}")
 
   def isAlpnSupportedByJDK: Boolean =
     // ALPN is supported starting with JDK 9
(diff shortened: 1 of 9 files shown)


jrudolph:
See the docs at https://github.com/jetty-project/jetty-alpn-agent:

> Starting with OpenJDK 1.8.0_252 the use of jetty-alpn-agent is not strictly required anymore since the standard ALPN APIs have been backported to OpenJDK 1.8.0_252 from Java 9.

> Using jetty-alpn-agent with OpenJDK 1.8.0_252 or later performs no class transformation; the -javaagent option can be left in the command line, but it is encouraged to remove it.

That's from using jetty-alpn-agent 2.0.10.

raboof:
> I tend to say that we should drop jetty-alpn support for 10.2.0 and just require a recent JDK 8 if you want to use HTTP/2.

That makes sense to me :+1: 

> This requirement will transitively propagate to akka-grpc as well so that should be taken into account as well.

Yes. The smoothest transition might be to stick with Akka HTTP 10.1.0 for Akka gRPC 1.0.0, and move to Akka HTTP 10.2.0 for Akka gRPC 1.1.0 - perhaps already recommending JDK >= 8u251 for 1.0.0 but still supporting jetty ALPN

jrudolph:
> Yes. The smoothest transition might be to stick with Akka HTTP 10.1.0 for Akka gRPC 1.0.0, and move to Akka HTTP 10.2.0 for Akka gRPC 1.1.0 - perhaps already recommending JDK >= 8u251 for 1.0.0 but still supporting jetty ALPN

Sounds good. In any case, we should make sure akka-grpc works well with 10.2.0.

raboof:
> In any case, we should make sure akka-grpc works well with 10.2.0.

Agreed! We tested the milestone at https://github.com/akka/akka-grpc/pull/937 and when we get closer to the release we can test with some snapshots.

jrudolph:
I rebased this one based on the release-10.1 version in #3125 and then added a commit which completely removes support for jetty-alpn-agent.

jrudolph:
Failing because the jenkins nodes needs a JDK update ;)

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2AlpnSupport.scala
```diff
@@ -32,48 +29,31 @@ private[http] object Http2AlpnSupport {
...
   def enableForServer(engine: SSLEngine, setChosenProtocol: String => Unit): SSLEngine =
     if (isAlpnSupportedByJDK) jdkAlpnSupport(engine, setChosenProtocol)
-    else jettyAlpnSupport(engine, setChosenProtocol)
+    else throw new RuntimeException(s"Need to run on a JVM >= 8u252 for ALPN support needed for HTTP/2. Running on ${sys.props("java.version")}")
 
   def isAlpnSupportedByJDK: Boolean =
     // ALPN is supported starting with JDK 9
-    JavaVersion.majorVersion >= 9
+    JavaVersion.majorVersion >= 9 ||
+      (classOf[SSLEngine].getMethods.exists(_.getName == "setHandshakeApplicationProtocolSelector")
```
jrudolph, line 37: This will probably not help anymore because class loading will now fail on older JDKs. Should we keep the reflective access below to avoid that? Or create another class, so we can have a reflective check before the actual one?

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -305,9 +305,7 @@ class Http2ServerSpec extends AkkaSpecWithMaterializer("""
 
       // Reproducing https://github.com/akka/akka-http/issues/2957
       "close the stream when we receive a RST after we have half-closed ourselves as well" in new WaitingForRequestData {
-        import akka.http.scaladsl.client.RequestBuilding._
         // Client sends the request, but doesn't close the stream yet. This is a bit weird, but it's whet grpcurl does ;)
-        val post = Post("/grpc.reflection.v1alpha.ServerReflection/ServerReflectionInfo")
```
jrudolph, line 310: Some unrelated cleanup. Can you remember, @raboof if that's just a left-over?

jrudolph:
PLS BUILD

jrudolph:
Updated JVM on the node.

jrudolph:
Update didn't really work since the adoptopenjdk PPA didn't have a newer version yet. Trying another way.

jrudolph:
PLS BUILD

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2AlpnSupport.scala
ennru: Just leaving reflection seems enough to me.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
raboof: Looks like a leftover from a previous version of that test, yes :+1:

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2AlpnSupport.scala
raboof: Referring to missing methods does not prevent class loading, only referring to unknown types does. For a missing method the class loads fine, but a `java.lang.NoSuchMethodError` will be thrown when we try to invoke the method.

So this seems :+1:

docs/src/main/paradox/server-side/http2.md
```diff
@@ -17,6 +17,10 @@ To use Akka HTTP2 Support, add the module to your project:
   version="$project.version$"
 }
 
+HTTP/2 needs support in TLS for [Application-Layer Protocol Negotiation (ALPN)](https://en.wikipedia.org/wiki/Application-Layer_Protocol_Negotiation)
+to negotiate whether both client and server support HTTP/2. The JVM provides ALPN support starting from JDK 8u252.
+Make sure you run a JVM greater than that.
```
raboof, line 22: At least that version :)

docs/src/main/paradox/migration-guide/migration-guide-10.2.x.md
```diff
@@ -12,14 +12,18 @@ If you find an unexpected incompatibility please let us know, so we can check wh
 
 ## Akka HTTP 10.1.11 -> 10.2.0
 
+### Jetty ALPN agent not supported any more for HTTP/2
```
ennru, line 15:
```suggestion
### HTTP/2 support requires JDK 8 update 252 or later
```

docs/src/main/paradox/release-notes/10.1.x.md
```diff
@@ -1,5 +1,13 @@
 # 10.1.x Release Notes
 
+## 10.1.12
+
+### ALPN support in JDK >= 8u252
```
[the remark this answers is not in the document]
raboof: (for 10.1.12 HTTP/2 support doesn't require update 252, as #3125 doesn't yet remove the jetty alpn support)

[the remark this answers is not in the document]
ennru: Thank you. I got that when I re-read and looked at #3125 again.
So I left it as suggested.

PR #3124, opened 2020-04-30
core: convert ClientServerSpec to use AkkaSpecWithMaterializer

jrudolph:


akka-http-core/src/test/scala/akka/http/scaladsl/ClientServerSpec.scala
```diff
@@ -35,30 +35,21 @@ import akka.util.ByteString
...
 import org.scalatest.concurrent.Eventually.eventually
-import org.scalatest.matchers.should.Matchers
-import org.scalatest.wordspec.AnyWordSpec
-
-class ClientServerSpec extends AnyWordSpec with Matchers with BeforeAndAfterAll with ScalaFutures with WithLogCapturing {
-  val testConf: Config = ConfigFactory.parseString("""
-    akka.loglevel = DEBUG
-    akka.loggers = ["akka.http.impl.util.SilenceAllTestEventListener"]
-    akka.stdout-loglevel = ERROR
-    windows-connection-abort-workaround-enabled = auto
```
jrudolph, line 49: This setting cannot have worked that way because it's in `akka.io.tcp.`.

PR #3125, opened 2020-04-30
[release 10.1] http2: support ALPN natively on JDK >= 8u252

jrudolph:
Crash if incompatible versions of JDK and jetty-alpn-agent are detected

10.1 version of #3117:
 * make it work on JDK 8 >= 8u252
 * fail if incompatible versions are detected
 * add a note to release notes

### DIFF
--- a/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2AlpnSupport.scala
+++ b/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2AlpnSupport.scala
@@ -16,6 +16,7 @@ import org.eclipse.jetty.alpn.ALPN
 import org.eclipse.jetty.alpn.ALPN.ServerProvider
 
 import scala.language.reflectiveCalls
+import scala.util.Try
 
 /**
  * INTERNAL API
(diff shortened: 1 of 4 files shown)


akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2AlpnSupport.scala
```diff
@@ -36,7 +37,16 @@ private[http] object Http2AlpnSupport {
...
   def isAlpnSupportedByJDK: Boolean =
     // ALPN is supported starting with JDK 9
-    JavaVersion.majorVersion >= 9
+    JavaVersion.majorVersion >= 9 ||
+      (classOf[SSLEngine].getMethods.exists(_.getName == "setHandshakeApplicationProtocolSelector")
+        && {
+          // This method only exists in the jetty-alpn provided implementation. If it exists an old version of the jetty-alpn-agent is active which is not supported
+          // on JDK>= 8u252. When running on such a JVM, you can either just remove the agent or (if you want to support older JVMs with the same command line),
+          // use jetty-alpn-agent >= 2.0.10
+          val jettyAlpnClassesAvailable = Try(Class.forName("sun.security.ssl.ALPNExtension")).toOption.exists(_.getDeclaredMethods.exists(_.getName == "init"))
```
raboof, line 46: Tested and works.

(I got into a situation where with JDK 252 the correct agent was loaded, but an older agent was also on the classpath, which confusingly but correctly triggered this error...)

PR #3126, opened 2020-04-30
build: build with Scala 2.13 by default

jrudolph:


ennru:
Do you think we should try to reach warning-free with Scala 2.13? ...and still work with Scala 2.12

raboof:
(those fatal warnings are fixed in #3127 )

jrudolph:
PLS BUILD

ennru:
Works nicely after rebase.

PR #3127, opened 2020-04-30
Fix more 2.13 warnings

raboof:
Moved 'full application' examples to their own files

refs #3127


build.sbt
```diff
@@ -334,9 +334,11 @@ lazy val docs = project("docs")
       // We use this for `complete`
       "-Xlint:-adapted-args",
       // TODO avoid this deprecation
-      "-P:silencer:globalFilters=Adaptation of argument list:adaptation of an empty argument list",
+      "-P:silencer:globalFilters=Adaptation of argument list;adaptation of an empty argument list",
       // Does not appear to lead to problems
       "-P:silencer:globalFilters=The outer reference in this type test cannot be checked at run time",
+      // Can be removed when we drop support for Scala 2.12
+      "-P:silencer:globalFilters=object JavaConverters in package collection is deprecated"
```
ennru, line 341: Isn't this solved with https://github.com/scala/scala-library-compat ?

raboof: We don't include that library because (at least at the time of writing) its binary compatibility story was unclear, so instead we just copied over the bits we need. Not sure if it's worth looking into that just to get rid of this warning...

Issue #3129, opened 2020-05-04
HeaderDirectivesExamplesSpec: compilation on 2.13


chbatey:
```
 [Times: user=0.42 sys=0.00, real=0.06 secs] 
[error] /home/jenkins/workspace/workspace/akka-http-nightly/AKKA_VERSION/default/Node/akka-http-nightly-node/SCALA_VERSION/2.13.1/jdk/AdoptOpenJDK 8/docs/src/test/scala/docs/http/scaladsl/server/directives/HeaderDirectivesExamplesSpec.scala:187:32: adaptation of an empty argument list by inserting () is deprecated: this is unlikely to be what you want
[error]         signature: HeaderDirectives.headerValueByType[T](magnet: akka.http.scaladsl.server.directives.HeaderMagnet[T]): akka.http.scaladsl.server.Directive1[T]
[error]   given arguments: <none>
[error]  after adaptation: HeaderDirectives.headerValueByType((): Unit)
[error]       headerValueByType[Origin]() { origin =>
[error]                                ^
[error] /home/jenkins/workspace/workspace/akka-http-nightly/AKKA_VERSION/default/Node/akka-http-nightly-node/SCALA_VERSION/2.13.1/jdk/AdoptOpenJDK 8/docs/src/test/scala/docs/http/scaladsl/server/directives/HeaderDirectivesExamplesSpec.scala:208:40: adaptation of an empty argument list by inserting () is deprecated: this is unlikely to be what you want
[error]         signature: HeaderDirectives.optionalHeaderValueByType[T <: akka.http.scaladsl.model.HttpHeader](magnet: akka.http.scaladsl.server.directives.HeaderMagnet[T]): akka.http.scaladsl.server.Directive1[Option[T]]
[error]   given arguments: <none>
[error]  after adaptation: HeaderDirectives.optionalHeaderValueByType((): Unit)
[error]       optionalHeaderValueByType[Origin]() {
[error]                                        ^
````

https://jenkins.akka.io:8498/job/akka-http-nightly/115/AKKA_VERSION=default,Node=akka-http-nightly-node,SCALA_VERSION=2.13.1,jdk=AdoptOpenJDK%208/consoleFull

ennru:
Fixed with #3130

PR #3130, opened 2020-05-04
Fix more Scala 2.13 warnings

ennru:
Fixes compilation error https://github.com/akka/akka-http/issues/3129


ennru:
The adaptation is being worked on in https://github.com/akka/akka-http/issues/2738

ennru:
The adaptation is being worked on in https://github.com/akka/akka-http/issues/2738

Issue #3135, opened 2020-05-07
HTTP 2 travis compilation fails on master


raboof:

raboof:
ah, seems like it's a matter of akka-http using a version of jdk8 that is no longer supported by HTTP 10.2.0 since 94ad70d145831c354d03963c29c14c80d1a86a5a

johanandren:
Also non javadoc on JDK 11: https://jenkins.akka.io:8498/job/akka-http-nightly/AKKA_VERSION=master,Node=akka-http-nightly-node,SCALA_VERSION=2.13.1,jdk=AdoptOpenJDK%2011/120/consoleFull

johanandren:
https://jenkins.akka.io:8498/job/akka-http-docs/425/console

raboof:
> https://jenkins.akka.io:8498/job/akka-http-nightly/AKKA_VERSION=master,Node=akka-http-nightly-node,SCALA_VERSION=2.13.1,jdk=AdoptOpenJDK%2011/120/consoleFull

I suspect this is due to jdk11 needing to be upgraded on `johannes`, since this version of jdk11 does not expose the backported API when running with `-release 8`

> https://jenkins.akka.io:8498/job/akka-http-docs/425/console

This is due to jdk8 needing to be upgraded on the moxie servers. This will make other builds fail (like akka-grpc because of https://github.com/akka/akka-grpc/issues/946), but I don't think that should hold up the upgrade.

jrudolph:
> > [jenkins.akka.io:8498/job/akka-http-nightly/AKKA_VERSION=master,Node=akka-http-nightly-node,SCALA_VERSION=2.13.1,jdk=AdoptOpenJDK%2011/120/consoleFull](https://jenkins.akka.io:8498/job/akka-http-nightly/AKKA_VERSION=master,Node=akka-http-nightly-node,SCALA_VERSION=2.13.1,jdk=AdoptOpenJDK%2011/120/consoleFull)
> 
> I suspect this is due to jdk11 needing to be upgraded on `johannes`, since this version of jdk11 does not expose the backported API when running with `-release 8`

Do you know if there is such an JDK 11 version yet?

jrudolph:
http://hg.openjdk.java.net/jdk-updates/jdk11u/file/d37f3b2d0da9/make/data/symbols/java.base-8.sym.txt doesn't seem to have been changed recently.

jrudolph:
[...]

PR #3136, opened 2020-05-07
Enable mima for akka-http2-support

raboof:
Even if we still call it 'experimental', we should protect ourselves against
accidental binary incompatible changes

Will fail on travis because of #3135


raboof:
Let's merge this together with #3137

raboof:
Needs #3137 (or https://github.com/shyiko/jabba/issues/648) for travis to succeed

jrudolph:
Is this module even supposed to have any public Api? I think it wasn't and that might be another reason we didn't enable compat checking?

raboof:
Hmm, that's a good point ;).

PR #3137, opened 2020-05-07
Use sdkman instead of jabba on travis

### LINKED ISSUE #3135: HTTP 2 travis compilation fails on master

raboof:
Refs #3135

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,6 @@
+version: ~> 1.0
+import: scala/scala-dev:travis/default.yml
+
 language: scala
 
 scala:
@@ -11,9 +14,6 @@ git:
 before_install:
   # make comparing to origin/master work
   - git remote set-branches --add origin master && git fetch
-  # using jabba for custom jdk management
-  - curl -sL https://raw.githubusercontent.com/shyiko/jabba/0.11.1/install.sh | bash && . ~/.jabba/jabba.sh
-  - jabba install adopt@1.8-0
 
 addons:
   apt:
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -208,7 +208,6 @@ lazy val http2Support = project("akka-http2-support")
     )
   }
   .enablePlugins(JavaAgent, BootstrapGenjavadoc)
-  .disablePlugins(MimaPlugin) // experimental module still
```
ennru, line 211: A bit unexpected to have this in the same PR.

.travis.yml
```diff
@@ -28,11 +28,9 @@ cache:
   directories:
     - $HOME/.ivy2/cache
     - $HOME/.sbt/boot
-    - $HOME/.jabba/jdk
```
ennru, line 31: Where does the JDK get cached with sdkman?
Or does this even import https://github.com/scala/scala-dev/blob/scala-dev/travis/default.yml#L29 ?

build.sbt
raboof: The relationship is that this would trigger #3135, so including it here proved that it fixes that problem.

I agree that now that we've verified that it's probably better to separate them again, rebased.

.travis.yml
raboof: I think it should but I haven't checked - I guess perhaps we could drop this whole section here.

ignasi35: Can't we remove `before_cache` and `cache` and use the values provided by `scala/scala-dev`'s default TravisCI setup?

Side note: IDK how redefinition/overloading works when using YML imports in Travis.

raboof:
Jabba has caught up, so this isn't strictly needed anymore

jrudolph:
Good that we don't have to do that. I'd rather copy a complete concrete script here than rely on any more abstraction layers for the travis build.

PR #3139, opened 2020-05-08
Test: Inspect chunked responses during termination

ennru:
Tests that a chunked HTTP response stream fails when the server terminates. 

The Akka HTTP client reacts as it should with "EntityStreamException: Entity stream truncation. The HTTP parser was receiving an entity when the underlying connection was closed unexpectedly."

https://github.com/akka/akka-http/issues/1210#issuecomment-488136833 made me curious how a client will see server termination on a chunked HTTP response stream.

I expect this test to be flaky, so I'm not sure if it is worth merging.

References #1210

### DIFF
--- a/akka-http-core/src/test/scala/akka/http/scaladsl/GracefulTerminationSpec.scala
+++ b/akka-http-core/src/test/scala/akka/http/scaladsl/GracefulTerminationSpec.scala
@@ -7,6 +7,7 @@ package akka.http.scaladsl
 import java.util.concurrent.atomic.AtomicInteger
 import java.util.concurrent.{ ArrayBlockingQueue, TimeUnit }
 
+import akka.Done
 import akka.actor.ActorSystem
 import akka.http.impl.util._
 import akka.http.scaladsl.model.HttpEntity._
@@ -15,6 +16,8 @@ import akka.http.scaladsl.model.headers.Connection
 import akka.http.scaladsl.settings.ClientConnectionSettings
 import akka.http.scaladsl.settings.{ ConnectionPoolSettings, ServerSettings }
 import akka.stream.scaladsl._
+import akka.stream.testkit.TestSubscriber.{ OnError, OnNext }
+import akka.stream.testkit.scaladsl.TestSink
 import akka.stream.{ Server => _, _ }
 import akka.testkit._
 import akka.util.ByteString
@@ -41,6 +44,20 @@ class GracefulTerminationSpec
 
   implicit override val patience = PatienceConfig(5.seconds.dilated(system), 200.millis)
 
+  "Unbinding" should {
+    "not allow new connections" in new TestSetup {
+      Await.result(serverBinding.unbind(), 1.second) should ===(Done)
+
+      // immediately trying a new connection should cause `Connection refused` since we unbind immediately:
+      val r = makeRequest(ensureNewConnection = true)
+      val ex = intercept[StreamTcpException] {
+        Await.result(r, 2.seconds)
+      }
+      ex.getMessage should include("Connection refused")
+      serverBinding.terminate(hardDeadline = 2.seconds)
+    }
+  }
+
   "Graceful termination" should {
 
     "stop accepting new connections" in new TestSetup {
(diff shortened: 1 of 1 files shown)


jrudolph:
We are missing such a test indeed, mainly because we don't seem to verify yet that streams that are still ongoing are really aborted in time. What we should strive for is that the connection is aborted, otherwise, responses that use `HttpEntity.CloseDelimited` will see a truncated body without an error.

The test case seems a bit fragile since it uses the same ActorSystem for server and client. We should change that as well, because we might want to add `Http().shutdownConnectionPools` to the coordinated shutdown as well and that might interfere with the testing logic.

jrudolph:
Unrelated failure that will be fixed with https://github.com/akka/akka-http/pull/3145.

jrudolph:
PLS BUILD

akka-http-core/src/test/scala/akka/http/scaladsl/GracefulTerminationSpec.scala
```diff
@@ -60,6 +77,53 @@ class GracefulTerminationSpec
...
+      // start reading the response
+      val response = r1.futureValue.entity.dataBytes
+        .via(Framing.delimiter(ByteString(","), 20))
+        .runWith(TestSink.probe[ByteString])
+      response.requestNext().utf8String should ===("reply1")
+
+      try {
+        val termination = serverBinding.terminate(hardDeadline = 50.millis)
+        response.request(20)
+        // local testing shows the stream fails long after the 50 ms deadline
```
raboof, line 100: So this test shows the deadline isn't actually enforced for streams like this?

akka-http-core/src/test/scala/akka/http/scaladsl/GracefulTerminationSpec.scala
```diff
@@ -60,6 +77,53 @@ class GracefulTerminationSpec
...
+        response.expectNext().utf8String should ===("reply2")
+        response.expectNext().utf8String should ===("reply3")
+        response.expectNext().utf8String should ===("reply4")
+        response.expectNext().utf8String should ===("reply5")
```
raboof, line 104: We could make this non-flakey by asserting that `eventually` `response.expectEvent()` should return an `OnError`?

akka-http-core/src/test/scala/akka/http/scaladsl/GracefulTerminationSpec.scala
raboof: (I guess that might not be a big problem since the actorsystem shutdown will shut down the materializer and fail the stream anyway?)

ennru:
> What we should strive for is that the connection is aborted, otherwise, responses that use HttpEntity.CloseDelimited will see a truncated body without an error.

Yes, I added a test for that and the client doesn't understand that it's over. The test is ignored for now.

akka-http-core/src/test/scala/akka/http/scaladsl/GracefulTerminationSpec.scala
jrudolph: I'm banging my head against this a bit... What I'm seeing is that the GracefulTerminationStage does something after those 50ms but that does not kill the connection immediately (especially not the response stream). My hypothesis is, since the termination stage does not monitor response entity streams, it does something which will only indirectly stop the connection at some point. In any case, it does not cancel connection but complete it, which will still keep proper semantics in this case (because chunked encoding can detect truncation) but not in the below one.

akka-http-core/src/test/scala/akka/http/scaladsl/GracefulTerminationSpec.scala
jrudolph: I agree, maybe we should remove all the extra replies and just keep the note that it takes longer than expected for those streams to fail?

(I've seen the current version fail with more log output, so it is likely still flaky).

jrudolph:
> I debugged the remaining issue a bit and will create a ticket to follow up.

#3209

akka-http-core/src/test/scala/akka/http/scaladsl/GracefulTerminationSpec.scala
```diff
@@ -60,6 +77,73 @@ class GracefulTerminationSpec
...
+        .via(Framing.delimiter(ByteString(","), 20))
+        .runWith(TestSink.probe[ByteString])
+      response.requestNext().utf8String should ===("reply1")
+
+      try {
+        val termination = serverBinding.terminate(hardDeadline = 50.millis)
+        response.request(20)
+        // local testing shows the stream fails long after the 50 ms deadline
+        response.expectNext().utf8String should ===("reply2")
+        response.expectNext().utf8String should ===("reply3")
```
raboof, line 102: Should we remove a couple of these as well?

ennru: Done.

PR #3151, opened 2020-05-12
core: use SystemMaterializer in AkkaSpecWithMaterializer

jrudolph:
We had to revert this before in #3145 because ClientServerSpecs started to fail in relation
with another change. The fix was to make sure to shut down all
connection pools in the end of tests that used Http().singleRequest.

### DIFF
--- a/akka-http-core/src/test/scala/akka/http/scaladsl/ClientServerSpec.scala
+++ b/akka-http-core/src/test/scala/akka/http/scaladsl/ClientServerSpec.scala
@@ -581,6 +581,7 @@ class ClientServerSpec extends AkkaSpecWithMaterializer(
       try {
         (1 to 10).foreach(runOnce)
       } finally server.foreach(_.unbind())
+      Http().shutdownAllConnectionPools().futureValue
     }
 
     "complete a request/response when the request side immediately closes the connection after sending the request" in Utils.assertAllStagesStopped {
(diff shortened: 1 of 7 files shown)


akka-http-core/src/test/scala/akka/http/scaladsl/ClientServerSpec.scala
```diff
@@ -776,6 +778,7 @@ Host: example.com
       }
 
       Await.result(binding.unbind(), 10.seconds)
```
ennru, line 780: Any particular reason to keep this `Await` and change others to `.futureValue`?

raboof: Not sure in this case, but generally: sadly `Await.result` is usually faster than `.futureValue` because `.futureValue` always polls. That might not always matter much, but is the reason why we still use `Await.result` in a lot of places even though `.futureValue` is a nicer API. https://github.com/scalatest/scalatest/pull/1794 avoids the polling for `.futureValue`.

jrudolph: Good point, but let's keep it like this. Maybe we'll have to revive `awaitResult` with a more concise syntax.

PR #3157, opened 2020-05-13
Simplify pool access

jrudolph:
The PoolGateway was a bit weird mixture of an identifier together with methods that gave access to the pool master (`gatewayRef`). This obstructed the view that every access to any client pool goes through the master.

I removed those methods and created a new interface to the pool master that is more explicit about what is happening.


jrudolph:
This is the important commit: https://github.com/akka/akka-http/pull/3157/commits/8a06f4608b84c58d741f809e1f5f383c29705661

jrudolph:
This is ready to be reviewed. Has a few parts:

 * rename `PoolGateway` to `PoolId`
 * remove action methods from `PoolId` and move them to new PoolMaster for clarity about what is doing what
 * adapt public `HostConnectionPool` API

akka-http-compatibility-tests/src/main/scala/example/HostConnectionPoolCompat.scala
```diff
@@ -0,0 +1,35 @@
...
+
+    require(hcp.productArity == 1)
+    require(hcp.productElement(0) == hcp.setup)
+    require(hcp.canEqual(hcp))
+    require(hcp.equals(hcp))
+
+    // This one didn't compile even before (private[http] constructor)
+    // new HostConnectionPool(hcp.setup)(null)
+
+    // This line compiles but only by giving null as the PoolGateway parameter which was private before
```
jrudolph, line 27: I tried to come up with examples that would crash and these do but I think they are not worthwhile to be supported.

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolId.scala
```diff
@@ -0,0 +1,44 @@
...
+import akka.http.impl.settings.HostConnectionPoolSetup
+
+/**
+ * Represents a pool by its target endpoint and settings and its designated usage (shared or unique).
+ *
+ * A [[PoolId]] is represented by its [[HostConnectionPoolSetup]] and its [[PoolUsage]]. If the later
+ * is [[SharedPool]], it means that a shared pool must be used for this particular [[HostConnectionPoolSetup]].
+ */
+private[http] final class PoolId(val hcps: HostConnectionPoolSetup, val usage: PoolUsage) {
+  override def toString = s"PoolGateway(hcps = $hcps)"
```
ennru, line 19:
```suggestion
  override def toString = s"PoolId(hcps = $hcps, usage = $usage)"
```

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
```diff
@@ -113,39 +163,39 @@ private[http] final class PoolMasterActor extends Actor with ActorLogging {
       def track(remaining: Iterator[Future[Done]]): Unit =
         if (remaining.hasNext) remaining.next().onComplete(_ => track(remaining))
         else shutdownCompletedPromise.trySuccess(Done)
-      track(poolStatus.keys.map(_.shutdown()).iterator)
+      track(statusById.keys.map(id => new PoolMaster(self).shutdown(id)).iterator) // FIXME
```
ennru, line 166: FIXME?

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
```diff
@@ -697,7 +700,7 @@ class HttpExt private[http] (private val config: Config)(implicit val system: Ex
    */
   def shutdownAllConnectionPools(): Future[Unit] = {
     val shutdownCompletedPromise = Promise[Done]()
-    poolMasterActorRef ! ShutdownAll(shutdownCompletedPromise)
+    poolMaster.ref ! ShutdownAll(shutdownCompletedPromise)
     shutdownCompletedPromise.future.map(_ => ())(system.dispatcher)
```
ennru, line 704: Should this move into `PoolMaster` now?

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
```diff
@@ -791,7 +792,7 @@ class HttpExt private[http] (private val config: Config)(implicit val system: Ex
   @InternalApi
   private[scaladsl] def poolSize: Future[Int] = {
     val sizePromise = Promise[Int]()
-    poolMasterActorRef ! PoolSize(sizePromise)
+    poolMaster.ref ! PoolSize(sizePromise)
     sizePromise.future
```
ennru, line 796: Move to `PoolMaster`?

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
jrudolph: Thanks, done.

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
jrudolph: Done

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolInterface.scala
```diff
@@ -236,19 +236,19 @@ private[http] object PoolInterface {
...
    */
-  private[http] val GatewayLogSource: LogSource[PoolGateway] =
-    new LogSource[PoolGateway] {
-      def genString(gateway: PoolGateway): String = {
-        val scheme = if (gateway.hcps.setup.connectionContext.isSecure) "https" else "http"
-        s"Pool(${gateway.gatewayId.name}->$scheme://${gateway.hcps.host}:${gateway.hcps.port})"
+  private[http] val PoolLogSource: LogSource[PoolId] =
+    new LogSource[PoolId] {
+      def genString(poolId: PoolId): String = {
+        val scheme = if (poolId.hcps.setup.connectionContext.isSecure) "https" else "http"
```
raboof, line 247: It feels a bit weird to 'reach into' the ID like that. Might be neater to have a method on `PoolId` that constructs this string? OTOH this keeps all logging-related logic here, might be fine.

jrudolph: I agree, it's a bit weird (but unchanged from before). If there would be better ways of showing that kind of info with default logging configuration (like putting it into the MDC), that would be preferable.

jrudolph:
squashed and rebased

PR #3160, opened 2020-05-14
core: decrease idle timeout in ClientServerSpec

### LINKED ISSUE #3159: Failed: Multiple ClientServerSpec tests fail in JDK 11
https://jenkins.akka.io:8498/job/akka-http-nightly/126/

jrudolph:
Fixes #3159.

Otherwise, the test will fail with assertAllStagesStopped in JDK 11.
In JDK 11 the SSLEngine seems to show a different closing behavior which
leads to the TLS stage being kept open until the idle timeout closes
the connection.


jrudolph:
>  Alternatively, we could increase the `akka.stream.testkit.all-stages-stopped-timeout`.

Basically, this is only a workaround around a bug in the JDK 11 TLS stage behavior so let's keep it as simple as possible.

ennru:
Merged to get the fix in, I'll repoen #3159 to evolve a "real" issue from it.

Issue #3155, opened 2020-05-13
Deprecate `HttpApp`


ennru:
The `HttpApp` which is marked `@ApiMayChange` would need to be adapted for use with the new actor APIs and Coordinated shutdown.

Instead of working on that, the examples for a simple server application should be improved and `HttpApp` deprecated.

See #960, #1741, #1300, #1460

jlprat:
Sad to see `HttpApp` go

ennru:
I agree it is nice for simple things, but it doesn't scale well.

ennru:
I guess one answer could be to link to https://developer.lightbend.com/guides/akka-http-quickstart-scala/ and https://developer.lightbend.com/guides/akka-http-quickstart-java/

jrudolph:
> Sad to see `HttpApp` go

Quite understable! Just to give a bit more of a background on that decision: HttpApp works best if you have an application that centered around akka-http routes (which is a perfectly valid use case, if you ask me). As soon, as you use other pieces of akka, you need some bootstrapping code anyway. In this case, `HttpApp` can get in the way because it tries to "manage" the ActorSystem for you. This is even more pronounced with the new akka typed APIs that recommend a bit different way of structuring/wrapping HTTP boot code, where the subclass-based style of HttpApp is not quite fitting.

The goal should be that bootstrapping an akka-http server alone is almost as simple as using HttpApp but does not hide creating the ActorSystem, so it is easy to expand that initial bootstrap code into an Akka application that uses more of Akka's features. That also goes the other way, where you start with an Akka application and add akka-http routes. Adding the server shouldn't be more than adding one extra line to your existing bootstrap code.

The feature we lose with HttpApp, is shutting down the ActorSystem by default when pressing ENTER. It seems that feature is mostly useful during testing. For that, I'd recommend to use sbt-revolver instead.

ennru:
Fixed with #3162.

PR #3162, opened 2020-05-15
Deprecate HttpApp

ennru:
References #3155


docs/src/main/paradox/routing-dsl/index.md
```diff
@@ -57,8 +56,6 @@ Scala
 Java
```
jrudolph, line 56: Did you keep the lines above intentionally?

docs/src/test/scala/docs/http/scaladsl/HttpAppExampleSpec.scala
```diff
@@ -4,10 +4,12 @@
 
 package docs.http.scaladsl
 
+import com.github.ghik.silencer.silent
 import docs.CompileOnlySpec
 import org.scalatest.matchers.should.Matchers
 import org.scalatest.wordspec.AnyWordSpec
 
+@silent("deprecated")
```
jrudolph, line 12: Is there a particular reason to keep the docs examples around?

ennru: Dropped them now.

docs/src/main/paradox/routing-dsl/index.md
ennru: How could I not see this? Removed.

PR #3166, opened 2020-05-18
HTTP/2 client implementation stub

jrudolph:
Refs #1883 

Doing the minimum to make the simplest test pass. Interestingly enough, only request rendering and response parsing had to be implemented anew. Everything else seems to mostly "just work", even the demux (which is somewhat surprising).

Things that are needed

 * Http2ServerDemux and Http2ClientDemux are almost identical which looks as if they could be the same
 * Parsing and rendering need more complete implementations
 * Backpressure handling in demux needs to be reviewed to make sure that peer-imposed limits work (most importantly, max_concurrent_streams).
 * Some initial handshake related things (settings exchange) might be different between server and client and need to be adapted
 * Settings classes need to be figured out
 * API needs to be reviewed
 * Proper TLS setup/testing
 * More tests for more scenarios
 * Ultimately, integration into the pool API would be nice (automatic negotiation of h2 vs http1, maybe if we know that an endpoint speaks h2 (which could be probed) the flow implemented here could be a drop-in replacement for the PoolInterface? That would be great)

Maybe we should strive to make that effort here as minimal and as least intrusive as possible, so we can do incremental improvements and try it out in akka-grpc as soon as possible.

### DIFF
--- a/akka-http-core/src/main/resources/reference.conf
+++ b/akka-http-core/src/main/resources/reference.conf
@@ -369,6 +369,60 @@ akka.http {
     # Int   : determines how many bytes should be logged per data chunk
     log-unencrypted-network-bytes = off
 
+    // FIXME: unify with server-side part (by importing or similar to parsing)
+    http2 {
+      # The maximum number of request per connection concurrently dispatched to the request handler.
+      max-concurrent-streams = 256
+
+      # The maximum number of bytes to receive from a request entity in a single chunk.
+      #
+      # The reasoning to limit that amount (instead of delivering all buffered data for a stream) is that
+      # the amount of data in the internal buffers will drive backpressure and flow control on the HTTP/2 level. Bigger
+      # chunks would mean that the user-level entity reader will have to buffer all that data if it cannot read it in one
+      # go. The implementation would not be able to backpressure further data in that case because it does not know about
+      # this user-level buffer.
+      request-entity-chunk-size = 65536 b
+
+      # The number of request data bytes the HTTP/2 implementation is allowed to buffer internally per connection. Free
+      # space in this buffer is communicated to the peer using HTTP/2 flow-control messages to backpressure data if it
+      # isn't read fast enough.
+      #
+      # When there is no backpressure, this amount will limit the amount of in-flight data. It might need to be increased
+      # for high bandwidth-delay-product connections.
+      #
+      # There is a relation between the `incoming-connection-level-buffer-size` and the `incoming-stream-level-buffer-size`:
+      # If incoming-connection-level-buffer-size < incoming-stream-level-buffer-size * number_of_streams, then
+      # head-of-line blocking is possible between different streams on the same connection.
+      incoming-connection-level-buffer-size = 10 MB
+
+      # The number of request data bytes the HTTP/2 implementation is allowed to buffer internally per stream. Free space
+      # in this buffer is communicated to the peer using HTTP/2 flow-control messages to backpressure data if it isn't
+      # read fast enough.
+      #
+      # When there is no backpressure, this amount will limit the amount of in-flight data per stream. It might need to
+      # be increased for high bandwidth-delay-product connections.
+      incoming-stream-level-buffer-size = 512kB
+
+      # The maximum number of outgoing control frames to buffer when the peer does not read from its TCP connection before
+      # backpressuring incoming frames.
+      #
+      # On a healthy HTTP/2 connection this setting should have little effect because control frames are given priority over
+      # data frames and should not be buffered for a long time.
+      #
+      # The limit is necessary to prevent a malicious peer to solicit buffering of outgoing control frames (e.g. by sending PINGs)
+      # without ever reading frames ultimately leading to an out of memory situation. With the limit in place, the implementation
+      # stops reading incoming frames when the number of outgoing control frames has reached the given amount. This way an attacker
+      # isn't able to communicate any further without first freeing space in the TCP window, draining the buffered control frames.
+      #
+      # See CVE-2019-9512 for an example of such an attack.
+      #
+      # Note that only control frames are affected because data frames, in contrast, are covered by the HTTP/2 flow control.
+      outgoing-control-frame-buffer-size = 1024
+
+      # Enable verbose debug logging for all ingoing and outgoing frames
+      log-frames = false
+    }
+
     websocket {
       # periodic keep alive may be implemented using by sending Ping frames
       # upon which the other side is expected to reply with a Pong frame,
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/settings/Http2ServerSettings.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/settings/Http2ServerSettings.scala
@@ -11,6 +11,23 @@ import akka.http.impl.util._
 import com.typesafe.config.Config
 
 /**
+ * INTERNAL API
+ *
+ * Settings which are common for server and client side.
+ */
+@InternalApi
+@DoNotInherit
+private[http] trait Http2CommonSettings {
+  def requestEntityChunkSize: Int
+  def incomingConnectionLevelBufferSize: Int
+  def incomingStreamLevelBufferSize: Int
+
+  def logFrames: Boolean
+  def maxConcurrentStreams: Int
+  def outgoingControlFrameBufferSize: Int
+}
+
+/**
  * Placeholder for any kind of internal settings that might be interesting for HTTP/2 (like custom strategies)
  */
 @InternalApi
@@ -19,7 +36,7 @@ private[http] trait Http2InternalServerSettings
 
 @ApiMayChange
 @DoNotInherit
-trait Http2ServerSettings extends javadsl.settings.Http2ServerSettings { self: Http2ServerSettings.Http2ServerSettingsImpl =>
+trait Http2ServerSettings extends javadsl.settings.Http2ServerSettings with Http2CommonSettings { self: Http2ServerSettings.Http2ServerSettingsImpl =>
   def requestEntityChunkSize: Int
   def withRequestEntityChunkSize(newValue: Int): Http2ServerSettings = copy(requestEntityChunkSize = newValue)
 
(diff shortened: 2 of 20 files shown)


akka-http2-support/src/main/scala/akka/http/scaladsl/model/http2/RequestResponseAssociation.scala
```diff
@@ -0,0 +1,11 @@
+/*
+ * Copyright (C) 2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.http.scaladsl.model.http2
+
+/**
+ * A marker trait for attribute values that should be (automatically) carried over from request to response.
+ */
+trait RequestResponseAssociation
```
jrudolph, line 10: We might want to move to akka-http-core and make use of it there. We could use it also for the regular pools to associate requests with responses (currently, we create a `RequestContext` there to carry over a Promise from request to response which we could get rid of by using an (internal) attribute instead).

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientServerSpec.scala
```diff
@@ -0,0 +1,78 @@
...
+
+class Http2ClientServerSpec extends AkkaSpecWithMaterializer(
+  """akka.http.server.remote-address-header = on
+     akka.http.server.http2.log-frames = on
+     akka.http.server.log-unencrypted-network-bytes = 100
+     akka.http.server.preview.enable-http2 = on
+     akka.actor.serialize-messages = false
+  """) with ScalaFutures {
+  case class RequestId(id: String) extends RequestResponseAssociation
+  val requestIdAttr = AttributeKey[RequestId]("requestId")
```
jrudolph, line 27: Here's how you use it: you create a token type and extend from `RequestResponseAssociation`. It can be any type.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientServerSpec.scala
```diff
@@ -0,0 +1,78 @@
...
+
+  "HTTP 2 implementation" should {
+    "support simple round-trips" in new TestSetup {
+      sendClientRequest(
+        HttpRequest(method = HttpMethods.POST, entity = "ping")
+          .addAttribute(requestIdAttr, RequestId("request-1"))
```
jrudolph, line 33: Then add it to the request...

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientServerSpec.scala
```diff
@@ -0,0 +1,78 @@
...
+
+      val serverRequest = expectServerRequest()
+      serverRequest.request.header[Http2StreamIdHeader] shouldBe Symbol("nonEmpty")
+      serverRequest.request.method shouldBe HttpMethods.POST
+      Unmarshal(serverRequest.request.entity).to[String].futureValue shouldBe "ping"
+      serverRequest.sendResponse(HttpResponse(entity = "pong"))
+
+      val response = expectClientResponse()
+      Unmarshal(response.entity).to[String].futureValue shouldBe "pong"
+      response.attribute(requestIdAttr).get.id shouldBe "request-1"
```
jrudolph, line 44: ... and the response will contain it as well.

jrudolph:
> Added a critical feature I forgot in the original comment: since request/response streams are multiplexed, we need a way to associate requests and responses. In the current solution, I use specially marked attributes which will be carried over from request to responses.

The same feature could be used on the server-side if we wanted to support a streaming interface there (`Http().bindAndHandle`). In this case, the user handler would have to make sure to carry over the attributes but we could provide helpers for that.

jrudolph:
Now ready to be reviewed.

This is now a working HTTP/2 client implementation. It's not complete or completely tested.

The focus here is on giving an overview and a first potential API of how many (or how few) changes would be necessary to build an HTTP/2 client implementation:

 * new: RequestRendering / ResponseParsing
 * new: state in the multiplexer for tracking when a request is being sent out
 * some settings infrastructure to reuse server settings (will have to be refined for GA)
 * some mostly trivial glue code in Http2Blueprint
 * integrated client-server round-trip testing
 * Only private API so far in `akka.http.scaladsl.Http2`

The roadmap would be:
 * review this PR
 * ticketify all the remaining issues as outlined above
 * merge and release with next 10.2 milestone/RC
 * incrementally improve and complete work as priorities allow
 * (now or later) try out client in akka-grpc (can use preliminary API and later public one)

WDYT? If you prefer I could also break this PR down in necessary ground-work and the actual implementation. In any case I can create more digestible commits for merging.

akka-http-core/src/main/resources/reference.conf
```diff
@@ -369,6 +369,60 @@ akka.http {
     # Int   : determines how many bytes should be logged per data chunk
     log-unencrypted-network-bytes = off
 
+    // FIXME: unify with server-side part (by importing or similar to parsing)
```
jrudolph, line 372: Verbatim copy of server-side settings so far.

akka-http-core/src/main/scala/akka/http/scaladsl/settings/Http2ServerSettings.scala
```diff
@@ -77,3 +94,71 @@ object Http2ServerSettings extends SettingsCompanion[Http2ServerSettings] {
     )
   }
 }
+
+/**
```
jrudolph, line 98: Adapted copy of server-side

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
```diff
@@ -61,22 +82,55 @@ private[http] object Http2Blueprint {
...
+    }
+
   def framing(log: LoggingAdapter): BidiFlow[FrameEvent, ByteString, ByteString, FrameEvent, NotUsed] =
     BidiFlow.fromFlows(
       Flow[FrameEvent].via(new Http2FrameRendering),
       Flow[ByteString].via(new Http2FrameParsing(shouldReadPreface = true, log)))
 
+  def framingClient(log: LoggingAdapter): BidiFlow[FrameEvent, ByteString, ByteString, FrameEvent, NotUsed] =
+    BidiFlow.fromFlows(
+      Flow[FrameEvent].via(new Http2FrameRendering).prepend(Source.single(Http2Protocol.ClientConnectionPreface)),
```
jrudolph, line 131: Reverse as server framing: send preface but don't expect it.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -83,10 +86,10 @@ private[http2] class Http2ServerDemux(http2Settings: Http2ServerSettings, initia
     new GraphStageLogic(shape) with Http2MultiplexerSupport with Http2StreamHandling with GenericOutletSupport with StageLogging {
       logic =>
 
-      override def settings: Http2ServerSettings = http2Settings
+      override def settings: Http2CommonSettings = http2Settings
       override def isUpgraded: Boolean = upgraded
 
-      override protected def logSource: Class[_] = classOf[Http2ServerDemux]
+      override protected def logSource: Class[_] = if (isServer) classOf[Http2ServerDemux] else classOf[Http2ClientDemux]
```
jrudolph, line 92: `isServer` only used here so far to be able to distinguish logs when running server and client at the same time in tests.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -98,28 +107,44 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with StageLoggi
       Closed
     }
 
+    protected def expectIncomingStream(
```
jrudolph, line 110: Moved up here, so that it can be reused in multiple states.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -62,10 +62,15 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with StageLoggi
   def handleStreamEvent(e: StreamFrameEvent): Unit = {
     updateState(e.streamId, _.handle(e))
   }
+
+  def handleOutgoingCreated(stream: Http2SubStream): Unit =
```
jrudolph, line 66: Only clients are allowed to create streams, so we didn't track this event before.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/client/RequestRendering.scala
```diff
@@ -0,0 +1,47 @@
...
+import akka.annotation.InternalApi
+import akka.event.LoggingAdapter
+import akka.http.impl.engine.http2.FrameEvent.ParsedHeadersFrame
+import akka.http.scaladsl.model.http2.RequestResponseAssociation
+import akka.http.scaladsl.model.{ ContentTypes, HttpEntity, HttpRequest }
+
+import scala.collection.immutable.VectorBuilder
+
+@InternalApi
+private[http2] object RequestRendering {
```
jrudolph, line 19: For parsing and rendering, I adopted the style of the server-side components. In the future, more of those components could be shared.

akka-http2-support/src/main/scala/akka/http/scaladsl/Http2.scala
```diff
@@ -180,6 +180,29 @@ final class Http2Ext(private val config: Config)(implicit val system: ActorSyste
     ProtocolSwitch(_ => getChosenProtocol(), http1, http2) join
       tls
   }
+
+  def outgoingConnection(
```
jrudolph, line 184: Already has same signature as the existing API in `Http()`. Due to multiplexing the stream has different semantics than for HTTP/1.1, so it's more similar in usage as the `newHostConnectionPool` API in `Http()`.

For the public APIs we have to consider how to integrate it. (Will create a ticket about it)

jrudolph:
Rebased and squashed

jrudolph:
Created tickets and overview over remaining work in https://github.com/akka/akka-http/issues/3226

PR #3172, opened 2020-05-19
Make scalatest versions consistent

raboof:


project/Dependencies.scala
```diff
@@ -12,23 +12,20 @@ object Dependencies {
...
-  lazy val specs2Version = settingKey[String]("The version of Specs2 to use")
-  lazy val scalaCheckVersion = settingKey[String]("The version of ScalaCheck to use.")
+  val scalaTestVersion = "3.1.2"
+  val specs2Version = "4.9.4"
+  val scalaCheckVersion = "1.14.3"
 
   val Versions = Seq(
     crossScalaVersions := Seq("2.13.1", "2.12.10"),
     scalaVersion := crossScalaVersions.value.head,
-    scalaCheckVersion := System.getProperty("akka.build.scalaCheckVersion", "1.14.3"),
```
jrudolph, line 29: Good to finally get rid of that.

raboof:
> The question would be if we would ever notice to update to a new hotfix version (like + ".1")?

Probably not.... but probably not that important either?

jrudolph:
> > The question would be if we would ever notice to update to a new hotfix version (like + ".1")?
> 
> Probably not.... but probably not that important either?

I agree, if there's a need, we'd probably notice and update ourselves.

PR #3183, opened 2020-05-26
Avoid warning in HTTP/2 code and add more logging

jrudolph:


akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -165,7 +165,8 @@ private[http2] class Http2ServerDemux(http2Settings: Http2ServerSettings, initia
                 multiplexer.pushControlFrame(SettingsAckFrame(settings))
               }
 
-            case PingFrame(true, _) =>
+            case SettingsAckFrame(Nil) => // This should be the ack to the initial settings frame
```
raboof, line 168: So we don't care and just drop the ack? Seems a bit suspicious but :+1:

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -67,10 +67,13 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with StageLoggi
...
   }
   private def updateState(streamId: Int, handle: IncomingStreamState => IncomingStreamState): Unit = {
-    handle(streamFor(streamId)) match {
+    val oldState = streamFor(streamId)
+    val newState = handle(oldState)
+    newState match {
       case Closed   => incomingStreams -= streamId
       case newState => incomingStreams += streamId -> newState
     }
+    log.debug(s"Incoming stream state updated for [$streamId] ${oldState.stateName} -> ${oldState.stateName}")
```
raboof, line 76:
```suggestion
    log.debug(s"Incoming stream state updated for [$streamId] ${oldState.stateName} -> ${newState.stateName}")
```

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
ennru: It was just logged as unhandled before (debug level).

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
jrudolph: Ehem, yes, thanks :)

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
raboof: Yes, in that sense the behavior doesn't really change, but handling it explicitly like this gives the impression that it is 'correct' to just ignore the Ack, which was surprising to me - you'd expect it to be in the protocol for a reason.

jrudolph: Yes, exactly, and it's an answer to the frame we send in `preStart`. Should we start sending proper things in there, we need to revise.

raboof: for now perhaps add a comment here explaining what we should ideally do (even though we don't do it now)?

jrudolph: Good point, added https://github.com/akka/akka-http/issues/3185 and linked to it from the comment.

Issue #3187, opened 2020-05-27
Remove LegacyHostConnectionPool


jrudolph:
It has been in legacy mode quite a while, still fails some tests regularly and all in all hasn't seen much use it seems.

PR #3188, opened 2020-05-27
core: remove legacy host connection pool including settings and tests

jrudolph:
Refs #3187


akka-http-core/src/main/mima-filters/10.1.x.backwards.excludes/3187-remove-legacy-pool.excludes
```diff
@@ -0,0 +1,16 @@
+# Removal of @ApiMayChange methods
+ProblemFilters.exclude[MissingClassProblem]("akka.http.javadsl.settings.PoolImplementation$")
```
jrudolph, line 2: We could have kept (part of) those methods around for a while for better binary compatibility but that might not be necessary (because people would have used config settings in most cases anyway).

akka-http-core/src/main/scala/akka/http/impl/settings/ConnectionPoolSettingsImpl.scala
```diff
@@ -76,10 +72,6 @@ private[akka] object ConnectionPoolSettingsImpl extends SettingsCompanionImpl[Co
       c.getFiniteDuration("max-connection-backoff"),
       c.getPotentiallyInfiniteDuration("idle-timeout"),
       ClientConnectionSettingsImpl.fromSubConfig(root, c.getConfig("client")),
-      c.getString("pool-implementation").toLowerCase match {
```
jrudolph, line 79: We could keep a warning if the setting is still found to be set to `legacy`. We have no logger here, though, so it didn't seem worth the effort.

PR #3189, opened 2020-05-27
Remove deprecated methods

jrudolph:


akka-http-core/src/main/scala/akka/http/scaladsl/ConnectionContext.scala
```diff
@@ -17,7 +17,8 @@ import scala.collection.immutable
 import scala.compat.java8.OptionConverters._
 
 trait ConnectionContext extends akka.http.javadsl.ConnectionContext {
-  final def defaultPort = getDefaultPort
+  @deprecated("Internal method, left for binary compatibility", since = "10.2.0")
+  protected[http] def defaultPort: Int
```
jrudolph, line 21: Couldn't be removed right away because deprecation notice was missing.

jrudolph:
(sorry got bigger than originally planned)

jrudolph:
I'll break it down a bit for easier consumption.

PR #3193, opened 2020-05-28
http: remove deprecated javadsl.server.Directives.route

jrudolph:
Has been replaced by `concat`

This has been deprecated in 10.1.6 which was released in Dec 2018, so it's probably good to go.

### DIFF
--- a/akka-http-tests/src/test/java/akka/http/javadsl/server/directives/RouteDirectivesTest.java
+++ b/akka-http-tests/src/test/java/akka/http/javadsl/server/directives/RouteDirectivesTest.java
@@ -94,9 +94,4 @@ public class RouteDirectivesTest extends JUnitRouteTest {
 	      "a decoder limit (set via `akka.http.routing.decode-max-size`), " +
               "or a custom limit set with `withSizeLimit`.");
   }
-
-  @Test(expected = IllegalArgumentException.class)
-  public void testEmptyRoutesConcatenation() {
-    route();
-  }
 }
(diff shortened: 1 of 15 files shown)


jrudolph:
> `route(..*)` seems quite central, which upgrade guidance will we give? (go to 10.1.latest first and than to 10.2.x, ?)
> I guess a list of removed methods wouldn't help the user.

It has been deprecated since 18 months in 10.1.6 and people should have migrated to use `concat`. Is your concern that this method has been a central part in older versions and people might still be using it a lot?

In any case, it makes sense to add a section to the migration guide, to go to the latest 10.1.x first and fix deprecation warnings if you need some guidance.

raboof:
> It has been deprecated since 18 months in 10.1.6

Hmm. Previously we promised binary compatibility that when a library depends on 10.1.x and uses no deprecated methods, it can be used from 10.2.0. This weakens this promise to only apply for x>=6. Can we do that?

ennru:
I agree, we can't drop it for 10.2. Let's just hide it as much as possible for now.

jrudolph:
I think that promise is overly strict (and also a bit useless since there are lots of other changes since then which might impact compatibility other than "binary"). But let's keep route for now since it still has been shown in docs imports and it used to have quite a central role.

jrudolph:
Replaced by #3207

raboof:
> I think that promise is overly strict

I can live with weakening it, but perhaps we should document what we're weakening it to, then?

PR #3195, opened 2020-05-28
http: remove deprecated fileIODispatcher setting

jrudolph:
Has been deprecated a long time ago in 10.1.6


jrudolph:
As we didn't remove the other post 10.1.0 methods, maybe we shouldn't here as well. So closing for now.

PR #3202, opened 2020-05-29
core: introduce custom hostname resolution client transport

ennru:
Based on the work in #3070 this offers a `ClientTransportWithCustomResolver` to select different means to get from the hostname to an `InetSocketAddress`.

References #92, #3070


ennru:
I wonder if it would make sense to defer resolving the hostname to stream materialization (eg. with `Flow.setup`). For the default `ClientTransport` that is achieved by creating the `InetSocketAddress` as unresolved.

ennru:
Instead of Akka Discovery, it might be relevant to have a pre-baked variant that uses Akka async DNS instead of the JDK's/OS's DNS.

akka-http-core/src/main/scala/akka/http/scaladsl/ClientTransportWithCustomResolver.scala
```diff
@@ -0,0 +1,36 @@
...
+import akka.stream.scaladsl.{ Flow, Tcp }
+import akka.util.ByteString
+
+import scala.concurrent.{ ExecutionContext, Future }
+
+@ApiMayChange
+trait ClientTransportWithCustomResolver extends ClientTransport {
+  override def connectTo(host: String, port: Int, settings: ClientConnectionSettings)(implicit system: ActorSystem): Flow[ByteString, ByteString, Future[Http.OutgoingConnection]] = {
+    implicit val ec = system.dispatcher
+    futureFlow(
```
jrudolph, line 22: Maybe `lazyInitAsync` would work on 2.5 and 2.6? Maybe you could prepend an empty `ByteString()` to that flow, so it will be initialized immediately on materialization.

akka-http-core/src/main/scala/akka/http/scaladsl/ClientTransportWithCustomResolver.scala
```diff
@@ -0,0 +1,43 @@
...
+    Flow.setup { (mat, _) =>
+      implicit val ec: ExecutionContext = mat.executionContext
+      futureFlow {
+        inetSocketAddress(host, port, settings).map { address =>
+          Tcp()(mat.system).outgoingConnection(address, settings.localAddress,
+            settings.socketOptions, halfClose = true, settings.connectingTimeout, settings.idleTimeout)
+            .mapMaterializedValue(_.map(tcpConn => OutgoingConnection(tcpConn.localAddress, tcpConn.remoteAddress)))
+        }
+      }
+    }.mapMaterializedValue(_.flatten.flatten)
```
ennru, line 31: The heart of this PR. But the doubly deferred construction looks like a monster.

akka-http-core/src/main/scala/akka/http/scaladsl/ClientTransportWithCustomResolver.scala
```diff
@@ -0,0 +1,43 @@
...
+import akka.stream.scaladsl.{ Flow, Tcp }
+import akka.util.ByteString
+
+import scala.concurrent.{ ExecutionContext, Future }
+
+@ApiMayChange
+trait ClientTransportWithCustomResolver extends ClientTransport {
+  override def connectTo(host: String, port: Int, settings: ClientConnectionSettings)(implicit unused: ActorSystem): Flow[ByteString, ByteString, Future[Http.OutgoingConnection]] = {
+    // delay hostname resolution until stream materialization
+    Flow.setup { (mat, _) =>
```
jrudolph, line 22: Do we really need the `mat` here? We've got a system which we could use for the materializer or the ec.

akka-http-core/src/main/scala/akka/http/scaladsl/ClientTransportWithCustomResolver.scala
```diff
@@ -0,0 +1,43 @@
...
+import akka.annotation.ApiMayChange
+import akka.http.scaladsl.Http.OutgoingConnection
+import akka.http.scaladsl.settings.ClientConnectionSettings
+import akka.stream.scaladsl.{ Flow, Tcp }
+import akka.util.ByteString
+
+import scala.concurrent.{ ExecutionContext, Future }
+
+@ApiMayChange
+trait ClientTransportWithCustomResolver extends ClientTransport {
```
jrudolph, line 19: Should we make it a class/trait based API? Otherwise, it could be just a

```scala
def withCustomResolver(resolve: (String, Int) => Future[InetSocketAddress]): ClientTransport
```

or create an extra `trait` just for the resolution part if we want to give it a name and more parameters.

akka-http-core/src/main/scala/akka/http/scaladsl/ClientTransportWithCustomResolver.scala
```diff
@@ -0,0 +1,43 @@
...
+        }
+      }
+    }.mapMaterializedValue(_.flatten.flatten)
+  }
+
+  /**
+   * Return an address to which the requests for this connection should be directed.
+   * Called when a new connection is to be established.
+   */
+  protected def inetSocketAddress(host: String, port: Int, settings: ClientConnectionSettings)(implicit ec: ExecutionContext): Future[InetSocketAddress]
```
jrudolph, line 38: I guess it returns an `InetSocketAddress` so it can also override the port?

ennru: Yes, this allows the lookup to map a hostname/port to select the target address and the port.

akka-http-core/src/main/scala/akka/http/scaladsl/ClientTransportWithCustomResolver.scala
ennru: We need the `Flow.setup` so resolving kicks in during materialisation. We can use the actor system parameter instead.

jrudolph:
I added some documentation around this. Is there anything else we should do here?

ennru:
Thank you. This offers a small and flexible API that solves #92, so we're good.

akka-http-core/src/main/scala/akka/http/scaladsl/ClientTransportWithCustomResolver.scala
jrudolph: I added a commit that does that so we don't need the forward port for now (might be slightly less efficient because of an extra indirection in the stream).

akka-http-core/src/main/scala/akka/http/scaladsl/ClientTransport.scala
```diff
@@ -93,6 +98,14 @@ object ClientTransport {
     httpsProxy(InetSocketAddress.createUnresolved(settings.host, settings.port), proxyCredentials)
   }
 
+  /**
+   * Returns a [[ClientTransport]] that allows to customize host name resolution.
+   * @param lookup A function that will be called with hostname and port and that should (potentially asynchronously resolve the given host/port
+   *               to an [[InetSocketAddress]]
+   */
+  def withCustomResolver(lookup: (String, Int) => Future[InetSocketAddress]): ClientTransport =
```
raboof, line 106: Do we also want a javadsl?

docs/src/main/paradox/client-side/client-transport.md
```diff
@@ -104,6 +104,18 @@ Java
 :  @@snip [WebSocketClientExampleTest.java]($test$/java/docs/http/javadsl/WebSocketClientExampleTest.java) { #auth-https-proxy-singleWebSocket-request-example }
 
 
+## Custom Host Name Resolution Transport
+
+You can use `ClientTransport.withCustomResolver` to customize host name resolution. The given resolution function will be called for every connection attempt to resolve
```
raboof, line 109:
```suggestion
You can use @apidoc[ClientTransport.withCustomResolver](ClientTransport) to customize host name resolution. The given resolution function will be called for every connection attempt to resolve
```

docs/src/main/paradox/client-side/client-transport.md
```diff
@@ -104,6 +104,18 @@ Java
...
+
+You can use @apidoc[ClientTransport.withCustomResolver](ClientTransport) to customize host name resolution. The given resolution function will be called for every connection attempt to resolve
+a hostname / port combination (potentially asynchronously) to an `InetSocketAddress`.
+
+As a backend to implement the resolution function you can use Akka's [Async DNS Resolution](https://doc.akka.io/docs/akka/current/io-dns.html#dns-extension).
+
+Potential use cases:
+
+ * in a managed setting this can be used to query for `SRV` DNS records that contain both address and port for a service.
+ * if the DNS server returns multiple addresses, you can implement a load balancing algorithm to select different target address for each connection
```
jrudolph, line 117:
```suggestion
 * if the DNS server returns multiple addresses, you can implement a load balancing algorithm to select a different target address for each connection      
```

docs/src/main/paradox/technologies.md
```diff
@@ -28,6 +28,11 @@ and @ref[Client Websocket Support](client-side/websocket-support.md) for more in
 Akka HTTP provides server-side HTTP/2 support currently in a preview version. See @ref[Server HTTP/2 Support](server-side/http2.md)
 for more information.
 
+## DNS
+
+Akka HTTP provides pluggable client transport which can be used to @ref[customize host name resolution](client-side/client-transport.md#custom-host-name-resolution-transport)
```
jrudolph, line 33:
```suggestion
Akka HTTP provides a pluggable client transport which can be used to @ref[customize host name resolution](client-side/client-transport.md#custom-host-name-resolution-transport)
```

Issue #3179, opened 2020-05-25
Make sure not to break `ParserSettings(system)` usages for now


jrudolph:
Follow up to #3098.

Using code like this:

```scala
ClientConnectionSettings(system)
  .withParsingSettings(
     ParsingSettings(system).withXYZ()
  )  
```

might now fail at runtime when maxContentLength is accessed. `ParsingSettings.apply` is deprecated but we should either make sure it fails directly when passed to `withParsingSettings`, or make sure it still works (by providing the old defaults somehow).

See https://github.com/akka/akka-http/pull/3098#pullrequestreview-399111399

PR #3208, opened 2020-06-02
Fail early when creating client settings with incomplete parser settings

raboof:
refs #3179


raboof:
(travis is green but apparently failed to report back)

raboof:
(manually checked travis, is green)

raboof:
> Overall, it's still a bit of a mess with the remaining usages of ParserSettings.apply.

True, #2999 fixes some, but we can't fix all while keeping compatibility

PR #3210, opened 2020-06-03
Clear some warnings

jrudolph:


build.sbt
```diff
@@ -39,7 +39,10 @@ inThisBuild(Def.settings(
   onLoad in Global := {
     sLog.value.info(s"Building Akka HTTP ${version.value} against Akka ${AkkaDependency.akkaVersion}")
     (onLoad in Global).value
-  }
+  },
+  scalacOptions ++= Seq(
+    "-P:silencer:globalFilters=is deprecated \\(since 2\\.13\\." // remove when 2.12 support is dropped
```
raboof, line 44: Hmm, I guess we do want to see this warning for things that are deprecated in 2.13 for which already in 2.12 there was an alternative available... but perhaps that's rare.

jrudolph: Yeah, it seems rarer than the other way round currently.

project/Common.scala
```diff
@@ -39,6 +41,11 @@ object Common extends AutoPlugin {
     javacOptions in (Compile, compile) ++=
       // From jdk9 onwards this is covered by the '-release' flag above
       onlyOnJdk8("-target", "1.8"),
+
+    // in test code we often use destructing assignment, which now produces an exhaustiveness warning
+    // when the type is asserted
+    scalacOptions in (Test, compile) += "-P:silencer:globalFilters=match may not be exhaustive",
```
jrudolph, line 47: I added this. Do you of any other way of avoiding these, @raboof?

raboof: No, this seems reasonable to me!

PR #3211, opened 2020-06-03
tests: use AkkaSpecWithMaterializer in more places and make it slightly nicer

jrudolph:


akka-http-core/src/test/scala/akka/http/impl/util/AkkaSpecWithMaterializer.scala
```diff
@@ -25,7 +34,20 @@ abstract class AkkaSpecWithMaterializer(s: String)
...
       materializer.asInstanceOf[ActorMaterializer].shutdown()
+      Http().shutdownAllConnectionPools()
       // materializer shutdown is async but cannot be watched
       Thread.sleep(10)
     }
 }
+object AkkaSpecWithMaterializer {
+  // adapted version of AkkaSpec.getCallerName that also works for `AkkaSpecWithMaterializer`
+  def getCallerName(clazz: Class[_]): String = {
+    val s = (Thread.currentThread.getStackTrace map (_.getClassName) drop 1)
```
ennru, line 45: @jrudolph is in the house!

jrudolph: Wasn't my idea but just copied from AkkaSpec ;) I'm not even quite sure why it's needed, somehow it seemed that due to the initializers or something, the concrete class wasn't available otherwise.

jrudolph:
> which does seem a bit harsh on the `SystemMaterializer`?

That's "as designed". There's a comment explaining that reasoning. Since it's in `beforeTermination` it's right before the ActorSystem is terminated. I guess it might still be harsh if some infrastructure tries to use streams during shutdown but it seems to work so far...

akka-http-core/src/test/scala/akka/http/impl/util/AkkaSpecWithMaterializer.scala
jrudolph: This came back after all and needed another fix: [https://github.com/akka/akka-http/pull/3218/commits/458807442d5a88d219297f7f54b35b7d78667fda ](https://github.com/akka/akka-http/pull/3218/commits/458807442d5a88d219297f7f54b35b7d78667fda) ;)

PR #3213, opened 2020-06-03
tests: proxy transport instead of disabling hostname verification

raboof:
Keep one test explicitly testing disabling hostname verification

### DIFF
--- a/akka-http-core/src/test/scala/akka/http/scaladsl/GracefulTerminationSpec.scala
+++ b/akka-http-core/src/test/scala/akka/http/scaladsl/GracefulTerminationSpec.scala
@@ -21,8 +21,6 @@ import akka.stream.testkit.scaladsl.TestSink
 import akka.stream.{ Server => _, _ }
 import akka.testkit._
 import akka.util.ByteString
-import com.typesafe.sslconfig.akka.AkkaSSLConfig
-import com.typesafe.sslconfig.ssl.{ SSLConfigSettings, SSLLooseConfig }
 import org.scalactic.Tolerance
 import org.scalatest.concurrent.Eventually
 import org.scalatest.Assertion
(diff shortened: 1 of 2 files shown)


akka-http-core/src/test/scala/akka/http/scaladsl/GracefulTerminationSpec.scala
```diff
@@ -34,6 +32,7 @@ import scala.util.Success
 
 class GracefulTerminationSpec
   extends AkkaSpecWithMaterializer("""
+    akka.loglevel = debug
```
jrudolph, line 35: already set in `AkkaSpecWithMaterializer`

akka-http-core/src/test/scala/akka/http/scaladsl/GracefulTerminationSpec.scala
```diff
@@ -199,7 +198,8 @@ class GracefulTerminationSpec
         super.basePoolSettings
           .withTransport(new ClientTransport {
             override def connectTo(host: String, port: Int, settings: ClientConnectionSettings)(implicit system: ActorSystem): Flow[ByteString, ByteString, Future[Http.OutgoingConnection]] = {
-              ClientTransport.TCP.connectTo(host, port, settings)
+              ExampleHttpContexts.proxyTransport(serverBinding.localAddress)
```
jrudolph, line 201: Does that change anything here? You could probably also just keep using `TCP.connectTo` and pass in the `serverBinding`'s host and port.

raboof: Good one, indeed

akka-http-core/src/test/scala/akka/http/scaladsl/GracefulTerminationSpec.scala
raboof: oops

raboof:
(travis is green but not reporting back)

PR #3215, opened 2020-06-04
build: Set build-time jvm locale to en_US.UTF-8

raboof:
This prevents problems with 'akka-http-tests/lib/i have sp��ces.jar' in
environments where the locale is not UTF-8


jrudolph:
Do we also need to adapt `sun.jnu.encoding`? (I use my own past issue as a reference for an environment where utf8 did not work: https://github.com/AdoptOpenJDK/openjdk-docker/issues/105)

raboof:
> Do we also need to adapt `sun.jnu.encoding`?

Good one, added

PR #3216, opened 2020-06-04
Properly mark javadsl ConnectionContext.https deprecated

raboof:


akka-http-core/src/main/scala/akka/http/javadsl/ConnectionContext.scala
```diff
@@ -41,8 +41,12 @@ object ConnectionContext {
...
       OptionConverters.toScala(sslParameters))
 
-  /** Used to serve HTTPS traffic. */
-  // for binary-compatibility, since 2.4.7
+  /**
+   * Used to serve HTTPS traffic.
+   *
+   * @deprecated since 2.4.7
+   */
+  @Deprecated
```
jrudolph, line 49: Shouldn't we just remove it? I.e. how likely is it that it was more recently chosen by javac's overload resolution?

raboof:
Hmm, yeah, also the only change is dropping the AkkaSSLConfig we're moving away from in #2983 - perhaps let's just take care of it there

PR #3218, opened 2020-06-04
build: update to Scala 2.13.2 / 2.12.11

jrudolph:
And latest genjavadoc to support those.

### DIFF
--- a/akka-http-core/src/test/scala/akka/http/impl/util/AkkaSpecWithMaterializer.scala
+++ b/akka-http-core/src/test/scala/akka/http/impl/util/AkkaSpecWithMaterializer.scala
@@ -11,17 +11,17 @@ import akka.testkit.AkkaSpec
 import akka.testkit.EventFilter
 import com.typesafe.config.ConfigFactory
 
-abstract class AkkaSpecWithMaterializer(s: String)
+abstract class AkkaSpecWithMaterializer(configOverrides: String)
   extends AkkaSpec(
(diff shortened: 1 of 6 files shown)


akka-http-core/src/test/scala/akka/http/impl/util/AkkaSpecWithMaterializer.scala
```diff
@@ -41,13 +41,8 @@ abstract class AkkaSpecWithMaterializer(s: String)
...
+  def callerName(): String =
+    (Thread.currentThread.getStackTrace map (_.getClassName) drop 1)
       .dropWhile(_ matches "(java.lang.Thread|.*AkkaSpecWithMaterializer.?$|.*StreamSpec.?$)")
-    val reduced = s.lastIndexWhere(_ == clazz.getName) match {
-      case -1 => s
-      case z  => s drop (z + 1)
-    }
-    reduced.head.replaceFirst(""".*\.""", "").replaceAll("[^a-zA-Z_0-9]", "_")
-  }
+      .head.replaceFirst(""".*\.""", "").replaceAll("[^a-zA-Z_0-9]", "_")
```
raboof, line 47: Hard to see what we lose here (if anything)

jrudolph: Maybe there were some setups where `getClass` worked differently but not here.

PR #3219, opened 2020-06-04
tests: remove useless getClass call in AkkaSpecWithMaterializer

jrudolph:
This call failed at least one build. After all `getClass` here always means
`Predef.getClass`, so it will not help at all.

(See https://www.scala-lang.org/old/node/1648 for more information)


jrudolph:
Will add to update to 2.13.2 PR, since the fix is needed there.

PR #3223, opened 2020-06-05
Update javadsl bindAndHandle to take a Java function and System

raboof:


raboof:
> have we decided on the akka Function vs. Java Function? I.e. is it ok for us that users have to deal with checked exceptions?

Ah, no, I think I opened this PR to start that discussion but then never linked those ;)

So the reasoning here is:
* we want to gracefully deal with unchecked exceptions from these user-provided functions anyway
* in Scala we deal with checked exceptions from user-provided functions as well
* so in Java we also want users to be able to throw checked exceptions
* so we need to introduce a function type other than `java.util.function.Function`

WDYT, should we just drop this PR or update the functions to take `akka.japi.function.Function` instead of the old `akka.japi.Function`?

jrudolph:
> update the functions to take `akka.japi.function.Function` instead of the old `akka.japi.Function`?

That seems a useful update anyway. And also the change to use the system seems useful.

raboof:
> That seems a useful update anyway. And also the change to use the system seems useful.

Yeah it's a happy coincidence that we can change both and avoid ambiguity - it does mean that we should pick the right Function type now since there likely won't be a similar opportunity in the future ;)

jrudolph:
Does that fix all the occurrences of the old `japi.Function`? @ennru was a bit concerned that the additional overloads could confuse Java users.

akka-http-core/src/main/scala/akka/http/javadsl/Http.scala
```diff
@@ -288,6 +387,30 @@ class Http(system: ExtendedActorSystem) extends akka.actor.Extension {
       .map(new ServerBinding(_))(ec).toJava
   }
 
+  /**
+   * Convenience method which starts a new HTTP server at the given endpoint and uses the given `handler`
+   * [[akka.stream.javadsl.Flow]] for processing all incoming connections.
```
ennru, line 392: This handler is not a flow.

akka-http-core/src/main/scala/akka/http/javadsl/Http.scala
```diff
@@ -232,7 +258,30 @@ class Http(system: ExtendedActorSystem) extends akka.actor.Extension {
...
+    connect: ConnectHttp,
+    system:  ClassicActorSystemProvider): CompletionStage[ServerBinding] = {
+    val connectionContext = connect.effectiveConnectionContext(defaultServerHttpContext).asScala
+    delegate.bindAndHandleSync(handler.apply(_).asScala, connect.host, connect.port, connectionContext)(Materializer.matFromSystem(system))
+      .map(new ServerBinding(_))(ec).toJava
+  }
+
+  /**
+   * Convenience method which starts a new HTTP server at the given endpoint and uses the given `handler`
+   * [[akka.stream.javadsl.Flow]] for processing all incoming connections.
```
ennru, line 271: This handler is not a Flow.

jrudolph:
Still marked as draft. But please merge if ready :)

PR #3224, opened 2020-06-05
Make the http2 stream id an attribute

raboof:


akka-http2-support/src/main/scala/akka/http/impl/engine/http2/ResponseRendering.scala
```diff
@@ -41,7 +40,7 @@ private[http2] object ResponseRendering {
     val serverHeader = settings.serverHeader.map(h => h.lowercaseName -> h.value)
 
     { (response: HttpResponse) =>
-      val streamId = response.header[Http2StreamIdHeader].getOrElse(failBecauseOfMissingHeader).streamId
+      val streamId = response.attribute(Http2.streamId).getOrElse(failBecauseOfMissingHeader)
```
jrudolph, line 43: The exception needs a fix/rewording.

akka-http2-support/src/main/scala/akka/http/scaladsl/Http2.scala
```diff
@@ -183,6 +183,8 @@ final class Http2Ext(private val config: Config)(implicit val system: ActorSyste
 }
 
 object Http2 extends ExtensionId[Http2Ext] with ExtensionIdProvider {
+  val streamId = AttributeKey[Int]("x-http2-stream-id")
```
jrudolph, line 186: Maybe rename to `StreamIdAttr` or `StreamIdAttribute`?

Not sure that is best place for it but as long as it is still internal, it doesn't really matter.

raboof: This looks so repetitive:

    response.attribute(Http2.streamIdAttribute)

... maybe:

    response.attribute(`x-http2-stream-id`)

?

jrudolph: Not sure, that looks as if we'd like to pretend that the attribute is still a header...

Then maybe just keep it as `streamId`.

raboof: Oh, I kindof liked the idea ;) - but no big deal since it's internal anyway.

> Not sure that is best place for it

Agreed, but didn't see an obvious place for it in `akka.http.impl.engine.http2` either

jrudolph:
(sorry, tried to resolve some conflicts, but messed up the imports)

jrudolph:
(also fixed the newly introduced Http2ClientServerSpec)

Issue #2463, opened 2019-03-13
Support HTTP2 from bindAndHandle


raboof:
Currently HTTP2 is only supported for "bindAndHandleAsync", it would be nice to also support it for "bindAndHandle"

heraklos:
Hello @raboof .
Can I work on this issue?

I plan to Implement HTTP2-supporting  "bindAndHandle" and let the function be called from "bindAndHandle" in HttpExt according to passed argument.

raboof:
That sounds great! Balancing readability and performance might be challenging here, I'm looking forward to see what you will come up with!

heraklos:
I've got stuck in how to implement this... 
Please anyone interested in this issue take over it.
Sorry for holding this issue so long 🙇

jrudolph:
Yeah, no worries, let us know if you want to continue at some point and need some pointers.

jrudolph:
The functionality should be simple to add by doing a bit of refactoring in `Http2`. In the `bindAndHandle` API, users will have to propagate the stream id header manually. We should probably that to an attribute before (right now it's internal anyway). If we merge the API idea from the HTTP/2 client PR, we could provide some API in HttpRequest or HttpResponse to make it easier to propagate attributes that are marked as such.

raboof:
This is superseded by the newServerAt API's

PR #3225, opened 2020-06-08
Http2 bind and handle

raboof:
References #2463 

* not 'hooked in' to the Http extension yet
* needs testing
* especially the http 1 upgrade support could probably be more efficient.


akka-http2-support/src/main/scala/akka/http/scaladsl/Http2.scala
```diff
@@ -38,6 +38,23 @@ final class Http2Ext(private val config: Config)(implicit val system: ActorSyste
...
 
+  def bindAndHandle(
+    handler:   Flow[HttpRequest, HttpResponse, Any],
+    interface: String, port: Int = DefaultPortForProtocol,
+    connectionContext: ConnectionContext = http.defaultServerHttpContext,
+    settings:          ServerSettings    = ServerSettings(system),
+    parallelism:       Int               = 1,
+    log:               LoggingAdapter    = system.log)(implicit fm: Materializer): Future[ServerBinding] = {
+    bindAndHandleImpl(
+      Http2Blueprint.handleWithStreamIdHeader(handler),
```
jrudolph, line 49: I see how that is nice for existing, non-reordering handler flows.

However, it also loses a significant part of HTTP2's appeal because it reintroduces head-of-line blocking. It would still be able to multiplex response streams, yes, but any request handling that can only return a response with an appropriate status code after some delay (like having to wait for the result of async callbacks), will block other responses on the stream from being sent out (and even from being processed when blocked responses pile up).

In the end, the `bindAndHandle` API could offer two different semantics under the same signature which both have drawbacks for the user:
 * the one suggested here where the user flow must be ordered one-to-one (which matches the original HTTP/1.1 semantics with HTTP pipelining enabled)
 * the one offered by `bindAndHandleImpl` where there are no constraints on flow ordering but the flow has to ensure that the `streamId` attribute is manually propagated

The second one is the functionally more useful, so we should provide it in some way.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/StreamIdBidiFlow.scala
```diff
@@ -0,0 +1,64 @@
...
+  val fromWrapped = Inlet[HttpResponse]("StreamIdBidi.fromWrapped")
+
+  override def initialAttributes = Attributes.name("StreamIdBidi")
+
+  val shape = BidiShape(in, toWrapped, fromWrapped, out)
+
+  override def toString = "One2OneBidi"
+
+  override def createLogic(effectiveAttributes: Attributes): GraphStageLogic = new GraphStageLogic(shape) {
+    private val idsInFlight = mutable.ArrayDeque[Option[Int]]()
```
jrudolph, line 30: This has the same problem as the One2OneBidi implementation in that it ever only pulls a single element. It needs a parallelism parameter that governs how many requests are pulled and then buffers them until they are fetched from downstream.

In the best case, this flow should be part of `FlowWithContext` (and maybe we already have it?) as a general way to add context propagation to ordered one-to-one Flows.

akka-http2-support/src/main/scala/akka/http/scaladsl/Http2.scala
```diff
@@ -133,29 +168,29 @@ final class Http2Ext(private val config: Config)(implicit val system: ActorSyste
...
           // is not present or if more than one is present
           case _ =>
             log.debug("Invalid upgrade request (http2-settings header missing or repeated)")
-            handler(req)
+            Source.single(req).via(handler).runWith(Sink.head)
         }
 
       case _ =>
-        handler(req)
+        Source.single(req).via(handler).runWith(Sink.head)
```
jrudolph, line 180: As you said, that probably needs an optimization. Otherwise, HTTP/1 performance will take quite a hit when we need one materialization per request on a persistent connection.

akka-http2-support/src/main/scala/akka/http/scaladsl/Http2.scala
```diff
@@ -57,7 +92,7 @@ final class Http2Ext(private val config: Config)(implicit val system: ActorSyste
       else settings.defaultHttpPort
 
     val http1 = Flow[HttpRequest].mapAsync(parallelism)(handleUpgradeRequests(handler, settings, parallelism, log)).join(http.serverLayer(settings, log = log))
-    val http2 = Http2Blueprint.handleWithStreamIdHeader(parallelism)(handler)(system.dispatcher).join(Http2Blueprint.serverStackTls(settings, log))
+    val http2 = handler.join(Http2Blueprint.serverStackTls(settings, log))
```
jrudolph, line 95: That's a nice and simple reuse :)

akka-http2-support/src/main/scala/akka/http/scaladsl/Http2.scala
raboof: Right - and the current `Route.toFlow` produces a non-parallel one-to-one `Flow`, so after that the 'damage is done' and we can't really untangle it anymore.

Does this mean we need a `bindAndHandle` signature that doesn't accept a `Flow`, but accepts something that gets turned into the right kind of `Flow` only after negotiation?

It kindof sounds like what we really want is indeed `bindAndHandleAsync`, perhaps with an utility for automatically converting the `Route` to `HttpRequest => Future[HttpResponse]` (we already have `Route.toFunction`). Of course the `bindAndHandle` that takes a `Flow` must stay, but then how do we prevent users from accidentally using that one with a `Route` that was meant to support Http2?

Removing the implicit `Route.toFlow` and instead making `Route.toFunction` implicit and providing a `bindAndHandle` that takes such a function might work, but seems like a pretty invasive change.

jrudolph: Could `Route.toFunction` be implicit as well? Then, we could say, that you should use `bindAndHandleAsync` with Routes if you want to support both Http 1 and 2.

We could still fix `Route.toFlow` as well to propagate the attribute but using that you would still lose proper multiplexing with HTTP/2.

But I'm also open for the more drastic approach of deprecating `RouteResult.routeToFlow` pointing to using `bindAndHandleAsync` instead (or i.e. to change the deprecation warning of `route2handlerflow` and adapting the message to say that). After all, the high-level DSL is not stream-based and is unlikely to become stream-based in the future.

raboof: (first step, allowing a `Route` in `bindAndHandleAsync`, is in https://github.com/akka/akka-http/pull/3237)

jrudolph: :+1: for a `bind` overload if that works out.

raboof:
See https://github.com/akka/akka-http/pull/3225#discussion_r437240045 on why this isn't attractive, and we want #3271 instead.

PR #3237, opened 2020-06-11
http: allow 'bindAndHandleAsync' to take a 'Route'

raboof:
Related to #2463

This makes H2Spec '6.4. RST_STREAM' emit a 'Trying to materialize stream after
materializer has been shutdown. Completing with 500 Internal Server Error
response.', but that might be a separate issue?

### DIFF
--- a/akka-http/src/main/mima-filters/10.1.x.backwards.excludes/route-to-function.backwards.excludes
+++ b/akka-http/src/main/mima-filters/10.1.x.backwards.excludes/route-to-function.backwards.excludes
@@ -1,0 +1,1 @@
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.javadsl.server.Route.function")
--- a/akka-http/src/main/mima-filters/10.1.x.backwards.excludes/route-to-function.backwards.excludes
+++ b/akka-http/src/main/mima-filters/10.1.x.backwards.excludes/route-to-function.backwards.excludes
@@ -1,1 +1,2 @@
+# New method in @DoNotInherit class
 ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.javadsl.server.Route.function")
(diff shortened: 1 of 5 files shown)


akka-http/src/main/mima-filters/10.1.x.backwards.excludes/route-to-function.backwards.excludes
```diff
@@ -0,0 +1 @@
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.javadsl.server.Route.function")
```
jrudolph, line 1:
```suggestion
# New method in @DoNotInherit class
ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.javadsl.server.Route.function")
```

akka-http/src/main/scala/akka/http/scaladsl/server/RouteResult.scala
```diff
@@ -43,6 +43,17 @@ object RouteResult {
...
 
+  /**
+   * Turns a `Route` into a server function.
+   *
+   * This implicit conversion is defined here because `Route` is an alias for
+   * `RequestContext => Future[RouteResult]`, and the fact that `RouteResult`
+   * is in that type means this implicit conversion come into scope whereever
+   * a `Route` is given but a `Function[HttpRequest, Future[HttpResponse]` is expected.
+   */
+  implicit def routeToFunction(route: Route)(implicit system: ClassicActorSystemProvider): Function[HttpRequest, Future[HttpResponse]] =
```
jrudolph, line 54: Was that syntax intentional? Had to lookup what that `Function` even refers to...  (it's `Predef.Function` which is an alias for `Function1`.

```suggestion
  implicit def routeToFunction(route: Route)(implicit system: ClassicActorSystemProvider): HttpRequest => Future[HttpResponse] =
```

raboof:
> is that just an extra log message in the test?

It's just logging (though with stacktrace), yes

> If yes, maybe create a ticket to look into it?

Added #3250

jrudolph:
Do we want to make this the default when used with the routing DSL? Needs an entry in the changelog then and updates in the documentation examples.

Added https://github.com/akka/akka-http/issues/3271

PR #3239, opened 2020-06-11
http: new `handle` directive to create Route from function

raboof:


akka-http/src/main/scala/akka/http/javadsl/server/directives/RouteDirectives.scala
```diff
@@ -295,6 +294,11 @@ abstract class RouteDirectives extends RespondWithDirectives {
     D.complete(value.asScala.fast.map(v => ToResponseMarshallable(v)(marshaller)).recover(unwrapCompletionException))
   }
 
+  def fromFunction(handler: akka.japi.Function[HttpRequest, CompletionStage[HttpResponse]]): Route = {
```
ennru, line 297: What did we say about `akka.japi.Function`?

akka-http/src/main/scala/akka/http/scaladsl/server/directives/RouteDirectives.scala
```diff
@@ -54,6 +56,14 @@ trait RouteDirectives {
    */
   def failWith(error: Throwable): StandardRoute =
     StandardRoute(_.fail(error))
+
+  /**
+   *
+   * @group route
+   */
+  def fromFunction(handler: HttpRequest => Future[HttpResponse]): StandardRoute =
+    { ctx => handler(ctx.request).map(RouteResult.Complete)(ctx.executionContext) }
```
jrudolph, line 65:
```suggestion
    { ctx => handler(ctx.request).fast.map(RouteResult.Complete)(ctx.executionContext) }
```

akka-http/src/main/scala/akka/http/javadsl/server/directives/RouteDirectives.scala
```diff
@@ -295,6 +294,11 @@ abstract class RouteDirectives extends RespondWithDirectives {
     D.complete(value.asScala.fast.map(v => ToResponseMarshallable(v)(marshaller)).recover(unwrapCompletionException))
   }
 
+  def fromFunction(handler: akka.japi.function.Function[HttpRequest, CompletionStage[HttpResponse]]): Route = {
+    import akka.http.impl.util.JavaMapping._
+    RouteAdapter { ctx => handler(ctx.request).asScala.map(response => RouteResult.Complete(response.asScala)) }
```
jrudolph, line 299: Is there a way to delegate to Scala version instead of reimplementing? Or is it too complicated to be worth it?

```suggestion
    RouteAdapter { ctx => handler(ctx.request).asScala.fast.map(response => RouteResult.Complete(response.asScala)) }
```

raboof:
Added tests, but not so sure if this should be the expected behaviour ;)

raboof:
> not so sure if this should be the expected behaviour ;)

I think this is the right behavior: failing (rather than rejecting) both when the function fails (consistent with e.g. extractRequest) and when the future fails (you probably don't want to use failing asynchronous calls for 'flow control', or if you do it should be intentional)

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/RouteDirectivesSpec.scala
```diff
@@ -157,6 +158,30 @@ class RouteDirectivesSpec extends AnyFreeSpec with GenericRoutingSpec {
     }
   }
 
+  "the fromFunction directive" should {
+    "use a function to complete a request" in {
+      Get(Uri("https://akka.io/foo")) ~> {
+        fromFunction(req => Future.successful(HttpResponse(OK, entity = req.uri.toString)))
```
raboof, line 164: Is 'fromFunction' the right name? It seems rather 'technical', but looking for a more 'functional' name I can only come up with rather generic terms like 'handle'.

There is already `Route` apply method that takes a function, so we should probably not overload that ;).

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/RouteDirectivesSpec.scala
```diff
@@ -157,6 +158,30 @@ class RouteDirectivesSpec extends AnyFreeSpec with GenericRoutingSpec {
...
+      } ~> check { response shouldEqual HttpResponse(200, entity = "https://akka.io/foo") }
+    }
+    "reject the request when the future fails" in {
```
jrudolph, line 167:
```suggestion
    "fail the request when the future fails" in {
```

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/RouteDirectivesSpec.scala
```diff
@@ -157,6 +158,30 @@ class RouteDirectivesSpec extends AnyFreeSpec with GenericRoutingSpec {
...
+      Get(Uri("https://akka.io/foo")) ~> {
+        concat(
+          fromFunction(req => Future.failed(new IllegalStateException("Some error"))),
+          complete(ImATeapot)
+        )
+      } ~> check { response shouldEqual HttpResponse(500, entity = "There was an internal server error.") }
+    }
+    "reject the request when the function throws fails" in {
```
jrudolph, line 175:
```suggestion
    "fail the request when the function throws" in {
```

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/RouteDirectivesSpec.scala
```diff
@@ -157,6 +158,30 @@ class RouteDirectivesSpec extends AnyFreeSpec with GenericRoutingSpec {
...
+      Get(Uri("https://akka.io/foo")) ~> {
+        concat(
+          fromFunction(req => throw new IllegalStateException("Some error")),
+          complete(ImATeapot)
+        )
+      } ~> check { response shouldEqual HttpResponse(500, entity = "There was an internal server error.") }
+    }
+  }
+
```
jrudolph, line 184: Maybe we could also support the `RejectionError` to reject from such a function? Though, seems a bit weird mix of APIs if you go that way...

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/RouteDirectivesSpec.scala
```diff
@@ -15,11 +16,11 @@ import akka.http.scaladsl.server._
...
 import StatusCodes._
-import org.scalatest.freespec.AnyFreeSpec
+import org.scalatest.wordspec.AnyWordSpec
 
-class RouteDirectivesSpec extends AnyFreeSpec with GenericRoutingSpec {
+class RouteDirectivesSpec extends AnyWordSpec with GenericRoutingSpec {
 
-  "The `complete` directive should" - {
+  "The `complete` directive" should {
     "by chainable with the `&` operator" in {
```
jrudolph, line 24:
```suggestion
    "be chainable with the `&` operator" in {
```

Issue #3251, opened 2020-06-15
[Failure] value setApplicationProtocols is not a member of javax.net.ssl.SSLParameters


ignasi35:
seen in https://jenkins.akka.io:8498/job/akka-http-nightly/160/AKKA_VERSION=master,Node=akka-http-nightly-node,SCALA_VERSION=2.13.1,jdk=AdoptOpenJDK%2011/console (can reproduce locally).

```
[info] Compiling 138 Scala sources and 8 Java sources to /home/jenkins/workspace/workspace/akka-http-nightly/AKKA_VERSION/master/Node/akka-http-nightly-node/SCALA_VERSION/2.13.1/jdk/AdoptOpenJDK 11/akka-http/target/scala-2.13/classes ...
[error] /home/jenkins/workspace/workspace/akka-http-nightly/AKKA_VERSION/master/Node/akka-http-nightly-node/SCALA_VERSION/2.13.1/jdk/AdoptOpenJDK 11/akka-http2-support/src/main/scala/akka/http/scaladsl/Http2.scala:195:14: value setApplicationProtocols is not a member of javax.net.ssl.SSLParameters
[error]       params.setApplicationProtocols(Array("h2"))
[error]              ^
[error] one error found
```

It's a weird issue since `setApplicationProtocols` was introduced in `javax.net.ssl.SSLParameters` on java 8.

ignasi35:
I'm suspicious of https://github.com/akka/akka-http/blob/529d33b31d8bd64c7dbc382620bde570adf1cd4f/project/Common.scala#L33-L40 but I'm more confused by the fact that `setApplicationProtocols` was available in JDK8.

ignasi35:
>  but I'm more confused by the fact that `setApplicationProtocols` was available in JDK8.

IT was backported in some versions of JDK8 _and_ documented as being available in JDK8, but JDK11 running with `-release 8` doesn't _see_ it.

ignasi35:
This failure snowballed into other jobs: https://jenkins.akka.io:8498/job/akka-http-docs/445/

jrudolph:
Fixed in #3256

PR #3256, opened 2020-06-16
http2: use reflection for `setApplicationProtocols`

jrudolph:
That's necessary because JDK11 with `-release 8` does not yet know about
the backported JDK 8 APIs.

refs #3251

### DIFF
--- a/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2AlpnSupport.scala
+++ b/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2AlpnSupport.scala
@@ -67,4 +67,14 @@ private[http] object Http2AlpnSupport {
     else None
 
   def applySessionParameters(engine: SSLEngine, sessionParameters: NegotiateNewSession): Unit = TlsUtils.applySessionParameters(engine, sessionParameters)
+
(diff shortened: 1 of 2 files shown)


akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2AlpnSupport.scala
```diff
@@ -67,4 +67,14 @@ private[http] object Http2AlpnSupport {
     else None
 
   def applySessionParameters(engine: SSLEngine, sessionParameters: NegotiateNewSession): Unit = TlsUtils.applySessionParameters(engine, sessionParameters)
+
+  private type SSLParametersWithALPNSupport = {
+    def setApplicationProtocols(protocols: Array[String]): Unit
+  }
+  def clientSetApplicationProtocols(engine: SSLEngine, protocols: Array[String]): Unit = {
+    val params = engine.getSSLParameters
+    params.asInstanceOf[SSLParametersWithALPNSupport].setApplicationProtocols(Array("h2"))
```
jrudolph, line 76: Will crash if not supported, but that seems fine for me right now (HTTP/2 client side not yet officially supported).

Issue #3255, opened 2020-06-16
No SNI when using explicit port


mpdn:
When making a HTTPS request, the TLS client hello does not include the SNI extension if using an explicit port.

The [docs](https://doc.akka.io/docs/akka-http/current/client-side/client-https-support.html#server-name-indication-sni-) seem to indicate that SNI is enabled by default, that it protects against MITM attacks (this seems incorrect), and that turning it off is an unsafe option and a "very bad idea".

I don't really see how SNI can be a security feature, but due to how it was presented in the docs I originally sent this to the security@akka.io. According to @jroper SNI should not be a security issue, so an open GH issue here is better.

Example
=======

```scala
import akka.actor.ActorSystem
import akka.http.scaladsl.Http
import akka.http.scaladsl.model.HttpRequest

import scala.concurrent.duration._
import scala.concurrent.{Await, ExecutionContextExecutor, Future}

object Main {
  def main(args: Array[String]): Unit = {
    implicit val system: ActorSystem = ActorSystem()
    implicit val executionContext: ExecutionContextExecutor = system.dispatcher

    val timeout = Duration(10, SECONDS)

    def await[T](fut: Future[T]) = Await.result(fut, timeout)

    def sniTest(uri: String) = {
      val response = Http().singleRequest(HttpRequest(uri = uri))
      println(await(await(response).entity.toStrict(timeout)).getData().utf8String)
    }

    println("Without explicit port:")
    sniTest("https://check-tls.akamaized.net/v1/tlssni.json")

    println("With explicit port:")
    sniTest("https://check-tls.akamaized.net/v1/tlssni.json:80")
  }
}

```

Outputs:

```
Without explicit port:
{
"tls_sni_status": "present",
"scheme": "https",
"tls_version": "tls1.2",
"tls_sni_value": "check-tls.akamaized.net",
"host_header_value": "check-tls.akamaized.net",
"tls_cipher_name": "ECDHE-ECDSA-AES128-GCM-SHA256",
"alpn": "",
"client_ip": "194.182.8.167",
"client_ip_version": "ipv4",
"client_port": 57700,
"user_agent": "akka-http%2f10.1.12",
"output_version": "0.1.21",
"timestamp": 1592290268
}

With explicit port:
Not found
```

Client hello output from wireshark:

```
Frame 87: 369 bytes on wire (2952 bits), 369 bytes captured (2952 bits) on interface lo, id 0
[...]

PR #3257, opened 2020-06-16
docs: don't claim that SNI is a security feature

jrudolph:
Refs #3255

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
@@ -1051,8 +1051,7 @@ trait DefaultSSLContextCreation {
         WarningAboutGlobalLoose)
 
     if (sslConfig.config.loose.disableSNI) {
-      log.warning("Detected that Server Name Indication (SNI) is disabled globally (via ssl-config's akka.ssl-config.loose.disableSNI) for the Http extension! " +
-        WarningAboutGlobalLoose)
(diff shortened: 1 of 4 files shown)


akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
```diff
@@ -1051,8 +1051,7 @@ trait DefaultSSLContextCreation {
         WarningAboutGlobalLoose)
 
     if (sslConfig.config.loose.disableSNI) {
-      log.warning("Detected that Server Name Indication (SNI) is disabled globally (via ssl-config's akka.ssl-config.loose.disableSNI) for the Http extension! " +
-        WarningAboutGlobalLoose)
+      log.warning("Detected that Server Name Indication (SNI) is disabled globally (via ssl-config's akka.ssl-config.loose.disableSNI) for the Http extension!")
```
jrudolph, line 1054: Removed the drastic part of that warning. Might still be good to warn if it globally disabled, as it may prevent accessing multi-homed servers.

docs/src/test/scala/docs/http/scaladsl/HttpsExamplesSpec.scala
```diff
@@ -18,13 +18,13 @@ class HttpsExamplesSpec extends AnyWordSpec with Matchers with CompileOnlySpec {
 
   "disable SNI for connection" in compileOnlySpec {
```
jroper, line 19:
```suggestion
  "disable hostname verification for connection" in compileOnlySpec {
```

jrudolph: Thanks missed that one :)

Issue #3247, opened 2020-06-12
Downstream connection failure is not propagated into WebSockets flow


raboof:
Even when using Akka 2.6, when a request is handled using a websocket flow, and the connection is aborted by the outgoing TCP connection, this failure is not propagated to the websocket flows' `watchTermination`.

I suspect this is because the `ProtocolSwitchStage` completes the stage when downstream finishes,  (https://github.com/akka/akka-http/blob/master/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala#L602), without taking into account any cancellation reason.

Setting a cancellation strategy of `AfterDelay` (to make sure we pick up the upstream failure before handling the downstream cancellation) didn't appear to help, but perhaps I added it in the wrong way.

jrudolph:
Is that about the server or the client side?

raboof:
This is about the server side

jrudolph:
Seems this is fixed in some regards in 10.2.0-M1. I guess the reason is that we only backported the client-side cancellation race fixes to 10.1.12 but not the server-side fix which came later and also fixed it on the server-side.

Also we still need a fix for the ProtocolSwitchStage to materialize the inner flow with the cancellation delay enabled, also for 10.2.x.

jrudolph:
@raboof this should be fixed in master and with #3258 also in release-10.1. Could you convert your test repro into a test to make sure it's really fixed?

raboof:
> and with #3258 also in release-10.1

Yes, verified

> Could you convert your test repro into a test to make sure it's really fixed?

I'll have a look.

jrudolph:
Test added in #3276.

jrudolph:
10.1 backport of test in #3288.

PR #3258, opened 2020-06-16
[10.1 backport] server-side streamCancellationDelay

jrudolph:
Backport of #3022

The client-side part was backported as part of the big bunch of backports but this one came afterwards and was missed.

I looked at #3022 to find out whether a backport was intended or not but I couldn't find one. It's a rather intrusive change so there's certainly an argument to be made to not go there for 10.1.x (on the other hand, we did already change client-side behavior for 10.1.12)

Refs #3247


jrudolph:
Ok, so let's pull it in so we have a fix for #3247 also in 10.1.

PR #3261, opened 2020-06-17
core: fix some warnings

jrudolph:


akka-http-core/src/test/scala/akka/http/impl/engine/client/ConnectionPoolSpec.scala
```diff
@@ -349,7 +349,8 @@ class NewConnectionPoolSpec extends AkkaSpecWithMaterializer("""
 
       responseOutSub.request(1)
       acceptIncomingConnection()
-      val (Success(_), 42) = responseOut.expectNext()
+      val (Success(_), result) = responseOut.expectNext()
+      result shouldEqual 42
```
raboof, line 353: Sad that this is no longer allowed :). I'd be fine with silencing those warnings for the test scope...

akka-http-core/src/test/scala/akka/http/scaladsl/settings/SettingsEqualitySpec.scala
```diff
@@ -22,13 +22,13 @@ class SettingsEqualitySpec extends AnyWordSpec with Matchers {
...
 
-    "hold for ParserSettings" in {
+    "hold for ParserSettings.forServer" in {
       val s1 = ParserSettings(config)
       val s2 = ParserSettings(config)
 
       s1 shouldBe s2
       s1.toString should startWith("ParserSettings(")
-    }
+    }: @silent("apply in object ParserSettings is deprecated")
```
raboof, line 31: Cool trick! ... but shouldn't we actually use `forServer` (or `forClient`) here instead of silencing the warning?

jrudolph: Yes, but those only work from an ActorSystem currently.

raboof: ahh didn't think of that.

akka-http-core/src/test/scala/akka/http/impl/engine/client/ConnectionPoolSpec.scala
jrudolph: Actually, I don't think those usages are forbidden, I just don't understand the warning or the solution :)

```
[warn] /home/johannes/git/opensource/akka-http/akka-http-core/src/test/scala/akka/http/impl/engine/client/ConnectionPoolSpec.scala:352:11: Pattern definition introduces Unit-valued member of $anon; consider wrapping it in `locally { ... }`.
```

jrudolph: Ah, I know understood. Because with is wrapped in an anonymous `new TestSetup` this pattern match would actually create a field (which doesn't matter in the test).

jrudolph: Extra braces around the statement helped.

Issue #3232, opened 2020-06-09
HTTP/2 demux: unify incoming and outgoing stream handling


jrudolph:
Currently, incoming and outgoing ends of the same bidirectional substream in HTTP/2 are treated separately. This already led to confusion and bugs. We should have only one state machine per stream where states hold the particular incoming and outgoing stream handles as needed.

jrudolph:
Fixed in #3552.

PR #3267, opened 2020-06-22
Fix warning in Http2 server side

### LINKED ISSUE #3264: Http2ServerSpec.The Http/2 server implementation should support stream for request entity data should close the stream when we receive a RST after we have half-closed ourselves as well
https://jenkins.akka.io:8498/job/akka-http-nightly/166/AKKA_VERSION=default,Node=akka-http-nightly-node,SCALA_VERSION=2.13.2,jdk=AdoptOpenJDK%2011/

Note: this is a detailed failure report. See [this](https://github.com/akka/akka-http/issues?q=is%3Aopen+label%3Afailed+Http2ServerSpec) other failures in `Http2ServerSpec` 

```
sbt.ForkMain$ForkError: java.lang.AssertionError: received 1 excess messages on WarningFilter(None,Left(),false)
	at akka.testkit.EventFilter.intercept(TestEventListener.scala:121)
	at akka.http.impl.engine.http2.Http2ServerSpec$$anon$12.<init>(Http2ServerSpec.scala:317)
	at akka.http.impl.engine.http2.Http2ServerSpec.$anonfun$new$17(Http2ServerSpec.scala:308)
```

jrudolph:
The client-side PR introduced a new serious looking warning on the server-side because I introduced another signal which came in at an unexpected time. This PR adds another state for the regular server case, so that the warning is not shown any more.

Also, it adds a bit more tooling to allow wholesale checking for extra warnings/error log messages in tests.

Fixes #3264

### DIFF
--- a/akka-http-core/src/test/scala/akka/http/impl/util/WithLogCapturing.scala
+++ b/akka-http-core/src/test/scala/akka/http/impl/util/WithLogCapturing.scala
@@ -10,13 +10,15 @@ import akka.actor.ActorSystem
 import akka.event.Logging
 import akka.event.Logging._
 import akka.testkit.{ EventFilter, TestEventListener }
-import org.scalatest.{ Outcome, SuiteMixin, TestSuite }
+import org.scalatest.{ Failed, Outcome, SuiteMixin, TestSuite }
 
 /**
  * Mixin this trait to a test to make log lines appear only when the test failed.
  */
 trait WithLogCapturing extends SuiteMixin { this: TestSuite =>
   implicit def system: ActorSystem
+
+  def failOnSevereMessages: Boolean = false
 
   abstract override def withFixture(test: NoArgTest): Outcome = {
     // When filtering just collects events into this var (yeah, it's a hack to do that in a filter).
@@ -38,14 +40,32 @@ trait WithLogCapturing extends SuiteMixin { this: TestSuite =>
       r
     }
 
-    if (!(res.isSucceeded || res.isPending)) {
+    def flushLog(): Unit = {
       println(s"--> [${Console.BLUE}${test.name}${Console.RESET}] Start of log messages of test that [$res]")
       val logger = new StdOutLogger {}
-      withPrefixedOut("| ") { events.reverse.foreach(logger.print) }
+      withPrefixedOut("| ") {
+        events.reverse.foreach { ev =>
+          if (ev.level == WarningLevel) print(Console.YELLOW)
+          else if (ev.level == ErrorLevel) print(Console.RED)
+          logger.print(ev)
+          if (ev.level <= WarningLevel) print(Console.RESET)
+        }
+      }
       println(s"<-- [${Console.BLUE}${test.name}${Console.RESET}] End of log messages of test that [$res]")
     }
 
-    res
+    if (!(res.isSucceeded || res.isPending)) {
+      flushLog()
+      res
+    } else if (failOnSevereMessages && events.exists(_.level <= Logging.WarningLevel)) {
+      val stats = events.groupBy(_.level).view.mapValues(_.size).toMap.withDefaultValue(0)
+      flushLog()
+
+      Failed(new AssertionError(
+        s"No severe log messages should be emitted during test run but got [${stats(Logging.WarningLevel)}] warnings and [${stats(Logging.ErrorLevel)}] errors (see marked lines above)"
+      ))
+    } else res
+
   }
 
   /** Adds a prefix to every line printed out during execution of the thunk. */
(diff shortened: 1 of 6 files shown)


akka-http-core/src/test/scala/akka/http/impl/util/WithLogCapturing.scala
```diff
@@ -38,14 +40,32 @@ trait WithLogCapturing extends SuiteMixin { this: TestSuite =>
...
-    res
+    if (!(res.isSucceeded || res.isPending)) {
+      flushLog()
+      res
+    } else if (failOnSevereMessages && events.exists(_.level <= Logging.WarningLevel)) {
+      val stats = events.groupBy(_.level).view.mapValues(_.size).toMap.withDefaultValue(0)
+      flushLog()
+
+      Failed(new AssertionError(
+        s"No severe log messages should be emitted during test run but got [${stats(Logging.WarningLevel)}] warnings and [${stats(Logging.ErrorLevel)}] errors (see marked lines above)"
```
raboof, line 65: :+1: nice

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -307,20 +309,16 @@ class Http2ServerSpec extends AkkaSpecWithMaterializer("""
       // Reproducing https://github.com/akka/akka-http/issues/2957
       "close the stream when we receive a RST after we have half-closed ourselves as well" in new WaitingForRequestData {
         // Client sends the request, but doesn't close the stream yet. This is a bit weird, but it's whet grpcurl does ;)
-        sendHEADERS(streamId = 1, endStream = false, endHeaders = true, encodeRequestHeaders(request))
-        sendDATA(streamId = 1, endStream = false, ByteString(0, 0, 0, 0, 0x10, 0x22, 0x0e) ++ ByteString.fromString("GreeterService"))
+        sendDATA(streamId = TheStreamId, endStream = false, ByteString(0, 0, 0, 0, 0x10, 0x22, 0x0e) ++ ByteString.fromString("GreeterService"))
```
raboof, line 312: Why remove the headers here?

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -229,6 +235,10 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with StageLoggi
     override def shutdown(): IncomingStreamState = Closed
   }
   case object Closed extends IncomingStreamState {
+
+    override def handleOutgoingCreated(stream: Http2SubStream): IncomingStreamState = this
```
raboof, line 239: When would this happen?

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
jrudolph: Because the test setup already includes sending a request. It seems we ignore later HEADERS frames, we need them to support trailers but maybe not in the middle of the stream? Created https://github.com/akka/akka-http/issues/3268 to dig further into this.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
jrudolph: I agree, it's a bit hand-wavy to just ignore those signals here.

IIRC, it could happen in cases, when a stream is cancelled before the response has been sent out. Right now this kind of coordination between incoming and outgoing stream side is hard to coordinate because the split of the two ends. In the commit comment I refer to the cleanup planned with #3232 to look into these kind of scenarios more deeply.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -307,20 +309,16 @@ class Http2ServerSpec extends AkkaSpecWithMaterializer("""
...
-          // Wait to give the warning (that we hope not to see) time to pop up.
-          Thread.sleep(3000)
-        }
+        emitResponse(streamId = TheStreamId, HttpResponse(StatusCodes.NotFound))
+
+        // The client closes the stream with a protocol error. This is somewhat questionable but it's what grpc-go does
+        sendRST_STREAM(streamId = TheStreamId, ErrorCode.PROTOCOL_ERROR)
+        entityDataIn.expectError()
+        // Wait to give the warning (that we hope not to see) time to pop up.
+        Thread.sleep(100)
```
jrudolph, line 321: Reduced time to wait here, maybe we could remove it completely or add another assertion that would ensure that the server has processed the frame.

PR #3275, opened 2020-06-23
tests: don't use `view.mapValues` as it breaks 2.12 compilation

jrudolph:


jrudolph:
Going forward with merging since it breaks the 2.12 nightlies (interestingly nothing else because all other 2.12 infra on travis doesn't even compile tests).

PR #3276, opened 2020-06-23
Add test for failure propagation into websocket flows

raboof:
Refs #3247

### DIFF
--- a/akka-http-core/src/test/scala/akka/http/impl/engine/ws/WebSocketIntegrationSpec.scala
+++ b/akka-http-core/src/test/scala/akka/http/impl/engine/ws/WebSocketIntegrationSpec.scala
@@ -194,6 +194,40 @@ class WebSocketIntegrationSpec extends AkkaSpec("akka.stream.materializer.debug.
       binding.unbind()
     }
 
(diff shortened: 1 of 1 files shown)


akka-http-core/src/test/scala/akka/http/impl/engine/ws/WebSocketIntegrationSpec.scala
```diff
@@ -194,6 +194,40 @@ class WebSocketIntegrationSpec extends AkkaSpec("akka.stream.materializer.debug.
...
+        port = 0)
+      val binding = Await.result(bindingFuture, 3.seconds.dilated)
+      val myPort = binding.localAddress.getPort
+
+      val ((switch, connection), completion) =
+        Source.maybe
+          .viaMat {
+            Http().webSocketClientLayer(WebSocketRequest("ws://localhost:" + myPort))
+              .atop(TLSPlacebo())
+              .atopMat(KillSwitches.singleBidi[ByteString, ByteString])(Keep.right)
```
jrudolph, line 217: I guess that could be using a `ClientTransport` instead for hooking in the kill switch but I guess that's ok as well.

PR #3277, opened 2020-06-23
build: stay on ScalaTest 3.1.x

ennru:
Replaces #3273

### DIFF
--- a/.scala-steward.conf
+++ b/.scala-steward.conf
@@ -13,7 +13,7 @@ updates.ignore = [
 ]
 
 updates.pin = [
-  // Stay on ScalaTest 3.1 as long as  Akka TestKit does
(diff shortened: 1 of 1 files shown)


.scala-steward.conf
```diff
@@ -13,6 +13,8 @@ updates.ignore = [
 ]
 
 updates.pin = [
+  // Stay on ScalaTest 3.1 as long as  Akka TestKit does
```
raboof, line 16: Are `//` allowed here for comments?

ennru: I'd believe this is Typesafe Config, so yes. For consistence I changed to `#`.

raboof: TIL HOCON supports both `#` and `//` :laughing:

Issue #3266, opened 2020-06-22
avoid adapted arguments in headerValueByType


raboof:
Subtask of #2738

`headerValueByType[Foo]() { foo => ... }` works through arguments adaptation: it's actually `headerValueByType[Foo](()) { foo => ... }`, and the `Unit` value is converted to a `magnet: HeaderMagnet[T]` parameter by implicit conversions.

The reason the magnet pattern is used here is because `Foo` can either be a 'normal' `HttpHeader`, or a `ModeledCustomHeader` in which case we also implicitly find the `ModeledCustomHeaderCompanion`.

Unfortunately I don't see a way to change the current implicit conversion (of unit) to an implicit parameter, as AFAIK values (unlike conversions) can't depend on further implicit resolution to find the right candidate.

jrudolph:
We could probably match by companion in most cases (i.e. not passing a type/ClassTag but the companion) which would avoid the issue but it's a bit hard to make that kind of change right now.

raboof:
Do you mean so we'd have `headerValueByType[TheClass]` for 'normal' headers and `headerValueByType(TheCompanion)` for 'modeled' ones? That could probably work, though it looks a bit surprising API-wise ;).

jrudolph:
> Do you mean so we'd have `headerValueByType[TheClass]` for 'normal' headers and `headerValueByType(TheCompanion)` for 'modeled' ones?

The current API allows to have header classes without a companion but we recommend to have companions in all cases, even for the `modeled-by-user` ones. A simple "fix" could thus be to deprecate the from-Unit implicit and point to using a version that uses a companion instead. The question would then be if we should provide those as 

 * new magnets for the existing signature
 * new overloads for the existing directive
 * create a completely new directive for that

raboof:
> we recommend to have companions in all cases

Ah, I didn't realize - so we could move to `headerValueByXxx(TheCompanion)` for all.

Doing this as new magnets for the existing signature seems fine to me: you could see the companion as 'part of the type', and that also keeps it somewhat more consistent with the javadsl.

I'll give this a try.

jrudolph:
Fixed in #3279.

PR #3279, opened 2020-06-23
Allow passing a companion object to headerValueByType

raboof:
Refs #3266


akka-http/src/main/scala/akka/http/scaladsl/server/directives/HeaderDirectives.scala
```diff
@@ -171,8 +171,12 @@ object HeaderMagnet extends LowPriorityHeaderMagnetImplicits {
 
   /**
    * If possible we want to apply the special logic for [[ModeledCustomHeader]] to extract custom headers by type,
-   * otherwise the default `fromUnit` is good enough (for headers that the parser emits in the right type already).
+   * otherwise the default `fromCompanion` is good enough (for headers that the parser emits in the right type already).
    */
+  implicit def fromCompanionForModeledCustomHeader[T <: ModeledCustomHeader[T], H <: ModeledCustomHeaderCompanion[T]](companion: ModeledCustomHeaderCompanion[T])(implicit tag: ClassTag[T]): HeaderMagnet[T] =
+    fromClassTagForModeledCustomHeader[T, H](tag, companion)
+
+  @deprecated("Pass the companion object to headerValueByType as a parameter instead", since = "10.2.0")
```
jrudolph, line 179: Let's add an example to that particular deprecation note to make it concrete how to do the conversion as so many usages will be affected.

Issue #3278, opened 2020-06-23
UpgradeToWebSocket as Attribute


raboof:
UpgradeToWebSocket is a 'virtual' header, perhaps we should start introducing an attribute for this and deprecate the header?

jrudolph:
> UpgradeToWebSocket is a 'virtual' header, perhaps we should start introducing an attribute for this and deprecate the header?

Yes, agreed, would be nice, but not sure if it should block 10.2.0. In any case, we could already do the deprecation without the alternative because it's public API but mostly hidden higher-level abstractions.

jrudolph:
Deprecation done in #3280. Leaving open until attribute is done.

PR #3280, opened 2020-06-23
Deprecate UpgradeToWebSocket

raboof:
Refs #3278

Recommend using the higher-level handleWebSocketMessages, or the
(yet to be implemented) attributes-based approach.


jrudolph:
This is missing the Scala side of things. Would also be good if warnings in code would be silenced where necessary because of this.

Issue #2916, opened 2020-01-31
Remove Deflate/GZip stream stages and use ones from akka-stream


jrudolph:

jrudolph:
We might be able to remove the decompressor side but the compressor side still is public API right now.

jrudolph:
At least make sure that all non-intended to be public classes be marked as deprecated / internal etc.

krestenkrab:
FYI, we (humio) have a rewritten GZip stream stage for http using `com.jcraft.jzlib` because the JVM native zip implementation ends up holding up the garbage collector all the time (it uses the `JNIEnv:: GetPrimitiveArrayCritical` that ends up blocking global GC).  Let us know if you want to incorporate it.

jrudolph:
Thanks for the great comment, @krestenkrab. I didn't know about it. I guess it would mean that if you have code that uses Gzip a lot, your parallelism might suffer when you are also allocation bottlenecked. I guess that's not an unlikely situation as you will use Gzip often when you have lots of data (e.g. parsing big files). In fact, in some of my personal projects I saw a few cases of inexplicably low CPU saturation during streaming parsing that might be related to this.

Another thing that would be enabled by using jzlib would be the WebSocket compression extension support (#659) where some gzip limits need to be configured that aren't exposed by the JDK.

So, initial support for this would be very welcome and would ideally go into akka-stream directly. It seems jzlib has a Deflate-like interface, so it might almost be a drop-in replacement for the JDK? If there's no good reason to keep the JDK implementation it would then be easy. Otherwise, we'd have to build an interface to support both JDK and jzlib at the same time.

Do you know of any performance benchmarks? (I would expect similar performance, low-level JVM translation might be slightly slower (in some cases) but JNI has it's own overhead).

jrudolph:
@raboof did you get anywhere with this already? I also started a branch a few months ago. The basic blocker I found, is that the implementation classes `Compressor` and `Decompressor` classes are public right now. The other thing is that `Encoder` has strict methods that cannot be implemented by a streams version without blocking.

One suggestion would be:

 * deprecate `Compressor` and `Decompressor` in the 10.1.x branch and remove them right away, it's somewhat unlikely that they have been used by users
 * deprecate methods in Encoder that cannot be nicely implemented anymore and just accept that they will block now (and, of course, make sure that we don't use them any more)

The more conservative approach to the first step would be to deprecate and keep the implementation classes for another version.

jrudolph:
[...]

PR #3294, opened 2020-06-24
http: deprecate more of coding infrastructure

jrudolph:
Refs #2916

Previously, we exposed Scala objects as API directly which exposes implementation
details and makes it hard to evolve the implementation. Instances in coders
are upcasted to `Coder` so that implementation details are not leaked any more.

### DIFF
--- a/akka-http/src/main/scala/akka/http/scaladsl/coding/Coders.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/coding/Coders.scala
@@ -1,0 +1,19 @@
+package akka.http.scaladsl.coding
+
+import akka.http.scaladsl.model.HttpMessage
+import com.github.ghik.silencer.silent
+
+import scala.collection.immutable
+
+@silent("in package coding is deprecated")
+object Coders {
+  def Gzip: Coder = akka.http.scaladsl.coding.Gzip
+  def Gzip(messageFilter: HttpMessage => Boolean): Coder = new Gzip(messageFilter)
+
+  def Deflate: Coder = akka.http.scaladsl.coding.Gzip
+  def Deflate(messageFilter: HttpMessage => Boolean): Coder = new Deflate(messageFilter)
+
+  def NoCoding: Coder = akka.http.scaladsl.coding.NoCoding
+
+  val DefaultCoders: immutable.Seq[Coder] = immutable.Seq(Gzip, Deflate, NoCoding)
+}
--- a/akka-http/src/main/scala/akka/http/scaladsl/coding/Coders.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/coding/Coders.scala
@@ -1,3 +1,7 @@
+/*
+ * Copyright (C) 2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
 package akka.http.scaladsl.coding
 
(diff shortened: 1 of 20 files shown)


ennru:
The doc examples show `Gzip.withLevel(9)` which I silenced for now.

jrudolph:
> The doc examples show `Gzip.withLevel(9)` which I silenced for now.

Thanks, I tried to go through the warnings, but it seems adding a deprecation didn't recompile dependencies so I missed all of those.

akka-http/src/main/scala/akka/http/scaladsl/coding/Coders.scala
```diff
@@ -0,0 +1,23 @@
...
+import com.github.ghik.silencer.silent
+
+import scala.collection.immutable
+
+@silent("in package coding is deprecated")
+object Coders {
+  def Gzip: Coder = akka.http.scaladsl.coding.Gzip
+  def Gzip(messageFilter: HttpMessage => Boolean): Coder = new Gzip(messageFilter)
+
+  def Deflate: Coder = akka.http.scaladsl.coding.Gzip
```
jrudolph: oops

akka-http/src/main/scala/akka/http/scaladsl/coding/Coders.scala
```diff
@@ -0,0 +1,19 @@
+package akka.http.scaladsl.coding
+
+import akka.http.scaladsl.model.HttpMessage
+import com.github.ghik.silencer.silent
+
+import scala.collection.immutable
+
+@silent("in package coding is deprecated")
+object Coders {
```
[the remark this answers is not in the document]
raboof: Good to be consistent with e.g. `StatusCodes` I think, indeed

akka-http/src/main/scala/akka/http/scaladsl/coding/Gzip.scala
```diff
@@ -4,10 +4,13 @@
 
 package akka.http.scaladsl.coding
 
+import akka.annotation.InternalApi
 import akka.http.scaladsl.model._
 import akka.http.scaladsl.model.headers.HttpEncodings
 
-class Gzip private (compressionLevel: Int, val messageFilter: HttpMessage => Boolean) extends Coder with StreamDecoder {
+@InternalApi
+@deprecated("Actual implementation of Gzip is internal, use Coders.Gzip instead", since = "10.2.0")
```
raboof, line 12: It's a bit strange to mark internal API as deprecated, but indeed as it used to be public API I guess it makes sense :+1:

ennru:
I did another round on this to use the non-deprecated variants in the tests.

akka-http/src/main/scala/akka/http/scaladsl/coding/Gzip.scala
jrudolph: Yeah, guess I went over the top with deprecating stuff, just to be sure. Maybe we can create an internal copy later as the future version while keeping that one around only for compatibility.

jrudolph:
> I did another round on this to use the non-deprecated variants in the tests.

Thanks, much appreciated :)

I'm planning to do a more thorough review of warnings and documentation for 10.2.0 next week.

Issue #3287, opened 2020-06-24
Deprecate `headerValueByName(Symbol)`


jrudolph:
There's no good reason to use this over either

 * preferred: `headerValueByType(Referer)` or
 * `headerValueByName("Referer")`

PR #3295, opened 2020-06-24
Deprecate header value extraction with Scala Symbols

ennru:
Refs #3287


ennru:
Oh, it's too warm.
Duplicate ot #3291

PR #3296, opened 2020-06-24
Deprecate scaladsl UpgradeToWebSocket

ennru:
Refs #3278


jrudolph:
Deprecation warning fails docs compilation.

docs/src/main/paradox/server-side/websocket-support.md
```diff
@@ -55,6 +55,12 @@ The WebSocket specification mandates that details of the WebSocket connection ar
 HTTP-headers into request and response of the HTTP upgrade. In Akka HTTP these HTTP-level details of the WebSocket
 handshake are hidden from the application and don't need to be managed manually.
 
+@@@ note
+
+The current implementation with a synthetic headers will be replaced with using request attributes (see [issue #3278](https://github.com/akka/akka-http/issues/3278)).
```
raboof, line 60: Maybe point to the higher-level 'route' section below (that is not expected to change) from here?

Issue #3306, opened 2020-06-26
Avoid/deprecate RouteDirectives.complete variations that use argument adaptation


raboof:
Part of #2738

jrudolph:
Done in #3307

PR #3307, opened 2020-06-26
http: allow 'complete' directive with status without adapted arguments

raboof:
References #3306

### DIFF
--- a/akka-http/src/main/scala/akka/http/scaladsl/server/directives/RouteDirectives.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/server/directives/RouteDirectives.scala
@@ -62,7 +62,7 @@ trait RouteDirectives {
    *
    * @group route
    */
-  def complete[T](status: StatusCode, headers: immutable.Seq[HttpHeader], v: T)(implicit m: ToEntityMarshaller[T]): StandardRoute =
+  def complete[T](status: StatusCode, headers: immutable.Seq[HttpHeader], v: => T)(implicit m: ToEntityMarshaller[T]): StandardRoute =
     complete((status, headers, v))
 
   /**
--- a/akka-http/src/main/scala/akka/http/scaladsl/server/directives/RouteDirectives.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/server/directives/RouteDirectives.scala
@@ -6,8 +6,7 @@ package akka.http.scaladsl.server
 package directives
 
 import scala.concurrent.Future
-
-import akka.http.scaladsl.marshalling.ToResponseMarshallable
+import akka.http.scaladsl.marshalling.{ ToEntityMarshaller, ToResponseMarshallable }
 import akka.http.scaladsl.model._
 import StatusCodes._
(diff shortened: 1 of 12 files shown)


akka-http/src/main/scala/akka/http/scaladsl/server/directives/RouteDirectives.scala
```diff
@@ -49,6 +49,14 @@ trait RouteDirectives {
   def complete(m: => ToResponseMarshallable): StandardRoute =
     StandardRoute(_.complete(m))
 
+  /**
+   * Completes the request using the given arguments.
+   *
+   * @group route
+   */
+  def complete[T](status: StatusCode, m: ToResponseMarshallable): StandardRoute =
```
raboof, line 57: This should probably be
```suggestion
  def complete[T](status: StatusCode, v: T)(implicit marshaller: ToEntityMarshaller[T]): StandardRoute =
```
and then we should put those together so we can call the regular `ctx.complete(ToResponseMarshallable)`

akka-http/src/main/scala/akka/http/scaladsl/server/directives/RouteDirectives.scala
```diff
@@ -49,6 +49,22 @@ trait RouteDirectives {
   def complete(m: => ToResponseMarshallable): StandardRoute =
     StandardRoute(_.complete(m))
 
+  /**
+   * Completes the request using the given arguments.
+   *
+   * @group route
+   */
+  def complete[T](status: StatusCode, v: T)(implicit m: ToEntityMarshaller[T]): StandardRoute =
```
jrudolph, line 57: The main `complete` directive is by-name. Could the strict version lead to surprises?

akka-http/src/main/scala/akka/http/scaladsl/server/directives/RouteDirectives.scala
```diff
@@ -49,6 +49,22 @@ trait RouteDirectives {
...
+   */
+  def complete[T](status: StatusCode, v: => T)(implicit m: ToEntityMarshaller[T]): StandardRoute =
+    StandardRoute(_.complete((status, v)))
+
+  /**
+   * Completes the request using the given arguments.
+   *
+   * @group route
+   */
+  def complete[T](status: StatusCode, headers: immutable.Seq[HttpHeader], v: T)(implicit m: ToEntityMarshaller[T]): StandardRoute =
```
jrudolph, line 65:
```suggestion
  def complete[T](status: StatusCode, headers: immutable.Seq[HttpHeader], v: => T)(implicit m: ToEntityMarshaller[T]): StandardRoute =
```

jrudolph: Incidentally, while I think it's good to have those as by-name for consistency, it's less likely that they are used in the wrong way (compared to the one-argument `complete`). The one argument `complete` could be used like that:


```scala
complete {
  // run some side-effects in here
}
```

but that's less likely with multiple arguments.

Issue #3234, opened 2020-06-10
Update to Scala 2.13.3


raboof:
There were a couple of regressions in Scala 2.13.2.

While it is unclear if any of those would impact the Akka HTTP artifacts, it might be good to be on the safe side and update to 2.13.3 before releasing Akka HTTP 10.2.0

jrudolph:
Done in #3309.

PR #3309, opened 2020-06-26
build: update to Scala 2.13.3

raboof:
Refs #3234

### DIFF
--- a/akka-http-core/src/test/scala/akka/http/impl/engine/parsing/HttpHeaderParserSpec.scala
+++ b/akka-http-core/src/test/scala/akka/http/impl/engine/parsing/HttpHeaderParserSpec.scala
@@ -29,80 +29,80 @@ abstract class HttpHeaderParserSpec(mode: String, newLine: String) extends AkkaS
 ) {
   s"The HttpHeaderParser (mode: $mode)" should {
     "insert the 1st value" in new TestSetup(testSetupMode = TestSetupMode.Unprimed) {
-      insert("Hello", Symbol("Hello"))
+      insert("Hello", "Hello")
       check {
         s"""nodes: 0/H, 0/e, 0/l, 0/l, 0/o, 1/Ω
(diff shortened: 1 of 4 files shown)


akka-http-core/src/test/scala/akka/http/impl/engine/parsing/HttpHeaderParserSpec.scala
```diff
@@ -33,10 +33,10 @@ abstract class HttpHeaderParserSpec(mode: String, newLine: String) extends AkkaS
       check {
         s"""nodes: 0/H, 0/e, 0/l, 0/l, 0/o, 1/Ω
            |branchData:${" " /* explicit trailing space */ }
-           |values: 'Hello""" -> parser.formatRawTrie
+           |values: Symbol(Hello)""" -> parser.formatRawTrie
```
jrudolph, line 36: Would that still work with older Scala versions?

raboof: Ah, of course it wouldn't.

We could:

```suggestion
           |values: ${Symbol(Hello)}""" -> parser.formatRawTrie
```

... but at that point perhaps it'd be better to replace the symbols with regular strings here - WDYT?

jrudolph: Yes, good point. I don't think anything speaks against using strings here.

raboof:
Yes, I released 0.16 for 2.13.3 (https://github.com/lightbend/genjavadoc/issues/239#issuecomment-650186057)

jrudolph:
> Yes, I released 0.16 for 2.13.3 ([lightbend/genjavadoc#239 (comment)](https://github.com/lightbend/genjavadoc/issues/239#issuecomment-650186057))

Great, so we don't even have to update the version :+1:

Issue #2405, opened 2019-01-30
FAILED: EntityStreamingSpec.client-consume-streaming-json


jrudolph:
https://jenkins.akka.io:8498/job/akka-http-nightly/363/AKKA_VERSION=default,Node=akka-http-nightly-node,SCALA_VERSION=2.11.12,jdk=AdoptOpenJDK%2011/testReport/junit/akka.http.scaladsl.server/EntityStreamingSpec/client_consume_streaming_json/

```
Error Message

A timeout occurred waiting for a future to complete. Queried 11 times, sleeping 15 milliseconds between each query.

Stacktrace

org.scalatest.concurrent.Futures$FutureConcept$$anon$1: A timeout occurred waiting for a future to complete. Queried 11 times, sleeping 15 milliseconds between each query.
	at org.scalatest.concurrent.Futures$FutureConcept$class.tryTryAgain$1(Futures.scala:538)
	at org.scalatest.concurrent.Futures$FutureConcept$class.futureValueImpl(Futures.scala:550)
	at org.scalatest.concurrent.ScalaFutures$$anon$1.futureValueImpl(ScalaFutures.scala:275)
	at org.scalatest.concurrent.Futures$FutureConcept$class.futureValue(Futures.scala:476)
	at org.scalatest.concurrent.ScalaFutures$$anon$1.futureValue(ScalaFutures.scala:275)
	at akka.http.scaladsl.server.EntityStreamingSpec$$anonfun$8.apply(EntityStreamingSpec.scala:123)
```

johanandren:
https://jenkins.akka.io:8498/job/akka-http-nightly-release-10.1/AKKA_VERSION=master,Node=akka-http-nightly-node,SCALA_VERSION=2.13.1,jdk=AdoptOpenJDK%208/886/testReport/junit/akka.http.scaladsl.server/EntityStreamingSpec/client_consume_streaming_json/

ennru:
Akka HTTP 10.1
Scala 2.11
JDK 8

https://jenkins.akka.io:8498/job/akka-http-nightly-release-10.1/AKKA_VERSION=default,Node=akka-http-nightly-node,SCALA_VERSION=2.11.12,jdk=AdoptOpenJDK%208/lastCompletedBuild/testReport/akka.http.scaladsl.server/EntityStreamingSpec/client_consume_streaming_json/

PR #3312, opened 2020-06-29
tests: set a more reasonable PatienceConfig for RoutingSpecs

jrudolph:
Otherwise, it will use the scalatest default of 150 millis.

Refs #2405.


raboof:
manually verified travis is green

jrudolph:
> manually verified travis is green

Hehe was just doing the same thing :)

PR #3316, opened 2020-06-29
Upgrade to websocket as attribute

raboof:


akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpRequestParser.scala
```diff
@@ -180,15 +180,17 @@ private[http] final class HttpRequestParser(
...
             else headers
 
-          val allHeaders =
+          val requestStart =
             if (method == HttpMethods.GET) {
               Handshake.Server.websocketUpgrade(headers, hostHeaderPresent, websocketSettings, headerParser.log) match {
-                case OptionVal.Some(upgrade) => upgrade :: allHeaders0
-                case OptionVal.None          => allHeaders0
+                case OptionVal.Some(upgrade) =>
+                  RequestStart(method, uri, protocol, Map(AttributeKeys.webSocketUpgrade -> upgrade), upgrade :: allHeaders0, createEntity, expect100continue, closeAfterResponseCompletion)
```
jrudolph, line 187: So, we would have both the header and the attribute? I guess that's what we want for compatibility?

akka-http-core/src/main/scala/akka/http/scaladsl/model/ws/WebSocketUpgrade.scala
```diff
@@ -0,0 +1,96 @@
...
+import java.lang.Iterable
+import scala.collection.immutable
+import akka.NotUsed
+import akka.stream._
+import akka.http.impl.util.JavaMapping
+import akka.http.javadsl.{ model => jm }
+import akka.http.scaladsl.model.HttpResponse
+
+/**
+ * A custom header that will be added to an WebSocket upgrade HttpRequest that
```
jrudolph, line 16: Scaladoc needs update.

akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpRequestParser.scala
raboof: I think so, yes - we can drop the header in 10.3 I guess?

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -134,6 +134,7 @@ private[http] object HttpServerBluePrint {
 
           val entity = createEntity(entityCreator) withSizeLimit settings.parserSettings.maxContentLength
           val httpRequest = HttpRequest(effectiveMethod, uri, effectiveHeaders, entity, protocol)
+            .withAttributes(attrs)
```
jrudolph, line 137: I guess this might have a small performance impact. In that particular case, we can probably optimize `withAttributes` to be a no-op for `Map.empty` but we should also monitor if we should introduce a private constructor that includes attributes as well to avoid extra copying in cases like this.

jrudolph:
Still marked as draft, @raboof. Anything still missing? Otherwise, merge at your discretion.

raboof:
> Anything still missing?

I think we should also update docs, examples and tests in this PR - but for those we'd want to have the directives for extracting attributes first (#3315)

raboof:
> I think we should also update docs, examples and tests in this PR

That would look like #3358

raboof:
let's merge this as part of #3358

PR #3317, opened 2020-06-29
[WIP] fluent builder API for creating websocket handlers

jrudolph:
This is basically an alternative for #2133 which I had lying around.

It's based on the observation that providing WebSocket handlers on both client and server side has a lot of potential bits and options that are mostly orthogonal. The current API is highly flexible because it accepts a `Flow[Message, Message]` but that flexibility currently leads to a lot of essentially broken code because it's non-trivial to provide implementation for concrete, simpler scenarios in terms of the flexible Flow-based API.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/ws/WebSocketHandler.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/ws/WebSocketHandler.scala
@@ -89,7 +89,7 @@ object WebSocketHandlerExamples {
   implicit def mat: Materializer = ???
   val handler: Flow[Message, Message, Any] =
     Flow[Message]
-      .mapAsync[Seq[Message]](1) {
+      .mapAsync[immutable.Seq[Message]](1) {
         case tm: TextMessage =>
           tm.toStrict(WebSocketHandler.defaultSettings.perMessageTimeout)
             // FIXME: also check maxMessageSize here (or better in toStrict)
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/ws/WebSocketHandler.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/ws/WebSocketHandler.scala
@@ -1,0 +1,120 @@
+package akka.http.scaladsl.model.ws
+
+import akka.actor.ActorRef
+import akka.annotation.ApiMayChange
+import akka.stream.Materializer
+import akka.stream.scaladsl.{ Flow, Sink, Source }
+import akka.util.{ ByteString, Timeout }
+
+import scala.collection.immutable
+import scala.concurrent.{ ExecutionContext, Future }
+import scala.concurrent.duration._
+
+@ApiMayChange
+object WebSocketHandler {
+  // auto-strict variants
+  def forText(): Builder[String] = ???
+  def forBinary(): Builder[ByteString] = ???
+
+  // potentially streaming variants
+  def forTextMessage(): Builder[TextMessage] = ???
+  def forBinaryMessage(): Builder[BinaryMessage] = ???
+
+  trait Builder[T] {
+    def handle(handler: Flow[T, T, Any]): SettingsBuilder
+    def mapMessages(handler: T => T): SettingsBuilder = handle(Flow[T].map(handler))
+    def flatMapMessages(handler: T => immutable.Iterable[T]): SettingsBuilder = handle(Flow[T].mapConcat(handler))
+    def mapMessagesAsync(handler: T => Future[T], parallelism: Int = 1, unordered: Boolean = false): SettingsBuilder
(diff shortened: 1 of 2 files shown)


akka-http-core/src/main/scala/akka/http/scaladsl/model/ws/WebSocketHandler.scala
```diff
@@ -0,0 +1,119 @@
...
+import akka.stream.scaladsl.{ Flow, Sink, Source }
+import akka.util.{ ByteString, Timeout }
+
+import scala.concurrent.{ ExecutionContext, Future }
+import scala.concurrent.duration._
+
+@ApiMayChange
+object WebSocketHandler {
+  // auto-strict variants
+  def forText(): Builder[String] = ???
```
jrudolph, line 19: First choice: support WS binary or text messages

akka-http-core/src/main/scala/akka/http/scaladsl/model/ws/WebSocketHandler.scala
```diff
@@ -0,0 +1,119 @@
...
+  def forBinary(): Builder[ByteString] = ???
+
+  // potentially streaming variants
+  def forTextMessage(): Builder[TextMessage] = ???
+  def forBinaryMessage(): Builder[BinaryMessage] = ???
+
+  trait Builder[T] {
+    def handle(handler: Flow[T, T, Any]): SettingsBuilder
```
jrudolph, line 27: Second choice: what kind of handler do you need?

akka-http-core/src/main/scala/akka/http/scaladsl/model/ws/WebSocketHandler.scala
```diff
@@ -0,0 +1,119 @@
...
+    def flatMapMessages(handler: T => Seq[T]): SettingsBuilder = handle(Flow[T].mapConcat(handler))
+    def mapMessagesAsync(handler: T => Future[T], parallelism: Int = 1, unordered: Boolean = false): SettingsBuilder
+    def flatMapMessagesAsync(handler: T => Future[Seq[T]], parallelism: Int = 1, unordered: Boolean = false): SettingsBuilder
+
+    def sinkAndSource(sink: Sink[T, Any], source: Source[T, Any]): SettingsBuilder
+    def source(source: Source[T, Any]): SettingsBuilder
+    def sink(sink: Sink[T, Any]): SettingsBuilder
+  }
+  trait SettingsBuilder {
+    def withMaxMessageSize(maxSize: Int): this.type
```
jrudolph, line 38: Third choice: what kind of additional settings would you like?

akka-http-core/src/main/scala/akka/http/scaladsl/model/ws/WebSocketHandler.scala
```diff
@@ -0,0 +1,119 @@
...
+    def withPerMessageTimeout(timeout: FiniteDuration): this.type
+    def withHalfClosedConnections(): this.type
+    def withIgnoreWrongMessageType(): this.type
+
+    def build(): Flow[Message, Message, Any]
```
jrudolph, line 43: Finally, a Flow is served to be put into the WS APIs.

akka-http-core/src/main/scala/akka/http/scaladsl/model/ws/WebSocketHandler.scala
```diff
@@ -0,0 +1,119 @@
...
+    override def build(): Flow[Message, Message, Any] = builder(settings)
+  }
+}
+
+object WebSocketHandlerExamples {
+  // simple text-based Request / Response pattern:
+  // synchronous
+
+  val strictHandler: String => String = txt => s"echo: $txt"
+  WebSocketHandler
```
jrudolph, line 81: Here's a simple example for a `map(String => String)` handler. Below is how users would have to write it themselves right now.

akka-http-core/src/main/scala/akka/http/scaladsl/model/ws/WebSocketHandler.scala
```diff
@@ -0,0 +1,119 @@
...
+  val handler: Flow[Message, Message, Any] =
+    Flow[Message]
+      .mapAsync[Seq[Message]](1) {
+        case tm: TextMessage =>
+          tm.toStrict(WebSocketHandler.defaultSettings.perMessageTimeout)
+            // FIXME: also check maxMessageSize here (or better in toStrict)
+            .map(t => TextMessage(strictHandler(t.text)) :: Nil)
+
+        // if ignoreWrongMessageType is false, which is currently the default
+        case _: BinaryMessage => Future.failed(new IllegalArgumentException("Unexpected binary WS message"))
```
jrudolph, line 98: If would want to ignore BinaryMessages instead, it would have to drain the data stream here. A common un-pattern to use `.collect` leads to leaked Sources and temporarily stalled streams when messages are ignored.

PR #3321, opened 2020-06-30
core: disable server-side HTTP pipelining by default

jrudolph:
It's somewhat of an obscure feature which doesn't seem to be tested well.

### DIFF
--- a/akka-http-core/src/main/resources/reference.conf
+++ b/akka-http-core/src/main/resources/reference.conf
@@ -95,7 +95,7 @@ akka.http {
     # connection can be "open" (i.e. being processed by the application) at any
     # time. Set to higher values to enable HTTP pipelining.
     # This value must be > 0 and <= 1024.
-    pipelining-limit = 16
+    pipelining-limit = 1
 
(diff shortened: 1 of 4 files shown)


jrudolph:
It's not about [persistence](https://tools.ietf.org/html/rfc7230#section-6.3) but about [pipelining](https://tools.ietf.org/html/rfc7230#section-6.3.2). Pipelining helps only with the latency of sending over requests because responses still need to be delivered in order.

The only place where I see it used in "practice" is in benchmarks where sending over multiple requests in one go helps with avoiding system calls.

> Wouldn't performance suffer because this would cause many more TCP (and perhaps TLS) handshakes in scenario's where there's a lot of small requests (and perhaps head-of-line blocking isn't such an issue)? Or am I thinking of a different 'pipelining' feature there?

This scenario would probably benefit from pipelining but you will first have to find a client which supports it.

After all we are just following the way that curl went: https://daniel.haxx.se/blog/2019/04/06/curl-says-bye-bye-to-pipelining/

Issue #3326, opened 2020-07-01
Optimize HostConnectionPoolSetup fields


jrudolph:
It currently has `host` and `port` fields. PoolInterface later needs a `Uri.Host` instance instead which leads to parsing the host string again. The `HostConnectionPoolSetup` is usually created from a `Uri` in the first place, so we could just use the `Uri.Host` from the original uri instead to avoid the extra per-request parsing.

Issue #3325, opened 2020-07-01
Optimize NewHostConnectionPool hasIdleSlots


jrudolph:
To avoid excessive scanning as hinted in an existing TODO:

https://github.com/akka/akka-http/blob/6934ba1353fe72475b48c1843b070a58412a330e/akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/NewHostConnectionPool.scala#L105-L107

jrudolph:
Fixed in #3357.

Issue #3324, opened 2020-07-01
Expensive NewHostConnectionPool debug logging


jrudolph:
Debug logs should be better guarded behind `isDebugEnabled` conditions (or be using the existing macro).

PR #3331, opened 2020-07-01
bench: add connection pool benchmark

jrudolph:
Showcases #3324, #3325, and #3326.

```
[info] Benchmark                              (maxConnections)   Mode  Cnt      Score       Error  Units
[info] ConnectionPoolBenchmark.singleRequest                 1  thrpt    3  52257.195 ± 49604.377  ops/s
[info] ConnectionPoolBenchmark.singleRequest                10  thrpt    3  51110.097 ± 32053.322  ops/s
[info] ConnectionPoolBenchmark.singleRequest               100  thrpt    3  37155.430 ±  6851.042  ops/s
[info] ConnectionPoolBenchmark.singleRequest              1000  thrpt    3  12741.234 ±  6250.936  ops/s
[info] ConnectionPoolBenchmark.singleRequest             10000  thrpt    3   5236.162 ±  9153.993  ops/s
```

So, the pool scales badly with higher numbers of concurrent connections. Most of the slowdown is due to various linear scanning to find idle connections (#3325).

### DIFF
--- a/akka-http-bench-jmh/src/main/scala/akka/http/impl/engine/ConnectionPoolBenchmark.scala
+++ b/akka-http-bench-jmh/src/main/scala/akka/http/impl/engine/ConnectionPoolBenchmark.scala
@@ -1,0 +1,107 @@
+package akka.http.impl.engine
+
+import java.net.InetSocketAddress
+import java.util.concurrent.CountDownLatch
+
+import akka.actor.ActorSystem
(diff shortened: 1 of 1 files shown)


jrudolph:
@raboof would be nice to merge that soonish, so that we have some baseline numbers before I try to fix the worst offenders here.

PR #3344, opened 2020-07-06
Fix compiler warnings

ennru:
Most warnings were "Auto-application to `()` is deprecated."
Very few "procedure syntax is deprecated".

### DIFF
--- a/akka-http-tests/src/test/scala/akka/http/scaladsl/marshalling/MarshallingSpec.scala
+++ b/akka-http-tests/src/test/scala/akka/http/scaladsl/marshalling/MarshallingSpec.scala
@@ -106,31 +106,31 @@ class MarshallingSpec extends AnyFreeSpec with Matchers with BeforeAndAfterAll w
     "multipartMarshaller should correctly marshal multipart content with" - {
       "no parts" in {
         marshal(Multipart.General(`multipart/mixed`)) shouldEqual HttpEntity(
-          contentType = (`multipart/mixed` withBoundary randomBoundary()).toContentType,
+          contentType = (`multipart/mixed` withBoundary randomBoundaryValue).toContentType,
           data = ByteString(s"""
-                      |--${randomBoundary()}--""".stripMarginWithNewline("\r\n")))
+                      |--$randomBoundaryValue--""".stripMarginWithNewline("\r\n")))
       }
       "one empty part" in {
         marshal(Multipart.General(`multipart/mixed`, Multipart.General.BodyPart.Strict(""))) shouldEqual HttpEntity(
-          contentType = (`multipart/mixed` withBoundary randomBoundary()).toContentType,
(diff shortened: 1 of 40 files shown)


akka-http-tests/src/test/scala/akka/http/scaladsl/marshalling/MarshallingSpec.scala
```diff
@@ -106,31 +106,31 @@ class MarshallingSpec extends AnyFreeSpec with Matchers with BeforeAndAfterAll w
     "multipartMarshaller should correctly marshal multipart content with" - {
       "no parts" in {
         marshal(Multipart.General(`multipart/mixed`)) shouldEqual HttpEntity(
-          contentType = (`multipart/mixed` withBoundary randomBoundary).toContentType,
+          contentType = (`multipart/mixed` withBoundary randomBoundary()).toContentType,
```
raboof, line 109: In MarshallingSpec, we override `multipartBoundaryRandom` to a fixed value, so `randomBoundary` actually returns a constant (which AFAICS this test relies on).

Wouldn't it be clearer to, instead of overriding `multipartBoundaryRandom`, just call `randomBoundary` once and use that value twice in the test?

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/DebuggingDirectivesSpec.scala
```diff
@@ -43,7 +43,7 @@ class DebuggingDirectivesSpec extends RoutingSpec {
       resetDebugMsg()
       Get("/hello") ~> route ~> check {
         response shouldEqual Ok
-        normalizedDebugMsg shouldEqual "1: HttpRequest(HttpMethod(GET),http://example.com/hello,List(),HttpEntity.Strict(none/none,0 bytes total),HttpProtocol(HTTP/1.1))\n"
+        normalizedDebugMsg() shouldEqual "1: HttpRequest(HttpMethod(GET),http://example.com/hello,List(),HttpEntity.Strict(none/none,0 bytes total),HttpProtocol(HTTP/1.1))\n"
```
jrudolph, line 46: Shouldn't have parentheses in the first place.

raboof: You could say it needs parens since its return value depends on a `var`, but indeed debatable

jrudolph: Ah, good catch, then let's just leave it like that.

ennru: Yes, I looked into that and kept it. Changed it after @jrudolph's comment, and back now...

PR #3347, opened 2020-07-06
core: keep request/response copy method public

jrudolph:
Otherwise, it would be source incompatible in a central piece of the model.
We usually prefer the `withXYZ` methods but since the case class style has
been around for such a long time maybe we should also keep the copy method
around until it becomes more of a problem.

The other alternative would be to deprecate the methods and point to the
`withXYZ` methods more prominently.

(And yes, I probably argued for the existing state at some point...)

See also https://github.com/akka/alpakka/issues/2355

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
@@ -306,7 +306,7 @@ final class HttpRequest(
     "HTTP/1.0 requests must not have a chunked entity")
 
   type Self = HttpRequest
-  def self = this
+  def self: Self = this
 
   override def isRequest = true
   override def isResponse = false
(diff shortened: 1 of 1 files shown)


ennru:
I think deprecation and pointing to `withXyz` would be good so that they might be hidden/dropped later.

jrudolph:
> I think deprecation and pointing to `withXyz` would be good so that they might be hidden/dropped later.

Can we find a conclusion about this? 

Pros for keeping it a public method:

 * no visible change on a central API (we don't know exactly how common `copy` is in user code)
 * some evolution is still possible:
   * adding parameters

Cons:
 
 * further evolution requires to keep copy working as is, not allowed are
   * reordering of parameters
   * removal of parameters
   * renaming of parameters

I don't see a situation coming where the disadvantages become a blocker, so I'd currently vote just to keep things as they are and defer the more intrusive changes to a time when such a change is foreseeable. WDYT?

(Incidentally, the most intrusive change I currently have in mind to do eventually is about the internal representation of headers. Two things could make sense: transparent lazy parsing of headers in general, special representation of some headers for performance reasons. As long as those are internal, they shouldn't affect the public APIs.)

ennru:
I favour keeping it public but adding a deprecation.

jrudolph:
Now deprecated the copy methods.

PR #3348, opened 2020-07-06
Fix some more warnings

jrudolph:


akka-http/src/main/scala/akka/http/scaladsl/server/PathMatcher.scala
```diff
@@ -48,7 +48,7 @@ abstract class PathMatcher[L](implicit val ev: Tuple[L]) extends (Path => PathMa
   }
 
   /** Operator alternative to [[PathMatchers.not]] */
-  def unary_!(): PathMatcher0 = PathMatchers.not(self)
+  def unary_! : PathMatcher0 = PathMatchers.not(self)
```
raboof, line 51: ha ;)

jrudolph: Btw. needs that extra space to compile, I guess that might be a reason the parentheses were added in the first place.

PR #3350, opened 2020-07-06
build: fail on JavaConverter warnings in docs

ennru:


build.sbt
```diff
@@ -15,8 +15,6 @@ inThisBuild(Def.settings(
   organizationName := "Lightbend",
   organizationHomepage := Some(url("https://www.lightbend.com")),
   homepage := Some(url("https://akka.io")),
-  // https://github.com/dwijnand/sbt-dynver/issues/23
-  isSnapshot :=  { isSnapshot.value || hasCommitsAfterTag(dynverGitDescribeOutput.value) },
```
ennru, line 19: This issue was fixed way back...

ennru:
> It's not needed any more because the same ignore rule is already in project/Common.scala.

That explains it ofc.

Issue #3342, opened 2020-07-03
[10.2.0-RC1] Http().singleRequest() is sending query params cached from the last request


mushtaq:
**Steps**

- clone repo: https://github.com/mushtaq/akka-http-client-bug
- run QuickstartApp
- run test UserRoutesSpec
- server log prints 
```txt
http://127.0.0.1:8080/users?id=100
http://127.0.0.1:8080/users?id=100
```

**Expected**

- server should print

```txt
http://127.0.0.1:8080/users?id=100
http://127.0.0.1:8080/users
```

raboof:
Thanks for the report and reproducer! That indeed looks wrong.

Looking with wireshark, it seems actually the client is sending the correct request, it is the server that seems to behave incorrectly.

Indeed the problem appears to be introduced between M1 and RC1.

We will look into this!

raboof:
The offending change is d3b1322d4526171370935ec7b4648bc45fabc672 . I have to go now, but this is definitely something we want to fix ;)

jrudolph:
Ugh, indeed!

jrudolph:
Fix in #3352, embarrassing that I didn't think about those other fields at all. Mutable state, what could go wrong...

mushtaq:
> Fix in #3352, embarrassing that I didn't think about those other fields at all. Mutable state, what could go wrong...

Mutable state is dangerous.... especially for people who have become too comfortable using the immutable data structures 😄

jrudolph:
Fixed in #3352.

PR #3352, opened 2020-07-06
core: make sure to reset all UriParser fields before reusing

jrudolph:
Refs #3342


jrudolph:
> This makes the optimization of keeping one instance a bit less attractive, but this is not something that changes a lot so OK :)

Yeah, indeed. I introduced the `reset` call for that reason that it is easier to keep track of what code is trying to reuse it in the first place (and it's only the HttpRequestParser so far).

PR #3357, opened 2020-07-07
Track idle slots explicitly to avoid O(n^2) work for big pools

jrudolph:
Before linear scanning for idle slots was the number one hot spot for pools.
In experiments that showed for pools of size > 500.

Refs #3325

```
Before:

Benchmark                              (maxConnections)   Mode  Cnt      Score      Error  Units
ConnectionPoolBenchmark.singleRequest              1000  thrpt    3  10459.168 ±  948.674  ops/s
ConnectionPoolBenchmark.singleRequest             10000  thrpt    3   4302.828 ± 1031.308  ops/s

After:

Benchmark                              (maxConnections)   Mode  Cnt      Score       Error  Units
ConnectionPoolBenchmark.singleRequest              1000  thrpt    3  28747.105 ± 44447.131  ops/s
ConnectionPoolBenchmark.singleRequest             10000  thrpt    3  29018.001 ±  2771.695  ops/s
```

I.e.
~ 3x speedup for a size of 1000 connections
~ 7x speedup for a size of 10000 connections

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/NewHostConnectionPool.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/NewHostConnectionPool.scala
@@ -23,10 +23,10 @@ import akka.stream.scaladsl.{ Flow, Keep, Sink, Source }
 import akka.stream.stage.{ GraphStage, GraphStageLogic, InHandler, OutHandler }
 import akka.util.OptionVal
 
+import scala.collection.JavaConverters._
 import scala.concurrent.Future
 import scala.concurrent.duration._
-import scala.util.control.NoStackTrace
-import scala.util.control.NonFatal
+import scala.util.control.{ NoStackTrace, NonFatal }
 import scala.util.{ Failure, Random, Success, Try }
 
 /**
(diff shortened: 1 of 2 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/NewHostConnectionPool.scala
```diff
@@ -114,9 +133,8 @@ private[client] object NewHostConnectionPool {
             push(responsesOut, ResponseContext(req, result))
 
         def dispatchRequest(req: RequestContext): Unit = {
-          val slot =
-            slots.find(_.isIdle)
-              .getOrElse(throw new IllegalStateException("Tried to dispatch request when no slot is idle"))
+          val slot = idleSlots.first()
```
ennru, line 136: Are we sure there is an idle slot when we get here?

jrudolph: Yes, we should. In any case, we threw before as well.

jrudolph:
See https://jenkins.akka.io:8498/job/akka-http-bench-adhoc/177/jmh-run-report/ for a comparison between current master and a branch with #3356 and #3357 applied. 20-30% improvement for small pools and 5x-8x improvement for big ones.

akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/NewHostConnectionPool.scala
```diff
@@ -102,9 +115,15 @@ private[client] object NewHostConnectionPool {
...
 
-        def hasIdleSlots: Boolean =
-          // TODO: optimize by keeping track of idle connections?
-          slots.exists(_.isIdle)
+        def hasIdleSlots: Boolean = {
+          if (log.isDebugEnabled) { // somewhat sneaky way of enabling extra assertions in "debug-mode"
+            // Helps debugging if you suspect that idleSlots are not consistent with actual state any more
+            val idle = idleSlots.asScala.map(_.slotId).toSet
+            val idleAll = slots.filter(_.isIdle).map(_.slotId).toSet
+            require(idle == idleAll, s"Managed idle [${idle.mkString(", ")}] != real idle [${idleAll.mkString(", ")}]")
```
raboof, line 123: Should we make a note to eventually remove this check?

jrudolph: The question is when. Right now there probably are no bugs. The check helps especially when after you made any changes.

jrudolph: Maybe the right way to do would be to create a real option to enable more assertions and enable it when running tests.

Issue #3241, opened 2020-06-11
`discardEntityBytes` should allow `ClassicActorSystemProvider`


ennru:
`discardEntityBytes` should be usable without a materializer and allow usage with the system's materializer instead.

https://github.com/akka/akka-http/blob/6b79bfec6e536da20fa6fb3d8e15a9aa75a35540/akka-http-core/src/main/java/akka/http/javadsl/model/HttpMessage.java#L100

jrudolph:
Just as a clarification: this is for the java-side where there is no implicit doing that conversion from `ClassicActorSystemProvider` to `Materializer`.

jrudolph:
Fixed in #3361

Issue #3240, opened 2020-06-11
`Unmarshaller` should accept `ClassicActorSystemProvider` as alternative to `Materializer`


ennru:
The `Unmarshaller`s should be able to work without a materializer at hand.

https://github.com/akka/akka-http/blob/6b79bfec6e536da20fa6fb3d8e15a9aa75a35540/akka-http/src/main/scala/akka/http/javadsl/unmarshalling/Unmarshaller.scala#L121-L130

jrudolph:
Fixed in #3361

PR #3361, opened 2020-07-08
Accept classic provider in more places

jrudolph:
Refs #3240, #3241


jrudolph:
Added a few more changes to remove Materializer from docs examples

PR #3362, opened 2020-07-08
http: new ServerBuilder API

jrudolph:
The Javadsl-side of the Http entrypoints was always a bit unpolished. Especially the need to create a `ConnectHttp.toHost` to bind a server seemed quite wild.

In the process of finding simpler ways to bind servers on the Java-side, I created a little fluent builder to ease the creation of server and find something against the sprouting variety of overloads in javadsl.Http over time.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/javadsl/Http.scala
+++ b/akka-http-core/src/main/scala/akka/http/javadsl/Http.scala
@@ -83,6 +83,14 @@ class Http(system: ExtendedActorSystem) extends akka.actor.Extension {
     adaptServerLayer(delegate.serverLayer(settings.asScala, remoteAddress.asScala, log))
 
   /**
+   * Main entry point to create a server binding.
+   *
+   * @param interface The interface to bind to.
+   * @param port The port to bind to or `0` if the port should be automatically assigned.
+   */
+  def newServerAt(interface: String, port: Int): ServerBuilder = ServerBuilder(interface, port, system)
+
+  /**
    * Creates a [[akka.stream.javadsl.Source]] of [[IncomingConnection]] instances which represents a prospective HTTP server binding
    * on the given `endpoint`.
    *
@@ -96,7 +104,11 @@ class Http(system: ExtendedActorSystem) extends akka.actor.Extension {
    *
    * The server will be bound using HTTPS if the [[ConnectHttp]] object is configured with an [[HttpsConnectionContext]],
    * or the [[defaultServerHttpContext]] has been configured to be an [[HttpsConnectionContext]].
+   *
+   * @deprecated since 10.2.0: Use Http.get(system).newServerAt(interface, port).bind() instead
    */
+  @Deprecated
+  @deprecated("Use newServerAt instead", since = "10.2.0")
   def bind(connect: ConnectHttp): Source[IncomingConnection, CompletionStage[ServerBinding]] = {
     val connectionContext = connect.effectiveConnectionContext(defaultServerHttpContext).asScala
     new Source(delegate.bind(connect.host, connect.port, connectionContext)
@@ -118,7 +130,11 @@ class Http(system: ExtendedActorSystem) extends akka.actor.Extension {
    *
    * The server will be bound using HTTPS if the [[ConnectHttp]] object is configured with an [[HttpsConnectionContext]],
    * or the [[defaultServerHttpContext]] has been configured to be an [[HttpsConnectionContext]].
+   *
+   * @deprecated since 10.2.0: Use Http.get(system).newServerAt(interface, port).withSettings(settings).bind() instead
    */
+  @Deprecated
+  @deprecated("Use newServerAt instead", since = "10.2.0")
   def bind(
     connect:  ConnectHttp,
     settings: ServerSettings): Source[IncomingConnection, CompletionStage[ServerBinding]] = {
@@ -142,7 +158,11 @@ class Http(system: ExtendedActorSystem) extends akka.actor.Extension {
    *
    * The server will be bound using HTTPS if the [[ConnectHttp]] object is configured with an [[HttpsConnectionContext]],
    * or the [[defaultServerHttpContext]] has been configured to be an [[HttpsConnectionContext]].
+   *
+   * @deprecated since 10.2.0: Use Http.get(system).newServerAt(interface, port).withSettings(settings).logTo(log).bind() instead
    */
+  @Deprecated
+  @deprecated("Use newServerAt instead", since = "10.2.0")
   def bind(
     connect:  ConnectHttp,
     settings: ServerSettings,
@@ -163,7 +183,11 @@ class Http(system: ExtendedActorSystem) extends akka.actor.Extension {
    *
    * The server will be bound using HTTPS if the [[ConnectHttp]] object is configured with an [[HttpsConnectionContext]],
    * or the [[defaultServerHttpContext]] has been configured to be an [[HttpsConnectionContext]].
+   *
+   * @deprecated since 10.2.0: Use Http.get(system).newServerAt(interface, port).bindFlow(handler) instead.
    */
+  @Deprecated
+  @deprecated("Use newServerAt instead", since = "10.2.0")
   def bindAndHandle(
     handler:      Flow[HttpRequest, HttpResponse, _],
     connect:      ConnectHttp,
@@ -185,29 +209,11 @@ class Http(system: ExtendedActorSystem) extends akka.actor.Extension {
    *
    * The server will be bound using HTTPS if the [[ConnectHttp]] object is configured with an [[HttpsConnectionContext]],
    * or the [[defaultServerHttpContext]] has been configured to be an [[HttpsConnectionContext]].
-   */
-  def bindAndHandle(
-    handler: Flow[HttpRequest, HttpResponse, _],
-    connect: ConnectHttp,
-    system:  ClassicActorSystemProvider): CompletionStage[ServerBinding] = {
-    val connectionContext = connect.effectiveConnectionContext(defaultServerHttpContext).asScala
-    delegate.bindAndHandle(
-      handler.asInstanceOf[Flow[sm.HttpRequest, sm.HttpResponse, _]].asScala,
-      connect.host, connect.port, connectionContext)(Materializer.matFromSystem(system))
-      .map(new ServerBinding(_))(ec).toJava
-  }
-
-  /**
-   * Convenience method which starts a new HTTP server at the given endpoint and uses the given `handler`
-   * [[akka.stream.javadsl.Flow]] for processing all incoming connections.
    *
-   * The number of concurrently accepted connections can be configured by overriding
-   * the `akka.http.server.max-connections` setting. Please see the documentation in the reference.conf for more
-   * information about what kind of guarantees to expect.
-   *
-   * The server will be bound using HTTPS if the [[ConnectHttp]] object is configured with an [[HttpsConnectionContext]],
-   * or the [[defaultServerHttpContext]] has been configured to be an [[HttpsConnectionContext]].
+   * @deprecated since 10.2.0: Use Http.get(system).newServerAt(interface, port).withSettings(settings).logTo(log).bindFlow(handler) instead.
    */
+  @Deprecated
+  @deprecated("Use newServerAt instead", since = "10.2.0")
   def bindAndHandle(
     handler:      Flow[HttpRequest, HttpResponse, _],
     connect:      ConnectHttp,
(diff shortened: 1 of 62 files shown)


akka-http-core/src/main/scala/akka/http/javadsl/ServerBuilder.scala
```diff
@@ -0,0 +1,81 @@
...
+import akka.http.javadsl.settings.ServerSettings
+import akka.http.scaladsl
+import akka.http.scaladsl.{ model => sm }
+import akka.japi.function.Function
+import akka.stream.javadsl.Flow
+
+import scala.compat.java8.FutureConverters._
+import scala.concurrent.ExecutionContext
+
+trait HandlerProvider {
```
jrudolph, line 19: Smallest common denominator between akka-http-core and akka-http to allow passing a route directly into `bind` without needing to know about `Route` here.

akka-http-core/src/main/scala/akka/http/javadsl/ServerBuilder.scala
```diff
@@ -0,0 +1,81 @@
...
+  def handler(system: ClassicActorSystemProvider): Function[HttpRequest, CompletionStage[HttpResponse]]
+}
+
+/**
+ * The new API to create server bindings
+ */
+@ApiMayChange
```
jrudolph, line 26: New main server building API for the java side.

akka-http-tests/src/test/java/akka/http/javadsl/server/JavaTestServer.java
```diff
@@ -142,11 +139,11 @@ public static void main(String[] args) throws InterruptedException {
 
   private void run() throws InterruptedException {
     final ActorSystem system = ActorSystem.create();
-    final ActorMaterializer mat = ActorMaterializer.create(system);
 
-    final Flow<HttpRequest, HttpResponse, ?> flow = createRoute().flow(system, mat);
     final CompletionStage<ServerBinding> binding =
-      Http.get(system).bindAndHandle(flow, ConnectHttp.toHost("127.0.0.1", 8080), mat);
+      Http.get(system)
```
jrudolph, line 144: Look ma, no materializer necessary, no double usages of system, no `ConnectHttp`.

akka-http/src/main/scala/akka/http/javadsl/server/Route.scala
```diff
@@ -54,7 +55,8 @@ trait Route {
   def flow(system: ClassicActorSystemProvider): Flow[HttpRequest, HttpResponse, NotUsed] =
     flow(system.classicSystem, SystemMaterializer(system).materializer)
 
-  def function(system: ClassicActorSystemProvider): Function[HttpRequest, CompletionStage[HttpResponse]]
+  def function(system: ClassicActorSystemProvider): Function[HttpRequest, CompletionStage[HttpResponse]] = handler(system)
```
jrudolph, line 58: `function` was only recently introduced, if we go with this, we can probably just remove it (or keep only one of `function` or `handler`.

akka-http-core/src/main/scala/akka/http/javadsl/Http.scala
```diff
@@ -153,6 +153,8 @@ class Http(system: ExtendedActorSystem) extends akka.actor.Extension {
       .mapMaterializedValue(_.map(new ServerBinding(_))(ec).toJava))
   }
 
+  def newServerAt(interface: String): ServerBuilder = ServerBuilder(interface, system)
```
ennru, line 156: I guess a `newServerAt(interface: String, port: Int)` overload would make sense.

akka-http-core/src/main/scala/akka/http/scaladsl/ServerBuilder.scala
```diff
@@ -0,0 +1,71 @@
...
+ */
+trait ServerBuilder {
+  def onPort(port: Int): ServerBuilder
+  def logTo(log: LoggingAdapter): ServerBuilder
+  def settings(settings: ServerSettings): ServerBuilder
+  def adaptSettings(f: ServerSettings => ServerSettings): ServerBuilder
+  def enableHttps(context: HttpsConnectionContext): ServerBuilder
+
+  def bind(): Source[Http.IncomingConnection, Future[ServerBinding]]
+  def bindFlow(handlerFlow: Flow[HttpRequest, HttpResponse, _]): Future[ServerBinding]
```
jrudolph, line 28: This cannot be an overload called `bind` because otherwise, it's ambiguous for routes for conversions are defined both for flow and function (not a blocker for me, we want to deemphasize the flow variant anyway).

akka-http-core/src/main/scala/akka/http/scaladsl/ServerBuilder.scala
```diff
@@ -0,0 +1,71 @@
...
+
+import scala.concurrent.Future
+
+/**
+ * The new API to create server bindings
+ */
+trait ServerBuilder {
+  def onPort(port: Int): ServerBuilder
+  def logTo(log: LoggingAdapter): ServerBuilder
+  def settings(settings: ServerSettings): ServerBuilder
```
jrudolph, line 23: `settings` or `withSettings`?

docs/src/test/scala/docs/http/scaladsl/server/WebSocketExampleSpec.scala
```diff
@@ -122,14 +122,10 @@ class WebSocketExampleSpec extends AnyWordSpec with Matchers with CompileOnlySpe
...
-    val customWebsocketSettings =
-      defaultSettings.websocketSettings
-        .withPeriodicKeepAliveData(() => ByteString(s"debug-${pingCounter.incrementAndGet()}"))
-
-    val customServerSettings =
-      defaultSettings.withWebsocketSettings(customWebsocketSettings)
 
-    Http().bindAndHandle(route, "127.0.0.1", settings = customServerSettings)
+    Http().newServerAt("127.0.0.1")
+      .adaptSettings(_.mapWebsocketSettings(_.withPeriodicKeepAliveData(() => ByteString(s"debug-${pingCounter.incrementAndGet()}"))))
```
jrudolph, line 127: That's not super nice. We could introduce first-class lenses for those at some point. Or we could at least add an easy way to add a string based config override (which wouldn't help here).

jrudolph:
Seeing all the deprecation warnings this PR produced, I went ahead and wrote a scalafix rule to fix those usages. For now I included it into the docs module directly but in the end it should go into its own module or even a separate project.

akka-http-core/src/main/scala/akka/http/javadsl/ServerBuilder.scala
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * The new API to create server bindings
+ */
+trait ServerBuilder {
+  def onPort(port: Int): ServerBuilder
+  def logTo(log: LoggingAdapter): ServerBuilder
+  def settings(settings: ServerSettings): ServerBuilder
+  def adaptSettings(f: Function[ServerSettings, ServerSettings]): ServerBuilder
+  def enableHttps(context: HttpsConnectionContext): ServerBuilder
```
raboof, line 35: Do we also want an `enableHttps()` that enables HTTPS with the default context?

akka-http-core/src/main/scala/akka/http/javadsl/ServerBuilder.scala
raboof: Perhaps clarify in the javadoc that `Route` is the typically used implementation?

akka-http-core/src/test/scala/akka/http/scaladsl/TestServer.scala
```diff
@@ -38,7 +38,7 @@ object TestServer extends App {
     .withInputBuffer(128, 128)
   implicit val fm = ActorMaterializer(settings)
   try {
-    val binding = Http().bindAndHandleSync({
+    val binding = Http().newServerAt("localhost", 9001).bindSync {
```
raboof, line 41: This is actually a lot nicer to read as well ;)

akka-http-core/src/main/scala/akka/http/javadsl/Http.scala
raboof: wouldn't `onPort` be sufficient?

raboof: (arguably we could have `newServer` bind to localhost and have `onAddress` ;) - I'd say this is fine though)

docs/src/test/scala/docs/http/scaladsl/server/WebSocketExampleSpec.scala
raboof: (agreed, though it's not much worse than before)

docs/src/test/scala/docs/http/scaladsl/server/ExceptionHandlerExamplesSpec.scala
```diff
@@ -39,7 +39,7 @@ object MyExplicitExceptionHandler {
         null // #hide
       }
 
-    Http().bindAndHandle(route, "localhost", 8080)
+    Http().newServerAt("localhost", 8080).bindFlow(route)
```
raboof, line 42: Coulnd't this now simply be `.bind()`?

docs/src/scalafix/scala/BindToServerBuilderApi.scala
```diff
@@ -0,0 +1,72 @@
...
+        Patch.replaceTree(t, s"${builder(argExps)}.bind(${argExps("handler")})")
+
+      case t@q"Http().bindAndHandle(..$params)" =>
+
+        val args = Seq("handler", "interface", "port", "connectionContext", "settings", "log")
+        val argExps = namedArgMap(args, params)
+
+        val handler = argExps("handler")
+        handler.synthetics match {
+          case ApplyTree(fun, _) :: Nil if fun.symbol.exists(_.displayName == "routeToFlow") /* FIXME: could use real symbol instead */ =>
```
raboof, line 25: wow!

akka-http-core/src/main/scala/akka/http/javadsl/ServerBuilder.scala
jrudolph: The default context does not work as far as I know because it cannot contain any reasonable key material (and if only through ssl-config, which we'd like to get rid of).

akka-http-core/src/main/scala/akka/http/javadsl/Http.scala
jrudolph: It would be somewhat more consistent to only have one way of doing things.

My idea of why to add another overload having the port was that you will basically always use that one. The situation where you don't want to specify a port is almost exclusively from tests. So, maybe we should instead remove this original overload and only have one with interface and port, so you'd have to be explicit about these at least.

akka-http-core/src/main/scala/akka/http/javadsl/ServerBuilder.scala
jrudolph: Done now also in the docs for `bind(HandlerProvider)`.

akka-http-core/src/main/scala/akka/http/scaladsl/ServerBuilder.scala
jrudolph: Settled now on `withSettings`.

akka-http-core/src/main/scala/akka/http/javadsl/Http.scala
```diff
@@ -185,29 +209,11 @@ class Http(system: ExtendedActorSystem) extends akka.actor.Extension {
    *
    * The server will be bound using HTTPS if the [[ConnectHttp]] object is configured with an [[HttpsConnectionContext]],
    * or the [[defaultServerHttpContext]] has been configured to be an [[HttpsConnectionContext]].
-   */
-  def bindAndHandle(
```
jrudolph, line 189: These were introduced for 10.2.0 but are not necessary any more, so I removed them directly.

jrudolph:
This is now ready from my side. I squashed everything into two commits for easier review consumption: one with the new API and one with all the necessary test changes.

jrudolph:
(I removed the scalafix rule for now)

docs/src/test/scala/docs/http/scaladsl/server/ExceptionHandlerExamplesSpec.scala
jrudolph: Yep, fixed!

akka-http-core/src/main/scala/akka/http/javadsl/ServerBuilder.scala
raboof: of course :+1:

akka-http-core/src/main/scala/akka/http/javadsl/ServerBuilder.scala
```diff
@@ -0,0 +1,180 @@
...
+   * If no `port` is explicitly given (or the port value is negative) the protocol's default port will be used,
+   * which is 80 for HTTP and 443 for HTTPS.
+   */
+  def bind(): Source[IncomingConnection, CompletionStage[ServerBinding]]
+}
+
+/**
+ * A HandlerProvider can provide an asynchronous request handler given an ClassicActorSystemProvider.
+ *
+ * The main use case for this class is to enable passing a Route to ServerBuilder.bind().
```
raboof, line 126: :+1:

akka-http-core/src/main/scala/akka/http/javadsl/Http.scala
```diff
@@ -82,6 +82,14 @@ class Http(system: ExtendedActorSystem) extends akka.actor.Extension {
     log:           LoggingAdapter): BidiFlow[HttpResponse, SslTlsOutbound, SslTlsInbound, HttpRequest, NotUsed] =
     adaptServerLayer(delegate.serverLayer(settings.asScala, remoteAddress.asScala, log))
 
+  /**
+   * Main entry point to create a server binding.
+   *
+   * @param interface The interface to bind to.
+   * @param port The port to bind to or `0` if the port should be automatically assigned.
+   */
+  def newServerAt(interface: String, port: Int): ServerBuilder = ServerBuilder(interface, port, system)
```
raboof, line 91: so we landed on always specifying the port? works for me.

Issue #3271, opened 2020-06-22
Make Http.bindAndHandleAsync the default entry point to use with routing DSL


jrudolph:
Two benefits:
 * it matches its structure better (routing DSL corresponds to async function rather than to Flow)
 * works with HTTP/2 out of the box if enabled

raboof:
Perhaps making it available as just `bind` (https://github.com/akka/akka-http/pull/3225#discussion_r440122171)

jrudolph:
:+1:

jrudolph:
> Perhaps making it available as just `bind` ([#3225 (comment)](https://github.com/akka/akka-http/pull/3225#discussion_r440122171))

`bind` doesn't work because the existing bind method also has default arguments.

Alternatives suggested before were:

 * `bindServer`
 * `serve`
 * `handle`

jrudolph:
Looking further into this, one problem is also that it will be hard to support the alias on the Java side. Even now we have a long list of overloads that we will have to copy for the new name.

jrudolph:
The combination of this issue with #3363 to also try to get rid of the `parallelism` parameters, uncovered a few obstacles:
 * on the Java side we need a multitude of overloads to support an alias (and even those are a somewhat arbitrary selection)
 * there's no way to gracefully deprecate the `parallelism` parameter, i.e. the only way to go there is to introduce the new name without a parallelism and deprecate `bindAndHandleAsync`

Going forward there are a few alternatives:

 * provide a new `bindServer` on Scala and Java side (without a `parallelism` parameter) and ignore for now that we cannot get rid of `parallelism` for `bindAndHandleAsync`
 * provide a new `bindServer` on Scala and Java side (while keeping the arbitrary set of overloads on the Java side) *and* deprecate `bindAndHandleAsync` (proposal in #3365)
 * provide a new ServerBuilder-based API, which is the idiomatic way of providing default-parameter-like APIs that can be evolved more easily. (Proposal for Java in #3362, Scala-side would basically be a copy with scaladsl types instead)

@raboof and @ennru (and anyone else interested), do you have any preference of where to go for 10.2.0?

raboof:
I could live with each of those proposals, though I think I'd prefer '1' and '3'. Perhaps even simpler might be: provide `bindServer` on the Scala side, and the 'builder'-based API on the Java side?

jrudolph:
[...]

PR #3364, opened 2020-07-08
core: introduce bindServer as an alias for bindAndHandleAsync

jrudolph:
And change usages with routes to use the new method.

Refs #3271

We still have the opportunity for bike-shedding to find better names for `bindServer` as suggested in https://github.com/akka/akka-http/issues/3271#issuecomment-648124081.


jrudolph:
The java side requires introducing a new set of overloads. I started #3362 to try out a builder pattern instead.

raboof:
(manually verified: travis is green)

jrudolph:
@ennru seemed supportive of the ServerBuilder approach, so I'm going to try that instead also for the Scala-side and then see if this one still makes sense.

jrudolph:
Closing for now.

PR #3365, opened 2020-07-09
Deprecate binadAndHandleAsync to get rid of parallelism parameter

jrudolph:
On top of #3364

Somewhat drastic change but it seems it would be the only way to get ultimately rid of the `parallelism` parameter.

Javadsl changes are somewhat in conflict with suggestion of new ServerBuilder API in #3362.


jrudolph:
Closed in favor of #3362

PR #3366, opened 2020-07-09
core: cleanup HttpMessage.copy chaos introduced in #3347

jrudolph:
Solution:
 * remove newly introduced and then deprecated variant that allows to set attributes
 * keep deprecated original copy method
 * create an internal copyImpl that allows to copy all fields
 * use copyImpl internally

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
@@ -341,7 +341,7 @@ final class HttpRequest(
    * http://tools.ietf.org/html/rfc7230#section-5.5
    */
   def withEffectiveUri(securedConnection: Boolean, defaultHostHeader: Host = Host.empty): HttpRequest =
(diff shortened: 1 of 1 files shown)


akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
```diff
@@ -354,48 +354,48 @@ final class HttpRequest(
...
-  @deprecated("Use the `withXYZ` methods instead", "10.2.0")
+  @deprecated("Use the `withXYZ` methods instead. Kept for binary compatibility", "10.2.0")
   def copy(
+    method:   HttpMethod                = method,
+    uri:      Uri                       = uri,
+    headers:  immutable.Seq[HttpHeader] = headers,
+    entity:   RequestEntity             = entity,
+    protocol: HttpProtocol              = protocol) = copyImpl(method, uri, headers, entity = entity, protocol = protocol)
+
+  private def copyImpl(
```
raboof, line 391: this way you can have default values for both the old and the new one :+1:

Issue #3263, opened 2020-06-17
Directive creating a Route from a partial function


raboof:
Similar to #3239, it would be helpful to have a directive that creates a Route for a partial function, and rejects the route (rather than failing) when the partial function is not defined for a given request.

jrudolph:
Done in #3367.

PR #3367, opened 2020-07-14
http: add handlePF directive

jrudolph:
Refs #3239, #3263

Before I continue, I'd like some feedback on API and naming.


akka-http/src/main/scala/akka/http/scaladsl/server/directives/RouteDirectives.scala
```diff
@@ -82,6 +83,47 @@ trait RouteDirectives {
...
+   *
+   * This directive can be used to include external components request processing components defined as PartialFunction
+   * (like those provided by akka-grpc) into a routing tree defined as routes.
+   *
+   * @param rejections The list of rejections to reject with if the handler is not defined for a request. By default a request
+   *                   is rejected with an empty list of rejections in that case, which is interpreted as "Not Found".
+   *
+   * @group route
+   */
+  def handlePF(handler: PartialFunction[HttpRequest, Future[HttpResponse]], rejections: Seq[Rejection] = Nil): StandardRoute =
```
raboof, line 105: :+1:, just `handle` would also be kind of nice but perhaps too ambiguous

akka-http/src/main/scala/akka/http/scaladsl/server/directives/RouteDirectives.scala
```diff
@@ -82,6 +83,47 @@ trait RouteDirectives {
...
+    { ctx =>
+      if (handler.isDefinedAt(ctx.request)) handler(ctx.request).fast.map(RouteResult.Complete)(ctx.executionContext)
```
raboof, line 107: unfortunately `isDefinedAt` is evil, because in some implementations it will actually evaluate the function to determine if it is defined for a certain input (so in this case it'd be evaluated twice: once by `isDefinedAt` and once by the `apply`). `applyOrElse` should work though?

jrudolph: Thanks. I changed that. The problem is a bit that you cannot cheaply apply a post-processing step in `applyOrElse`, so you will have to use `andThen` before the `applyOrElse` which isn't quite ideal (but might still be better than running the filter twice).

akka-http/src/main/scala/akka/http/scaladsl/server/directives/RouteDirectives.scala
jrudolph: It seems that would work. I think the original reason I put something else in here was to make it easier to document but maybe it's not a real problem to show both signatures on one page.

jrudolph: I made that change, but now it doesn't work any more because the wrong overload is chosen, I guess because of the default parameter in one case. I'll see if that can be fixed by even more overloads.

jrudolph: Added more overloads, so the overload resolution now works correctly ;)

jrudolph:
> Might be interesting to think about if we could come up with something for the javadsl, but it's not so obvious what a PartialFunction should look like there.

Yes, good point, I think akka-grpc also generates the scala signatures for `partial` on the Java side?

raboof:
> Yes, good point, I think akka-grpc also generates the scala signatures for `partial` on the Java side?

Not currently, but if we add a directive to convert a scala `PartialFunction` to a javadsl `Route` I think changing it to return a scaladsl `PartialFunction` might make sense!

(it currently returns a non-partial java function with a 'magic constant' representing 'no return value')

PR #3368, opened 2020-07-14
Fix scenario where buffered element is lost

raboof:
Feels like this is more complicated (and harder to test) than it
should be, but I don't see an obvious way to simplify.

### DIFF
--- a/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/ProtocolSwitch.scala
+++ b/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/ProtocolSwitch.scala
@@ -26,8 +26,8 @@ private[http] object ProtocolSwitch {
       new GraphStage[FlowShape[SslTlsInbound, SslTlsOutbound]] {
 
         // --- outer ports ---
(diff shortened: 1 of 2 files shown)


jrudolph:
Can you explain how the scenario could be observed?

raboof:
> Can you explain how the scenario could be observed?

I saw it when I sent a request and then immediately completed the source.

I guess in 'real life' that might be triggered by half-closed TCP connections, though it is unclear how a HTTP/2 server should handle half-closed TCP connections (everything I can find is about HTTP/1 and it seems there's no consensus there either...).

The spec says https://tools.ietf.org/html/rfc7540#section-5.4.3 about automatically retrying failed requests from half-closed connections, but nothing about whether requests may be actively aborted in that case.

jrudolph:
HTTP/2 has an indication of which requests have been processed in the GOAWAY frame.

jrudolph:
> I saw it when I sent a request and then immediately completed the source.

Which source was that? For the HTTP/2 client or server? I guess for the client a nice behavior would be that we don't close the TCP connection for a time when there's still a response outstanding.

raboof:
> Which source was that?

(this was the 'test source' exercising the server stack in #3369)

PR #3369, opened 2020-07-14
Http2 benchmarks

raboof:
Very simplistic so far, but this way at least we have some baseline

### DIFF
--- a/akka-http-bench-jmh/src/main/scala/akka/http/impl/engine/http2/H2ServerProcessingBenchmark.scala
+++ b/akka-http-bench-jmh/src/main/scala/akka/http/impl/engine/http2/H2ServerProcessingBenchmark.scala
@@ -82,7 +82,7 @@ class H2ServerProcessingBenchmark extends CommonBenchmark {
     })
     val http2 =
       Http2Blueprint.handleWithStreamIdHeader(1)(req => {
-        req.entity.dataBytes.runWith(Sink.ignore).map(_ => response)
+        req.discardEntityBytes().future.map(_ => response)
       })(system.dispatcher)
         .join(Http2Blueprint.serverStackTls(settings, log))
     httpFlow = Http2.priorKnowledge(http1, http2)
--- a/akka-http-bench-jmh/src/main/scala/akka/http/impl/engine/http2/H2ServerProcessingBenchmark.scala
+++ b/akka-http-bench-jmh/src/main/scala/akka/http/impl/engine/http2/H2ServerProcessingBenchmark.scala
@@ -1,0 +1,95 @@
+package akka.http.impl.engine.http2
+
+import java.util.concurrent.CountDownLatch
+
+import scala.concurrent.Await
+import scala.concurrent.Future
+import scala.concurrent.duration._
+import akka.actor.ActorSystem
+import akka.http.CommonBenchmark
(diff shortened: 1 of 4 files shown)


jrudolph:
(sorry for the spam, trying to get a new node up)

akka-http-bench-jmh/src/main/scala/akka/http/impl/engine/http2/H2ServerProcessingBenchmark.scala
```diff
@@ -0,0 +1,95 @@
...
+    mat = ActorMaterializer()
+    val settings = implicitly[ServerSettings]
+    val log = system.log
+    implicit val ec = system.dispatcher
+    val http1 = Flow[SslTlsInbound].mapAsync(1)(_ => {
+      Future.failed[SslTlsOutbound](new IllegalStateException("Failed h2 detection"))
+    })
+    val http2 =
+      Http2Blueprint.handleWithStreamIdHeader(1)(req => {
+        req.entity.dataBytes.runWith(Sink.ignore).map(_ => response)
```
jrudolph, line 85:
```suggestion
        req.discardEntityBytes().future.map(_ => response)
```

which will avoid materializing an empty source.

akka-http-bench-jmh/src/main/scala/akka/http/impl/engine/http2/H2ServerProcessingBenchmark.scala
```diff
@@ -0,0 +1,95 @@
...
+  val packedResponse = ByteString(-62, -63, -64, -65, -66)
+
+  @Benchmark
+  @OperationsPerInvocation(10000)
+  def benchRequestProcessing(): Unit = {
+    val numRequests = 10000
+    val latch = new CountDownLatch(numRequests)
+
+    val (in, done) =
+      Source.queue(numRequests, OverflowStrategy.fail)
```
jrudolph, line 42: Could this just be a fixed source that concats the preface with a number of requests?

raboof:
> LGTM, is the fix from the other PR required? Otherwise we could already merge it without.

No, not in this form, so I'll rebase

> Could this just be a fixed source that concats the preface with a number of requests?

No, since then it'd have a half-closed problem - so then it'd at least need to have a `Source.maybe` attached at the end, kindof defeating the purpose

jrudolph:
> > Could this just be a fixed source that concats the preface with a number of requests?
> 
> No, since then it'd have a half-closed problem - so then it'd at least need to have a `Source.maybe` attached at the end, kindof defeating the purpose

The purpose would be not to avoid extra overhead of the queue, so having a Source.maybe might be just fine.

raboof:
> The purpose would be to avoid extra overhead of the queue, so having a Source.maybe might be just fine.

Interestingly the throughput didn't really seem to change but I didn't look for hotspots yet :)

raboof:
(travis is happy but didn't report back)

jrudolph:
> > The purpose would be to avoid extra overhead of the queue, so having a Source.maybe might be just fine.
> 
> Interestingly the throughput didn't really seem to change but I didn't look for hotspots yet :)

Yes, I guess that's expected. It might become more significant ones other optimizations have been done.

PR #3382, opened 2020-07-21
core: polishing and docs for ServerBuilder API

jrudolph:
* One significant change over #3362: I renamed `bind()` to `connectionSource` to make it clear that calling the method will not immediately create a binding and also to avoid overloading `bind()` (which became awkward in docs)
 * Converted remaining internal and docs usages of `Http().bindXYZ` usages
 * Brief release notes and migration notes entry

### DIFF
--- a/akka-http-core/src/main/resources/reference.conf
+++ b/akka-http-core/src/main/resources/reference.conf
@@ -21,11 +21,11 @@ akka.http {
     # "PREVIEW" features that are not yet fully production ready.
     # These flags can change or be removed between patch releases.
     preview {
-      # ONLY WORKS WITH `bindAndHandleAsync` (currently)
-      #
       # If this setting is enabled AND the akka-http2-support is found
-      # on the classpath the usual Http().bind... method calls will bind
-      # using HTTP/2. Please note that you must configure HTTPS while doing so.
+      # on the classpath, `Http().newServerAt(...).bind` and `bindSync`
+      # will be enabled to use HTTP/2. Please note that you must configure HTTPS while doing so.
+      #
+      # `Http().newServerAt(...).bindFlow` and `connectionSource(): Source` are not supported.
       enable-http2 = off
     }
 
@@ -71,8 +71,8 @@ akka.http {
     # Set to 'infinite' to disable automatic connection closure (which will risk to leak connections).
     linger-timeout = 1 min
 
-    # The maximum number of concurrently accepted connections when using the
-    # `Http().bindAndHandle` methods.
+    # The maximum number of concurrently accepted connections when binding a server using
+    # `Http().newServerAt().bindXYZ()` methods.
     #
     # This setting doesn't apply to the `Http().bind` method which will still
     # deliver an unlimited backpressured stream of incoming connections.
--- a/akka-http-core/src/main/resources/reference.conf
+++ b/akka-http-core/src/main/resources/reference.conf
@@ -23,7 +23,7 @@ akka.http {
     preview {
       # If this setting is enabled AND the akka-http2-support is found
       # on the classpath, `Http().newServerAt(...).bind` and `bindSync`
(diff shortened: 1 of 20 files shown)


akka-http-core/src/main/resources/reference.conf
```diff
@@ -21,11 +21,11 @@ akka.http {
     # "PREVIEW" features that are not yet fully production ready.
     # These flags can change or be removed between patch releases.
     preview {
-      # ONLY WORKS WITH `bindAndHandleAsync` (currently)
-      #
       # If this setting is enabled AND the akka-http2-support is found
-      # on the classpath the usual Http().bind... method calls will bind
-      # using HTTP/2. Please note that you must configure HTTPS while doing so.
+      # on the classpath, `Http().newServerAt(...).bind` and `bindSync`
+      # will be enabled to use HTTP/2. Please note that you must configure HTTPS while doing so.
```
raboof, line 26:
```suggestion
      # will be enabled to use HTTP/2.
```

I know this wasn't introduced in this PR, but I don't think this is true anymore.

akka-http-core/src/main/scala/akka/http/javadsl/ServerBuilder.scala
```diff
@@ -106,18 +106,17 @@ trait ServerBuilder {
...
-   * If the configured  port is non-zero subsequent materialization attempts of the produced source will immediately
+   *  * if the configured  port is non-zero subsequent materialization attempts of the produced source will immediately
    * fail, unless the first materialization has already been unbound. Unbinding can be triggered via the materialized
    * [[akka.http.javadsl.ServerBinding]].
-   *
-   * If no `port` is explicitly given (or the port value is negative) the protocol's default port will be used,
-   * which is 80 for HTTP and 443 for HTTPS.
    */
-  def bind(): Source[IncomingConnection, CompletionStage[ServerBinding]]
+  def connectionSource(): Source[IncomingConnection, CompletionStage[ServerBinding]]
```
raboof, line 119: We also call it `Tcp.bind()` in Akka, but I agree something like this is better.

Just `connections` or `source`/`asSource` would work for me as well, this is :+1: too.

docs/src/test/scala/docs/http/scaladsl/server/HttpsServerExampleSpec.scala
```diff
@@ -61,12 +61,12 @@ abstract class HttpsServerExampleSpec extends AnyWordSpec with Matchers
     //#both-https-and-http
     // you can run both HTTP and HTTPS in the same application as follows:
     val commonRoutes: Route = get { complete("Hello world!") }
-    Http().newServerAt("127.0.0.1", 443).enableHttps(https).bindFlow(commonRoutes)
-    Http().newServerAt("127.0.0.1", 80).bindFlow(commonRoutes)
+    Http().newServerAt("127.0.0.1", 443).enableHttps(https).bind(commonRoutes)
+    Http().newServerAt("127.0.0.1", 80).bind(commonRoutes)
```
raboof, line 65: :+1: weird that we had `bindFlow` here before

akka-http-core/src/main/scala/akka/http/javadsl/ServerBuilder.scala
jrudolph: > Just `connections` or `source`/`asSource` would work for me as well, this is +1 too.

All good suggestions, but lets keep `connectionSource` as I like the extra verbosity for that particular method.

docs/src/test/scala/docs/http/scaladsl/server/HttpsServerExampleSpec.scala
jrudolph: I think that might have been missed by scalafix for some reason.

PR #3384, opened 2020-07-22
docs: add 'troubleshooting' page, populate with 'PRI' error message

raboof:


docs/src/main/paradox/troubleshooting.md
```diff
@@ -0,0 +1,17 @@
+# Troubleshooting
```
jrudolph, line 1: Can you make this an overview page and add a single page for that particular issue? That will help us to add a link for that particular error page later.

raboof: :+1: done

PR #3386, opened 2020-07-22
Convert more javadsl doc examples from bindAndHandle to ServerBuilder

jrudolph:
Last time I missed Java examples that still used old `bindAndHandle` APIs and also a `scaladsl.Source` slipped into `javadsl.ServerBuilder`.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/javadsl/ServerBuilder.scala
+++ b/akka-http-core/src/main/scala/akka/http/javadsl/ServerBuilder.scala
@@ -16,8 +16,7 @@ import akka.http.scaladsl
 import akka.http.scaladsl.util.FastFuture
 import akka.http.scaladsl.{ model => sm }
 import akka.japi.function.Function
-import akka.stream.javadsl.Flow
-import akka.stream.scaladsl.Source
+import akka.stream.javadsl.{ Flow, Source }
 import akka.stream.{ Materializer, SystemMaterializer }
 
 import scala.compat.java8.FutureConverters._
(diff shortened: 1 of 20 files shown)


akka-http-core/src/main/scala/akka/http/javadsl/ServerBuilder.scala
```diff
@@ -16,8 +16,7 @@ import akka.http.scaladsl
 import akka.http.scaladsl.util.FastFuture
 import akka.http.scaladsl.{ model => sm }
 import akka.japi.function.Function
-import akka.stream.javadsl.Flow
-import akka.stream.scaladsl.Source
```
raboof, line 20: :hushed: totally missed that

docs/src/test/java/docs/http/javadsl/server/testkit/MyAppService.java
```diff
@@ -40,13 +38,10 @@ public Route createRoute() {
...
   public static void main(String[] args) throws IOException {
     final ActorSystem system = ActorSystem.create();
-    final ActorMaterializer materializer = ActorMaterializer.create(system);
 
     final SimpleServerApp app = new SimpleServerApp();
 
-    final ConnectHttp host = ConnectHttp.toHost("127.0.0.1");
-
-    Http.get(system).bindAndHandle(app.createRoute().flow(system, materializer), host, materializer);
+    Http.get(system).newServerAt("127.0.0.1", 8080).bind(app.createRoute());
```
raboof, line 44: Much nicer. Maybe add newlines?

akka-http2-support/src/test/scala/akka/http/scaladsl/Http2BindingViaConfigSpec.scala
```diff
@@ -41,7 +39,7 @@ class Http2BindingViaConfigSpec extends AkkaSpec("""
 
         val connectionContext = ExampleHttpContexts.exampleServerContext
         binding =
-          Http().newServerAt(host, port)
+          Http().newServerAt("localhost", 0)
```
raboof, line 42: Safer :+1:

docs/src/test/java/docs/http/javadsl/server/testkit/MyAppService.java
jrudolph: It's already shorter than before and there are so many similar instances of this, so I'd rather keep it like this than touching it again right now. The whole documentation will need an overhaul so I'll try to fix those cosmetics in a run next week.

PR #3409, opened 2020-07-30
build: quote more parts in PR validation report

jrudolph:
Currently on top of 4ce9afb which is known to fail with output in that segment.

### DIFF
--- a/project/ValidatePullRequest.scala
+++ b/project/ValidatePullRequest.scala
@@ -393,7 +393,7 @@ object AggregatePRValidation extends AutoPlugin {
         def hasTests(result: Tests.Output): Boolean = result.events.exists(e => hasExecutedTests(e._2))
 
         def printTestResults(result: KeyValue[Tests.Output]): Unit = {
-          write(s"Test result for `${showKey(result.key)}``")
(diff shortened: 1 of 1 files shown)


project/ValidatePullRequest.scala
```diff
@@ -393,7 +393,7 @@ object AggregatePRValidation extends AutoPlugin {
         def hasTests(result: Tests.Output): Boolean = result.events.exists(e => hasExecutedTests(e._2))
 
         def printTestResults(result: KeyValue[Tests.Output]): Unit = {
-          write(s"Test result for '${showKey(result.key)}'")
+          write(s"Test result for `${showKey(result.key)}``")
```
jrudolph, line 396:
```suggestion
          write(s"Test result for `${showKey(result.key)}`")
```

jrudolph:
See the failures above as an example. Now only the fixes are included here.

project/ValidatePullRequest.scala
ennru: This still needs to be applied, right?

jrudolph: Ah, good catch, wonder what happened there...

PR #3415, opened 2020-08-04
docs: create actor system from a guardian instead

ennru:
Updates the minimal examples with the new actors API to create the Akka HTTP binding in a guardian actor instead.

### DIFF
--- a/docs/src/test/java/docs/http/javadsl/HttpServerMinimalExampleTest.java
+++ b/docs/src/test/java/docs/http/javadsl/HttpServerMinimalExampleTest.java
@@ -5,6 +5,7 @@
 package docs.http.javadsl;
 //#minimal-routing-example
 import akka.actor.typed.ActorSystem;
+import akka.actor.typed.Behavior;
 import akka.actor.typed.javadsl.Behaviors;
 import akka.http.javadsl.Http;
 import akka.http.javadsl.ServerBinding;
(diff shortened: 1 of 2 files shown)


jrudolph:
> Not sure about this, it makes it less of a minimal example than before... Do we gain anything by doing it like this?

To be more clear, if you actually have a typed actor based application, I understand that you should follow the typed guidelines, but if you just have a plain Akka Stream / Akka HTTP based application, that seems like extra boilerplate that only complicates stuff.

patriknw:
I don't have a strong opinion and I can understand the argument about plain Akka Stream/HTTP applications. At least, because that is what we and existing users are used to.

The drawback is that we show different ways and if the plain HTTP app is expanded into using actors the structure must be changed. For example spawning a child actor for handling a HTTP request can not be done from the outside.

Somewhat related is my proposal of an Extensions for spawning actors from the outside. HTTP routes were one of the use cases I had in mind. Other's were not enthusiastic about it and maybe the suggestion with a directive would be more interesting.

jrudolph:
> The drawback is that we show different ways and if the plain HTTP app is expanded into using actors the structure must be changed. For example spawning a child actor for handling a HTTP request can not be done from the outside.

I guess for global actors that would indeed be the place to create them but if you don't need those, the extra nesting seems confusing.

> Somewhat related is my proposal of an Extensions for spawning actors from the outside. HTTP routes were one of the use cases I had in mind. Other's were not enthusiastic about it and maybe the suggestion with a directive would be more interesting.

Yes, that would be nice especially for dynamically created actors. If people try to create child actors manually from the route, it seems hard to have a context available to actually create those actors.

ennru:
Ok, let us keep the minimal examples minimal.

PR #3431, opened 2020-08-10
core: add debug logging when Websocket impl closes connections after timeout

### LINKED ISSUE #75: Websocket client (or server) disconnect after 6 seconds
<a href="https://github.com/johanandren"></img></a> **Issue by [johanandren](https://github.com/johanandren)**
_Monday Mar 07, 2016 at 12:17 GMT_
_Originally opened as https://github.com/akka/akka/issues/19957_

---

This sample based on a question on the mailing list: https://gist.github.com/johanandren/858bc3d0cfeaf66ee116

Expected behavior is that the client would keep consuming one tick per second from the server until it is killed but in reality it gets 5 events and then the connection is closed.

On the server side this is logged: 

```
[akka://test/system/IO-TCP/selectors/$a/1] Closing connection due to IO error java.io.IOException: Broken pipe
```

jrudolph:
Also converted to more idiomatic Deadline.

Fixes #75

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/ws/FrameOutHandler.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/ws/FrameOutHandler.scala
@@ -9,13 +9,12 @@ import akka.annotation.InternalApi
 import akka.event.LoggingAdapter
 import akka.http.impl.engine.ws.FrameHandler.{ UserHandlerErredOut, _ }
 import akka.http.impl.engine.ws.WebSocket.Tick
-import akka.http.impl.util.Timestamp
 import akka.stream.scaladsl.Flow
 import akka.stream.stage._
 import akka.stream.{ Attributes, FlowShape, Inlet, Outlet }
 
 import scala.annotation.tailrec
-import scala.concurrent.duration.FiniteDuration
+import scala.concurrent.duration.{ Deadline, FiniteDuration }
 
 /**
  * Implements the transport connection close handling at the end of the pipeline.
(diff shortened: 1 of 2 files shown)


raboof (approved):
Much more readable with the 'deadline' as well :+1:

akka-http-core/src/main/scala/akka/http/impl/engine/ws/FrameOutHandler.scala
```diff
@@ -105,16 +104,18 @@ private[http] class FrameOutHandler(serverSide: Boolean, _closeTimeout: FiniteDu
...
-            else pull(in)
+            if (deadline.isOverdue()) {
+              if (log.isDebugEnabled) log.debug(s"Peer did not acknowledge CLOSE frame after ${_closeTimeout}, closing underlying connection now.")
+              completeStage()
+            } else pull(in)
           case PeerClosed(code, reason) =>
             if (serverSide) completeStage()
             else {
-              setHandler(in, new WaitingForTransportClose())
+              setHandler(in, new WaitingForTransportClose(deadline))
```
raboof, line 118: Hmm, so previously we had `_closeTimeout` for the peer to send its close frame, and another `_closeTimeout` for the transport to close, while now we expect to see both within one `_closeTimeout`.

Is that what we want?

jrudolph: I think we can redefine it's meaning as we see fit. Previously, the actual delay was pretty confusing because it didn't seem to correspond to any of the configured values because it was twice as long as expected.

raboof: I like having separate timeouts for each phase: if each phase 'normally' takes 30ms and 'exceptionally' 2s, you can set the timeout to 2s. Then, when the first action completes normally, the second one will be aborted within 30ms+2s. When using just one timeout, you'd have to set it to 4s, and even when the first half completes normally within 30ms you won't abort the second half before the full 4s have expired.

But it's not important :D

jrudolph: > But it's not important :D

Yeah, the question would be why you would be interested in each of those phases. In the end, what you care about is that a connection that might be hung actually is torn down after the deadline.

PR #3437, opened 2020-08-13
http: fail storeUploadedFile(s) directive when IO operation fails

### LINKED ISSUE #1488: `uploadedFile` directive doesn't work correctly on request cancellation
It gets executed with a partially uploaded file and logs an error (`Tearing down FileSink ... EntityStreamException: Entity stream truncation`).

Looks like `FileIO.toPath` returned `IOResult` but the directive didn't check for failure and turned it into a file like this `.map(_ ⇒ (fileInfo, destination))` (placeholder is `IOResult`). I was able to fix it like this. https://gist.github.com/RobertoUa/34f360a9e9a1e4c2bab0abce361775fe
Notice that I use flatMap on `IOResult`'s `status` so that error gets propagated. `FileSubscriber` still logs an error which is undesirable.

jrudolph:
Fixes #1488

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
@@ -9,6 +9,7 @@ import akka.actor.Cancellable
 import akka.annotation.InternalApi
 import akka.dispatch.ExecutionContexts
 import akka.http.scaladsl.model.HttpEntity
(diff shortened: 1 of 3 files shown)


ennru (approved):
Sad to see that https://github.com/akka/akka/issues/23951 hits again.
LGTM.

raboof (approved):
:+1: to have the test so we can easily see if an future upstream fix has the desired effect later.

PR #3439, opened 2020-08-13
core: don't leak internal limits to the clients for EntityStreamSizeException

### LINKED ISSUE #1015: Akka HTTP configuration details exposed to client on EntityStreamSizeException
Hi,

The current implementation of the [EntityStreamSizeException](https://github.com/akka/akka-http/blob/3c966af2255b58b02fbdacd6abec89c47f78ce48/akka-http-core/src/main/scala/akka/http/scaladsl/model/ErrorInfo.scala#L83) handling exposes a detailed message about which Akka settings to use (to allow bigger contents in requests) to the client, even though `verbose-error-messages` is set to `off`.

I changed the Class to extend [ExceptionWithErrorInfo](https://github.com/akka/akka-http/blob/3c966af2255b58b02fbdacd6abec89c47f78ce48/akka-http-core/src/main/scala/akka/http/scaladsl/model/ErrorInfo.scala#L36) to distinguish between a summary and a detailed message.
There is a check for the verbosity setting in [HttpServerBluePrint:459](https://github.com/akka/akka-http/blob/3c966af2255b58b02fbdacd6abec89c47f78ce48/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala#L459,) but it seems like it is never called and the `ErrorInfo.formatPretty` result is returned to the client (which may also just be the result of `RuntimeException.getMessage` as `ExceptionWithErrorInfo` extends `RuntimeException`).

Do you have an idea why or what to change to only display the summary in that case?
The mentioned classes are in the akka-http-core module.

You can trigger it by setting `akka.http.server.parsing.max-content-length` to `1` and send some data with your request.

I pushed my current changes to https://github.com/makubi/akka-http/commit/546af4fa4ac4c5c7e78df2c8b3ba083aa64b3c2e.

Thanks in advance.

jrudolph:
Fixes #1015.

~Also never include details from ErrorInfo in the exception message to avoid
accidental leakage.~

~This is now a rather big change because it's now much more likely that details will be missing also from log messages in many instances. To retain the old behavior we might want to review error logging to make sure they use `info.formatPretty` to log the problem.~

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/ErrorInfo.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/ErrorInfo.scala
@@ -7,6 +7,8 @@ package akka.http.scaladsl.model
 import StatusCodes.ClientError
 import akka.annotation.InternalApi
 
+import scala.runtime.AbstractFunction2
+
 /**
  * Two-level model of error information.
  * The summary should explain what is wrong with the request or response *without* directly
@@ -69,7 +71,7 @@ object ErrorInfo {
 }
 
 /** Marker for exceptions that provide an ErrorInfo */
-abstract class ExceptionWithErrorInfo(val info: ErrorInfo, cause: Throwable) extends RuntimeException(info.formatPretty, cause) {
+abstract class ExceptionWithErrorInfo(val info: ErrorInfo, cause: Throwable) extends RuntimeException(info.format(withDetail = true), cause) {
   def this(info: ErrorInfo) = this(info, null)
 }
 
@@ -118,16 +120,19 @@ object EntityStreamException {
  * The limit can also be configured in code, by calling [[HttpEntity#withSizeLimit]]
  * on the entity before materializing its `dataBytes` stream.
  */
-final case class EntityStreamSizeException(limit: Long, actualSize: Option[Long] = None) extends RuntimeException {
-
-  override def getMessage = toString
-
-  override def toString = {
-    s"EntityStreamSizeException: incoming entity size (${actualSize.getOrElse("while streaming")}) exceeded size limit ($limit bytes)! " +
-      "This may have been a parser limit (set via `akka.http.[server|client].parsing.max-content-length`), " +
-      "a decoder limit (set via `akka.http.routing.decode-max-size`), " +
-      "or a custom limit set with `withSizeLimit`."
-  }
+final case class EntityStreamSizeException(limit: Long, actualSize: Option[Long] = None)
+  extends ExceptionWithErrorInfo(EntityStreamSizeException.errorInfo(limit, actualSize)) {
+  override def getMessage: String = super.getMessage // for bin compat
+}
+object EntityStreamSizeException /* for bin compat */ extends AbstractFunction2[Long, Option[Long], EntityStreamSizeException] {
+  private[EntityStreamSizeException] def errorInfo(limit: Long, actualSize: Option[Long]): ErrorInfo =
+    ErrorInfo(
+      "The incoming entity size exceeded the size limit of the server.",
+      s"Incoming entity size (${actualSize.getOrElse("while streaming")}) exceeded size limit ($limit bytes)! " +
+        "This may have been a parser limit (set via `akka.http.[server|client].parsing.max-content-length`), " +
+        "a decoder limit (set via `akka.http.routing.decode-max-size`), " +
+        "or a custom limit set with `withSizeLimit`."
+    )
 }
 
 case class RequestTimeoutException(request: HttpRequest, message: String) extends RuntimeException(message)
--- a/akka-http-core/src/test/scala/akka/http/impl/engine/server/HttpServerSpec.scala
+++ b/akka-http-core/src/test/scala/akka/http/impl/engine/server/HttpServerSpec.scala
@@ -1225,7 +1225,7 @@ class HttpServerSpec extends AkkaSpec(
                 val error = origError.getCause
 
                 error shouldEqual EntityStreamSizeException(limit, Some(actualSize))
-                error.getMessage should include("exceeded size limit")
+                error.getMessage should include("exceeded the size limit")
 
                 responses.sendError(error.asInstanceOf[Exception])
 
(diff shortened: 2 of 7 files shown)


jrudolph:
I marked it now as draft as I'm not 100% sure we should do it (now). It seems to be the right thing as it is much too easy to accidentally slip exception messages into outgoing messages (as witnessed here as well) but on the other hand, detailed error info will now also be missing from log messages (which sometimes is still considered the right thing).

akka-http-core/src/main/scala/akka/http/scaladsl/model/ErrorInfo.scala
```diff
@@ -69,7 +71,7 @@ object ErrorInfo {
 }
 
 /** Marker for exceptions that provide an ErrorInfo */
-abstract class ExceptionWithErrorInfo(val info: ErrorInfo, cause: Throwable) extends RuntimeException(info.formatPretty, cause) {
+abstract class ExceptionWithErrorInfo(val info: ErrorInfo, cause: Throwable) extends RuntimeException(info.format(withDetail = false), cause) {
```
raboof, line 74: This is tricky indeed.

Ideally, we should not output 'possibly sensitive' information to clients unless verbose `verbose-error-messages` is on. Without further information we should consider the `message` of an exception as 'possibly sensitive' and not output it, so from that perspective `withDetail = true` would be reasonable.

In the case of EntityStreamSizeException, this exception is not caught by the exception handler, but turned into a `MalformedRequestContentRejection`. Since rejections are meant to be shown to the client (and not necessarily go into the logging), I guess the rejection message should not contain 'possibly sensitive' information.

Unfortunately FormFieldDirectives does just put the exception message into the rejection message. Perhaps we could follow the approach in MarshallingDirectives and look at `verbose-error-messages` when turning an exception that has an `ErrorInfo` into a rejection, like https://github.com/jrudolph/akka-http/pull/2 ?

jrudolph: These are good points. I guess we should try to untangle a few related but distinct aspects of it to make some progress:

 * The particular issue in question about the EntitySize exception
 * Which information is allowed to be included to users?
 * Which information stack traces are allowed to carry and how they should be logged

> The particular issue in question about the EntitySize exception

This can and maybe should be solved in isolation.

> Which information is allowed to be included in message to users in general?

I think we have quite good guidelines on this. In doubt, be conservative and make the messages as generic as possible to avoid attack vectors.

Neither this PR nor your suggestion actual addresses the real problem here: 

We should never include a clause like this which trusts generic exception implementations to do the right thing:

```scala
        case Failure(x)                                  => reject(MalformedRequestContentRejection(x.getMessage.nullAsEmpty, x))
```

If we want to fix something generically it should be these places where we shouldn't include any concrete information about an exception with `verbose-error-message = off`.

> Which information stack traces are allowed to carry and how they should be logged

There are different levels of sensitive data. Critically, secrets should never be included in any messages that might be logged. Other details should be allowed to be logged to help with debugging though not be exposed to users in responses.

raboof: > Neither this PR nor your suggestion actual addresses the real problem here

I agree: they happen to solve the EntitySize situation specifically, but...

> We should never include a clause like this which trusts generic exception implementations to do the right thing:
> 
> ```scala
>         case Failure(x)                                  => reject(MalformedRequestContentRejection(x.getMessage.nullAsEmpty, x))
> ```

Indeed this is the 'real problem'. Unfortunately this won't be such an easy change, as we'd have to review how *else* to convey the details of this error then...

> If we want to fix something generically it should be these places where we shouldn't include any concrete information about an exception with `verbose-error-message = off`.

:+1: 

> > Which information stack traces are allowed to carry and how they should be logged
> 
> There are different levels of sensitive data. Critically, secrets should never be included in any messages that might be logged. Other details should be allowed to be logged to help with debugging though not be exposed to users in responses.

:+1:

jrudolph: > Indeed this is the 'real problem'. Unfortunately this won't be such an easy change, as we'd have to review how _else_ to convey the details of this error then...

Yep, probably still use your suggestion for that and deal with the more generic things in another PR.

jrudolph: > We should never include a clause like this which trusts generic exception implementations to do the right thing:
> 
> ```scala
>         case Failure(x)                                  => reject(MalformedRequestContentRejection(x.getMessage.nullAsEmpty, x))
> ```
> 
> If we want to fix something generically it should be these places where we shouldn't include any concrete information about an exception with `verbose-error-message = off`.

On potential problem with this is that we use `IllegalArgumentExceptions` to transport parsing errors in various places. In those cases the details would now be hidden behind `verbose-error-messages` as well.

jrudolph: Another aspect that I found, is that the processing varies between the `FormFieldDirectives`.

 * The `formField` variants use `toStrictEntity` which will fail with an exception directly without giving control to the form field processing at all, so the response is created from the exception handler ultimately.
 * The `formFieldSeq` / `Map` / `MultiMap` directives however do no (yet?) use `toStrictEntity` (because we didn't expect people to nest those with `formField` which might be to optimistic anyway). So those fail while actually running the unmarshaller so that the extra handling logic above is run to wrap the exceptions with rejections as we see fit.

jrudolph:
I revised the akka-http parts of the problem. What remains is that there's another layer of handling on the lowest layer which still reports to much information.

jrudolph:
Now I also fixed the remaining error message from the internals.

akka-http-core/src/test/scala/akka/http/impl/engine/server/HttpServerSpec.scala
```diff
@@ -1174,27 +1174,27 @@ class HttpServerSpec extends AkkaSpec(
       netIn.sendComplete()
     })
 
-    "support request length verification" which afterWord("is defined via") {
-
-      class LengthVerificationTest(maxContentLength: Int) extends TestSetup(maxContentLength) {
-        val entityBase = "0123456789ABCD"
-        def sendStrictRequestWithLength(bytes: Int) =
-          send(s"""POST /foo HTTP/1.1
+    def supportRequestLengthVerification(verboseErrorMessages: Boolean = true) =
```
jrudolph, line 1177: Most of the changes here are because of the added indentation. What I actually added was running the test both with `verboseErrorMessage = true` and `false` and adapting the expected error messages accordingly.

PR #3440, opened 2020-08-13
core: make entities of HEAD requests discard

### LINKED ISSUE #1495: Connectionpool leak in http client
When doing HEAD requests using http.singleRequest, I can only perform 4 requests before the threadpool is exhausted. If I perform GET requests, the connections are closed properly.
If I discard the entitiy bytes after doing a HEAD request, this bug does also not occur.

Example Code broken:
`                val eventualHead: Future[HttpResponse] = http.singleRequest(HttpRequest(HttpMethods.HEAD, uri = url))
                onComplete(eventualHead) { r =>
                  r.map(_.status) match {
                    case Success(code) =>
                      complete(HttpResponse(code, entity = asJson(response.toString())))
                    case _ => complete(HttpResponse(StatusCodes.NotFound))
                  }
                }`

Workaround:
`                val eventualHead: Future[HttpResponse] = http.singleRequest(HttpRequest(HttpMethods.HEAD, uri = url))
                  .map { res =>
                    res.discardEntityBytes()
                    res
                  }
                onComplete(eventualHead) { r =>
                  r.map(_.status) match {
                    case Success(code) =>
                      complete(HttpResponse(code, entity = asJson(response.toString())))
                    case _ => complete(HttpResponse(StatusCodes.NotFound))
                  }
                }`

jrudolph:
Fixes #1495

It seems without this fix responses to HEAD requests will have to discard
the entities manually or the entity subscription timeout will kick in.

This is more of a hack but it might work well enough for now.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/NewHostConnectionPool.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/NewHostConnectionPool.scala
@@ -16,7 +16,7 @@ import akka.http.impl.engine.client.PoolFlow.{ RequestContext, ResponseContext }
 import akka.http.impl.engine.client.pool.SlotState._
 import akka.http.impl.util.{ RichHttpRequest, StageLoggingWithOverride, StreamUtils }
 import akka.http.scaladsl.Http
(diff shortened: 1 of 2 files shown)


raboof (approved):

ennru (approved):
LGTM.

PR #3448, opened 2020-08-24
Updated Scala style guide example to match the Java one (removed duplication)

agemooij:
The Java example for `path-compose` already does the right thing but the Scala example duplicates the `customer` path.
I also fixed the name of the extracted `customerId` value.


ennru (approved):
LGTM, thank you.

raboof (approved):
Thanks!

Issue #3209, opened 2020-06-03
GracefulTermination does not take response entity streams into account


jrudolph:
As observed in https://github.com/akka/akka-http/pull/3139#discussion_r434433886, graceful termination does not take response entity streams into account. I.e. those streams could still be running when the deadline triggers. What currently happens is that due to https://github.com/akka/akka-http/blob/27d987f922a08cdb44ad1ce0ba3e043920724891/akka-http-core/src/main/scala/akka/http/impl/engine/server/ServerTerminator.scala#L264, the top of the server stack is immediately completed on graceful termination (i.e. even before the deadline). However, the controller prevents that this completion is propagated to the connection. Instead, after some time (related to `server.stream-cancellation-delay`) also cancellation arrives at the controller stage in which case the rest of the stack is completed. There are a few issues with that:

 * The stack should fail (where it just regularly completes currently), so that the connection is abandoned and the peer knows that something bad happened. Especially, for `CloseDelimited` entities, the stream will otherwise be silently truncated.
 * The stack should be completed only after the deadline and regardless of the `stream-cancellation-delay`.
 * (incidentally) the controller should be reviewed, why it closes the connection regularly after it has received cancellation from the user side

johanandren:
Guessing there is something left here given that it was pushed across previous milestones

Issue #2532, opened 2019-05-16
akka.http.scaladsl.ClientServerSpec.The low-level HTTP infrastructure should complete a request/response over https when request has `Connection: close` set failing on 2.13


jrudolph:
https://jenkins.akka.io:8498/job/akka-http-nightly/487/AKKA_VERSION=default,Node=akka-http-nightly-node,SCALA_VERSION=2.13.0-RC1,jdk=AdoptOpenJDK%208/testReport/junit/akka.http.scaladsl/ClientServerSpec/The_low_level_HTTP_infrastructure_should_complete_a_request_response_over_https_when_request_has__Connection__close__set/

```
Error Message

timeout (3 seconds) waiting for 1 messages on WarningFilter(None,Right(Hostname verification failed),false)

Stacktrace

java.lang.AssertionError: timeout (3 seconds) waiting for 1 messages on WarningFilter(None,Right(Hostname verification failed),false)
	at akka.testkit.EventFilter.intercept(TestEventListener.scala:119)
	at akka.http.scaladsl.ClientServerSpec.$anonfun$new$62(ClientServerSpec.scala:611)
	at akka.stream.testkit.scaladsl.StreamTestKit$.assertAllStagesStopped(StreamTestKit.scala:32)
	at akka.stream.testkit.Utils$.assertAllStagesStopped(Utils.scala:24)
	at akka.http.scaladsl.ClientServerSpec.$anonfun$new$61(ClientServerSpec.scala:588)
```

johanandren:
Also on 2.11 in a nightly this weekend:
https://jenkins.akka.io:8498/job/akka-http-nightly-release-10.1/AKKA_VERSION=default,Node=akka-http-nightly-node,SCALA_VERSION=2.11.12,jdk=AdoptOpenJDK%208/944/testReport/junit/akka.http.scaladsl/ClientServerSpec/The_low_level_HTTP_infrastructure_should_complete_a_request_response_when_request_has__Connection__close__set/

johanandren:
And 2.12
https://jenkins.akka.io:8498/job/akka-http-nightly/AKKA_VERSION=master,Node=akka-http-nightly-node,SCALA_VERSION=2.12.11,jdk=AdoptOpenJDK%208/190/

jrudolph:
[...]

PR #3456, opened 2020-09-01
core: reenable lingerTimeout

### LINKED ISSUE #3452: `CloseDelimited` or `Connection: close` responses may cancel connections prematurely
As observed in https://github.com/akka/akka-http/issues/2532.

In that case, the server sends a very big chunk of data and sends completion to the network afterwards. Basically, the high-level part of the server is done at that point. There's nothing to do for it, so after the cancellation timeout also the downstream is cancelled. However, that cancellation is then ultimately interpreted as a request to reset the connection. So, if the client is too slow receiving the data (which is simulated in this test because it uses small network buffers and TLS) the connection is forced to be closed at some point.

The problem here is that a higher layer is basically done with its work but still would like to monitor that the lower layers are cleaned up at some point. However, since there's no way to monitor actual progress of the big chunk of data being transmitted to the other side on the lower layers this is hard to achieve.

How could we handle this more gracefully?

Here are a few alternatives:

 * Recommend sending "reasonably sized" chunks in relation to network speeds and buffers, so that completion is delayed until the last chunk is actually going to be sent in a reasonable time frame.
 * Try to monitor progress more closely at the end of connections by sending empty canary chunks at the end of connections with the intention to delay completion until the point where all the intermediate buffers have space after a potential big last chunk.
 * Try to introduce a more graceful cancellation option that prevents forcing a connection to close. That could work with Akka 2.6, by adding a special kind of cancellation cause that could be interpreted as a more graceful way of closing a connection. All layers that have still data buffered could try to send out the data and only shutdown afterwards.
 * Prevent complete shutdown of higher-level components completely by avoiding the use of `completeStage`. Instead, only completion would be used in most places and we would rely on the lowest layers (= TCP or idle timeout above) to track progress and close the connection when all work is done (or cancel if progress stalls or is too slow).

jrudolph:
Fixes #3452, #2532

In 294df230af0e1e1b3a7aba575975a634bd017050, the stream-cancellation-delay was
introduced. However, at the same time I removed the lingerTimeout functionality.
While both features are related, they serve different purposes:

 * `streamCancellationDelay` tries to avoid cancellation races by giving a
   stream graph more time to propagate errors/completion on the forward path.
   The default is 100 millis because that suffices in most conditions.
 * `lingerTimeout` gives clients more time to read a response slowly when the
   server is finished with a connection. lingerTimeout is set to a more generous
   1 minute, because it is only a safe-guard that a connection, that isn't "managed"
   any more by the server (because the server stack was already done), is
   still ultimately closed if the client is too slow (or stuck at all).

We backported the same change on 10.1.13, so it needs to be fixed there as well.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
@@ -9,7 +9,7 @@ import java.util.concurrent.atomic.AtomicReference
 import scala.concurrent.{ Future, Promise }
 import scala.concurrent.duration.{ Deadline, Duration, FiniteDuration }
 import scala.collection.immutable
-import scala.util.control.NonFatal
+import scala.util.control.{ NoStackTrace, NonFatal }
 import akka.NotUsed
 import akka.actor.Cancellable
 import akka.annotation.InternalApi
@@ -120,6 +120,16 @@ private[http] object HttpServerBluePrint {
           entitySource.complete()
         }
         completeStage()
+      }
+
+      override def onUpstreamFinish(): Unit = super.onUpstreamFinish()
+      override def onUpstreamFailure(ex: Throwable): Unit = {
+        if (entitySource ne null) {
+          // application layer has cancelled or only partially consumed response entity:
+          // connection will be closed
+          entitySource.fail(ex)
+        }
+        super.onUpstreamFailure(ex)
       }
 
       override def onPush(): Unit = grab(in) match {
@@ -423,7 +433,9 @@ private[http] object HttpServerBluePrint {
         def onPull(): Unit =
           if (oneHundredContinueResponsePending) pullSuppressed = true
           else if (!hasBeenPulled(requestParsingIn)) pull(requestParsingIn)
-        override def onDownstreamFinish() = cancel(requestParsingIn)
+        override def onDownstreamFinish(): Unit =
+          if (openRequests.isEmpty) completeStage()
+          else failStage(new IllegalStateException("User handler flow was cancelled with ongoing request") with NoStackTrace)
       })
 
       setHandler(httpResponseIn, new InHandler {
@@ -484,10 +496,7 @@ private[http] object HttpServerBluePrint {
             case IllegalUriException(errorInfo) =>
               finishWithIllegalRequestError(StatusCodes.BadRequest, errorInfo)
 
-            case _: ServerTerminationDeadlineReached =>
-              val response = settings.terminationDeadlineExceededResponse
-              log.warning(s"Closing connection with [{}] response, termination deadline exceeded. Server is shutting down...", response.status)
-              emitErrorResponse(response)
+            case ex: ServerTerminationDeadlineReached => failStage(ex)
 
             case NonFatal(e) =>
               log.error(e, "Internal server error, sending 500 response")
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/server/ServerTerminator.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/server/ServerTerminator.scala
@@ -260,8 +260,9 @@ private[http] final class GracefulTerminatorStage(settings: ServerSettings)
       })
 
       def installTerminationHandlers(deadline: Deadline): Unit = {
-        // when no inflight requests, complete stage right away
-        if (!pendingUserHandlerResponse) completeStage()
+        // when no inflight requests, fail stage right away, could probably be a complete
+        // when https://github.com/akka/akka-http/issues/3209 is fixed
+        if (!pendingUserHandlerResponse) failStage(new ServerTerminationDeadlineReached)
 
         setHandler(fromUser, new InHandler {
           override def onPush(): Unit = {
--- a/akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
@@ -15,9 +15,9 @@ import akka.stream.impl.fusing.GraphInterpreter
 import akka.stream.impl.fusing.GraphStages.SimpleLinearGraphStage
 import akka.stream.scaladsl._
 import akka.stream.stage._
-import akka.util.ByteString
+import akka.util.{ ByteString, OptionVal }
 
-import scala.concurrent.duration.FiniteDuration
+import scala.concurrent.duration.{ Duration, FiniteDuration }
 import scala.concurrent.{ ExecutionContext, Future, Promise }
 import scala.util.Failure
 import scala.util.Success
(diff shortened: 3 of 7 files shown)


raboof (approved):
Code looks straightforward enough, not sure about the test failures though?

jrudolph:
Already looking into the test failures, they are real assumptions that are not met any more.

akka-http-core/src/test/scala/akka/http/scaladsl/GracefulTerminationSpec.scala
```diff
@@ -75,35 +75,38 @@ class GracefulTerminationSpec
...
+              .throttle(1, 100.millis)
+            reply(_ => HttpResponse(entity = HttpEntity.Chunked(ContentTypes.`text/plain(UTF-8)`, chunks)))
+
+            // start reading the response
+            val responseEntity = r1.futureValue.entity.dataBytes
+              .via(Framing.delimiter(ByteString(","), 20))
+              .runWith(TestSink.probe[ByteString])(SystemMaterializer(clientSystem).materializer)
+            responseEntity.requestNext().utf8String should ===("reply1")
+
+            val termination = serverBinding.terminate(hardDeadline = 1.second)
```
jrudolph, line 99: Basically, now the delay between calling `terminate` and the hardDeadline is basically broken if only the response entity hasn't been streamed out. This never worked correctly (#3209) but some delay was introduced but for other reasons that have now been taken out of the picture.

akka-http-core/src/test/scala/akka/http/scaladsl/GracefulTerminationSpec.scala
```diff
@@ -75,35 +75,38 @@ class GracefulTerminationSpec
       ex.getMessage should include("Connection refused")
     }
 
-    "fail chunked response streams" in new TestSetup {
+    "fail chunked response streams" in {
+      // client system created on the outside, so we simulate that client is not shut down at the same time as server
```
jrudolph, line 79: A bit of reorganization here to be able to add `assertAllStagesStopped`.

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -122,6 +122,16 @@ private[http] object HttpServerBluePrint {
         completeStage()
       }
 
+      override def onUpstreamFinish(): Unit = super.onUpstreamFinish()
+      override def onUpstreamFailure(ex: Throwable): Unit = {
```
jrudolph, line 126: Before the stage was just closed and the entity source potentially left hanging. There seemed to be some kind of behavior which actually prevented that but let's be rather be explicitly and propagate the failure if we are not able to send any more data to it.

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -423,7 +433,9 @@ private[http] object HttpServerBluePrint {
         def onPull(): Unit =
           if (oneHundredContinueResponsePending) pullSuppressed = true
           else if (!hasBeenPulled(requestParsingIn)) pull(requestParsingIn)
-        override def onDownstreamFinish() = cancel(requestParsingIn)
+        override def onDownstreamFinish(): Unit =
+          if (openRequests.isEmpty) completeStage()
+          else failStage(new IllegalStateException("User handler flow was cancelled with ongoing request") with NoStackTrace)
```
jrudolph, line 438: This new behavior is more aggressive than before: if the user handler cancels the request stream, we will kill the connection. Would be nice to be able to get rid of that change as it's somewhat of an invitation of new race conditions. In practice it wouldn't probably not be so serious as user-handlers usually do not complete explicitly.

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -484,10 +496,7 @@ private[http] object HttpServerBluePrint {
             case IllegalUriException(errorInfo) =>
               finishWithIllegalRequestError(StatusCodes.BadRequest, errorInfo)
 
-            case _: ServerTerminationDeadlineReached =>
-              val response = settings.terminationDeadlineExceededResponse
-              log.warning(s"Closing connection with [{}] response, termination deadline exceeded. Server is shutting down...", response.status)
-              emitErrorResponse(response)
+            case ex: ServerTerminationDeadlineReached => failStage(ex)
```
jrudolph, line 499: The previous behavior is part of the problem described in #3209: in the controller error handler we try to handle errors gracefully by trying to converts errors into a response + completion. In the renderer that extra (and not asked-for) response is then stuck while a response entity is streamed out. That means there's (too?) little control from stages on top of the stack like the user-handler or the GracefulTerminationStage to effectively kill bring down a connection.

akka-http-core/src/main/scala/akka/http/impl/engine/server/ServerTerminator.scala
```diff
@@ -260,8 +260,9 @@ private[http] final class GracefulTerminatorStage(settings: ServerSettings)
       })
 
       def installTerminationHandlers(deadline: Deadline): Unit = {
-        // when no inflight requests, complete stage right away
-        if (!pendingUserHandlerResponse) completeStage()
+        // when no inflight requests, fail stage right away, could probably be a complete
+        // when https://github.com/akka/akka-http/issues/3209 is fixed
+        if (!pendingUserHandlerResponse) failStage(new ServerTerminationDeadlineReached)
```
jrudolph, line 265: Another part of the puzzle in #3209: when `!pendingUserHandlerResponse` but there's an ongoing response stream, we completed before which is just to tame to have any effect. Now we bring the connection down immediately even before the hard deadline in such a case.

akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
```diff
@@ -191,6 +191,56 @@ private[http] object StreamUtils {
     }
   }
 
+  /**
+   * INTERNAL API
+   *
+   * Returns a flow that is almost identity but delays propagation of cancellation from downstream to upstream.
+   */
+  def delayCancellation[T](cancelAfter: Duration): Flow[T, T, NotUsed] = Flow.fromGraph(new DelayCancellationStage(cancelAfter))
```
jrudolph, line 199: This and below is just a revert of the previous removal of the `lingerTimeout` functionality.

jrudolph:
Added a few explaining comments above. After all I didn't try to fix #3209 but had to add more small fixes to the server logic than I would have liked. All of that is somewhat aggravated that this fix would technically also need a backport to 10.1 :fearful:

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
raboof: Right - not as nice, but indeed perhaps better than having 2 responses for the same request floating around.

raboof (approved):
Good to have the `lingerTimeout` back. Tricky to review for unintended consequences in the cancellation/failure behaviour, but logically makes sense

ennru (approved):
LGTM.

PR #3462, opened 2020-09-03
http2: ignore unexpected DATA frames in state closed

### LINKED ISSUE #3461: HTTP/2 drops a connection if frames are received on a stream that was recently closed
As observed in https://github.com/akka/akka-grpc/issues/1081.

There are at least some legitimate reasons why frames would arrive on a stream that was recently closed, especially if our side cancelled itself, moving the state to "Closed" immediately.

A simple fix would be to just fail less and ignore more packets received on closed streams. A more elaborate scheme would introduce a cool-down period for a stream when it gets closed in which we ignore frames and only move to a more strict closed state later. Not sure we care enough to force connections to be closed as we do now?

jrudolph:
Fixes #3461

Those frames are somewhat expected at least in the case where we just sent a
RST_STREAM frame and the peer might not have yet processed it.

Of course, it might also be an indication of a misbehaving peer but as long
as there's no indication that this can lead to a non-negligible overhead on
our side, we might be fine with this simple solution.


raboof (approved):

ennru (approved):
LGTM.

Issue #3391, opened 2020-07-23
Make SSLSession information available as an attribute


raboof:
Rather than through `Tls-Session-Info`

jrudolph:
Yep, good point.

PR #3472, opened 2020-09-08
Expose the SslSession via an attribute

raboof:
Refs #3391


akka-http-core/src/main/scala/akka/http/scaladsl/model/SslSession.scala
```diff
@@ -0,0 +1,48 @@
...
+
+package akka.http.scaladsl.model
+
+import java.security.Principal
+import java.security.cert.Certificate
+
+import javax.net.ssl.{ SSLPeerUnverifiedException, SSLSession }
+
+/** Value class to allow access to an SSLSession with Scala types */
+class SslSession(val session: SSLSession) extends AnyVal {
```
jrudolph, line 13: Do we know the compatibility constraints a value class will impose? Not sure we would gain a lot here with a value class, so maybe this could be a trait instead?

akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpResponseParser.scala
```diff
@@ -122,11 +124,17 @@ private[http] class HttpResponseParser(protected val settings: ParserSettings, p
...
-                                  expect100continue: Boolean, hostHeaderPresent: Boolean, closeAfterResponseCompletion: Boolean): StateResult = {
+                                  expect100continue: Boolean, hostHeaderPresent: Boolean, closeAfterResponseCompletion: Boolean,
+                                  sslSession: SSLSession): StateResult = {
 
     def emitResponseStart(
       createEntity: EntityCreator[ResponseOutput, ResponseEntity],
       headers:      List[HttpHeader]                              = headers) = {
+
+      val attributes: Map[AttributeKey[_], Any] =
+        if (settings.includeSslSessionAttribute) Map(AttributeKeys.sslSession -> new SslSession(sslSession))
```
jrudolph, line 135: :+1: didn't even remember it's even supported on the client side.

akka-http-core/src/main/java/akka/http/javadsl/model/AttributeKeys.java
```diff
@@ -6,9 +6,13 @@
...
 import akka.http.javadsl.model.ws.WebSocketUpgrade;
 
+import javax.net.ssl.SSLSession;
+
 public final class AttributeKeys {
     public static final AttributeKey<RemoteAddress> remoteAddress =
             (AttributeKey<RemoteAddress>)(Object)akka.http.scaladsl.model.AttributeKeys.remoteAddress();
     public static final AttributeKey<WebSocketUpgrade> webSocketUpgrade =
             (AttributeKey<WebSocketUpgrade>)(Object)akka.http.scaladsl.model.AttributeKeys.webSocketUpgrade();
+    public static final AttributeKey<SSLSession> sslSession =
```
jrudolph, line 16: This cast is so sneaky that I don't think it will even work ;)

If a value class instance is used in a generic environment, it will be/stay boxed, so retrieving the value like this will likely fail.

jrudolph (commented):
Mostly LGTM, some doubts about the value-class approach.

akka-http-core/src/main/java/akka/http/javadsl/model/AttributeKeys.java
raboof: You're quite right ;). I guess we have to wrap the session for both Java and Scala.

akka-http-core/src/main/scala/akka/http/javadsl/model/SslSession.scala
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright (C) 2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.http.javadsl.model
+
+import javax.net.ssl.SSLSession
+
+trait SslSession {
```
ennru, line 9: A bit hairy to have `SslSession` and `SSLSession` in the same context.
Would `SslSessionValue` be too ugly?

ennru (approved):
Looks regular and consistent to me.

akka-http-core/src/main/scala/akka/http/scaladsl/model/SslSession.scala
```diff
@@ -0,0 +1,28 @@
...
+ * Copyright (C) 2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.http.scaladsl.model
+
+import akka.http.javadsl.{ model => jm }
+import akka.stream.scaladsl.ScalaSessionAPI
+import javax.net.ssl.SSLSession
+
+class SslSession(val session: SSLSession) extends jm.SslSession with ScalaSessionAPI {
```
jrudolph, line 11: What's the benefit of wrapping the session at all?

akka-http-core/src/test/java/akka/http/javadsl/model/HttpMessageTest.java
```diff
@@ -52,4 +52,9 @@ public void testAttributeKeys() {
 
         assert(newRequest.getAttribute(AttributeKeys.remoteAddress).get().equals(remoteAddress));
     }
+
+    @Test
+    public void testSslAttribute() {
+
```
jrudolph, line 58: Test still missing? :)

jrudolph (commented):
Some comments

akka-http-core/src/main/scala/akka/http/scaladsl/model/SslSession.scala
raboof: To get the `ScalaSessionAPI` scala accessors... I guess we could do without them and expose the SSLSession directly, but then users would have to deal with that java-collection-based null-returning API...

jrudolph: Ah missed that. Makes sense to keep that I guess.

akka-http-core/src/main/scala/akka/http/javadsl/model/SslSession.scala
jrudolph: I'd prefer that as well.

raboof: Hmm, 'value' is a bit ugly - maybe `SslSessionInfo`?

jrudolph: Yep, better and more in line with what we had before.

jrudolph (approved):
LGTM

What do we think about that the naming thing that `sslSession` attribute (and `ssl-session-attribute`) now returns a `SslSessionInfo`? I'd say it's fine even if the others attribute key names correspond more tightly with the class names.

raboof:
> What do we think about that the naming thing that `sslSession` attribute (and `ssl-session-attribute`) now returns a `SslSessionInfo`? I'd say it's fine even if the others attribute key names correspond more tightly with the class names.

I agree

Issue #3429, opened 2020-08-10
Failed: org.pegdown.ParsingTimeoutException while building paradox documentation on travis


jrudolph:
Often happens on master builds.

E.g. here: https://travis-ci.org/github/akka/akka-http/jobs/716500980

This might just be a too tight timeout. I've seen it at least twice today.

jrudolph:
Still happening afterwards: https://travis-ci.org/github/akka/akka-http/jobs/724154843

jrudolph:
This time it was for "custom-directives.md".

In #3430 the timeout was changed to 10 seconds and I think that's per file (?) which seems reasonable. I think we haven't understood why the processing times would be so long or so volatile.

jrudolph:
path-directives.md

https://travis-ci.org/github/akka/akka-http/jobs/725277631

marshalling.md

https://travis-ci.org/github/akka/akka-http/jobs/725253509

http2.md

https://travis-ci.org/github/akka/akka-http/jobs/725254337

Seems not really correlated to the file so far.

jrudolph:
This still happens regularly. I wonder now if there are just too many things running in parallel when that happens? Or a GC / memory problem?

raboof:
Good question.

What should we do to find out? Enable GC logging and show a stack dump when we see this exception?

PR #3487, opened 2020-09-29
Enable gc logging when generating docs

raboof:
To track down #3429


jrudolph (approved):
Thanks, let's see if it shows something interesting.

Issue #3427, opened 2020-08-10
Route => Flow used to need just an implicit Materializer now it requires a ClassicActorSystemProvider


jrudolph:
E.g. see https://github.com/akka/akka-minion/pull/174#r466863894

So, on one hand requiring the system is the more "honest" requirement because, in fact, before we required the materializer to provide a system which is kind of backbwards.

On the other hand, it would be nice to be more source compatible, so if we are doing a 10.2.1 early enough we could try to offer the old functionality implicit conversion as well.

PR #3489, opened 2020-09-29
http: provide implicit conversion from route and materializer to flow

raboof:
For better backwards compatibility with 10.1.x #3427

I think putting it in a superclass like this should avoid any
ambiguous import problems, but a second pair of eyes would not
hurt here.

### DIFF
--- a/akka-http/src/main/scala/akka/http/scaladsl/server/RouteResult.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/server/RouteResult.scala
@@ -9,7 +9,7 @@ import akka.actor.ClassicActorSystemProvider
 import akka.http.javadsl
 import akka.http.scaladsl.model.{ HttpRequest, HttpResponse }
 import akka.http.scaladsl.settings.{ ParserSettings, RoutingSettings }
-import akka.stream.Materializer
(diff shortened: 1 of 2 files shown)


jrudolph (approved):
LGTM

akka-http/src/main/scala/akka/http/scaladsl/server/RouteResult.scala
```diff
@@ -72,3 +72,17 @@ object RouteResult {
...
+sealed abstract class LowerPriorityRouteResultImplicits {
+  /**
+   * Turns a `Route` into a server flow.
+   *
+   * This implicit conversion is defined here because `Route` is an alias for
+   * `RequestContext => Future[RouteResult]`, and the fact that `RouteResult`
+   * is in that type means this implicit conversion come into scope whereever
+   * a `Route` is given but a `Flow` is expected.
+   */
+  implicit def routeToFlowViaMaterializer(route: Route)(implicit materializer: Materializer): Flow[HttpRequest, HttpResponse, NotUsed] =
```
jrudolph, line 85: One thing I wonder about, is if we should deprecate it right away, so we can remove it again in the next version?

ennru (approved):
LGTM.

Issue #2507, opened 2019-04-13
Uri.collapseDotSegments fails on some relative paths


yonran:
It seems that when `Uri` is given a relative URI with `..`s after the first `Segment` of a relative path, the `Uri` becomes incorrectly normalized.

Testing with `"com.typesafe.akka" %% "akka-http" % "10.1.8"`, the last two `should be`s fail (the actual value is in the comment):

```scala
import akka.http.scaladsl.model.Uri
import akka.http.scaladsl.model.Uri.Path
import org.scalatest.{FlatSpec, Matchers}

class UriTest extends FlatSpec with Matchers {
  "Uri" should "normalize dot dot" in {
    Uri(scheme = "file", path = Path("a/./b")) should be (Uri(scheme = "file", path = Path("a/b")))  // ok
    Uri(scheme = "file", path = Path("../c")) should be (Uri(scheme = "file", path = Path("c")))  // ok
    Uri(scheme = "file", path = Path("a/b/c/../../d")) should be (Uri(scheme = "file", path = Path("a/d")))  // ok
    Uri(scheme = "file", path = Path("/a/../b")) should be (Uri(scheme = "file", path = Path("/b")))  // ok
    Uri(scheme = "file", path = Path("a/../../c")) should be (Uri(scheme = "file", path = Path("c")))  // c/..//c
    Uri(scheme = "file", path = Path("a/../b")) should be (Uri(scheme = "file", path = Path("b")))  // b//b
  }
}
```

jrudolph:
Interesting. Thanks for the report, @yonran (and sorry for the delay).

Would you want to open a PR to fix that?

jrudolph:
Fixed in #3490

PR #3490, opened 2020-09-29
core: Uri.collapseDotSegments fails on some relative paths #2507

Yash0215:
Uri.collapseDotSegments fails on some relative paths #2507
Description:
the conclusion is that the 'collapseDotSegments' doesnt work as described in 'http://tools.ietf.org/html/rfc3986#section-5.2.4' when relative path doent start with a '/'. the easy solution is to check whether the given Uri starts with a '/' or not. and call 'process' accordingly.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/Uri.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/Uri.scala
@@ -866,6 +866,7 @@ object Uri {
   }
 
   private[http] def collapseDotSegments(path: Path): Path = {
+    import Path._
     @tailrec def hasDotOrDotDotSegment(p: Path): Boolean = p match {
       case Path.Empty => false
       case Path.Segment(".", _) | Path.Segment("..", _) => true
@@ -873,7 +874,6 @@ object Uri {
     }
     // http://tools.ietf.org/html/rfc3986#section-5.2.4
     @tailrec def process(input: Path, output: Path = Path.Empty): Path = {
-      import Path._
       input match {
         case Path.Empty                       => output.reverse
         case Segment("." | "..", Slash(tail)) => process(tail, output)
@@ -889,7 +889,9 @@ object Uri {
         case Segment(string, tail)     => process(tail, string :: output)
       }
     }
-    if (hasDotOrDotDotSegment(path)) process(path) else path
+    if (hasDotOrDotDotSegment(path))
+      if (path.startsWithSlash) process(path) else process(Slash(path)).tail
+    else path
   }
 
   private[http] def fail(summary: String, detail: String = "") = throw IllegalUriException(summary, detail)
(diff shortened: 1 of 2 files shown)


lightbend-cla-validator:
Hi @Yash0215,

jrudolph:
Thanks, @Yash0215. This is looking good. I wonder where those URIs would turn up though, as file URIs in particular don't support relative paths according to https://tools.ietf.org/html/rfc8089.

That said, http://tools.ietf.org/html/rfc3986#section-5.2.4 which you refer to above, has an explicit example using a relative path, so it would probably make sense to support it correctly.

Yash0215:
such types of URI raises "IllegalUriException" when host is present but are allowed when authority is empty for some reason. so I thought it is better to support it correctly instead of producing unexpected results.

raboof:
Running the `UriParserBenchmark` this does seem to have a somewhat visible impact:

`jmh:run -i 3 -wi 3 -f 1 .*UriParserBenchmark`

Before:
```
[info] Iteration   1: 621.705 ops/ms
[info] Iteration   2: 625.823 ops/ms
[info] Iteration   3: 627.470 ops/ms
[info] Iteration   1: 284.254 ops/ms
[info] Iteration   2: 282.141 ops/ms
[info] Iteration   3: 282.596 ops/ms

[info] Iteration   1: 608.121 ops/ms
[info] Iteration   2: 638.491 ops/ms
[info] Iteration   3: 635.331 ops/ms

[info] Iteration   1: 282.200 ops/ms
[info] Iteration   2: 282.684 ops/ms
[info] Iteration   3: 278.946 ops/ms
```

After:
```
[info] Iteration   1: 612.108 ops/ms
[info] Iteration   2: 610.006 ops/ms
[info] Iteration   3: 618.760 ops/ms
[info] Iteration   1: 277.478 ops/ms
[info] Iteration   2: 276.727 ops/ms
[info] Iteration   3: 275.033 ops/ms

[info] Iteration   1: 618.215 ops/ms
[info] Iteration   2: 615.976 ops/ms
[info] Iteration   3: 624.117 ops/ms
[info] Iteration   1: 268.670 ops/ms
[info] Iteration   2: 272.804 ops/ms
[info] Iteration   3: 269.275 ops/ms
```

Now this is of course a bit of a microbenchmarks and as such to be taken with a grain of salt, but as we don't really seem to have a strong use case for this feature perhaps we should not take the hit?

jrudolph:
Sorry for keeping this around for so long.

I cannot see any significant performance difference. On my machine with JDK 8, I can see an (at least) bimodal distribution of running times when using higher fork values for JMH (as you should always run it with). As so often the Hotspot not always arrives at the same steady JIT-generated code, which seems to make a big difference here. In any case, the difference between before and after seems to be small and not definitely in one direction, so this is fine for me the be included.

jrudolph (approved):
LGTM

jrudolph:
(I refreshed the original branch)

PR #3507, opened 2020-10-07
Don't upgrade akka via scala-steward

raboof:


jrudolph (approved):
LGTM, ah funny that this is how it happened.

PR #3508, opened 2020-10-07
build: really increase paradox parsing timeout

### LINKED ISSUE #3429: Failed: org.pegdown.ParsingTimeoutException while building paradox documentation on travis

jrudolph:
Fixes #3429

Previously in #3430 the wrong key was updated so that the change had no effect and the timeout was still 2 seconds which is arguably slow enough to trip over now and then.

### DIFF
--- a/project/ParadoxSupport.scala
+++ b/project/ParadoxSupport.scala
@@ -18,13 +18,7 @@ import scala.io.{Codec, Source}
 
 object ParadoxSupport {
   val paradoxWithCustomDirectives = Seq(
-    paradoxDirectives ++= Def.taskDyn {
-      Def.task { Seq(
-        { context: Writer.Context =>
-            new SignatureDirective(context.location.tree.label, context.properties, context)
-        },
-      )}
-    }.value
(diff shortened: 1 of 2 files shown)


project/ParadoxSupport.scala
```diff
@@ -18,13 +18,7 @@ import scala.io.{Codec, Source}
 
 object ParadoxSupport {
   val paradoxWithCustomDirectives = Seq(
-    paradoxDirectives ++= Def.taskDyn {
```
jrudolph, line 21: What kind of pattern is that weird `taskDyn { task { } }`? Is that somehow imposing some execution ordering? But seems not useful in this particular case (I've seen it used in paradox it seems to delete target folders before repopulating them with another task).

raboof: I think that was there previously to get access to the logger, but that's no longer needed since c9f3e55c86c5a5f90bca855296c4676df6d0ba59

raboof (approved):

Issue #3230, opened 2020-06-09
HTTP/2 client: public connection-level API


jrudolph:
It could be made available under the signature of `Http().outgoingConnection`, but as a user, you would have to know that the responses will come out unordered. So, maybe it would make sense to introduce another API that would clearly state that constraint for both HTTP/1.1 and HTTP/2.

Make sure to add documentation and examples for public APIs.

jrudolph:
We currently already have `Http2.outgoingConnection` but this should stay internal.

As stated above, we could retrofit `Http().outgoingConnection` with HTTP/2 functionality but that would change semantics because with HTTP/2  multiple requests can be ongoing and the strict ordering of requests and responses is not given any more.

When you don't have fixed ordering between request and responses you need another way of associating requests with responses.

Our current idea for how to associate requests with responses can be seen in Http2ClientServerSpec: you can create custom attributes with a value deriving from `RequestResponseAssociation` in which case the implementation will make sure carry over those attribute from request to response.

This could work the same for both HTTP/1.1 and HTTP/2 if we want to have a single entrypoint that serves both.

(A basic wrapper usage of that could be a non-pooled version of `singleRequest` by putting a queue in front of the connection and adding a promise to each request as such an attribute that would be completed for each response coming out of the connection.

For HTTP/1.1 such a usage wouldn't be to useful because of the throughput constraints of a single connection (compared to a pool) but for HTTP/2 it would be enough for many use cases.

That said, the pool has other benefits, so it will still make sense to provide more above that. See #3231)

jrudolph:
Done in #3511.

PR #3511, opened 2020-10-08
core: HTTP/2 connection level API

johanandren:
References #3230

Drew inspiration from the new server API and tried to cover all previous variants of outgoingConnection.


akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
```diff
@@ -458,6 +458,23 @@ class HttpExt private[http] (private val config: Config)(implicit val system: Ex
...
+   * so therefore requests needs to have a [[akka.http.scaladsl.model.http2.RequestResponseAssociation]]
+   * which Akka HTTP will carry over to the corresponding response for a request.
+   *
+   * FIXME: the association lives in the http2 model
+   * FIXME: do want something even more convenient for the normal case like `connectionTo("http://host")` and use the
+   *        protocol to figure out port + .withPort for custom cases?
+   *
+   * @return A builder to configure more specific setup for the connection and then build a `Flow[Request, Response, Future[OutgoingConnection]]`.
+   */
+  def connectionTo(host: String, port: Int): SingleConnectionBuilder
```
johanandren, line 476: Should it be even more explicit? `unorderedConnectionTo`?

akka-http-core/src/main/scala/akka/http/scaladsl/OutgoingConnectionBuilder.scala
```diff
@@ -0,0 +1,110 @@
...
+
+    override def withLocalAddress(localAddress: InetAddress): OutgoingConnectionBuilder = copy(localAddress = Some(localAddress))
+
+    override def withConnectionContext(context: ConnectionContext): OutgoingConnectionBuilder = copy(connectionContext = context)
+
+    override def withClientConnectionSettings(settings: ClientConnectionSettings): OutgoingConnectionBuilder = copy(clientConnectionSettings = settings)
+
+    override def logTo(logger: LoggingAdapter): OutgoingConnectionBuilder = copy(log = logger)
+
+    override def connectionFlow(): Flow[HttpRequest, HttpResponse, Future[OutgoingConnection]] = ???
```
johanandren, line 108: Some questions:
 * How would the actual protocol negotiation work?
 * Should the user be able to explicitly choose between HTTP/1.1 and HTTP/2 as well?

raboof: I like the builder approach, indeed consistent with the new server API's, and gives the opportunity to introduce additional options later. Might get tricky when some options are mutually exclusive, I guess, but OK.

I definitely think it's useful to be able to explicitly choose between various connection/negotiation types, but maybe just start with one?

I agree it's not super obvious the `Flow[HttpRequest, HttpResponse]` is *not* ordered. Perhaps rather than renaming `connectTo` we could rename `connectionFlow` to something that conveys it's unordered? Not sure of a great name though.

An HTTP2 connection can have a max number of concurrent streams (https://tools.ietf.org/html/rfc7540#section-6.5.2). Since each `HttpResponse` corresponds to a stream, you could in theory get stuck if you're not consuming the HttpResponses. Would that cause trouble if you're for example trying to populate a tree-like structure with a depth-first streaming approach? Or nudge you towards a better approach anyway? Not sure, but this seems like a good start :)

johanandren: One more question, akka-http2-support depends on akka-http-core where Http lives, so it is not possible to just add a dependency to get access to the HTTP/2 classes. Is the long term goal to keep them as separate modules or to merge them or how do we make using HTTP/2 more seamless moving forward?

johanandren: Moved it to the http2-support module for now

raboof: For those reading along: for the server-side Http2 support, the actual implementation is in `Http2Ext` in the `akka-http2-support` module, available via `Http2()`.

We use a reflection trick to use that implementation from the regular `Http` extension when the module is available and enabled. We could possibly do the same for the client API. For now having it just in the http2-support module might be the quickest way to iterate, though.

jrudolph: See also #3514

jrudolph (commented):
Creating a builder-like interface for clients as well is an interesting idea.

The original ticket was also about how to integrate HTTP/1.1 and HTTP/2 connection-level interfaces. What do we currently think about that? Our options seem to be:

 * don't support an integrated interface for HTTP/1.1 and HTTP/2 on this lowest level, i.e. just add another entry point for HTTP/2 connections
 * literally reuse the existing `Http().outgoingConnection` and try to find a way how to specify that HTTP/2 should be used. In that case, we need to make sure that it is clear that the behavior changes to "unordered" for HTTP/2 connections (somewhat unsatisfying for that reason).
 * create a new entry-point that supports both HTTP/1.1 and HTTP/2 at the same time (this could be a builder type interface, though we should not overdo it, as we expect the main user interface to be one based on `Http().singleRequest`)

If we want to support an integrated interface
 * How would the user specify the preference for HTTP/2 or HTTP/1.1?
 * Do we want to support protocol negotiation on that level? (This is less obvious than one could think: a single HTTP/2 connection is more like a whole HTTP/1.1 connection pool, so maybe the negotiation support should live on a higher level in the pool implementation?)

(Btw. we consider akka-http2-support completely internal API, so everything should ultimately go into the main `Http` in akka-http-core)

johanandren:
Pushed another iteration of this that perhaps makes it more clear how I thought the builder API should be the same for HTTP/1.1 and HTTP/2, hooking it in using the pre-existing shadow-hack that I didn't notice earlier. So the latest is basically your option 3 @jrudolph 

Some thoughts:
 * with negotiated automagic protocol making the starting point minimal (`connectTo(host)`) requires defaulting to TLS and 443, which can then be switched away from using the builder. Given that many if not most HTTP/1.1 endpoints also are HTTPS nowadays and this is a super user api so perhaps that is ok?
 * when the http2 support is not available/not enabled we'd need a HTTP/1.1-only impl of the builder, did not go there yet because wanted to see that this is a reasonable direction first

I think if we would deem it important to force a protocol we can add that in the builder.

raboof:
> with negotiated automagic protocol making the starting point minimal (connectTo(host)) requires defaulting to TLS and port 443, which can then be switched away from using the builder. Given that many if not most HTTP/1.1 endpoints also are HTTPS nowadays and this is a super user api so perhaps that is ok?

It's a bit weird since the server-side is unsecured by default, but there that makes sense since there you need to select the server certificate anyway. Agree it might be OK.

> when the http2 support is not available/not enabled we'd need a HTTP/1.1-only impl of the builder, did not go there yet because wanted to see that this is a reasonable direction first

I guess we could use the same builder and throw if you don't have http2 enabled but are selecting a variation that *needs* http2?

For the 'unsecured' case there are a couple of options:
* Send the first request over http1, using the `Upgrade` headers to allow the server to decide whether to continue with http1 or http2
* First do a synthetic 'pre-flight' http1 request to find out whether the server allows http2, and only send the `HttpRequest` from the user once you know
* Just start talking HTTP/2 without headers (but this will fail when the server doesn't actually support it)

Perhaps we should start with providing the first and third from this list. How should that look like in the API?

johanandren:
What about the current `unsecure()` for upgrade (which should always work but has a slight performance overhead) and explicit `unsecure(forceHttp2 = true)` for HTTP/2 without headers, plus throwing if you try to do that when HTTP/2 support is not available on classpath/disabled?

raboof:
> What about the current `unsecure()` for upgrade (which should always work but has a slight performance overhead)

Sounds good

> and explicit `unsecure(forceHttp2 = true)` for HTTP/2 without headers, plus throwing if you try to do that when HTTP/2 support is not available on classpath/disabled?

Or maybe even a completely separate method?

jrudolph (commented):
My main concern is that we introduce a somewhat complex API here, that we don't expect or want to be used.

Most importantly, let's not do any automatic negotiation at those levels for now.

The easiest would be just to offer a `Http.outgoingConnectionHttp2` which would probably work for its intended low-level use-cases but will suffer from binary compatibility issues if we need to evolve it.

So, maybe a builder API is the way to go but if in doubt, let's make it very explicit and don't try to guess anything.

Which connection options do we want to support?

 * [x] plain text HTTP/1.1
 * [x] HTTP/1.1 over TLS
 * [x] HTTP/2 over TLS
 * [ ] HTTP/2 over TLS with HTTP/1.1 fallback if not supported by server (not now)
 * [ ] plain text HTTP/2 with Upgrade
 * [ ] plain text HTTP/2 with prior knowledge

Do we need any of the latter ones for gRPC purposes? Let's only do the uncommon ones if strictly necessary.

akka-http-core/src/main/scala/akka/http/javadsl/OutgoingConnectionBuilder.scala
```diff
@@ -0,0 +1,75 @@
...
+  def logTo(logger: LoggingAdapter): OutgoingConnectionBuilder
+
+  /**
+   * Create flow that when materialized creates a single connection to the HTTP server.
+   *
+   * The responses are not guaranteed to arrive in the same order as the requests go out
+   * so therefore requests needs to have a [[akka.http.scaladsl.model.http2.RequestResponseAssociation]]
+   * which Akka HTTP will carry over to the corresponding response for a request.
+   */
+  def unorderedFlow(): Flow[HttpRequest, HttpResponse, CompletionStage[OutgoingConnection]]
```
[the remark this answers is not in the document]
ignasi35: After reading more on the PR, I don't think we should make it so explicit that the `Flow` may be unordered. It's exposing too much about the nature of the underlying connection. If the user wants to connect via HTTP/2, they should know the `Flow` will be unordered. If the user wants HTTP/1.x (let's assume we support pieplining), then they'll know the `Flow` is ordered.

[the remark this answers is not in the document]
ignasi35: If we want to keep the `unordered` bit in the name of the method, then I'd only add that method on `Htt2p.scala` but not on `Http` (which would require specializations on `OutgoingConnectionBuilder`).

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
```diff
@@ -386,6 +386,31 @@ class HttpExt private[http] (private val config: Config)(implicit val system: Ex
...
+   *
+   * @return A builder to configure more specific setup for the connection and then build a `Flow[Request, Response, Future[OutgoingConnection]]`.
+   */
+  /*
+   * FIXME: I think we'd want this in the HTTP extension for easy access but it cannot refer to classes in the http2-support module
+   * FIXME: for having the same entry point for HTTP/2 and HTTP/1.1 there is a mismatch with port always reasonably defaulting to 443
+   *        for HTTP/2 but not really for HTTP1/1
+   * FIXME also for same API and automatic protocol negotiation, how do we deal with explicitly saying TLS/No TLS for HTTP/1 but defaulting to TLS for HTTP/2?
+   */
+  def connectionTo(host: String): OutgoingConnectionBuilder =
```
ignasi35, line 407: These `FIXME`'s and the actual location of `connectionTo` (`Http` instead of `Http2`) confused me a bit regarding how this will look in the end.

ignasi35: I was thinking a user would use `val flow: Flow[] = Http2().someMethods()...` to connect to a server using HTTP/2 (over TLS or WithPriorKnowledge).

Alternatively, they would use `val flow: Flow[] = Http().someMethods()...` to connect to a server using HTTP/1.x (over TLS or cleartext).

Using `Http2()` or `Http()` would be enough for the user to be aware that the obtained `Flow` is ordered or not.

ignasi35 (commented):
Left a handful of comments but they are quite connected.

akka-http-core/src/main/scala/akka/http/javadsl/Http.scala
```diff
@@ -390,6 +390,21 @@ class Http(system: ExtendedActorSystem) extends akka.actor.Extension {
...
+   *
+   * The responses are not guaranteed to arrive in the same order as the requests go out (In the case of a HTTP/2 server)
+   * so therefore requests needs to have a [[akka.http.scaladsl.model.http2.RequestResponseAssociation]]
+   * which Akka HTTP will carry over to the corresponding response for a request.
+   *
+   *
+   * @return A builder to configure more specific setup for the connection and then build a `Flow[Request, Response, Future[OutgoingConnection]]`.
+   */
+  def connectionTo(host: String): OutgoingConnectionBuilder =
+    delegate.connectionTo(host).toJava
```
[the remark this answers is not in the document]
johanandren: I don't think that it is a good idea with a factory method returning an invalid builder that you cannot build (or that would throw if you did build immediately), so host plus implicit port (from implicit knowledge that it is HTTP/1 or HTTP/2) is minimal IMO.

akka-http-core/src/main/scala/akka/http/javadsl/OutgoingConnectionBuilder.scala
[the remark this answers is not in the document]
jrudolph: Http2 is not public API. So, we need to figure out something for Http, which this whole effort is about :)

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
jrudolph: See the comment above regarding private Http2 vs. public Http.

jrudolph:
How about this naming:

```scala
// main usage
def http(): Flow[...]
def https(): Flow[...]
def http2(): Flow[...]
def http2WithPriorKnowledge(): Flow[...]

// customization

// only applicable for https() and http2(), overrides defaultHttpsContext
def withCustomHttpsConnectionContext(httpsConnectionContext: HttpsConnectionContext): OutgoingConnectionBuilder

// or `toCustomPort`? overrides the default ports based on TLS selection
def toPort(port: Int): OutgoingConnectionBuilder

// optional overloads that would allow choosing protocol, TLS, and context at the same time
// violates complete separation of concerns but might be useful for internal usage
// Maybe should fail if withCustomHttpsConnectionContext was used before?
def generic(useHttp2: Boolean, connectionContext: ConnectionContext): Flow[...]
```

johanandren:
Sounds good enough to me, I'll update.

akka-http-core/src/main/scala/akka/http/javadsl/model/RequestResponseAssociation.scala
```diff
@@ -0,0 +1,26 @@
...
+
+import akka.annotation.ApiMayChange
+
+import scala.concurrent.Promise
+
+/**
+ * A marker trait for attribute values that should be (automatically) carried over from request to response.
+ */
+@ApiMayChange
+trait RequestResponseAssociation extends akka.http.scaladsl.model.RequestResponseAssociation
```
jrudolph, line 17: Usually we have the dependency hierarchy the other way round. Would that work as well?

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
```diff
@@ -386,12 +387,25 @@ class HttpExt private[http] (private val config: Config)(implicit val system: Ex
   private[http] val poolMaster: PoolMaster = PoolMaster()
   private[this] val systemMaterializer = SystemMaterializer(system).materializer
 
+  /**
+   * Creates a builder which will create a single connection to a host every time the built flow is materialized. There
+   * is no pooling and you are yourself responsible for lifecycle management of the connection. For a more convenient
+   * Request level API see [[singleRequest()]]
+   *
+   * The builder defaults to build a connection using HTTP/1.1 over a plaintext connection to port 80.
```
jrudolph, line 395: This doesn't seem applicable any more.

akka-http-core/src/main/scala/akka/http/scaladsl/model/RequestResponseAssociation.scala
```diff
@@ -16,4 +17,14 @@ trait RequestResponseAssociation
  * A simple value holder class implementing RequestResponseAssociation.
  */
 @ApiMayChange
-final case class SimpleRequestResponseAttribute[T](value: T) extends RequestResponseAssociation
+case class SimpleRequestResponseAttribute[T](value: T) extends RequestResponseAssociation
```
jrudolph, line 20: Was that `final` removed intentionally?

docs/src/main/paradox/client-side/connection-level.md
```diff
@@ -41,6 +42,25 @@ eventually be slowed down in sending requests.
...
 
+## Request-response ordering for HTTP/2
+
+For HTTP/2 connections the responses are not guaranteed to arrive in the same order that the requests were emitted to
+the server, for example a request with a quickly available response may outrun a previous request that the server is 
+slower to respond to. For HTTP/2 it is therefore often important to have a way to correlate the response with what request
+it was made for. This can be achieved through a @apidoc[RequestResponseAssociation] set on the request, Akka HTTP will pass
+such association objects on to the response.
+
+In this sample the built in @apidoc[ResponseFuture] `RequestResponseAssociation`  is used to return
```
jrudolph, line 53: :+1:

docs/src/test/java/docs/http/javadsl/Http2ClientApp.java
```diff
@@ -0,0 +1,104 @@
+/*
```
jrudolph, line 1: Thanks for adding that example!

akka-http-core/src/main/scala/akka/http/scaladsl/model/RequestResponseAssociation.scala
```diff
@@ -16,4 +17,14 @@ trait RequestResponseAssociation
...
+
+/**
+ * An association for completing a future when the response arrives.
+ */
+final class ResponseFuture(val promise: Promise[HttpResponse]) extends RequestResponseAssociation
```
jrudolph, line 25: Maybe `ResponsePromise` or `ResponseCallback`, instead?

akka-http-core/src/main/scala/akka/http/scaladsl/OutgoingConnectionBuilder.scala
```diff
@@ -0,0 +1,88 @@
...
+import akka.http.scaladsl.model.HttpResponse
+import akka.http.scaladsl.settings.ClientConnectionSettings
+import akka.stream.scaladsl.Flow
+import akka.http.javadsl.{ OutgoingConnectionBuilder => JOutgoingConnectionBuilder }
+
+import scala.concurrent.Future
+
+/**
+ * Builder for setting up a flow that will create one single connection per materialization to the specified host.
+ * When customization is done, the flow is created using [[#connectionFlow()]].
```
jrudolph, line 21: Not true anymore.

akka-http-core/src/main/scala/akka/http/javadsl/OutgoingConnectionBuilder.scala
```diff
@@ -0,0 +1,80 @@
...
+import akka.event.LoggingAdapter
+import akka.http.javadsl.model.HttpRequest
+import akka.http.javadsl.model.HttpResponse
+import akka.http.javadsl.settings.ClientConnectionSettings
+import akka.stream.javadsl
+import akka.stream.javadsl.Flow
+
+/**
+ * Builder for setting up a flow that will create one single connection per materialization to the specified host.
+ * When customization is done, the flow is created using [[#connectionFlow()]].
```
jrudolph, line 19: Not true any more.

akka-http-core/src/main/scala/akka/http/javadsl/OutgoingConnectionBuilder.scala
```diff
@@ -0,0 +1,75 @@
...
+ *
+ * Not for user extension
+ */
+@DoNotInherit
+trait OutgoingConnectionBuilder {
+  /**
+   * Change which host flows built with this builder connects to
+   */
+  def toHost(host: String): OutgoingConnectionBuilder
```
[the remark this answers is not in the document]
jrudolph: I think reusing is not the main reason but more a possibility to override the host after the fact if you want to abstract setting some features of the builder in user-code which would otherwise require to do the host thing first.

But in general, I agree, and it would also be possible to remove this one completely.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/OutgoingConnectionBuilderImpl.scala
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    override def https(): Flow[HttpRequest, HttpResponse, Future[OutgoingConnection]] = {
+      // http/1.1 tls
+      Http(system).outgoingConnectionHttps(host, port.getOrElse(443), connectionContext.getOrElse(Http(system).defaultClientHttpsContext), None, clientConnectionSettings, log)
+    }
+
+    override def http2(): Flow[HttpRequest, HttpResponse, Future[OutgoingConnection]] = {
+      // http/2 tls
+      val port = this.port.getOrElse(443)
```
jrudolph, line 85: All of this should stay / be implemented in `Http2`.

jrudolph (approved):
Looking good with only a few cosmetic things remaining.

akka-http-core/src/main/scala/akka/http/javadsl/model/RequestResponseAssociation.scala
johanandren: Sure, I think I just misremembered and didn't double check. I'll fix!

akka-http-core/src/main/scala/akka/http/scaladsl/model/RequestResponseAssociation.scala
johanandren: Hmm, maybe not

akka-http-core/src/main/scala/akka/http/scaladsl/model/RequestResponseAssociation.scala
johanandren: Heh, I agree, I had it as `ResponsePromise` but couldn't get the `@apidoc` paradox tag to work with two different names. The best reason for naming things. Perhaps I'll just skip those links.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/OutgoingConnectionBuilderImpl.scala
johanandren: Alright, so both http2 variations go in there as internal api and just delegate from here?

jrudolph: Yep.

jrudolph (approved):
LGTM

raboof (approved):
LGTM!

jrudolph:
:partying_face:  Great stuff, thanks @johanandren.

Issue #3514, opened 2020-10-12
Discuss: is splitting out the akka-http2-support module still worth it


jrudolph:
When working on akka-http2-support we opted for putting HTTP/2 support into its own module, so it could be optional for users.

akka-http2-support and akka-http-core have a cyclic dependency:

 * akka-http2-support depends on akka-http-core for utils, models, HTTP/1 implementation internals, etc.
 * akka-http-core depends on akka-http2-support to be able to provide transparent HTTP/2 support under the existing APIs

The cyclic dependency is resolved by using SPI-like reflection layer to find out if akka-http2-support is included and access the implementation.

If we had the freedom to rearrange our modules, we could split out utils and models and similar stuff to the bottom of our module hierarchy, then put implementations on top of this and then create a new public akka-http-core which only contains the public APIs on top of that (or move those APIs to akka-http).

It's somewhat unlikely that we can do the perfect refactoring any time soon (or in the 10.x release line).

At this point we can make this decision:

 * keep the current structure and keep HTTP/2 support optional (maybe improve `Http2Shadow` to use a trait instead of a structural type)
 * move HTTP/2 into akka-http-core and provide HTTP/2 support out of the box (but optionally still behind the flag)

Benefits of the current structure:

 * HTTP/2 support is optional for users (smaller code footprint if you don't need HTTP/2 support)
 * HTTP/2 implementation somewhat well-hidden behind the reflective `Http2Shadow` interface (it's made harder to refer to internal HTTP/2 bits)

Disadvantages of the current structure:

 * Users need to include akka-http2-support manually if they need HTTP/2 support
 * Adding references to new entrypoints in HTTP/2 is made harder
 * The dependency structure is somewhat confusing

Thanks to @johanandren for raising this issue.

jrudolph:
@raboof WDYT?

raboof:
Good question. My current feeling is, that while the disadvantages of the current setup aren't that bad, the advantages aren't so convincing (anymore) either.

Looking ahead to 11.0.0, splitting up across models/http1/http2/api's sounds feasible, but whether or not we keep akka-http2-support separate doesn't seem to make that any harder or easier.

So I don't see a decisive argument yet - I'm good with the status quo, but wouldn't be opposed to rolling akka-http2-support into akka-http-core either.

jrudolph:
If we want to include it in akka-http-core (and it seems to become the consensus) we should probably do it soon to avoid conflicts and extra work like with #3511.

[...]

PR #3525, opened 2020-10-14
http2: move internal classes to impl.engine.http2 package

jrudolph:
Most importantly, move `Http2` from `scaladsl` to `impl.engine.http2` and mark it explicitly as internal.

This is a general cleanup and preparation move for the potential move into akka-http-core (#3514).


jrudolph:
Now part of #3526.

PR #3526, opened 2020-10-14
Move http2 implementation into core

jrudolph:
Includes former #3525.

Only the implementation has been moved into `akka.http.impl.engine.http2` and the `Http2Shadow` has been removed.

Tests still live in the now otherwise empty `akka-http2-support` module to avoid messing up akka-http-core or akka-http-tests module with purely http2-specific setup bits.

Don't squash when merging to help git understanding when files have been moved.


ennru (approved):
I suggest we keep the jar around (even if it becomes empty) so people don't run into strange dependency problems when upgrading.
As we keep the project for the tests it isn't too strange, right? 
LGTM.

jrudolph:
> I suggest we keep the jar around (even if it becomes empty) so people don't run into strange dependency problems when upgrading.
> As we keep the project for the tests it isn't too strange, right?

Yes, at least until 10.3.0.

raboof (approved):

jrudolph:
Rebased before upcoming merge.

Issue #3522, opened 2020-10-14
HTTP/2: SETTINGs send and enforce SETTINGS_MAX_CONCURRENT_STREAMS


ignasi35:
This is a specific task off of #3185 (see https://github.com/akka/akka-http/issues/3185#issuecomment-708287733):

> there's already a max-concurrent-streams in the config to control the concurrency

> I think these should be connected. Basically, we should announce the setting value to the peer at the start of the connection so the peer has to backpressure new substreams instead of having us pile up incoming substreams in front of the user-handler.

jrudolph:
#774 is also about this setting but maybe for the other side.

ignasi35:
> #774 is also about this setting but maybe for the other direction (receiving this setting and respecting it).

Yeha, #774 is about controlling the size of [outstreams](https://github.com/akka/akka-http/blob/1b8fd7fa1d/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Multiplexer.scala#L180)

Issue #774, opened 2017-01-16
HTTP2: Respect SETTINGS_MAX_CONCURRENT_STREAMS


ktoso:
SETTINGS_MAX_CONCURRENT_STREAMS (0x3):
Indicates the maximum number of concurrent streams that the sender will allow. This limit is directional: it applies to the number of streams that the sender permits the receiver to create. Initially, there is no limit to this value. It is recommended that this value be no smaller than 100, so as to not unnecessarily limit parallelism.

A value of 0 for SETTINGS_MAX_CONCURRENT_STREAMS SHOULD NOT be treated as special by endpoints. A zero value does prevent the creation of new streams; however, this can also happen for any limit that is exhausted with active streams. Servers SHOULD only set a zero value for short durations; if a server does not wish to accept requests, closing the connection is more appropriate.

johanandren:
Setting added but not enforced so we can keep this ticket open

jrudolph:
This has been better specified in #3228. Closing this as a duplicate.

PR #3529, opened 2020-10-14
http2: send and enforce max concurrent streams setting

ignasi35:
References #3522 (related to #774 but doesn't fix it)

### DIFF
--- a/akka-http-core/src/main/resources/reference.conf
+++ b/akka-http-core/src/main/resources/reference.conf
@@ -228,6 +228,7 @@ akka.http {
       # This limit is enforced as soon as the connection between the peers is established. Enforcing
       # the limit even before the SETTINGS/SETTINGS_ACK echange happens between the peers is more
       # strict than required in the HTTP/2 Spec.
+      # Set to 'infinite' to disable this setting.
       max-concurrent-streams = 256
 
       # The maximum number of bytes to receive from a request entity in a single chunk.
--- a/akka-http-core/src/main/resources/reference.conf
+++ b/akka-http-core/src/main/resources/reference.conf
@@ -226,8 +226,9 @@ akka.http {
     http2 {
       # The maximum number of request per connection concurrently dispatched to the request handler.
       # This limit is enforced as soon as the connection between the peers is established. Enforcing
-      # the limit even before the SETTINGS/SETTINGS_ACK echange happens between the peers is more
-      # strict than required in the HTTP/2 Spec.
+      # the limit even before the SETTINGS/SETTINGS_ACK exchange has completed, means
+      # that we will refuse extra streams/requests that were sent by the client right after 
+      # the connection was established but before it received our SETTINGS.
       # Set to 'infinite' to disable this setting.
       max-concurrent-streams = 256
 
--- a/akka-http-core/src/main/resources/reference.conf
+++ b/akka-http-core/src/main/resources/reference.conf
@@ -229,7 +229,6 @@ akka.http {
       # the limit even before the SETTINGS/SETTINGS_ACK exchange has completed, means
       # that we will refuse extra streams/requests that were sent by the client right after 
       # the connection was established but before it received our SETTINGS.
-      # Set to 'infinite' to disable this setting.
       max-concurrent-streams = 256
 
       # The maximum number of bytes to receive from a request entity in a single chunk.
--- a/akka-http-core/src/main/resources/reference.conf
+++ b/akka-http-core/src/main/resources/reference.conf
@@ -380,8 +380,9 @@ akka.http {
     http2 {
       # The maximum number of request per connection concurrently dispatched to the request handler.
       # This limit is enforced as soon as the connection between the peers is established. Enforcing
-      # the limit even before the SETTINGS/SETTINGS_ACK echange happens between the peers is more
-      # strict than required in the HTTP/2 Spec.
+      # the limit even before the SETTINGS/SETTINGS_ACK exchange has completed, means
+      # that we will refuse extra streams/push promises that were sent by the client right after 
+      # the connection was established but before it received our SETTINGS.
       max-concurrent-streams = 256
 
       # The maximum number of bytes to receive from a request entity in a single chunk.
--- a/akka-http-core/src/main/resources/reference.conf
+++ b/akka-http-core/src/main/resources/reference.conf
@@ -225,6 +225,9 @@ akka.http {
 
     http2 {
       # The maximum number of request per connection concurrently dispatched to the request handler.
+      # This limit is enforced as soon as the connection between the peers is established. Enforcing
+      # the limit even before the SETTINGS/SETTINGS_ACK echange happens between the peers is more
+      # strict than required in the HTTP/2 Spec.
       max-concurrent-streams = 256
 
       # The maximum number of bytes to receive from a request entity in a single chunk.
@@ -374,6 +377,9 @@ akka.http {
     // FIXME: unify with server-side part (by importing or similar to parsing)
     http2 {
       # The maximum number of request per connection concurrently dispatched to the request handler.
+      # This limit is enforced as soon as the connection between the peers is established. Enforcing
+      # the limit even before the SETTINGS/SETTINGS_ACK echange happens between the peers is more
+      # strict than required in the HTTP/2 Spec.
       max-concurrent-streams = 256
 
       # The maximum number of bytes to receive from a request entity in a single chunk.
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/FrameEvent.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/FrameEvent.scala
@@ -44,9 +44,9 @@ private[http] object FrameEvent {
   }
 
   /**
-    * Representation of the wire-level HEADERS frame. Together with ContinuationFrame's and instance of
-    * HeadersFrame will be parsed into a logical ParsedHeadersFrame.
-    */
+   * Representation of the wire-level HEADERS frame. Together with ContinuationFrame's and instance of
+   * HeadersFrame will be parsed into a logical ParsedHeadersFrame.
+   */
   final case class HeadersFrame(
     streamId:            Int,
     endStream:           Boolean,
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/FrameEvent.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/FrameEvent.scala
@@ -43,6 +43,10 @@ private[http] object FrameEvent {
     def sizeInWindow: Int = payload.size // FIXME: take padding size into account, #1313
   }
 
+  /**
+    * Representation of the wire-level HEADERS frame. Together with ContinuationFrame's and instance of
+    * HeadersFrame will be parsed into a logical ParsedHeadersFrame.
+    */
   final case class HeadersFrame(
     streamId:            Int,
     endStream:           Boolean,
@@ -92,6 +96,10 @@ private[http] object FrameEvent {
     streamId: Int,
     payload:  ByteString) extends StreamFrameEvent
 
+  /**
+   * Convenience (logical) representation of a parsed HEADERS frame with zero, one or
+   * many CONTINUATIONS Frames into a single, decompressed object.
+   */
   final case class ParsedHeadersFrame(
     streamId:      Int,
     endStream:     Boolean,
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Compliance.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Compliance.scala
@@ -28,13 +28,6 @@ private[http2] object Http2Compliance {
   private[akka] sealed class IllegalHttp2FrameSize(size: Int, expected: String)
     extends Http2ProtocolException(ErrorCode.FRAME_SIZE_ERROR, s"Illegal HTTP/2 frame size: [$size]. $expected!")
 
-  // If the current count exceeds the maximum we let them complete. So the verification is
-  // only meant to be used when a new stream is created (but not when the setting is updated.
-  // 5.1.2 "An endpoint that wishes to reduce the value of SETTINGS_MAX_CONCURRENT_STREAMS to a
-  // value that is below the current number of open streams can either close streams that
-  // exceed the new value or allow streams to complete."
-  final def compliesWithMaxConcurrentStreams(currentConcurrentStreams: Int, maxConcurrentStreams: Int): Boolean = currentConcurrentStreams < maxConcurrentStreams
-
   // require methods use `if` because `require` allocates
 
   /** Validate value of MAX_FRAME_SIZE setting. */
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Compliance.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Compliance.scala
@@ -28,6 +28,13 @@ private[http2] object Http2Compliance {
   private[akka] sealed class IllegalHttp2FrameSize(size: Int, expected: String)
     extends Http2ProtocolException(ErrorCode.FRAME_SIZE_ERROR, s"Illegal HTTP/2 frame size: [$size]. $expected!")
 
+  // If the current count exceeds the maximum we let them complete. So the verification is
+  // only meant to be used when a new stream is created (but not when the setting is updated.
+  // 5.1.2 "An endpoint that wishes to reduce the value of SETTINGS_MAX_CONCURRENT_STREAMS to a
+  // value that is below the current number of open streams can either close streams that
+  // exceed the new value or allow streams to complete."
+  final def compliesWithMaxConcurrentStreams(currentConcurrentStreams: Int, maxConcurrentStreams: Int): Boolean = currentConcurrentStreams < maxConcurrentStreams
+
   // require methods use `if` because `require` allocates
 
   /** Validate value of MAX_FRAME_SIZE setting. */
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
@@ -162,8 +162,6 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
               // above in preStart. Since, only some settings are supported, and those
               // settings are non-modifiable and known at construction time, these settings
               // are enforced from the start of the connection.
-              // Also, since we only expect an ack for the initial settings frame there's
-              // no need to correlate the ACK to a particular SETTINGS frame.
               // Related: https://github.com/akka/akka-http/issues/3185
               enforceSettings(initialLocalSettings)
 
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
@@ -104,7 +104,7 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
       // Send settings initially based on our configuration. For simplicity, these settings are
       // enforced immediately even before the acknowledgement is received.
       // Reminder: the receiver of a SETTINGS frame must process them in the order they are received.
-      private def initialLocalSettings = immutable.Seq(Setting(SettingIdentifier.SETTINGS_MAX_CONCURRENT_STREAMS, value))
+      private def initialLocalSettings = immutable.Seq(Setting(SettingIdentifier.SETTINGS_MAX_CONCURRENT_STREAMS, maxConcurrentStreams))
 
       override def preStart(): Unit = {
         if (initialDemuxerSettings.nonEmpty) {
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
@@ -99,8 +99,6 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
       override def settings: Http2CommonSettings = http2Settings
       override def isUpgraded: Boolean = upgraded
 
-      def maxConcurrentStreams: Int = http2Settings.maxConcurrentStreams
-
       override protected def logSource: Class[_] = if (isServer) classOf[Http2ServerDemux] else classOf[Http2ClientDemux]
 
       val multiplexer = createMultiplexer(frameOut, StreamPrioritizer.first())
@@ -108,7 +106,7 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
       // Send settings initially based on our configuration. For simplicity, these settings are
       // enforced immediately even before the acknowledgement is received.
       // Reminder: the receiver of a SETTINGS frame must process them in the order they are received.
-      private def initialLocalSettings = immutable.Seq(Setting(SettingIdentifier.SETTINGS_MAX_CONCURRENT_STREAMS, maxConcurrentStreams))
+      private def initialLocalSettings = immutable.Seq(Setting(SettingIdentifier.SETTINGS_MAX_CONCURRENT_STREAMS, http2Settings.maxConcurrentStreams))
 
       override def preStart(): Unit = {
         if (initialRemoteSettings.nonEmpty) {
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
@@ -5,22 +5,25 @@
 package akka.http.impl.engine.http2
 
 import akka.annotation.InternalApi
+import akka.http.impl.engine.http2.FrameEvent._
 import akka.http.impl.engine.http2.Http2Protocol.ErrorCode
 import akka.http.impl.engine.http2.Http2Protocol.ErrorCode.FLOW_CONTROL_ERROR
+import akka.http.impl.engine.http2.Http2Protocol.SettingIdentifier
 import akka.http.scaladsl.settings.Http2CommonSettings
+import akka.macros.LogHelper
 import akka.stream.Attributes
 import akka.stream.BidiShape
 import akka.stream.Inlet
 import akka.stream.Outlet
 import akka.stream.impl.io.ByteStringParser.ParsingException
-import akka.stream.stage.{ GraphStage, GraphStageLogic, InHandler, StageLogging }
+import akka.stream.stage.GraphStage
+import akka.stream.stage.GraphStageLogic
+import akka.stream.stage.InHandler
+import akka.stream.stage.StageLogging
 import akka.util.ByteString
 
 import scala.collection.immutable
 import scala.util.control.NonFatal
-import FrameEvent._
-import akka.http.impl.engine.http2.Http2Protocol.SettingIdentifier
-import akka.macros.LogHelper
 
 /** Currently only used as log source */
 @InternalApi
@@ -92,9 +95,20 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
       override def settings: Http2CommonSettings = http2Settings
       override def isUpgraded: Boolean = upgraded
 
+      override def maxConcurrentStreams: Int = http2Settings.maxConcurrentStreams
+
       override protected def logSource: Class[_] = if (isServer) classOf[Http2ServerDemux] else classOf[Http2ClientDemux]
 
       val multiplexer = createMultiplexer(frameOut, StreamPrioritizer.first())
+
+      // TODO: we only support the initial SETTINGS message (the only one that's compulsory)
+      //  so we're keeping the settings as a field instead of implementing SETTINGS-SETTINGS_ACK
+      //  correlation logic.
+      // TODO: the receiver of a SETTINGS frame must apply them in the order they
+      //  are received. Review order of the settings in the sequence since it's relevant.
+      val localSettings = immutable.Seq(
+        Setting(SettingIdentifier.SETTINGS_MAX_CONCURRENT_STREAMS, maxConcurrentStreams)
+      )
 
       override def preStart(): Unit = {
         if (initialDemuxerSettings.nonEmpty) {
@@ -105,13 +119,8 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
         pullFrameIn()
         pull(substreamIn)
 
-        // TODO: review order of the settings in the sequence
-        val settings = Seq(
-          Setting(SettingIdentifier.SETTINGS_MAX_CONCURRENT_STREAMS, http2Settings.maxConcurrentStreams)
-        )
         // both client and server must send a settings frame as first frame
-        multiplexer.pushControlFrame(SettingsFrame(settings))
-        //                multiplexer.pushControlFrame(SettingsFrame(Nil))
+        multiplexer.pushControlFrame(SettingsFrame(localSettings))
       }
 
       override def pushGOAWAY(errorCode: ErrorCode, debug: String): Unit = {
@@ -167,13 +176,15 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
               }
 
             case SettingsAckFrame(Nil) =>
-            // Currently, we only expect an ack for the initial (currently empty) settings frame, sent
-            // above in preStart. Since, it was empty, there's nothing to do here.
-            // If we want to support setting and enforcing settings, we'll need to act here to commit
-            // to the settings we sent out before.
-            // https://github.com/akka/akka-http/issues/3185
+              // Currently, we only expect an ack for the initial settings frame, sent
+              // above in preStart. Since, only some settings are supported, and those
+              // settings are non-modifiable and known at construction time, there's not much to do here.
+              // Also, since we only expect an ack for the initial settings frame there's
+              // no need to correlate the ACK to a particular SETTINGS frame.
+              // Related: https://github.com/akka/akka-http/issues/3185
+              enforceSettings(localSettings)
 
-            case PingFrame(true, _)    =>
+            case PingFrame(true, _) =>
             // ignore for now (we don't send any pings)
             case PingFrame(false, data) =>
               multiplexer.pushControlFrame(PingFrame(ack = true, data))
@@ -226,6 +237,29 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
         }
       })
 
+      /**
+       * Tune this peer to enforce the settings configured from this peer.
+       * @return TODO: it's a Boolean but I don't think it has to be.
+       */
+      private def enforceSettings(settings: immutable.Seq[Setting]): Boolean = {
+        var settingsAppliedOk = true
+
+        settings.foreach {
+          case Setting(Http2Protocol.SettingIdentifier.SETTINGS_MAX_CONCURRENT_STREAMS, value) =>
+          // TODO: Http2Compliance.compliesWithMaxConcurrentStreams should only be enabled once
+          //  we get this SETTINGS_ACK.
+        }
+
+        settingsAppliedOk
+      }
+
+      /**
+       * Tune this peer to honour the settings sent from the other peer.
+       * @param settings settings sent from the other peer (or injected via the
+       *                 "HTTP2-Settings" in "h2c").
+       * @return true if settings were applied successfully, false if some ERROR
+       *         was raised (and pushed back to the peer)
+       */
       private def applySettings(settings: immutable.Seq[Setting]): Boolean = {
         var settingsAppliedOk = true
 
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
@@ -101,12 +101,9 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
 
       val multiplexer = createMultiplexer(frameOut, StreamPrioritizer.first())
 
-      // TODO: we only support the initial SETTINGS message (the only one that's compulsory)
-      //  so we're keeping the settings as a field instead of implementing SETTINGS-SETTINGS_ACK
-      //  correlation logic.
-      // TODO: the receiver of a SETTINGS frame must apply them in the order they
-      //  are received. Review order of the settings in the sequence since it's relevant.
-      val localSettings = immutable.Seq(
+      // Reminder: the receiver of a SETTINGS frame must apply them in the order they
+      //  are received. Place first the settings you want processed early.
+      private val initialLocalSettings = immutable.Seq(
         Setting(SettingIdentifier.SETTINGS_MAX_CONCURRENT_STREAMS, maxConcurrentStreams)
       )
 
@@ -120,7 +117,7 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
         pull(substreamIn)
 
         // both client and server must send a settings frame as first frame
-        multiplexer.pushControlFrame(SettingsFrame(localSettings))
+        multiplexer.pushControlFrame(SettingsFrame(initialLocalSettings))
       }
 
       override def pushGOAWAY(errorCode: ErrorCode, debug: String): Unit = {
@@ -182,7 +179,7 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
               // Also, since we only expect an ack for the initial settings frame there's
               // no need to correlate the ACK to a particular SETTINGS frame.
               // Related: https://github.com/akka/akka-http/issues/3185
-              enforceSettings(localSettings)
+              enforceSettings(initialLocalSettings)
 
             case PingFrame(true, _) =>
             // ignore for now (we don't send any pings)
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
@@ -95,7 +95,7 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
       override def settings: Http2CommonSettings = http2Settings
       override def isUpgraded: Boolean = upgraded
 
-      def maxConcurrentStreams: Int = http2Settings.maxConcurrentStreams
+      def maxConcurrentStreams: Option[Int] = http2Settings.maxConcurrentStreams
 
       override protected def logSource: Class[_] = if (isServer) classOf[Http2ServerDemux] else classOf[Http2ClientDemux]
 
@@ -103,9 +103,8 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
 
       // Reminder: the receiver of a SETTINGS frame must apply them in the order they
       //  are received. Place first the settings you want processed early.
-      private val initialLocalSettings = immutable.Seq(
-        Setting(SettingIdentifier.SETTINGS_MAX_CONCURRENT_STREAMS, maxConcurrentStreams)
-      )
+      private val initialLocalSettings = immutable.Seq.empty ++
+        maxConcurrentStreams.toSeq.map(value => Setting(SettingIdentifier.SETTINGS_MAX_CONCURRENT_STREAMS, value))
 
       override def preStart(): Unit = {
         if (initialDemuxerSettings.nonEmpty) {
@@ -251,15 +250,14 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
       }
 
       protected override def checkMaxConcurrentStreamsCompliance(currentConcurrentStreams: => Int): Unit = {
-        if (currentConcurrentStreams >= maxConcurrentStreams) {
-          // 5.1.2 An endpoint that receives a HEADERS frame that causes its advertised
-          // concurrent stream limit to be exceeded MUST treat this as a stream error
-          // (Section 5.4.2) of type PROTOCOL_ERROR or REFUSED_STREAM.
-          pushGOAWAY(
-            ErrorCode.PROTOCOL_ERROR,
-            s"Peer tried to open too many streams. Number of open streams=[$currentConcurrentStreams], max concurrent streams=[$maxConcurrentStreams]."
-          )
-        }
+        maxConcurrentStreams.foreach(limit =>
+          if (currentConcurrentStreams >= limit) {
+            pushGOAWAY(
+              ErrorCode.PROTOCOL_ERROR,
+              s"Peer tried to open too many streams. Number of open streams=[$currentConcurrentStreams], max concurrent streams=[$maxConcurrentStreams]."
+            )
+          }
+        )
       }
 
       /**
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
@@ -236,12 +236,6 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
         settingsAppliedOk
       }
 
-      protected override def checkMaxConcurrentStreamsCompliance(currentConcurrentStreams: => Int): Unit = {
-        if (currentConcurrentStreams >= maxConcurrentStreams) {
-          resetStream(streamId, ErrorCode.REFUSED_STREAM)
-        }
-      }
-
       /**
        * Tune this peer to the remote Settings.
        * @param settings settings sent from the other peer (or injected via the
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
@@ -243,8 +243,8 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
 
         settings.foreach {
           case Setting(Http2Protocol.SettingIdentifier.SETTINGS_MAX_CONCURRENT_STREAMS, value) =>
-          // TODO: Http2Compliance.compliesWithMaxConcurrentStreams should only be enabled once
-          //  we get this SETTINGS_ACK.
+            // Enforcing of SETTINGS_MAX_CONCURRENT_STREAMS is enabled even before getting the SETTINGS_ACK
+            // so there's nothing to do here. See https://github.com/akka/akka-http/issues/3551
         }
 
         settingsAppliedOk
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
@@ -103,7 +103,7 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
 
       // Reminder: the receiver of a SETTINGS frame must apply them in the order they
       //  are received. Place first the settings you want processed early.
-      private val initialLocalSettings = immutable.Seq.empty ++
+      private def initialLocalSettings = immutable.Seq.empty ++
         maxConcurrentStreams.toSeq.map(value => Setting(SettingIdentifier.SETTINGS_MAX_CONCURRENT_STREAMS, value))
 
       override def preStart(): Unit = {
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
@@ -171,7 +171,7 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
                 multiplexer.pushControlFrame(SettingsAckFrame(settings))
               }
 
-            case SettingsAckFrame(Nil) =>
+            case SettingsAckFrame(_) =>
               // Currently, we only expect an ack for the initial settings frame, sent
               // above in preStart. Since, only some settings are supported, and those
               // settings are non-modifiable and known at construction time, these settings
@@ -252,10 +252,7 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
 
       protected override def checkMaxConcurrentStreamsCompliance(currentConcurrentStreams: => Int): Unit = {
         if (currentConcurrentStreams >= maxConcurrentStreams) {
-          pushGOAWAY(
-            ErrorCode.PROTOCOL_ERROR,
-            s"Peer tried to open too many streams. Number of open streams=[$currentConcurrentStreams], max concurrent streams=[$maxConcurrentStreams]."
-          )
+          resetStream(streamId, ErrorCode.REFUSED_STREAM)
         }
       }
 
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
@@ -95,7 +95,7 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
       override def settings: Http2CommonSettings = http2Settings
       override def isUpgraded: Boolean = upgraded
 
-      def maxConcurrentStreams: Option[Int] = if (http2Settings.maxConcurrentStreams == Int.MaxValue) None else Some(http2Settings.maxConcurrentStreams)
+      def maxConcurrentStreams: Int = http2Settings.maxConcurrentStreams
 
       override protected def logSource: Class[_] = if (isServer) classOf[Http2ServerDemux] else classOf[Http2ClientDemux]
 
@@ -104,8 +104,7 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
       // Send settings initially based on our configuration. For simplicity, these settings are
       // enforced immediately even before the acknowledgement is received.
       // Reminder: the receiver of a SETTINGS frame must process them in the order they are received.
-      private def initialLocalSettings = immutable.Seq.empty ++
-        maxConcurrentStreams.toSeq.map(value => Setting(SettingIdentifier.SETTINGS_MAX_CONCURRENT_STREAMS, value))
+      private def initialLocalSettings = immutable.Seq(Setting(SettingIdentifier.SETTINGS_MAX_CONCURRENT_STREAMS, value))
 
       override def preStart(): Unit = {
         if (initialDemuxerSettings.nonEmpty) {
@@ -252,14 +251,12 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
       }
 
       protected override def checkMaxConcurrentStreamsCompliance(currentConcurrentStreams: => Int): Unit = {
-        maxConcurrentStreams.foreach(limit =>
-          if (currentConcurrentStreams >= limit) {
-            pushGOAWAY(
-              ErrorCode.PROTOCOL_ERROR,
-              s"Peer tried to open too many streams. Number of open streams=[$currentConcurrentStreams], max concurrent streams=[$maxConcurrentStreams]."
-            )
-          }
-        )
+        if (currentConcurrentStreams >= maxConcurrentStreams) {
+          pushGOAWAY(
+            ErrorCode.PROTOCOL_ERROR,
+            s"Peer tried to open too many streams. Number of open streams=[$currentConcurrentStreams], max concurrent streams=[$maxConcurrentStreams]."
+          )
+        }
       }
 
       /**
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
@@ -101,8 +101,9 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
 
       val multiplexer = createMultiplexer(frameOut, StreamPrioritizer.first())
 
-      // Reminder: the receiver of a SETTINGS frame must apply them in the order they
-      //  are received. Place first the settings you want processed early.
+      // Send settings initially based on our configuration. For simplicity, these settings are
+      // enforced immediately even before the acknowledgement is received.
+      // Reminder: the receiver of a SETTINGS frame must process them in the order they are received.
       private def initialLocalSettings = immutable.Seq.empty ++
         maxConcurrentStreams.toSeq.map(value => Setting(SettingIdentifier.SETTINGS_MAX_CONCURRENT_STREAMS, value))
 
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
@@ -75,9 +75,13 @@ private[http2] sealed abstract class Http2ClientDemux
  * In the best case we could just flattenMerge the outgoing side (hoping for the best) but this will probably
  * not work because the sending decision relies on dynamic window size and settings information that will be
  * only available in this stage.
+ *
+ * @param initialRemoteSettings sequence of settings received on the initial header sent from
+ *                              the client in an ' HTTP2-Settings:' header. This parameter should only be used
+ *                              on the server end of a connection.
  */
 @InternalApi
-private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initialDemuxerSettings: immutable.Seq[Setting], upgraded: Boolean, isServer: Boolean) extends GraphStage[BidiShape[Http2SubStream, FrameEvent, FrameEvent, Http2SubStream]] { stage =>
+private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initialRemoteSettings: immutable.Seq[Setting], upgraded: Boolean, isServer: Boolean) extends GraphStage[BidiShape[Http2SubStream, FrameEvent, FrameEvent, Http2SubStream]] { stage =>
   val frameIn = Inlet[FrameEvent]("Demux.frameIn")
   val frameOut = Outlet[FrameEvent]("Demux.frameOut")
 
@@ -107,9 +111,9 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
       private def initialLocalSettings = immutable.Seq(Setting(SettingIdentifier.SETTINGS_MAX_CONCURRENT_STREAMS, maxConcurrentStreams))
 
       override def preStart(): Unit = {
-        if (initialDemuxerSettings.nonEmpty) {
-          debug(s"Applying ${initialDemuxerSettings.length} initial settings!")
-          applySettings(initialDemuxerSettings)
+        if (initialRemoteSettings.nonEmpty) {
+          debug(s"Applying ${initialRemoteSettings.length} initial settings!")
+          applyRemoteSettings(initialRemoteSettings)
         }
 
         pullFrameIn()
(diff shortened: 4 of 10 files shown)


ignasi35:
I'm looking into a failure of `akka.http.h2spec.H2SpecIntegrationSpec`. I wonder if we'd need to relive #1589

ignasi35:
> Increasing the terminal buffer got me to the actual error message...

The failing test is asserting that the new `SETTINGS_MAX_CONCURRENT_STREAMS` sent from the server is enforced in the server.

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -230,6 +250,29 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
...
+       * @return TODO: it's a Boolean but I don't think it has to be.
+       */
+      private def enforceSettings(settings: immutable.Seq[Setting]): Boolean = {
+        var settingsAppliedOk = true
+
+        settings.foreach {
+          case Setting(Http2Protocol.SettingIdentifier.SETTINGS_MAX_CONCURRENT_STREAMS, value) =>
+          // nothing to do here: the count of open incoming streams
+          // depends on `def maxConcurrentStreams` above which is currently not modifiable
+        }
```
ignasi35, line 265: The default `SETTINGS_MAX_CONCURRENT_STREAMS` is _infinite_ and we're enforcing it to `max-concurrent-streams` at construction time already. That could be improved if we enabled the _enforcing_ in this piece of code. Not sure the added complexity is worth it at this point.

ignasi35:
```suggestion
          case Setting(Http2Protocol.SettingIdentifier.SETTINGS_MAX_CONCURRENT_STREAMS, value) =>
          // TODO: Http2Compliance.compliesWithMaxConcurrentStreams should only be enabled once 
          //  we get this SETTINGS_ACK.
        }
```

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -230,6 +250,29 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
...
+
+        settingsAppliedOk
+      }
+
+      /**
+       * Tune this peer to honour the settings sent from the other peer.
```
ignasi35, line 271: Added to disambiguate what settings this was referring to (the local settings VS the settings received from the peer).

akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -74,11 +76,38 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
...
+    val needsMaxConcurrentComplianceCheck = (oldState, newState) match {
+      case (Idle, _) => true
+      case _         => false
+    }
+
+    if (needsMaxConcurrentComplianceCheck) {
+      // incomingStreams contains all the open streams and an extra one that's Idle9
+      val currentConcurrentStreams = incomingStreams.size - 1
+      if (!Http2Compliance.compliesWithMaxConcurrentStreams(currentConcurrentStreams, maxConcurrentStreams)) {
+        pushGOAWAY(
```
ignasi35, line 93:
```suggestion
        // 5.1.2 An endpoint that receives a HEADERS frame that causes its advertised 
        // concurrent stream limit to be exceeded MUST treat this as a stream error 
        // (Section 5.4.2) of type PROTOCOL_ERROR or REFUSED_STREAM. The choice of error 
        // code determines whether the endpoint wishes to enable automatic retry 
        // (see Section 8.1.4) for details).
        pushGOAWAY(
```

I chose GOAWAY so the peer can retry.

ignasi35 (commented):
notes to reviewers and notes to self...

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -922,6 +922,35 @@ class Http2ServerSpec extends AkkaSpecWithMaterializer("""
...
+        )
+
+        // ... and then send one more!
+        val lastValidStreamId = 1 + (maxStreams) * 2
+        val firstInvalidStreamId = 1 + (maxStreams + 1) * 2
+        sendRequestHEADERS(firstInvalidStreamId, req1, endStream = false)
+        // TODO: track lastStreamId more accurately
+        //  (see https://github.com/akka/akka-http/blob/5b645e9a2ffb52fb25e5984a42a794a952856291/akka-http2-support/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala#L126-L133)
+        // val (_, code) = expectGOAWAY(lastValidStreamId)
+        val (_, code) = expectGOAWAY()
```
ignasi35, line 949: #3537 :-)

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -142,7 +158,10 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
           in match {
             case WindowUpdateFrame(streamId, increment) => multiplexer.updateWindow(streamId, increment) // handled specially
             case p: PriorityFrame                       => multiplexer.updatePriority(p)
-            case s: StreamFrameEvent                    => handleStreamEvent(s)
+            case s: StreamFrameEvent => {
+              log.debug(s"StreamFrameEvent= $s")
+              handleStreamEvent(s)
+            }
```
ignasi35, line 164:
```suggestion
             case s: StreamFrameEvent                    => handleStreamEvent(s) 
```

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2FrameProbe.scala
```diff
@@ -173,6 +175,11 @@ private[http] object Http2FrameProbe extends Matchers {
         (lastStreamId, ErrorCode.byId(reader.readIntBE()))
       }
 
+      override def expectSETTINGS(): Unit = {
+        // (6.5) The stream identifier for a SETTINGS frame MUST be zero (0x0)
+        expectFramePayload(FrameType.SETTINGS, Flags.NO_FLAGS, 0) // ignore the payload for the time being.
```
jrudolph, line 180: Just return the payload as a ByteString if it is not yet parsed.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -922,6 +922,35 @@ class Http2ServerSpec extends AkkaSpecWithMaterializer("""
       }
     }
 
+    "enforce settings" should {
+
+      "reject new substreams when exceeding SETTINGS_MAX_CONCURRENT_STREAMS" in new TestSetup with RequestResponseProbes with Http2FrameHpackSupport {
+        // this test exceeds the server setup so we use the server config value as input for the test
+        private val maxStreams: Int = system.settings.config.getInt("akka.http.server.http2.max-concurrent-streams")
```
jrudolph, line 929: Usually we set a fixed value in the settings (via TestSetup constructor).

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -922,6 +922,35 @@ class Http2ServerSpec extends AkkaSpecWithMaterializer("""
...
+            entity = HttpEntity(ContentTypes.`text/plain(UTF-8)`, ""))
+
+        private val streamIds: IndexedSeq[Int] = (0 to maxStreams).map(id => 1 + id * 2)
+        streamIds.foreach(streamId =>
+          sendRequestHEADERS(streamId, req1, endStream = false)
+        )
+
+        // ... and then send one more!
+        val lastValidStreamId = 1 + (maxStreams) * 2
+        val firstInvalidStreamId = 1 + (maxStreams + 1) * 2
```
jrudolph, line 944: Or just `lastValidStreamId + 2` ;)

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -922,6 +922,35 @@ class Http2ServerSpec extends AkkaSpecWithMaterializer("""
...
+            protocol = HttpProtocols.`HTTP/2.0`,
+            entity = HttpEntity(ContentTypes.`text/plain(UTF-8)`, ""))
+
+        private val streamIds: IndexedSeq[Int] = (0 to maxStreams).map(id => 1 + id * 2)
+        streamIds.foreach(streamId =>
+          sendRequestHEADERS(streamId, req1, endStream = false)
+        )
+
+        // ... and then send one more!
+        val lastValidStreamId = 1 + (maxStreams) * 2
```
jrudolph, line 943: Or `streamIds.max`

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -74,11 +76,38 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
...
+    // only validate maxConcurrentStreams compliance if we'd increment the number
+    // of streams in one of the open states. This way, when the value of SETTINGS_MAX_CONCURRNET_STREAMS
+    // decreases, inflight streams can still finish and close.
+    val needsMaxConcurrentComplianceCheck = (oldState, newState) match {
+      case (Idle, _) => true
+      case _         => false
+    }
+
+    if (needsMaxConcurrentComplianceCheck) {
+      // incomingStreams contains all the open streams and an extra one that's Idle9
```
jrudolph, line 90:
```suggestion
      // incomingStreams contains all the open streams and an extra one that's Idle
```

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -74,11 +76,38 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
 
   private def updateState(streamId: Int, handle: IncomingStreamState => IncomingStreamState): Unit = {
     val oldState = streamFor(streamId)
-    val newState = handle(oldState)
+    val newState: IncomingStreamState = handle(oldState)
+
+    // only validate maxConcurrentStreams compliance if we'd increment the number
```
jrudolph, line 81: I'd like to keep this method simple. Could we find a better place to put this logic? E.g. we know that only HEADERS and PUSH_PROMISE frames can create a stream, so we can put the logic into `handleStreamEvent`.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -230,6 +248,29 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
...
+      /**
+       * Tune this peer to enforce the settings configured from this peer.
+       * @return TODO: it's a Boolean but I don't think it has to be.
+       */
+      private def enforceSettings(settings: immutable.Seq[Setting]): Boolean = {
+        var settingsAppliedOk = true
+
+        settings.foreach {
+          case Setting(Http2Protocol.SettingIdentifier.SETTINGS_MAX_CONCURRENT_STREAMS, value) =>
+          // TODO: Http2Compliance.compliesWithMaxConcurrentStreams should only be enabled once
```
jrudolph, line 260: Yes, indeed. Especially if you set it to low values on the server, it's not completely unrealistic that the client already sent you a higher number of requests in the meantime which should work.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -74,11 +76,38 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
...
+    // of streams in one of the open states. This way, when the value of SETTINGS_MAX_CONCURRNET_STREAMS
+    // decreases, inflight streams can still finish and close.
+    val needsMaxConcurrentComplianceCheck = (oldState, newState) match {
+      case (Idle, _) => true
+      case _         => false
+    }
+
+    if (needsMaxConcurrentComplianceCheck) {
+      // incomingStreams contains all the open streams and an extra one that's Idle9
+      val currentConcurrentStreams = incomingStreams.size - 1
```
jrudolph, line 91: Not sure why you would remove that `Idle` one? It should only Idle because you haven't applied the state update further down...

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -922,6 +922,35 @@ class Http2ServerSpec extends AkkaSpecWithMaterializer("""
       }
     }
 
+    "enforce settings" should {
+
+      "reject new substreams when exceeding SETTINGS_MAX_CONCURRENT_STREAMS" in new TestSetup with RequestResponseProbes with Http2FrameHpackSupport {
```
jrudolph, line 927: I think we should also add a test about what happens if we skip stream IDs. In that case, the skipped streams should be automatically be set to `Closed` state.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
jrudolph: And then let's move the enforcing logic into `Http2ServerDemux` and just call that with the current number of open streams.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Compliance.scala
```diff
@@ -28,6 +28,13 @@ private[http2] object Http2Compliance {
   private[akka] sealed class IllegalHttp2FrameSize(size: Int, expected: String)
     extends Http2ProtocolException(ErrorCode.FRAME_SIZE_ERROR, s"Illegal HTTP/2 frame size: [$size]. $expected!")
 
+  // If the current count exceeds the maximum we let them complete. So the verification is
+  // only meant to be used when a new stream is created (but not when the setting is updated.
+  // 5.1.2 "An endpoint that wishes to reduce the value of SETTINGS_MAX_CONCURRENT_STREAMS to a
+  // value that is below the current number of open streams can either close streams that
+  // exceed the new value or allow streams to complete."
+  final def compliesWithMaxConcurrentStreams(currentConcurrentStreams: Int, maxConcurrentStreams: Int): Boolean = currentConcurrentStreams < maxConcurrentStreams
```
jrudolph, line 36: Not sure if it's worth putting that here if it's just this condition. :)

jrudolph (commented):
Looking promising, a few comments below

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2FrameProbe.scala
ignasi35: Isn't the payload of a SETTINGS frame empty?

ignasi35: I guess I'll edit the comment, then 😅

ignasi35:
```suggestion
        // ignore the payload for the time being, SETTINGS frame 
        // should have empty payloads anyway
        expectFramePayload(FrameType.SETTINGS, Flags.NO_FLAGS, 0)         
```

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
ignasi35: Yeah, I have another PR built on top of this one with that improvement:
https://github.com/akka/akka-http/pull/3543/files#diff-708fe2ad051369bad850a1616aecc6abb02a09b61d6a2045b1b762cae57a39a0R929-R932

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
ignasi35: This was an off-by-one error in concordance with the condition. Fixed the condition, fixed the subtraction.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
ignasi35: > it's not completely unrealistic that the client already sent you a higher number of requests in the meantime which should work.

Agreed, what I don't like about the spec is that the default value for `SETTINGS_MAX_CONCURRENT_STREAMS` is _infinite_. A malicious client may ignore the server `SETTINGS` frame (or delay the delivery of a `SETTINGS_ACK` enough) and create too many streams. The SETTINGS_ACK timeout should help prevent/hold that kind of attack but enforcing the limit eagerly, as it is currently implemented, sounds good enough for the time being.

ignasi35: raised https://github.com/akka/akka-http/issues/3551 so we can improve this in the future.

ignasi35:
Still need to address https://github.com/akka/akka-http/pull/3529#discussion_r507604331

ignasi35:
@jrudolph is [`ef8ab04` (#3529)](https://github.com/akka/akka-http/pull/3529/commits/ef8ab04de13907e9db00f2482e16b16fb6e3823e) what you meant?

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2FrameProbe.scala
```diff
@@ -173,6 +175,13 @@ private[http] object Http2FrameProbe extends Matchers {
         (lastStreamId, ErrorCode.byId(reader.readIntBE()))
       }
 
+      override def expectSETTINGS(): Unit = {
+        // (6.5) The stream identifier for a SETTINGS frame MUST be zero (0x0).
+        // Ignore the payload for the time being. SETTINGS frames
+        // should have an empty payload anyway.
```
jrudolph, line 181: No, the payload contains the settings. Let's either return the payload as a ByteString as suggested before or use `Http2FrameParsing` to actually parse the settings.

raboof, line 181: I'm not sure I understand, aren't the settings part of the payload?

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -219,6 +237,41 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
...
+
+        settingsAppliedOk
+      }
+
+      protected override def checkMaxConcurrentStreamsCompliance(currentConcurrentStreams: => Int): Unit = {
+        if (currentConcurrentStreams >= maxConcurrentStreams) {
+          // 5.1.2 An endpoint that receives a HEADERS frame that causes its advertised
+          // concurrent stream limit to be exceeded MUST treat this as a stream error
+          // (Section 5.4.2) of type PROTOCOL_ERROR or REFUSED_STREAM.
+          pushGOAWAY(
```
jrudolph, line 261: Didn't we want to use `RST_STREAM` with `REFUSED_STREAM` instead? Or is that for a follow up PR?

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -922,6 +934,55 @@ class Http2ServerSpec extends AkkaSpecWithMaterializer("""
...
+
+        // while we don't exceed the limit, there's silence on the line
+        expectNoBytes()
+
+        // When we exceed the limit, though...
+        val lastValidStreamId = streamIds.max
+        val firstInvalidStreamId = lastValidStreamId + 2
+        sendHEADERS(firstInvalidStreamId, endStream = false, endHeaders = true, requestHeaderBlock)
+        val (_, code) = expectGOAWAY(lastValidStreamId)
+        code should ===(ErrorCode.PROTOCOL_ERROR)
```
raboof, line 958: :+1:

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -922,6 +934,55 @@ class Http2ServerSpec extends AkkaSpecWithMaterializer("""
...
+      }
+
+      "reject new substreams when exceeding SETTINGS_MAX_CONCURRENT_STREAMS (with closed streams in between)" in new TestSetup with RequestResponseProbes with Http2FrameHpackSupport {
+        def maxStreams: Int = 32
+        val requestHeaderBlock: ByteString = HPackSpecExamples.C41FirstRequestWithHuffman
+        override def settings = super.settings.mapHttp2Settings(_.withMaxConcurrentStreams(maxStreams))
+
+        // the Seq of stream ids has gaps
```
raboof, line 966: If the implementation becomes more strict in the future it might not allow gaps and require you to actually open and close those streams - but OK to deal with that then ;)

raboof (commented):
Mostly LGTM, some comments

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -219,6 +237,41 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
...
+      /**
+       * Tune this peer to enforce the settings configured from this peer.
+       * @return TODO: it's a Boolean but I don't think it has to be.
+       */
+      private def enforceSettings(settings: immutable.Seq[Setting]): Boolean = {
+        var settingsAppliedOk = true
+
+        settings.foreach {
+          case Setting(Http2Protocol.SettingIdentifier.SETTINGS_MAX_CONCURRENT_STREAMS, value) =>
+          // TODO: Http2Compliance.compliesWithMaxConcurrentStreams should only be enabled once
```
jrudolph, line 249: Is this comment still correct? Maybe we should just remove the whole method?

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -160,13 +176,15 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
...
-            // Currently, we only expect an ack for the initial (currently empty) settings frame, sent
-            // above in preStart. Since, it was empty, there's nothing to do here.
-            // If we want to support setting and enforcing settings, we'll need to act here to commit
-            // to the settings we sent out before.
-            // https://github.com/akka/akka-http/issues/3185
-
-            case PingFrame(true, _)    =>
+              // Currently, we only expect an ack for the initial settings frame, sent
+              // above in preStart. Since, only some settings are supported, and those
+              // settings are non-modifiable and known at construction time, there's not much to do here.
```
jrudolph, line 181:
```suggestion
              // settings are non-modifiable and known at construction time, these settings
              // are enforced from the start of the connection.
```

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -91,10 +95,21 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
...
       override protected def logSource: Class[_] = if (isServer) classOf[Http2ServerDemux] else classOf[Http2ClientDemux]
 
       val multiplexer = createMultiplexer(frameOut, StreamPrioritizer.first())
 
+      // TODO: we only support the initial SETTINGS message (the only one that's compulsory)
+      //  so we're keeping the settings as a field instead of implementing SETTINGS-SETTINGS_ACK
+      //  correlation logic.
+      // TODO: the receiver of a SETTINGS frame must apply them in the order they
+      //  are received. Review order of the settings in the sequence since it's relevant.
+      val localSettings = immutable.Seq(
```
jrudolph, line 109: Let's make it a `def` and call it `initialSettings`.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -91,10 +95,21 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
...
+      def maxConcurrentStreams: Int = http2Settings.maxConcurrentStreams
+
       override protected def logSource: Class[_] = if (isServer) classOf[Http2ServerDemux] else classOf[Http2ClientDemux]
 
       val multiplexer = createMultiplexer(frameOut, StreamPrioritizer.first())
 
+      // TODO: we only support the initial SETTINGS message (the only one that's compulsory)
+      //  so we're keeping the settings as a field instead of implementing SETTINGS-SETTINGS_ACK
+      //  correlation logic.
+      // TODO: the receiver of a SETTINGS frame must apply them in the order they
```
jrudolph, line 107: The TODOs can be removed. Let's add a comment like this:

```
// Send settings initially based on our configuration. For simplicity, these settings are
// enforced immediately even before the acknowledgement is received.
```

akka-http-core/src/main/resources/reference.conf
```diff
@@ -225,6 +225,9 @@ akka.http {
 
     http2 {
       # The maximum number of request per connection concurrently dispatched to the request handler.
+      # This limit is enforced as soon as the connection between the peers is established. Enforcing
+      # the limit even before the SETTINGS/SETTINGS_ACK echange happens between the peers is more
+      # strict than required in the HTTP/2 Spec.
```
jrudolph, line 230:
```suggestion
      # the limit even before the SETTINGS/SETTINGS_ACK exchange has completed, means
      # that we will refuse extra streams/requests that were sent by the client right after 
      # the connection was established but before it received our SETTINGS.
```

akka-http-core/src/main/resources/reference.conf
```diff
@@ -374,6 +377,9 @@ akka.http {
     // FIXME: unify with server-side part (by importing or similar to parsing)
     http2 {
       # The maximum number of request per connection concurrently dispatched to the request handler.
+      # This limit is enforced as soon as the connection between the peers is established. Enforcing
+      # the limit even before the SETTINGS/SETTINGS_ACK echange happens between the peers is more
+      # strict than required in the HTTP/2 Spec.
```
jrudolph, line 382:
```suggestion
      # the limit even before the SETTINGS/SETTINGS_ACK exchange has completed, means
      # that we will refuse extra streams/push promises that were sent by the client right after 
      # the connection was established but before it received our SETTINGS.
```

jrudolph (commented):
LGTM, mostly some cosmetic issues left

akka-http-core/src/main/scala/akka/http/scaladsl/settings/Http2ServerSettings.scala
```diff
@@ -76,6 +76,7 @@ object Http2ServerSettings extends SettingsCompanion[Http2ServerSettings] {
     logFrames:                         Boolean,
     internalSettings:                  Option[Http2InternalServerSettings])
     extends Http2ServerSettings {
+    require(maxConcurrentStreams >= 0, "max-concurrent-streams must be >= 0")
```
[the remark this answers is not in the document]
jrudolph: The default by the spec is infinite, so it might make sense after all (in which case we wouldn't send out this settings value).

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
jrudolph: Is there any indication that they would change this?

It seems like a conscious feature. The way it is currently used, e.g. by firefox to set up a priority hierarchy with the first few streams but never actually use those streams. So, those streams will be immediately closed but if you track priorities correctly, you will use those stream identifiers as part of the priority hierarchy.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2FrameProbe.scala
raboof: (I already added an `expectFrame` that returns the parsed `FrameEvent` that you can then pattern-match on. perhaps it'd be neat to give that a generic parameter so `expectFrame[SettingsFrame]()` throws when another type of frame is encountered, and returns the frame in the correct type right away. Anyway, not necessarily for this PR :) )

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
raboof: Interesting! I thought the spec might disallow gaps but indeed it doesn't (and that won't change)

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -91,10 +95,21 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
...
 
+      def maxConcurrentStreams: Int = http2Settings.maxConcurrentStreams
+
       override protected def logSource: Class[_] = if (isServer) classOf[Http2ServerDemux] else classOf[Http2ClientDemux]
 
       val multiplexer = createMultiplexer(frameOut, StreamPrioritizer.first())
 
+      // TODO: we only support the initial SETTINGS message (the only one that's compulsory)
+      //  so we're keeping the settings as a field instead of implementing SETTINGS-SETTINGS_ACK
+      //  correlation logic.
```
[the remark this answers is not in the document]
ignasi35: Good call. Raised #3560 
I don't think we should support sending new `SETTINGS` frames once the connection is established.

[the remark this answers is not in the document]
raboof: If we shouldn't send new `SETTINGS` frames once the connection is established, perhaps we should just remove the TODO and the github issue? (I agree I don't see a reason we should)

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -91,10 +95,21 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
...
+      // TODO: the receiver of a SETTINGS frame must apply them in the order they
+      //  are received. Review order of the settings in the sequence since it's relevant.
```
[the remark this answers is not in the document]
ignasi35: I added this comment as a reminder that anytime we add a setting on the list, the ordering is meaningful according to the spec. It shouldn't be a `TODO`.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -219,6 +237,41 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
         }
       })
 
+      /**
+       * Tune this peer to enforce the settings configured from this peer.
+       * @return TODO: it's a Boolean but I don't think it has to be.
+       */
+      private def enforceSettings(settings: immutable.Seq[Setting]): Boolean = {
```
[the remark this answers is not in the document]
ignasi35: Correct.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -60,10 +60,23 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
           Idle
         }
     }
+
   def handleStreamEvent(e: StreamFrameEvent): Unit = {
+    // only validate maxConcurrentStreams compliance if we'd increment the number
+    // of streams in one of the open states. This way, when the value of SETTINGS_MAX_CONCURRENT_STREAMS
+    // decreases, in-flight streams can still finish and close.
```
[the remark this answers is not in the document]
ignasi35: If we implement enforcing of this setting correctly (even if we support a single, initial `SETTINGS` frame), the default value a peer sees for `max-concurrent-streams` is _infinite_ so a very fast client may have already open a number of frames over the value in the settings.
It's a small case but considering the `decrease` case already doesn't harm.
Related #3551

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -60,10 +60,23 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
...
+    e match {
+      case _: ParsedHeadersFrame =>
+        checkMaxConcurrentStreamsCompliance(incomingStreams.size)
+      case _: PushPromiseFrame =>
+        checkMaxConcurrentStreamsCompliance(incomingStreams.size)
+      case _ =>
+    }
```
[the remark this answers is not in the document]
ignasi35: > Should we check this after updating the state? When updating the state might increate the number of incoming streams?

Interesting. Let me have a look.
Hmm, it's a thin line. 

So, this check is only run when receiving a HEADERS or a PUSH_PROMISE frames. Should we process a HEADERS/PUSH_PROMISE which won't even create a stream? 
I mean, HEADERS/PUSH_PROMISE will generally create a stream. But there are cases in which a HEADERS/PUSH_PROMISE immediately transitions to `Closed` (when END_STREAM is set).

I agre that moving the check to _after_ `updateState()` sounds harmless but I'm not sure I see it being more correct.

What am I missing?

akka-http-core/src/main/scala/akka/http/scaladsl/settings/Http2ServerSettings.scala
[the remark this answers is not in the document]
ignasi35: Good. Then I'll add support for it and turn the `maxConcurrentStreams` field into `Option[Int]`).

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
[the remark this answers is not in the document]
jrudolph: I think it becomes even more relevant when using RST_STREAM instead of PROTOCOL_ERROR. In that case, we could move the logic to `expectIncomingStream`, use `resetStream` to refuse the stream right away and set the state to CLOSED. WDYT?

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2FrameProbe.scala
ignasi35: I don't know what I was thinking when I wrote that. The last sentence is not correct.

akka-http-core/src/main/scala/akka/http/scaladsl/settings/Http2ServerSettings.scala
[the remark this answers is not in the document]
jrudolph: Could we use Int.MaxValue?

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2FrameProbe.scala
ignasi35: I think there's a bit of a mess in the methods we have in the probe. For example, the new [`expectFrame()`](https://github.com/akka/akka-http/blob/master/akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2FrameProbe.scala#L113) is useful to expect a frame but, unlike other `expectXyz` methods it doesn't validate any field on the frame (e.g. `expectDATAFrame(streamId: Int)` validates the `streamId` is the value expected).

I think we have an opportunity to review these methods and, for example, add the generic param @raboof mentioned above.

I'd rather do it on a separate PR.

ignasi35: > Let's either return the payload as a ByteString as suggested before or use Http2FrameParsing to actually parse the settings.

+1 I've improved it to return frame.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
ignasi35: My bad. I overlooked the change.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
ignasi35: I used `localSettings` so it's clear it's not settings received from the peer. How about `initialLocalSettings`?

jrudolph: Sounds good

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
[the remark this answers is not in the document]
ignasi35: Let me rephrase:

> I don't think we should support sending new SETTINGS frames once the connection is established.
... in this PR.
I don't know if we should or shouldn't support it in the future. The issue is there now to discuss the convenience of that future effort. :-)

akka-http-core/src/main/scala/akka/http/scaladsl/settings/Http2ServerSettings.scala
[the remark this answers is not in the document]
ignasi35: I was just doing a pass at converting the field to `Option[Int]`. After initially regretting the change from `Int` to `Option[Int]` I realized we're using `maxConcurrentStreams` to setup the parallelism:

      Http2().bindAndHandleAsync(...
        definitiveSettings.http2Settings.maxConcurrentStreams, log)(fm)

I think moving to `Option[Int]` is the right choice in this setting.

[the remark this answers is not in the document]
jrudolph: Ah, I forgot about `parallelism`, how would we chose that then? Maybe let's not go into that right now and do that change in another PR.

akka-http-core/src/main/resources/reference.conf
```diff
@@ -374,6 +378,10 @@ akka.http {
     // FIXME: unify with server-side part (by importing or similar to parsing)
     http2 {
       # The maximum number of request per connection concurrently dispatched to the request handler.
+      # This limit is enforced as soon as the connection between the peers is established. Enforcing
+      # the limit even before the SETTINGS/SETTINGS_ACK exchange has completed, means
+      # that we will refuse extra streams/push promises that were sent by the client right after 
+      # the connection was established but before it received our SETTINGS.
```
ignasi35, line 384: As discussed, `"infinite"` is not supported and the value of `max-concurrent-streams` will also be used as the `parallelism` on the stream.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -71,9 +75,13 @@ private[http2] sealed abstract class Http2ClientDemux
...
  * not work because the sending decision relies on dynamic window size and settings information that will be
  * only available in this stage.
+ *
+ * @param initialRemoteSettings sequence of settings received on the initial header sent from
+ *                              the client in an ' HTTP2-Settings:' header. This parameter should only be used
+ *                              on the server end of a connection.
  */
 @InternalApi
-private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initialDemuxerSettings: immutable.Seq[Setting], upgraded: Boolean, isServer: Boolean) extends GraphStage[BidiShape[Http2SubStream, FrameEvent, FrameEvent, Http2SubStream]] { stage =>
+private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initialRemoteSettings: immutable.Seq[Setting], upgraded: Boolean, isServer: Boolean) extends GraphStage[BidiShape[Http2SubStream, FrameEvent, FrameEvent, Http2SubStream]] { stage =>
```
ignasi35, line 84: I renamed and documented this param because it confused me over and over.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -91,20 +99,28 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
...
+      // Reminder: the receiver of a SETTINGS frame must process them in the order they are received.
+      private def initialLocalSettings = immutable.Seq(Setting(SettingIdentifier.SETTINGS_MAX_CONCURRENT_STREAMS, maxConcurrentStreams))
+
       override def preStart(): Unit = {
-        if (initialDemuxerSettings.nonEmpty) {
-          debug(s"Applying ${initialDemuxerSettings.length} initial settings!")
-          applySettings(initialDemuxerSettings)
+        if (initialRemoteSettings.nonEmpty) {
+          debug(s"Applying ${initialRemoteSettings.length} initial settings!")
+          applyRemoteSettings(initialRemoteSettings)
```
ignasi35, line 116: Renaming the `applySettings` aswell.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -91,20 +99,28 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
...
         }
 
         pullFrameIn()
         pull(substreamIn)
 
-        multiplexer.pushControlFrame(SettingsFrame(Nil)) // both client and server must send an settings frame as first frame
+        // both client and server must send a settings frame as first frame
```
ignasi35, line 122: I just realized this comment (and implementation) may be wrong. This is not to be fixed in this PR, but when implementing the client, if we sent the initial settings as the `Upgrade` header then we should not push a `SettingsFrame`.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -136,37 +152,22 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
...
-                  }
-                case Setting(Http2Protocol.SettingIdentifier.SETTINGS_MAX_FRAME_SIZE, value) =>
-                  multiplexer.updateMaxFrameSize(value)
-                case Setting(Http2Protocol.SettingIdentifier.SETTINGS_MAX_CONCURRENT_STREAMS, value) =>
-                  debug(s"Setting max concurrent streams to $value (not enforced)")
-                case Setting(id, value) =>
-                  debug(s"Ignoring setting $id -> $value")
-              }
-
+              val settingsAppliedOk = applyRemoteSettings(settings)
```
ignasi35, line 155: Extracted code duplication (reuse method)

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -136,37 +152,22 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
...
-            // https://github.com/akka/akka-http/issues/3185
-
-            case PingFrame(true, _)    =>
+            case SettingsAckFrame(_) =>
+              // Currently, we only expect an ack for the initial settings frame, sent
+              // above in preStart. Since, only some settings are supported, and those
+              // settings are non-modifiable and known at construction time, these settings
+              // are enforced from the start of the connection.
+              // Also, since we only expect an ack for the initial settings frame there's
+              // no need to correlate the ACK to a particular SETTINGS frame.
```
ignasi35, line 166:
```suggestion
```

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -234,7 +259,7 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
           case Setting(Http2Protocol.SettingIdentifier.SETTINGS_MAX_FRAME_SIZE, value) =>
             multiplexer.updateMaxFrameSize(value)
           case Setting(Http2Protocol.SettingIdentifier.SETTINGS_MAX_CONCURRENT_STREAMS, value) =>
-            debug(s"Setting max concurrent streams to $value (not enforced)")
+            debug(s"Setting max concurrent streams to $value (not respected)")
```
ignasi35, line 262: I think `enforcing` applies to settings the local peer set and requires, while `respect` is what the local peer should do with remote settings.
This method applies remote settings and logs that "Hey, I received `SETTINGS_MAX_CONCURRENT_STREAMS` but I'm not respecting it".

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -143,7 +146,14 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
...
   case object Idle extends IncomingStreamState {
-    def handle(event: StreamFrameEvent): IncomingStreamState = expectIncomingStream(event, HalfClosedRemote, ReceivingDataFirst)
+    def handle(event: StreamFrameEvent): IncomingStreamState = {
+      if (incomingStreams.size > settings.maxConcurrentStreams) {
+        resetStream(event.streamId, ErrorCode.REFUSED_STREAM)
+        Closed
+      } else {
+        expectIncomingStream(event, HalfClosedRemote, ReceivingDataFirst)
+      }
+    }
```
ignasi35, line 156: this is the change where the setting is enforced. I only implemented enforcing on `Idle` instead of adding it into `expectIncomingStream` because `expectIncomingStream` is invoked from other states. This solution makes it possible to decrease the value of `maxConcurrentStreams` (in a future PR) and let the in-flight streams to finish.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2FrameProbe.scala
```diff
@@ -173,6 +175,12 @@ private[http] object Http2FrameProbe extends Matchers {
         (lastStreamId, ErrorCode.byId(reader.readIntBE()))
       }
 
+      override def expectSETTINGS(): FrameEvent = {
+        // (6.5) The stream identifier for a SETTINGS frame MUST be zero (0x0).
+        val payload = expectFramePayload(FrameType.SETTINGS, Flags.NO_FLAGS, 0)
+        Http2FrameParsing.parseFrame(FrameType.SETTINGS, Flags.NO_FLAGS, 0, new ByteReader(payload), system.log)
+      }
```
ignasi35, line 182: Expect a SETTINGS without flags (this is no ACK) and return the whole Frame

ignasi35 (commented):
Add a few notes for reviewers.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
raboof: So this is something we'll have to doublecheck when we implement the client negotiating HTTP/2 via the `Upgrade` mechanism :+1:

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
raboof: I can see that. TBQH 'remote' can still mean 'the settings we got from the remote' or 'the settings we sent to the remote', I wonder if 'receivedSettings' would be even less ambiguous?

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -922,6 +934,53 @@ class Http2ServerSpec extends AkkaSpecWithMaterializer("""
...
+        expectNoBytes()
+
+        // When we exceed the limit, though...
+        val lastValidStreamId = streamIds.max
+        val firstInvalidStreamId = lastValidStreamId + 2
+        sendHEADERS(firstInvalidStreamId, endStream = false, endHeaders = true, requestHeaderBlock)
+        expectRST_STREAM(firstInvalidStreamId, ErrorCode.REFUSED_STREAM)
+      }
+
+      "reject new substreams when exceeding SETTINGS_MAX_CONCURRENT_STREAMS (with closed streams in between)" in new TestSetup with RequestResponseProbes with Http2FrameHpackSupport {
```
raboof, line 960: It would be even more reliable when this test would actually open-and-close some streams rather than just skipping some stream ID's, but this is already :+1:

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -922,6 +934,53 @@ class Http2ServerSpec extends AkkaSpecWithMaterializer("""
       }
     }
 
+    "enforce settings" should {
+
+      "reject new substreams when exceeding SETTINGS_MAX_CONCURRENT_STREAMS" in new TestSetup with RequestResponseProbes {
```
raboof, line 939: We can remove the TODO test `received SETTINGS_MAX_CONCURRENT_STREAMS should limit the number of streams` on line 915 now ;)

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/H2SpecIntegrationSpec.scala
```diff
@@ -96,6 +96,7 @@ class H2SpecIntegrationSpec extends AkkaSpec(
       "4.3",
       "5.1",
       "5.1.1",
+      "5.1.2",
```
raboof, line 99: so h2spec expects us to produce a PROTOCOL_ERROR but we produce a REFUSED_STREAM, which is also allowed by the spec. Too bad but :+1:

raboof (approved):

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
ignasi35: Ah! No, we can't remove that test! ;-)
That test is about respecting remote settings from _this_ peer.

Note this PR is only about sending the setting and enforcing it on incoming streams. We should use `enforce` and `respect` carefully.

Bonus: the test in line 915 (respect the client-set `SETTINGS_MAX_CONCURRENT_STREAMS` from the server) only makes sense once we add support for PUSH_PROMISE.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/H2SpecIntegrationSpec.scala
ignasi35: No no. The problem is `h2spec` tries to create too many connections but sends `HEADERS` frames with the `END_STREAM` flag so we immediately close the streams not triggering the limit. 

I don't know what type of error is expected in `h2spec` when too many streams are open.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
raboof: Ah you're right of course :+1:

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/H2SpecIntegrationSpec.scala
raboof: Ah, right, I misread its outcome (and didn't remember the particulars of our previous discussion on this). I think it accepts all 4 variations (goaway/rst x protocol_error/refused).

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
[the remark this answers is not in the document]
jrudolph: Let's remove it :) Just add a comment above where handling `SettingsAckFrame` that we currently enforce all settings from the start of the connection.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
jrudolph: But that assumes that every received event here will open a stream. At least `PRIORITY` frames can be received in this stream without opening it.

> this is the change where the setting is enforced. I only implemented enforcing on `Idle` instead of adding it into `expectIncomingStream` because `expectIncomingStream` is invoked from other states. This solution makes it possible to decrease the value of `maxConcurrentStreams` (in a future PR) and let the in-flight streams to finish.

You could also state it differently: `maxConcurrentStreams` is directional and should only account for streams opened from the peer. Right now we count all streams which is only correct for the server side in cases without `PUSH_PROMISE`. On the client side, we move out of this state using `handleOutgoingCreated` so we side-step the issue.

Maybe it's enough to add a check that the event actually is a `ParsedHeaderFrame`?

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -922,6 +934,53 @@ class Http2ServerSpec extends AkkaSpecWithMaterializer("""
...
+        val requestHeaderBlock: ByteString = HPackSpecExamples.C41FirstRequestWithHuffman
+
+        // start as many streams as max concurrent...
+        private val streamIds: IndexedSeq[Int] = (0 until maxStreams).map(id => 1 + id * 2)
+        streamIds.foreach(streamId =>
+          sendHEADERS(streamId, endStream = false, endHeaders = true, requestHeaderBlock)
+        )
+
+        // while we don't exceed the limit, there's silence on the line
+        expectNoBytes()
```
jrudolph, line 951:
```suggestion
        expectNoBytes(100.millis)
```

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -922,6 +934,53 @@ class Http2ServerSpec extends AkkaSpecWithMaterializer("""
...
+
+        // the Seq of stream ids has gaps
+        // the skipped values should be represented as automatically-closed streams
+        private val streamIds: IndexedSeq[Int] = (0 until maxStreams).map(id => 1 + id * 4)
+        streamIds.foreach { streamId =>
+          sendHEADERS(streamId, endStream = false, endHeaders = true, requestHeaderBlock)
+        }
+
+        // while we don't exceed the limit, there's silence on the line
+        expectNoBytes()
```
jrudolph, line 973:
```suggestion
        expectNoBytes(100.millis)
```

jrudolph (commented):
A few more minor comments but now almost ready to go!

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
ignasi35: I didn't consider `PRIORITY` frames and I agree they should not trigger the verification. After a quick review, it makes sense to me to only check the number of concurrent streams when processing a `ParsedHeaderFrame`.

We may need to revisit this logic when supporting push but we'll cross that bridge when we get there.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/FrameEvent.scala
```diff
@@ -40,48 +48,62 @@ private[http] object FrameEvent {
...
      * That means this size amounts to data size + padding size field + padding.
      */
-    def sizeInWindow: Int = payload.size // FIXME: take padding size into account, #1313
+    def sizeInWindow: Int =
+      payload.size // FIXME: take padding size into account, #1313
   }
 
+  /**
+   * Representation of the wire-level HEADERS frame. Together with ContinuationFrame's and instance of
+   * HeadersFrame will be parsed into a logical ParsedHeadersFrame.
```
jrudolph, line 57:
```suggestion
   * Representation of the wire-level HEADERS frame. Together with ContinuationFrames an instance of
   * HeadersFrame will be parsed into a logical ParsedHeadersFrame.
```

akka-http-core/src/main/scala/akka/http/impl/engine/http2/FrameEvent.scala
```diff
@@ -40,48 +48,62 @@ private[http] object FrameEvent {
      *
      * That means this size amounts to data size + padding size field + padding.
      */
-    def sizeInWindow: Int = payload.size // FIXME: take padding size into account, #1313
+    def sizeInWindow: Int =
```
jrudolph, line 51: Can we please keep the compact form for those one-liners. :)

akka-http-core/src/main/scala/akka/http/impl/engine/http2/FrameEvent.scala
```diff
@@ -40,48 +48,62 @@ private[http] object FrameEvent {
...
     endHeaders:          Boolean,
     promisedStreamId:    Int,
-    headerBlockFragment: ByteString) extends StreamFrameEvent
+    headerBlockFragment: ByteString)
+    extends StreamFrameEvent
 
-  final case class RstStreamFrame(streamId: Int, errorCode: ErrorCode) extends StreamFrameEvent
-  final case class SettingsFrame(settings: immutable.Seq[Setting]) extends FrameEvent
-  final case class SettingsAckFrame(acked: immutable.Seq[Setting]) extends FrameEvent
+  final case class RstStreamFrame(streamId: Int, errorCode: ErrorCode)
```
jrudolph, line 78: Those should stay one-liners, too.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/FrameEvent.scala
```diff
@@ -40,48 +48,62 @@ private[http] object FrameEvent {
...
+    extends StreamFrameEvent
+  final case class SettingsFrame(settings: immutable.Seq[Setting])
+    extends FrameEvent
+  final case class SettingsAckFrame(acked: immutable.Seq[Setting])
+    extends FrameEvent
 
   case class PingFrame(ack: Boolean, data: ByteString) extends FrameEvent {
-    require(data.size == 8, s"PingFrame payload must be of size 8 but was ${data.size}")
+    require(
```
jrudolph, line 86: This one as well

akka-http-core/src/main/scala/akka/http/impl/engine/http2/FrameEvent.scala
```diff
@@ -40,48 +48,62 @@ private[http] object FrameEvent {
...
+    extends StreamFrameEvent
 
-  final case class Setting(
-    identifier: SettingIdentifier,
-    value:      Int)
+  final case class Setting(identifier: SettingIdentifier, value: Int)
 
   object Setting {
-    implicit def autoConvertFromTuple(tuple: (SettingIdentifier, Int)): Setting =
+    implicit def autoConvertFromTuple(
```
jrudolph, line 104: And this one

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -91,20 +99,28 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
       override def settings: Http2CommonSettings = http2Settings
       override def isUpgraded: Boolean = upgraded
 
+      def maxConcurrentStreams: Int = http2Settings.maxConcurrentStreams
```
jrudolph, line 102: Maybe we should call that

```suggestion
      def localMaxConcurrentStreamsLimit: Int = http2Settings.maxConcurrentStreams
```

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -28,15 +28,17 @@ import scala.util.control.NoStackTrace
...
+  self: GraphStageLogic with LogHelper =>
   // required API from demux
   def multiplexer: Http2Multiplexer
   def settings: Http2CommonSettings
   def pushGOAWAY(errorCode: ErrorCode, debug: String): Unit
   def dispatchSubstream(sub: Http2SubStream): Unit
   def isUpgraded: Boolean
 
-  def flowController: IncomingFlowController = IncomingFlowController.default(settings)
+  def flowController: IncomingFlowController =
```
jrudolph, line 40: keep one-liner

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -47,7 +49,8 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
     incomingStreams.get(streamId) match {
       case Some(state) => state
       case None =>
-        if (streamId <= largestIncomingStreamId) Closed // closed streams are never put into the map
+        if (streamId <= largestIncomingStreamId)
```
jrudolph, line 52: keep one-liner

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -79,20 +83,31 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
...
   def lastStreamId(): Int = largestIncomingStreamId
 
-  private def updateState(streamId: Int, handle: IncomingStreamState => IncomingStreamState): Unit = {
+  private def updateState(
+    streamId: Int,
+    handle:   IncomingStreamState => IncomingStreamState
+  ): Unit = {
     val oldState = streamFor(streamId)
-    val newState = handle(oldState)
+    val newState: IncomingStreamState = handle(oldState)
```
jrudolph, line 91:
```suggestion
    val newState = handle(oldState)
```

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -79,20 +83,31 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
...
-    debug(s"Incoming side of stream [$streamId] changed state: ${oldState.stateName} -> ${newState.stateName}")
+
+    debug(
+      s"Incoming side of stream [$streamId] changed state: ${oldState.stateName} -> ${newState.stateName}"
+    )
   }
+
   /** Called to cleanup any state when the connection is torn down */
-  def shutdownStreamHandling(): Unit = incomingStreams.keys.foreach(id => updateState(id, _.shutdown()))
+  def shutdownStreamHandling(): Unit =
```
jrudolph, line 104: keep one-liner

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -79,20 +83,31 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
...
+    incomingStreams.keys.foreach(id => updateState(id, _.shutdown()))
   def resetStream(streamId: Int, errorCode: ErrorCode): Unit = {
     incomingStreams -= streamId
-    debug(s"Incoming side of stream [$streamId]: resetting with code [$errorCode]")
+    debug(
```
jrudolph, line 108: keep one-liner

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -104,31 +119,43 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
 
     def stateName: String = productPrefix
     def handleOutgoingCreated(stream: Http2SubStream): IncomingStreamState = {
-      warning(s"handleOutgoingCreated received unexpectedly in state $stateName. This indicates a bug in Akka HTTP, please report it to the issue tracker.")
+      warning(
```
jrudolph, line 122: keep one-liner

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -104,31 +119,43 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
...
+        s"handleOutgoingCreated received unexpectedly in state $stateName. This indicates a bug in Akka HTTP, please report it to the issue tracker."
+      )
       this
     }
     def handleOutgoingEnded(): IncomingStreamState = {
-      warning(s"handleOutgoingEnded received unexpectedly in state $stateName. This indicates a bug in Akka HTTP, please report it to the issue tracker.")
+      warning(
```
jrudolph, line 128: keep one-liner

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -104,31 +119,43 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
...
+    ): IncomingStreamState =
       event match {
         case frame @ ParsedHeadersFrame(streamId, endStream, _, _) =>
           val (data, nextState) =
             if (endStream)
               (Source.empty, nextStateEmpty)
             else {
-              val subSource = new SubSourceOutlet[ByteString](s"substream-out-$streamId")
-              (Source.fromGraph(subSource.source), nextStateStream(new IncomingStreamBuffer(streamId, subSource)))
+              val subSource =
```
jrudolph, line 153: keep one-liner

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -104,31 +119,43 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
...
+      )
       Closed
     }
 
     protected def expectIncomingStream(
       event:           StreamFrameEvent,
       nextStateEmpty:  IncomingStreamState,
       nextStateStream: IncomingStreamBuffer => IncomingStreamState,
-      mapStream:       Http2SubStream => Http2SubStream            = identity): IncomingStreamState =
+      mapStream:       Http2SubStream => Http2SubStream            = identity
```
jrudolph, line 145: keep one-liner

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -143,44 +170,83 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
...
+      if (event
+        .isInstanceOf[ParsedHeadersFrame] && incomingStreams.size > settings.maxConcurrentStreams) {
+        // When trying to open a new Stream, if that op would exceed the maxConcurrentStreams, then refuse the op
+        resetStream(event.streamId, ErrorCode.REFUSED_STREAM)
+        Closed
+      } else {
+        expectIncomingStream(event, HalfClosedRemote, ReceivingDataFirst)
+      }
+    }
+    override def handleOutgoingCreated(
```
jrudolph, line 183: keep one-liner

jrudolph (commented):
I started reviewing but there are now so many unwanted (accidental?) whitespace changes that the change is hard to review and a huge pain to resolve conflicts with my ongoing PR. So, could you first remove all the whitespace changes? I think they were all introduced with the last commit, so it shouldn't be too hard.

ignasi35:
> I started reviewing but there are now so many unwanted (accidental?) whitespace changes...

Fixed. I re-imported the project into IntelliJ which enabled a few default settings re `scalafmt` and auto-format on save. Unfortunately, the changes introduced are changes `scalariform` pays no attention to.

jrudolph (approved):
LGTM

jrudolph:
Done :partying_face:, thanks @ignasi35.

PR #3532, opened 2020-10-14
http2: protocol-level client specs

raboof:
Infrastructure for 'protocol-level' tests for the client side: allows testing the translation from request object to network frames, and from response network frames to response object.

Heavily inspired by `Http2ServerSpec`, and sharing some of the code.

Kept most of `Http2ServerSpec` as commented-out `Http2ClientSpec`, as for some of those it might make sense to include the 'mirrored' scenario.

PR #3535, opened 2020-10-15
http2: use scheme portion of target URI in :scheme pseudo-header

raboof:
https://tools.ietf.org/html/rfc7540#section-8.1.2.3

Tests for this are already in #3532


jrudolph (approved):
Not sure the mentioned proxy use cases in the spec are relevant but I'm still fine with that change as it probably won't change anything in the common case.

jrudolph:
> Tests for this are already in #3532

Yes, but they are now failing, is that what you meant?

jrudolph:
> > Tests for this are already in #3532
> 
> Yes, but they are now failing, is that what you meant?

Fixed in https://github.com/akka/akka-http/pull/3526/commits/8edcd7cd82b281a5fbd2b1a1d05bbe687180d944

raboof:
> > Tests for this are already in #3532
> 
> Yes, but they are now failing, is that what you meant?

Yes, thanks for fixing :D

PR #3537, opened 2020-10-15
More accurately track the last-stream-id for GOAWAY

raboof:
Though this value is somewhat best-effort, if I'm reading the [spec](https://tools.ietf.org/html/rfc7540#section-6.8)
right we should over-estimate, not under-estimate.

This probably conflicts with the ongoing refactor in this area, I'm happy to rebase when that is done :)

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
@@ -107,19 +107,8 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
         multiplexer.pushControlFrame(SettingsFrame(Nil)) // both client and server must send an settings frame as first frame
       }
 
-      /**
-       * The "last peer-initiated stream that was or might be processed on the sending endpoint in this connection"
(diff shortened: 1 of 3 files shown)


ignasi35:
Had an early peek and it LGTM.

@raboof are you holding back (keeping it DRAFT) for any particular improvement that's pending? or just waiting for other PRs to land and then rebase this?

raboof:
> are you holding back (keeping it DRAFT) for any particular improvement that's pending? or just waiting for other PRs to land and then rebase this?

Just because it's a relatively small change that I don't want to interfere with @jrudolph 's currently ongoing refactor

jrudolph (approved):
LGTM, in fact the previous behavior wasn't just inaccurate but, in fact, a bug. The new behavior is more conservative because we say that we might have processed all incoming streams instead of claiming we had not done any processing on most streams...

jrudolph:
> > are you holding back (keeping it DRAFT) for any particular improvement that's pending? or just waiting for other PRs to land and then rebase this?
> 
> Just because it's a relatively small change that I don't want to interfere with @jrudolph 's currently ongoing refactor

No worries about that one, this one is small enough to avoid conflicts here.

ignasi35 (approved):

ignasi35:
I've merged and will rebase #3529 to use this already.

PR #3539, opened 2020-10-15
build: replace jabba by proper debian installation

jrudolph:


raboof (approved):
IIRC we moved to jabba because they make it easier to switch to specific other versions, but I agree it seems to be causing more trouble than it solves now.

ignasi35:
> IIRC we moved to jabba because they make it easier to switch to specific other versions, but I agree it seems to be causing more trouble than it solves now.

How about converging to `sdkman`? Either by copy-pasting the YAML from https://github.com/scala/scala-dev/blob/scala-dev/travis/default.yml or using it as a parent (related to https://github.com/akka/akka-meta/issues/141)

ignasi35:
> IIRC we moved to jabba because they make it easier to switch to specific other versions, but I agree it seems to be causing more trouble than it solves now.

Some of the issues in jabba come from caching the JDKs. TIL TravisCI [recommends _not_ caching JDKs](https://docs.travis-ci.com/user/caching/#things-not-to-cache). If I'm reading this PR correctly, the suggested change would not cache the JDK either, right?

Maybe an intermediate solution is to remain in jabba without caching JDKs?

jrudolph:
What's the precedence for using sdkman as an official way of installing packages? I'd vote for using an official repository instead of introducing unnecessary third-parties. I'd rather go back to manual downloading of an JDK instead of using another tool.

ignasi35:
> What's the precedence for using sdkman as ...

Mostly consistence with other projects in the org.

jrudolph:
Let's try to merge this and see if it works.

Issue #3185, opened 2020-05-27
HTTP/2: support enforcing SETTINGS from our side


jrudolph:
Currently, we never send out any settings from our side. What we could do is expose (some of) the settings with configuration settings, announce, and enforce them.

 * Decide which settings to support
 * Create settings in our configuration
 * Announce the setting values in the first SETTINGS frame (currently empty in `Http2ServerDemux.preStart`)
 * Commit to these settings when the ack comes back and start enforcing them afterwards

ignasi35:
Regarding `Decide which settings to support`, I think we can start off with:

* `SETTINGS_HEADER_TABLE_SIZE`
* `SETTINGS_MAX_HEADER_LIST_SIZE`
* `SETTINGS_MAX_FRAME_SIZE`

And leave:

* `SETTINGS_MAX_CONCURRENT_STREAMS`
* `SETTINGS_ENABLE_PUSH`
* `SETTINGS_INITIAL_WINDOW_SIZE`

for future efforts.

jrudolph:
I'd say the most important one (for the server) is `SETTINGS_MAX_CONCURRENT_STREAMS` as it controls the effective concurrency on the server.

All the other ones are more fine-tuning settings.

ignasi35:
> I'd say the most important one (for the server) is `SETTINGS_MAX_CONCURRENT_STREAMS` as it controls the effective concurrency on the server.

I need to clarify my sorting criteria: I picked first the settngs that looked simpler to implement. 

Regarding `SETTINGS_MAX_CONCURRENT_STREAMS`: if I read the code correctly, there's already a `max-concurrent-streams` in the config to control the concurrency but I'm not entirly sure the purpose is exactly the same.

jrudolph:
> Regarding `SETTINGS_MAX_CONCURRENT_STREAMS`: if I read the code correctly, there's already a `max-concurrent-streams` in the config to control the concurrency but I'm not entirly sure the purpose is exactly the same.

I think these should be connected. Basically, we should announce the setting value to the peer at the start of the connection so the peer has to backpressure new substreams instead of having us pile up incoming substreams in front of the user-handler.

ignasi35:
@jrudolph I've raised #3522 for `SETTINGS_MAX_CONCURRENT_STREAMS`. We can raise other specific issues if there are extra details to discuss for a particular setting or use case.

ignasi35:
Status:

[...]

Issue #1051, opened 2017-04-25
Http2 Server Push API


jrudolph:

raboof:
See preliminary low-level API work in https://github.com/akka/akka-http/pull/785

PR #3540, opened 2020-10-15
Minor improvements in preparation for SETTING_ENABLE_PUSH

ignasi35:
References #3185

I started working on supporting SETTING_ENABLE_PUSH but then realized #1051 is still open.

Decided to push these few changes but I'm happy to close the PR.

### DIFF
--- a/akka-http-core/src/main/mima-filters/10.2.1.backwards.excludes/add-enable-push-http2-client-setting.excludes
+++ b/akka-http-core/src/main/mima-filters/10.2.1.backwards.excludes/add-enable-push-http2-client-setting.excludes
@@ -1,0 +1,2 @@
+ProblemFilters.exclude[Problem]("akka.http.scaladsl.settings.Http2ClientSettings#Http2ClientSettingsImpl.*")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.scaladsl.settings.Http2ClientSettings.enablePush")
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
@@ -6,7 +6,7 @@ package akka.http.impl.engine.http2
 
 import akka.annotation.InternalApi
 import akka.http.impl.engine.http2.Http2Protocol.ErrorCode
-import akka.http.impl.engine.http2.Http2Protocol.ErrorCode.FLOW_CONTROL_ERROR
+import akka.http.impl.engine.http2.Http2Protocol.ErrorCode._
 import akka.http.scaladsl.settings.Http2CommonSettings
 import akka.stream.Attributes
 import akka.stream.BidiShape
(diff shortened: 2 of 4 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Protocol.scala
```diff
@@ -98,8 +98,6 @@ private[http] object Http2Protocol {
    */
   final val NoStreamId = 0
 
-  final val PushPromiseEnabledDefault = true
```
ignasi35, line 101: This is dead code and seems like it can be replaced with `enable-push = on` in reference.conf.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -245,7 +245,17 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
...
+            // TODO: respect the peer settings
             debug(s"Setting max concurrent streams to $value (not enforced)")
+          case Setting(Http2Protocol.SettingIdentifier.SETTINGS_ENABLE_PUSH, value) =>
+            // TODO: respect the peer settings
+            value match {
+              case 0 => debug(s"Client disabled PUSH (ignored, PUSH not supported)")
+              case 1 => debug(s"Client enabled PUSH (ignored, PUSH not supported)")
+              case _ =>
+                pushGOAWAY(PROTOCOL_ERROR, s"Invalid value for SETTINGS_ENABLE_PUSH: $value")
+                settingsAppliedOk = false
```
ignasi35, line 257: This will introduce a change in behavior. The server is not a bit more strict.

If client sends valid values for the setting they won't notice.

akka-http-core/src/main/scala/akka/http/scaladsl/settings/Http2ServerSettings.scala
```diff
@@ -114,6 +114,9 @@ trait Http2ClientSettings extends /*FIXME: javadsl.settings.Http2ClientSettings
   def incomingStreamLevelBufferSize: Int
   def withIncomingStreamLevelBufferSize(newValue: Int): Http2ClientSettings = copy(incomingStreamLevelBufferSize = newValue)
 
+  def enablePush: Boolean
+  def withEnablePush(newValue: Boolean): Http2ClientSettings = copy(enablePush = newValue)
```
ignasi35, line 118: This is a client-only setting so not adding it to `Http2CommonSettings`.

Mental note: having peer-specific settings will require refactoring `class Http2ServerDemux(http2Settings: Http2CommonSettings, ... `

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
ignasi35: Hmm, I implemented this with a server point of view, but when this code runs on the client, the behavior should differ. I need to improve this...

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Protocol.scala
ignasi35: This is not dead code. 

A server may send PUSH_PROMISE frames before the client had time to send the SETTINGS.

akka-http-core/src/main/mima-filters/10.2.1.backwards.excludes/add-enable-push-http2-client-setting.excludes
```diff
@@ -0,0 +1,8 @@
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.scaladsl.settings.Http2ClientSettings#Http2ClientSettingsImpl.apply")
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.scaladsl.settings.Http2ClientSettings#Http2ClientSettingsImpl.copy")
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.scaladsl.settings.Http2ClientSettings#Http2ClientSettingsImpl.this")
+ProblemFilters.exclude[IncompatibleSignatureProblem]("akka.http.scaladsl.settings.Http2ClientSettings#Http2ClientSettingsImpl.unapply")
+ProblemFilters.exclude[IncompatibleResultTypeProblem]("akka.http.scaladsl.settings.Http2ClientSettings#Http2ClientSettingsImpl.copy$default$2")
+ProblemFilters.exclude[IncompatibleResultTypeProblem]("akka.http.scaladsl.settings.Http2ClientSettings#Http2ClientSettingsImpl.copy$default$6")
+ProblemFilters.exclude[IncompatibleResultTypeProblem]("akka.http.scaladsl.settings.Http2ClientSettings#Http2ClientSettingsImpl.copy$default$7")
```
jrudolph, line 7:
```suggestion
ProblemFilters.exclude[Problem]("akka.http.scaladsl.settings.Http2ClientSettings#Http2ClientSettingsImpl")
```

jrudolph (commented):
Not sure we should add a setting for now as long as we don't support server pushing either on the server or the client side. If we want to do something right now, we can disable server push on the client side in the initial settings push.

ignasi35:
> Not sure we should add a setting for now as long as we don't support server pushing either on the server or the client side. If we want to do something right now, we can disable server push on the client side in the initial settings push.

I'll close this in the meantime, then.

PR #3541, opened 2020-10-16
http2: client: test three consecutive GET requests

raboof:


akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
```diff
@@ -119,6 +121,47 @@ class Http2ClientSpec extends AkkaSpecWithMaterializer("""
         val (_, errorCode) = expectGOAWAY(1)
         errorCode should ===(ErrorCode.COMPRESSION_ERROR)
       }
+      "Three consecutive GET requests" in new SimpleRequestResponseRoundtripSetup {
```
ignasi35, line 124: What's so special about sending three consecutive GET requests?
I don't understand the purpose of the test :-(

jrudolph: Basically it tests that the examples from the HPACK spec work correctly. HPACK introduces a dependency between consecutive messages in one direction because of the compression (i.e. later requests can refer to previously sent headers to save bandwidth).

Maybe we can improve the test name to point that out more clearly.

jrudolph (approved):
LGTM

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/HPackSpecExamples.scala
```diff
@@ -192,7 +193,17 @@ object HPackSpecExamples {
         headers.Date.parseFromValueString("Mon, 21 Oct 2013 20:13:22 GMT").right.get,
         headers.Location("https://www.example.com"),
         headers.`Content-Encoding`(HttpEncodings.gzip),
-        // `Set-Cookie` modeled header doesn't support extra params like "version=1"
+        // The RFC example spells 'max-age' in lowercase, while the modeled `Set-Cookie` header renders it uppercase.
+        // Rendering it uppercase is consistent with https://tools.ietf.org/html/rfc6265#section-5.2.2 .
+        // https://tools.ietf.org/html/rfc7540#section-8.1.2 specifies HTTP/2 header fields ('Set-Cookie') must be
+        // converted to lowercase, but AFAICS has no such provision for header parameters.
```
jrudolph, line 199: Ugh, yeah, for some reason the RFC 6265 uses mixed-case parameter names and later specified that you need to match them case-insensitively...

https://tools.ietf.org/html/rfc6265#section-4.1.1

And yes, the HTTP/1.1 and HTTP/2 specs treat header values opaquely and refer to the particular spec for parsing details.

jrudolph: So that's indeed fine here, that we need to use a RawHeader in cases where we want a particular rendering on the wire (which is not related to HTTP/2 itself after all).

Thanks for adding the comment!

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
```diff
@@ -119,6 +121,47 @@ class Http2ClientSpec extends AkkaSpecWithMaterializer("""
...
+          ":authority" -> "www.example.com",
+          ":path" -> "/",
+          "content-length" -> "0"
+        )
+        requestResponseRoundtrip(
+          streamId = 1,
+          request = HttpRequest(GET, "https://www.example.com/"),
+          expectedHeaders = requestHeaders,
+          response = Seq(
+            HeadersFrame(streamId = 2, endStream = true, endHeaders = true, HPackSpecExamples.C61FirstResponseWithHuffman, None)
```
jrudolph, line 139: You also annotated it above but what's going on with the streamIds here?!?

jrudolph: It also works with the correct stream ids, so could you fix it and create a ticket about looking into why using the wrong ids doesn't break anything?

raboof: Ah, I guess these tests don't exercise any parts that associate the requests and responses. The HPACK state is also independent for the request and the response, so perhaps it's fine that it didn't cause any problems. Fixed.

jrudolph (approved):
LGTM

Issue #3542, opened 2020-10-16
Send and enforce SETTINGS_MAX_HEADER_LIST_SIZE


ignasi35:
Related to #778

A peer should read a user-defined value for `SETTINGS_MAX_HEADER_LIST_SIZE` and send it on the initial `SETTINGS` frame. Once SETTINGS_ACK is received, the peer shall enforce it. Note that `This advisory setting informs a peer...` so the enforcement is more relaxed than other settings. Also, the behavior when receiving an offending frame differs between server and client:

```
A server that receives a larger header block than it is willing to handle can send an HTTP 431 (Request Header Fields Too Large) status code [RFC6585]. A client can discard responses that it cannot process. The header block MUST be processed to ensure a consistent connection state, unless the connection is closed.
```

PR #3543, opened 2020-10-16
Send settings max header list size

ignasi35:
References #3542

Built on top of #3529 

I'm not sure I will implement the enforcing of the setting on this PR. It is an advisory setting _and_ the enforcement is too peer-specific.

### DIFF
--- a/akka-http-core/src/main/resources/reference.conf
+++ b/akka-http-core/src/main/resources/reference.conf
@@ -230,6 +230,12 @@ akka.http {
       # strict than required in the HTTP/2 Spec.
       max-concurrent-streams = 256
 
+      # This advisory setting informs a peer of the maximum size of header list that the sender is
+      # prepared to accept, in octets. The value is based on the uncompressed size of header fields,
+      # including the length of the name and value in octets plus an overhead of 32 octets for each
+      # header field.
+      max-header-list-size = 65536
+
       # The maximum number of bytes to receive from a request entity in a single chunk.
       #
       # The reasoning to limit that amount (instead of delivering all buffered data for a stream) is that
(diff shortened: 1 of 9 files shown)


jrudolph:
> I'm not sure I will implement the enforcing of the setting on this PR. It is an advisory setting _and_ the enforcement is too peer-specific.

Not quite sure what you mean? Sending out limits and then not enforcing it seems a bit useless to me? The place to enforce the limit would be in Http2FrameParsing. I guess we could assume that the demux actually sends out the setting and then start enforcing it as soon as the ack arrives.

jrudolph:
(I'm not saying we have to do that work right now, only that just sending out without enforcing it seems not worthwhile)

ignasi35:
> > I'm not sure I will implement the enforcing of the setting on this PR. It is an advisory setting and the enforcement is too peer-specific.

> Not quite sure what you mean? Sending out limits and then not enforcing it seems a bit useless to me? The place to enforce the limit would be in Http2FrameParsing. I guess we could assume that the demux actually sends out the setting and then start enforcing it as soon as the ack arrives.

From the [spec](https://httpwg.org/specs/rfc7540.html#rfc.section.6.5.2):

> SETTINGS_MAX_HEADER_LIST_SIZE (0x6):
> 
> This advisory setting informs a peer of the maximum size of header list that the sender is prepared to accept, in octets. The value is based on the uncompressed size of header fields, including the length of the name and value in octets plus an overhead of 32 octets for each header field.
> 
> For any given request, a lower limit than what is advertised MAY be enforced. The initial value of this setting is unlimited.

So the setting is less strict than others and enforcement is optional.

On top of that, the behavior of a peer when detecting a setting violation on an incoming frame is [peer-type dependant](https://httpwg.org/specs/rfc7540.html#rfc.section.10.5.1):

> A server that receives a larger header block than it is willing to handle can send an HTTP 431 (Request Header Fields Too Large) status code [RFC6585]. A client can discard responses that it cannot process. The header block MUST be processed to ensure a consistent connection state, unless the connection is closed.

Summing up, sending this header is only a best-effort, _hoping_ the peer will honor the value. 

@jrudolph, you're right that because of all the above it probably doesn't make sense to send the setting. I'll close.

PR #3552, opened 2020-10-20
HTTP/2 unified stream incoming / outgoing stream handling

jrudolph:
This is only a preview of how it might look like, this is a version that mostly works and passes all tests (but still has some known bugs).

I've got a few FIXMEs, I'd like to work continue on.

Not for review, yet.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Multiplexer.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Multiplexer.scala
@@ -5,17 +5,13 @@
 package akka.http.impl.engine.http2
 
 import akka.annotation.InternalApi
-import akka.http.scaladsl.model.HttpEntity
-import akka.stream.scaladsl.Sink
-
-import scala.collection.mutable
-import scala.collection.immutable
-import akka.stream.stage.{ GraphStageLogic, InHandler, OutHandler, StageLogging }
-import akka.util.ByteString
-import FrameEvent._
 import akka.event.LoggingAdapter
+import akka.http.impl.engine.http2.FrameEvent._
 import akka.http.scaladsl.settings.Http2CommonSettings
 import akka.macros.LogHelper
+import akka.stream.stage.{ GraphStageLogic, OutHandler, StageLogging }
+
+import scala.collection.immutable
 
 /**
  * INTERNAL API
@@ -25,19 +21,29 @@ import akka.macros.LogHelper
 @InternalApi
 private[http2] trait Http2Multiplexer {
   def pushControlFrame(frame: FrameEvent): Unit
-  def registerSubStream(sub: Http2SubStream): Unit
-
-  /** Notifies the multiplexer that the peer decided to cancel the substream */
-  def cancelSubStream(streamId: Int): Unit
-  def updateWindow(streamId: Int, increment: Int): Unit
+  def updateConnectionLevelWindow(increment: Int): Unit
   def updateMaxFrameSize(newMaxFrameSize: Int): Unit
   def updateDefaultWindow(newDefaultWindow: Int): Unit
   def updatePriority(priorityFrame: PriorityFrame): Unit
 
+  def enqueueOutStream(streamId: Int): Unit
+  def closeStream(streamId: Int): Unit
+
+  def currentInitialWindow: Int
+
   def reportTimings(): Unit
 
-  /** Called to cleanup any state when the connection is torn down */
-  def shutdown(): Unit
+  def maxBytesToBufferPerSubstream: Int
+}
+
+@InternalApi
+private[http2] sealed abstract class PullFrameResult {
+  def frame: DataFrame
+}
+@InternalApi
+private[http2] object PullFrameResult {
+  final case class SendFrame(frame: DataFrame, hasMore: Boolean) extends PullFrameResult
+  final case class SendFrameAndTrailer(frame: DataFrame, trailer: FrameEvent) extends PullFrameResult
 }
 
 /**
@@ -47,6 +53,14 @@ private[http2] trait Http2Multiplexer {
  * Accepts the streams from the Akka HTTP side and turns them into `FrameEvent`s
  * to be passed to the network side.
  *
+ * The main interface between stream handling and multiplexing is this:
+ *
+ * - stream handling can call `enqueueOutStream` if a stream wants to send a data frame
+ * - if there's connection window and the network pulls and there are no control frames to send, the multiplexer
+ *   state machine calls `sendDataFrame` here. This calls `pullNextFrame` to get the next frame and more info about
+ *   what the state of the stream is afterwards.
+ * - stream handling can call `closeStream` to remove a potentially enqueued stream
+ *
  * Mixed into the Http2ServerDemux graph logic.
  */
 @InternalApi
@@ -54,15 +68,14 @@ private[http2] trait Http2MultiplexerSupport { logic: GraphStageLogic with Stage
   def isServer: Boolean
   def settings: Http2CommonSettings
 
-  /**
-   * Signal an outgoing stream has ended, so when the incoming side is also finished it can be cleaned up.
-   */
-  def handleOutgoingEnded(streamId: Int): Unit
-
-  /**
-   * Allows suspending reading of incoming frames.
-   */
+  /** Allows suspending reading of incoming frames */
   def allowReadingIncomingFrames(allow: Boolean): Unit
+
+  /** Called by the multiplexer when ready to send a data frame */
+  def pullNextFrame(streamId: Int, maxSize: Int): PullFrameResult
+
+  /** Called by the multiplexer after SETTINGS_INITIAL_WINDOW_SIZE has changed */
+  def distributeWindowDeltaToAllStreams(delta: Int): Unit
 
   def createMultiplexer(outlet: GenericOutlet[FrameEvent], prioritizer: StreamPrioritizer): Http2Multiplexer =
     new Http2Multiplexer with OutHandler with StateTimingSupport with LogHelper { self =>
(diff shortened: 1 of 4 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -49,62 +51,145 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
...
+   * Server states:
+   *   * Idle -> OpenReceivingDataFirst: on receiving request HEADERS with endStream = false
+   *   * Idle -> HalfClosedRemoteWaitingForOutgoingStream: on receiving HEADERS with endStream = true
+   *   * OpenReceivingDataFirst -> HalfClosedRemoteWaitingForOutgoingStream: after receiving endStream
+   *   * OpenReceivingDataFirst -> Open: after user provided response before request was fully streamed in
+   *   * HalfClosedRemoteWaitingForOutgoingStream -> HalfClosedRemoteSendingData: we sent response HEADERS with endStream = false
+   *   * HalfClosedRemoteWaitingForOutgoingStream -> Closed: we sent response HEADERS with endStream = true
+   *   * HalfClosedRemoteSendingData -> Closed: we sent DATA with endStream = true
+   *   * Open -> HalfClosedRemoteSendingData: on receiving request data with endStream = true
+   *   * Open -> HalfClosedLocal: on receiving response data with endStream = true before request has been fully received (uncommon)
```
jrudolph, line 160: This transition is quite interesting. HTTP/2 explicitly allows bidirectional streams and also half-closed states at the end of the exchange. Especially, this weird case (also for client, see below), where the response has been sent over completely before the request has been fully received. Even that case is explicitly covered in https://httpwg.org/specs/rfc7540.html#rfc.section.8.1:

> A server can send a complete response prior to the client sending an entire request if the response does not depend on any portion of the request that has not been sent and received. When this is true, a server MAY request that the client abort transmission of a request without error by sending a RST_STREAM with an error code of NO_ERROR after sending a complete response (i.e., a frame with the END_STREAM flag). Clients MUST NOT discard responses as a result of receiving such a RST_STREAM, though clients can always discard responses at their discretion for other reasons.

Do you know if that is actively used in gRPC, @raboof?

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Multiplexer.scala
```diff
@@ -54,212 +60,86 @@ private[http2] trait Http2MultiplexerSupport { logic: GraphStageLogic with Stage
...
 
       private[http2] sealed trait MultiplexerState extends Product {
         def name: String = productPrefix
 
-        def onPull(): Unit
-        def pushControlFrame(frame: FrameEvent): Unit
-        def connectionWindowAvailable(): Unit
-        def enqueueOutStream(outStream: OutStream): Unit
-        def closeStream(outStream: OutStream): Unit
+        def onPull(): MultiplexerState
```
jrudolph, line 118: The one significant change in multiplexer is returning states instead of relying on side-effects for state changes.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Multiplexer.scala
```diff
@@ -54,212 +60,86 @@ private[http2] trait Http2MultiplexerSupport { logic: GraphStageLogic with Stage
...
+        def pushControlFrame(frame: FrameEvent): MultiplexerState
+        def connectionWindowAvailable(): MultiplexerState
+        def enqueueOutStream(streamId: Int): MultiplexerState
+        def closeStream(streamId: Int): MultiplexerState
+
+        protected def sendDataFrame(streamId: Int, sendableOutstreams: immutable.Set[Int]): MultiplexerState = {
```
jrudolph, line 124: The other is the greatly simplified handling of sending outgoing data and the interface with `OutStream` / stream handling. The full interface is now:

 * stream handling can call `enqueueOutStream` if a stream wants to send a data frame
 * if there's connection window and the network pulls and there are no control frames to send, the multiplexer state machine calls `sendDataFrame` here. This calls `pullNextFrame` to get the next frame and more info about what the state of the stream is afterwards.
 * stream handling can call `closeStream` to remove a potentially enqueued stream

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -38,89 +40,173 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
...
+
+  /** Called by Http2ServerDemux to let the state machine handle StreamFrameEvents */
+  def handleStreamEvent(e: StreamFrameEvent): Unit =
     updateState(e.streamId, _.handle(e))
-  }
 
+  /** Called by Http2ServerDemux when a stream comes in from the user-handler */
   def handleOutgoingCreated(stream: Http2SubStream): Unit =
-    updateState(stream.streamId, _.handleOutgoingCreated(stream))
+    if (streamFor(stream.streamId) != Closed) {
```
jrudolph, line 80: Since stream handling now has to deal with outgoing streams, we create those here and pass those to the state `handleOutgoingCreated`. There's another callback `handleOutgoingCreatedAndFinished` if the stream has no entity, so we can skip the logic for tracking sending entity data.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -38,89 +40,173 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
...
+   *   * HalfClosedLocalWaitingForPeerStream -> HalfClosedLocal: on receiving response HEADERS with endStream = false
+   *   * HalfClosedLocalWaitingForPeerStream -> Closed: on receiving response HEADERS with endStream = true
+   *   * HalfClosedLocal -> Closed: on receiving response data with endStream = true
+   *   * Open -> HalfClosedLocal: on sending out request data with endStream = true
+   *   * Open -> HalfClosedRemoteSendingData: on receiving response data with endStream = true before request has been fully sent out (uncommon)
+   *   * HalfClosedRemoteSendingData -> Closed: on sending out request data with endStream = true
    */
-  sealed abstract class IncomingStreamState { _: Product =>
-    def handle(event: StreamFrameEvent): IncomingStreamState
+  sealed abstract class StreamState { _: Product =>
```
jrudolph, line 176: New name: this is not only for incoming streams any more.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -38,89 +40,173 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
...
+    def handle(event: StreamFrameEvent): StreamState
 
     def stateName: String = productPrefix
-    def handleOutgoingCreated(stream: Http2SubStream): IncomingStreamState = {
+
+    /** Called when we receive a user-created stream (that is open for more data) */
+    def handleOutgoingCreated(outStream: OutStream, correlationAttributes: Map[AttributeKey[_], _]): StreamState = {
```
jrudolph, line 182: A few more callbacks added which are basically called dealing with outgoing streams or on events dispatched from stream callbacks (onNext, onUpstreamFinished, etc) from OutStream.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -139,32 +225,80 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
...
-    def handle(event: StreamFrameEvent): IncomingStreamState = expectIncomingStream(event, HalfClosedRemote, ReceivingDataFirst)
-    override def handleOutgoingCreated(stream: Http2SubStream): IncomingStreamState = SendingData(stream)
-    override def shutdown(): IncomingStreamState = Closed
+
+  case object Idle extends StreamState {
+    def handle(event: StreamFrameEvent): StreamState = expectIncomingStream(event, HalfClosedRemoteWaitingForOutgoingStream(0), OpenReceivingDataFirst(_, 0))
+    override def handleOutgoingCreated(outStream: OutStream, correlationAttributes: Map[AttributeKey[_], _]): StreamState = OpenSendingData(outStream, correlationAttributes)
+    override def handleOutgoingCreatedAndFinished(correlationAttributes: Map[AttributeKey[_], _]): StreamState = HalfClosedLocalWaitingForPeerStream(correlationAttributes)
+  }
+  case class OpenReceivingDataFirst(buffer: IncomingStreamBuffer, extraInitialWindow: Int = 0) extends ReceivingData(HalfClosedRemoteWaitingForOutgoingStream(extraInitialWindow)) {
```
jrudolph, line 239: In some states, we need to track `extraInitialWindow` because window tracking is usually done in OutStream but in some states we don't have an OutStream yet so we need to track it manually until the OutStream is created later.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -139,32 +225,80 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
...
+    override def handleOutgoingEnded(): StreamState = Closed
+
+    override protected def onReset(streamId: Int): Unit = multiplexer.closeStream(streamId)
+
+    override def incrementWindow(delta: Int): StreamState = copy(extraInitialWindow = extraInitialWindow + delta)
   }
-  case class ReceivingDataFirst(buffer: IncomingStreamBuffer) extends ReceivingData(HalfClosedRemote) {
-    override def handleOutgoingCreated(stream: Http2SubStream): IncomingStreamState = Open(buffer)
-    override def handleOutgoingEnded(): IncomingStreamState = Closed
+  trait Sending extends StreamState { _: Product =>
```
jrudolph, line 251: A common super trait to deal with the intricacies of handling `pullNextFrame`.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -139,32 +225,80 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
...
+    protected def outStream: OutStream
+
+    override def pullNextFrame(maxSize: Int): (StreamState, PullFrameResult) = {
+      val frame = outStream.nextFrame(maxSize)
```
jrudolph, line 255: Here the old and somewhat unintuitive `nextFrame` / `endStreamIfPossible` interface is still existent and converted into the nicer `PullFrameResult` one. It's now internal to stream handling so less critical to remove (but might be removed in another PR).

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -332,6 +498,132 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
...
+  object OutStream {
+    def apply(sub: Http2SubStream): OutStream = {
+      val subIn = new SubSinkInlet[Any](s"substream-in-${sub.streamId}")
+      val info = new OutStreamImpl(sub.streamId, None, multiplexer.currentInitialWindow)
+      info.registerIncomingData(subIn)
+      sub.data.runWith(subIn.sink)(subFusingMaterializer)
+      info
+    }
+  }
+  class OutStreamImpl(
```
jrudolph, line 518: Carried over as much unchanged as possible from Http2Multiplexer. We can clean this up later. The important bit is that we avoid side-effects crossing between stream handling and multiplexer in an uncontrolled way.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -838,6 +838,12 @@ class Http2ServerSpec extends AkkaSpecWithMaterializer("""
         // now again on stream 1
         sendDataAndExpectOnNet(entity1DataOut, 1, "jklm")
 
+        // send two data bits first but only pull and expect later
```
jrudolph, line 841: Another test case bit I added here after I noticed that this worked but broke during the refactoring.

jrudolph:
This is now ready for review. I added a few comments here for easier consumption. (Now needs another rebase after merging the maxConcurrentStreams PR).

jrudolph:
> (Now needs another rebase after merging the maxConcurrentStreams PR).

Done

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -142,40 +226,87 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
...
+      this
+    }
   }
-  case class SendingData(outgoingStream: Http2SubStream) extends IncomingStreamState {
-    override def handle(event: StreamFrameEvent): IncomingStreamState = expectIncomingStream(event, HalfClosedRemote, Open, _.withCorrelationAttributes(outgoingStream.correlationAttributes))
-    override def handleOutgoingEnded(): IncomingStreamState = SendingDataHalfClosedLocal(outgoingStream)
-    override def shutdown(): IncomingStreamState = Closed
+
+  case class OpenSendingData(outStream: OutStream, correlationAttributes: Map[AttributeKey[_], _]) extends StreamState with Sending {
+    override def handle(event: StreamFrameEvent): StreamState = event match {
```
jrudolph, line 296: `handle` doesn't work too well any more because often different traits contribute handling code that needs to be composed. In a future update, we could do the pattern matching in a central place and provide handler methods for each of those, similar to what we already do for `handleWindowUpdate` and `handleResetStream` with ad-hoc solutions.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
jrudolph: Created https://github.com/akka/akka-http/issues/3575 to track that work

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
jrudolph: Create low-prio follow up for this issue in https://github.com/akka/akka-http/issues/3576

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
raboof: I'm not sure if that's actively mentioned anywhere, no.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -38,92 +39,175 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
...
+   */
+  def lastStreamId(): Int = largestIncomingStreamId
+
+  private def streamFor(streamId: Int): StreamState =
+    streamStates.get(streamId) match {
       case Some(state) => state
       case None =>
         if (streamId <= largestIncomingStreamId) Closed // closed streams are never put into the map
         else if (isUpgraded && streamId == 1) {
+          require(isServer)
```
raboof, line 60: :+1:

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -38,92 +39,175 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
...
   def handleOutgoingCreated(stream: Http2SubStream): Unit =
-    updateState(stream.streamId, _.handleOutgoingCreated(stream))
+    if (streamFor(stream.streamId) != Closed) {
+      multiplexer.pushControlFrame(stream.initialHeaders)
+
+      if (stream.initialHeaders.endStream) {
+        updateState(stream.streamId, _.handleOutgoingCreatedAndFinished(stream.correlationAttributes))
+      } else {
+        val outStream = OutStream(stream)
+        stream.initialHeaders.priorityInfo.foreach(multiplexer.updatePriority)
```
raboof, line 86: Shouldn't we keep this information even for requests without an entity? so when other streams depend on this stream, we can correctly infer that they should depend on the stream that this stream depended on?

raboof: I haven't totally digested https://tools.ietf.org/html/rfc7540#section-5.3.4 yet, but noticed:

```
   It is possible for a stream to become closed while prioritization
   information that creates a dependency on that stream is in transit.
   If a stream identified in a dependency has no associated priority
   information, then the dependent stream is instead assigned a default
   priority (Section 5.3.5).  This potentially creates suboptimal
   prioritization, since the stream could be given a priority that is
   different from what is intended.

   To avoid these problems, an endpoint SHOULD retain stream
   prioritization state for a period after streams become closed
```

raboof (approved):
Nice, returning rather than side-effecting can indeed make things clearer. The extra comments are helpful and clear, naming improvements look good.

One question around prioritization.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
jrudolph: I pulled that statement out, so that the multiplexer gets to know about priority changes in all cases (even the Closed case).

Right now it doesn't matter because we don't have a prioritization strategy that uses any of the priority information.

ignasi35 (approved):
A couple of minor suggestions (I haven't reviewed all code yet).

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -38,92 +39,176 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
...
+
+  /** Called by multiplexer to distribute changes from INITIAL_WINDOW_SIZE to all streams */
+  def distributeWindowDeltaToAllStreams(delta: Int): Unit =
+    updateAllStates {
+      case s: Sending => s.increaseWindow(delta)
+      case x          => x
+    }
 
-  private def updateState(streamId: Int, handle: IncomingStreamState => IncomingStreamState): Unit = {
+  /** Called by the multiplexer if ready to send a data frame */
```
ignasi35, line 108: Ah! The comment was here! 😅

ignasi35 (approved):

Issue #3553, opened 2020-10-20
SETTINGS_INITIAL_WINDOW_SIZE is applied (and frames sent out) before SETTINGS ack is sent


jrudolph:

jrudolph:
That means that applying settings must be a multi-stage process:

 * first check if all settings can be applied
 * if false: fail with PROTOCOL_ERROR
 * if true: send ack first, then apply settings (which might send out other frames already relying on new settings)

ignasi35:
> That means that applying settings must be a multi-stage process:
> 

That plan is a bit different from the recommendation on the spec: 

```
The values in the SETTINGS frame MUST be processed in the order they 
appear, with no other frame processing between values. Unsupported 
parameters MUST be ignored. Once all values have been processed, the 
recipient MUST immediately emit a SETTINGS frame with the ACK flag set. 
Upon receiving a SETTINGS frame with the ACK flag set, the sender of the 
altered parameters can rely on the setting having been applied.

If the sender of a SETTINGS frame does not receive an acknowledgement 
within a reasonable amount of time, it MAY issue a connection error 
(Section 5.4.1) of type SETTINGS_TIMEOUT.
```

That recommendation is a bit unclear on whether to accumulate errors (if any) while processing the settings or send an error immediately (abruptly stopping the processing of other Settings). My assumption is that as soon as a setting triggers an error it is sent back to the peer immediately.

jrudolph:
The problem is that during processing of the settings we might already take further action (e.g. if `MAX_CONCURRENT_STREAM` was increased, we might want to send out new streams, if `INITIAL_WINDOW_SIZE` was increased when can send new data frames, etc) but we are only allowed to do so after sending out the ACK (which acknowledges all settings altogether). So, first of all we need to  send out the ACK. However, we should only do that if we know that we are able (or that it is very likely) that we can actually apply the setting afterwards.

> The values in the SETTINGS frame MUST be processed in the order they 
> appear, with no other frame processing between values.

That's something, we might have to think about: right now we process one setting after the other and this processing might have the side-effect of sending out new frames. However, if two settings would have an effect that would depend on the combination of those settings, according to this sentence we would have to hold back, first apply all settings and then check if as a result to all those applied settings we can do something else.

This change would be somewhat complicated, so I'd say we hold back on that part and add a comment to the settings processing that applying settings should be an atomic process and when adding or changing the process, it needs to be reviewed if the effect of applying all settings together atomically or each one by one would have the same effect.

[...]

PR #3554, opened 2020-10-20
Adding a few more tests about WINDOW_UPDATE edge cases

jrudolph:


akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -452,6 +453,66 @@ class Http2ServerSpec extends AkkaSpecWithMaterializer("""
...
+        sendWINDOW_UPDATE(0, missingWindow) // SETTINGS_INITIAL_WINDOW_SIZE only has effect on stream-level window, so need to make room in connection
+
+        entityDataOut.sendNext(bytes(bytesToSend, 0x23))
+        expectDATA(TheStreamId, false, Http2Protocol.InitialWindowSize)
+        expectNoBytes(100.millis)
+
+        sendSETTING(Http2Protocol.SettingIdentifier.SETTINGS_INITIAL_WINDOW_SIZE, bytesToSend)
+        updateFromServerWindows(TheStreamId, _ + missingWindow) // test probe doesn't automatically update window
+        expectDATA(TheStreamId, false, missingWindow)
+        expectSettingsAck() // FIXME: bug: we must send ACK before making use of the new setting, see https://github.com/akka/akka-http/issues/3553
```
jrudolph, line 510: @ignasi35 this test found the issue around applying settings we discussed in #3553.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -452,6 +453,66 @@ class Http2ServerSpec extends AkkaSpecWithMaterializer("""
         toNet.expectNoBytes(100.millis) // the whole stage failed with bug #2236
       }
 
+      "handle WINDOW_UPDATE correctly when received before started sending out response and while receiving request data" in new WaitingForRequestData {
+        val bytesToSend = 70000 // > Http2Protocol.InitialWindowSize
+        val missingWindow = bytesToSend - Http2Protocol.InitialWindowSize
+        require(missingWindow >= 0)
+        sendWINDOW_UPDATE(0, missingWindow)
```
raboof, line 460: So this extends the window size with `missingWindow` bytes beyond the `InitialWindowSize`. Makes sense, but perhaps a comment might make this easier to follow ;)

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -452,6 +453,66 @@ class Http2ServerSpec extends AkkaSpecWithMaterializer("""
         toNet.expectNoBytes(100.millis) // the whole stage failed with bug #2236
       }
 
+      "handle WINDOW_UPDATE correctly when received before started sending out response and while receiving request data" in new WaitingForRequestData {
```
raboof, line 456: Does the request never finish? OK I guess :)

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -452,6 +453,66 @@ class Http2ServerSpec extends AkkaSpecWithMaterializer("""
...
+        expectDecodedResponseHEADERS(streamId = TheStreamId, endStream = false) shouldBe response.withEntity(HttpEntity.Empty.withContentType(ContentTypes.`application/octet-stream`))
+        entityDataOut.sendNext(bytes(bytesToSend, 0x23))
+
+        expectDATA(TheStreamId, false, bytesToSend)
+
+        entityDataOut.sendComplete()
+        expectDATA(TheStreamId, true, 0)
+      }
+
+      "handle WINDOW_UPDATE correctly when received before started sending out response" in new WaitingForResponseSetup {
```
raboof, line 476: Very similar to the previous one but now with the request already finished, seems OK

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
raboof: So when this bug is fixed we should see the Ack before we see the DATA? Make sense.

raboof (approved):

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
ignasi35: > perhaps a comment might make this easier to follow

I came here just to write that :-)

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
ignasi35: I'm not sure I'm familiar enought with the issue. In cases when the window is exhausted, no frames should be produced and buffered until (1) a `WINDOW_UPDATE` is received, or (2) a `SETTINGS` is received and the corresponding `SETTING_ACK` is sent out.

When #3553 is fixed, I agree that `SETTINGS_ACK` should arrive before the SETTINGS ACK.

The problem is, when the server window on the client side is exhausted, the client should not pre-render any frames until a WINDOW_UPDATE (or a SETTINGS_INITIAL_WINDOW_SIZE) is received since there are no guarantees a pre-rendered frame would fit on the window, right? The same happens with SETTINGS_FRAME_SIZE.

------

Hmm, increasing the window size via settings is very different from the processing of a `WINDOW_UPDATE`. 

In `WINDOW_UPDATE`, the sender increases the window value and then sends the frame out so the remote peer's value for the window is always less or equal to the sender's value.
Using `SETTINGS_INITIAL_WINDOW_SIZE` means the sender doesn't increase the window value until it gets an `SETTINGS_ACK` but the remote peer already uses the increased window value so the remote peer's value for the window may be greater than the sender's window value.

It sounds like `SETTINGS_INITIAL_WINDOW_SIZE` as a broadcast replacement of `WINDOW_UPDATE` is a hack.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -452,6 +453,66 @@ class Http2ServerSpec extends AkkaSpecWithMaterializer("""
...
+        require(missingWindow >= 0)
+        sendWINDOW_UPDATE(0, missingWindow)
+        sendWINDOW_UPDATE(TheStreamId, missingWindow)
+
+        val entityDataOut = TestPublisher.probe[ByteString]()
+
+        val response = HttpResponse(entity = HttpEntity(ContentTypes.`application/octet-stream`, Source.fromPublisher(entityDataOut)))
+        emitResponse(TheStreamId, response)
+        expectDecodedResponseHEADERS(streamId = TheStreamId, endStream = false) shouldBe response.withEntity(HttpEntity.Empty.withContentType(ContentTypes.`application/octet-stream`))
+        entityDataOut.sendNext(bytes(bytesToSend, 0x23))
```
ignasi35, line 468: How does this work? 
First, the server emits a response, and then it sends `bytesToSend` bytes? Doesn't that exceed the window size of `bytesToSend` octets?

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
ignasi35: I need to read more about `WaitingForResponseSetup` and `WaitingForRequestData` to see the difference between these two tests.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
jrudolph: > The problem is, when the server window on the client side is exhausted, the client should not pre-render any frames until a WINDOW_UPDATE (or a SETTINGS_INITIAL_WINDOW_SIZE) is received since there are no guarantees a pre-rendered frame would fit on the window, right? The same happens with SETTINGS_FRAME_SIZE.

Yes.

> In `WINDOW_UPDATE`, the sender increases the window value and then sends the frame out so the remote peer's value for the window is always less or equal to the sender's value.
> Using `SETTINGS_INITIAL_WINDOW_SIZE` means the sender doesn't increase the window value until it gets an `SETTINGS_ACK` but the remote peer already uses the increased window value so the remote peer's value for the window may be greater than the sender's window value.

For the right value of sender (i.e. the one that proposed the setting), that seems correct as well.

> It sounds like `SETTINGS_INITIAL_WINDOW_SIZE` as a broadcast replacement of `WINDOW_UPDATE` is a hack.

Agreed, that's not what it is supposed to be used for. I guess it was really originally thought to be used for setting the initial window size for new streams. Maybe it was changed this way because the sender can just ignore the SETTINGS ACK and assume the setting to be applied instantly and retroactively, so that it doesn't matter for the client if a received stream was created slightly before the corresponding ACK or slightly after.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -452,6 +453,66 @@ class Http2ServerSpec extends AkkaSpecWithMaterializer("""
...
+        val missingWindow = bytesToSend - Http2Protocol.InitialWindowSize
+        require(missingWindow >= 0)
+        sendWINDOW_UPDATE(0, missingWindow) // SETTINGS_INITIAL_WINDOW_SIZE only has effect on stream-level window, so need to make room in connection
+
+        entityDataOut.sendNext(bytes(bytesToSend, 0x23))
+        expectDATA(TheStreamId, false, Http2Protocol.InitialWindowSize)
+        expectNoBytes(100.millis)
+
+        sendSETTING(Http2Protocol.SettingIdentifier.SETTINGS_INITIAL_WINDOW_SIZE, bytesToSend)
+        updateFromServerWindows(TheStreamId, _ + missingWindow) // test probe doesn't automatically update window
```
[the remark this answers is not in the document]
jrudolph: No, I think that doesn't work because we consumed the initial part of the window already.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
jrudolph: The user level interface uses a simple `Source[ByteString]` which can send any amount of data (I talked about that in our call this morning). The internal logic for outgoing streams is this:

 * frames are pulled one-by-one by the lower layers from the Http2ServerDemux
 * we keep a buffer of control frames to send (control frames are not subject to flow-control / windows) and a list of streams that can send data
 * if the multiplexer is pulled, first the control frames are drained and only if there are no pending control frames, one ready stream is chosen to send the next data frame
 * for each stream we have an outgoing buffer and if there's pending data the stream is listed to be chosen to send a frame
 * into that buffer, we pull from the user until we have at least `2 * currentMaxFrameSize` number of bytes buffered

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
jrudolph: Yep, two different states in the stream state machine..

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
jrudolph: Done

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
jrudolph: So, in case that wasn't clear, here we use the user-level interface to send a ByteString bigger than the WINDOW_SIZE and expect the rest to be buffered and only flushed later when more WINDOW is available.

raboof: `entityDataOut` is a probe stream that is attached to the body of the `HttpResponse`. Only DATA frames count to the window size (https://tools.ietf.org/html/rfc7540#section-5.2.1 point 5), so this fits exactly

PR #3555, opened 2020-10-20
http2: disable Push (send SETTING)

ignasi35:
References #3185 (https://github.com/akka/akka-http/pull/3540#pullrequestreview-511554303)

Built on top of #3522

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
@@ -103,10 +103,18 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
 
       val multiplexer = createMultiplexer(frameOut, StreamPrioritizer.first())
 
-      // Send settings initially based on our configuration. For simplicity, these settings are
+      // Send initial settings based on the local application.conf. For simplicity, these settings are
       // enforced immediately even before the acknowledgement is received.
       // Reminder: the receiver of a SETTINGS frame must process them in the order they are received.
-      private def initialLocalSettings = immutable.Seq(Setting(SettingIdentifier.SETTINGS_MAX_CONCURRENT_STREAMS, http2Settings.maxConcurrentStreams))
+      val initialLocalSettings: Seq[Setting] = immutable.Seq(
+        Setting(SettingIdentifier.SETTINGS_MAX_CONCURRENT_STREAMS, http2Settings.maxConcurrentStreams)
+      ) :++ {
+          if (isServer) {
+            Seq.empty[Setting]
+          } else {
+            immutable.Seq(Setting(SettingIdentifier.SETTINGS_ENABLE_PUSH, 0))
+          }
+        }
 
       override def preStart(): Unit = {
         if (initialRemoteSettings.nonEmpty) {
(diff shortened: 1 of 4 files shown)


akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
```diff
@@ -32,7 +33,7 @@ import org.scalatest.concurrent.Eventually
  */
 class Http2ClientSpec extends AkkaSpecWithMaterializer("""
     akka.http.server.remote-address-header = on
-    akka.http.server.http2.log-frames = on
+    akka.http.client.http2.log-frames = on
   """)
```
ignasi35, line 37: I've enabled logging of frames on the client. Not sure about `server.remote-address-header`, though.

ignasi35:
Rebased this on `master`

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -103,10 +103,18 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
...
-      private def initialLocalSettings = immutable.Seq(Setting(SettingIdentifier.SETTINGS_MAX_CONCURRENT_STREAMS, http2Settings.maxConcurrentStreams))
+      val initialLocalSettings: Seq[Setting] = immutable.Seq(
+        Setting(SettingIdentifier.SETTINGS_MAX_CONCURRENT_STREAMS, http2Settings.maxConcurrentStreams)
+      ) :++ {
+          if (isServer) {
+            Seq.empty[Setting]
+          } else {
+            immutable.Seq(Setting(SettingIdentifier.SETTINGS_ENABLE_PUSH, 0))
+          }
+        }
```
jrudolph, line 117:
```suggestion
      ) ++
      Seq(Setting(SettingIdentifier.SETTINGS_ENABLE_PUSH, 0)).filter(_ => !isServer) // only on client
```

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
jrudolph: Can be removed I guess.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
```diff
@@ -158,6 +162,22 @@ class Http2ClientSpec extends AkkaSpecWithMaterializer("""
...
         )
       }
+
+    }
+
+    "send settings" should {
+      abstract class SettingsSetup extends TestSetupWithoutHandshake with NetProbes with Http2FrameSending {
+        def expectSetting(expected: Setting): Unit = {
+          toNet.expectBytes(Http2Protocol.ClientConnectionPreface)
+          val event: FrameEvent = expectFrame()
```
jrudolph, line 172: You added an `expectSETTINGS` which you can use here (it can also be changed to return a `SettingsFrame` directly).

jrudolph (approved):
LGTM with a few cosmetic suggestions.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
ignasi35: Good call. I was cherrypicking these changes from an old branch without that method.

ignasi35:
Note this PR sends the settings from the client to disable PUSH_PROMISE, but doesn't enforce it.

We're effectively enforcing `ENABLE_PUSH=0` because we don't support it but we may not be responding according to the Spec when an unexpected PUSH_PROMISE is received.

It's an unlikely scenario so I don't think we should do it now.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
```diff
@@ -158,6 +162,20 @@ class Http2ClientSpec extends AkkaSpecWithMaterializer("""
             .withEntity(Strict(ContentTypes.NoContentType, ByteString.empty))
         )
       }
+
+    }
+
+    "send settings" should {
+      abstract class SettingsSetup extends TestSetupWithoutHandshake with NetProbes with Http2FrameSending {
+        def expectSetting(expected: Setting): Unit = {
+          toNet.expectBytes(Http2Protocol.ClientConnectionPreface)
```
raboof, line 171: I would be OK with having those directly in the test rather than introducing `SettingsSetup`, but :+1:

raboof (approved):

jrudolph:
Sorry, don't know why I merged despite the test failures...

jrudolph:
https://github.com/akka/akka-http/pull/3582

PR #3559, opened 2020-10-21
bench: fix http2 compilation

jrudolph:
Benchmark compilation silently failed since the move of http2 into akka-http-core.

Added email notifications for akka-http-bench-nightly failures.

PR #3561, opened 2020-10-21
build: also aggregate akka-http-bench-jmh to fail if benchs fail to compile

jrudolph:


jrudolph:
Should fail until based on #3559

jrudolph:
Ok, didn't fail because I already merged the fixed :roll_eyes:

raboof (approved):
weird that they weren't there before :+1:

jrudolph:
The first benchmark and the module was contributed, so maybe we just missed it in the review.

Issue #3566, opened 2020-10-23
Add parameter name String to ParamsSpec field


umbreak:
The trait [ParamsSpec](https://github.com/akka/akka-http/blob/master/akka-http/src/main/scala/akka/http/scaladsl/server/directives/ParameterDirectives.scala#L73) could also contain the following:
```scala
def name: String
```
being `name` the requested query parameter field name. This would be very useful for us because we want to override the parameter directive adding some extra action (adding an attribute to the `HttpRequest`):

```scala
  private val atKey: AttributeKey[Set[String]] = AttributeKey("allowed-query-params")

  private def trackQueryParam(name: String): Directive0 =
    optionalAttribute(atKey).flatMap { paramsOpt =>
      mapRequest(addAttribute(atKey, paramsOpt.fold(Set(name))(_ + name)))
    }

def param(pdef1: ParamSpec): Directive[Tuple1[pdef1.Out]] =
    Directives.parameter(pdef1.name) & trackQueryParam(pdef1.name)
```

Since the field name is not currently accessible from `ParamSpec` we have to do some ugly work around to make that work currently.

jrudolph:
Thanks for the suggestion, @umbreak. Even with that addition, wouldn't it still require for you to provide all the overloads of parameters for different numbers of arguments (which we provide via code generation)?

For that reason, my concern is that this feature would not really be generally applicable.

Maybe we can rather discuss you use case on a higher-level to see if it can be provided in some way?

umbreak:
I don't think it is a problem that `ParamsSpec` has the information about which is the query parameter name, since all the calls to `ParamsSpec` must have it.

I'm not interested into supporting all the overrides, since I'm not interested in using `parameters` Directive.

My use case, as you can see in the example is to track the query parameters defined for a certain Route.
We keep track of them using an `AttributeKey` where we store the query parameters' name. And tracking them can be useful to allow this directive:

```scala
  val rejectInvalidParams: Directive0 =
    (parameterMap & optionalAttribute(atKey)).tflatMap { case (allParams, allowedParams) =>
      val invalid = allParams.keySet -- allowedParams.getOrElse(Set.empty[String])
      if (invalid.isEmpty) pass
      else reject(validationRejection(s"query parameters not valid: '${invalid.mkString(", ")}'"))
    }
```

This means that for every Route DSL I have I can make sure that no query parameters, besides the ones I've explicitly defined in the route DSL (using `parameter` directive) are present. And If they are, I reject.

[...]

PR #3573, opened 2020-10-27
Added unusedParameters to RequestContext

umbreak:
Fixes https://github.com/akka/akka-http/issues/3566

It is probably better than the solution I pushed on the PR https://github.com/akka/akka-http/pull/3567

The usefulness of adding this field is the ability to easily create the following Directive:

```scala
  val rejectOnUnusedParameters: Directive0 =
   extractUnusedParameters flatMap { 
      case unused if unused.isEmpty => pass
      case unused => reject(validationRejection(s"query parameters not valid: '${invalid.mkString(", ")}'"))
    }
```

### DIFF
--- a/akka-http/src/main/scala/akka/http/scaladsl/server/RequestContextImpl.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/server/RequestContextImpl.scala
@@ -23,6 +23,7 @@ import scala.concurrent.{ ExecutionContextExecutor, Future }
 private[http] class RequestContextImpl(
   val request:          HttpRequest,
   val unmatchedPath:    Uri.Path,
+  val unusedParameters: Set[String],
   val executionContext: ExecutionContextExecutor,
   val materializer:     Materializer,
   val log:              LoggingAdapter,
@@ -30,7 +31,7 @@ private[http] class RequestContextImpl(
   val parserSettings:   ParserSettings) extends RequestContext {
 
   def this(request: HttpRequest, log: LoggingAdapter, settings: RoutingSettings, parserSettings: ParserSettings)(implicit ec: ExecutionContextExecutor, materializer: Materializer) =
-    this(request, request.uri.path, ec, materializer, log, settings, parserSettings)
+    this(request, request.uri.path, request.uri.query().toMap.keySet, ec, materializer, log, settings, parserSettings)
 
   def reconfigure(executionContext: ExecutionContextExecutor, materializer: Materializer, log: LoggingAdapter, settings: RoutingSettings): RequestContext =
     copy(executionContext = executionContext, materializer = materializer, log = log, routingSettings = settings)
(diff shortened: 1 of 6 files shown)


umbreak:
All those "binary incompatibility" failures are due to the introduction of the following methods to `RequestContext`
- unusedParameters
- withUnusedParameters
- mapUnusedParameters

akka-http/src/main/scala/akka/http/scaladsl/server/RequestContextImpl.scala
```diff
@@ -23,14 +23,15 @@ import scala.concurrent.{ ExecutionContextExecutor, Future }
...
+  val unusedParameters: Set[String],
   val executionContext: ExecutionContextExecutor,
   val materializer:     Materializer,
   val log:              LoggingAdapter,
   val settings:         RoutingSettings,
   val parserSettings:   ParserSettings) extends RequestContext {
 
   def this(request: HttpRequest, log: LoggingAdapter, settings: RoutingSettings, parserSettings: ParserSettings)(implicit ec: ExecutionContextExecutor, materializer: Materializer) =
-    this(request, request.uri.path, ec, materializer, log, settings, parserSettings)
+    this(request, request.uri.path, request.uri.query().toMap.keySet, ec, materializer, log, settings, parserSettings)
```
raboof, line 34: I'm not sure we would want to do this here: we have made the parsing of the query string as a set of name-value parameters lazy (they are not parsed until you actually use them as name-value pairs), because it is not always needed (so postponing parsing them can be good for performance), and may even fail: there are query strings that are allowed by the URI spec, but cannot be parsed into name-value pairs.

raboof (commented):
To be honest I'm not quite sold on how useful such a `rejectOnUnusedParameters` directive is. From the issue at #3566, it looks like you'd want to use it during testing? What would that actually test? Wouldn't it be better to test that those parameters actually have some particular effect?

umbreak:
This functionality allows to provide feedback to the client (via this rejectOnUnusedParameters) when it provides query params  to the API that are not being used.

I feel that an API that silently ignores unused query parameters it misleads the client, since they don't know if the query parameter is being used or not in a certain endpoint. This [StackOverflow thread](https://stackoverflow.com/questions/15947074/extra-query-parameters-in-the-rest-api-url) talks about the same concept.

akka-http/src/main/scala/akka/http/scaladsl/server/RequestContextImpl.scala
umbreak: What is your suggestion?

umbreak: In the other hand, every time a call to `parameter(...)` or `parameters(...)` the following method is executed
```scala
Try(request.uri.query())
```

which does not seem very efficient either. Maybe just getting all the query parameters in the `RequestContext` (lazy if you want) and then having a set with the currently visited would help with this.

PR #3577, opened 2020-10-27
build: run mima on travis only for master builds or releases

jrudolph:
Instead, PR validation on Jenkins was changed to run mima for 2.12 as well

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,6 +50,7 @@ jobs:
 
 stages:
   - name: mima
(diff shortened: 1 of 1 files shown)


jrudolph:
This seems to add ~90-100 seconds to the Jenkins PR validation which is quite a lot but better than what Jenkins has to offer, probably even without factoring in the excessive queuing time.

raboof (approved):

ignasi35 (approved):

Issue #3228, opened 2020-06-09
HTTP/2 client: backpressure requests when server (peer) announced max-concurrent-streams is reached / Respect peer SETTINGS_MAX_CONCURRENT_STREAMS


jrudolph:
The default at connection start is an unlimited number of concurrent streams but both sides can limit the number of concurrent streams that the peer is allowed to create (though, only the client will create substreams aside from server push).

When the server announces that it will only accept a limited number of requests we need to keep track of the ongoing open streams and prevent new streams from being opened by sending new requests.

The rule is that a new setting received from a peer has to be respected only after having sent the corresponding SETTING ACK. As settings come in concurrently with sending requests out, we need to make sure the control frames for SETTING ACKS for new limits and opening new streams (sending requests) keep being ordered on the wire. This is a potential concern because control frames for opening streams that might have been opened under the old limit might still be buffered on our side and we might be tempted to reorder control frames in a way that the SETTING ACK might take precedence over HEADERS frames. We shouldn't do this.

So here are a few tasks to accomplish:

 * keep track of the number of open streams (if we expect to support server push at some point, we need to track this number in both directions)
 * after having submitted a request, only pull in a new request when there are spare slots (i.e. open streams < limit)
 * when a stream is closed, check if a slot is free and only then pull or grab next request (we might have pulled a new request before we knew that we are not allowed to push it if limit is decreased)
 * when limit is changed, check if we now can pull a new request

jrudolph:
@ignasi, assigned you to this instead of #774 because this one has more information.

PR #3578, opened 2020-10-27
Respect remote settings max concurrent streams (#3228)

ignasi35:
(I need to rebase, keeping this DRAFT for early reviewers)
References #3228

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
@@ -113,7 +113,7 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
         // FIXME: handle the connection closing according to the specification
       }
       private[this] var allowReadingIncomingFrames: Boolean = true
-      override def allowReadingIncomingFrames(allow: Boolean): Unit = {
+      override def      allowReadingIncomingFrames(allow: Boolean): Unit = {
         if (allow != allowReadingIncomingFrames)
           if (allow) {
             debug("Resume reading incoming frames")
@@ -124,6 +124,18 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
       }
       def pullFrameIn(): Unit = if (allowReadingIncomingFrames && !hasBeenPulled(frameIn)) pull(frameIn)
 
+      private[this] var allowWritingOutgoingFrames: Boolean = true
+      override def allowWritingOutgoingFrames(allow: Boolean): Unit = {
+        if (allow != allowWritingOutgoingFrames)
+          if (allow) {
+            debug("Resume writing outgoing frames")
+            if (!hasBeenPulled(substreamIn)) pull(substreamIn)
+          } else debug("Suspended writing outgoing frames") // can't retract pending pull but that's ok
+
+        allowReadingIncomingFrames = allow
+      }
+      def pullsubstreamIn(): Unit = if (allowWritingOutgoingFrames && !hasBeenPulled(substreamIn)) pull(substreamIn)
+
       setHandler(frameIn, new InHandler {
 
         def onPush(): Unit = {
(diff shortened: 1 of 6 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -38,25 +39,55 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
...
+    //  only account for open and halfclosed streams
+    // FIXME: when supporting PUSH_PROMISE this operation should only consider streams
+    //  initiated in this peer
+    def peerHasCapacity: Boolean = {
+      // FIXME: this off-by-one is required because we're evaluating
+      //  capacity after generating demand
+      size < getRemoteMaxConcurrentStreams - 1
+    }
+
+    def getRemoteMaxConcurrentStreams: Int = _maxConcurrentStreams
```
jrudolph, line 53: Even with PUSH_PROMISE there are only two groups: incoming and outgoing streams which need different behavior. For incoming streams we need to enforce our stream limit, for outgoing streams we need to respect the peer limit. So, it seems packing that logic in here, is not necessary.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -212,10 +229,13 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
 
       setHandler(substreamIn, new InHandler {
         def onPush(): Unit = {
-          val sub = grab(substreamIn)
-          pull(substreamIn)
+          val sub: Http2SubStream = grab(substreamIn)
```
jrudolph, line 232: The logic to pull must go somewhere else:

 * on the server side, a substream received here, will not open a new stream (but actually might in the future if we implement PUSH_PROMISE by letting responses "suggest" PUSH_PROMISES)
 * the stage has lots of entrypoints because of all the ports and substream handlers, etc. E.g. an action might be required if we read the last frame of a response in `IncomingStreamBuffer.onUpstreamFinish`.

I think we should add an action to the stream handling `updateState` which (on the client) tries to pull if there's capacity for outgoing capacity.

jrudolph: (This also one reason we need the refactoring, because it is much more principled about going through the state machine at all times)

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
```diff
@@ -193,20 +241,20 @@ class Http2ClientSpec extends AkkaSpecWithMaterializer("""
   }
 
   /** Basic TestSetup that has already passed the exchange of the connection preface */
-  abstract class TestSetup extends TestSetupWithoutHandshake with NetProbes with Http2FrameSending {
+  abstract class TestSetup(serverSettings: Setting*) extends TestSetupWithoutHandshake with NetProbes with Http2FrameSending {
```
jrudolph, line 244:
```suggestion
  abstract class TestSetup(initialServerSettings: Setting*) extends TestSetupWithoutHandshake with NetProbes with Http2FrameSending {
```

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
```diff
@@ -159,6 +165,48 @@ class Http2ClientSpec extends AkkaSpecWithMaterializer("""
         )
       }
     }
+
+    "respect settings" should {
+      "received SETTINGS_MAX_CONCURRENT_STREAMS should limit the number of outgoing streams" in new TestSetup(
```
jrudolph, line 170: We need another test when the limit is changed later. Especially, if the current number of streams is then already above the limit.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
```diff
@@ -159,6 +165,48 @@ class Http2ClientSpec extends AkkaSpecWithMaterializer("""
...
+        Setting(SettingIdentifier.SETTINGS_MAX_CONCURRENT_STREAMS, 3)
+      ) with NetProbes {
+        val request = HttpRequest(uri = "https://www.example.com/")
+        // server set SETTINGS_MAX_CONCURRENT_STREAMS=5 so an attempt from the client to open 6 streams
+        // should only produce 5 frames
+        emitRequest(1, request)
+        emitRequest(3, request)
+        emitRequest(5, request)
+        emitRequest(7, request) // this is backpressured
```
jrudolph, line 179: Shouldn't this fail then already? Or do the probes introduce a bit of buffering?

jrudolph (commented):
Test looks good, the implementation will have to change, see below

ignasi35:
With #3552 merged, I think I'd rather redo this from scratch.

ignasi35:
Redone in #3581

PR #3579, opened 2020-10-28
bench: avoid getting stuck in H2ServerProcessingBenchmark

jrudolph:
The nightly bench run was stuck because of this issue: https://jenkins.akka.io:8498/view/Akka-http/job/akka-http-bench-nightly/219/

### DIFF
--- a/akka-http-bench-jmh/src/main/scala/akka/http/impl/engine/http2/H2ServerProcessingBenchmark.scala
+++ b/akka-http-bench-jmh/src/main/scala/akka/http/impl/engine/http2/H2ServerProcessingBenchmark.scala
@@ -4,7 +4,7 @@
 
 package akka.http.impl.engine.http2
 
-import java.util.concurrent.CountDownLatch
+import java.util.concurrent.{ CountDownLatch, TimeUnit }
 
 import scala.concurrent.Await
 import scala.concurrent.Future
@@ -35,10 +35,11 @@ class H2ServerProcessingBenchmark extends CommonBenchmark {
 
   val packedResponse = ByteString(-62, -63, -64, -65, -66)
 
+  val numRequests = 10000
+
   @Benchmark
-  @OperationsPerInvocation(10000)
+  @OperationsPerInvocation(10000) // should be same as numRequest
   def benchRequestProcessing(): Unit = {
-    val numRequests = 10000
     val latch = new CountDownLatch(numRequests)
 
     val requests =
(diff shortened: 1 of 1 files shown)


raboof (approved):

akka-http-bench-jmh/src/main/scala/akka/http/impl/engine/http2/H2ServerProcessingBenchmark.scala
```diff
@@ -71,6 +71,7 @@ class H2ServerProcessingBenchmark extends CommonBenchmark {
            #akka.loglevel = debug
            akka.actor.default-dispatcher.fork-join-executor.parallelism-max = 1
            #akka.http.server.log-unencrypted-network-bytes = 100
+           akka.http.server.http2.max-concurrent-streams = 50000 # needs to be higher than the above `numRequests`
```
raboof, line 74: I guess we could use something like `${numRequests * 5}` to make this always true?

jrudolph:
Though, this is only a quick fix, it's not quite clear, why it would get stuck instead of failing outright.

akka-http-bench-jmh/src/main/scala/akka/http/impl/engine/http2/H2ServerProcessingBenchmark.scala
jrudolph: I think even `numRequests` is enough.

jrudolph:
> Though, this is only a quick fix, it's not quite clear, why it would get stuck instead of failing outright.

Seems it's the right fix after all: The benchmark creates one big ByteString with all the requests in one go, so if the number of concurrent requests is too low, the infra will now reject some requests which our bench infra doesn't count, so we'll get stuck on the latch.

akka-http-bench-jmh/src/main/scala/akka/http/impl/engine/http2/H2ServerProcessingBenchmark.scala
```diff
@@ -57,21 +58,22 @@ class H2ServerProcessingBenchmark extends CommonBenchmark {
         }))(Keep.both)
         .run()
 
-    latch.await()
+    require(latch.await(10, TimeUnit.SECONDS), "Not all responses were received in time")
```
jrudolph, line 61: Also added the timeout here to prevent getting stuck

akka-http-bench-jmh/src/main/scala/akka/http/impl/engine/http2/H2ServerProcessingBenchmark.scala
jrudolph: Done

PR #3581, opened 2020-10-29
Respect remote max concurrent streams (reprise)

ignasi35:
References #3228

PR #3584, opened 2020-10-29
http2: also go through state machine for `resetStream`

jrudolph:
To avoid uncontrolled reentry into state machine. Will conflict with #3581.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
@@ -130,8 +130,7 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper 
   /** Called to cleanup any state when the connection is torn down */
   def shutdownStreamHandling(): Unit = streamStates.keys.foreach(id => updateState(id, { x => x.shutdown(); Closed }))
   def resetStream(streamId: Int, errorCode: ErrorCode): Unit = {
-    streamStates -= streamId
-    debug(s"Incoming side of stream [$streamId]: resetting with code [$errorCode]")
+    updateState(streamId, _ => Closed) // force stream to be closed
     multiplexer.pushControlFrame(RstStreamFrame(streamId, errorCode))
   }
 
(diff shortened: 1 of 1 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -237,7 +236,8 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
     def handle(event: StreamFrameEvent): StreamState =
       if (event.isInstanceOf[ParsedHeadersFrame] && streamStates.size > settings.maxConcurrentStreams) {
         // When trying to open a new Stream, if that op would exceed the maxConcurrentStreams, then refuse the op
-        resetStream(event.streamId, ErrorCode.REFUSED_STREAM)
+        debug("Peer trying to open stream that would exceed `maxConcurrentStreams`, refusing stream")
+        multiplexer.pushControlFrame(RstStreamFrame(event.streamId, ErrorCode.REFUSED_STREAM))
```
raboof, line 240: hmm, so `resetStream` is only to be called from outside the state machine, while inside the state machine you have to push the RST 'yourself'?

raboof (approved):

jrudolph:
PLS BUILD

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -237,7 +236,8 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
     def handle(event: StreamFrameEvent): StreamState =
       if (event.isInstanceOf[ParsedHeadersFrame] && streamStates.size > settings.maxConcurrentStreams) {
         // When trying to open a new Stream, if that op would exceed the maxConcurrentStreams, then refuse the op
-        resetStream(event.streamId, ErrorCode.REFUSED_STREAM)
+        debug("Peer trying to open stream that would exceed `maxConcurrentStreams`, refusing stream")
```
ignasi35, line 239: Good addition. Could this be `info`, though?

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -130,8 +130,7 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
   /** Called to cleanup any state when the connection is torn down */
   def shutdownStreamHandling(): Unit = streamStates.keys.foreach(id => updateState(id, { x => x.shutdown(); Closed }))
   def resetStream(streamId: Int, errorCode: ErrorCode): Unit = {
-    streamStates -= streamId
-    debug(s"Incoming side of stream [$streamId]: resetting with code [$errorCode]")
+    updateState(streamId, _ => Closed) // force stream to be closed
```
ignasi35, line 133: 👍🏼

ignasi35 (approved):

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
jrudolph: I considered that but since it is somewhat expected that servers run into that case for some clients, I'd rather avoid spamming the log with client-induced log messages.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
jrudolph: Unfortunately, yes. The alternative would be either to introduce an effect system to make side-effects more explicit, or to introduce post-actions that automatically reset a stream if it moved to `Closed` state. I might experiment with one of them but for now this seems to be the right way.

PR #3585, opened 2020-10-29
[Preview] HTTP/2 test scripts

jrudolph:
This is an attempt to write at least some tests for the HTTP/2 (server) logic in form of test scripts.

Potential problems:

 * exposes too many implementation details like certain frame ordering details
 * can only indirectly assert on server handler user API by providing fixed server handlers and examining how they interact

### DIFF
--- a/akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2FrameProbe.scala
+++ b/akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2FrameProbe.scala
@@ -5,19 +5,23 @@
 package akka.http.impl.engine.http2
 
 import akka.actor.ActorSystem
+import akka.http.impl.engine.http2.FrameEvent.{ HeadersFrame, ParsedHeadersFrame }
 import akka.http.impl.engine.http2.Http2FrameProbe.FrameHeader
 import akka.http.impl.engine.http2.Http2Protocol.ErrorCode
 import akka.http.impl.engine.http2.Http2Protocol.Flags
 import akka.http.impl.engine.http2.Http2Protocol.FrameType
 import akka.http.impl.engine.http2.framing.Http2FrameParsing
+import akka.http.impl.engine.http2.hpack.ByteStringInputStream
 import akka.http.impl.engine.ws.ByteStringSinkProbe
 import akka.stream.impl.io.ByteStringParser.ByteReader
 import akka.stream.scaladsl.Sink
 import akka.util.ByteString
+import com.twitter.hpack.HeaderListener
 
 import scala.annotation.tailrec
 import org.scalatest.matchers.should.Matchers
 
+import scala.collection.immutable.VectorBuilder
 import scala.concurrent.duration.FiniteDuration
 
 private[http] trait Http2FrameProbe {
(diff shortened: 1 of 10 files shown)


ignasi35 (commented):
I'd rather use a strongly-typed, easy to refactor way to describe scenarios. Maybe we can keep the parser implementation introduced in this PR, and combine the existing senders and expecters with stuff like:

```
send("""
  HEADERS 5 END_STREAM
  :method    -> POST
  :scheme    -> https
  :path      -> /request-toString
  :authority -> www.example2.com
  accept-encoding -> gzip 
""")
```

I find the use of `<` and `>` for send and expect a bit confusing. My mental model of the `Demux` (being a bidiflow) is of things going one way or the other, and `<`/`>` make it hard to remember what is send and what is expect (considering send can be in either direction).

jrudolph:
> I'd rather use a strongly-typed

The point of the external DSL is to be more constrained than manually written tests. In this way it is more strictly typed than Scala code that executes side-effects freely.

The main remaining problem is that

 1. There are inherent race conditions that somehow need to be resolved without failing tests regularly (control frames vs. data frames, sending frames vs. expecting frames). We have the same problem with the regular tests but have more ways to resolve them (but often don't really do it)
 2. There are internals that we don't really want to test (exact chunking into DATA frames, exact values and timing of WINDOW_UPDATES, etc)
 3. The only way to specify server behavior is by writing a route in the test file. This might be too inaccurate for some tests

> easy to refactor way to describe scenarios. 

Not sure what you mean with that?

Maybe I should redefine the purpose of these tests: for some purposes it makes sense to record communication traces and keep asserting that those traces can be reproduced after changes. Even if they change it might be interesting to see what the change is (but, of course, it comes with similar caveats as described above)

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2FrameProbe.scala
```diff
@@ -113,17 +121,37 @@ private[http] object Http2FrameProbe extends Matchers {
       def expectNoBytes(timeout: FiniteDuration): Unit = probe.expectNoBytes(timeout)
 
       def expectFrame(): FrameEvent = {
-        // Not supporting large frames or high streamId's here for now, throw when we encounter those.
-        probe.expectBytes(2) should be(ByteString(0, 0))
-        val length = probe.expectByte()
+        val length =
```
raboof, line 124: Nice to support large frames

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2FrameProbe.scala
```diff
@@ -113,17 +121,37 @@ private[http] object Http2FrameProbe extends Matchers {
...
+          probe.expectByte() << 24 |
+            probe.expectByte() << 16 |
+            probe.expectByte()
         val _type = FrameType.byId(probe.expectByte()).get
         val flags = new ByteFlag(probe.expectByte())
+        // Not supporting high streamId's here for now, throw when we encounter those.
```
raboof, line 130: This comment is still accurate, though, right?

raboof (commented):
> Maybe I should redefine the purpose of these tests: for some purposes it makes sense to record communication traces and keep asserting that those traces can be reproduced after changes. Even if they change it might be interesting to see what the change is (but, of course, it comes with similar caveats as described above)

I'm not sure I'm convinced the scripted tests achieve that better than `Http2ServerSpec`/`Http2ClientSpec`. In `Http2ServerSpec` the mocked user handler is more intertwined in the test instead of having the route-under-test in a separate file, but that actually seems like an advantage of the `Spec` approach?

(I'm not opposed to introducing this style of tests, but I don't really see the advantage yet)

ignasi35:
I was a bit skeptical but this approach is growing on me. 
I'm not a great fan of new DSLs that come with no IDE support or discoverability but this new format is not meant to replace what we have bu simplify some of the tests.

jrudolph:
Not currently pursuing this further. Can be revisited later. The hardest problem to solve is how inherent race conditions could be specified in this script.

PR #3586, opened 2020-10-29
Allow TLS and non-TLS outgoing HTTP2 connections

raboof:


ignasi35 (approved):
LGTM

jrudolph (commented):
Looks simple enough but needs tests.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/ConnectionSpec.scala
```diff
@@ -0,0 +1,54 @@
...
+import akka.http.scaladsl.model.HttpResponse
+import akka.http.scaladsl.model.StatusCodes.OK
+import akka.http.scaladsl.server.Directives.complete
+import akka.http.scaladsl.settings.ClientConnectionSettings
+import akka.stream.OverflowStrategy
+import akka.stream.scaladsl.{Keep, Sink, Source}
+import akka.util.ByteString
+import org.scalatest.concurrent.ScalaFutures
+
+class ConnectionSpec extends AkkaSpecWithMaterializer(
```
jrudolph, line 20: Could we put that into `Http2ClientServerSpec` or is there some context that is different enough to warrant a new spec file?

raboof: `Http2ClientServerSpec` sets up the (secure) connection in the `TestSetup`.

Adding a test there that does not use `TestSetup` seemed a bit inconsistent.

Alternatively we could extend `TestSetup` to allow both secure and insecure tests. Originally I decided against that because I didn't want to make `TestSetup` more complicated just for one test. Thinking about it now, it might still make sense to add that, to make it easier to run a single test there and observe it with wireshark.

That said, I think there's still reason to keep this `ConnectionSpec`: it can be used to test the internal connection API in more detail. In particular, I noticed while writing this test that when using this API with a simple `Source.single` and `Sink.head`, the graph would complete even though the response was not written yet. I didn't want to get further into that for this PR, but that seems like something that is worth investigating.

jrudolph: > `Http2ClientServerSpec` sets up the (secure) connection in the `TestSetup`.
> 
> Adding a test there that does not use `TestSetup` seemed a bit inconsistent.

It's more about grouping tests by functionality so when adding features it's clear where to look for existing tests.

> Alternatively we could extend `TestSetup` to allow both secure and insecure tests. Originally I decided against that because I didn't want to make `TestSetup` more complicated just for one test. Thinking about it now, it might still make sense to add that, to make it easier to run a single test there and observe it with wireshark.

Either that, or don't use TestSetup which would be fine for me as well (as long as there not multiple TestSetups per spec).

> That said, I think there's still reason to keep this `ConnectionSpec`: it can be used to test the internal connection API in more detail. In particular, I noticed while writing this test that when using this API with a simple `Source.single` and `Sink.head`, the graph would complete even though the response was not written yet. I didn't want to get further into that for this PR, but that seems like something that is worth investigating.

Good point, created https://github.com/akka/akka-http/issues/3594 to track this feature.

raboof: > It's more about grouping tests by functionality so when adding features it's clear where to look for existing tests.

From that perspective I think the separate `ConnectionSpec` is better, as those are tests on a different 'level' after all.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2.scala
```diff
@@ -190,29 +190,40 @@ private[http] final class Http2Ext(private val config: Config)(implicit val syst
...
+      // TODO FIXME configure hostname verification for this case
+      case Left(ssl) =>
+        val e = ssl.sslContext.createSSLEngine(host, port)
+        TlsUtils.applySessionParameters(e, ssl.firstSession)
+        e
+      case Right(e) => e(Some((host, port)))
+    }
+    engine.setUseClientMode(true)
+    Http2AlpnSupport.clientSetApplicationProtocols(engine, Array("h2"))
+    engine
```
johanandren, line 226: Since this is gone in #3511 is there something from this PR to incorporate there rather than merge this?

jrudolph: I'd rather keep all the stream setup together in this file, in fact.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/ConnectionSpec.scala
jrudolph: I see it more like a settings variation of what's otherwise also tested in Http2ClientServerSpec using a reimplementation of basically the same infrastructure (queues instead of probes). :)

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/ConnectionSpec.scala
```diff
@@ -0,0 +1,58 @@
...
+  """akka.http.server.preview.enable-http2 = on
+    |""".stripMargin) with ScalaFutures {
+  val handler = complete(OK, "Test successful")
+
+  "The HTTP/2 connection API" should {
+    "establish a secure connection" in {
+      val binding =
```
jrudolph, line 27: Maybe, DRY up those two tests which test the same thing under different circumstances?

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/ConnectionSpec.scala
```diff
@@ -0,0 +1,58 @@
...
+import akka.http.scaladsl.model.StatusCodes.OK
+import akka.http.scaladsl.server.Directives.complete
+import akka.http.scaladsl.settings.ClientConnectionSettings
+import akka.stream.OverflowStrategy
+import akka.stream.scaladsl.{ Keep, Sink, Source }
+import akka.util.ByteString
+import org.scalatest.concurrent.ScalaFutures
+
+class ConnectionSpec extends AkkaSpecWithMaterializer(
+  """akka.http.server.preview.enable-http2 = on
```
jrudolph, line 21: Also add the usual debug loggers as in the other tests.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2.scala
johanandren: Let me rephrase that now that it is moved back into `Http2` in that PR, is there something for prior knowledge or regular HTTP/2 in here that I have missed in the connection methods over there. I added a test in `WithPriorKnowledgeSpec` which seems to pass, and so does the sample applications using regular HTTP2 against akka.io, but maybe I missed something.

raboof: I think  #3511 indeed makes this one redundant. Since that one is so close now I think we can focus on that and close this one.

Issue #3589, opened 2020-10-30
Failed: scala collection compat compiler errors for 2.12 on nightly


seglo:
Various builds failing: 

https://jenkins.akka.io:8498/job/akka-http-nightly/311//consoleFull
https://jenkins.akka.io:8498/job/akka-http-bench-nightly/221//consoleFull

Ex)

```
[info] Compiling 19 Scala sources and 1 Java source to /home/jenkins/workspace/workspace/akka-http-nightly/AKKA_VERSION/master/Node/akka-http-nightly-node/SCALA_VERSION/2.12.12/jdk/AdoptOpenJDK 8/akka-http2-support/target/scala-2.12/test-classes ...
[error] /home/jenkins/workspace/workspace/akka-http-nightly/AKKA_VERSION/master/Node/akka-http-nightly-node/SCALA_VERSION/2.12.12/jdk/AdoptOpenJDK 8/akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala:315:29: type mismatch;
[error]  found   : Seq[akka.http.impl.engine.http2.FrameEvent.Setting] (in scala.collection) 
[error]  required: Seq[akka.http.impl.engine.http2.FrameEvent.Setting] (in scala.collection.immutable) 
[error]     sendFrame(SettingsFrame(initialServerSettings))
[error]                             ^
[error] /home/jenkins/workspace/workspace/akka-http-nightly/AKKA_VERSION/master/Node/akka-http-nightly-node/SCALA_VERSION/2.12.12/jdk/AdoptOpenJDK 8/akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala:1299:29: type mismatch;
[error]  found   : Seq[akka.http.impl.engine.http2.FrameEvent.Setting] (in scala.collection) 
[error]  required: Seq[akka.http.impl.engine.http2.FrameEvent.Setting] (in scala.collection.immutable) 
[error]     sendFrame(SettingsFrame(initialClientSettings))
[error]                             ^
[error] two errors found
```

https://jenkins.akka.io:8498/job/akka-http-nightly/AKKA_VERSION=master,Node=akka-http-nightly-node,SCALA_VERSION=2.12.12,jdk=AdoptOpenJDK 8/311/console

jrudolph:
Fixed in #3590.

PR #3590, opened 2020-10-31
fix 3589

andreaTP:
Fix for #3589 
this encoding is cross-compatible across 2.12/2.13, since those are tests I assume we do not care about performance.

### DIFF
--- a/akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
+++ b/akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
@@ -24,6 +24,7 @@ import akka.stream.testkit.{ TestPublisher, TestSubscriber }
 import akka.util.ByteString
 import org.scalatest.concurrent.Eventually
 
+import scala.collection.immutable
 import scala.concurrent.duration._
 
 /**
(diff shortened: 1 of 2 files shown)


raboof (approved):
Hmm I guess we don't test PR's on 2.13 anymore since f8466e304c

jrudolph:
> Hmm I guess we don't test PR's on 2.13 anymore since [f8466e3](https://github.com/akka/akka-http/commit/f8466e304cf45ee8c3cdf24f43d5352db873e345)

No change in behavior, we run `mimaReportBinaryIssues` also for 2.12 but do not compile the tests. I will change the Jenkins job to compile the tests as well at least.

andreaTP:
The problem occurred because we don't compile the tests for 2.12, but I'm not sure about the whole chain.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
```diff
@@ -312,7 +313,7 @@ class Http2ClientSpec extends AkkaSpecWithMaterializer("""
     toNet.expectBytes(Http2Protocol.ClientConnectionPreface)
     expectSETTINGS()
 
-    sendFrame(SettingsFrame(initialServerSettings))
+    sendFrame(SettingsFrame(immutable.Seq.empty ++ initialServerSettings))
```
jrudolph, line 316:
```suggestion
    sendFrame(SettingsFrame(initialServerSettings.toVector))
```

jrudolph (approved):
LGTM

jrudolph:
Thanks, @andreaTP

PR #3591, opened 2020-11-01
http: optimize header directives by avoiding HttpHeader.unapply

nitikagarw:
The existing implementation of `optionalHeaderValueByName` is costly for the `Authorization` header due to the `unapply` on `value` which evaluates the header value for each such call.

For `Authorization` header, for example in case of `OAuth2BearerToken` evaluating the `value` causes the `render` to execute which causes a new string to be created and discarded for each such call.
In our case, the header is a jwt header which is 1kb in size, and such frequent eager evaluation this adds to ~ 9% of the time spent in our flow.

```
final case class OAuth2BearerToken(token: String) extends jm.headers.OAuth2BearerToken {
  def render[R <: Rendering](r: R): r.type = r ~~ "Bearer " ~~ token

  override def scheme: String = "Bearer"
  override def params: Map[String, String] = Map.empty
}
```
FlameGraph

**This PR avoids doing eager `value` evaluation and delays it till we have a matching header.**

### DIFF
--- a/akka-http/src/main/scala/akka/http/scaladsl/server/directives/HeaderDirectives.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/server/directives/HeaderDirectives.scala
@@ -135,7 +135,7 @@ trait HeaderDirectives {
   def optionalHeaderValueByName(headerName: String): Directive1[Option[String]] = {
     val lowerCaseName = headerName.toLowerCase
     extract(_.request.headers.collectFirst {
-      case HttpHeader(`lowerCaseName`, value) => value
+      case h: HttpHeader if h.is(lowerCaseName) => h.value
     })
   }
 


lightbend-cla-validator:
Hi @nitikagarw,

jrudolph (approved):
LGMT 

Great catch, thanks a lot @nitikagarw. We should maybe even deprecate the `unapply` for that reason.

johanandren (approved):

jrudolph:
Great, thanks again, @nitikagarw.

jrudolph:
There's another usage of `HttpHeader.unapply` in that file. Maybe you'd like to change that as well in another PR, @nitikagarw ?

nitikagarw:
@jrudolph Done the changes in #3604 

> There's another usage of HttpHeader.unapply in that file. Maybe you'd like to change that as well in another PR, @nitikagarw ?

Issue #3233, opened 2020-06-09
HTTP/2: investigate potential stream leak


jrudolph:
In Http2StreamHandling, an `IncomingStreamBuffer` reference might be dropped, while that is still serving stream data to the user. That sounds dangerous because the outlet might be let hanging if the connection dies before the remaining data has been delivered.

Maybe you could call it a bug in akka-stream, that still running `SubSourceOutlet` or a `SubSinkInlet` are not automatically aborted when the hosting GraphStage is stopped.

From a superficial glance at the implementations it seems that in this case, `SubSource` and `SubSink` call into async callbacks that are not backed by anything anymore (i.e. the actions are not dispatched any more).

jrudolph:
I cannot remember which states I meant specifically and how they would drop the reference. It seems the below condition in `IncomingStreamBuffer.onDataFrame` might be affected in an edge case of the peer sending too much unwanted data:

```
if (outstandingStreamWindow < 0) {
```

jrudolph:
I think I found more edge cases in `onPull` not going through the state machine. To be investigated further.

jrudolph:
Let's consider this done since #3621 is now merged.

PR #3592, opened 2020-11-02
Use assertAllStagesStopped in Http2ServerSpec and fix two edge case stream leaks

jrudolph:
Refs #3233

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
@@ -338,6 +338,7 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper 
         totalBufferedData += d.payload.size // padding can be seen as instantly discarded
 
         if (outstandingConnectionLevelWindow < 0) {
+          buffer.shutdown()
           pushGOAWAY(ErrorCode.FLOW_CONTROL_ERROR, "Received more data than connection-level window would allow")
           Closed
         } else {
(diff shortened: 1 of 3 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -338,6 +338,7 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
         totalBufferedData += d.payload.size // padding can be seen as instantly discarded
 
         if (outstandingConnectionLevelWindow < 0) {
+          buffer.shutdown()
```
jrudolph, line 341: These seem a bit like ad-hoc solutions to a more general problem:

 * it should not be possible to leak SubSinkInlets / SubSourceOutlets if the stage that handles them is shutdown (akka-stream problem)
 * we somehow need to keep track of those substreams and we currently do it explicitly in the state. Should we other logic to close those down automatically if we move to CLOSED state?

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -1212,6 +1223,19 @@ class Http2ServerSpec extends AkkaSpecWithMaterializer("""
...
+  implicit class InWithStoppedStages(name: String) {
+    def inAssertAllStagesStopped(runTest: => TestSetup) =
+      name in StreamTestKit.assertAllStagesStopped {
+        val setup = runTest
+
+        // force connection to shutdown (in case it is an invalid state)
+        setup.fromNet.sendError(new RuntimeException)
+        setup.toNet.cancel()
+
+        // and then assert that all stages, substreams in particular, are stopped
```
raboof, line 1235: .. which this taken care of by `StreamTestKit.assertAllStagesStopped` :+1:

raboof (approved):

johanandren (approved):

PR #3593, opened 2020-11-02
http2: run handler code in its own task on the server

jrudolph:
I tested the current server implementation on some real projects and noticed some unexpected slowness which was caused by routing DSL code being run directly in the handler and blocking progress of the HTTP/2 infrastructure and sibling requests.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
@@ -180,12 +180,17 @@ private[http] object Http2Blueprint {
   def handleWithStreamIdHeader(parallelism: Int)(handler: HttpRequest => Future[HttpResponse])(implicit ec: ExecutionContext): Flow[HttpRequest, HttpResponse, NotUsed] =
     Flow[HttpRequest]
       .mapAsyncUnordered(parallelism) { req =>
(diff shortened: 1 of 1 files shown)


raboof (approved):
hmm, surprising. would be good to have a benchmark but the comment is a good first step ;)

jrudolph:
> hmm, surprising. would be good to have a benchmark but the comment is a good first step ;)

Yep, for HTTP/1 we wouldn't notice it because with HTTP pipelining, only one request is ever run concurrently per connection, so the whole parallelism comes from using multiple connections.

johanandren (approved):

jrudolph:
> hmm, surprising. would be good to have a benchmark but the comment is a good first step ;)

Are you concerned about performance regressions or about the significance of this change?

This is basically about isolation of different requests and the infrastructure. In general, big parts of a routing structure are evaluated strictly (not least because of the usage of `FastFuture`). In the case where I found this issue, I used `parameters` and other directives to build and run a query on some data in memory. So, to fix that in user code would mean, that I would have to run CPU-constrained calculations in `Future {}` manually (which might be reasonable once you realize that this can be a source of starvation). If the calculation takes tens or hundreds of milliseconds, that's something you should consider anyway. However, even just executing a more complex route can take significant time alone, at least more time than the extra task dispatching will cost you.

Issue #3563, opened 2020-10-21
http2: support for receiving response trailers


jrudolph:
Trailers are received as a second and final HEADERS frame on a stream.

This is probably needed for the gRPC client as sending response trailers was needed for the server-side. (correct, @raboof?)

One difficulty is, that in the HTTP model we can encode trailers only with HttpEntity.Chunked. Currently, we create chunked entities when a stream comes in which doesn't have a `Content-Length` header set.

We could also expand that for receiving request trailers on the server (but only if required).

It seems, we cannot support trailers for responses with a `Content-Length` set (or we will have to generated Default entities in all cases).

raboof:
> This is probably needed for the gRPC client as sending response trailers was needed for the server-side. (correct, @raboof?)

Yes, the grpc status code and other metadata are sent in the trailing header, so we'd need to be able to get to those.

> We could also expand that for receiving request trailers on the server (but only if required).

I'm not aware of use cases for that (though I could think of them I guess ;) )

> It seems, we cannot support trailers for responses with a `Content-Length` set

`Strict` is `final`, so I guess we're free to add methods to that... but perhaps there is a nicer way.

jrudolph:
> > It seems, we cannot support trailers for responses with a `Content-Length` set
> 
> `Strict` is `final`, so I guess we're free to add methods to that... but perhaps there is a nicer way.

I think we should only do it if there's a use case with gRPC. Do gRPC requests ever have a `Content-Length` set for streamed channels?

raboof:
> Do gRPC ~requests~ responses ever have a `Content-Length` set for streamed channels?

I'm sure Akka gRPC never sends a Content-Length for a gRPC response, but I'm not sure what other frameworks do. I don't think the [spec](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#responses) forbids it (you can have 'custom metadata' in the response headers and content-length could be one of those AFAICS)

jrudolph:
So, the conclusion could be to emit a chunked entity in all cases where no `Content-Length` header was received, so we can publish the trailer as a `LastChunk`. This should work for gRPC which doesn't seem to use `Content-Length` commonly.

What would we do if we receive a `Content-Length` and also get a trailer? We could either just ignore the trailer or fail or warn.

raboof:
[...]

PR #3599, opened 2020-11-03
Accepting trailing headers in the HTTP/2 client

raboof:
A first stab at #3563.

I'm not too fond of the use of the promise/future here, but the obvious alternative (to make the data substream a stream of `ChunkStreamPart` instead of `ByteString`) seems invasive and inelegant as well... thoughts?

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
@@ -38,7 +38,7 @@ private[http2] class Http2ClientDemux(http2Settings: Http2CommonSettings, initia
   override def createSubstream(initialHeaders: ParsedHeadersFrame, data: Source[ByteString, Any], trailingHeaders: Future[ParsedHeadersFrame], correlationAttributes: Map[AttributeKey[_], _])(implicit ec: ExecutionContext): ChunkedHttp2SubStream =
     ChunkedHttp2SubStream(initialHeaders, data.map(ChunkStreamPart(_)).concat(Source.lazilyAsync(() => trailingHeaders.map(frame => LastChunk(extension = "", frame.keyValuePairs.map {
       case (k, v) => RawHeader(k, v)
-    })))), correlationAttributes)
+    }.toList)))), correlationAttributes)
 }
 
 private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initialRemoteSettings: immutable.Seq[Setting], upgraded: Boolean)
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
@@ -9,6 +9,8 @@ import akka.http.impl.engine.http2.FrameEvent._
 import akka.http.impl.engine.http2.Http2Protocol.ErrorCode
 import akka.http.impl.engine.http2.Http2Protocol.ErrorCode.FLOW_CONTROL_ERROR
 import akka.http.impl.engine.http2.Http2Protocol.SettingIdentifier
+import akka.http.scaladsl.model.AttributeKey
+import akka.http.scaladsl.model.HttpEntity.ChunkStreamPart
 import akka.http.scaladsl.settings.Http2CommonSettings
 import akka.macros.LogHelper
 import akka.stream.Attributes
@@ -16,6 +18,7 @@ import akka.stream.BidiShape
 import akka.stream.Inlet
 import akka.stream.Outlet
 import akka.stream.impl.io.ByteStringParser.ParsingException
+import akka.stream.scaladsl.Source
 import akka.stream.stage.GraphStage
 import akka.stream.stage.GraphStageLogic
 import akka.stream.stage.InHandler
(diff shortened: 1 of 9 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -489,6 +486,13 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
         None // don't change state
       }
     }
+    def onTrailingHeaders(headers: ParsedHeadersFrame): Unit = {
+      if (headers.endStream) {
+        onDataFrame(DataFrame(headers.streamId, endStream = true, ByteString.empty)) // simulate end stream by empty dataframe
```
jrudolph, line 491: Maybe add a comment why return value can be ignored.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -9,25 +9,44 @@ import akka.http.impl.engine.http2.FrameEvent._
...
 import scala.util.control.NonFatal
 
 /** Currently only used as log source */
 @InternalApi
-private[http2] sealed abstract class Http2ClientDemux
+private[http2] class Http2ClientDemux(http2Settings: Http2CommonSettings, initialRemoteSettings: immutable.Seq[Setting])
+  extends Http2Demux[ChunkedHttp2SubStream](http2Settings, initialRemoteSettings, upgraded = false, isServer = false) {
+
+  override def createSubstream(initialHeaders: ParsedHeadersFrame, data: Source[ByteString, Any], trailingHeaders: Future[ParsedHeadersFrame], correlationAttributes: Map[AttributeKey[_], _])(implicit ec: ExecutionContext): ChunkedHttp2SubStream =
+    ChunkedHttp2SubStream(initialHeaders, data.map(ChunkStreamPart(_)).concat(Source.lazilyAsync(() => trailingHeaders.map(frame => LastChunk(extension = "", frame.keyValuePairs.map {
```
jrudolph, line 39: That feels like a hack. I guess since it's a somewhat non-essential part, it might be good enough? Maybe add a comment about the reasoning behind that implementation.

Alternatively, could the logic to create the Http2SubStream be part of `IncomingStreamState` and emit the right elements there right away? I guess that would be more invasive because of the different element types that `IncomingStreamBuffer` would have to support.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -229,23 +230,24 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
...
 
     protected def expectIncomingStream(
-      event:           StreamFrameEvent,
-      nextStateEmpty:  StreamState,
-      nextStateStream: IncomingStreamBuffer => StreamState,
-      mapStream:       Http2SubStream => Http2SubStream    = identity): StreamState =
+      event:                 StreamFrameEvent,
+      nextStateEmpty:        StreamState,
+      nextStateStream:       IncomingStreamBuffer => StreamState,
+      correlationAttributes: Map[AttributeKey[_], _]             = Map.empty): StreamState =
```
jrudolph, line 236: :+1:

jrudolph (commented):
Looks quite good, though appending the trailer using a side-channel feels like a hack.

One thing we don't want to lose is to create `HttpEntity.Default` entities. Could the logic be based on whether a `Content-Length` exists or not?

 * `Content-Length` exists: HttpEntity.Default and trailers are ignored
 * `Content-Length` missing: HttpEntity.Chunked + trailers

Or do we expect any obvious problems with that strategy?

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -9,25 +9,44 @@ import akka.http.impl.engine.http2.FrameEvent._
...
+      case (k, v) => RawHeader(k, v)
```
jrudolph, line 40: Should these be converted to converted headers right away or later?

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -489,6 +486,13 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
         None // don't change state
       }
     }
+    def onTrailingHeaders(headers: ParsedHeadersFrame): Unit = {
+      if (headers.endStream) {
+        onDataFrame(DataFrame(headers.streamId, endStream = true, ByteString.empty)) // simulate end stream by empty dataframe
+      } else pushGOAWAY(Http2Protocol.ErrorCode.PROTOCOL_ERROR, "Got unexpected mid-stream HEADERS frame")
+
+      trailer.success(headers)
```
jrudolph, line 494: What happens in the case where the last frame is a DATA frame with `endStream = true`? Shouldn't `trailer` be completed in that case as well?

jrudolph: (Looking through IncomingStreamState, there seem to be a few loopholes, where we don't go through the state machine which should leave the stream state in a non-closed state after the final response data frame has been sent out)

ignasi35 (approved):
+1 to the cleanup and disambiguation around Demux and Server vs Client.

Some implementations detail goa bit over my head, though.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
raboof: Yes, that's the dilemma I meant in the PR message ;). I'll first fix up this PR and then have another go at coming up with an alternative approach.

raboof:
> Looks quite good, though appending the trailer using a side-channel feels like a hack.
> 
> One thing we don't want to lose is to create `HttpEntity.Default` entities. Could the logic be based on whether a `Content-Length` exists or not?
> 
>     * `Content-Length` exists: HttpEntity.Default and trailers are ignored
> 
>     * `Content-Length` missing: HttpEntity.Chunked + trailers
> 
> 
> Or do we expect any obvious problems with that strategy?

I think it's a tricky one. For example when a gRPC server returns an error, I think it would be valid to respond with headers (including a 200 status code and a content-length of 0), no data, and trailing headers including the grpc error status code. I'm OK with betting on the assumption that there's no implementations that would do that, but it's a risk.

raboof:
We should close this in favor of #3602

Issue #3268, opened 2020-06-22
Http2: fail as specified if server receives multiple HEADERS frames on the same stream


jrudolph:
As observed [here](https://github.com/akka/akka-http/pull/3267#discussion_r443466922), currently the second headers frame seems to just be ignored.

PR #3602, opened 2020-11-04
Accepting trailing headers in the HTTP/2 client

raboof:
Refs #3563.

Added a generic parameter to pass along the knowledge of what kinds of objects should be inserted into the streams (byte strings or chunks), so we can share code while still keeping it safe. The wrapData/wrapTrailingHeaders methods still feel hacky, but at least now they're pure functions and applied immediately when the respective frames come in, so I think this is at least better than the promise/future approach in #3599.

Another alternative would be to always use chunks (even for streams that will eventually be passed to the user code as non-chunked entities). That makes some of the generics simpler, but seems weird/wasteful in general.

There is 1 failing test that appears to be a bug that I haven't figured out yet: it seems IncomingStreamBuffer is not pulled when it has emitted all data but the trailing header is still pending.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
@@ -15,6 +15,7 @@ import akka.http.impl.engine.http2.hpack.{ HeaderCompression, HeaderDecompressio
 import akka.http.impl.engine.parsing.HttpHeaderParser
 import akka.http.impl.util.LogByteStringTools.logTLSBidiBySetting
 import akka.http.impl.util.StreamUtils
+import akka.http.scaladsl.model.HttpEntity.Chunk
 import akka.http.scaladsl.model._
 import akka.http.scaladsl.settings.{ ClientConnectionSettings, Http2CommonSettings, ParserSettings, ServerSettings }
 import akka.stream.TLSProtocol._
@@ -66,8 +67,11 @@ private[http2] final case class ChunkedHttp2SubStream(
     copy(correlationAttributes = attributes)
 
   @silent("never used")
-  def createResponseEntity(contentLength: Long, contentType: ContentType)(implicit ec: ExecutionContext): RequestEntity =
-    HttpEntity.Chunked(contentType, data)
+  def createResponseEntity(contentLength: Long, contentType: ContentType): RequestEntity =
+    // Ignore trailing headers when content-length is defined
+    if (contentLength == 0) HttpEntity.Empty
+    else if (contentLength > 0) HttpEntity.Default(contentType, contentLength, data.collect { case Chunk(bytes, _) => bytes })
+    else HttpEntity.Chunked(contentType, data)
 }
 
 /** INTERNAL API */
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
@@ -90,7 +90,7 @@ private[http] object Http2Blueprint {
       initialDemuxerSettings: immutable.Seq[Setting] = Nil,
       upgraded: Boolean = false): BidiFlow[HttpResponse, ByteString, ByteString, HttpRequest, NotUsed] =
     httpLayer(settings, log) atop
-      demux(settings.http2Settings, initialDemuxerSettings, upgraded) atop
+      serverDemux(settings.http2Settings, initialDemuxerSettings, upgraded) atop
       FrameLogger.logFramesIfEnabled(settings.http2Settings.logFrames) atop // enable for debugging
       hpackCoding() atop
       framing(log) atop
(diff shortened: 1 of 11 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
```diff
@@ -51,18 +52,25 @@ private[http2] final case class ByteHttp2SubStream(
...
-  override def withCorrelationAttributes(attributes: Map[AttributeKey[_], _]): Http2SubStream =
+) extends Http2SubStream[HttpEntity.ChunkStreamPart] {
+  override def withCorrelationAttributes(attributes: Map[AttributeKey[_], _]): Http2SubStream[HttpEntity.ChunkStreamPart] =
     copy(correlationAttributes = attributes)
+
+  def createResponseEntity(contentLength: Long, contentType: ContentType): RequestEntity = {
+    // Ignore trailing headers when content-length is defined
+    if (contentLength == 0) HttpEntity.Empty
+    else if (contentLength > 0) HttpEntity.Default(contentType, contentLength, data.collect { case Chunk(bytes, _) => bytes })
+    else HttpEntity.Chunked(contentType, data)
```
johanandren, line 72: If contentLength is not 0 and not > 0, when will it ever go here, can it be negative to signify undefined?

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -9,25 +9,55 @@ import akka.http.impl.engine.http2.FrameEvent._
...
 import akka.stream.stage.GraphStageLogic
 import akka.stream.stage.InHandler
 import akka.stream.stage.StageLogging
 import akka.util.ByteString
+import com.github.ghik.silencer.silent
 
 import scala.collection.immutable
 import scala.util.control.NonFatal
 
 /** Currently only used as log source */
```
johanandren, line 33: not true anymore

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -491,24 +490,41 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
...
+        outlet.push(wrapData(buffer.take(dataSize)))
         buffer = buffer.drop(dataSize)
 
         totalBufferedData -= dataSize
 
         debug(s"Dispatched chunk of $dataSize for stream [$streamId], remaining window space now $outstandingStreamWindow, buffered: ${buffer.size}")
         updateWindows()
       }
-      if (buffer.isEmpty && wasClosed) outlet.complete()
+      if (buffer.isEmpty && wasClosed) {
```
johanandren, line 517: Shouldn't this also have a `outlet.isAvailable` condition like the block above? The block above mutates the buffer so if it contained one last chunk it could try to do a second push with the trailing headers here (which would throw).

What triggers `outlet.complete()` if there was headers?

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
```diff
@@ -274,6 +279,31 @@ class Http2ClientSpec extends AkkaSpecWithMaterializer("""
...
+        val response = expectResponse()
+        response.entity shouldBe a[Chunked]
+
+        sendDATA(streamId, endStream = false, ByteString("asdf"))
+        sendHEADERS(streamId, endStream = true, Seq(RawHeader("grpc-status", "0")))
+
+        val chunks = response.entity.asInstanceOf[Chunked].chunks.runWith(Sink.seq).futureValue
+        chunks(0) should be(Chunk("asdf"))
+        chunks(1) should be(LastChunk(extension = "", List(RawHeader("grpc-status", "0"))))
+      }
```
johanandren, line 304: 👍 

Perhaps add coverage for failing if headers are incorrectly emitted mid stream as well.

johanandren (commented):
Looks promising, added some thoughts and questions.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
raboof: > can it be negative to signify undefined?

Exactly. I guess we should make this an `OptionVal` to make that clearer - but that might be for a different PR.

johanandren: That would require boxing to make it nullable though, so perhaps this is better even if a bit more obscure.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
raboof: > Shouldn't this also have a outlet.isAvailable condition like the block above?

`outlet.isAvailable` only has to be checked when there are trailers to be sent, so it happens a few lines below.

> What triggers `outlet.complete()` if there was headers?

Oooh good point. thanks.

raboof:
(doing the modeled headers now)

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -491,24 +490,42 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
...
+        trailingHeaders match {
+          case Some(trailer) =>
+            if (outlet.isAvailable) {
+              outlet.push(trailer)
+              trailingHeaders = None
+              outlet.complete()
+            }
```
johanandren, line 524: Ok, so in the scenario where it was not available, we leave here, and on the next pull we will end up here again.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -31,17 +32,19 @@ import scala.collection.immutable
...
   override def createSubstream(initialHeaders: ParsedHeadersFrame, data: Source[ChunkStreamPart, Any], correlationAttributes: Map[AttributeKey[_], _]): ChunkedHttp2SubStream =
     ChunkedHttp2SubStream(initialHeaders, data, correlationAttributes)
 
   def wrapData(bytes: akka.util.ByteString): ChunkStreamPart = Chunk(bytes)
-  def wrapTrailingHeaders(headers: ParsedHeadersFrame): Option[ChunkStreamPart] = Some(LastChunk(extension = "", headers.keyValuePairs.map {
-    // TODO convert to modeled headers
-    case (k, v) => RawHeader(k, v)
-  }.toList))
+  def wrapTrailingHeaders(headers: ParsedHeadersFrame): Option[ChunkStreamPart] = {
+    val headerParser = masterHttpHeaderParser.createShallowCopy()
```
johanandren, line 43: and trailing headers is optional and usually not there so it makes sense to defer copy until we se some, right?

johanandren (approved):
LGTM

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
```diff
@@ -78,7 +83,7 @@ class Http2ClientSpec extends AkkaSpecWithMaterializer("""
           ),
           expectedResponse =
             HPackSpecExamples.FirstResponse
-              .withEntity(Strict(ContentTypes.NoContentType, ByteString.empty))
+              .withEntity(Chunked(ContentTypes.`application/octet-stream`, Source.empty))
```
jrudolph, line 86: What happened here? That seems wrong for empty entities. We want that to come out as `HttpEntity.Empty`. What is the result for entities with a `content-length: 0` and a different content-type set?

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
johanandren: BTW failing on headers mid stream has a ticket #3268

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
jrudolph: Never mind, there are few things weird about how empty entities or responses that do not have entities are parsed but we can fix those later.

jrudolph (approved):
Basically LGTM, I have a few ideas how to remove the remaining type parameters as well unifying request and response handling even more, but maybe that can be done in after merging this.

jrudolph:
@raboof please merge when done!

jrudolph (changes requested):
Seems something went wrong with the rebase / merge?

raboof:
> Seems something went wrong with the rebase / merge?

Oops, looks like I wasn't quite up-to-date locally when I did the rebase. Should be OK now.

jrudolph (approved):
Yep, thanks.

Issue #812, opened 2017-01-24
Http2: add correctness checks to `RequestParsing` for http2 header correctness


gosubpl:
Add correctness checks to `RequestParsing` for checking http2 headers for correctness (like duplicated `content-length` issue). Model after `HttpMessageParser`

jrudolph:
Fixed in #3603.

PR #3603, opened 2020-11-04
Check HTTP2 headers for correctness

johanandren:
References #812 

Tried to read RFC7540 for MUST rules around headers and crosscheck with the HTTP/1.1 message parsing, but not super confident I did not miss something or get something wrong.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/RequestParsing.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/RequestParsing.scala
@@ -117,12 +117,16 @@ private[http2] object RequestParsing {
             malformedRequest("Pseudo-header ':status' is for responses only; it cannot appear in a request")
 
           case ("content-type", ct) =>
-            val contentType = ContentType.parse(ct).right.getOrElse(malformedRequest(s"Invalid content-type: '$ct'"))
-            rec(remainingHeaders.tail, method, scheme, authority, pathAndRawQuery, contentType, contentLength, cookies, true, headers)
+            if (contentType eq ContentTypes.`application/octet-stream`) {
+              val contentTypeValue = ContentType.parse(ct).right.getOrElse(malformedRequest(s"Invalid content-type: '$ct'"))
+              rec(remainingHeaders.tail, method, scheme, authority, pathAndRawQuery, contentTypeValue, contentLength, cookies, true, headers)
+            } else malformedRequest("HTTP message must not contain more than one content-type header")
 
           case ("content-length", length) =>
-            val contentLength = length.toLong
-            rec(remainingHeaders.tail, method, scheme, authority, pathAndRawQuery, contentType, contentLength, cookies, true, headers)
+            if (contentLength == -1) {
+              val contentLengthValue = length.toLong
+              rec(remainingHeaders.tail, method, scheme, authority, pathAndRawQuery, contentType, contentLengthValue, cookies, true, headers)
+            } else malformedRequest("HTTP message must not contain more than one content-length header")
 
           case ("cookie", value) =>
             // Compress cookie headers as described here https://tools.ietf.org/html/rfc7540#section-8.1.2.5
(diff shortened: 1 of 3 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/http2/RequestParsing.scala
```diff
@@ -117,12 +117,16 @@ private[http2] object RequestParsing {
             malformedRequest("Pseudo-header ':status' is for responses only; it cannot appear in a request")
 
           case ("content-type", ct) =>
-            val contentType = ContentType.parse(ct).right.getOrElse(malformedRequest(s"Invalid content-type: '$ct'"))
-            rec(remainingHeaders.tail, method, scheme, authority, pathAndRawQuery, contentType, contentLength, cookies, true, headers)
+            if (contentType eq ContentTypes.`application/octet-stream`) {
```
jrudolph, line 120: But what if the header *does* actually contain `application/octet-stream`? Maybe we should use `OptionVal` for the `contentType` parameter instead to figure that out?

akka-http-core/src/main/scala/akka/http/impl/engine/http2/RequestParsing.scala
```diff
@@ -117,12 +117,16 @@ private[http2] object RequestParsing {
...
+              val contentTypeValue = ContentType.parse(ct).right.getOrElse(malformedRequest(s"Invalid content-type: '$ct'"))
+              rec(remainingHeaders.tail, method, scheme, authority, pathAndRawQuery, contentTypeValue, contentLength, cookies, true, headers)
+            } else malformedRequest("HTTP message must not contain more than one content-type header")
 
           case ("content-length", length) =>
-            val contentLength = length.toLong
-            rec(remainingHeaders.tail, method, scheme, authority, pathAndRawQuery, contentType, contentLength, cookies, true, headers)
+            if (contentLength == -1) {
+              val contentLengthValue = length.toLong
```
jrudolph, line 127: Maybe check that value is not negative.

jrudolph (commented):
One potential problem with `application/octet-stream` content types but otherwise LGTM

akka-http-core/src/main/scala/akka/http/impl/engine/http2/client/ResponseParsing.scala
```diff
@@ -44,18 +44,23 @@ private[http2] object ResponseParsing {
           rec(remainingHeaders.tail, statusCodeValue.toInt, contentType, contentLength, seenRegularHeader, headers)
 
         case ("content-type", ct) =>
-          val contentType = ContentType.parse(ct).right.getOrElse(malformedRequest(s"Invalid content-type: '$ct'"))
-          rec(remainingHeaders.tail, status, contentType, contentLength, seenRegularHeader, headers)
+          if (contentType eq ContentTypes.`application/octet-stream`) {
```
jrudolph, line 47: Same fix needed here

johanandren: D'oh!

akka-http-core/src/main/scala/akka/http/impl/engine/http2/RequestParsing.scala
```diff
@@ -163,4 +170,17 @@ private[http2] object RequestParsing {
...
+  private[http2] def validateHeader(httpHeader: HttpHeader) = httpHeader.lowercaseName match {
+    case "connection" =>
+      // https://tools.ietf.org/html/rfc7540#section-8.1.2.2
+      malformedRequest("Header 'Connection' must not be used with HTTP/2")
+    case "transfer-encoding" =>
+      // https://tools.ietf.org/html/rfc7540#section-8.1.2.2
+      malformedRequest("Header 'Transfer-Encoding' must not be used with HTTP/2")
+    case "te" =>
+      // https://tools.ietf.org/html/rfc7540#section-8.1.2.2
+      if (httpHeader.value.compareToIgnoreCase("trailers") != 0) malformedRequest(s"Header 'TE' must not contain value other than 'trailers', value was '${httpHeader.value}")
```
raboof, line 182: I think TE is used as short for Transfer-Encoding in the spec here, so:

```suggestion
    case "transfer-encoding" =>
      // https://tools.ietf.org/html/rfc7540#section-8.1.2.2
      if (httpHeader.value.compareToIgnoreCase("trailers") != 0) malformedRequest(s"Header 'TE' must not contain value other than 'trailers', value was '${httpHeader.value}")
```

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/RequestParsingSpec.scala
```diff
@@ -110,6 +110,29 @@ class RequestParsingSpec extends AkkaSpec() with Inside with Inspectors {
         }
       }
 
+      "not accept TE with other values than 'trailers'" in {
```
raboof, line 113: (also needs to be reflected here)

raboof (commented):
Looks good, I think 'TE' just means 'Transfer-Encoding'

akka-http-core/src/main/scala/akka/http/impl/engine/http2/RequestParsing.scala
johanandren: I may be wrong but I think it actually is a distinct header: https://tools.ietf.org/html/rfc7230#section-4.3

raboof: huh, looks like you're completely right, sorry about he noise :(

johanandren: No worries, it sure is confusing with two headers meaning the same but not quite the same...

jrudolph (approved):
LGTM

jrudolph:
@raboof good to go from your side?

raboof (approved):
Yes, looks good!

PR #3604, opened 2020-11-04
http: optimize header directives by avoiding HttpHeader.unapply

nitikagarw:
Related PR: https://github.com/akka/akka-http/pull/3591
Please refer to the above PR description for more details

### DIFF
--- a/akka-http/src/main/scala/akka/http/scaladsl/server/directives/HeaderDirectives.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/server/directives/HeaderDirectives.scala
@@ -133,7 +133,7 @@ trait HeaderDirectives {
    * @group header
    */
   def optionalHeaderValueByName(headerName: String): Directive1[Option[String]] = {
-    val lowerCaseName = headerName.toLowerCase
+    val lowerCaseName = headerName.toRootLowerCase
     extract(_.request.headers.collectFirst {
       case h: HttpHeader if h.is(lowerCaseName) => h.value
     })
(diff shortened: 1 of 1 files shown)


akka-http/src/main/scala/akka/http/scaladsl/server/directives/HeaderDirectives.scala
```diff
@@ -133,7 +133,7 @@ trait HeaderDirectives {
    * @group header
    */
   def optionalHeaderValueByName(headerName: String): Directive1[Option[String]] = {
-    val lowerCaseName = headerName.toLowerCase
+    val lowerCaseName = headerName.toRootLowerCase
```
jrudolph, line 136: Good catch!

jrudolph (approved):
LGTM, thanks a lot :)

akka-http/src/main/scala/akka/http/scaladsl/server/directives/HeaderDirectives.scala
raboof, line 136: This makes sense to me, though I wonder if there aren't more places where we should do this

raboof (approved):

akka-http/src/main/scala/akka/http/scaladsl/server/directives/HeaderDirectives.scala
jrudolph: Definitely, but that one was related to the previous PR, so makes sense to do that here.

jrudolph:
Thanks again, @nitikagarw!

PR #3605, opened 2020-11-05
Http2SubStream cleanup

jrudolph:
Now we are back to a single version of Http2SubStream that can either be supplied by `ByteString` or `ChunkStreamPart` as it fits.

I also put all of the substream <=> entity conversions into that class so that it will be more easily reused from parsing / rendering.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
@@ -15,12 +15,11 @@ import akka.http.impl.engine.http2.hpack.{ HeaderCompression, HeaderDecompressio
 import akka.http.impl.engine.parsing.HttpHeaderParser
 import akka.http.impl.util.LogByteStringTools.logTLSBidiBySetting
 import akka.http.impl.util.StreamUtils
-import akka.http.scaladsl.model.HttpEntity.Chunk
 import akka.http.scaladsl.model._
 import akka.http.scaladsl.settings.{ ClientConnectionSettings, Http2CommonSettings, ParserSettings, ServerSettings }
 import akka.stream.TLSProtocol._
 import akka.stream.scaladsl.{ BidiFlow, Flow, Source }
-import akka.util.ByteString
+import akka.util.{ ByteString, OptionVal }
 
 import scala.concurrent.duration.{ Duration, FiniteDuration }
 import scala.concurrent.{ ExecutionContext, Future }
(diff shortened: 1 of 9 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
```diff
@@ -29,60 +28,35 @@ import scala.collection.immutable
...
-    copy(correlationAttributes = attributes)
-
-  def createResponseEntity(contentLength: Long, contentType: ContentType): RequestEntity =
-    // Ignore trailing headers when content-length is defined
-    if (contentLength == 0) HttpEntity.Empty
-    else if (contentLength > 0) HttpEntity.Default(contentType, contentLength, data.collect { case Chunk(bytes, _) => bytes })
-    else HttpEntity.Chunked(contentType, data)
+) {
+  def streamId: Int = initialHeaders.streamId
+  def hasEntity: Boolean = !initialHeaders.endStream
```
raboof, line 37: Oh neat :+1:

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
```diff
@@ -29,60 +28,35 @@ import scala.collection.immutable
...
+        // ignore: HttpEntity.LastChunk
+      }
+      HttpEntity.Default(contentType, contentLength, byteSource)
+    } else {
+      val chunkSource: Source[HttpEntity.ChunkStreamPart, Any] = data.map {
+        case b: ByteString                 => HttpEntity.Chunk(b)
+        case p: HttpEntity.ChunkStreamPart => p
+      }
+      HttpEntity.Chunked(contentType, chunkSource)
+    }
```
raboof, line 59: Aah, so the streams can just mix and match raw bytestrings and chunks, postponing selecting the right wrapping until here. That provides a lot of flexibility. Would we expect an impact on performance?

jrudolph: WDYM, because of the matching?

raboof: adding the additional `map` to the stream to do the wrapping here - though I guess that's all fused so it shouldn't matter?

jrudolph: It could matter but at least it's run in a user stream, i.e. not starving the HTTP/2 processing itself.

I'm a bit more concerned about the general performance implications of putting each frame separately through the whole HTTP/2 pipeline on input and output side. It seems that when TLS is used, it doesn't really matter because even on JDK 11, HTTP/2 seems to be TLS/crypto constrained but that may change in the future and we should continue to monitor potential performance issues.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
```diff
@@ -29,60 +28,50 @@ import scala.collection.immutable
...
-        if (contentLength > 0) HttpEntity.Default(contentType, contentLength, s.data)
-        else HttpEntity.Chunked.fromData(contentType, s.data)
-      /* contentLength undefined */
-      case c: ChunkedHttp2SubStream => HttpEntity.Chunked(contentType, c.data)
+  def hasEntity: Boolean = !initialHeaders.endStream
+
+  def withCorrelationAttributes(newAttributes: Map[AttributeKey[_], _]): Http2SubStream =
+    copy(correlationAttributes = newAttributes)
+
+  def createEntity(contentLength: Long, contentTypeOption: OptionVal[ContentType]): RequestEntity = {
```
raboof, line 43:
```suggestion
  def createEntity(contentLength: Long, contentTypeOption: OptionVal[ContentType]): ResponseEntity = {
```

raboof (approved):

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
jrudolph: `RequestEntity <: ResponseEntity`, so `RequestEntity` is the more concrete type. Why should we change it?

johanandren (approved):

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
jrudolph: Actually it wouldn't compile because it is used for requests as well which have tighter requirements.

raboof: hm, not sure what I was thinking :laughing:

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
```diff
@@ -29,60 +28,50 @@ import scala.collection.immutable
...
+    def contentType: ContentType = contentTypeOption.getOrElse(ContentTypes.`application/octet-stream`)
```
adityamundra, line 44: Why this is by default set to application/octet-stream? Some client like K9s do not set Content-type header and on the k8s server side it defaults missing content type header to application/json. Using a code written with akka-http to proxy this connection leads to errors.

raboof: The content type is not optional in Akka HTTP, so we have to choose something. I suspect this is consistent with the HTTP/1.1 implementation, but I haven't checked - do you know?

jrudolph: Yes, it's the same for our HTTP/1.1 implementation: 

https://github.com/akka/akka-http/blob/eb7bc5803ac1ed561f85fff961abc3203ae3cefc/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpMessageParser.scala#L357-L360

[RFC 7231](https://datatracker.ietf.org/doc/html/rfc7231#section-3.1.1.5) says:

> If a Content-Type header field is not present, the recipient
> MAY either assume a media type of "application/octet-stream"
> ([RFC2046], Section 4.5.1) or examine the data to determine its type.

Issue #2640, opened 2019-08-14
Configurable HTTP/2 ping support


jroper:
It'd be great if Akka HTTP provided configurable HTTP/2 ping support, to ensure liveness of connections. This is particularly useful when streaming in gRPC, in cases where a streamed gRPC call may still be alive, but there just haven't been any events in either direction, so that that case can be distinguished from a dropped connection.

Technically, the configured idle timeout should probably not consider pings, though in many cases where pings are enabled, idle timeout would be disable because it doesn't make sense. What might make sense as an alternative is a timeout for when there are no active requests on the connection (or, a stream idle timeout for no active requests in the stream, and a connection idle timeout that's triggered when there's no streams).

jrudolph:
Thanks, James, that looks like a useful addition. Maybe we can take some inspiration from the existing ping support for WebSockets.

jrudolph:
https://github.com/grpc/grpc/blob/master/doc/keepalive.md has some ideas for how ping support could be structured:

 * only send PINGs when streams are open (otherwise fall back to idle-timeout)
 * restrict time / number of pings to send if no data is sent at all for a longer period
 * monitor incoming pings and close connection on overly frequent usages of pings

We should take those with a grain of salt and check whether they make sense for more general usage (especially being overly strict on incoming PINGs probably doesn't make sense if you don't control both client and server).

jrudolph:
Done in #3617.

PR #3607, opened 2020-11-06
Configurable HTTP/2 Ping support

johanandren:
References #2640 

More questions than solutions but a starting point

### DIFF
--- a/akka-http-core/src/main/resources/reference.conf
+++ b/akka-http-core/src/main/resources/reference.conf
@@ -431,6 +431,14 @@ akka.http {
 
       # Enable verbose debug logging for all ingoing and outgoing frames
       log-frames = false
+
+      # When there is no data transmitted, send a HTTP/2 ping frame with this interval to make sure the connection is
+      # kept alive, 0 disables sending ping frames. Must be full seconds.
+      ping-interval = 0s
+
+      # Fail the connection if a ping frame is sent and does not come back from the server within this timeout.
+      # Must be full seconds.
+      ping-timeout = 5s
     }
 
     #client-settings
--- a/akka-http-core/src/main/resources/reference.conf
+++ b/akka-http-core/src/main/resources/reference.conf
@@ -433,12 +433,16 @@ akka.http {
       log-frames = false
 
       # When there is no data transmitted, send a HTTP/2 ping frame with this interval to make sure the connection is
-      # kept alive, 0 disables sending ping frames. Must be full seconds.
+      # kept alive, 0s disables sending ping frames. Must be full seconds.
       ping-interval = 0s
 
       # Fail the connection if a ping frame is sent and does not come back from the server within this timeout.
-      # Must be full seconds.
+      # Must be full seconds and less than the ping interval.
       ping-timeout = 5s
+
+      # Fail the connection if more ping interchanges than this are done without any data or header frames.
+      # Setting to 0 disables this limit. Only in effect if ping-interval is enabled.
+      max-pings-without-data = 300000
     }
 
     #client-settings
--- a/akka-http-core/src/main/resources/reference.conf
+++ b/akka-http-core/src/main/resources/reference.conf
@@ -278,6 +278,18 @@ akka.http {
 
       # Enable verbose debug logging for all ingoing and outgoing frames
       log-frames = false
+
+      # When there is no data transmitted, send a HTTP/2 ping frame with this interval to make sure the connection is
+      # kept alive, 0s disables sending ping frames. Must be full seconds.
+      keepalive-time = 0s
+
+      # Fail the connection if a ping frame is sent and does not come back from the server within this timeout.
+      # Must be full seconds and less than the ping interval.
+      keepalive-timeout = 5s
+
+      # Fail the connection if more ping interchanges than this are done without any data or header frames.
+      # Setting to 0 disables this limit. Only in effect if ping-interval is enabled.
+      max-keepalives-without-data = 300000
     }
 
     websocket {
@@ -434,15 +446,15 @@ akka.http {
 
       # When there is no data transmitted, send a HTTP/2 ping frame with this interval to make sure the connection is
       # kept alive, 0s disables sending ping frames. Must be full seconds.
-      ping-interval = 0s
+      keepalive-time = 0s
 
       # Fail the connection if a ping frame is sent and does not come back from the server within this timeout.
       # Must be full seconds and less than the ping interval.
-      ping-timeout = 5s
+      keepalive-timeout = 5s
 
       # Fail the connection if more ping interchanges than this are done without any data or header frames.
       # Setting to 0 disables this limit. Only in effect if ping-interval is enabled.
-      max-pings-without-data = 300000
+      max-keepalives-without-data = 0
     }
 
     #client-settings
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
@@ -18,6 +18,7 @@ import akka.http.impl.util.StreamUtils
 import akka.http.scaladsl.model.HttpEntity.Chunk
 import akka.http.scaladsl.model._
 import akka.http.scaladsl.settings.Http2ClientSettings
+import akka.http.scaladsl.settings.Http2ServerSettings
 import akka.http.scaladsl.settings.{ ClientConnectionSettings, Http2CommonSettings, ParserSettings, ServerSettings }
 import akka.stream.TLSProtocol._
 import akka.stream.scaladsl.{ BidiFlow, Flow, Source }
@@ -103,6 +104,7 @@ private[http] object Http2Blueprint {
       upgraded: Boolean = false): BidiFlow[HttpResponse, ByteString, ByteString, HttpRequest, NotUsed] =
     httpLayer(settings, log) atop
       serverDemux(settings.http2Settings, initialDemuxerSettings, upgraded) atop
+      serverKeepaliveIfConfigured(settings.http2Settings) atop
       FrameLogger.logFramesIfEnabled(settings.http2Settings.logFrames) atop // enable for debugging
       hpackCoding() atop
       framing(log) atop
@@ -117,7 +119,7 @@ private[http] object Http2Blueprint {
     val masterHttpHeaderParser = HttpHeaderParser(settings.parserSettings, log)
     httpLayerClient(masterHttpHeaderParser, log) atop
       clientDemux(settings.http2Settings, masterHttpHeaderParser) atop
-      periodicPingIfEnabled(settings.http2Settings) atop
+      clientKeepaliveIfConfigured(settings.http2Settings) atop
       FrameLogger.logFramesIfEnabled(settings.http2Settings.logFrames) atop // enable for debugging
       hpackCoding() atop
       framingClient(log) atop
@@ -137,9 +139,13 @@ private[http] object Http2Blueprint {
     )
   }
 
-  def periodicPingIfEnabled(settings: Http2ClientSettings): BidiFlow[FrameEvent, FrameEvent, FrameEvent, FrameEvent, NotUsed] =
-    if (settings.pingInterval == Duration.Zero) BidiFlow.identity[FrameEvent, FrameEvent]
-    else PingTimeoutBidi(settings)
+  def clientKeepaliveIfConfigured(settings: Http2ClientSettings): BidiFlow[FrameEvent, FrameEvent, FrameEvent, FrameEvent, NotUsed] =
+    if (settings.keepaliveTime == Duration.Zero) BidiFlow.identity[FrameEvent, FrameEvent]
+    else Keepalive(settings.keepaliveTime, settings.keepaliveTimeout, settings.maxKeepalivesWithoutData)
+
+  def serverKeepaliveIfConfigured(settings: Http2ServerSettings): BidiFlow[FrameEvent, FrameEvent, FrameEvent, FrameEvent, NotUsed] =
+    if (settings.keepaliveTime == Duration.Zero) BidiFlow.identity[FrameEvent, FrameEvent]
+    else Keepalive(settings.keepaliveTime, settings.keepaliveTimeout, settings.maxKeepalivesWithoutData).reversed
 
   def idleTimeoutIfConfigured(timeout: Duration): BidiFlow[ByteString, ByteString, ByteString, ByteString, NotUsed] =
     timeout match {
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
@@ -17,6 +17,7 @@ import akka.http.impl.util.LogByteStringTools.logTLSBidiBySetting
 import akka.http.impl.util.StreamUtils
 import akka.http.scaladsl.model.HttpEntity.Chunk
 import akka.http.scaladsl.model._
+import akka.http.scaladsl.settings.Http2ClientSettings
 import akka.http.scaladsl.settings.{ ClientConnectionSettings, Http2CommonSettings, ParserSettings, ServerSettings }
 import akka.stream.TLSProtocol._
 import akka.stream.scaladsl.{ BidiFlow, Flow, Source }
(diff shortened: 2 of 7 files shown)


akka-http-core/src/main/resources/reference.conf
```diff
@@ -431,6 +431,14 @@ akka.http {
 
       # Enable verbose debug logging for all ingoing and outgoing frames
       log-frames = false
+
+      # When there is no data transmitted, send a HTTP/2 ping frame with this interval to make sure the connection is
+      # kept alive, 0 disables sending ping frames. Must be full seconds.
+      ping-interval = 0s
```
raboof, line 437: should we name this `keepalive-time` to be consistent with https://github.com/grpc/grpc/blob/master/doc/keepalive.md ?

akka-http-core/src/main/resources/reference.conf
```diff
@@ -431,6 +431,14 @@ akka.http {
...
+
+      # Fail the connection if a ping frame is sent and does not come back from the server within this timeout.
+      # Must be full seconds.
+      ping-timeout = 5s
```
raboof, line 441: (then this would be `keepalive-timeout`)

akka-http-core/src/main/resources/reference.conf
```diff
@@ -431,6 +431,14 @@ akka.http {
 
       # Enable verbose debug logging for all ingoing and outgoing frames
       log-frames = false
+
+      # When there is no data transmitted, send a HTTP/2 ping frame with this interval to make sure the connection is
```
raboof, line 435: Perhaps allow those to be configured separately for the server and the client side?

akka-http-core/src/main/scala/akka/http/impl/engine/http2/PingTimeoutBidi.scala
```diff
@@ -0,0 +1,99 @@
...
+
+object PingTimeoutBidi extends {
+
+  def apply(settings: Http2ClientSettings): BidiFlow[FrameEvent, FrameEvent, FrameEvent, FrameEvent, NotUsed] =
+    BidiFlow.fromGraph(new PingTimeoutStage(settings))
+
+  private val tick = "tick"
+  private val Ping = PingFrame(false, ByteString("abcdefgh")) // FIXME should we have unique ping payloads so we can identify?
+
+  // FIXME ignore ping/pong for the connection idle timeout? (by merging that logic into this stage perhaps?)
```
raboof, line 38: I'm not sure that is necessary: since missed pings would lead to a disconnect as well, perhaps it's no problem that the idle timeout is suppressed when there's pings?

raboof: Even if we don't implement the proposed logic from https://github.com/grpc/grpc/blob/master/doc/keepalive.md to not ping when there's no requests in flight, this would allow users to choose between 'active' pinging and a 'passive' idle timeout.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/PingTimeoutBidi.scala
```diff
@@ -0,0 +1,99 @@
...
+    val in1 = Inlet[FrameEvent]("in1")
+    val out1 = Outlet[FrameEvent]("out1")
+    val in2 = Inlet[FrameEvent]("in2")
+    val out2 = Outlet[FrameEvent]("out2")
+    val shape: BidiShape[FrameEvent, FrameEvent, FrameEvent, FrameEvent] = BidiShape.of(in1, out1, in2, out2)
+
+    override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new TimerGraphStageLogic(shape) {
+      // FIXME this requires settings to be at least seconds, is that fine grained enough
+      private val maxPingsWithoutResponse = settings.pingTimeout.toSeconds
+      private val pingEveryNTickWithoutData = settings.pingTimeout.toSeconds
```
raboof, line 50:
```suggestion
      private val pingEveryNTickWithoutData = settings.pingInterval.toSeconds
```
?

akka-http-core/src/main/scala/akka/http/impl/engine/http2/PingTimeoutBidi.scala
```diff
@@ -0,0 +1,99 @@
...
+  // FIXME perhaps not quite related but, protection against something like CVE-2019-9512 (Ping Flood)
+  private final class PingTimeoutStage(settings: Http2ClientSettings) extends GraphStage[BidiShape[FrameEvent, FrameEvent, FrameEvent, FrameEvent]] {
+    val in1 = Inlet[FrameEvent]("in1")
+    val out1 = Outlet[FrameEvent]("out1")
+    val in2 = Inlet[FrameEvent]("in2")
+    val out2 = Outlet[FrameEvent]("out2")
+    val shape: BidiShape[FrameEvent, FrameEvent, FrameEvent, FrameEvent] = BidiShape.of(in1, out1, in2, out2)
+
+    override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new TimerGraphStageLogic(shape) {
+      // FIXME this requires settings to be at least seconds, is that fine grained enough
```
raboof, line 48: sounds fine-grained enough to me (we should check/document for it I suppose)

akka-http-core/src/main/scala/akka/http/impl/engine/http2/PingTimeoutBidi.scala
```diff
@@ -0,0 +1,99 @@
...
+      private var outstandingPings = 0L
+
+      // to limit overhead rather than constantly rescheduling a timer and looking at system time we use a constant timer
+      // FIXME does that really matter (I think the idle timeout logic touches nanotime for every frame)?
+      schedulePeriodically(tick, 1.second)
+
+      override protected def onTimer(timerKey: Any): Unit = {
+        ticksWithoutData += 1L
+        if (outstandingPings > maxPingsWithoutResponse) {
+          ??? // FIXME what fail action, should this trigger a GOAWAY?
```
raboof, line 61: Good question, I don't know if GOAWAY or RST would make most sense.

We'd also have to make sure the connection is nicely cleaned up - failing the stream feels like an attractive mechanism for this, though then I guess we'll drop the connection without sending GOAWAY *or* RST, so that might not be great.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/PingTimeoutBidi.scala
```diff
@@ -0,0 +1,99 @@
...
+        }
+        if (ticksWithoutData % pingEveryNTickWithoutData == 0) {
+          // FIXME only emit when there are active calls/streams?
```
raboof, line 64: That could be an optional feature

akka-http-core/src/main/scala/akka/http/impl/engine/http2/PingTimeoutBidi.scala
```diff
@@ -0,0 +1,99 @@
...
+          pull(in1)
+        }
+      })
+      setHandlers(in2, out2, new InHandler with OutHandler {
+        override def onPush(): Unit = {
+          val frame = grab(in2)
+          frame match {
+            case PingFrame(true, _) =>
+              // FIXME should we even reset here rather, or else a missed response will stay missed forever
+              // do we need something more elaborate than a counter - unique ping messages and a timestamp for each/or ordering they were sent?
```
raboof, line 84: A counter feels reasonable to me...

raboof (commented):
Looks like a good direction to me!

akka-http-core/src/main/resources/reference.conf
johanandren: Hmm, I thought about that but that it also applies when not using gRPC and the HTTP/2 term is ping/pong?

Otoh keepalive is the purpose more than the underlying terminology so maybe that is better.

akka-http-core/src/main/resources/reference.conf
johanandren: Yeah, it recently struck me that it should be for both client and server, I started out only thinking of it for client side.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/PingTimeoutBidi.scala
johanandren: Ah, yes, ups.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/PingTimeoutBidi.scala
johanandren: Inspired by the grpc setup I went with a combination of counter and a pong-timeout. So each ping must get a response within a configured timeout (should be possible to disable that as well) or if there are more than N ping-pongs without any data that fails the connection as well.

Not sure if it needs be something more elaborate like the failure detection in Akka where one missing reply does not mean failure?

akka-http-core/src/main/scala/akka/http/impl/engine/http2/PingTimeoutBidi.scala
johanandren: Yeah, I haven't figured out how I would do that without quite a bit of inspection of the frames, which feels wasteful.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/PingTimeoutBidi.scala
raboof: > Not sure if it needs be something more elaborate like the failure detection in Akka where one missing reply does not mean failure?

I don't think so. In Akka, a broken TCP connection could be re-established, so then messages could be lost. In this case this is all tied to a TCP connection, and if it's broken then that must be handled at a higher level anyway, so I think it's less useful to allow missed messages.

akka-http-core/src/main/resources/reference.conf
johanandren: Updated to use keepalive as name for it.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
```diff
@@ -102,6 +104,7 @@ private[http] object Http2Blueprint {
       upgraded: Boolean = false): BidiFlow[HttpResponse, ByteString, ByteString, HttpRequest, NotUsed] =
     httpLayer(settings, log) atop
       serverDemux(settings.http2Settings, initialDemuxerSettings, upgraded) atop
+      serverKeepaliveIfConfigured(settings.http2Settings) atop
```
johanandren, line 107: Had some bad luck thinking here, in addition to the bidi incorrectly being reversed in the current PR state, it is trickier than just re-using the exact same logic: I tried it with a short strict request and a slow throttled streaming response, in which case the current logic keeping track if it saw any incoming data frames between pings does not work, because there will not be any incoming frames other than the pongs. Need to rethink that a bit.

jrudolph (commented):
We need to think about what we want to achieve with the pings. Usually, it's there to make sure that the connection is still healthy when the connection is currently unused but it is still useful (i.e. ongoing streams or the expectation that we soon want to run more requests).

So, we need some idea how to track open streams. E.g.

 * we always ping if there's no incoming (outgoing?) data regardless of open streams
 * we remove the TCP-level idle-timeout (which will never triggered because of PINGs) and reimplement the idle-timeout in the demux that ignores pings and just closes the connection when all streams are closed for a certain amount of time

alternatively

 * we track open streams in the ping logic (might mean that it must move into demux) and only ping when there are open streams

Open questions:

 * Should we consider outgoing and incoming data differently? There are some TCP error conditions that can be detected by sending packets (peer RSTs after reception or route is lost) while others can only be detected by timing out waiting for PONGs (packets are silently dropped on the way to peer). So, I guess for detecting an idle connection, we should conservatively only consider incoming packets.

akka-http-core/src/main/resources/reference.conf
```diff
@@ -431,6 +443,18 @@ akka.http {
...
+      # kept alive, 0s disables sending ping frames. Must be full seconds.
+      keepalive-time = 0s
```
jrudolph, line 449: Maybe just `ping-interval`?

akka-http-core/src/main/resources/reference.conf
```diff
@@ -431,6 +443,18 @@ akka.http {
...
+
+      # Fail the connection if a ping frame is sent and does not come back from the server within this timeout.
+      # Must be full seconds and less than the ping interval.
+      keepalive-timeout = 5s
```
jrudolph, line 453: I'd say the default is too low. Too easy to be stuck behind data that was just send at the same time as the ping was sent.

Maybe just `ping-ack-timeout`.

akka-http-core/src/main/resources/reference.conf
```diff
@@ -431,6 +443,18 @@ akka.http {
...
+
+      # Fail the connection if more ping interchanges than this are done without any data or header frames.
+      # Setting to 0 disables this limit. Only in effect if ping-interval is enabled.
+      max-keepalives-without-data = 0
```
jrudolph, line 457: What's the goal of this setting? A second level of guards, that still closes the connection after a while?

akka-http-core/src/main/resources/reference.conf
```diff
@@ -278,6 +278,18 @@ akka.http {
...
+      # kept alive, 0s disables sending ping frames. Must be full seconds.
+      keepalive-time = 0s
+
+      # Fail the connection if a ping frame is sent and does not come back from the server within this timeout.
+      # Must be full seconds and less than the ping interval.
+      keepalive-timeout = 5s
+
+      # Fail the connection if more ping interchanges than this are done without any data or header frames.
+      # Setting to 0 disables this limit. Only in effect if ping-interval is enabled.
+      max-keepalives-without-data = 300000
```
jrudolph, line 292: Where does that default value come from?

akka-http-core/src/main/resources/reference.conf
jrudolph: We could also treat that as a different feature. Would probably be nicer to configure just with a timeout.

akka-http-core/src/main/resources/reference.conf
raboof: those names/defaults were inspired by https://github.com/grpc/grpc/blob/master/doc/keepalive.md

akka-http-core/src/main/resources/reference.conf
johanandren: Yeah, it corresponds with the gRPC GRPC_ARG_HTTP2_MAX_PINGS_WITHOUT_DATA setting, so I guess it would somewhat replaces/duplicates the idle connection timeout.

akka-http-core/src/main/resources/reference.conf
johanandren: For naming I started out with ping as that is the protocol feature but then after some feedback from raboof I reconsidered and thought that it was better to name it after the purpose, hoping that is what I'd be looking for if I wanted to use it. It's not a super convinced stance though.

akka-http-core/src/main/resources/reference.conf
johanandren: I think it should perhaps not have a default at all since it needs to be lower than the interval with the current implementation, and the interval does not have a default but is set by the user to enable pinging.

akka-http-core/src/main/resources/reference.conf
jrudolph: I see. Keep-alive is just one of the purposes, though, the other one is detecting when a connection is gone. If we want to name it after that, those names seem a bit misleading. `keepalive-time` is not the time the connection is kept alive and `keepalive-timeout` is misleading because it's not the timeout after which the keepalive is stopped.

johanandren:
I was hoping I could keep this isolated from other logic but since there is no option that does not involve putting either pings or idle timeouts in the demuxing I think we can just as well put the ping there and only ping when there are active streams (but no data).

ignasi35 (commented):
Haven't reviewed the `BidiFlow` in detail. (brain is fried)

akka-http-core/src/main/resources/reference.conf
johanandren: The PR name is simple taken from the issue name, and is not super important.

The comment describes pretty much exactly what it does, when there is no data transmitted over the connection, a ping is sent with this interval, including first interval between last data and first ping. As soon as there is data the pinging stops, because there already is data moving across the wire so no keepalive is needed. I think interval is a better name than time.

johanandren:
Closing this as #3617 is a better way to go.

Issue #3606, opened 2020-11-05
[failure] Http2ServerSpec. [...]should fail if more data is received than stream-level window allows


ignasi35:
The test `akka.http.impl.engine.http2.Http2ServerSpec.The Http/2 server implementation should support stream for request entity data should fail if more data is received than stream-level window allows` seems to fail consistently. In https://jenkins.akka.io:8498/job/akka-http-nightly/318/ the same tests failed for all possible combinations of scala version and JDK version in `master`.

<details>
<summary>StackTrace</summary>

```
sbt.ForkMain$ForkError: org.scalatest.exceptions.TestFailedException: WINDOW_UPDATE was not equal to RST_STREAM
	at org.scalatest.matchers.MatchersHelper$.indicateFailure(MatchersHelper.scala:339)
	at org.scalatest.matchers.should.Matchers$AnyShouldWrapper.shouldBe(Matchers.scala:6982)
	at akka.http.impl.engine.http2.Http2FrameProbe$$anon$1.expectFrameFlagsStreamIdAndPayload(Http2FrameProbe.scala:206)
	at akka.http.impl.engine.http2.Http2FrameProbe$$anon$1.expectFrameFlagsAndPayload(Http2FrameProbe.scala:200)
	at akka.http.impl.engine.http2.Http2FrameProbe$$anon$1.expectFramePayload(Http2FrameProbe.scala:195)
	at akka.http.impl.engine.http2.Http2FrameProbe$$anon$1.expectRST_STREAM(Http2FrameProbe.scala:163)
	at akka.http.impl.engine.http2.Http2FrameProbe$$anon$1.expectRST_STREAM(Http2FrameProbe.scala:160)
	at akka.http.impl.engine.http2.Http2FrameProbeDelegator.expectRST_STREAM(Http2FrameProbe.scala:85)
	at akka.http.impl.engine.http2.Http2FrameProbeDelegator.expectRST_STREAM$(Http2FrameProbe.scala:85)
	at akka.http.impl.engine.http2.Http2ServerSpec$TestSetupWithoutHandshake.expectRST_STREAM(Http2ServerSpec.scala:1239)
	at akka.http.impl.engine.http2.Http2ServerSpec$$anon$20.<init>(Http2ServerSpec.scala:414)
	at akka.http.impl.engine.http2.Http2ServerSpec.$anonfun$new$28(Http2ServerSpec.scala:408)
[...]

Issue #3610, opened 2020-11-09
Discuss whether WINDOW_UPDATEs should be sent out eagerly


jrudolph:
The "WINDOW goals" for incoming data are configured using the `http2.incoming-connection-level-buffer-size` and `http2.incoming-stream-level-buffer-size` settings. However, these values are not immediately conveyed to the peer:

 * receiving the first DATA frame will trigger a connection-level WINDOW_UPDATE
 * pulling data from a message entity stream will trigger the first stream-level WINDOW_UPDATE

The potential disadvantage of the current behavior is that peers (especially clients) might be not allowed to use the full bandwidth of a connection (as configured through the settings) until data is first received (on the connection or the entity stream).

For servers that might be a conservative behavior after all: requests don't include entity data in most cases (like usual GET requests). On a connection that only ever uses GET requests, no WINDOW_UPDATEs will ever be sent.

For clients, (response) entity data is quite likely on the other hand, so we might optimize there by sending out WINDOW_UPDATEs eagerly, so that servers are immediately allowed to send as much response data as configured. (If we go that way, we should use the same behavior for servers and clients and don't care about those little extra WINDOW_UPDATE frames that we would send eagerly).

ignasi35:
I think eagerly sending WINDOW_UPDATE is preferable in gRPC:

 - request payload is very likely (even if small)
 - network bandwith (especially in microservices setups) is way higher than general web use cases (browser-server)

So in that case eagerly sending WINDOW_UPDATE seems more convenient.

PS. I think Firefox sends a WINDOW_UPDATE immediately after the initial SETTINGS.

jrudolph:
> I think eagerly sending WINDOW_UPDATE is preferable in gRPC:
> 
>     * request payload is very likely (even if small)
> 
>     * network bandwidth (especially in microservices setups) is way higher than general web use cases (browser-server)

Not sure it would really fit the use case then. In any case the default window size is 65536, which should be enough for non-streaming normal gRPC requests. Sending out the WINDOW_UPDATE eagerly will mainly benefit peers behind a high-bandwidth-delay product connection. In the worst case, sending lazily will add a half RTT to the first big data packet.

PR #3611, opened 2020-11-09
Stream window test fail fix

johanandren:
Make sure the stream window resize always happens.
Previously it would only be triggered in Akka 2.6 by a earlier pull causing the test to fail
while with Akka 2.5 would pass. This adds an explicit pull to trigger the same behavior on both.

Credits for fix mostly goes to jrudolph who helped me a lot.

References #3606

### DIFF
--- a/akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
+++ b/akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
@@ -409,8 +409,10 @@ class Http2ServerSpec extends AkkaSpecWithMaterializer("""
         // trigger a connection-level WINDOW_UPDATE
         sendDATA(TheStreamId, endStream = false, ByteString("0000"))
         expectWindowUpdate()
+        entityDataIn.expectUtf8EncodedString("0000")
+        expectWindowUpdate() // window resize/update triggered
 
-        sendFrame(DataFrame(TheStreamId, endStream = false, ByteString("0" * 100000)))
+        sendFrame(DataFrame(TheStreamId, endStream = false, ByteString("0" * 512001)))
         expectRST_STREAM(TheStreamId, ErrorCode.FLOW_CONTROL_ERROR)
       }
       "fail entity stream if advertised content-length doesn't match" in pending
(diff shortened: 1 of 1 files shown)


raboof (approved):
Seems OK, though it's a bit unfortunate we're now adding overly-specific 'noise' to this test: we don't care how many window updates we get, we only care that we get a RST when we exceed the window. You'd almost want something like `fishForMessage` in Akka, though that is not super nice to use either...

In any case this :+1: !

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -409,8 +409,10 @@ class Http2ServerSpec extends AkkaSpecWithMaterializer("""
         // trigger a connection-level WINDOW_UPDATE
         sendDATA(TheStreamId, endStream = false, ByteString("0000"))
         expectWindowUpdate()
+        entityDataIn.expectUtf8EncodedString("0000")
+        expectWindowUpdate() // window resize/update triggered
 
-        sendFrame(DataFrame(TheStreamId, endStream = false, ByteString("0" * 100000)))
+        sendFrame(DataFrame(TheStreamId, endStream = false, ByteString("0" * 512001)))
```
jrudolph, line 415:
```suggestion
        sendFrame(DataFrame(TheStreamId, endStream = false, ByteString("0" * 512001))) // more than default `incoming-stream-level-buffer-size = 512kB`
```

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -409,8 +409,10 @@ class Http2ServerSpec extends AkkaSpecWithMaterializer("""
         // trigger a connection-level WINDOW_UPDATE
         sendDATA(TheStreamId, endStream = false, ByteString("0000"))
         expectWindowUpdate()
+        entityDataIn.expectUtf8EncodedString("0000")
+        expectWindowUpdate() // window resize/update triggered
```
jrudolph, line 413:
```suggestion
        entityDataIn.expectUtf8EncodedString("0000")
        pollForWindowUpdates(500.millis) // window resize/update triggered
```

jrudolph (approved):
LGTM with a suggestion following @raboof's comment.

jrudolph:
> Seems OK, though it's a bit unfortunate we're now adding overly-specific 'noise' to this test: we don't care how many window updates we get, we only care that we get a RST when we exceed the window. 

Partly it's because of https://github.com/akka/akka-http/issues/3610.

> You'd almost want something like `fishForMessage` in Akka, though that is not super nice to use either...

Yes, good point, we have that :) See above.

PR #3612, opened 2020-11-09
Include stack traces in scalatest failures

raboof:
That makes failures triggered in helper methods easier to read


johanandren (approved):

jrudolph (approved):
Nice. I sometimes used it directly in the sbt shell, since it feels sometimes arbitrary if stack traces are shown or not. So, having them on always, seems much better.

raboof:
> it feels sometimes arbitrary if stack traces are shown or not

Yeah, IIRC `TestFailedExceptions` are displayed according to this option, but other exceptions are shown with stack trace regardless of it...

Issue #3588, opened 2020-10-29
Failed: HttpAppSpec, Java should start providing an ActorSystem and Settings


seglo:
```
java.util.concurrent.TimeoutException
	at java.base/java.util.concurrent.CompletableFuture.timedGet(CompletableFuture.java:1886)
	at java.base/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:2021)
	at akka.http.javadsl.server.HttpAppSpec.$anonfun$new$12(HttpAppSpec.scala:112)
	at akka.http.javadsl.server.HttpAppSpec.withMinimal(HttpAppSpec.scala:33)
	at akka.http.javadsl.server.HttpAppSpec.$anonfun$new$11(HttpAppSpec.scala:106)
	at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.scala:18)
	at org.scalatest.OutcomeOf.outcomeOf(OutcomeOf.scala:85)
	at org.scalatest.OutcomeOf.outcomeOf$(OutcomeOf.scala:83)
	at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
	at org.scalatest.Transformer.apply(Transformer.scala:22)
	at org.scalatest.Transformer.apply(Transformer.scala:20)
	at org.scalatest.wordspec.AnyWordSpecLike$$anon$3.apply(AnyWordSpecLike.scala:1076)
	at akka.http.impl.util.WithLogCapturing.$anonfun$withFixture$1(WithLogCapturing.scala:38)
	at akka.testkit.EventFilter.intercept(TestEventListener.scala:116)
	at akka.http.impl.util.WithLogCapturing.withFixture(WithLogCapturing.scala:36)
	at akka.http.impl.util.WithLogCapturing.withFixture$(WithLogCapturing.scala:23)
	at akka.http.impl.util.AkkaSpecWithMaterializer.withFixture(AkkaSpecWithMaterializer.scala:14)
	at org.scalatest.wordspec.AnyWordSpecLike.invokeWithFixture$1(AnyWordSpecLike.scala:1074)
	at org.scalatest.wordspec.AnyWordSpecLike.$anonfun$runTest$1(AnyWordSpecLike.scala:1086)
	at org.scalatest.SuperEngine.runTestImpl(Engine.scala:306)
	at org.scalatest.wordspec.AnyWordSpecLike.runTest(AnyWordSpecLike.scala:1086)
[...]

PR #3613, opened 2020-11-09
Make sure to fail binding promise if binding fails in MinimalHttpApp

johanandren:
Not sure this is the issue making it fail, because I'd expect that failure showing up in logs anyway, but it could provide more info if an exception is what is going on making the promise never complete.

References #3588


raboof (approved):
Nicer behavior in any case!

jrudolph (approved):
I looked into that quite a few times without understanding what might be going on. This should give us more clues, indeed!

One interesting thing is that the HttpApp will create another ActorSystem which might lead to the logging situation.

PR #3617, opened 2020-11-11
http2: configurable ping support through demuxer

johanandren:
Potentially superseeding #3607 by implementing the pings in the demuxer to be able to only send pings when there are active streams.

References #2640


akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -97,6 +97,9 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
         }
     }
 
+  def activeStreamCount(): Int =
+    streamStates.size // FIXME is this good enough?
```
ignasi35, line 101:
```
// FIXME is this good enough?
```

I think it is good, yeah. If we introduce this method, I think we should replace other usages of `streamStates.size` with it (see `hasCapacityToCreateStreams` above, for example).

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -304,6 +358,36 @@ private[http2] abstract class Http2Demux(http2Settings: Http2CommonSettings, ini
...
+      override protected def onTimer(timerKey: Any): Unit = timerKey match {
+        case Keepalive.Tick =>
+          pingState match {
+            case OptionVal.Some(state) =>
+              // don't do anything unless there are active streams
+              debug(s"keepalive tick, activeStreamCount: ${activeStreamCount()}, ticksWithoutData: ${state.ticksWithoutData}, pingsWithoutData: ${state.pingsWithoutData}, pingEveryN: ${state.pingEveryNTickWithoutData}")
+              if (activeStreamCount() > 0) {
+                state.ticksWithoutData += 1
+                val now = System.nanoTime()
+                if (state.lastPingTimestamp > 0L && (state.lastPingTimestamp - now) > state.maxKeepaliveTimeoutNanos) {
```
ignasi35, line 370: `now - state.lastPingTimestamp`  instead of `state.lastPingTimestamp - now`?

Should `maxKeepaliveTimeoutNanos` be checked when receiving a Ping(ACK)? As it is here, it looks like certain settings will make this code branch useless (if pings are sent more often than the configured `maxKeepaliveTimeoutNanos`), no?

A hacky way to check the ping ACK was received faster than `maxKeepaliveTimeoutNanos` would be to put the value of `now` as the ping payload.

ignasi35 (commented):
Looking good.

I don't know how to make outgoing data frames reset the ping counters either, though (there's a FIXME somewhere in the PR). 😅

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
johanandren: ups. yes, `now - lastPing` for sure.

`timeout` must be lower than `interval`, will be verified in settings loading before done.

I think it is better to not trust the remote side with data we use for something (I think the spec mandates that it is the same ping content that is in the pong though, so maybe we should verify that)

akka-http-core/src/main/resources/reference.conf
```diff
@@ -278,6 +278,18 @@ akka.http {
...
+      # delay and subsequent interval to make sure the connection is kept alive, 0s disables sending ping frames. Must be full seconds.
+      ping-interval = 0s
+
+      # Fail the connection if a ping frame is sent and does not come back from the server within this timeout.
+      # Must be full seconds and less than the ping interval.
+      ping-timeout = 0s
+
+      # Fail the connection if more ping interchanges than this are done without any data or header frames.
+      # Setting to 0 disables this limit. Only in effect if ping-interval is enabled.
+      max-pings-without-data = 0
```
jrudolph, line 292: Do we have a better understanding about why this is needed and what we want to achieve with it? Why would the user want to set this in units of "number of pings", would that correspond to a time `ping-interval * max-pings`? Maybe can move this out of this PR and discuss and work on it later.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -63,6 +74,22 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
...
 }
 
+/**
+ * INTERNAL API
+ */
+@InternalApi
+private[http2] object Keepalive {
+  case object Tick
+  val Ping = PingFrame(false, ByteString("abcdefgh"))
+  final class PingState(val maxKeepaliveTimeoutNanos: Long, val pingEveryNTickWithoutData: Long) {
```
jrudolph, line 84: Good to move this into an extra class. Could we move all of the ping functionality in there? Especially mutating the vars?

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -218,8 +277,18 @@ private[http2] abstract class Http2Demux(http2Settings: Http2CommonSettings, ini
...
             // `enforceSettings(initialLocalSettings)`
 
-            case PingFrame(true, _)  =>
-            // ignore for now (we don't send any pings)
+            case PingFrame(true, _) =>
+              pingState match {
+                case OptionVal.Some(state) =>
+                  state.pingsWithoutData += 1L
+                  if (settings.maxPingsWithoutData > 0) {
+                    // FIXME fail with specific exception or some more graceful failure?
```
jrudolph, line 285: Yep, should be GOAWAY

akka-http-core/src/main/resources/reference.conf
johanandren: I can see how you wouldn't want the pings to keep a stale request stream alive _forever_, that it is pings rather than some time span was mostly because that's what the gRPC one seems to do but I agree that maybe makes more sense.

No strong opinions on my part about this being important so I'm ok with skipping it in this PR.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
johanandren: Sure, I'll give it a try.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -218,8 +277,18 @@ private[http2] abstract class Http2Demux(http2Settings: Http2CommonSettings, ini
...
+                    if (state.pingsWithoutData > settings.maxPingsWithoutData)
+                      throw new RuntimeException("HTTP/2 more than " + settings.maxPingsWithoutData + " pings exchanged without any data frames. This is configureable by akka.http2.[client, server].max-keepalives-without-data")
+                  }
+                // FIXME also fail if too much time passed since last ping was sent?
```
jrudolph, line 289: Is this about the race condition between receiving the ack and running `onTimer`? I think no need to care for that, after all we received an ack. (Just need to make sure that `onTimer` doesn't kill the connection afer we handled the ack here, I cannot see if that's done currently?)

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -304,6 +373,33 @@ private[http2] abstract class Http2Demux(http2Settings: Http2CommonSettings, ini
...
 
+      override protected def onTimer(timerKey: Any): Unit = timerKey match {
+        case Keepalive.Tick =>
+          pingState match {
+            case OptionVal.Some(state) =>
+              // don't do anything unless there are active streams
+              if (activeStreamCount() > 0) {
+                state.ticksWithoutData += 1
+                val now = System.nanoTime()
+                if (state.lastPingTimestamp > 0L && (now - state.lastPingTimestamp) > state.maxKeepaliveTimeoutNanos) {
```
jrudolph, line 384: Is `lastPingTimestamp` reset somewhere (when receiving an ack)? How is it otherwise prevented that an ack is received and this logic here still triggers?

johanandren: I think I missed that reset when copying from the other PR, it should be on receiving ack.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
johanandren: I think maybe that race isn't very interesting and we can keep it only failing on timer if too much time passed without seeing ack.

jrudolph: Agreed.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
johanandren: Now changed, however `pushGOAWAY` does not lead to the connection being torn down, I guess that is what the 
` // FIXME: handle the connection closing according to the specification` is about?

akka-http-core/src/main/resources/reference.conf
johanandren: Removed now

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -1223,6 +1223,75 @@ class Http2ServerSpec extends AkkaSpecWithMaterializer("""
...
+
+        val responseStream = TestPublisher.probe[ByteString]()
+        val response1 = HttpResponse(entity = HttpEntity(ContentTypes.`application/octet-stream`, Source.fromPublisher(responseStream)))
+        emitResponse(1, response1)
+        expectDecodedHEADERS(streamId = 1, endStream = false) shouldBe response1.withEntity(HttpEntity.Empty.withContentType(ContentTypes.`application/octet-stream`))
+        // No data for 2s to trigger the ping
+        Thread.sleep(2000) // slow test but 2s is current minimum :(
+        expectFrame(FrameType.PING, ByteFlag.Zero, 0, ConfigurablePing.Ping.data)
+
+        // FIXME how to actually shut the connection down?
```
raboof, line 1265: `fromNet.sendComplete()` I think?

akka-http-core/src/main/mima-filters/10.2.1.backwards.excludes/http2-ping-support.excludes
```diff
@@ -0,0 +1,19 @@
...
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.scaladsl.settings.Http2ClientSettings#Http2ClientSettingsImpl.copy")
+ProblemFilters.exclude[IncompatibleResultTypeProblem]("akka.http.scaladsl.settings.Http2ClientSettings#Http2ClientSettingsImpl.copy$default$7")
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.scaladsl.settings.Http2ClientSettings#Http2ClientSettingsImpl.this")
+ProblemFilters.exclude[IncompatibleSignatureProblem]("akka.http.scaladsl.settings.Http2ServerSettings#Http2ServerSettingsImpl.unapply")
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.scaladsl.settings.Http2ServerSettings#Http2ServerSettingsImpl.copy")
+ProblemFilters.exclude[IncompatibleResultTypeProblem]("akka.http.scaladsl.settings.Http2ServerSettings#Http2ServerSettingsImpl.copy$default$7")
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.scaladsl.settings.Http2ServerSettings#Http2ServerSettingsImpl.this")
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.scaladsl.settings.Http2ServerSettings#Http2ServerSettingsImpl.apply")
```
jrudolph, line 11: Fine to use a broad single exclusion for implementation classes

```suggestion
ProblemFilters.exclude[Problem]("akka.http.scaladsl.settings.Http2ClientSettings#Http2ClientSettingsImpl.*")
```

akka-http-core/src/main/resources/reference.conf
```diff
@@ -278,6 +278,14 @@ akka.http {
 
       # Enable verbose debug logging for all ingoing and outgoing frames
       log-frames = false
+
+      # When there is no data transmitted in either direction, but there are active streams, send a HTTP/2 ping frame with this initial
+      # delay and subsequent interval to make sure the connection is kept alive, 0s disables sending ping frames. Must be full seconds.
+      ping-interval = 0s
+
+      # Fail the connection if a ping frame is sent and does not come back from the server within this timeout.
```
jrudolph, line 286: Maybe slightly simpler (also below for client):

```suggestion
      # Fail the connection if a sent ping is not acknowledged within this timeout.
```

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Multiplexer.scala
```diff
@@ -77,12 +77,10 @@ private[http2] trait Http2MultiplexerSupport { logic: GraphStageLogic with Stage
   /** Called by the multiplexer after SETTINGS_INITIAL_WINDOW_SIZE has changed */
   def distributeWindowDeltaToAllStreams(delta: Int): Unit
 
-  def createMultiplexer(outlet: GenericOutlet[FrameEvent], prioritizer: StreamPrioritizer): Http2Multiplexer =
+  def createMultiplexer(pushFrameOut: FrameEvent => Unit, prioritizer: StreamPrioritizer): Http2Multiplexer with OutHandler =
```
jrudolph, line 80: You can just define `pushFrameOut` above as we did for other callbacks back to `Http2MultiplexerSupport`.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientServerSpec.scala
```diff
@@ -108,20 +109,31 @@ class Http2ClientServerSpec extends AkkaSpecWithMaterializer(
     }
   }
   class TestSetup {
+    def serverSettings: Option[ServerSettings] = None
```
jrudolph, line 112: Just initialize to `ServerSettings(system)` to get rid of option.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientServerSpec.scala
```diff
@@ -108,20 +109,31 @@ class Http2ClientServerSpec extends AkkaSpecWithMaterializer(
...
+    def clientSettings: Option[ClientConnectionSettings] = None
```
jrudolph, line 113: Same here.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerWithPings.scala
```diff
@@ -0,0 +1,49 @@
...
+
+object Http2ServerWithPings {
+
+  def main(args: Array[String]): Unit = {
+    val config = ConfigFactory.parseString(
+      """
+         akka.loglevel = debug
+         akka.http.server.preview.enable-http2 = on
+         akka.http.server.http2.log-frames = true
+         akka.http.server.http2.ping-interval = 2s
```
jrudolph, line 29: Would be fine to merge that with `Http2ServerTest`.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
```diff
@@ -304,6 +305,66 @@ class Http2ClientSpec extends AkkaSpecWithMaterializer("""
       }
 
     }
+
+    "support for configurable ping" should {
+      "send pings when there is an active but slow stream from client" in new TestSetup with NetProbes with Http2FrameHpackSupport {
+        override def settings = {
+          val default = super.settings
+          default.withHttp2Settings(default.http2Settings.withPingInterval(2.seconds).withPingTimeout(1.second))
```
jrudolph, line 313: You can use `mapHttp2Settings`. Then it should be something like

```scala
super.settings.mapHttp2Settings(_.withPingInterval(...).withPingTimeout(...))
```

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
```diff
@@ -304,6 +305,66 @@ class Http2ClientSpec extends AkkaSpecWithMaterializer("""
...
+        }
+        val streamId = 0x1
+        val requestStream = TestPublisher.probe[ByteString]()
+        emitRequest(streamId, HttpRequest(
+          protocol = HttpProtocols.`HTTP/2.0`,
+          entity = HttpEntity(ContentTypes.`application/octet-stream`, Source.fromPublisher(requestStream))))
+        expectDecodedHEADERS(streamId, endStream = false)
+        Thread.sleep(2000) // no data for this interval should trigger ping
```
jrudolph, line 321: Maybe `expectNoBytes(2.seconds)` instead?

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
```diff
@@ -304,6 +305,66 @@ class Http2ClientSpec extends AkkaSpecWithMaterializer("""
...
+          protocol = HttpProtocols.`HTTP/2.0`,
+        ))
+        expectDecodedHEADERS(streamId)
+
+        sendHEADERS(streamId, endStream = false, Seq(
+          RawHeader(":status", "200"),
+          RawHeader("content-type", "application/octet-stream")
+        ))
+        val response = expectResponse()
+        Thread.sleep(2000) // no data for this interval should trigger ping
```
jrudolph, line 342: `expectNoBytes()`

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
```diff
@@ -304,6 +305,66 @@ class Http2ClientSpec extends AkkaSpecWithMaterializer("""
...
+          val default = super.settings
+          default.withHttp2Settings(default.http2Settings.withPingInterval(2.seconds).withPingTimeout(1.second))
+        }
+        val streamId = 0x1
+        val requestStream = TestPublisher.probe[ByteString]()
+        emitRequest(streamId, HttpRequest(
+          protocol = HttpProtocols.`HTTP/2.0`,
+          entity = HttpEntity(ContentTypes.`application/octet-stream`, Source.fromPublisher(requestStream))))
+        expectDecodedHEADERS(streamId, endStream = false)
+        Thread.sleep(2000) // no data for this interval should trigger ping
```
jrudolph, line 359: `expectNoBytes(...)`

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
```diff
@@ -304,6 +305,66 @@ class Http2ClientSpec extends AkkaSpecWithMaterializer("""
...
+        expectFrame(FrameType.PING, ByteFlag.Zero, 0, ConfigurablePing.Ping.data)
+        Thread.sleep(1000)
+        // no ack sent back
+        expectGOAWAY(streamId)
+
+        fromNet.sendComplete()
```
jrudolph, line 365: `expectNoBytes(...)`

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -1223,6 +1223,68 @@ class Http2ServerSpec extends AkkaSpecWithMaterializer("""
...
+        override def settings: ServerSettings = {
+          val default = super.settings
+          default.withHttp2Settings(default.http2Settings.withPingInterval(2.seconds).withPingTimeout(1.second))
+        }
+
+        sendRequestHEADERS(1, HttpRequest(protocol = HttpProtocols.`HTTP/2.0`), endStream = false)
+        expectRequest()
+
+        // No data for 2s to trigger the ping
+        Thread.sleep(2000) // slow test but 2s is current minimum :(
```
jrudolph, line 1238: `expectNoBytes(...)`

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -1223,6 +1223,68 @@ class Http2ServerSpec extends AkkaSpecWithMaterializer("""
...
+        val theRequest = HttpRequest(protocol = HttpProtocols.`HTTP/2.0`)
+        sendRequest(1, theRequest)
+        expectRequest() shouldBe theRequest
+
+        val responseStream = TestPublisher.probe[ByteString]()
+        val response1 = HttpResponse(entity = HttpEntity(ContentTypes.`application/octet-stream`, Source.fromPublisher(responseStream)))
+        emitResponse(1, response1)
+        expectDecodedHEADERS(streamId = 1, endStream = false) shouldBe response1.withEntity(HttpEntity.Empty.withContentType(ContentTypes.`application/octet-stream`))
+        // No data for 2s to trigger the ping
+        Thread.sleep(2000) // slow test but 2s is current minimum :(
```
jrudolph, line 1260: `expectNoBytes(...)`

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -1223,6 +1223,68 @@ class Http2ServerSpec extends AkkaSpecWithMaterializer("""
...
+        override def settings: ServerSettings = {
+          val default = super.settings
+          default.withHttp2Settings(default.http2Settings.withPingInterval(2.seconds).withPingTimeout(1.second))
+        }
+
+        sendRequestHEADERS(1, HttpRequest(protocol = HttpProtocols.`HTTP/2.0`), endStream = false)
+        expectRequest()
+
+        // No data for 2s to trigger the ping
+        Thread.sleep(2000) // slow test but 2s is current minimum :(
```
jrudolph, line 1276: `expectNoBytes(...)`

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -1223,6 +1223,68 @@ class Http2ServerSpec extends AkkaSpecWithMaterializer("""
...
+        expectFrame(FrameType.PING, ByteFlag.Zero, 0, ConfigurablePing.Ping.data)
+        // no ack to make it time out
+        Thread.sleep(1000)
```
jrudolph, line 1279: `expectNoBytes(...)`

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -147,7 +192,24 @@ private[http2] abstract class Http2Demux(http2Settings: Http2CommonSettings, ini
 
       override protected def logSource: Class[_] = if (isServer) classOf[Http2ServerDemux] else classOf[Http2ClientDemux]
 
-      val multiplexer = createMultiplexer(frameOut, StreamPrioritizer.first())
+      def pushFrameOut(event: FrameEvent): Unit = {
+        pingState match {
+          case OptionVal.None =>
+          case OptionVal.Some(state) =>
+            event match {
+              case _: StreamFrameEvent => state.onDataFrameSeen()
```
jrudolph, line 200: Why only these frames and not also control frames?

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -168,6 +230,12 @@ private[http2] abstract class Http2Demux(http2Settings: Http2CommonSettings, ini
 
         // both client and server must send a settings frame as first frame
         multiplexer.pushControlFrame(SettingsFrame(initialLocalSettings))
+
+        if (pingState.isDefined) {
+          // to limit overhead rather than constantly rescheduling a timer and looking at system time we use a constant timer
```
jrudolph, line 235: I think it might matter enough, because otherwise, we would have to cancel and reschedule the timer basically for every single incoming and outgoing frame.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -168,6 +230,12 @@ private[http2] abstract class Http2Demux(http2Settings: Http2CommonSettings, ini
...
+          // FIXME does that really matter (I think the idle timeout logic touches nanotime for every frame)?
+          schedulePeriodically(ConfigurablePing.Tick, 1.second)
```
jrudolph, line 237: What would happen if we just take `pingInterval` as the tick interval to allow shorter intervals? We could document that `pingTimeout` will be checked in those same units (or even use the minimum of `pingInterval` and `pingTimeout`).

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -147,7 +192,24 @@ private[http2] abstract class Http2Demux(http2Settings: Http2CommonSettings, ini
...
+              case _                   =>
+            }
+        }
+        frameOut.push(event)
+      }
+
+      val multiplexer = createMultiplexer(pushFrameOut, StreamPrioritizer.first())
+      frameOut.setHandler(multiplexer)
+
+      val pingState: OptionVal[ConfigurablePing.PingState] =
```
jrudolph, line 210: Maybe you could make `PingState` a trait and provide a no-op instance, so we get rid of the extra `OptionVal` stuff?

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -304,6 +381,26 @@ private[http2] abstract class Http2Demux(http2Settings: Http2CommonSettings, ini
...
+      override protected def onTimer(timerKey: Any): Unit = timerKey match {
+        case ConfigurablePing.Tick =>
+          val state = pingState.get // no tick scheduled if state is not set
+
+          // don't do anything unless there are active streams
+          if (activeStreamCount() > 0) {
+            state.onTick()
+            val now = System.nanoTime()
+            if (state.pingAckOverdue(now)) {
+              // FIXME this does not actually close connection currently
```
jrudolph, line 393: I think you can remove that comment here as the same problem exists everywhere. `pushGOAWAY` should do the full job in the future.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -304,6 +381,26 @@ private[http2] abstract class Http2Demux(http2Settings: Http2CommonSettings, ini
...
+              pushGOAWAY(ErrorCode.CANCEL, "Ping ack timeout")
```
jrudolph, line 394: `CANCEL` is used to reset streams, not sure if that would match here. Maybe just use `PROTOCOL_ERROR` as a peer is required to ack pings.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -218,8 +291,12 @@ private[http2] abstract class Http2Demux(http2Settings: Http2CommonSettings, ini
             // are enforced from the start of the connection so there's no need to invoke
             // `enforceSettings(initialLocalSettings)`
 
-            case PingFrame(true, _)  =>
-            // ignore for now (we don't send any pings)
+            case PingFrame(true, _) =>
+              // FIXME should we verify the pong data?
```
jrudolph, line 295: Yes, maybe let's do at least that even if we use static data.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -147,7 +192,24 @@ private[http2] abstract class Http2Demux(http2Settings: Http2CommonSettings, ini
...
+            event match {
+              case _: StreamFrameEvent => state.onDataFrameSeen()
+              case _                   =>
+            }
+        }
+        frameOut.push(event)
+      }
+
+      val multiplexer = createMultiplexer(pushFrameOut, StreamPrioritizer.first())
+      frameOut.setHandler(multiplexer)
```
jrudolph, line 208: Hmm, this got a bit weird now, but I guess, I don't have any better idea...

jrudolph (commented):
Shaping up nicely, some comments below

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
johanandren: Yeah, good idea, was considering but could not decide.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
johanandren: Ok, good, that was my thinking but it was just gut feeling which is why I left the comment.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
johanandren: Yeah maybe that's better, and more efficient for actual intervals you'd want which may be more like 1 per minute or so.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerWithPings.scala
johanandren: This was mostly for manual checking with curl while developing, I think we can get rid of it completely now that there is some actual test coverage.

johanandren:
I think I addressed all the feedback comments except redoing timing to something more clever.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
ignasi35: Why do you think it's weird? My first impression was it looked cleaner. 🤔

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -63,6 +72,58 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
   def wrapTrailingHeaders(headers: ParsedHeadersFrame): Option[ByteString] = None
 }
 
+/**
+ * INTERNAL API
+ */
+@InternalApi
+private[http2] object ConfigurablePing {
+  case object Tick
+  val Ping = PingFrame(false, ByteString("abcdefgh"))
```
ignasi35, line 81: We can use hardcoded data because we require ping-timeout to be smaller than ping-interval. If/when we remove that restriction, using hardcoded `data` will break ping ACK's.
It'd be good to add a comment mentioning it.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -63,6 +72,58 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
...
+  object DisabledPingState extends PingState {
+    def onDataFrameSeen(): Unit = ()
+    def onPingAck(): Unit = ()
+    def onTick(): Unit = ()
+    def clear(): Unit = ()
+    def shouldEmitPing(): Boolean = false
+    def sendingPing(timestamp: Long): Unit = ()
+    def pingAckOverdue(timestampNanos: Long): Boolean = false
+  }
+  final class EnabledPingState(pingEveryNTickWithoutData: Long, maxKeepaliveTimeoutNanos: Long) extends PingState {
```
ignasi35, line 100:
```suggestion
  final class EnabledPingState(pingEveryNTickWithoutData: Long, pingTimeoutNanos: Long) extends PingState {
```

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -63,6 +72,58 @@ private[http2] class Http2ServerDemux(http2Settings: Http2CommonSettings, initia
...
+    def this(settings: Http2CommonSettings) = this(settings.pingInterval.toSeconds, settings.pingTimeout.toNanos)
```
ignasi35, line 101: Using `settings.pingInterval.toSeconds` directly works because the scheduler will send a ping per second. If/when we change that `this(settings.pingInterval.toSeconds...` will set the wrong value on `pingEveryNTickWithoutData`.

Maybe adding a constant on  the `case object Tick` to hold that `1.second` and a comment re this relation between schedule period, `pingEveryNTickWithoutData` and `settings.pingInterval` will be enough.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
ignasi35: Hmm, I realized there are more details on this implementation that rely on having only one ping in-flight. I guess a generic comment about it (not just focused on `data`) could be useful.

ignasi35 (commented):
Left a couple of comments.

Looking good.

ignasi35:
PLS BUILD

akka-http-core/src/main/mima-filters/10.2.1.backwards.excludes/http2-ping-support.excludes
```diff
@@ -0,0 +1,11 @@
...
+ProblemFilters.exclude[Problem]("akka.http.scaladsl.settings.Http2ClientSettings#Http2ClientSettingsImpl.*")
+ProblemFilters.exclude[Problem]("akka.http.scaladsl.settings.Http2ServerSettings#Http2ServerSettingsImpl.*")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.scaladsl.settings.Http2ServerSettings.pingInterval")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.scaladsl.settings.Http2ServerSettings.pingTimeout")
+ProblemFilters.exclude[InheritedNewAbstractMethodProblem]("akka.http.scaladsl.settings.Http2ServerSettings.pingInterval")
+ProblemFilters.exclude[InheritedNewAbstractMethodProblem]("akka.http.scaladsl.settings.Http2ServerSettings.pingTimeout")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.scaladsl.settings.Http2ClientSettings.pingInterval")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.scaladsl.settings.Http2ClientSettings.pingTimeout")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.scaladsl.settings.Http2CommonSettings.pingInterval")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.scaladsl.settings.Http2CommonSettings.pingTimeout")
```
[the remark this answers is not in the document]
johanandren: Yup, good catch

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
johanandren: Yes, simple 2.seconds wont work because the interval (and timeout) is from sending on the other side, not from receiving the frame. So in reality I rather had to do something like you describe Ignasi.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -256,12 +257,15 @@ private[http2] abstract class Http2Demux(http2Settings: Http2CommonSettings, ini
       setHandler(frameIn, new InHandler {
 
         def onPush(): Unit = {
-          grab(frameIn) match {
+          val frame = grab(frameIn)
+          frame match {
+            case _: PingFrame => // handle later
+            case _            => pingState.onDataFrameSeen()
+          }
```
ignasi35, line 264: Nice, less chatty. :-)

akka-http-core/src/main/scala/akka/http/scaladsl/settings/Http2ServerSettings.scala
```diff
@@ -26,6 +29,24 @@ private[http] trait Http2CommonSettings {
...
+/**
+ * INTERNAL API
+ */
+@InternalApi
+private[http] object Http2CommonSettings {
+  def validate(settings: Http2CommonSettings): Unit = {
+    import settings._
+    if (pingInterval.toSeconds > 0) {
+      require(pingInterval.toSeconds.seconds == pingInterval, s"ping-interval must be whole seconds, was $pingInterval")
+      require(pingTimeout == 0.seconds || pingTimeout < pingInterval, "ping-timeout must be zero, or positive and smaller than ping-interval")
```
ignasi35, line 46:
```suggestion
      require(pingTimeout == 0.seconds || (pingTimeout > 0.seconds && pingTimeout < pingInterval), "ping-timeout must be zero, or positive and smaller than ping-interval")
```
?

ignasi35: I think the case for `0s` disabling the timeout checks is not documented in `reference.conf` and would break the condition that only one PING can be in-flight at any given time.

ignasi35 (commented):
I get the impression from the implementation of `def pingAckOverdue` that setting `ping-timeout = 0s` would disable the timeout check?

Or is the validation code in the Settings class wrong and `ping-timeout = 0s` should not be a valid input?

akka-http-core/src/main/scala/akka/http/scaladsl/settings/Http2ServerSettings.scala
johanandren: I'm leaving that for now since I'm looking at implementing a non static tick interval which changes this a bit.

johanandren: It's tricky because as soon as we leave the possibility to send two pings without first seeing an ack, we don't know which ack we got back and the protocol requires the other side to respond etc. Given the value of the feature I'd prefer to not make this more complicated by having several outstanding pings and keeping track of them. In reality I think you will likely set a ping interval of a minute or higher and the responses will either come in milliseconds or not at all.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
jrudolph: Because multiplexer before "owned" the outlet but now it's more complex...

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
jrudolph: Still using `CANCEL`?

akka-http-core/src/main/scala/akka/http/javadsl/settings/Http2ClientSettings.scala
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+package akka.http.javadsl.settings
+
+import java.time.Duration
+
+import akka.http.scaladsl
+import scala.concurrent.duration._
+
+trait Http2ClientSettings { self: scaladsl.settings.Http2ClientSettings.Http2ClientSettingsImpl =>
```
jrudolph, line 12: Great, thanks for adding that.

akka-http-core/src/main/scala/akka/http/scaladsl/settings/Http2ServerSettings.scala
```diff
@@ -106,24 +138,30 @@ private[http] trait Http2InternalClientSettings
 
 @ApiMayChange
 @DoNotInherit
-trait Http2ClientSettings extends /*FIXME: javadsl.settings.Http2ClientSettings with*/ Http2CommonSettings { self: Http2ClientSettings.Http2ClientSettingsImpl =>
```
jrudolph, line 109: At least I documented my negligence...

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientServerSpec.scala
```diff
@@ -108,22 +109,24 @@ class Http2ClientServerSpec extends AkkaSpecWithMaterializer(
     }
   }
   class TestSetup {
+    def serverSettings: ServerSettings = ServerSettings(system)
```
jrudolph, line 112: :+1:

akka-http-core/src/main/scala/akka/http/scaladsl/settings/Http2ServerSettings.scala
```diff
@@ -26,6 +29,24 @@ private[http] trait Http2CommonSettings {
...
+ * INTERNAL API
+ */
+@InternalApi
+private[http] object Http2CommonSettings {
+  def validate(settings: Http2CommonSettings): Unit = {
+    import settings._
+    if (pingInterval > Duration.Zero && pingTimeout > Duration.Zero) {
+      require(
+        pingTimeout <= pingInterval && pingInterval.toMillis % pingTimeout.toMillis == 0,
+        s"ping-timeout must be less than and evenly divisible by the ping-interval ($pingInterval)")
```
jrudolph, line 47: I guess that's not too much of a restriction? Right now I guess we envision settings like 10 seconds for interval and 1 second for timeout. Might become weird if you want to set the timeout to very specific values (but maybe it's not necessary to worry about that?).

The description seems to be backwards, though?

```suggestion
        s"ping-timeout must be less than and a divisor of the ping-interval ($pingInterval)")
```

jrudolph (approved):
LGTM with a few comments. Maybe we should think about enabling pings by default with conservative settings? Or maybe only on the client side?

akka-http-core/src/main/scala/akka/http/scaladsl/settings/Http2ServerSettings.scala
johanandren: 🤦

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
johanandren: Could have sworn I fixed it already. :/ 

Will fix.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
johanandren: "What's the relation with the outlet and the multiplexer?"
"It's complicated."

johanandren:
I'm not sure what would be a good universal default, even for the client so I'd vote for opt-in only and not enabled by default.

ignasi35 (approved):

jrudolph:
> I'm not sure what would be a good universal default, even for the client so I'd vote for opt-in only and not enabled by default.

Agreed, let's keep it like this for now.

jrudolph:
Great, thanks @johanandren for trying out both of the implementation alternatives :) :fireworks:

PR #3619, opened 2020-11-11
Continue accepting window updates even when done sending

raboof:
Since those may come in asynchronously


jrudolph (approved):
LGTM (a bit weird to receive those on an `endStream = true` stream, but race condition is still there, of course)

raboof:
> LGTM (a bit weird to receive those on an `endStream = true` stream, but race condition is still there, of course)

good point, made it a bit more realistic

johanandren (approved):

PR #3621, opened 2020-11-12
http2: more robust handling of IncomingStreamBuffer.onPull

jrudolph:
Refs #3233

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
@@ -137,6 +137,10 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper 
   def pullNextFrame(streamId: Int, maxSize: Int): PullFrameResult =
     updateStateAndReturn(streamId, _.pullNextFrame(maxSize))
 
+  /** Entry-point to handle IncomingStreamBuffer.onPull through the state machine */
+  def incomingStreamPulled(streamId: Int): Unit =
+    updateState(streamId, _.incomingStreamPulled())
+
   private def updateAllStates(handle: StreamState => StreamState): Unit =
     streamStates.keys.foreach(updateState(_, handle))
 
@@ -257,6 +261,7 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper 
       }
 
     def pullNextFrame(maxSize: Int): (StreamState, PullFrameResult) = throw new IllegalStateException(s"pullNextFrame not supported in state $stateName")
+    def incomingStreamPulled(): StreamState = throw new IllegalStateException(s"incomingStreamPulled not supported in state $stateName")
 
     /** Called to cleanup any state when the connection is torn down */
     def shutdown(): Unit = ()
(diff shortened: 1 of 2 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -359,8 +364,8 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
             outstandingConnectionLevelWindow += windowSizeIncrement
           }
 
-          buffer.onDataFrame(d).getOrElse(
-            maybeFinishStream(d.endStream))
+          buffer.onDataFrame(d)
```
jrudolph, line 367: The main change is that, before, we moved out of the state receiving data when we received the frame with `END_STREAM` which is right if we watch the states from the outside. However, internally, the final frames might not yet have been dispatched in which case we would have dropped the buffer on the floor and leaked it in cases where the connection was aborted in this state.

Alternatively we could have spelled this out as another state (END_STREAM received but not yet all data dispatched), however, since we already have a state matrix with the product of the state of all the four ends of each stream (network and user in and out), this would have easily let to a multitude of new states. That's why I went with the slightly impure solution of extending the meaning of receiving states until after the user handler has actually pulled all the data.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -268,7 +268,7 @@ class Http2ServerSpec extends AkkaSpecWithMaterializer("""
       entityDataIn.ensureSubscription()
     }
 
-    abstract class WaitingForRequest(request: HttpRequest) extends RequestEntityTestSetup {
+    abstract class WaitingForRequest(request: HttpRequest) extends RequestEntityTestSetup with Serializable {
```
johanandren, line 271: What's up with this?

johanandren (approved):
LGTM (but not 100% I get it)

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
jrudolph: Oops, will remove. Was a test whether I can get rid of scalatest spamming the console with stack traces.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -473,31 +481,38 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
...
 
-      outstandingStreamWindow -= data.sizeInWindow
-      if (outstandingStreamWindow < 0) {
+    def onDataFrame(data: DataFrame): Unit =
+      if (wasClosed) {
         shutdown()
-        multiplexer.pushControlFrame(RstStreamFrame(streamId, ErrorCode.FLOW_CONTROL_ERROR))
-        // also close response delivery if that has already started
-        multiplexer.closeStream(streamId)
-        Some(Closed)
```
raboof, line 493: so the transition to `Closed` is no longer signaled from here.

How does it go into that state now? Is it an effect of `shutdown` failing the outlet?

raboof (approved):
Looks good. I didn't understand yet how the state machine goes to `Closed` now, but I guess this is via the outlet failure/completion?

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
jrudolph: It's now using `afterBufferEvent` consistently to figure out the next state after each call into `IncomingStreamBuffer` where before this return value was sometimes used and sometimes ignored.

raboof: right, which looks at `buffer.isDone` which checks `outlet.isClosed` :+1:

PR #3623, opened 2020-11-13
Test coverage for GOAWAY on mid-stream headers 

johanandren:
References #3268

Actual fix was implemented in trailing headers work (I think)

### DIFF
--- a/akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
+++ b/akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
@@ -141,13 +141,13 @@ class Http2ClientSpec extends AkkaSpecWithMaterializer("""
         response.entity shouldBe a[Chunked]
 
         sendDATA(streamId, endStream = false, ByteString("asdf"))
+        expectFrameFlagsStreamIdAndPayload(FrameType.WINDOW_UPDATE)
+
         sendHEADERS(streamId, endStream = false, Seq(RawHeader("X-mid-stream", "badvalue")))
 
-        // FIXME need something like the Http2ServerSpec.expectWindowUpdate() here
-        expectFrameFlagsStreamIdAndPayload(FrameType.WINDOW_UPDATE)
-        
-        val (_, error) = expectGOAWAY(streamId)
-        error should ===(ErrorCode.PROTOCOL_ERROR)
+        val goAwayFrame = expect[GoAwayFrame]()
+        goAwayFrame.errorCode should ===(ErrorCode.PROTOCOL_ERROR)
+        goAwayFrame.debug.utf8String should ===("Got unexpected mid-stream HEADERS frame")
       }
 
       "Three consecutive GET requests" in new SimpleRequestResponseRoundtripSetup {
--- a/akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
+++ b/akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
@@ -8,6 +8,7 @@ import akka.NotUsed
 import akka.event.Logging
 import akka.http.impl.engine.http2.FrameEvent._
 import akka.http.impl.engine.http2.Http2Protocol.ErrorCode
+import akka.http.impl.engine.http2.Http2Protocol.FrameType
 import akka.http.impl.engine.http2.Http2Protocol.SettingIdentifier
 import akka.http.impl.engine.ws.ByteStringSinkProbe
 import akka.http.impl.util.{ AkkaSpecWithMaterializer, LogByteStringTools }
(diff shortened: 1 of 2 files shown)


johanandren:
Now I actually noticed a `reject mid-stream HEADERS with endStream = false` test case in Http2ServerSpec so perhaps at least the server side was already covered?

jrudolph:
> Now I actually noticed a `reject mid-stream HEADERS with endStream = false` test case in Http2ServerSpec so perhaps at least the server side was already covered?

Yes, looked into that again, it seems that the test which I found weird, didn't fail because it never pulled the network to see the GOAWAY and we never close or fail the connection itself properly.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -336,6 +336,13 @@ class Http2ServerSpec extends AkkaSpecWithMaterializer("""
         // Wait to give the warning (that we hope not to see) time to pop up.
         Thread.sleep(100)
       }
+      "fail if headers are sent mid request stream" inAssertAllStagesStopped new WaitingForRequestData {
```
jrudolph, line 339: Maybe we can remove that one and keep the client one.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
```diff
@@ -126,6 +126,19 @@ class Http2ClientSpec extends AkkaSpecWithMaterializer("""
...
       }
 
+      "GOAWAY when the response has headers mid stream" in new SimpleRequestResponseRoundtripSetup {
+        val streamId = 0x1
+        emitRequest(streamId, HttpRequest(uri = "http://www.example.com/"))
+        expect[HeadersFrame]()
+
+        sendDATA(streamId, endStream = false, ByteString("wowdata"))
+        val headerBlock = encodeHeaders(Seq(RawHeader("X-Mid-Stream", "Wowsuchmid")))
+        sendFrame(HeadersFrame(streamId, endStream = false, endHeaders = true, headerBlock, None))
```
jrudolph, line 136: Why would that be the second one? Isn't that interpreted as the first HEADERS of the response?

johanandren: Yeah, I don't know what I was thinking here TBH, this doesn't test the right thing at all.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
```diff
@@ -126,6 +127,29 @@ class Http2ClientSpec extends AkkaSpecWithMaterializer("""
...
+        response.entity shouldBe a[Chunked]
+
+        sendDATA(streamId, endStream = false, ByteString("asdf"))
+        sendHEADERS(streamId, endStream = false, Seq(RawHeader("X-mid-stream", "badvalue")))
+
+        // FIXME need something like the Http2ServerSpec.expectWindowUpdate() here
+        expectFrameFlagsStreamIdAndPayload(FrameType.WINDOW_UPDATE)
+        
+        val (_, error) = expectGOAWAY(streamId)
+        error should ===(ErrorCode.PROTOCOL_ERROR)
```
johanandren, line 150: Also manually verified by failing this test and seeing that there is a debug text in the log mentioning mid stream headers in addition to the PROTOCOL_ERROR

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
```diff
@@ -126,6 +127,29 @@ class Http2ClientSpec extends AkkaSpecWithMaterializer("""
...
+        ))
+
+        val response = expectResponse()
+        response.entity shouldBe a[Chunked]
+
+        sendDATA(streamId, endStream = false, ByteString("asdf"))
+        sendHEADERS(streamId, endStream = false, Seq(RawHeader("X-mid-stream", "badvalue")))
+
+        // FIXME need something like the Http2ServerSpec.expectWindowUpdate() here
+        expectFrameFlagsStreamIdAndPayload(FrameType.WINDOW_UPDATE)
```
johanandren, line 147: Is there a better way to do this (less fragile)?

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
jrudolph: You could also use `expect[GoAwayFrame]()` if you want to assert on the debug string.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
jrudolph: It's triggered by `sendDATA` so you could at least pull it up a line. The infrastructure in Http2ServerSpec (`pollForWindowUpdates`) would probably also work here but we haven't pulled it out so far.

jrudolph (approved):
LGTM

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
johanandren: Ah, nice, that's good to make sure the test isn't accidentally failing from abusing the protocol in some other way, I'll change.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
johanandren: I'll leave it like this for now (but pulled it up next to the sendDATA).

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
```diff
@@ -126,6 +127,29 @@ class Http2ClientSpec extends AkkaSpecWithMaterializer("""
...
+        response.entity shouldBe a[Chunked]
+
+        sendDATA(streamId, endStream = false, ByteString("asdf"))
+        expectFrameFlagsStreamIdAndPayload(FrameType.WINDOW_UPDATE)
+
+        sendHEADERS(streamId, endStream = false, Seq(RawHeader("X-mid-stream", "badvalue")))
+
+        val goAwayFrame = expect[GoAwayFrame]()
+        goAwayFrame.errorCode should ===(ErrorCode.PROTOCOL_ERROR)
+        goAwayFrame.debug.utf8String should ===("Got unexpected mid-stream HEADERS frame")
```
jrudolph, line 150: :+1:

jrudolph (approved):
LGTM

raboof (approved):

PR #3624, opened 2020-11-13
docs: add example for Get request with query parameters

nikhilpatil:
This is quite a common requirement and nowhere mentioned in the docs explicitely. A question on stackoverflow (with 11k views) has an [answer](https://stackoverflow.com/a/31939776/154325) to this. But I believe inclusion of a simple example in official documentation will help developers a great deal.

References #xxxx

### DIFF
--- a/akka-http-core/src/test/java/akka/http/javadsl/model/UriTest.java
+++ b/akka-http-core/src/test/java/akka/http/javadsl/model/UriTest.java
@@ -65,6 +65,14 @@ public class UriTest extends JUnitSuite {
     //#dont-double-decode
   }
 
+  @Test
+  public void testAddingQueryParams() {
+    //#create-uri-with-query
+    Uri uri = Uri.create("http://foo.com").query(Query.create(Pair.create("foo","bar")));
+    //#create-uri-with-query
+    assertEquals(Optional.of("foo=bar"),uri.rawQueryString());
+  }
+
   //#illegal-scheme
   @Test(expected = IllegalUriException.class)
   public void testIllegalScheme() {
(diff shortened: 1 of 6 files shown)


lightbend-cla-validator:
Hi @nikhilpatil,

raboof (approved):
I like having more 'simple' examples early on.

This is already an improvement, so LGTM'ing. I think we could go a few steps further, though: perhaps here `Uri("https://akka.io?foo=bar")` would be nicer as a first step, and then linking to https://doc.akka.io/docs/akka-http/current/common/uri-model.html#query-string-in-uri and having this example over there?

Also it would be great to also have a Java version of this example

nikhilpatil:
> I like having more 'simple' examples early on.
> 
> This is already an improvement, so LGTM'ing. I think we could go a few steps further, though: perhaps here `Uri("https://akka.io?foo=bar")` would be nicer as a first step, and then linking to https://doc.akka.io/docs/akka-http/current/common/uri-model.html#query-string-in-uri and having this example over there?
> 
> Also it would be great to also have a Java version of this example

That makes sense. First part is straight forward. I will need help - especially with wording - for adding example to the query-string-in-uri section. I'll give it a shot though and will await comments on it 👍

lightbend-cla-validator:

lightbend-cla-validator:

nikhilpatil:
@raboof have tried to distribute examples over two pages as suggested. I couldn't put the original example (from the first commit) at https://doc.akka.io/docs/akka-http/current/common/uri-model.html#query-string-in-uri because that didn't seem like the right place to talk about `HttpRequest` (or `Get`). So just put a relevant example there, but If the reader starts from HttpRequest's page, I hope with the link in the "Note", they will get a clear picture about how to use the Uri and Query API with HttpRequest! 
Awaiting review comments, if any :)

raboof (approved):
Thanks, nice improvement!

jrudolph (approved):
LGTM

Issue #3625, opened 2020-11-16
Failed: Http2ClientSpec.The Http/2 client implementation should support stream support for receiving response entity data should support trailing headers for responses


johanandren:
https://jenkins.akka.io:8498/job/akka-http-nightly/AKKA_VERSION=master,Node=akka-http-nightly-node,SCALA_VERSION=2.13.3,jdk=AdoptOpenJDK%2011/327/testReport/junit/akka.http.impl.engine.http2/Http2ClientSpec/The_Http_2_client_implementation_should_support_stream_support_for_receiving_response_entity_data_should_support_trailing_headers_for_responses/

```
ScalaFutures$$anon$1$$anon$2: A timeout occurred waiting for a future to complete. Waited 150 milliseconds.
	at org.scalatest.concurrent.ScalaFutures$$anon$1.futureValueImpl(ScalaFutures.scala:339)
	at org.scalatest.concurrent.Futures$FutureConcept.futureValue(Futures.scala:476)
	at org.scalatest.concurrent.Futures$FutureConcept.futureValue$(Futures.scala:475)
	at org.scalatest.concurrent.ScalaFutures$$anon$1.futureValue(ScalaFutures.scala:281)
	at akka.http.impl.engine.http2.Http2ClientSpec$$anon$10.<init>(Http2ClientSpec.scala:307)
	at akka.http.impl.engine.http2.Http2ClientSpec.$anonfun$new$16(Http2ClientSpec.scala:295)
```

jrudolph:
Wonder if that was just too slow or related to https://github.com/akka/akka-http/pull/3605.

jrudolph:
Note that this is observed against Akka master where we observed timing differences before.

What happened here is that running the entity stream never seemed to pull the `IncomingStreamBuffer`. But maybe it's really just a timing issue...

Let's close this for now with the changes in #3628.

PR #3628, opened 2020-11-16
Increase patience in Http2ClientSpec

raboof:
To reduce timing-sensitivity of tests on CI. Might (or might
not) help for #3625.


johanandren (approved):

jrudolph (approved):
LGTM, won't hurt. This seems to be related to (or even fixed by) changes I proposed in #3621 (which is not yet merged) but I couldn't figure out how the current code would lead to the behavior described.

PR #3631, opened 2020-11-18
docs: release notes for 10.1.13

jrudolph:
After I reverted the big backport of fixing server-side cancellation races (88a6a75f3e59d11ef8bb03c73b0193b5ad4e78d6) because of it breaking the `lingerTimeout` there's actually not much left but a fix for the connection pool and an performance optimization for strict entities.

I'm seeing this as a good sign that maintenance on 10.1.x has naturally wound down.


raboof (approved):

ennru (approved):
LGTM.

PR #3638, opened 2020-11-18
http2 tests: Make 'network' and 'user' sides more explicit

raboof:
In Http2ServerSpec information flows in 4 directions:
frames from the network to the server, requests from the server
to the user code, responses from the user code to the server, and
frames from the server to the network.

That can get fairly hard to follow at times. I wonder if it helps
when we explicitly put the test methods in 'user' and 'network'
objects?

(if we like this we should also apply it to Http2ClientSpec of
course)

### DIFF
--- a/akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
+++ b/akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
@@ -442,8 +442,8 @@ class Http2ServerSpec extends AkkaSpecWithMaterializer("""
         network.sendDATA(TheStreamId, endStream = true, ByteString.empty)
 
         // DATA is left in IncomingStreamBuffer because we never pulled
-        toNet.cancel()
-        fromNet.sendError(new RuntimeException("connection crashed"))
+        network.toNet.cancel()
+        network.fromNet.sendError(new RuntimeException("connection crashed"))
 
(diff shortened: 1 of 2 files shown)


akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -73,11 +73,11 @@ class Http2ServerSpec extends AkkaSpecWithMaterializer("""
...
           expectedResponseHeaderBlock: ByteString
         ): Unit = {
-          sendHEADERS(streamId, endStream = true, endHeaders = true, requestHeaderBlock)
-          expectRequest() shouldBe expectedRequest
+          network.sendHEADERS(streamId, endStream = true, endHeaders = true, requestHeaderBlock)
+          user.expectRequest() shouldBe expectedRequest
 
-          responseOut.sendNext(response.addAttribute(Http2.streamId, streamId))
-          val headerPayload = expectHeaderBlock(streamId)
+          user.responseOut.sendNext(response.addAttribute(Http2.streamId, streamId))
```
jrudolph, line 79: Is that the same as `user.emitResponse`? Maybe that can be changed then as well.

jrudolph (approved):
LGTM, great idea, bit more verbose but helps being explicit in this case.

ignasi35 (approved):
love it!

jrudolph:
@ignasi35 please squash commits when merging if they are not meant to be conserved.

PR #3637, opened 2020-11-18
Move to travis-ci.com

ennru:
Travis nowadays operates on master builds only.

References https://github.com/akka/akka-meta/issues/143

PR #3645, opened 2020-11-19
CI: move env secrets

ennru:
...to the Travis GUI.

References https://github.com/akka/akka-http/pull/3637


jrudolph (approved):
I guess we don't know yet if it works but it seems to be setup correctly. If not we can still revert.

PR #3650, opened 2020-11-24
Update formFields example

kag0:
With the new `formFields` signature, parameters that extract nothing still need to be accounted for in the output. Code example is updated demonstrating that.
The discussion on magnets and signatures at the top of https://doc.akka.io/docs/akka-http/current/routing-dsl/directives/form-field-directives/formFields.html also probably needs to be updated, but I'm not sure what to replace it with.

### DIFF
--- a/docs/src/test/scala/docs/http/scaladsl/server/directives/FormFieldDirectivesExamplesSpec.scala
+++ b/docs/src/test/scala/docs/http/scaladsl/server/directives/FormFieldDirectivesExamplesSpec.scala
@@ -13,7 +13,7 @@ class FormFieldDirectivesExamplesSpec extends RoutingSpec with CompileOnlySpec {
   "formFields" in {
     //#formFields
     val route =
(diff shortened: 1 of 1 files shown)


docs/src/test/scala/docs/http/scaladsl/server/directives/FormFieldDirectivesExamplesSpec.scala
```diff
@@ -13,12 +13,12 @@ class FormFieldDirectivesExamplesSpec extends RoutingSpec with CompileOnlySpec {
   "formFields" in {
     //#formFields
     val route =
-      formFields("color", "age".as[Int]) { (color, age) =>
+      formFields("color", "age".as[Int], "direction" ! "up") { (color, age, _) =>
```
jrudolph, line 16: I moved it to the end. Having it at the front changed the result if the required field value was missing.

jrudolph (approved):
LGTM, good to have that in there.

raboof (approved):
:+1:

PR #3654, opened 2020-11-24
Verify signatures from the markdown actually exist

raboof:


johanandren (approved):

ignasi35 (approved):

docs/src/main/paradox/routing-dsl/directives/parameter-directives/parameter.md
```diff
@@ -1,13 +1,5 @@
 # parameter
 
-@@@ div { .group-scala }
```
jrudolph, line 3: I guess it wasn't super useful before and below we refer to the canonical `parameters`, so I guess that's good enough :+1:

Issue #3647, opened 2020-11-23
Fail substreams earlier in demuxer


johanandren:
The demuxer expect substreams would leak if not closed from `postStop` but they are now already failed by calling failStage, we'll have to fail them earlier if we want them failed with a custom exception with latest Akka master.

Original comment/PR:
https://github.com/akka/akka/issues/29823#issuecomment-732083018_

PR #3651, opened 2020-11-24
tests: any exception is fine in "close substreams" test

jrudolph:
To fix nightly failure here: https://jenkins.akka.io:8498/view/Akka-http/job/akka-http-nightly-release-10.1/AKKA_VERSION=master,Node=akka-http-nightly-node,SCALA_VERSION=2.12.10,jdk=AdoptOpenJDK%208/1089/testReport/junit/akka.http.impl.engine.http2/Http2ServerSpec/The_Http_2_server_implementation_should_support_multiple_concurrent_substreams_should_close_substreams_when_connection_is_shutting_down/

(Behavior will be slightly different between latest Akka 2.6 and previous versions but we don't care to much there.)

PR #3656, opened 2020-11-24
Close the substreams earlier from the demuxer

johanandren:
Refs #3647

To emit the custom failure before auto-substream-cancelation kicks in.

Needed to pass tests with Akka master after akka/akka#29823 was merged.

Manually ran tests against master (won't show in PR validation)


akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -129,6 +129,12 @@ private[http2] abstract class Http2Demux(http2Settings: Http2CommonSettings, ini
 
       override protected def logSource: Class[_] = if (isServer) classOf[Http2ServerDemux] else classOf[Http2ClientDemux]
 
+      def frameOutFinished(): Unit = {
+        // make sure we clean up/fail substreams with a custom failure before stage is canceled
+        // and substream autoclean kicks in
+        shutdownStreamHandling()
```
johanandren, line 135: `Http2Multiplexer` does not know that `this` is also a `Http2StreamHandling` so that's why we need a call back out here to call `shutdownStreamHandling()`

raboof (approved):
:+1:

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Multiplexer.scala
```diff
@@ -111,6 +114,11 @@ private[http2] trait Http2MultiplexerSupport { logic: GraphStageLogic with Stage
       /** Network pulls in new frames */
       def onPull(): Unit = updateState(_.onPull())
 
+      override def onDownstreamFinish(): Unit = {
+        frameOutFinished()
+        super.onDownstreamFinish()
```
johanandren, line 119: This will trigger cancelStage which now also cancels all substreams with the given cause, and we want our custom `Http2StreamHandling.ConnectionWasAbortedException` as cause for the user stream.

raboof (approved):
should/can we also revert #3651 as part of this PR?

johanandren:
That was against 10.1 while this is against master, Johannes stated it is ok to just adapt the test for 10.1

jrudolph (approved):
LGTM, interesting that it failed this way. We are probably missing other cases this way that are currently not covered but let's merge this to fix the nightlies.

Issue #3229, opened 2020-06-09
HTTP/2 client: review rendering parsing


jrudolph:
And see if things can be DRY'd with server implementations

johanandren:
Hmm, can't see any obvious repetition in the rendering and the parsing. Am I missing something?

jrudolph:
> Hmm, can't see any obvious repetition in the rendering and the parsing. Am I missing something?

Rendering has somewhat duplicated logic wrt `content-length` and substream creation. Maybe that was the main thing I had in mind.

ignasi35:
@jrudolph, with #3520 merged is there anything else pending for this issue?

jrudolph:
These four lines could still be dry'd up between client and server:

```
      addContentHeaders(headerPairs, response.entity)
      renderHeaders(response.headers, headerPairs, serverHeader, log, isServer = true)

      val headers = ParsedHeadersFrame(streamId, endStream = response.entity.isKnownEmpty, headerPairs.result(), None)
      substreamFor(response.entity, headers)
```

but it's somewhat low priority do go this step as well.

ignasi35:
> These four lines could still be dry'd up between client and server:

Done in #3657 :-)

PR #3657, opened 2020-11-24
http2: extract common rendering + add userAgent on requests

ignasi35:
References #3229 (and https://github.com/akka/akka-http/issues/3229#issuecomment-713436683 in particular)


ignasi35:
this PR extracts code from `ResponseRendering` into the new `CommonRendering` object and introduces a `CommonRendering` trait that can be instantiated for requests or responses.

I think there's still rooom for naming reviews and moving code around. For example, I'm not super happy with the fact that `CommonRendering` lives in the `ResponseRendering.scala` file. I almost moved `RequestRendering`, `ResponseRendering` and `CommonRendering` into a new file so all three lived closer together but is required moving `RequestRendering` into a different package. It's private and it's probably fine. I may do it on a new commit.

I may also improve the code a bit and avoid using anonymous subclasses of the `trait CommonRendering` and, instead, create a better abstraction.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
```diff
@@ -103,18 +103,18 @@ private[http] object Http2Blueprint {
...
       framingClient(log) atop
       idleTimeoutIfConfigured(settings.idleTimeout)
   }
 
-  def httpLayerClient(masterHttpHeaderParser: HttpHeaderParser, log: LoggingAdapter): BidiFlow[HttpRequest, Http2SubStream, Http2SubStream, HttpResponse, NotUsed] = {
+  def httpLayerClient(masterHttpHeaderParser: HttpHeaderParser, settings: ClientConnectionSettings, log: LoggingAdapter): BidiFlow[HttpRequest, Http2SubStream, Http2SubStream, HttpResponse, NotUsed] = {
     BidiFlow.fromFlows(
       Flow[HttpRequest].statefulMapConcat { () =>
-        val renderer = RequestRendering.createRenderer(log)
+        val renderer = RequestRendering.createRenderer(settings, log)
```
ignasi35, line 117: Passing in the settings so the renderer can add a `User-Agent` header. It also makes the signature more similar to the server counterpart.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/ResponseRendering.scala
```diff
@@ -71,28 +93,37 @@ private[http2] object ResponseRendering {
...
+  /**
+   * Mutates `headerPairs` adding any valid header from `headersSeq`.
+   * @param peerIdHeader a header providing extra information (e.g. vendor and version) about the
+   *                     peer. For example, a User-Agent on the client or a Server header on the server.
+   */
   private[http2] def renderHeaders(
     headersSeq:   immutable.Seq[HttpHeader],
     headerPairs:  VectorBuilder[(String, String)],
-    serverHeader: Option[(String, String)],
+    peerIdHeader: Option[(String, String)],
```
ignasi35, line 104: renamed from `server` to `peerId` since that may be used for `Server` or `User-Agent` now.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/ResponseRendering.scala
```diff
@@ -71,28 +93,37 @@ private[http2] object ResponseRendering {
...
         header match {
-          case x: Server =>
+          case x: Server if isServer =>
+            addHeader(x)
+            peerIdSeen = true
+
+          case x: `User-Agent` if !isServer =>
             addHeader(x)
-            serverSeen = true
+            peerIdSeen = true
```
ignasi35, line 126: Meh

akka-http-core/src/main/scala/akka/http/impl/engine/http2/ResponseRendering.scala
```diff
@@ -5,17 +5,61 @@
...
+    CommonRendering.renderHeaders(r.headers, headerPairs, peerIdHeader, log, isServer = r.isResponse)
+
+    val headersFrame = ParsedHeadersFrame(nextStreamId(r), endStream = r.entity.isKnownEmpty, headerPairs.result(), None)
+
+    Http2SubStream(r.entity, headersFrame, r.attributes.filter(_._2.isInstanceOf[RequestResponseAssociation]))
+  }
+}
+
+private[http2] object CommonRendering {
+
```
ignasi35, line 62: I first moved a few methods and fields into the new `CommonRendering` and later created the `CommonRendering` trait.

I think the current names for the object and the trait are not grat but keeping some code on the `object` came in quite handy on tests.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
```diff
@@ -50,6 +52,16 @@ class Http2ClientSpec extends AkkaSpecWithMaterializer("""
...
       abstract class SimpleRequestResponseRoundtripSetup extends TestSetup with NetProbes {
+
+        val defaultExpectedHeaders = Seq(
+          ":method" -> "GET",
+          ":scheme" -> "https",
+          ":authority" -> "www.example.com",
+          ":path" -> "/",
+          "content-length" -> "0",
+          "user-agent" -> ClientConnectionSettings(system).userAgentHeader.get.value
+        )
```
ignasi35, line 63: I've extracted this block (copy/pasted multiple times) to add the `User-Agent` header (which requires a system to load the settings).

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/ResponseRenderingSpec.scala
```diff
@@ -69,14 +69,14 @@ class ResponseRenderingSpec extends AnyWordSpec with Matchers {
 
     "add server header if default provided" in {
       val builder = new VectorBuilder[(String, String)]
-      ResponseRendering.renderHeaders(Seq.empty, builder, Some(("server", "default server")), NoLogging, isServer = true)
+      CommonRendering.renderHeaders(Seq.empty, builder, Some(("server", "default server")), NoLogging, isServer = true)
```
ignasi35, line 72: Oh, I probably should add a test for the `User-Agent`...

ignasi35 (commented):
notes to self and reviewers.

raboof:
> I think there's still rooom for naming reviews and moving code around.

I think having classes in files that don't match their name (unless it's a 'private' class only used within that file) makes the project harder to navigate, I'd prefer to stick to the file-per-class convention.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/HttpMessageRendering.scala
```diff
@@ -0,0 +1,188 @@
...
+    CommonRendering.addContentHeaders(headerPairs, r.entity)
+    CommonRendering.renderHeaders(r.headers, headerPairs, peerIdHeader, log, isServer = r.isResponse)
+
+    val headersFrame = ParsedHeadersFrame(nextStreamId(r), endStream = r.entity.isKnownEmpty, headerPairs.result(), None)
+
+    Http2SubStream(r.entity, headersFrame, r.attributes.filter(_._2.isInstanceOf[RequestResponseAssociation]))
+  }
+}
+
+private[http2] object CommonRendering {
```
raboof, line 79: perhaps `MessageRendering` (since `HttpRequest` and `HttpResponse` both extend `HttpMessage`?

akka-http-core/src/main/scala/akka/http/impl/engine/http2/HttpMessageRendering.scala
```diff
@@ -0,0 +1,188 @@
...
+    headerPairs += ":scheme" -> request.uri.scheme
+    headerPairs += ":authority" -> request.uri.authority.toString
+    headerPairs += ":path" -> request.uri.toHttpRequestTargetOriginForm.toString
+    headerPairs
+  }
+
+  override lazy val peerIdHeader: Option[(String, String)] = settings.userAgentHeader.map(h => h.lowercaseName -> h.value)
+}
+
+abstract class RenderFactory[R <: HttpMessage](log: LoggingAdapter) {
```
raboof, line 60: Perhaps this could *also* be `MessageRendering`, so the current `object CommonRendering` becomes the companion object to this class? Seems to make sense, as it contains the static helpers used (among other places) in instances of this class?

akka-http-core/src/main/scala/akka/http/impl/engine/http2/HttpMessageRendering.scala
```diff
@@ -0,0 +1,188 @@
...
+
+  def nextStreamId(r: R): Int
+  def initialHeaderPairs(r: R): VectorBuilder[(String, String)]
+  def peerIdHeader: Option[(String, String)]
+
+  val renderer: R => Http2SubStream = { (r: R) =>
```
raboof, line 66: we could also make this a `def render(message: R): Http2SubStream` and update the call sites accordingly - or would that have adverse effects?

jrudolph (commented):
I added a PR with a few cosmetic things, part of which were also annotated by @raboof: https://github.com/ignasi35/akka-http/pull/1

akka-http-core/src/main/scala/akka/http/impl/engine/http2/HttpMessageRendering.scala
jrudolph: In https://github.com/ignasi35/akka-http/pull/1 I made the whole class a `R => Http2SubStream` but your suggestion would also work.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/CommonRenderingSpec.scala
```diff
@@ -36,7 +36,7 @@ class ResponseRenderingSpec extends AnyWordSpec with Matchers {
         RawHeader("raw", "whatever"),
         new MyCustomHeader("whatever", renderInResponses = true)
       )
-      ResponseRendering.renderHeaders(headers, builder, None, NoLogging, isServer = true)
+      CommonRendering.renderHeaders(headers, builder, None, NoLogging, isServer = true)
```
jrudolph, line 39: Can that call be abstracted out? Seems to be mostly the same everywhere.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/CommonRenderingSpec.scala
```diff
@@ -59,28 +59,42 @@ class ResponseRenderingSpec extends AnyWordSpec with Matchers {
...
     }
 
+    "add user-agent header if default provided (in client mode)" in {
+      val builder = new VectorBuilder[(String, String)]
+      CommonRendering.renderHeaders(Seq.empty, builder, Some(("user-agent", "fancy browser")), NoLogging, isServer = false)
+      val result = builder.result().find(_._1 == "user-agent").map(_._2)
+      result shouldEqual Some("fancy browser")
+    }
+
+    "keep user-agent header if explicitly provided (in server mode)" in {
```
jrudolph, line 91: Copy paste error?

```suggestion
    "keep user-agent header if explicitly provided (in client mode)" in {
```

akka-http-core/src/main/scala/akka/http/impl/engine/http2/HttpMessageRendering.scala
```diff
@@ -0,0 +1,188 @@
...
+          case x: Connection =>
+            suppressionWarning(x, "`Connection` header is not allowed for HTTP/2")
+
+          case x =>
+            addHeader(x)
+        }
+      }
+    }
+
+    if (!dateSeen) {
```
jrudolph, line 175: The `Date` header should only be added to responses, how comes it didn't fail any tests?

jrudolph (commented):
A few more comments

akka-http-core/src/main/scala/akka/http/impl/engine/http2/HttpMessageRendering.scala
jrudolph: Seems to be copy pasta in Http2ClientSpec where the `Date` header is filtered out which is convinient on the server side but shouldn't be done on the client side.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/CommonRenderingSpec.scala
ignasi35: I don't think we can abstract it away. Each test invokes `renderHeaders` with slightly different arguments.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/HttpMessageRendering.scala
ignasi35: I've added a new unit test for `Date` on requests in `HttpMessageRenderingSpec`.

ignasi35:
Thanks @raboof and @jrudolph for the reviews. I went with the suggestion to extend `(R => Http2SubsStream)`. I think that's one less allocation (not a big deal) _and_, more importantly, I think code looks cleaner.

johanandren:
Good to include the TLS header and attributes in the common logic as well if possible, those are practically identical now: https://github.com/akka/akka-http/pull/3668#discussion_r530859737

ignasi35:
> Good to include the TLS header and attributes in the common logic as well if possible, those are practically identical now: [#3668 (comment)](https://github.com/akka/akka-http/pull/3668#discussion_r530859737)

I'd rather do that and its set of tests on a separate PR.

ignasi35:
> I'd rather do that and its set of tests on a separate PR.

Scratch that. 

This PR only addresses code duplication on rendering (not on parsing).

raboof (approved):
LGTM! (2 optional suggestions)

jrudolph (approved):
LGTM

PR #3666, opened 2020-11-25
[Preview] http2: implement backpressure on the user-handler side

jrudolph:
Especially on the client-side, before this change there would be no
backpressure between responseOut and requestIn, i.e. when request would be
done as observed on the network (resposne was received and dispatched), a
new request would be immediately dispatched. In the worst case of slow (or
no) processing on the response side, we would accumulate more and more ready
responses in the bufferedSubstreamOut.

In practice, it might be less significant as many responses would be backpressured
themselves on response entity data.

One problem with the current solution is that the backpressue is based on the peer-supplied
`SETTING_MAX_CONCURRENT_STREAMS`, so it is still the peer which can control how many outgoing
responses to buffer.

I guess what we should do instead is to make that buffering configurable:

 * either by overloading the `max-concurrent-stream` setting on the client to mean outgoing streams (instead of server-pushed requests which isn't supported anyway currently)
 * or by introducing another setting with that semantics

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
@@ -63,17 +63,17 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper 
    */
   def lastStreamId(): Int = largestIncomingStreamId
 
-  private var maxConcurrentStreams = Http2Protocol.InitialMaxConcurrentStreams
+  protected var maxConcurrentStreams = Http2Protocol.InitialMaxConcurrentStreams
   def setMaxConcurrentStreams(newValue: Int): Unit = maxConcurrentStreams = newValue
   /**
    * @return true if the number of outgoing Active streams (Active includes Open
-   *         and any variant of HalfClosedXxx) doesn't exceed MaxConcurrentStreams
+   *         and any variant of HalfClosedXxx) plus any outgoing streams that haven't yet fetched
+   *         doesn't exceed MaxConcurrentStreams
    */
-  def hasCapacityToCreateStreams: Boolean = {
+  def hasCapacityToCreateStreams(numBufferedOut: Int): Boolean =
     // StreamStates only contains streams in active states (active states are any variation
     // of Open, HalfClosed) so using the `size` works fine to compute the capacity
-    activeStreamCount() < maxConcurrentStreams
-  }
+    numBufferedOut + activeStreamCount() < maxConcurrentStreams
 
   private def streamFor(streamId: Int): StreamState =
     streamStates.get(streamId) match {
@@ -115,7 +115,6 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper 
     } else
       // stream was cancelled by peer before our response was ready
       stream.data.runWith(Sink.cancelled)(subFusingMaterializer)
-
   }
 
   // Called by the outgoing stream multiplexer when that side of the stream is ended.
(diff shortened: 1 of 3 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -63,17 +63,17 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
...
+   *         and any variant of HalfClosedXxx) plus any outgoing streams that haven't yet fetched
+   *         doesn't exceed MaxConcurrentStreams
    */
-  def hasCapacityToCreateStreams: Boolean = {
+  def hasCapacityToCreateStreams(numBufferedOut: Int): Boolean =
     // StreamStates only contains streams in active states (active states are any variation
     // of Open, HalfClosed) so using the `size` works fine to compute the capacity
-    activeStreamCount() < maxConcurrentStreams
-  }
+    numBufferedOut + activeStreamCount() < maxConcurrentStreams
```
raboof, line 76: makes sense to count those :+1:

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
```diff
@@ -552,6 +569,36 @@ class Http2ClientSpec extends AkkaSpecWithMaterializer("""
...
+
+      def sendOneRequest(streamId: Int): Unit = {
+        user.emitRequest(HttpRequest())
+        network.expectDecodedHEADERS(streamId, endStream = true)
+      }
+
+      user.responseIn.ensureSubscription()
+
+      (1 to 5).foreach(i => sendOneRequest(i * 2 - 1))
+      user.emitRequest(HttpRequest()) // one extra is requested (some buffering?) but not send on the wire
```
raboof, line 585:
```suggestion
      user.emitRequest(HttpRequest()) // one extra is requested (some buffering?) but not sent on the wire
```

raboof (commented):
> I guess what we should do instead is to make that buffering configurable:
> 
> * either by overloading the max-concurrent-stream setting on the client to mean outgoing streams (instead of server-pushed requests which isn't supported anyway currently)
> * or by introducing another setting with that semantics

Also, SETTING_MAX_CONCURRENT_STREAMS is per-connection, but we might also want to have a maximum that spans all connections?

jrudolph:
> > I guess what we should do instead is to make that buffering configurable:
> > 
> > * either by overloading the max-concurrent-stream setting on the client to mean outgoing streams (instead of server-pushed requests which isn't supported anyway currently)
> > * or by introducing another setting with that semantics
> 
> Also, SETTING_MAX_CONCURRENT_STREAMS is per-connection, but we might also want to have a maximum that spans all connections?

You mean per host? That sounds like a feature that might be part of the load-balancer part of the implementation. Without the load balancer I would imagine there would only ever be one open connection to one host.

johanandren (commented):
Looks like it makes sense.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
```diff
@@ -552,6 +569,36 @@ class Http2ClientSpec extends AkkaSpecWithMaterializer("""
...
+      network.expectNoBytes(100.millis)
+      user.requestOut.pending shouldBe 0
+      user.requestOut.expectNoMessage(100.millis) // no further request expected
+
+      network.sendHEADERS(3, endStream = false, Seq(RawHeader(":status", "404")))
+      network.expectNoBytes(100.millis)
+      network.sendDATA(3, endStream = true, ByteString.empty)
+      network.expectWindowUpdate()
```
ignasi35, line 593: Hmmm, where does this WINDOW_UPDATE come from?
I'm guessing it was held back until `3` is closed in the previous line?
Can we add a comment with extra info about that `network.expectWindowUpdate`? Thanks :)

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
```diff
@@ -552,6 +569,36 @@ class Http2ClientSpec extends AkkaSpecWithMaterializer("""
...
+
+      user.requestOut.pending shouldBe 0
+      user.requestOut.expectNoMessage(100.millis)
+
+      user.expectResponse() // now we fetch the first response
+      network.expectDecodedHEADERS(11, endStream = true) // expect request 6 to be sent
```
ignasi35, line 599:
```suggestion
      network.expectDecodedHEADERS(11, endStream = true) // expect request 6 (aka streamId==11) to be sent
```

PR #3671, opened 2020-11-26
http2: handle RST while sending data + more tests

raboof:


akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -329,9 +332,13 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
...
     override def handle(event: StreamFrameEvent): StreamState = event match {
-      case _: ParsedHeadersFrame => expectIncomingStream(event, HalfClosedRemoteSendingData(outStream), Open(_, outStream), correlationAttributes)
-      case w: WindowUpdateFrame  => handleWindowUpdate(w)
-      case _                     => receivedUnexpectedFrame(event)
+      case _: ParsedHeadersFrame =>
+        expectIncomingStream(event, Some(outStream), HalfClosedRemoteSendingData(outStream), Open(_, outStream), correlationAttributes)
+      case w: WindowUpdateFrame =>
+        handleWindowUpdate(w)
+      case _ =>
+        outStream.cancelStream()
```
johanandren, line 340: `receivedUnexpectedFrame` tries to push GOAWAY, but that is for entire connection so no problem with canceling the outstream first right?

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
```diff
@@ -409,6 +409,47 @@ class Http2ClientSpec extends AkkaSpecWithMaterializer("""
       "fail entity stream if advertised content-length doesn't match" in pending
     }
 
+    "support stream support for sending request entity data" should {
```
johanandren, line 412: support stream support stream support

johanandren (approved):

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -256,7 +257,9 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
           dispatchSubstream(frame, data, correlationAttributes)
           nextState
 
-        case x => receivedUnexpectedFrame(x)
+        case x =>
```
jrudolph, line 260: Would be good to have tests for each of these added cases, especially since it seems to apply both to client and server and also to incoming and outgoing streams.

jrudolph (commented):
I wonder if we should deal with `receivedUnexpectedFrame` differently in the first place. Could we make it an exception that is dealt with in `updateState`?  So we don't have to sprinkle that kind of logic all over the place?

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
raboof: Hmm, yeah - I'll roll this back for now and add it again in one or more separate PRs with specific tests.

johanandren (approved):

jrudolph (approved):
LGTM

PR #3672, opened 2020-11-26
http2: 'pull' was lost when closing while waiting for window

raboof:
~on top of #3671~ (rebased)

### DIFF
--- a/akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
+++ b/akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
@@ -452,6 +452,28 @@ class Http2ClientSpec extends AkkaSpecWithMaterializer("""
 
         connectionShouldStillBeUsable()
       }
+      "handle RST_STREAM while waiting for a window update" in new WaitingForRequestData {
+        val entitySize = 70000
+        entityDataOut.sendNext(ByteString(Array.fill[Byte](entitySize)(0x23))) // 70000 > Http2Protocol.InitialWindowSize
+        network.sendWINDOW_UPDATE(TheStreamId, 10000) // enough window for the stream but not for the window
+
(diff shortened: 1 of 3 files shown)


johanandren (approved):

jrudolph:
Needs a rebase

jrudolph:
Seemed it was easy enough to merge/resolve conflicts.

jrudolph (approved):
LGTM, great catch, so the effect was that the connection was somewhat stalled after that happened?

jrudolph:
Hmm not sure the new scenario is done yet? Should we just merge the original fix or is there a problem with it?

raboof:
> LGTM, great catch, so the effect was that the connection was somewhat stalled after that happened?

Yes, the test added in this PR would time out instead of continue (though I don't remember exactly at which step)

raboof:
> Hmm not sure the new scenario is done yet? Should we just merge the original fix or is there a problem with it?

I think we need additional fixes from #3673 to complete this scenario (so it would be good to merge this and follow-up with that), but I'll double-check

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
```diff
@@ -452,6 +452,60 @@ class Http2ClientSpec extends AkkaSpecWithMaterializer("""
...
+        entityDataOut.expectCancellation()
+        network.expectGOAWAY(otherRequestStreamId)
+
+        // make sure the demuxer also moved back from WaitingForConnectionWindow to Idle
+        network.sendWINDOW_UPDATE(0, 10000)
+        network.expectNoBytes(100.millis) // don't expect anything, stream has been cancelled in the meantime
+
+        // TODO the client stack should not accept new connections anymore
+        // TODO what to do with requests that are already in the queue at this point?
+        // user.requestOut.expectCancellation()
```
raboof, line 503: This indeed needs further changes, which are coming in #3673

Issue #3227, opened 2020-06-09
HTTP/2 client: more complete test suite


jrudolph:
List potential tests here:

 * [ ] more multiplexing interactions (request and response streams)
 * [ ] failure conditions

PR #3676, opened 2020-12-01
Test coverage for request HTTP2 client entity stream failing

johanandren:
References #3227

### DIFF
--- a/akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
+++ b/akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
@@ -512,9 +512,7 @@ class Http2ClientSpec extends AkkaSpecWithMaterializer("""
       abstract class StreamingRequestSent extends TestSetup {
         val streamId = 0x1
         val requestStream = TestPublisher.probe[ByteString]()
-        user.emitRequest(HttpRequest(
-          protocol = HttpProtocols.`HTTP/2.0`,
-          entity = HttpEntity(ContentTypes.`application/octet-stream`, Source.fromPublisher(requestStream))))
+        user.emitRequest(HttpRequest(entity = HttpEntity(ContentTypes.`application/octet-stream`, Source.fromPublisher(requestStream))))
         network.expectDecodedHEADERS(streamId, endStream = false)
         requestStream.expectRequest()
       }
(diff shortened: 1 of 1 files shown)


akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
```diff
@@ -411,6 +412,34 @@ class Http2ClientSpec extends AkkaSpecWithMaterializer("""
...
+    "support stream for request data" should {
+      abstract class StreamingRequestSent extends TestSetup {
+        val streamId = 0x1
+        val requestStream = TestPublisher.probe[ByteString]()
+        user.emitRequest(HttpRequest(
+          protocol = HttpProtocols.`HTTP/2.0`,
+          entity = HttpEntity(ContentTypes.`application/octet-stream`, Source.fromPublisher(requestStream))))
+        network.expectDecodedHEADERS(streamId, endStream = false)
+        requestStream.expectRequest()
+      }
```
johanandren, line 424: Kind of duplicating things in #3671 here but the actual test case is not covered by that PR from what I can see.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
```diff
@@ -411,6 +412,34 @@ class Http2ClientSpec extends AkkaSpecWithMaterializer("""
       "fail entity stream if advertised content-length doesn't match" in pending
     }
 
+    "support stream for request data" should {
+      abstract class StreamingRequestSent extends TestSetup {
+        val streamId = 0x1
+        val requestStream = TestPublisher.probe[ByteString]()
+        user.emitRequest(HttpRequest(
+          protocol = HttpProtocols.`HTTP/2.0`,
```
raboof, line 420: I don't think setting the protocol is needed

raboof (approved):

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
johanandren: Copy pasta from other test where it perhaps makes more sense :) I'll remove.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
jrudolph: Would those also be useful for the server side?

jrudolph (approved):
LGTM

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
johanandren: Yes, good point. 

I thought they were but now that I look at it only `send RST_STREAM when entity data stream fails` is there, which is about the response entity only.

raboof:
(needs a rebase)

Issue #3677, opened 2020-12-01
Flow based server handler support for HTTP/2


johanandren:
Stepping through the logic from ServerBuilder I noticed that it only actually binds a HTTP/2 server for the `Req => Future[Res]` `bind`,  the `Flow[Req, Res]` variation ends up always binding a HTTP/1.1 server.

Not sure if this is a priority in any way, but if we don't intend to support it I think it should at least emit some form of warning when `settings.previewServerSettings.enableHttp2` is true?

johanandren:
Noticed now the docs actually say so here: https://doc.akka.io/docs/akka-http/current/server-side/http2.html#server-side-http-2-preview-

jrudolph:
There should be previous tickets on this situation. IIRC, so far, we concluded that we don't think it's worth it right now.

johanandren:
Found https://github.com/akka/akka-http/issues/3627 and https://github.com/akka/akka-http/issues/246#issuecomment-712098233 which I missed when creating this. I guess we can close this as a duplicate then.

PR #3678, opened 2020-12-01
Fail if trying to bind flow/source with HTTP2 

johanandren:
Instead of silently ignoring and binding HTTP/1.1 

Refs #3677

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
@@ -180,6 +180,9 @@ class HttpExt private[http] (private val config: Config)(implicit val system: Ex
            connectionContext: ConnectionContext = defaultServerHttpContext,
            settings:          ServerSettings    = ServerSettings(system),
            log:               LoggingAdapter    = system.log): Source[Http.IncomingConnection, Future[ServerBinding]] = {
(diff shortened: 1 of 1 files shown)


raboof (commented):
Hmm, this might be too harsh - you could have an application that supports HTTP/2 on one port with `.bind()` and has a `.bindFlow()`-based interface on another (HTTP/1-only) port...

johanandren:
Yeah, maybe. Don't you think that the intention if you have it enabled in the settings is to use it though and want to notice as early as possible?

johanandren:
Should it be a warning or an error?

jrudolph:
> Yeah, maybe. Don't you think that the intention if you have it enabled in the settings is to use it though and that you'd want to notice as early as possible?

That's an interesting point, you can always enable and disable it per binding, so failing wouldn't make some usages impossible but only a bit more awkward.

ignasi35:
I'm on the fence on this one. 
I like it to crash early but the use-case @raboof describes is common (early adopting gRPC comes to mind).

I think we can exploit the fact that this new code is only on deprecated methods so we may rewrite the message adding an alternative?

jrudolph (approved):
LGTM, this is safe to do in any case

raboof (approved):

PR #3689, opened 2020-12-03
h2: close down connections after sending or receiving GOAWAY

raboof:
But not before processing remaining in-flight requests.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
@@ -220,10 +220,21 @@ private[http2] abstract class Http2Demux(http2Settings: Http2CommonSettings, ini
 
       override protected def logSource: Class[_] = if (isServer) classOf[Http2ServerDemux] else classOf[Http2ClientDemux]
 
-      def frameOutFinished(): Unit = {
+      override def frameOutFinished(): Unit = {
         // make sure we clean up/fail substreams with a custom failure before stage is canceled
         // and substream autoclean kicks in
         shutdownStreamHandling()
+      }
+
+      override def frameOutClosed(): Unit = {
+        complete(frameOut)
+        if (isServer) {
+          // TODO #3687 just cancelling the substream input should be sufficient
+          //  (since the other inputs/outputs are cancelled/completed elsewhere),
+          //  but tests show the stage is leaked unless we complete the stage here:
+          // cancel(substreamIn)
+          completeStage()
+        }
       }
 
       override def pushFrameOut(event: FrameEvent): Unit = {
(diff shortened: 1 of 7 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
```diff
@@ -324,7 +341,9 @@ private[http2] abstract class Http2Demux(http2Settings: Http2CommonSettings, ini
               }
             case PingFrame(false, data) =>
               multiplexer.pushControlFrame(PingFrame(ack = true, data))
-
+            case GoAwayFrame(lastStreamId, _, _) =>
+              // TODO do we log the error code and debug message on some level?
```
johanandren, line 345: Doesn't it surface in the user level stream as a failure somehow?

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
```diff
@@ -75,7 +76,7 @@ class Http2ClientSpec extends AkkaSpecWithMaterializer("""
         }
       }
 
-      "GET request in one HEADERS frame" in new SimpleRequestResponseRoundtripSetup {
+      "GET request in one HEADERS frame" inAssertAllStagesStopped new SimpleRequestResponseRoundtripSetup {
```
johanandren, line 79: 👍

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -1185,7 +1185,11 @@ class Http2ServerSpec extends AkkaSpecWithMaterializer("""
 
       "reject other frame than HEADERS/PUSH_PROMISE in idle state with connection-level PROTOCOL_ERROR (5.1)" inAssertAllStagesStopped new SimpleRequestResponseRoundtripSetup {
         network.sendDATA(9, endStream = true, HPackSpecExamples.C41FirstRequestWithHuffman)
-        network.expectGOAWAY()
+        val (lastStreamId, error) = network.expectGOAWAY()
+        // We are over-estimating the last stream for which 'data may have been processed' here.
+        // That is allowed. Returning '0' would be better here, since the stream is not actually processed.
```
johanandren, line 1190: Is that something we should improve?

johanandren (approved):
LGTM!

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
raboof: Not currently, at least.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
raboof: It doesn't seem very high-priority, but it would be nice. I'll create a ticket.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2ServerDemux.scala
johanandren: I guess in general it shouldn't happen when a request is in progress though, but let requests complete, so maybe logging something is more reasonable.

raboof:
I think we should hold off merging this until #3685 is in, as that is more urgent and the behavior may conflict

raboof:
(setting back to WIP because behavior conflicts with other work on master - I'm not planning on continuing on this in the near future though)

Issue #3311, opened 2020-06-29
FutureDirectivesExamplesSpec: 


raboof:
similar to #1419

https://jenkins.akka.io:8498/job/akka-http-nightly/AKKA_VERSION=release-2.5,Node=akka-http-nightly-node,SCALA_VERSION=2.12.11,jdk=AdoptOpenJDK%2011/178/console

```
[info] FutureDirectivesExamplesSpec:
[info] - onComplete
[info] - onCompleteWithBreaker *** FAILED ***
[info]   Request was rejected with rejection CircuitBreakerOpenRejection(akka.pattern.CircuitBreakerOpenException: Circuit Breaker is open; calls are failing fast) (FutureDirectivesExamplesSpec.scala:96)
```

chbatey:
https://jenkins.akka.io:8498/job/akka-http-nightly/AKKA_VERSION=release-2.5,Node=akka-http-nightly-node,SCALA_VERSION=2.13.3,jdk=AdoptOpenJDK%208/317/testReport/junit/docs.http.scaladsl.server.directives/FutureDirectivesExamplesSpec/onCompleteWithBreaker/

jrudolph:
Hopefully fixed in #3694.

PR #3694, opened 2020-12-07
FutureDirectivesExampleSpec fail fix

johanandren:
References #3311

I think the cause is that the reset is timer based so needs to "happen" before the call, but if it is a bad JVM day those 200ms of leeway is not enough for it to complete before the sample http call expecting the breaker to have been reset/closed again.


raboof (approved):

jrudolph (approved):
LGTM, cannot hurt to try that.

PR #3701, opened 2020-12-09
update ScalaCheck (1.14->1.15) and ScalaTest (3.1->3.2)

SethTisue:


raboof:
We have considered this before in #3273 and at that point decided against it.

What was the reason for proposing this change? Perhaps that would be a good motivation to do the update in Akka as well (https://github.com/akka/akka/issues/29368)?

SethTisue:
I was looking at the Lightbend stack repos in the community build and seeing if any of the changes I had made locally in the build could be upstreamed to the original repos.

But I can't use akka-http master in the community build currently because Play & Lagom haven't moved. So this was just future-proofing and not something that benefits me now.

I'll close this PR and leave a breadcrumb for myself in the akka-http community build config file so I remember what the situation is for next time I run across this.

Waiting for Akka to upgrade first makes sense to me, but in a general way, I do think Akka should upgrade. Forcing end users to upgrade isn't great, but forcing end users *not* to upgrade is worse IMO. Regardless, I trust the team's judgment, I lack context.

jrudolph:
> Waiting for Akka to upgrade first makes sense to me, but I do (in a general way, at least) think Akka should upgrade. Forcing end users to upgrade isn't great, but forcing end users _not_ to upgrade is worse IMO. Regardless, I trust the team's judgment, I lack full context.

I guess, the reluctance was because scalatest 3.2 is not binary compatible with 3.1 and we have a stack of testkits between the different Akka projects that need to align scalatest versions.

Strictly spoken, we don't provide compatibility guarantees for testkits but even for our projects it can make things difficult because we also have to deal with an internal compat matrix e.g. that Akka HTTP 10.2 still works with Akka 2.5 which will not get the scalatest upgrade etc.

But I agree at some point we'll have to move and maybe during that move we'll have to figure whether we can decrease the dependency coupling between the testkits.

PR #3702, opened 2020-12-10
Telemetry SPI for the HTTP/2 client

johanandren:

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2.scala
@@ -19,6 +19,7 @@ import akka.http.scaladsl.model.headers.{ Connection, RawHeader, Upgrade, Upgrad
 import akka.http.scaladsl.model.http2.Http2SettingsHeader
 import akka.http.scaladsl.settings.ClientConnectionSettings
 import akka.http.scaladsl.settings.ServerSettings
+import akka.stream.Attributes
 import akka.stream.TLSClosing
 import akka.stream.TLSProtocol.{ SslTlsInbound, SslTlsOutbound }
 import akka.stream.impl.io.TlsUtils
@@ -78,7 +79,7 @@ private[http] final class Http2Ext(private val config: Config)(implicit val syst
       .mapAsyncUnordered(settings.maxConnections) {
         incoming: Tcp.IncomingConnection =>
           try {
-            httpPlusSwitching(http1, http2).addAttributes(prepareAttributes(settings, incoming))
+            httpPlusSwitching(http1, http2).addAttributes(prepareServerAttributes(settings, incoming))
               .watchTermination()(Keep.right)
               .join(incoming.flow)
               .run().recover {
@@ -102,11 +103,11 @@ private[http] final class Http2Ext(private val config: Config)(implicit val syst
       }.to(Sink.ignore).run()
   }
 
-  def prepareAttributes(settings: ServerSettings, incoming: Tcp.IncomingConnection) = {
+  private def prepareServerAttributes(settings: ServerSettings, incoming: Tcp.IncomingConnection) = {
     val attrs = Http.prepareAttributes(settings, incoming)
     if (telemetry == NoOpTelemetry) attrs
     else {
-      attrs.and(TelemetryAttributes.prepareConnectionAttributes(incoming))
+      attrs.and(TelemetryAttributes.prepareServerFlowAttributes(incoming))
     }
   }
 
@@ -214,7 +215,7 @@ private[http] final class Http2Ext(private val config: Config)(implicit val syst
       engine
     }
 
-    val stack = Http2Blueprint.clientStack(clientConnectionSettings, log, telemetry) atop
+    val stack = Http2Blueprint.clientStack(clientConnectionSettings, log, telemetry).addAttributes(prepareClientAttributes(host, port)) atop
       Http2Blueprint.unwrapTls atop
       LogByteStringTools.logTLSBidiBySetting("client-plain-text", clientConnectionSettings.logUnencryptedNetworkBytes) atop
       TLS(createEngine _, closing = TLSClosing.eagerClose)
@@ -223,12 +224,19 @@ private[http] final class Http2Ext(private val config: Config)(implicit val syst
   }
 
   def outgoingConnectionPriorKnowledge(host: String, port: Int, clientConnectionSettings: ClientConnectionSettings, log: LoggingAdapter): Flow[HttpRequest, HttpResponse, Future[OutgoingConnection]] = {
-    val stack = Http2Blueprint.clientStack(clientConnectionSettings, log, telemetry) atop
+    val stack = Http2Blueprint.clientStack(clientConnectionSettings, log, telemetry).addAttributes(prepareClientAttributes(host, port)) atop
       Http2Blueprint.unwrapTls atop
       LogByteStringTools.logTLSBidiBySetting("client-plain-text", clientConnectionSettings.logUnencryptedNetworkBytes) atop
       TLSPlacebo()
 
     stack.joinMat(clientConnectionSettings.transport.connectTo(host, port, clientConnectionSettings)(system.classicSystem))(Keep.right)
+  }
+
+  private def prepareClientAttributes(serverHost: String, port: Int): Attributes = {
+    if (telemetry == NoOpTelemetry) Attributes.none
+    else {
+      TelemetryAttributes.prepareClientFlowAttributes(serverHost, port)
+    }
   }
 
 }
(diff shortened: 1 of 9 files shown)


johanandren:
I think this covers all that we need, so unmarking it as wip now even though it in some sense still is.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
```diff
@@ -88,22 +88,27 @@ private[http] object Http2Blueprint {
       settings: ServerSettings,
       log: LoggingAdapter,
       initialDemuxerSettings: immutable.Seq[Setting] = Nil,
-      upgraded: Boolean = false): BidiFlow[HttpResponse, ByteString, ByteString, HttpRequest, NotUsed] =
-    httpLayer(settings, log) atop
+      upgraded: Boolean = false,
+      telemetry: TelemetrySpi): BidiFlow[HttpResponse, ByteString, ByteString, HttpRequest, NotUsed] = {
+    telemetry.serverConnection atop
```
johanandren, line 93: This is what made me think of akka/akka#29901 for when its not enabled

jrudolph: That optimization would be good in any case for other parts of implementation stacks.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2.scala
```diff
@@ -213,19 +226,27 @@ private[http] final class Http2Ext(private val config: Config)(implicit val syst
...
-    val stack = Http2Blueprint.clientStack(clientConnectionSettings, log) atop
+    val stack = Http2Blueprint.clientStack(clientConnectionSettings, log, telemetry).addAttributes(prepareClientAttributes(host, port)) atop
       Http2Blueprint.unwrapTls atop
       LogByteStringTools.logTLSBidiBySetting("client-plain-text", clientConnectionSettings.logUnencryptedNetworkBytes) atop
       TLSPlacebo()
 
     stack.joinMat(clientConnectionSettings.transport.connectTo(host, port, clientConnectionSettings)(system.classicSystem))(Keep.right)
   }
 
+  private def prepareClientAttributes(serverHost: String, port: Int): Attributes = {
```
jrudolph, line 237: nit: remove extra braces

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2.scala
```diff
@@ -100,6 +104,15 @@ private[http] final class Http2Ext(private val config: Config)(implicit val syst
       }.to(Sink.ignore).run()
   }
 
+  private def prepareServerAttributes(settings: ServerSettings, incoming: Tcp.IncomingConnection) = {
+    val attrs = Http.prepareAttributes(settings, incoming)
+    if (telemetry == NoOpTelemetry) attrs
+    else {
+      // transfer attributes allowing context propagation from serverBinding interceptor to request-response interceptor
```
jrudolph, line 111: That's really subtle (and potentially dangerous depending on the concrete set of attribute we might set on `IncomingConnection.flow`). This tight coupling to implementations of `IncomingConnection` sounds like a viability if we ever have to change details in `IncomingConnection`.

jrudolph (commented):
Nice to see how little is needed to make it work technically.

I'd prefer an interface that is more general, though.

Instead of the `TelemetrySpi` interface and single setting that only works for http2, could we introduce several config settings `akka.http.server.intercept-connection-flow`, `akka.http.server.intercept-request-response-flow`, and `akka.http.client.intercept-request-response-flow`? No need to implement them right now also for HTTP/1.1, it would just be nice to have more generally useful interfaces.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2.scala
johanandren: Yeah, I agree. This is basically doing the exact same thing that cinnamon does today, I tried hard coming up with another solution for passing context between the right places but in the end I gave up and went with this.

johanandren:
@jrudolph I can see how it makes sense with more general/and more granular SPI hooks, however I'm at least slightly worried that it would also increase the risk (chance) that they are used for all kinds of things, potentially conflicting with use of Cinnamon, and also giving us a head ache to support, because you can do pretty much _anything_ with them. With a more specific API, even though you could still do _anything_ it at least gives a hint about the purpose.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2.scala
johanandren: We have now reverted to do less HTTP/2 specifics in Cinnamon and rely more on the existing HTTP/1 instrumentation, so doing this some other way is even less of an option without lots of extra work now.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2.scala
```diff
@@ -213,19 +226,26 @@ private[http] final class Http2Ext(private val config: Config)(implicit val syst
...
       TLSPlacebo()
 
     stack.joinMat(clientConnectionSettings.transport.connectTo(host, port, clientConnectionSettings)(system.classicSystem))(Keep.right)
   }
 
+  private def prepareClientAttributes(serverHost: String, port: Int): Attributes =
+    if (telemetry == NoOpTelemetry) Attributes.none
+    else {
+      TelemetryAttributes.prepareClientFlowAttributes(serverHost, port)
+    }
```
jrudolph, line 241:
```suggestion
    else TelemetryAttributes.prepareClientFlowAttributes(serverHost, port)
```

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/TelemetrySpiSpec.scala
```diff
@@ -0,0 +1,237 @@
...
+  """
+     akka.http.server.preview.enable-http2 = on
+     akka.actor.serialize-messages = false
+     akka.http.http2-telemetry-class = "akka.http.impl.engine.http2.TestTelemetryImpl"
+  """) with ScalaFutures with BeforeAndAfterAll {
+
+  case class RequestId(id: String) extends RequestResponseAssociation
+  val requestIdAttr = AttributeKey[RequestId]("requestId")
+
+  case class ConnectionId(is: String) extends Attribute
```
jrudolph, line 63:
```suggestion
  case class ConnectionId(id: String) extends Attribute
```

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/TelemetrySpiSpec.scala
```diff
@@ -0,0 +1,237 @@
...
+          .viaMat(http2ClientFlow)(Keep.left)
+          .toMat(Sink.actorRef(responseProbe.ref, "done"))(Keep.both)
+          .run()
+      requestQueue.offer(HttpRequest())
+
+      probe.expectMsg("connection-seen")
+      val connId = probe.expectMsgType[ConnectionId]
+
+      probe.expectMsg("request-seen")
+      probe.expectMsgType[ConnectionId] should ===(connId)
```
jrudolph, line 205: Ok, so this tests that the attribute is propagated from the connection stream to the request stream.

jrudolph (approved):
LGTM (some cosmetics)

Good that we can have tests here for the interaction between SPI and HTTP infrastructure (even if those don't cover everything of what Telemetry needs).

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2.scala
```diff
@@ -213,19 +226,24 @@ private[http] final class Http2Ext(private val config: Config)(implicit val syst
...
+  private def prepareClientAttributes(serverHost: String, port: Int): Attributes =
+    if (telemetry == NoOpTelemetry) Attributes.none
+    else TelemetryAttributes.prepareClientFlowAttributes(serverHost, port)
+
 }
 
 /** INTERNAL API */
 @InternalApi
 private[http] object Http2 extends ExtensionId[Http2Ext] with ExtensionIdProvider {
+  @InternalStableApi
```
raboof, line 246: Why?

raboof (approved):
LGTM! One `@InternalStableApi` that I'm not sure is (still) relevant?

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2.scala
johanandren: We use the streamId key in telemetry, even though it is internal, to correlate client response with the request it is for.

johanandren: Sorry, I'm wrong, we stopped doing that and are using our own attribute, I'll remove this

PR #3703, opened 2020-12-10
build: upgrade to Scala 2.13.4 / scalafix 0.9.24

### LINKED ISSUE #3680: Update to Scala 2.13.4
This also seems to require upgrading semanticdb (though it's not obvious how), and scalafix (#3674).

Scala 2.13.4 introduces a bunch more warnings. We should probably look into those and silence/fix them as needed.

jrudolph:
Fixes #3680

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -27,7 +27,7 @@ object Dependencies {
   val scalafixVersion = _root_.scalafix.sbt.BuildInfo.scalafixVersion // grab from plugin
 
   val scala212Version = "2.12.12"
-  val scala213Version = "2.13.3"
+  val scala213Version = "2.13.4"
 
   val Versions = Seq(
     crossScalaVersions := Seq(scala213Version, scala212Version),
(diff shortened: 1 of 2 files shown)


raboof (approved):
Nice. This adds a bunch of warnings - I guess we should just 'gradually' triage (and then fix/silence) those as we do with other warnings?

jrudolph:
> Nice. This adds a bunch of warnings - I guess we should just 'gradually' triage (and then fix/silence) those as we do with other warnings?

Yeah, a bit unfortunate that the signal to noise ratio doesn't necessarily seem to increase.

jrudolph:
With sbt 1.4.4 and Scala 2.13.4 I see two things:

 * increased memory usage, so my usual memory settings run into OOM after a while
 * This error message related to separate compilation:

```
[error] /home/johannes/git/opensource/akka-http/akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala:986:53: could not find implicit value for parameter ev: akka.http.impl.util.Renderer[Seq[akka.http.scaladsl.model.TransferEncoding]]
[error]   def renderValue[R <: Rendering](r: R): r.type = r ~~ acceptableEncodings
[error]                                                     ^
[error] one error found
[error] (akka-http-core / Compile / compileIncremental) Compilation failed
```

If that is confirmed, I'd rather stay on whatever fixes those issues.

raboof:
> This error message related to separate compilation:

I see you created https://github.com/sbt/sbt/issues/6197

> I'd rather stay on whatever fixes those issues.

:+1:

jrudolph:
Follow up in #3780.

PR #3705, opened 2020-12-14
Content-Disposition filename encoding fix #1240

### LINKED ISSUE #1240: Incorrect handling of `filename`, `filename*` in Content-Disposition header

guntiso:


lightbend-cla-validator:
Hi @guntiso,

raboof (approved):
This looks great! 

Refs #1066 & #1240

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
```diff
@@ -455,13 +456,29 @@ final case class `Content-Location`(uri: Uri) extends jm.headers.ContentLocation
...
  */
 object `Content-Disposition` extends ModeledCompanion[`Content-Disposition`]
 final case class `Content-Disposition`(dispositionType: ContentDispositionType, params: Map[String, String] = Map.empty)
   extends jm.headers.ContentDisposition with RequestResponseHeader {
   def renderValue[R <: Rendering](r: R): r.type = {
+    import Rendering.{ contentDispositionFilenameSafeChars => safeChars }
     r ~~ dispositionType
-    params foreach { case (k, v) => r ~~ "; " ~~ k ~~ '=' ~~#! v }
+    val renderExtFilename =
+      params.get("filename").map(safeChars.indexOfFirstMismatch(_) >= 0) getOrElse false
```
jrudolph, line 468: Can this be rewritten as something like `params.get("filename").exists(safeChars(_))`?

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
```diff
@@ -455,13 +456,29 @@ final case class `Content-Location`(uri: Uri) extends jm.headers.ContentLocation
...
+    val withExtParams =
+      if (renderExtFilename && !params.contains("filename*"))
```
jrudolph, line 470: I think we should document that render logic somewhere. So, it is:

 * if `filename` exists and contains unsafe chars => also render `filename*`
 * if explicit `filename*` exists => don't generate `filename*` automatically

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
```diff
@@ -455,13 +456,29 @@ final case class `Content-Location`(uri: Uri) extends jm.headers.ContentLocation
...
+        params + ("filename*" -> params("filename"))
+      else params
+    withExtParams foreach {
+      case (k, v) if k == "filename" =>
```
jrudolph, line 474: Ok, so here we always edit `filename` to contain only safe chars.

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
```diff
@@ -455,13 +456,29 @@ final case class `Content-Location`(uri: Uri) extends jm.headers.ContentLocation
...
+        r ~~ "; " ~~ k ~~ '=' ~~ '"'
+        r.putReplaced(v, keep = safeChars, placeholder = '?') ~~ '"'
+      case (k, v) if k endsWith "*" =>
+        r ~~ "; " ~~ k ~~ '=' ~~ "UTF-8''"
```
jrudolph, line 478: So, we always render UTF-8, sounds reasonable.

jrudolph: Also, we never render a language tag, which is fine as we don't know any.

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
```diff
@@ -455,13 +456,29 @@ final case class `Content-Location`(uri: Uri) extends jm.headers.ContentLocation
...
+      params.get("filename").map(safeChars.indexOfFirstMismatch(_) >= 0) getOrElse false
+    val withExtParams =
+      if (renderExtFilename && !params.contains("filename*"))
+        params + ("filename*" -> params("filename"))
+      else params
+    withExtParams foreach {
+      case (k, v) if k == "filename" =>
+        r ~~ "; " ~~ k ~~ '=' ~~ '"'
+        r.putReplaced(v, keep = safeChars, placeholder = '?') ~~ '"'
+      case (k, v) if k endsWith "*" =>
```
jrudolph, line 477: I guess this whole logic should be moved to a more general place for wider adoption, but for this PR we can leave it here.

jrudolph (commented):
Looks very promising, thanks, @guntiso! I haven't fully reviewed yet but annotated a few assumptions / implementation choices that we need to document somewhere.

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
jrudolph: > if `filename` exists and contains unsafe chars => also render `filename*`

IIRC there's some contention about whether to render both `filename*` and `filename`. Is there any data about what is best (basically, how browsers interpret those)? Does order matter?

raboof: https://tools.ietf.org/html/rfc6266#appendix-D suggests they should both be rendered and `filename` should come first - but I'm not sure how outdated that advice might be.

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
guntiso: .exists(!safeChars(_)) ?

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
guntiso: > if filename exists and contains unsafe chars => also render filename*

This is because of the decision to hide `filename*` as implementation detail and put unicode under `filename` key in model. Then spec requires to render `filename*`

> if explicit filename* exists => don't generate filename* automatically

This is to allow control over ascii `filename` without resorting to `RawHeader`. For example, provide `"telephone"` instead of default `"t?l?phone"` (derived automatically from `"téléphone"`). It would also be expected to correctly render parameters mentioned in spec.

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
guntiso: > So, we always render UTF-8, sounds reasonable.

This is according to spec - https://tools.ietf.org/html/rfc8187#section-3.2.1 says:
> Producers MUST use the "UTF-8" ([RFC3629]) character encoding. Extension character encodings (mime-charset) are reserved for future use.

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
jrudolph: > This is because of the decision to hide `filename*` as implementation detail and put unicode under `filename` key in model. Then spec requires to render `filename*`

Makes sense.

> > if explicit filename* exists => don't generate filename* automatically
> 
> This is to allow control over ascii `filename` without resorting to `RawHeader`. For example, to provide `"telephone"` instead of default `"t?l?phone"` (derived automatically from `"téléphone"`). It would also be expected behaviour to correctly render both parameters mentioned in spec, if provided explicitly.

I agree, that's nice when needed.

akka-http-core/src/test/scala/akka/http/impl/model/parser/HttpHeaderSpec.scala
```diff
@@ -196,6 +196,28 @@ class HttpHeaderSpec extends AnyFreeSpec with Matchers {
         `Content-Disposition`(ContentDispositionTypes.attachment, Map("filename" -> "file/txt", "name" -> "field1"))
       "Content-Disposition: attachment; name=\"field1\"; other=\"\"" =!=
         `Content-Disposition`(ContentDispositionTypes.attachment, Map("name" -> "field1", "other" -> ""))
+      "Content-Disposition: attachment; filename=\"x=2*2+1(?)\"; filename*=UTF-8''x%3D2%2A2+1%28%E2%82%AC%29" =!=
```
jrudolph, line 199: Is there a rule in which order `filename` and `filename*` should appear in the parameter list? Because currently the "accidental" solution seems to be that the last `filename` parameter "wins" (probably due to the `TreeMap` constructor in https://github.com/akka/akka-http/pull/3705/files#diff-ad128a726c2bd0651c962a329a9b0d62d2e763a6d76fc3fa5b96cf7084f3173bR23).

If there's no strict rule, could you also add tests going the other way round? I suspect they might be failing ("filename" value overwrites "filename*") in which case we need a smarter resolution rule in the parser.

guntiso: I think `TreeMap` was used to handle rendering, but I do not know the reasoning behind "last parameter wins" in parser. This should be fixed, too.

akka-http-core/src/main/scala/akka/http/impl/model/parser/ContentDispositionHeader.scala
```diff
@@ -20,7 +20,13 @@ private[parser] trait ContentDispositionHeader { this: Parser with CommonRules w
 
   // http://tools.ietf.org/html/rfc6266#section-4.1
   def `content-disposition` = rule {
-    `disposition-type` ~ zeroOrMore(ws(';') ~ `disposition-parm`) ~ EOI ~> (p => TreeMap(p: _*)) ~> (`Content-Disposition`(_, _))
+    `disposition-type` ~ zeroOrMore(ws(';') ~ `disposition-parm`) ~ EOI ~> { p =>
+      val all = TreeMap(p: _*)
+      // https://tools.ietf.org/html/rfc6266#section-4.3
+      // when both "filename" and "filename*" are present in a single header field value,
+      //   recipients SHOULD pick "filename*" and ignore "filename"
```
jrudolph, line 27: Great, thanks for looking that up.

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
jrudolph: Ah, yes, indeed :)

akka-http-core/src/test/scala/akka/http/impl/model/parser/HttpHeaderSpec.scala
jrudolph: Thanks

jrudolph (approved):
LGTM, thanks a lot, @guntiso, that has been a long-standing issue now finally fixed!

Issue #2317, opened 2018-12-11
Missing documentation for `path.?/`


dsilvasc:
Hi, I tried to search the akka-http docs at https://doc.akka.io/docs/akka-http/current for `?/` but the search box won't let me type the slash -- it gets stuck in the question mark.

I tried looking in a few pages but no luck in those either:
https://doc.akka.io/docs/akka-http/current/routing-dsl/overview.html
https://doc.akka.io/docs/akka-http/current/common/uri-model.html
https://doc.akka.io/docs/akka-http/current/routing-dsl/path-matchers.html

I found the source at https://github.com/akka/akka-http/blob/ff1258ec2be4d658f143d1b7dd95d70d31744d1e/akka-http-core/src/main/scala/akka/http/scaladsl/model/Uri.scala#L502 and I think it's combining two path segments while avoiding a double slash between them, but would be nice to have scaladoc there confirming it.

I tried searching github for example usages but found none:
https://github.com/akka/akka-http/search?q=%3F%2F&unscoped_q=%3F%2F

No luck with google either:
https://www.google.co.uk/search?q=akka-http+"%3F%2F"

Or stackoverflow:
https://stackoverflow.com/search?q=%3F%2F

Should this operator be discoverable somewhere?

jrudolph:
Hi @dsilvasc, thanks for the report (and sorry for the delay, the issue was buried deep down).

I agree that operator is hard to find, there's not much to do about this because searching for operators is always hard. We should certainly add proper scaladoc to it so you can look at scaladoc or the sources to find out what it's doing.

nitikagarw:
Hi @jrudolph 
Can I please pick this issue?

raboof:
@nitikagarw that would be great!

nitikagarw:
Hi @raboof, 
I have opened https://github.com/akka/akka-http/pull/3711 for this issue. Please review. 
Thanks!

ennru:
Scaladoc added with #3711

PR #3711, opened 2020-12-19
Add scaladoc for path.?/

nitikagarw:
References #2317


raboof (approved):
weird, this method doesn't seem to get rendered into the html scaladoc at all... but having it here is already a nice improvement :+1:

ennru (approved):
LGTM.

ennru:
Thank you!

PR #3716, opened 2021-01-04
2020 -> 2021

gokyo:
References #xxxx


raboof:
OK TO TEST

we should probably also update the copyright headers (I think those are managed via sbt-headers-plugin)

raboof (approved):

Marcus-Rosti:
:rocket:

PR #3724, opened 2021-01-08
Added support to fetch chunks in the testkit without waiting for the whole response to finish

umbreak:
It is useful using `ScalatestRouteTest` but anything inside the body of `check` method ends up calling some sort of `awaitAllElements`, effectively blocking until the whole response has been retrieved.

There are cases where we might have an infinite stream (i.e. [SSE](https://doc.akka.io/docs/akka-http/10.0/sse-support.html)) and we want to verify that part of the stream is returned as expected.

I've added `chunksStream` method for this.

### DIFF
--- a/akka-http-testkit/src/main/scala/akka/http/scaladsl/testkit/RouteTest.scala
+++ b/akka-http-testkit/src/main/scala/akka/http/scaladsl/testkit/RouteTest.scala
@@ -7,6 +7,7 @@ package akka.http.scaladsl.testkit
 import akka.actor.ActorSystem
 import akka.http.scaladsl.Http
 import akka.http.scaladsl.client.RequestBuilding
+import akka.http.scaladsl.model.HttpEntity.ChunkStreamPart
 import akka.http.scaladsl.model._
 import akka.http.scaladsl.model.headers.{ Host, Upgrade, `Sec-WebSocket-Protocol` }
 import akka.http.scaladsl.server._
@@ -16,6 +17,7 @@ import akka.http.scaladsl.settings.ServerSettings
 import akka.http.scaladsl.unmarshalling._
 import akka.http.scaladsl.util.FastFuture._
 import akka.stream.SystemMaterializer
+import akka.stream.scaladsl.Source
 import akka.testkit.TestKit
 import akka.util.ConstantFun
 import com.typesafe.config.{ Config, ConfigFactory }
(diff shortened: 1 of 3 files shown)


akka-http-testkit/src/main/scala/akka/http/scaladsl/testkit/RouteTestResultComponent.scala
```diff
@@ -47,6 +47,12 @@ trait RouteTestResultComponent {
         case _                             => Nil
       }
 
+    def chunksStream: Source[ChunkStreamPart, Any] =
+      rawResponse.entity match {
+        case HttpEntity.Chunked(_, data) => data
+        case _                           => Source.empty
```
raboof, line 53: hmm, perhaps this should throw an exception?

raboof (commented):
Thanks for looking into this!

I agree it would be useful to be able to check things on an infinite stream. The reason the current implementation awaits the whole entity is because it makes it easier to write tests that refer to the entity multiple times without running into problems because you can only consume a non-strict stream once.

Having specific access to the `chunksStream` seems OK to me, though I'd also be OK with allowing access to the `rawResponse` somehow (under a clear name so people using/reading it are aware they might only be able to consume the stream once).

One small comment on the fallback when the response is not actually chunked.

raboof:
OK TO TEST

akka-http-testkit/src/main/scala/akka/http/scaladsl/testkit/RouteTestResultComponent.scala
umbreak: I can do that. What exception do you want this to throw?
I returned an empty source for consistency since the method `chunks` returns an empty List in the case the entity is not `Chunked`

raboof: > I returned an empty source for consistency since the method `chunks` returns an empty `List` in the case the entity is not Chunked

I see, you're right, that makes sense.

You still won't be able to inspect any other aspects of the response except for the chunkStream, though, since those still all go through `entity` and thus `awaitAllElements`, right? How do you think we should support that? We could probably adapt the methods that don't touch the entity's data to look at `rawResponse` instead, or provide access to `rawResponse` somehow?

umbreak: You are right, there are several methods that do not require the response data to be consumed in order for them to be available. 

I've made the necessary changes.

raboof (approved):

johanandren (approved):
LGTM

akka-http-testkit/src/main/scala/akka/http/scaladsl/testkit/RouteTest.scala
```diff
@@ -72,14 +76,14 @@ trait RouteTest extends RequestBuilding with WSTestRequestBuilding with RouteTes
     def msg(e: Throwable) = s"Could not unmarshal response to type '${implicitly[ClassTag[T]]}' for `responseAs` assertion: $e\n\nResponse was: $responseSafe"
     Await.result(Unmarshal(response).to[T].fast.recover[T] { case error => failTest(msg(error)) }, timeout)
   }
-  def contentType: ContentType = responseEntity.contentType
+  def contentType: ContentType = rawResponse.entity.contentType
```
jrudolph, line 79: :+1: sounds nice to be able to assert on these things immediately.

Issue #3707, opened 2020-12-16
Http/2: support remote-address-attribute setting


jrudolph:
It seems in HTTP/2 only the `remote-address-header` setting works but not `remote-address-attribute`.

See https://github.com/akka/akka-http/blob/e77800c5df41c51418cbc2df5b337cd7c89e7482/akka-http-core/src/main/scala/akka/http/impl/engine/http2/RequestParsing.scala#L31

PR #3726, opened 2021-01-08
Include remote address attribute in HTTP/2 requests

johanandren:
References #3707

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/RequestParsing.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/RequestParsing.scala
@@ -17,6 +17,7 @@ import akka.util.ByteString
 import akka.util.OptionVal
 import com.github.ghik.silencer.silent
 
+import java.net.InetSocketAddress
 import scala.annotation.tailrec
 import scala.collection.immutable.VectorBuilder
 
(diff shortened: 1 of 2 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/http2/RequestParsing.scala
```diff
@@ -28,12 +29,18 @@ private[http2] object RequestParsing {
...
+
     val remoteAddressHeader: Option[`Remote-Address`] =
       if (serverSettings.remoteAddressHeader) {
         attributes.get[HttpAttributes.RemoteAddress].map(remote => model.headers.`Remote-Address`(RemoteAddress(remote.address)))
         // in order to avoid searching all the time for the attribute, we need to guard it with the setting condition
       } else None // no need to emit the remote address header
 
+    val remoteAddressAttribute: Option[RemoteAddress] =
+      if (serverSettings.remoteAddressAttribute) {
+        attributes.get[HttpAttributes.RemoteAddress].map(remote => RemoteAddress(remote.address))
```
raboof, line 41: In retrospect perhaps we should have called the `attributes` parameter `streamAttributes` instead for contrast ;)

raboof (approved):

akka-http-core/src/main/scala/akka/http/impl/engine/http2/RequestParsing.scala
johanandren: Yes, very confusing when there are more kinds of attributes :/

johanandren: Oh, but that is internal so fine to change. I'll do that.

Issue #3712, opened 2020-12-20
Support for responses with multiple Content-Type headers


armanbilge:
Today while testing my Alpakka connector akka/alpakka#2548 against the Google BigQuery REST API I was very frustrated to encounter `HTTP message must not contain more than one Content-Type header`. Unfortunately, it seems that one of the endpoints responds with two `Content-Type` headers making it impossible to use from Akka. 

Similarly to #3112, is it possible to make the Akka HTTP client more lenient with regard to this? If this is agreeable, I'm happy to make the PR. Thanks in advance for your thoughts.

raboof:
> Unfortunately, it seems that one of the endpoints responds with two `Content-Type` headers

Ouch, so conflicting ones?

> Similarly to #3112, is it possible to make the Akka HTTP client more lenient with regard to this? If this is agreeable, I'm happy to make the PR. Thanks in advance for your thoughts.

I think I'd be OK with allowing this this (perhaps as an option?) in principle, but it's not so trivial to decide how to do it: the content type is not kept in the list of headers, but as a field on the entity - so there's no obvious way to keep both and let the user decide how to deal with them. Do you have ideas on how to approach that? I guess we could just drop one of the values, but that seems questionable enough that it should perhaps be opt-in.

armanbilge:
> Ouch, so conflicting ones?

Yup, `application/json; charset=UTF-8` and `text/plain; charset=utf-8`. To add insult to injury the response body is empty anyway (pure-header response). FWIW I've filed an issue on their end as well.

> I guess we could just drop one of the values, but that seems questionable enough that it should perhaps be opt-in.

This was my first instinct too. As far as I can tell a response with more than one `Content-Type` is already questionable and while a warning is more than appropriate supporting this is a necessary evil for me. Perhaps while opting-in the user can specify whether to keep the first or last? And maybe include the "dropped" headers in the header list?

armanbilge:
Sorry to bump, but I'm badly in need of a fix for this. I'm going to draft a PR along the lines of my previous comment unless there are any other suggestions for how to handle this. Thanks!

jrudolph:
Fixed in #3727

PR #3727, opened 2021-01-11
core: add option for handling conflicting Content-Type headers

armanbilge:
References #3712

Adds a new config option `akka.http.parsing.conflicting-response-content-type-header-processing-mode` to control how multiple conflicting `Content-Type` headers are handled when parsing responses.

The three modes are:

* `error`: This is Akka HTTP's current implementation and the default. If multiple `Content-Type` headers are encountered, so long as their values are identical they are ignored. Conflicting values cause parsing to fail with an error.
* `first`: Use the first encountered `Content-Type` header for the entity. Any subsequent `Content-Type` headers encountered are added to the list of headers but otherwise ignored.
* `last`: Use the last encountered `Content-Type` header for the entity. All previously encountered `Content-Type` headers are added to the list of headers but otherwise ignored.

Thanks for your consideration; this patch would help me out of a sticky situation with a Google API that I'm calling from an Alpakka connector.


raboof:
OK TO TEST

raboof (approved):
Feature LGTM.

HttpMessageParser is rather in the hot paths, so it might be good to run the benchmarks to verify this has no unexpected effect on performance (though it looks like it wouldn't).

We're preparing the 10.2.3 release right now and I suspect this won't make it in before that, but hopefully soon after!

armanbilge:
Thanks for the review! Alas if only I had made this PR sooner, but still I appreciate the fast turnaround here over waiting for the REST API in question to clean up its act.

I just ran the `HeaderParserBenchmark` before/after my PR and perhaps it did create a small decrease in throughput?

```
BEFORE
[info] Benchmark                                  (withCustomMediaTypes)   Mode  Cnt     Score    Error   Units
[info] HeaderParserBenchmark.bench_parse_headers                      no  thrpt    5  3160.588 ± 39.270  ops/ms
[info] HeaderParserBenchmark.bench_parse_headers                     yes  thrpt    5  2881.763 ± 45.288  ops/ms

AFTER
[info] Benchmark                                  (withCustomMediaTypes)   Mode  Cnt     Score     Error   Units
[info] HeaderParserBenchmark.bench_parse_headers                      no  thrpt    5  2865.017 ±  83.560  ops/ms
[info] HeaderParserBenchmark.bench_parse_headers                     yes  thrpt    5  2797.581 ± 445.163  ops/ms
```
[3727-benchmark.txt](https://github.com/akka/akka-http/files/5824195/3727-benchmark.txt)

armanbilge:
Oops, I just realized that it probably was not the right benchmark to run since I didn't change any code in the `HttpHeaderParser`, only in the `HttpMessageParser`. So which benchmark should I be using?

akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpMessageParser.scala
```diff
@@ -163,7 +164,14 @@ private[http] trait HttpMessageParser[Output >: MessageOutput <: ParserOutput] {
         case h: `Content-Type` => cth match {
           case None      => parseHeaderLines(input, lineEnd, headers, headerCount + 1, ch, clh, Some(h), teh, e100c, hh)
           case Some(`h`) => parseHeaderLines(input, lineEnd, headers, headerCount, ch, clh, cth, teh, e100c, hh)
-          case _         => failMessageStart("HTTP message must not contain more than one Content-Type header")
+          case Some(x) if isResponseParser =>
+            import ConflictingResponseContentTypeHeaderProcessingMode._
+            settings.conflictingResponseContentTypeHeaderProcessingMode match {
+              case Error => failMessageStart("HTTP message must not contain more than one Content-Type header")
+              case First => parseHeaderLines(input, lineEnd, headers += h, headerCount + 1, ch, clh, cth, teh, e100c, hh)
+              case Last  => parseHeaderLines(input, lineEnd, headers += x, headerCount + 1, ch, clh, Some(h), teh, e100c, hh)
```
ignasi35, line 172: I like that the colliding headers are not discarded but transparently provided as plain headers.

akka-http-core/src/test/scala/akka/http/impl/engine/parsing/ResponseParserSpec.scala
```diff
@@ -140,6 +141,40 @@ abstract class ResponseParserSpec(mode: String, newLine: String) extends AnyFree
         closeAfterResponseCompletion shouldEqual Seq(true)
       }
 
+      "a response with several identical Content-Type headers" in new Test {
+        """HTTP/1.1 200 OK
+          |Content-Type: text/plain; charset=UTF-8
+          |Content-Type: text/plain; charset=UTF-8
```
ignasi35, line 147: 👍🏼

akka-http-core/src/main/resources/reference.conf
```diff
@@ -698,6 +698,14 @@ akka.http {
...
     illegal-response-header-value-processing-mode = error
 
+    # Configures the processing mode when encountering conflicting `Content-Type` headers in response.
+    #
+    # Supported mode:
+    # `error` : default mode, throw an ParsingException and terminate the processing
+    # `first` : use the value of the first `Content-Type` header
+    # `last`  : use the value of the last `Content-Type` header
+    conflicting-response-content-type-header-processing-mode = error
+
```
ignasi35, line 708: Using `first` and `last` sounds a bit like using `random`. It is likely that when the server sends too many content-type headers, the order is the same but it wouldn't be surprising if the order of headers sent from the server changed in every call.

Maybe there should a list of precedences? A list with like: `["text/plain", "application/json"]` so `text/plain` takes precedence. It would be a bit more expensive but this expensive check to determine what header has higher precedence would only be invoked when the server sends an invalid response. I haven't put too much thought on this alternative, though.

armanbilge: You make a fair point about `first`/`last` being just as good as `random`. I like the idea of a precedence list (perhaps as another option) but it gets a bit tricky, such as how to handle response with Content-Types not in the precedence list, etc.

So how about this: in situations with conflicting Content-Types, we set the entity to a default Content-Type (specified in the config) e.g. `application/octet-stream`, place all the parsed Content-Type headers from the response in the header list, and leave it up to the user to sort it all out on their own?

armanbilge:
I just pushed a revision based on @ignasi35's comments. The two changes are:

1. I renamed the mode `first` to `arbitrary` to emphasize the (likely) arbitrary ordering of multiple `Content-Type` headers in the response and removed the `last` mode.
2. I added a new mode `no-content-type` which sets the entity's content type to `ContentTypes.NoContentType`. Users can then extract the `Content-Type` headers from the headers list and set the entity's correct content type as they see fit.

akka-http-core/src/main/mima-filters/10.2.2.backwards.excludes/issue-3712-conflicting-content-type-headers.excludes
```diff
@@ -0,0 +1,4 @@
+ProblemFilters.exclude[Problem]("akka.http.impl.settings.ParserSettingsImpl*")
```
jrudolph, line 1:
```suggestion
# implementation classes
ProblemFilters.exclude[Problem]("akka.http.impl.settings.ParserSettingsImpl*")
```

akka-http-core/src/main/mima-filters/10.2.2.backwards.excludes/issue-3712-conflicting-content-type-headers.excludes
```diff
@@ -0,0 +1,4 @@
...
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.impl.model.parser.HeaderParser.Settings")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.javadsl.settings.ParserSettings.getConflictingResponseContentTypeHeaderProcessingMode")
```
jrudolph, line 3:
```suggestion

# adding fields to @DoNotInherit settings classes
ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.javadsl.settings.ParserSettings.getConflictingResponseContentTypeHeaderProcessingMode")
```

akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpMessageParser.scala
```diff
@@ -161,9 +162,20 @@ private[http] trait HttpMessageParser[Output >: MessageOutput <: ParserOutput] {
...
-          case None      => parseHeaderLines(input, lineEnd, headers, headerCount + 1, ch, clh, Some(h), teh, e100c, hh)
-          case Some(`h`) => parseHeaderLines(input, lineEnd, headers, headerCount, ch, clh, cth, teh, e100c, hh)
-          case _         => failMessageStart("HTTP message must not contain more than one Content-Type header")
+          case None =>
+            parseHeaderLines(input, lineEnd, headers, headerCount + 1, ch, clh, Some(h), teh, e100c, hh)
+          case Some(`h`) =>
+            parseHeaderLines(input, lineEnd, headers, headerCount, ch, clh, cth, teh, e100c, hh)
+          case Some(`Content-Type`(ContentTypes.`NoContentType`)) =>
+            parseHeaderLines(input, lineEnd, headers += h, headerCount + 1, ch, clh, cth, teh, e100c, hh)
+          case Some(x) if isResponseParser =>
```
jrudolph, line 171: Why only for responses?

jrudolph (commented):
Feature LGTM

Is there a reason it is specific to responses? I think it could work the same for both requests and responses (even if request Content-Types are less common).

> I renamed the mode `first` to `arbitrary` to emphasize the (likely) arbitrary ordering of multiple `Content-Type` headers in the response and removed the `last` mode.

I think `first` was a good name. I'm actually also fine with keeping the `last` strategy. After all these are the most likely options to do the right thing (for a certain server stack that produces these messages).

Putting the remaining ones into the header list makes sense in any case, to use any kind of other custom resolution that might be needed.

jrudolph:
> Thanks for your consideration; this patch would help me out of a sticky situation with a Google API that I'm calling from an Alpakka connector.

Btw. what API is that, that you have trouble with? Is that something we could fix in Alpakka itself, when we have added the feature here? Or at least document how to get unstuck with?

armanbilge:
Thanks for the review! I'm happy to remove the restriction to just responses; my assumption was that we don't want to encourage this behavior (are conflicting `Content-Type` headers allowed by the HTTP spec?).

> what API is that, that you have trouble with?

I'm working on a Google BigQuery connector for Alpakka akka/alpakka#2548. While testing it I encountered Google APIs that return conflicting `Content-Type` headers such that they are completely unusable from Akka; I've filed an issue with Google as well ([176002651](https://issuetracker.google.com/176002651)).

akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpMessageParser.scala
```diff
@@ -161,9 +162,21 @@ private[http] trait HttpMessageParser[Output >: MessageOutput <: ParserOutput] {
...
         case h: `Content-Type` => cth match {
-          case None      => parseHeaderLines(input, lineEnd, headers, headerCount + 1, ch, clh, Some(h), teh, e100c, hh)
-          case Some(`h`) => parseHeaderLines(input, lineEnd, headers, headerCount, ch, clh, cth, teh, e100c, hh)
-          case _         => failMessageStart("HTTP message must not contain more than one Content-Type header")
+          case None =>
+            parseHeaderLines(input, lineEnd, headers, headerCount + 1, ch, clh, Some(h), teh, e100c, hh)
+          case Some(`h`) =>
+            parseHeaderLines(input, lineEnd, headers, headerCount, ch, clh, cth, teh, e100c, hh)
+          case Some(`Content-Type`(ContentTypes.`NoContentType`)) =>
+            parseHeaderLines(input, lineEnd, headers += h, headerCount + 1, ch, clh, cth, teh, e100c, hh)
```
jrudolph, line 170: Why is this one handled specially? Is that even a content-type that we would expect to see on the wire at all?

akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpMessageParser.scala
```diff
@@ -161,9 +162,21 @@ private[http] trait HttpMessageParser[Output >: MessageOutput <: ParserOutput] {
...
+          case Some(x) =>
+            import ConflictingContentTypeHeaderProcessingMode._
+            settings.conflictingContentTypeHeaderProcessingMode match {
+              case Error         => failMessageStart("HTTP message must not contain more than one Content-Type header")
+              case First         => parseHeaderLines(input, lineEnd, headers += h, headerCount + 1, ch, clh, cth, teh, e100c, hh)
+              case Last          => parseHeaderLines(input, lineEnd, headers += x, headerCount + 1, ch, clh, Some(h), teh, e100c, hh)
+              case NoContentType => parseHeaderLines(input, lineEnd, headers += x += h, headerCount + 1, ch, clh, Some(`Content-Type`(ContentTypes.`NoContentType`)), teh, e100c, hh)
+            }
+          case _ => failMessageStart("HTTP message must not contain more than one Content-Type header")
```
jrudolph, line 179: Is that case needed?

akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpMessageParser.scala
jrudolph: Ah, I see it below. Maybe add a comment that this is how the `NoContentType` setting is handled.

akka-http-core/src/main/scala/akka/http/scaladsl/settings/ParserSettings.scala
```diff
@@ -163,7 +167,7 @@ object ParserSettings extends SettingsCompanion[ParserSettings] {
         case "error"  => Error
         case "warn"   => Warn
         case "ignore" => Ignore
-        case x        => throw new IllegalArgumentException(s"[$x] is not a legal `illegal-response-header-value-processing-mode` setting")
+        case x        => throw new IllegalArgumentException(s"[$x] is not a legal `illegal-header-value-processing-mode` setting")
```
jrudolph, line 170: Was that change intended?

akka-http-core/src/main/scala/akka/http/scaladsl/settings/ParserSettings.scala
```diff
@@ -182,6 +186,23 @@ object ParserSettings extends SettingsCompanion[ParserSettings] {
...
+    case object Last extends ConflictingContentTypeHeaderProcessingMode
+    case object NoContentType extends ConflictingContentTypeHeaderProcessingMode
+
+    def apply(string: String): ConflictingContentTypeHeaderProcessingMode =
+      string.toRootLowerCase match {
+        case "error"           => Error
+        case "first"           => First
+        case "last"            => Last
+        case "no-content-type" => NoContentType
+        case x                 => throw new IllegalArgumentException(s"[$x] is not a legal `conflicting-response-content-type-header-processing-mode` setting")
```
jrudolph, line 202: Seems the above change should have been applied here :)

jrudolph (commented):
Looking good, some minor comments below

akka-http-core/src/test/scala/akka/http/impl/engine/parsing/ResponseParserSpec.scala
```diff
@@ -140,6 +141,60 @@ abstract class ResponseParserSpec(mode: String, newLine: String) extends AnyFree
...
+        """HTTP/1.1 200 OK
+          |Content-Type: text/plain; charset=UTF-8
+          |Content-Type: application/json; charset=utf-8
+          |Content-Length: 0
+          |
+          |""" should parseTo(HttpResponse(headers = List(`Content-Type`(ContentTypes.`application/json`)), entity = HttpEntity.empty(ContentTypes.`text/plain(UTF-8)`)))
+        closeAfterResponseCompletion shouldEqual Seq(false)
+      }
+
+      "a response with several conflicting Content-Type headers with conflicting-response-content-type-header-processing-mode = last" in new Test {
```
jrudolph, line 166: here as well

akka-http-core/src/test/scala/akka/http/impl/engine/parsing/ResponseParserSpec.scala
```diff
@@ -140,6 +141,60 @@ abstract class ResponseParserSpec(mode: String, newLine: String) extends AnyFree
...
+        """HTTP/1.1 200 OK
+          |Content-Type: text/plain; charset=UTF-8
+          |Content-Type: application/json; charset=utf-8
+          |Content-Length: 0
+          |
+          |""" should parseTo(HttpResponse(headers = List(`Content-Type`(ContentTypes.`text/plain(UTF-8)`)), entity = HttpEntity.empty(ContentTypes.`application/json`)))
+        closeAfterResponseCompletion shouldEqual Seq(false)
+      }
+
+      "a response with several conflicting Content-Type headers with conflicting-response-content-type-header-processing-mode = no-content-type" in new Test {
```
jrudolph, line 178: and here

jrudolph (commented):
Found a few more minor things. Would you mind adding some tests for the request side as well?

armanbilge:
Thanks for catching those, sorry for all the little mistakes! All fixed now and added request tests.

jrudolph (approved):
LGTM, thanks a lot @armanbilge.

armanbilge:
Thanks for merging! Do we know yet which version of Akka HTTP this will be in? I'd like to mention this workaround in the docs for the Alpakka BigQuery connector.

jrudolph:
Yes, it will be 10.2.4 (which will probably not be released any time soon, though, as we just released 10.2.3). There is (or will be) an automatic snapshot build which can be used until then, see https://doc.akka.io/docs/akka-http/snapshot/contributing.html.

Issue #3594, opened 2020-11-02
HTTP/2: react gracefully on user-handler completion


jrudolph:
Regular user-handler completion should not immediately shut down the HTTP/2 pipeline but let existing requests / responses being finished for some time during a grace period.

One question is whether we should wait with sending out a `GOAWAY` frame until we have received everything that we want to receive.

On the client:
 * The functionality is needed so an HTTP/2 connection can be regularly closed by the user without requiring the user to monitor the state of ongoing requests.
 * We can wait with GOAWAY until all responses have been received, in which case the connection can be closed.
 * The grace period can be long or even infinite by default because we control the connection and the resources.

On the server:
 * The user-handler is not really expected to be completed by the user, but more likely by the graceful shutdown logic. Supporting graceful shutdown logic only makes sense in the HTTP/2 server infrastructure when it is actually implemented on top.
 * We might want to send a GOAWAY frame with `NO_ERROR` right away to prevent the client from opening new streams while still streaming out responses.
 * The grace period should be reasonable but not infinite so that a slow client can be dropped after a while.

PR #3734, opened 2021-01-13
Graceful shutdown: block Http2Demux stage completion until in-flight ops complete.

ignasi35:
This is the first step to solving #3594.

When the inlet in `Http2Demux` receiving elements from the user handler finishes, the internals of Http2Demux block the stage completion until all currently-open streams complete. 

When all streams are closed, though, any buffered data outgoing via the substream outlet (sent to the user handler) must also be delivered. To implement this part we've replaced (thanks @raboof for the assistance on this) the custom `BufferedOutlet` for an imperative implementation based on `emit` similar to what the FrameEvent outlet is already doing. As a consequence, `BufferedOutlet` is no longer used and removed. 

NOTE: this PR is focused on the client-side since the server-side still doesn't support PUSH_PROMISE so stream creation is not possible, and graceful shutdown on the server takes the whole stack down.

Out of scope for this PR:
 - complete the stage after a timeout to avoid infinitely waiting for in-flight streams to complete.
 - properly implement any GOAWAY edge case.


akka-http-core/src/main/scala/akka/http/impl/engine/http2/BufferedOutletSupport.scala
```diff
@@ -21,78 +21,6 @@ private[http2] trait GenericOutlet[T] {
...
-
-  final def push(element: T): Unit = pushWithTrigger(element, () => ())
-  final def pushWithTrigger(elem: T, trigger: () => Unit): Unit =
-    if (outlet.canBePushed && buffer.isEmpty) doPush(ElementAndTrigger(elem, trigger))
-    else buffer.addLast(ElementAndTrigger(elem, trigger))
-
-  def tryFlush(): Unit =
-    if (outlet.canBePushed && !buffer.isEmpty)
-      doPush(buffer.pop())
-}
```
ignasi35, line 94: None of this is used anymore.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Demux.scala
```diff
@@ -277,7 +280,7 @@ private[http2] abstract class Http2Demux(http2Settings: Http2CommonSettings, ini
 
         allowReadingIncomingFrames = allow
       }
-      def pullFrameIn(): Unit = if (allowReadingIncomingFrames && !hasBeenPulled(frameIn)) pull(frameIn)
+      def pullFrameIn(): Unit = if (allowReadingIncomingFrames && !hasBeenPulled(frameIn) && !isClosed(frameIn)) pull(frameIn)
```
ignasi35, line 283: when improving the shutdown sequence I stumbled upon scenarios where we tried to pull from a closed inlet.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Demux.scala
```diff
@@ -287,6 +290,7 @@ private[http2] abstract class Http2Demux(http2Settings: Http2CommonSettings, ini
         }
       }
 
+      // -----------------------------------------------------------------
```
ignasi35, line 293: Added some separators to let the code breathe a bit.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Demux.scala
```diff
@@ -357,21 +361,40 @@ private[http2] abstract class Http2Demux(http2Settings: Http2CommonSettings, ini
...
+      //        after a while or buffer only a limited amount?
+      substreamOut.setHandler(new OutHandler() {
+        // This is a noop handler because we're never pushing on this outlet and,
+        // instead, we always emit (which teporarily swaps this handler with an
+        // internal handler that implements the emit.
+        override def onPull(): Unit = ()
+      })
       override def dispatchSubstream(initialHeaders: ParsedHeadersFrame, data: Source[Any, Any], correlationAttributes: Map[AttributeKey[_], _]): Unit =
-        bufferedSubStreamOutput.push(Http2SubStream(initialHeaders, data, correlationAttributes))
+        emit(substreamOut, Http2SubStream(initialHeaders, data, correlationAttributes))
```
ignasi35, line 375: This is the key change: we're dropping `BufferedOutlet` in favor of using the internal buffer `emit` provides. 

In order to use `emit` we're setting a noop handler (since `emit` fails to operate on outlets without handler).

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Demux.scala
```diff
@@ -357,21 +361,40 @@ private[http2] abstract class Http2Demux(http2Settings: Http2CommonSettings, ini
...
+          //     super.onUpstreamFinish()
+          // flag the stream handling that the stage is finishing. When the stream handling is ready, it
+          // must invoke the complete the stage.
+          // Summing up, this passed the responsibility of completing the stage to Http2StreamHandling.
+          onUserUpstreamFinish()
+          //  - when receiving a response -> if(onFinishingState && no more Inflight Responses) then cancel the
+          //    timer and proceed with the finish logic
+          // start a timer
+          //  - on timer, log the timer event and proceed with the finish logic
+          //
```
ignasi35, line 396: The timer handling I'll do in a separate PR.

ignasi35 (commented):
Some comments for reviewers

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
```diff
@@ -753,6 +767,62 @@ class Http2ClientSpec extends AkkaSpecWithMaterializer("""
...
+
+        // When the missing response arrives
+        Seq(
+          HeadersFrame(streamId = 1, endStream = true, endHeaders = true, HPackSpecExamples.C61FirstResponseWithHuffman, None)
+        ).foreach(network.sendFrame)
+        // Then the user can still consume the response
+        user.expectResponse()
+      }
+
+      "until in-flight streams deliver the request and responses are received" inAssertAllStagesStopped new TestSetup(
```
ignasi35, line 787: I've added this extra test to exercise 2 things at once:

 - new graceful shutdown interactions with multiple open streams
 - new graceful shutdown interactions with backpressured streams (SETTINGS_MAX_CONCURRENT_STREAMS)

ignasi35:
hmm, if the stream completion happens when there's no stream in-flight, the completion is not fulfilled.

johanandren:
PLS BUILD

akka-http-core/src/main/scala/akka/http/impl/engine/http2/BufferedOutletSupport.scala
johanandren: Would be good to get some input from @jrudolph on why this was used in the first place, seems strange that a custom buffering thing was introduced if emit works just as well.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Demux.scala
```diff
@@ -357,21 +361,40 @@ private[http2] abstract class Http2Demux(http2Settings: Http2CommonSettings, ini
         }
       })
 
+      // -----------------------------------------------------------------
       // FIXME: What if user handler doesn't pull in new substreams? Should we reject them
-      //        after a while or buffer only a limited amount? We should also be able to
-      //        keep the buffer limited to the number of concurrent streams as negotiated
-      //        with the other side.
-      val bufferedSubStreamOutput = new BufferedOutlet[Http2SubStream](substreamOut)
+      //        after a while or buffer only a limited amount?
```
johanandren, line 366: I guess this as well as limiting the size of the buffer becomes harder/impossible with the custom buffered outlet replaced with emit. The buffer is now in fact an Akka streams internal linked list of emits with no API to look at size/interrupt it etc?

akka-http-core/src/main/scala/akka/http/impl/engine/http2/BufferedOutletSupport.scala
```diff
@@ -21,78 +21,6 @@ private[http2] trait GenericOutlet[T] {
   def canBePushed: Boolean
 }
```
johanandren, line 22: Are the rest of the stuff in this file used when the BufferedOutlet is removed?

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Demux.scala
```diff
@@ -357,21 +361,40 @@ private[http2] abstract class Http2Demux(http2Settings: Http2CommonSettings, ini
...
+      substreamOut.setHandler(new OutHandler() {
+        // This is a noop handler because we're never pushing on this outlet and,
+        // instead, we always emit (which teporarily swaps this handler with an
+        // internal handler that implements the emit.
+        override def onPull(): Unit = ()
+      })
       override def dispatchSubstream(initialHeaders: ParsedHeadersFrame, data: Source[Any, Any], correlationAttributes: Map[AttributeKey[_], _]): Unit =
-        bufferedSubStreamOutput.push(Http2SubStream(initialHeaders, data, correlationAttributes))
+        emit(substreamOut, Http2SubStream(initialHeaders, data, correlationAttributes))
```
johanandren, line 374: Do you think you could explain exactly what is handled differently with emit vs the `BufferedOutlet`, is it when completion happens with buffered elements?

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -96,6 +97,15 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
...
   def activeStreamCount(): Int = streamStates.size
 
+  /**
+   * This method signals the Http2StreamHandling that the enclosing stage (Http2Demux) is
+   * completing because the userland inlet was finished. Do not invoke this when the cause
+   * is the connection is closed or in case of failure in any of the inlets.
+   */
+  def onUserUpstreamFinish(): Unit = {
+    userUpstreamFinishing = true
+  }
```
johanandren, line 107: Does it need to be a method when we only call it from one place and all it does is toggling a boolean named the same? (explanation could go on field instead if important)

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Demux.scala
ignasi35: I'm not sure I understand your question. Actually, I think Github UI failed to mark it as resolved since I've rolled-back the use of `emit` to restore the use of the `BufferedOutlet`.

The buffer (in `BufferedOutlet` or in Akka Stream internals) can only blow up when:

 - the remote peer sets a very high MAX_CONCURRENT_STREAMS
 - the user handler doesn't grab the emitted elements

The server is protected because we don't support PUSH_PROMISE.

The client could blow up if:

 - the servers supports PUSH_PROMISE and the client set an unlimited MAX_CONCURRENT_STREAMS, or
 - the user creates more requests than it pulls responses (and the server sets an unlimited MAX_CONCURRENT_STREAMS)

There is room for improvement here, but I don't think it's in scope for this PR.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Demux.scala
```diff
@@ -357,21 +360,39 @@ private[http2] abstract class Http2Demux(http2Settings: Http2CommonSettings, ini
...
           handleOutgoingCreated(sub)
           // Once the incoming stream is handled, we decide if we need to pull more.
           tryPullSubStreams()
         }
+
+        override def onUpstreamFinish(): Unit = {
+          // marks StreamHandling as ready for completion
+          tryComplete()
+          // start a timer
+          //  - on timer, log the timer event and proceed with the finish logic
```
ignasi35, line 394: We'll do the timeout check to ensure the stage eventually completes (even if not all responses were delivered back) in a separate PR.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Demux.scala
```diff
@@ -357,21 +360,39 @@ private[http2] abstract class Http2Demux(http2Settings: Http2CommonSettings, ini
...
 
+      // -----------------------------------------------------------------
+      // This is a manually crafted sequence of shutdowns replacing the one on `completeStage`
+      // so we use the custom bufferedSubStreamOutput.complete() to ensure the buffer is empty
+      // before completing the outlet
+      override def onComplete(): Unit = {
+        cancel(substreamIn)
+        cancel(frameIn)
+        complete(frameOut)
+        bufferedSubStreamOutput.complete()
```
ignasi35, line 378: `completeStage` would invoke:

```
        cancel(substreamIn)
        cancel(frameIn)
        complete(frameOut)
        complete(substreamOut)
```
which is wrong because it doesn't flush the buffer in `bufferedSubStreamOutput` in from of `substreamOut`.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
```diff
@@ -753,6 +766,82 @@ class Http2ClientSpec extends AkkaSpecWithMaterializer("""
...
+        //... and it is possible to consume all responses (from stream 1 to stream 7)
+        val receivedResponse3: HttpResponse = user.expectResponse()
+        val receivedResponse5: HttpResponse = user.expectResponse()
+        val receivedResponse7: HttpResponse = user.expectResponse()
+      }
+
+      "until a timeout occurs" ignore {
+
+        // test timed completion
+      }
```
ignasi35, line 824: In a separate PR.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/TelemetrySpiSpec.scala
```diff
@@ -151,69 +151,74 @@ abstract class TelemetrySpiSpec(useTls: Boolean) extends AkkaSpecWithMaterialize
...
-      probe.expectMsg("close-seen")
-      responseProbe.expectMsg("done")
+      // As a consequence the whole stream completes.
+      telemetryProbe.expectMsg("close-seen")
+      responseProbe.expectMsg("onCompleteMessage")
 
       serverBinding.terminate(3.seconds).futureValue
-      probe.expectMsg("unbind-seen")
+      telemetryProbe.expectMsg("unbind-seen")
     }
```
ignasi35, line 222: This last bit is the only place where there are actual differences.

ignasi35:
Thanks @johanandren for the review. The removal of the `BufferedOutlet` in favor of `emit` was the wrong solution so we've rolled it back and applied the correct fix: use a carefully crafted sequence of inlet cancelations and outlet completions with an explicit completion of the `BufferedOutlet`.

ignasi35:
Warning: I've created the MiMa excludes on the `10.2.3.../` folder already as I don't expect this to be merged before the 10.2.3 release!

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Demux.scala
johanandren: Ah, I indeed missed the restore of BufferedOutlet, you can ignore this comment!

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -153,6 +172,7 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
     newState match {
       case Closed =>
         streamStates -= streamId
+        if (streamStates.isEmpty && completing) onComplete()
```
raboof, line 175: since `completing` is only set when `substreamIn` completes, I wonder if it would be simpler to check if `isClosed(substreamIn)` here

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -75,6 +78,22 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
...
+  def tryComplete(): Unit = {
+    completing = true
+    if (streamStates.isEmpty) onComplete()
+  }
+
+  /**
+   * Subclasses must override this with the code to run when all streams in StreamHandling
+   * buffers are closed
+   */
+  def onComplete()
```
raboof, line 95: this looks like an event handler that is called when the stage completes, but actually it is the function that does the completing. Perhaps call it `complete()`?

akka-http-core/src/main/scala/akka/http/impl/engine/http2/BufferedOutletSupport.scala
```diff
@@ -38,9 +38,10 @@ private[http2] class BufferedOutlet[T](outlet: GenericOutlet[T]) extends OutHand
   outlet.setHandler(this)
 
   def onPull(): Unit = tryFlush()
-  def push(elem: T): Unit =
+  def push(elem: T): Unit = {
     if (outlet.canBePushed && buffer.isEmpty) doPush(elem)
     else buffer.addLast(elem)
+  }
```
johanandren, line 44: I think curlies are generally avoided in Akka HTTP sources when there is just a single expression

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Demux.scala
johanandren: Good catch

raboof (commented):
change looks good, nice to see how (relatively) simple things have eventually become.

some naming thing to think about

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Demux.scala
johanandren: Mark it as a FIXME?

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -75,6 +78,22 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
...
+  /**
+   * Marks this object to be shutting down. If the buffer is empty, it also invokes [[onComplete()]].
+   */
+  def tryComplete(): Unit = {
+    completing = true
+    if (streamStates.isEmpty) onComplete()
+  }
+
+  /**
+   * Subclasses must override this with the code to run when all streams in StreamHandling
```
johanandren, line 92: Nitpick: that would be implement, since there is nothing to override.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -75,6 +78,22 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
     activeStreamCount() < maxConcurrentStreams
   }
 
+  private var completing = false
+
+  /**
+   * Marks this object to be shutting down. If the buffer is empty, it also invokes [[onComplete()]].
```
johanandren, line 84: Nitpick: A bit superflous repeating exactly what is in the code with the "if a then calls b"?

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
```diff
@@ -753,6 +766,82 @@ class Http2ClientSpec extends AkkaSpecWithMaterializer("""
...
+        user.emitRequest(HttpRequest(uri = "https://www.example.com/"))
+        network.expectDecodedResponseHEADERSPairs(1)
+        // ... and a completed user handler
+        user.requestOut.sendComplete()
+
+        network.sendFrame(
+          HeadersFrame(streamId = 1, endStream = true, endHeaders = true, HPackSpecExamples.C61FirstResponseWithHuffman, None)
+        )
+        // Then the user can still consume the response
+        user.expectResponse()
```
johanandren, line 783: Isn't there a missing assert/expect here that the response stream is also completed after that response?

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
```diff
@@ -753,6 +766,82 @@ class Http2ClientSpec extends AkkaSpecWithMaterializer("""
...
+        user.emitRequest(request)
+        user.emitRequest(request)
+        user.emitRequest(request)
+        // ... and backpressured requests...
+        user.emitRequest(request) // this emit succeeds but is buffered
+
+        network.expect[HeadersFrame]().streamId shouldBe (1)
+        network.expect[HeadersFrame]().streamId shouldBe (3)
+        network.expect[HeadersFrame]().streamId shouldBe (5)
+        network.expectNoBytes(100.millis)
```
johanandren, line 800: Is this important to verify (wouldn't that belong in some other tests to verify nothing but the headers are emitted)?

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
```diff
@@ -753,6 +766,82 @@ class Http2ClientSpec extends AkkaSpecWithMaterializer("""
...
+        // When the first few response-frames arrive...
+        network.sendFrame(HeadersFrame(streamId = 3, endStream = true, endHeaders = true, HPackSpecExamples.C61FirstResponseWithHuffman, None))
+        // Then the backpressured in-flight streams proceed (sending pending requests)...
+        network.expect[HeadersFrame]().streamId shouldBe (7)
+        network.sendFrame(HeadersFrame(streamId = 5, endStream = true, endHeaders = true, HPackSpecExamples.C61FirstResponseWithHuffman, None))
+        network.sendFrame(HeadersFrame(streamId = 7, endStream = true, endHeaders = true, HPackSpecExamples.C61FirstResponseWithHuffman, None))
+        //... and it is possible to consume all responses (from stream 1 to stream 7)
+        val receivedResponse3: HttpResponse = user.expectResponse()
+        val receivedResponse5: HttpResponse = user.expectResponse()
+        val receivedResponse7: HttpResponse = user.expectResponse()
```
johanandren, line 818: Shouldn't this also consume the response bodies and then verify response stream completed though?

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
```diff
@@ -753,6 +766,82 @@ class Http2ClientSpec extends AkkaSpecWithMaterializer("""
...
+        network.sendFrame(
+          HeadersFrame(streamId = 1, endStream = true, endHeaders = true, HPackSpecExamples.C61FirstResponseWithHuffman, None)
+        )
+        user.expectResponse()
+
+        // When the user completes the stream
+        user.requestOut.sendComplete()
+
+        // Then all stages are stopped
+        // (implicitly asserted via `inAssertAllStagesStopped`
```
johanandren, line 840: would anyways be nice with an explicit expect response stream completed imo

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
```diff
@@ -47,7 +60,7 @@ import scala.concurrent.duration._
  */
 class Http2ClientSpec extends AkkaSpecWithMaterializer("""
     akka.http.client.remote-address-header = on
-    akka.http.client.http2.log-frames = on
+    akka.http.client.http2.log-frames = off
```
johanandren, line 63: Isn't this good to have for failures, should anyway only be logged when failing since `AkkaSpecWithMaterializer` mixes in `WithLogCapturing`?

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Demux.scala
johanandren: A matter of taste but I would have skipped the three liner comment since some of it is implicit by overriding and maybe rather added one above the `bufferedSubStreamOutput` saying something like `// important to delegate to bufferedSubStreamOutput complete(substreamOut)` since that is the only thing that is "manually crafted"

johanandren (commented):
Some feedback, mostly nitpicking/style etc, looks good good to go except for that.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/TelemetrySpiSpec.scala
johanandren: Hmmmm, I do't see a difference except renaming the probe, adding comments and changing the "done" string to "onCompleteMessage"? Am I missing something?

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
ignasi35: I like the idea but I’ve tried and can’t seem to make it work. 

Either (1) I overwrite the `onUpstreamFinish` for `substreamIn` to do nothing (then it’s not closed), or (2) implement `onUpstreamFinish` to `cancel(substreamIn)` which seems to not close the stage (?).

For the record, code looked like:

```scala
      override def complete(): Unit = {
        if (!isClosed(substreamIn)) // may be close if the invocation to complete() was delayed
          cancel(substreamIn)
        cancel(frameIn)
        complete(frameOut)
        bufferedSubStreamOutput.complete()
      }
      override def completing(): Boolean = isClosed(substreamOut)
      setHandler(substreamIn, new InHandler {
	    ...
        override def onUpstreamFinish(): Unit = {
          if (activeStreamCount() == 0) complete() // no buffer so proceed with complete()
          else cancel(substreamIn)  // work pending, close the inlet

          // FIXME: start a timer
          //  - on timer, log the timer event and proceed with the finish logic
        }
      })
```

I think it's good enough as we have it.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
ignasi35: that's a mistake. I forgot to enable this again.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
ignasi35: This ensures that the stream `7` is not just open, but not even seen on the network.

Then, this test ensures that both (1) streams expecting response, and (2) backpressured streams with a valid request ready but backpressured by MAX_CONCURRENT_STREAMS can complete.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
ignasi35: Good point.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
```diff
@@ -753,6 +766,86 @@ class Http2ClientSpec extends AkkaSpecWithMaterializer("""
...
+        network.sendFrame(HeadersFrame(streamId = 7, endStream = true, endHeaders = true, HPackSpecExamples.C61FirstResponseWithHuffman, None))
+        //... and it is possible to consume all responses (from stream 1 to stream 7)
+        val receivedResponse3: HttpResponse = user.expectResponse()
+        val receivedResponse5: HttpResponse = user.expectResponse()
+        val receivedResponse7: HttpResponse = user.expectResponse()
+      }
+
+      "until in-flight responses with chunked payload are fully received and consumed" ignore {
+        // Assert that not only the response but the full entity is available
+      }
```
ignasi35, line 824: I've added this test skeleton as a response to https://github.com/akka/akka-http/pull/3734/files#r558348294.

We can merge this PR and implement the test on a new PR to get the feature in 10.2.3 or we can do things right and implement the test before merging this PR :-)

johanandren: What happens right now, it receives the headers/request but fails the body source?

johanandren: I think this should be covered by this PR as well.

jrudolph: Isn't it already tested above? Response data is always streamed with HTTP/2 because headers and body arrive in distinct frames.

jrudolph: Or is it about the distinction of reporting trailers with a chunked response?

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
```diff
@@ -753,6 +767,104 @@ class Http2ClientSpec extends AkkaSpecWithMaterializer("""
...
+
+      "until a timeout occurs" ignore {
+        // FIXME: test timed completion
+      }
+
+      "actually complete as soon as required if there are no open (in-flight) streams" inAssertAllStagesStopped new TestSetup with NetProbes {
+        // Given a request...
+        user.emitRequest(HttpRequest(uri = "https://www.example.com/"))
+        network.expectDecodedResponseHEADERSPairs(1)
+        // ... and a demanded response for that request (so there's nothing in-flight)
```
jrudolph, line 853:
```suggestion
        // ... and return response for that request (so there's nothing in-flight)
```

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/TelemetrySpiSpec.scala
```diff
@@ -34,6 +34,8 @@ import org.scalatest.concurrent.ScalaFutures
 import java.util.UUID
 import scala.concurrent.Future
 import scala.concurrent.duration._
+import scala.util.Failure
```
jrudolph, line 37: Are these changes related to the PR?

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
raboof: Looks like they're now tested above! (previously they weren't, this placeholder/TODO test can be removed now ;) )

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -153,6 +169,7 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
     newState match {
       case Closed =>
         streamStates -= streamId
+        if (streamStates.isEmpty && completing) complete()
```
jrudolph, line 172: I think I would prefer it to keep the whole inlet/outlet logic to the Demux. So, could this be something like 

```
if (streamStates.isEmpty) onAllStreamsClosed()
```

and implement that in the Demux (together with the logic from `tryComplete`). The flag `completed` probably wouldn't be necessary in that case as you could use `isClosed()` there directly.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/TelemetrySpiSpec.scala
ignasi35: Mostly not, but some are. 
In this file I renamed a value but also had to reorder some expectations since the shutdown sequence varied a bit.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
ignasi35: Ah, good idea. Looks cleaner too.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
```diff
@@ -753,6 +767,104 @@ class Http2ClientSpec extends AkkaSpecWithMaterializer("""
...
+        network.sendFrame(DataFrame(streamId = 7, endStream = true, ByteString("Hello World 7!")))
+        //... and it is possible to consume all responses (from stream 1 to stream 7)
+        val resp3: HttpResponse = user.expectResponse()
+        val resp5: HttpResponse = user.expectResponse()
+        val resp7: HttpResponse = user.expectResponse()
+
+        private val eventualEntities: Seq[Future[HttpEntity.Strict]] =
+          Seq(resp1, resp3, resp5, resp7).map(_.entity.toStrict(100.millis))
+        private val entities: Seq[HttpEntity.Strict] = Future.sequence(eventualEntities).futureValue
+        // The complete entity of the responses si available
```
jrudolph, line 828:
```suggestion
        // The complete entity of the responses is available
```

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Demux.scala
```diff
@@ -357,25 +360,50 @@ private[http2] abstract class Http2Demux(http2Settings: Http2CommonSettings, ini
...
         def onPush(): Unit = {
           val sub = grab(substreamIn)
           handleOutgoingCreated(sub)
           // Once the incoming stream is handled, we decide if we need to pull more.
           tryPullSubStreams()
         }
+
+        override def onUpstreamFinish(): Unit = {
+          // marks StreamHandling as ready for completion
+          tryComplete()
```
jrudolph, line 398: Do we still need `tryComplete`?

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Demux.scala
```diff
@@ -357,25 +360,50 @@ private[http2] abstract class Http2Demux(http2Settings: Http2CommonSettings, ini
...
-      //        after a while or buffer only a limited amount? We should also be able to
-      //        keep the buffer limited to the number of concurrent streams as negotiated
-      //        with the other side.
+      //        after a while or buffer only a limited amount?
       val bufferedSubStreamOutput = new BufferedOutlet[Http2SubStream](substreamOut)
       override def dispatchSubstream(initialHeaders: ParsedHeadersFrame, data: Source[Any, Any], correlationAttributes: Map[AttributeKey[_], _]): Unit =
         bufferedSubStreamOutput.push(Http2SubStream(initialHeaders, data, correlationAttributes))
 
+      // -----------------------------------------------------------------
+      private var completing = false
```
jrudolph, line 371: Do we still need the extra flag? Could we use `isClosed(substreamIn)` instead?

jrudolph (commented):
Mostly looks good. Some final cosmetic remarks below.

Is the behavior reasonable for the server side? I don't think so because streams will be stalled in an open state if they are waiting for responses on the then closed response `substreamIn`.

Even if we don't want to implement the completely correct behavior for the server-side in this PR, we should make sure the situation doesn't get worse and we risk leaking connections (more than we do even now).

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Demux.scala
ignasi35: I've tried it and it doesn't work. IIUC, `isClosed(inlet)` only returns true _after_ the invoking `complete(inlet)`.

I think the flag and `tryComplete` are necessary. `isClosed(substreamIn)` won't make it because `completing` (the flag) is already `true` when `isClosed(substreamIn)` is still false. While `completing` is already `true` we still need to have the inlet open to push elements.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Demux.scala
ignasi35: see https://github.com/akka/akka-http/pull/3734#discussion_r561190431

johanandren (approved):
LGTM, but I agree with Johannes on the need to verify server side behavior

ignasi35:
> Is the behavior reasonable for the server side? I don't think so because streams will be stalled in an open state if they are waiting for responses on the then closed response `substreamIn`.

As long as we don't support PUSH_PROMISE, I think this implementation is OK. Let me try to explain my train of thought in case I'm missing something. The following is how I think the graceful shutdown on a server works:

 - there's a connection
 - there's been a handful of requests
 - each request has been delivered to the user handler and a response has been provided back and the Demux stage has pushed it so it's now on the network
 - !! At this point there is no open HTTP/2 stream
 - a new request comes in
 - an HTTP/" stream is created
 - the server tries to shutdown or the connection is closing
 - the completing signal comes into Demux via the `frameIn` inlet
 - the completion propagates and eventually drains the `BufferedOutlet` 
 - the user handler provides a response for the in-flight request
 - the completion signal travels to the user handler and comes back via the `substreamIn` inlet
 - the new `tryComplete` is invoked and sets `completing` to `true`
 - at this point there may be a HTTP/2 stream in the `Http2StreamHandling` set of open streams, or maybe the response already traversed that part and the stream is already closed
 - eventually, the last stream in `Http2StreamHandling` is closed and `onAllStreamsClosed` is invoked
 - the invocation to `onAllStreamsClosed` cancels the inlets and completes the outlets

Is this the correct thinking?

ignasi35:
I just wrote a test on `Http2ServerSpec` and I am, indeed, getting `OnComplete` before getting `OnNext` in the `toNet` probe. I'm missing something.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Demux.scala
```diff
@@ -357,25 +362,57 @@ private[http2] abstract class Http2Demux(http2Settings: Http2CommonSettings, ini
...
+      private def onTryComplete(): Unit = {
+        completing = true
+        if (safeToComplete) complete()
+      }
+      override def onAllStreamsClosed(): Unit = if (safeToComplete) complete()
+      override def onMultiplexerIdle(): Unit = if (safeToComplete) complete()
+
+      private def safeToComplete = {
+        if (isServer) completing && multiplexer.isIdle && activeStreamCount() == 0
```
raboof, line 382: Instead of waiting for `multiplexer.isIdle` before declaring `safeToComplete` and calling `complete()`, I wonder if we could call `complete()` immediately but replace `complete(frameOut)` by a (to be implemented) `multiplexer.complete()`, and make the multiplexer responsible for eventually completing `frameOut` - similar to what we did for `bufferedSubStreamOutput`.

ignasi35:
Took me a bit to align all bits and pieces. I ended up moving the timeout handling on the Demux (https://github.com/akka/akka-http/pull/3744) to this PR to ensure all stages complete (which is required in tests).

I'll rebase to solve the conflicts.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/client/PersistentConnection.scala
```diff
@@ -185,11 +185,9 @@ private[http2] object PersistentConnection {
         }
         override def onPull(): Unit = responseIn.pull()
 
-        override def onUpstreamFinish(): Unit = {
+        override def onUpstreamFinish(): Unit =
           requestOut.complete()
-          responseIn.cancel()
-          completeStage()
-        }
```
ignasi35, line 192: `PersistentConnection` should not cancel `responseIn` and complete the stage so eagerly.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Demux.scala
```diff
@@ -357,25 +367,64 @@ private[http2] abstract class Http2Demux(http2Settings: Http2CommonSettings, ini
...
         }
+
+        override def onUpstreamFinish(): Unit = {
+          // marks StreamHandling as ready for completion
+          onTryComplete()
+          scheduleOnce(CompletionTimeout, settings.completionTimeout)
+        }
+
+        override def onUpstreamFailure(ex: Throwable): Unit = {
+          onTryComplete()
```
jrudolph, line 420: This converts the failure to regular completion when the completion condition holds. Is that what we want?

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -1326,6 +1326,42 @@ class Http2ServerSpec extends AkkaSpecWithMaterializer("""
...
+        network.fromNet.sendComplete()
+
+        user.emitResponse(1, HPackSpecExamples.FirstResponse)
+        val headerPayload = network.expectHeaderBlock(1)
+        headerPayload shouldBe HPackSpecExamples.C61FirstResponseWithHuffman
+        network.expectComplete()
+
+      }
+
+      //      "actually complete as soon as required if there are no open (in-flight) streams" inAssertAllStagesStopped new TestSetup with NetProbes {
```
jrudolph, line 1346: Is this still WIP?

akka-http-core/src/main/scala/akka/http/impl/engine/http2/client/PersistentConnection.scala
jrudolph: We should document here, that we here require "reasonable behavior" from downstream stages, i.e. that the downstream stage will eventually close all remaining inputs/outputs.

(A safer behavior would be to add another completion timeout here, but maybe we can leave that for now).

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Multiplexer.scala
```diff
@@ -121,12 +127,16 @@ private[http2] trait Http2MultiplexerSupport { logic: GraphStageLogic with Stage
 
       private var _state: MultiplexerState = Idle
 
+      def hasFlushedAllData: Boolean = allDataFlushed(_state)
+      private def allDataFlushed(state: MultiplexerState): Boolean = (state == WaitingForData || state == Idle)
```
jrudolph, line 131: :+1: important to have that bit to avoid losing the last few frames.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Demux.scala
```diff
@@ -332,6 +337,11 @@ private[http2] abstract class Http2Demux(http2Settings: Http2CommonSettings, ini
           pullFrameIn()
         }
 
+        override def onUpstreamFinish(): Unit = {
+          onTryComplete()
```
jrudolph, line 341: Do we need this? If we get completion from the connection/TLS it's not likely that we will be able to send anything else. So maybe better to keep it simple.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Demux.scala
jrudolph: This might be because of the `onUpstreamFinish` for `frameIn`. If we can remove that handling, we should also remove this one.

jrudolph (commented):
A few more comments

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Demux.scala
ignasi35: That's a leftover.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
ignasi35: I've implemented it.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Demux.scala
ignasi35: As discussed on the4 call, I think we can/have_to keep this.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/client/PersistentConnection.scala
ignasi35: > We should document here, that we here require "reasonable behavior" from downstream stages, i.e. that the downstream stage will eventually close all remaining inputs/outputs.

Will do.

> (A safer behavior would be to add another completion timeout here, but maybe we can leave that for now).

I think it's safe to leave that out for now.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Demux.scala
jrudolph: That's not quite what I meant :)

Without this the new tests in Http2ServerSpec are failing, yes. But I don't think we should change the behavior for incoming completion of the network for now. As I said before half-closed TLS/HTTP/2 connections are not really supported. In any case, the peer would have to send a GOAWAY first before closing the stream.

When I commented about server behavior on a previous review I meant something else, i.e. a potential issue on the server side if the user side closes the response side of the handler which will now go into the special shutdown state with no opportunity to actually complete potentially open streams which are still waiting for a response.

jrudolph: I will look into those issue to get this PR done.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Demux.scala
jrudolph: That part of the complexity is also due to the handling of the `frameIn` `onUpstreamFinish` handler. Only when called from there it can be that `completing && !isClosed(substreamIn)`.

jrudolph:
Added a commit that removes the `frameIn` handling, removes the server tests and simplifies things a bit. Now working on a new test for the server side.

jrudolph:
I came to the conclusion that we shouldn't change the server behavior in any way here so I added a condition to keep the old behavior on the server and only apply the new behavior on the client.

jrudolph:
I also fixed an edge case on the client side that shows that we also need that flushing behavior for the multiplexer on the client side as well.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Demux.scala
jrudolph: This is indeed still needed even without the `frameIn` handling.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2PersistentClientSpec.scala
```diff
@@ -63,197 +97,160 @@ class Http2PersistentClientSpec extends AkkaSpecWithMaterializer(
...
-          Unmarshal(response.entity).to[String].futureValue shouldBe "pong"
-          response.attribute(requestIdAttr).get.id shouldBe "request-1"
-        }
+        val serverRequest2 = server.expectRequest()
+        serverRequest2.entityAsString shouldBe "ping2"
+        // request should have come in through another connection
+        serverRequest2.clientPort should not be (clientPort)
       }
-      "eventually fail" should {
-        "when connecting keeps failing" inAssertAllStagesStopped new TestSetup(tls) {
```
jrudolph, line 223: @ignasi35, why did you delete this test?

jrudolph:
This is now functionally done from my side. There are still two remaining issues (which I can resolve on Monday):

 * one test was removed from Http2PersistentConnectionSpec but it's not clear to me why
 * We now introduced a `completion-timeout` for the server side but I removed its behavior, we should therefore probably also remove the setting again? That will need a minor refactoring to put the setting into the right place.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2PersistentClientSpec.scala
ignasi35: This was a mistake. The tests must be restored (sending a commit now...).

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
```diff
@@ -753,6 +768,158 @@ class Http2ClientSpec extends AkkaSpecWithMaterializer("""
...
+          HttpEntity.Strict(ContentTypes.`application/octet-stream`, ByteString("Hello World 7!"))
+        )
+
+        // Completion doesn't happen until entities are consumed
+        user.responseIn.expectComplete()
+
+      }
+
+      "until in-flight streaming requests are fully sent (after response is already complete) (uncommon)" inAssertAllStagesStopped new TestSetup(
+        Setting(SettingIdentifier.SETTINGS_MAX_CONCURRENT_STREAMS, 3)
```
ignasi35, line 842: we can avoid this SETTING for this test.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
```diff
@@ -753,6 +768,158 @@ class Http2ClientSpec extends AkkaSpecWithMaterializer("""
...
+        requestStreamProbe.sendNext(ByteString("almost"))
+        requestStreamProbe.sendNext(ByteString(" done"))
+        requestStreamProbe.sendComplete()
+        network.expectDATA(1, endStream = true, ByteString("almost done"))
+
+        // Completion doesn't happen until entities are sent
+        user.responseIn.expectComplete()
+        network.toNet.expectComplete()
+        network.fromNet.expectCancellation()
+      }
```
ignasi35, line 877: This is a nice test. I didn't think of the scenario and making the client await until both outgoing and incoming streams complete is the most reasonable behavior.

ignasi35 (commented):
LGTM

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ClientSpec.scala
jrudolph: :+1:

jrudolph:
> We now introduced a `completion-timeout` for the server side but I removed its behavior, we should therefore probably also remove the setting again? That will need a minor refactoring to put the setting into the right place.

Fixed this last remaining bit.

jrudolph:
Finished the last bits.

PR #3735, opened 2021-01-13
[h2 client] 'connecting' phase for persistent connections

raboof:
Avoids responding with 502 responses when we fail to connect to a host at all, so we can retry (possibly to a different endpoint).

Test still fails: the reconnection seems to happen, but no request is/becomes available on `requestIn` - I can't explain that yet...

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/client/PersistentConnection.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/client/PersistentConnection.scala
@@ -6,10 +6,14 @@ package akka.http.impl.engine.http2.client
 
 import akka.NotUsed
 import akka.annotation.InternalApi
+import akka.http.scaladsl.Http.OutgoingConnection
 import akka.http.scaladsl.model.{ AttributeKey, HttpRequest, HttpResponse, RequestResponseAssociation, StatusCodes }
-import akka.stream.scaladsl.{ Flow, Source }
+import akka.stream.scaladsl.{ Flow, Keep, Source }
 import akka.stream.stage.{ GraphStage, GraphStageLogic, InHandler, OutHandler, StageLogging }
 import akka.stream.{ Attributes, FlowShape, Inlet, Outlet }
+
+import scala.concurrent.Future
+import scala.util.{ Failure, Success }
 
 /** INTERNAL API */
 @InternalApi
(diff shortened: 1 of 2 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/http2/client/PersistentConnection.scala
```diff
@@ -57,18 +61,73 @@ private[http2] object PersistentConnection {
...
 
-        Source.fromGraph(requestOut.source)
-          .via(connectionFlow)
-          .runWith(responseIn.sink)(subFusingMaterializer)
+        var requestOutPulled = false
+        requestOut.setHandler(new OutHandler {
+          override def onPull(): Unit =
+            requestOutPulled = true
+          override def onDownstreamFinish(): Unit = ()
+        })
```
johanandren, line 95: We have already materialized the SubSourceOutlet, what does that mean for setting the handler after that is the pull lost perhaps?

akka-http-core/src/main/scala/akka/http/impl/engine/http2/client/PersistentConnection.scala
```diff
@@ -55,20 +59,74 @@ private[http2] object PersistentConnection {
...
+        requestOut: SubSourceOutlet[HttpRequest],
+        responseIn: SubSinkInlet[HttpResponse]
+      ) extends State {
+        implicit val ec = materializer.executionContext
+        connected.onComplete {
+          case Success(_) =>
+            onConnected.invoke(())
+          case Failure(_) =>
+            onFailed.invoke(())
+        }
```
johanandren, line 90: Probably does not matter much but this could be done on parasitic/samethread since async callback invoke should never fail.

johanandren (commented):
Pretty much LGTM if we don't want to add a cap on failures to connect without a success inbetween somehow. Perhaps nice to treat that as a separate PR?

akka-http-core/src/main/scala/akka/http/impl/engine/http2/client/PersistentConnection.scala
raboof: makes sense, done

raboof:
> Perhaps nice to treat that as a separate PR?

Yes, good idea

johanandren (approved):

ignasi35 (approved):

PR #3736, opened 2021-01-14
Race condition in TelemetrySpiSpec fix

johanandren:


raboof (approved):
makes sense for close-seen and done to be concurrent :+1:

Issue #2648, opened 2019-08-15
Add flag to log WS frame traffic


jrudolph:
Similar as done for HTTP/2 a while ago in #2561.

nitikagarw:
Hi @jrudolph, 
Looks like it requires similar changes in `WebSocketSettings.scala`. If yes, Can I please pick this issue?
Thanks, 
Nitika Agarwal

jrudolph:
Yes, please! (sorry for the delay over the change of the year)

jrudolph:
Implemented in #3739

PR #3739, opened 2021-01-18
core: add flag to log WebSocket frame traffic

nitikagarw:
References #2648


raboof:
OK TO TEST

ignasi35 (commented):
Looking good. 

Do you plan on adding the `log-frames` setting for `akka.http.client.websocket` aswell? (it's only on `akka.http.client` at the moment :-) )

nitikagarw:
Hi @jrudolph, 
I would need some help to figure out the right function to enable the Frame logging as [it has been done for HTTP2](https://github.com/akka/akka-http/blob/master/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala#L96).
I think it has to somewhere in [WebSocket](https://github.com/akka/akka-http/blob/master/akka-http-core/src/main/scala/akka/http/impl/engine/ws/WebSocket.scala) or [WebSocketClientBlueprint](https://github.com/akka/akka-http/blob/master/akka-http-core/src/main/scala/akka/http/impl/engine/ws/WebSocketClientBlueprint.scala)

johanandren:
Try somewhere in the blueprint stack: https://github.com/akka/akka-http/blob/master/akka-http-core/src/main/scala/akka/http/impl/engine/ws/WebSocketClientBlueprint.scala#L46

johanandren (commented):
Looks good to go except for the mima filters

akka-http-core/src/main/mima-filters/10.1.8.backwards.excludes
```diff
@@ -17,6 +17,9 @@ ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.javadsl.settings
 ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.scaladsl.settings.Http2ServerSettings.logFrames")
 ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.scaladsl.settings.Http2ServerSettings.withLogFrames")
 ProblemFilters.exclude[IncompatibleSignatureProblem]("akka.http.scaladsl.settings.Http2ServerSettings#Http2ServerSettingsImpl.unapply")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.javadsl.settings.WebSocketSettings.logFrames")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.javadsl.settings.WebSocketSettings.withLogFrames")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.scaladsl.settings.WebSocketSettings.logFrames")
```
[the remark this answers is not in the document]
jrudolph: Yep, so can be removed here for now.

akka-http-core/src/main/mima-filters/10.2.0.backwards.excludes/websocket-log-frames-flag.excludes
```diff
@@ -0,0 +1,10 @@
+# Extension of @DoNotInherit class
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.javadsl.settings.WebSocketSettings.logFrames")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.javadsl.settings.WebSocketSettings.withLogFrames")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.scaladsl.settings.WebSocketSettings.logFrames")
+ProblemFilters.exclude[IncompatibleSignatureProblem]("akka.http.impl.settings.WebSocketSettingsImpl.unapply")
+
+# Internal class changes
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.impl.settings.WebSocketSettingsImpl.apply")
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.impl.settings.WebSocketSettingsImpl.copy")
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.impl.settings.WebSocketSettingsImpl.this")
```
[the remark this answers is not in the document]
jrudolph: Ah, I see you added things to many of those files. In general, exclusions are only necessary once for the latest released version. Mima knows how to treat them for older versions.

jrudolph (commented):
Looking good, can you post an example of how the output looks like? Have you checked that `DOWN` and `UP` have the same direction as for HTTP/2? It seems the Websocket stack is built in the other direction as in HTTP/2, so it might be the wrong way round (I didn't try it, you might have checked for that already).

akka-http-core/src/main/mima-filters/10.2.0.backwards.excludes/websocket-log-frames-flag.excludes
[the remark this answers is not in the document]
nitikagarw: Thanks, removed.

nitikagarw:
HI @jrudolph, 
I am having some issues testing the output generated. 
I ran the HTTP2 test cases as below but didn't see any output. Maybe there are some settings which I need to do?
`akka-http-root > testOnly akka.http.impl.engine.http2.Http2ClientServerSpec`

johanandren:
`AkkaSpecWithMaterializer` pulls in logic to silence all logs unless a test fails (`WithLogCapturing`), you can temporarily add a `fail()` in a test to get the log output.

nitikagarw:
Thanks, @johanandren.
Here is the screenshot of the logs
Screenshot 2021-01-26 at 9 45 21 PM

akka-http-core/src/main/scala/akka/http/impl/engine/ws/FrameLogger.scala
```diff
@@ -0,0 +1,57 @@
...
+      val ellipsis = if (num < bytes.size) s" [... ${bytes.size - num} more bytes]" else ""
+      bytes
+        .take(num)
+        .map(_ formatted "%02x")
+        .mkString(" ") + ellipsis
+    }
+
+    def entryForFrame(frameEvent: FrameEvent): LogEntry =
+      frameEvent match {
+        case FrameStart(_, data) => LogEntry("FRAMESTART", hex(data))
```
jrudolph, line 46: The important thing here is to also log the frame header:

```
private[http] final case class FrameHeader(
  opcode: Protocol.Opcode,
  mask:   Option[Int],
  length: Long,
  fin:    Boolean,
  rsv1:   Boolean         = false,
  rsv2:   Boolean         = false,
  rsv3:   Boolean         = false)
```

 * The Opcode should be a textual representation of the opcode, maybe just the `toString.toUperString` for now. It can be used instead of `FRAMESTART`.
 * At the point where it's logged, no mask should be set, so that can be ignored.
 * Length should be shown before the actual data
 * The flags can be simple strings that are only shown if the flag is set, so add `FIN` `RSV1`, etc when the flag is set and leave it out otherwise.

jrudolph: Also it should add some indication if all data is contained in that event or not (i.e. the value of `FrameStart.lastPart`)

akka-http-core/src/main/scala/akka/http/impl/engine/ws/FrameLogger.scala
```diff
@@ -0,0 +1,57 @@
...
+        case FrameData(data, _)  => LogEntry("FRAMEDATA", hex(data))
```
jrudolph, line 47: This event is basically a continuation of the `FrameStart` one (i.e. it has no correspondence with anything in the WS specs, it's more about how frames are represented in akka-http).

The text should indicate that it's a continuation of a previous frame and whether more data is expected. Maybe we could just use ellipsis for that?

jrudolph (commented):
Looking pretty good. I'd suggest to add more detail to the messages as commented below.

It would be nice to interpret some of the data if possible, i.e. show the textual representation especially for `Text` frames. The difficulty with that is that data might be split over several `FrameStart / FrameData` events and multiple frames might be split using `Text` and `Continuation` frames. They might be split in a way that UTF-8 codepoints could be split between different events which makes it somewhat hard to do the decoding properly. So, maybe we would leave that part for another improvement in an extra PR.

akka-http-core/src/main/scala/akka/http/impl/engine/ws/FrameLogger.scala
nitikagarw: Updated the PR, The logs now look like below: 
Screenshot 2021-02-03 at 12 09 16 AM

akka-http-core/src/main/scala/akka/http/impl/engine/ws/FrameLogger.scala
nitikagarw: Are you suggesting to just use the ellipsis for `FrameData` ?

jrudolph: Yes, I made that change.

jrudolph:
Hi @nitikagarw, sorry for my long absence, I had to take a hiatus from work during the first half of the year and have only now come back to the backlog of PRs.

This was already looking quite good. I did a few changes to get it over the finish line:

 * changed the integration point so that the logger applies both to client and to server more easily
 * changed a bit of the display logic to also show ASCII representation
 * show `lastPart` as `...` after the frame and `FrameData` as type `...` as suggested before
 * fixed mima because we released 10.2.4 in the mean time

akka-http-core/src/main/scala/akka/http/impl/engine/ws/WebSocket.scala
```diff
@@ -40,6 +40,7 @@ private[http] object WebSocket {
     closeTimeout:      FiniteDuration    = 3.seconds, // TODO put close timeout into the settings?
     log:               LoggingAdapter): BidiFlow[FrameEvent, Message, Message, FrameEvent, NotUsed] =
     masking(serverSide, websocketSettings.randomFactory) atop
+      FrameLogger.logFramesIfEnabled(websocketSettings.logFrames) atop
```
jrudolph, line 43: To fit here, the type had to be adapted to accept `FrameEventOrError`. The benefit is that frames are shown unmasked.

akka-http-core/src/test/scala/akka/http/impl/engine/ws/WebSocketIntegrationSpec.scala
```diff
@@ -26,10 +27,14 @@ import akka.testkit._
 
 import scala.util.{ Failure, Success }
 
-class WebSocketIntegrationSpec extends AkkaSpec("akka.stream.materializer.debug.fuzzing-mode=off")
-  with Eventually {
-
-  implicit val materializer = ActorMaterializer()
+class WebSocketIntegrationSpec extends AkkaSpecWithMaterializer(
```
jrudolph, line 30: Needs to be `AkkaSpecWithMaterializer` to log automatically when a test fails.

akka-http-core/src/main/scala/akka/http/impl/engine/ws/FrameLogger.scala
jrudolph: The `lastPart` / `T` / `F` rendering was redundant as it should already included in the flags rendering.

johanandren (approved):

jrudolph:
Thanks, @nitikagarw!

nitikagarw:
Thanks a lot, @jrudolph for taking this PR to completion and @johanandren for the review.

PR #3744, opened 2021-01-18
[http2] Timeout and complete when in-flight responses don't finish

ignasi35:
Targets #3734


ignasi35:
Once #3734 is merged I'll rebase and resolve cronflicts.

ignasi35:
moved this improvement to #3734

PR #3750, opened 2021-01-23
Update current year in CopyrightHeader

nitikagarw:
References #xxxx

### DIFF
--- a/project/CopyrightHeader.scala
+++ b/project/CopyrightHeader.scala
@@ -33,7 +33,7 @@ object CopyrightHeader extends AutoPlugin {
 
   // We hard-code this so PR's created in year X will not suddenly fail in X+1.
   // Of course we should remember to update it early in the year.
-  val CurrentYear = "2020"
+  val CurrentYear = "2021"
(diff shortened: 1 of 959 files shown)


project/CopyrightHeader.scala
```diff
@@ -33,7 +33,7 @@ object CopyrightHeader extends AutoPlugin {
 
   // We hard-code this so PR's created in year X will not suddenly fail in X+1.
   // Of course we should remember to update it early in the year.
-  val CurrentYear = "2020"
+  val CurrentYear = "2021"
```
hepin1989, line 36: use Date?

raboof: As noted in the comment above this line, we prefer to hard-code the year, because otherwise all PR's that are open on December 31st and master suddenly start failing on January 1st

raboof (commented):
you'll probably also have to run `sbt headerCreate test:headerCreate` to update the actual headers in files.

raboof:
OK TO TEST

raboof (approved):
Thanks!

PR #3691, opened 2020-12-04
Avoid temporaryServerHostnameAndPort

raboof:
Since it has an inherent race condition.

Refs #3679. I thought we had an 'umbrella ticket' for this
but can't find it right now.

Issue #2947, opened 2020-02-06
Get rid of SocketUtils.temporaryHostnameAndPort


jrudolph:
Because, it has an inherent race condition and makes tests fail more easily.

jrudolph:
#3460 hopefully alleviates the situation for now since a bug was found that makes `SocketUtils.temporaryHostnameAndPort` fail on its own (i.e. without even taking the race condition into account).

PR #3751, opened 2021-01-23
tests: get rid of SocketUtils.temporaryHostnameAndPort in some tests

nitikagarw:
References #2947
Relates to https://github.com/akka/akka-http/pull/3691

### DIFF
--- a/akka-http-core/src/test/scala/akka/http/scaladsl/ClientServerSpec.scala
+++ b/akka-http-core/src/test/scala/akka/http/scaladsl/ClientServerSpec.scala
@@ -82,7 +82,8 @@ class ClientServerSpec extends AkkaSpecWithMaterializer(
     }
 
     "report failure if bind fails" in EventFilter[BindException](occurrences = 2).intercept {
-      val binding = Http().newServerAt("localhost", 0).connectionSource()
+      val port = 1025 // non-root users typically can't bind to port numbers below 1024
+      val binding = Http().newServerAt("localhost", port).connectionSource()
       val probe1 = TestSubscriber.manualProbe[Http.IncomingConnection]()
       // Bind succeeded, we have a local address
       val b1 = Await.result(binding.to(Sink.fromSubscriber(probe1)).run(), 3.seconds.dilated)
--- a/akka-http-core/src/test/scala/akka/http/scaladsl/ClientServerSpec.scala
+++ b/akka-http-core/src/test/scala/akka/http/scaladsl/ClientServerSpec.scala
@@ -82,8 +82,7 @@ class ClientServerSpec extends AkkaSpecWithMaterializer(
     }
 
     "report failure if bind fails" in EventFilter[BindException](occurrences = 2).intercept {
-      val (hostname, port) = SocketUtil.temporaryServerHostnameAndPort()
-      val binding = Http().newServerAt(hostname, port).connectionSource()
+      val binding = Http().newServerAt("localhost", 0).connectionSource()
       val probe1 = TestSubscriber.manualProbe[Http.IncomingConnection]()
       // Bind succeeded, we have a local address
       val b1 = Await.result(binding.to(Sink.fromSubscriber(probe1)).run(), 3.seconds.dilated)
@@ -903,7 +902,6 @@ Host: example.com
 
   class TestSetup {
     val hostname = "localhost"
-    //val (hostname, port) = SocketUtil.temporaryServerHostnameAndPort()
     def configOverrides = ""
 
     // automatically bind a server
--- a/akka-http-tests/src/multi-jvm/scala/akka/http/AkkaHttpServerLatencyMultiNodeSpec.scala
+++ b/akka-http-tests/src/multi-jvm/scala/akka/http/AkkaHttpServerLatencyMultiNodeSpec.scala
@@ -14,7 +14,7 @@ import akka.http.scaladsl.server.{ Directives, Route }
 import akka.http.scaladsl.Http
 import akka.remote.testkit.{ MultiNodeConfig, MultiNodeSpec }
 import akka.stream.scaladsl.Source
-import akka.testkit.{ ImplicitSender, LongRunningTest, SocketUtil }
+import akka.testkit.{ ImplicitSender, LongRunningTest }
 import akka.util.{ ByteString, Timeout }
 import com.typesafe.config.ConfigFactory
 import org.scalatest.concurrent.ScalaFutures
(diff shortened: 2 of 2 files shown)


akka-http-core/src/test/scala/akka/http/scaladsl/ClientServerSpec.scala
```diff
@@ -82,8 +82,7 @@ class ClientServerSpec extends AkkaSpecWithMaterializer(
     }
 
     "report failure if bind fails" in EventFilter[BindException](occurrences = 2).intercept {
-      val (hostname, port) = SocketUtil.temporaryServerHostnameAndPort()
-      val binding = Http().newServerAt(hostname, port).connectionSource()
+      val binding = Http().newServerAt("localhost", 0).connectionSource()
```
nitikagarw, line 85: I think we probably need a non-zero port here for this test to work?
I tried passing port = `1` but it gave me `java.net.SocketException: Permission denied` instead of `BindException`.

raboof: non-root users typically can't bind to port numbers below 1024, so a larger number might work?

nitikagarw: Thanks, @raboof. It worked.

raboof:
OK TO TEST

akka-http-core/src/test/scala/akka/http/scaladsl/ClientServerSpec.scala
```diff
@@ -903,7 +903,6 @@ Host: example.com
 
   class TestSetup {
     val hostname = "localhost"
-    //val (hostname, port) = SocketUtil.temporaryServerHostnameAndPort()
```
raboof, line 906: remove the line entirely?

akka-http-core/src/test/scala/akka/http/scaladsl/ClientServerSpec.scala
```diff
@@ -82,8 +82,8 @@ class ClientServerSpec extends AkkaSpecWithMaterializer(
     }
 
     "report failure if bind fails" in EventFilter[BindException](occurrences = 2).intercept {
-      val (hostname, port) = SocketUtil.temporaryServerHostnameAndPort()
-      val binding = Http().newServerAt(hostname, port).connectionSource()
+      val port = 1025 // non-root users typically can't bind to port numbers below 1024
```
raboof, line 85: hmm, so this test will no longer succeed when something is bound to port 1025... *might* be OK?

akka-http-tests/src/multi-jvm/scala/akka/http/AkkaHttpServerLatencyMultiNodeSpec.scala
```diff
@@ -176,7 +176,7 @@ class AkkaHttpServerLatencyMultiNodeSpec extends MultiNodeSpec(AkkaHttpServerLat
         enterBarrier("load-gen-ready")
 
         runOn(server) {
-          val (_, port) = SocketUtil.temporaryServerHostnameAndPort()
+          val port = 0
```
raboof, line 179: this means we'll do `setServerPort(0)` later - I'd say that probably would make tests that try to use that port fail? Perhaps it'd be OK to wait for the binding to complete here and pass in the actual port.

akka-http-core/src/test/scala/akka/http/scaladsl/ClientServerSpec.scala
nitikagarw: yes, this commented line was present in the code and I have removed it.

raboof: oops, sorry, good change indeed :D

akka-http-core/src/test/scala/akka/http/scaladsl/ClientServerSpec.scala
nitikagarw: I think it should be okay because even if we pick any random port, there might be a possibility that something is bound to that random port?

raboof:
(I think jenkins succeeded here because it does not run the multi-jvm tests, those are only tested locally or in nightly builds)

akka-http-tests/src/multi-jvm/scala/akka/http/AkkaHttpServerLatencyMultiNodeSpec.scala
nitikagarw: Are you suggesting something like this: `setServerPort(_binding.map(_.localAddress.getPort).getOrElse(port))`?

akka-http-core/src/test/scala/akka/http/scaladsl/ClientServerSpec.scala
raboof: I agree picking a random port won't help. Using `temporaryServerHostnameAndPort` is a little better because it'll pick a port that at least was free right before the binding, but still leaves a race condition. Using port '0' for the first bind, and then binding the second materialization to that same port would be the most reliable solution, but also a somewhat more complicated refactor. As said, I guess perhaps using 1025 is good enough.

akka-http-tests/src/multi-jvm/scala/akka/http/AkkaHttpServerLatencyMultiNodeSpec.scala
raboof: No, since the `getOrElse(port)` would still produce an unusable port. I'm not sure we would get away with `_binding.futureValue.localAddress.getPort` here?

nitikagarw: Changed to:

```         
val binding = Http().newServerAt("0.0.0.0", port).bind(routes).futureValue
_binding = Some(binding)
setServerPort(binding.localAddress.getPort)
```

akka-http-tests/src/multi-jvm/scala/akka/http/AkkaHttpServerLatencyMultiNodeSpec.scala
```diff
@@ -176,12 +176,12 @@ class AkkaHttpServerLatencyMultiNodeSpec extends MultiNodeSpec(AkkaHttpServerLat
...
           info(s"Binding Akka HTTP Server to port: $port @ ${myself}")
-          val futureBinding = Http().newServerAt("0.0.0.0", port).bind(routes)
+          val binding = Http().newServerAt("0.0.0.0", port).bind(routes).futureValue
 
-          _binding = Some(futureBinding.futureValue)
```
raboof, line 183: oh I didn't notice before this was already here - looks like a safe change then :+1:

raboof (approved):
looks OK to me

jrudolph (approved):
LGTM, thanks good to get rid of a few more instances.

PR #3754, opened 2021-02-08
core: disallow parsing of Transfer-Encoding other than chunked

johanandren:


johanandren:
Missing test coverage for responses.

ignasi35 (approved):

jrudolph (commented):
I think we can be even more strict than that. Basically, akka-http is implementing the "Transfer", so whatever we cannot decode, cannot be decoded.

The part that ends up in the entity should be the "Content", i.e. the representation that is described by the `Content-...` headers.

Especially in a proxy case, preserving any additional Transfer-Encodings seems dangerous, as we might be missing something to allow a fully functional round-trip in that case, or the user misses something trying to inspect a half-decoded entity.

So, I would suggest, for now we only allow a single `chunked` encoding here and fail everything else.

Btw. the current HTTP spec is [RFC 7230](https://tools.ietf.org/html/rfc7230#section-3.3).

The question is whether we need an escape hatch? I'm somewhat inclined to be lazy about it and wait until someone complains and explains why we should support their custom use-case (or contributes that feature). WDYT?

akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpMessageParser.scala
```diff
@@ -178,8 +178,19 @@ private[http] trait HttpMessageParser[Output >: MessageOutput <: ParserOutput] {
             }
         }
         case h: `Transfer-Encoding` => teh match {
-          case None    => parseHeaderLines(input, lineEnd, headers, headerCount + 1, ch, clh, cth, Some(h), e100c, hh)
-          case Some(x) => parseHeaderLines(input, lineEnd, headers, headerCount, ch, clh, cth, Some(x append h.encodings), e100c, hh)
+          case None =>
+            h.encodings match {
+              case Seq(TransferEncodings.chunked) =>
+                // A single chunked is the only one we support
+                parseHeaderLines(input, lineEnd, headers, headerCount + 1, ch, clh, cth, Some(h), e100c, hh)
```
jrudolph, line 185: We could change the interface here to avoid any ambiguities and make that flag a `isChunked: Boolean`. In any case, the `Content-Length` vs. `Transfer-Encoding` check in `parseEntity` needs to be changed, so that the check will work for any Transfer-Encoding header passed to the method.

akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpMessageParser.scala
```diff
@@ -178,8 +178,19 @@ private[http] trait HttpMessageParser[Output >: MessageOutput <: ParserOutput] {
...
+              case Seq(unknown) =>
+                failMessageStart(s"HTTP unsupported Transfer-Encoding entry '${unknown.name}'")
```
jrudolph, line 187:
```suggestion
                failMessageStart(s"Unsupported Transfer-Encoding '${unknown.name}'")
```

akka-http-core/src/test/scala/akka/http/impl/engine/parsing/RequestParserSpec.scala
```diff
@@ -667,6 +654,51 @@ abstract class RequestParserSpec(mode: String, newLine: String) extends AnyFreeS
           |
           |""" should parseToError(422: StatusCode, ErrorInfo("TRACE requests must not have an entity"))
       }
+
+      "a request with an unsupported transfer encoding" in new Test {
```
jrudolph, line 658: Is there already a test checking that `Transfer-Encoding` and `Content-Length` must not be mixed? In any case, it might make sense to explicitly also add a test with multiple `Transfer-Encoding` entries and also a `Content-Length` header (in case we later add back some support for other encodings, it should then still fail but maybe with a different error message).

akka-http-core/src/test/scala/akka/http/impl/engine/parsing/ResponseParserSpec.scala
```diff
@@ -354,6 +327,21 @@ abstract class ResponseParserSpec(mode: String, newLine: String) extends AnyFree
...
       }
+
+      "multiple transfer encodings in one header" in new Test {
+        """HTTP/1.1 200 OK
+          |Transfer-Encoding: fancy, chunked
+          |
+          |""" should parseToError(MessageStartError(BadRequest, ErrorInfo("Multiple HTTP Transfer-Encoding entries not supported")))
+      }
+
+      "multiple transfer encoding headers" in new Test {
```
jrudolph, line 338: Ok, for having less response tests, but we should also have one for the `Transfer-Encoding` / `Content-Length` combination.

akka-http-core/src/test/scala/akka/http/impl/engine/parsing/RequestParserSpec.scala
johanandren: I'l see if there is one and add one if not.

akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpMessageParser.scala
johanandren: Went with changing it into a boolean since that is more explicit about what is actually supported.

One curious thing now is that we support rendering with whatever Transfer-Encoding but not parsing, but maybe that is to be expected?

akka-http-core/src/test/scala/akka/http/impl/engine/parsing/ResponseParserSpec.scala
johanandren: Added

akka-http-core/src/test/scala/akka/http/impl/engine/parsing/RequestParserSpec.scala
johanandren: Added

akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpMessageParser.scala
johanandren: Needed changes in tests as well, fixed in latest commit

akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpResponseParser.scala
```diff
@@ -180,41 +180,37 @@ private[http] class HttpResponseParser(protected val settings: ParserSettings, p
...
+                    val data = entityParts.collect { case EntityPart(bytes) => bytes }
+                    HttpEntity.CloseDelimited(contentType(cth), data)
+                  }
+                }
+                setCompletionHandling(HttpMessageParser.CompletionOk)
+                parseToCloseBody(input, bodyStart, totalBytesRead = 0)
+            }
+          } else {
+            if (clh.isEmpty) {
+              emitResponseStart(chunkedEntity(cth), headers)
```
johanandren, line 210: Not 100% sure about just passing the headers here, vs addTransferEncodingWithChunkedPeeled like before, but it's now signalled by the boolean so I _think_ it is right?

Same logic in the `HttpRequestParser`

akka-http-core/src/main/mima-filters/10.2.3.backwards.excludes/stricter-transfer-encoding-header-parsing.excludes
```diff
@@ -0,0 +1,10 @@
+# parsing internals only
+ProblemFilters.exclude[IncompatibleMethTypeProblem]("akka.http.impl.engine.parsing.HttpResponseParser.parseHeaderLines")
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.impl.engine.parsing.HttpResponseParser.addTransferEncodingWithChunkedPeeled")
+ProblemFilters.exclude[IncompatibleResultTypeProblem]("akka.http.impl.engine.parsing.HttpResponseParser.parseHeaderLines$default$8")
+ProblemFilters.exclude[IncompatibleMethTypeProblem]("akka.http.impl.engine.parsing.HttpResponseParser.parseEntity")
+ProblemFilters.exclude[IncompatibleMethTypeProblem]("akka.http.impl.engine.parsing.HttpMessageParser.parseEntity")
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.impl.engine.parsing.HttpMessageParser.parseHeaderLines")
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.impl.engine.parsing.HttpMessageParser.parseHeaderLines$default$8")
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.impl.engine.parsing.HttpMessageParser.addTransferEncodingWithChunkedPeeled")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.impl.engine.parsing.HttpMessageParser.parseEntity")
```
jrudolph, line 10:
```suggestion
ProblemFilters.exclude[Problem]("akka.http.impl.engine.parsing.*")
```

akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpMessageParser.scala
```diff
@@ -149,57 +149,69 @@ private[http] trait HttpMessageParser[Output >: MessageOutput <: ParserOutput] {
...
-        case h: `Transfer-Encoding` => teh match {
-          case None    => parseHeaderLines(input, lineEnd, headers, headerCount + 1, ch, clh, cth, Some(h), e100c, hh)
-          case Some(x) => parseHeaderLines(input, lineEnd, headers, headerCount, ch, clh, cth, Some(x append h.encodings), e100c, hh)
-        }
+        case h: `Transfer-Encoding` =>
+          if (!isChunked) {
+            h.encodings match {
+              case Seq(TransferEncodings.chunked) =>
+                // A single chunked is the only one we support
+                parseHeaderLines(input, lineEnd, headers, headerCount + 1, ch, clh, cth, true, e100c, hh)
```
jrudolph, line 185:
```suggestion
                parseHeaderLines(input, lineEnd, headers, headerCount + 1, ch, clh, cth, isChunked = true, e100c, hh)
```

akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpResponseParser.scala
jrudolph: Yes, we don't provide custom `Transfer-Encoding` any more, so :+1:.

akka-http-core/src/test/scala/akka/http/impl/engine/parsing/RequestParserSpec.scala
jrudolph: Great, thanks!

jrudolph (approved):
LGTM, only some cosmetics below

akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpMessageParser.scala
jrudolph: Added https://github.com/akka/akka-http/issues/3761 to follow up on that.

akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpMessageParser.scala
jrudolph: Thanks

jrudolph:
Thanks for fixing, @johanandren.

thjaeckle:
Will this be released soon?
There is a ongoing medium CVE without a released Akka HTTP version fixing this (the `Recommendation` is wrong as this fix is not available in version `10.2.3`): https://snyk.io/vuln/SNYK-JAVA-COMTYPESAFEAKKA-1075043

mhenke1:
I there any chance to get this change into 10.1.x too? We also need to close CVE-2021-23339.

ignasi35:
> Is there any chance to get this change into 10.1.x too? We also need to close CVE-2021-23339.

See https://github.com/akka/akka-http/pull/3765 for the backport.

Issue #3717, opened 2021-01-04
Prevent the serialization of header containing a line terminator


gontard:
akka-http should throw an `Exception` when an header value containing a line terminator (LF) is sent in a response

For example

```scala
package com.example

import akka.actor.typed.ActorSystem
import akka.actor.typed.scaladsl.Behaviors
import akka.http.scaladsl.Http
import akka.http.scaladsl.model._
import akka.http.scaladsl.model.headers.{ModeledCustomHeader, ModeledCustomHeaderCompanion}
import akka.http.scaladsl.server.Directives._

import scala.util.Try

object HttpServerRoutingMinimal {
  def main(args: Array[String]): Unit = {
    implicit val system = ActorSystem(Behaviors.empty, "my-system")
    implicit val executionContext = system.executionContext
    val route =
      path("hello") {
        get {
          complete(StatusCodes.OK, Seq[HttpHeader](
            MyHeader("abc\ndef")
          ), "<h1>Say hello to akka-http</h1>")
        }
      }

    Http().newServerAt("localhost", 8080).bind(route)
  }

  final class MyHeader(token: String) extends ModeledCustomHeader[MyHeader] {
    override def renderInRequests = true
    override def renderInResponses = true
    override val companion = MyHeader
    override def value: String = token
  }
  object MyHeader extends ModeledCustomHeaderCompanion[MyHeader] {
    override val name = "my-header"
    override def parse(value: String) = Try(new MyHeader(value))
  }
}

```

We spotted this issue because our[ envoy](https://www.envoyproxy.io/) reverse proxy complains of an `Invalid HTTP header field was received`.

jrudolph:
Good point, thanks for reporting (and sorry for the delay), @gontard.

I don't have a good idea how to check this cheaply. I don't think we want that check for our own modeled headers. It would certainly make sense for `RawHeader` but not sure where to put this for more generic user-defined modeled headers. Any ideas?

gontard:
Thanks for your answer.

>  I don't think we want that check for our own modeled headers

Is it a performance concern? 

> Any ideas?

IDK the akka-http internals, but naively i would check that systematically.

nitikagarw:
Hi @jrudolph, 
How about adding the check [here][1] and [here][2]?

[1]: https://github.com/akka/akka-http/blob/master/akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala#L101
[...]

PR #3763, opened 2021-02-20
Prevent the serialization of header containing a line terminator

nitikagarw:
References #3717

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpHeader.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpHeader.scala
@@ -95,6 +95,10 @@ object HttpHeader {
       }
     } else ParsingResult.Error(ErrorInfo(s"Illegal HTTP header name", name))
 
+  def containsLineBreak(s: String): Boolean = {
+    s.contains("\n") || s.contains("\r") || s.contains("\r\n")
+  }
+
   /** INTERNAL API */
   @InternalApi
   private[akka] def fastFind[T >: Null <: jm.HttpHeader](clazz: Class[T], headers: immutable.Seq[HttpHeader]): OptionVal[T] = {
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
@@ -122,6 +122,7 @@ abstract class ModeledCustomHeader[H <: ModeledCustomHeader[H]] extends CustomHe
 
   final override def name = companion.name
   final override def lowercaseName = name.toRootLowerCase
+  
 }
 
 import akka.http.impl.util.JavaMapping.Implicits._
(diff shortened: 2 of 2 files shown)


akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
```diff
@@ -122,6 +122,7 @@ abstract class ModeledCustomHeader[H <: ModeledCustomHeader[H]] extends CustomHe
 
   final override def name = companion.name
   final override def lowercaseName = name.toRootLowerCase
+
```
nitikagarw, line 125: Hi @jrudolph , Are you suggesting something like below in this class:

```
  def apply(value: String): H = {
    companion.parse(value) match {
      case Success(parsed) =>
        if(HttpHeader.containsLineBreak(parsed.value()))
          throw new IllegalArgumentException("value must not contain line break characters")
        else parsed
      case Failure(ex)     => throw new IllegalArgumentException(s"Unable to construct custom header by parsing: '$value'", ex)
    }
  }
```

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpHeader.scala
```diff
@@ -95,6 +95,10 @@ object HttpHeader {
       }
     } else ParsingResult.Error(ErrorInfo(s"Illegal HTTP header name", name))
 
+  def containsLineBreak(s: String): Boolean = {
+    s.contains("\n") || s.contains("\r") || s.contains("\r\n")
```
jrudolph, line 99: Mmh, would be good to avoid traversing over the data multiple times. The latter check for `\r\n` should be redundant anyways.

You should be able to use `CharPredicate` for that. Maybe like that:

```scala
private val CRLF = CharPredicate('\r', '\n')

def validateHeaderValue(value: String): Boolean =
  // quote relevant RFC section here
  CRLF.matchesAny(s)
```

jrudolph (approved):
Thanks, @nitikagarw, see some comments below.

jrudolph (commented):
(sorry, didn't mean to approve yet, but it's looking good)

jrudolph:
OK TO TEST

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
nitikagarw: Not sure if I got it right, Something like below?

```
  override def value: String =
    if (HttpHeader.validateHeaderValue(super.value()))
      throw new IllegalArgumentException("value must not be contain line break characters")
    else super.value()
```

jrudolph:
Thanks again, @nitikagarw. I'm currently not available too much, so I took the liberty to amend the PR with a few changes and tests. WDYT?

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
```diff
@@ -82,6 +82,7 @@ private[headers] sealed trait SyntheticHeader extends ModeledHeader
  * as they allow the custom header to be matched from [[RawHeader]] and vice-versa.
  */
 abstract class CustomHeader extends jm.headers.CustomHeader {
+  HttpHeader.validateHeaderValue(value)
```
jrudolph, line 85: Hmm, this one cannot work because of initialization issues. I.e. if the custom implementation of `def value(): String` depends on a custom field that will not yet have been initialized when the validation method is called. Maybe we need to move it to `render` in that case?

nitikagarw: @jrudolph, How about moving it to [apply](https://github.com/akka/akka-http/pull/3763/files#diff-f6cd1e0396f76b9123778a986d9c55a3b43a69c27c949320c30456a2f4a61e29R100) method?

PR #3780, opened 2021-03-08
build: update to Scala 2.13.5, 2.12.13 and scalafix 0.9.26

jrudolph:
Also update silencer to compatible version.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -18,7 +18,7 @@ object Dependencies {
   val h2specName = s"h2spec_${DependencyHelpers.osName}_amd64"
   val h2specExe = "h2spec" + DependencyHelpers.exeIfWindows
   val h2specUrl = s"https://github.com/summerwind/h2spec/releases/download/v${h2specVersion}/${h2specName}.zip"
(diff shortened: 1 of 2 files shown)


raboof (approved):
Are the things we saw in https://github.com/akka/akka-http/pull/3703#issuecomment-742594353 fixed?

jrudolph:
> Are the things we saw in [#3703 (comment)](https://github.com/akka/akka-http/pull/3703#issuecomment-742594353) fixed?

I haven't yet completely checked. I think the sbt update might be more difficult than the Scala update. But I'll do another test.

jrudolph:
> I haven't yet completely checked. I think the sbt update might be more difficult than the Scala update. But I'll do another test.

The separate compilation issue in headers.scala is still there with 2.13.5

jrudolph:
> > I haven't yet completely checked. I think the sbt update might be more difficult than the Scala update. But I'll do another test.
> 
> The separate compilation issue in headers.scala is still there with 2.13.5

Just checked with 2.13.3 which has the same problem, so at least it's not a regression.

PR #3781, opened 2021-03-08
Update to sbt 1.4.7

jrudolph:
Also fix an implicit issue which non-deterministically turned up now and then probably because sometimes implicit resolution depends on definition order (which in turn can depend on the order that
sbt or the file system enumerate files, separate compilation etc.)

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
@@ -1008,8 +1008,7 @@ object TE extends ModeledCompanion[TE] {
 }
 final case class TE(acceptableEncodings: immutable.Seq[TransferEncoding]) extends jm.headers.TE with RequestHeader {
   def append(encodings: immutable.Seq[TransferEncoding]) = TE(this.acceptableEncodings ++ encodings)
-  import `Transfer-Encoding`.encodingsRenderer
-  def renderValue[R <: Rendering](r: R): r.type = r ~~ acceptableEncodings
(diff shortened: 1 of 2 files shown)


akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
```diff
@@ -1008,8 +1008,7 @@ object TE extends ModeledCompanion[TE] {
 }
 final case class TE(acceptableEncodings: immutable.Seq[TransferEncoding]) extends jm.headers.TE with RequestHeader {
   def append(encodings: immutable.Seq[TransferEncoding]) = TE(this.acceptableEncodings ++ encodings)
-  import `Transfer-Encoding`.encodingsRenderer
-  def renderValue[R <: Rendering](r: R): r.type = r ~~ acceptableEncodings
+  def renderValue[R <: Rendering](r: R): r.type = r.~~(acceptableEncodings)(`Transfer-Encoding`.encodingsRenderer)
```
raboof, line 1011: Great if that fixes the compilation... I think we use the same pattern elsewhere as well but if we run into that now we know what to do ;).

raboof (approved):

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
jrudolph: Yeah, for these simple kinds of problems it's perfectly fine, even a few characters saved... It was more problematic in this other test file which exercised the contrived FormField implicits where giving the params explicitly would have been quite a pain.

PR #3786, opened 2021-03-15
core: don't send '100 Continue' if 'Content-Length' header is 0, doesn't exist or data got sent early

mkurz:
Just have a look at the tests I addded, also the title says it all:
[If a request (which is _not_ a chunked one)](https://github.com/akka/akka-http/blob/b071bd67547714bd8bed2ccd8170fbbc6c2dbd77/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpRequestParser.scala#L202-L208) does [not have a `Content-Length` header or it is 0](https://github.com/akka/akka-http/blob/b071bd67547714bd8bed2ccd8170fbbc6c2dbd77/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpMessageParser.scala#L362-L363), it means it will not send a body/entitiy.
Therefore we do not have to send `100 Continue`, because obviously there will be nothing to continue with from the client.

Right now akka-http _does_ send `100 Continue` and then **waits/hangs indefinitely** (=until the request timeout) for data which will never be send... 

Original report: https://github.com/playframework/playframework/issues/10733

BTW: As you can see in the original report, netty does already behave correctly.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
@@ -410,8 +410,16 @@ private[http] object HttpServerBluePrint {
               openRequests = openRequests.enqueue(r)
               messageEndPending = r.createEntity.isInstanceOf[StreamedEntityCreator[_, _]]
               val rs = if (r.expect100Continue) {
-                oneHundredContinueResponsePending = true
-                r.copy(createEntity = with100ContinueTrigger(r.createEntity))
+                r.createEntity match {
+                  case StrictEntityCreator(HttpEntity.Strict(_, _)) =>
+                    // This covers two cases:
+                    // - Either: The strict entity got all its data send already, so no need to wait for more data
+                    // - Or: The strict entity contains no data (Content-Length header value was 0 or it did not exist), the client will not send any data
+                    r
+                  case _ =>
+                    oneHundredContinueResponsePending = true
+                    r.copy(createEntity = with100ContinueTrigger(r.createEntity))
+                }
               } else r
               push(requestPrepOut, rs)
             case MessageEnd =>
--- a/akka-http-core/src/test/scala/akka/http/impl/engine/server/HttpServerSpec.scala
+++ b/akka-http-core/src/test/scala/akka/http/impl/engine/server/HttpServerSpec.scala
@@ -648,6 +648,121 @@ class HttpServerSpec extends AkkaSpec(
       netOut.expectComplete()
     })
 
+    "produce a `100 Continue` response when requested by a `Default` entity and some data sent early" in assertAllStagesStopped(new TestSetup {
+      send("""POST / HTTP/1.1
+             |Host: example.com
+             |Expect: 100-continue
+             |Content-Length: 24
+             |
+             |0123456789ABCDEF""") // only 16 bytes get send at first
+      inside(expectRequest()) {
+        case HttpRequest(POST, _, _, Default(ContentType(`application/octet-stream`, None), 24, data), _) =>
+          val dataProbe = TestSubscriber.manualProbe[ByteString]()
+          data.to(Sink.fromSubscriber(dataProbe)).run()
+          val dataSub = dataProbe.expectSubscription()
+          netOut.expectNoBytes(50.millis.dilated)
+          dataSub.request(1) // triggers `100 Continue` response
+          expectResponseWithWipedDate(
+            """HTTP/1.1 100 Continue
+              |Server: akka-http/test
+              |Date: XXXX
+              |
+              |""")
+          dataProbe.expectNext(ByteString("0123456789ABCDEF"))
+          dataSub.request(1)
+          dataProbe.expectNoMessage(50.millis)
+          send("GHIJKLMN") // send missing 8 bytes
+          dataProbe.expectNext(ByteString("GHIJKLMN"))
+          dataSub.request(1)
+          dataProbe.expectComplete()
+          responses.sendNext(HttpResponse(entity = "Yeah"))
+          expectResponseWithWipedDate(
+            """HTTP/1.1 200 OK
+              |Server: akka-http/test
+              |Date: XXXX
+              |Content-Type: text/plain; charset=UTF-8
+              |Content-Length: 4
+              |
+              |Yeah""")
+      }
+
+      netIn.sendComplete()
+      netOut.expectComplete()
+    })
+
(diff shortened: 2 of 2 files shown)


jrudolph:
OK TO TEST

akka-http-core/src/test/scala/akka/http/impl/engine/server/HttpServerSpec.scala
```diff
@@ -648,6 +648,52 @@ class HttpServerSpec extends AkkaSpec(
...
+    "Don't produce a `100 Continue` response if `Strict` entity is empty because `Content-Length` header is 0" in assertAllStagesStopped(new TestSetup {
+      send("""POST / HTTP/1.1
+             |Host: example.com
+             |Expect: 100-continue
+             |Content-Length: 0
+             |
+             |""")
+
+      val HttpRequest(POST, _, _, entity @ Strict(ContentTypes.NoContentType, ByteString.empty), _) = expectRequest()
+      responses.sendNext(HttpResponse(entity = entity))
```
jrudolph, line 660: Sending back the request entity here is allowed but it's also a bit of a stretch of the HTTP protocol. We should at least also have a test sending back an arbitrary response that doesn't reference the request entity.

jrudolph (commented):
Thanks a lot @mkurz for tackling this. This solution already makes a lot of sense. We should also test what happens when

1. the request handling doesn't touch the entity (or echo it back in the response)
2. what happens with other strict / default request entities that are sent before the `100 Continue` response. The clients are not required to wait for a `100 Continue` response but can immediately send some data. (If this turns out to be buggy, I'd be fine with creating an issue and dealing with it later).

mkurz:
@jrudolph Thanks for the review! I addressed your feedback: You are right, now I am just sending "Yeah" without touching the entity, it makes more sense.
Like requested, I also added tests for cases when data gets send before the `100 Continue` response: That was working out of the box for `Default` entities, when more data is expected.
For `Strict` entities (when all the data gets send already) however I had to adjust my small patch, because when sending _all_ the data early already there is no need to send `100 Continue` and wait for more data.
I squashed everything into one commit and changed the commit and pull request title to reflect the changes.

Please have a look again, thanks!

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -410,8 +410,16 @@ private[http] object HttpServerBluePrint {
               openRequests = openRequests.enqueue(r)
               messageEndPending = r.createEntity.isInstanceOf[StreamedEntityCreator[_, _]]
               val rs = if (r.expect100Continue) {
-                oneHundredContinueResponsePending = true
-                r.copy(createEntity = with100ContinueTrigger(r.createEntity))
+                r.createEntity match {
+                  case StrictEntityCreator(HttpEntity.Strict(_, _)) =>
```
mkurz, line 414: Before the git force push,  this line was:
```scala
case StrictEntityCreator(HttpEntity.Strict(_, ByteString.empty)) =>
```
As you can see I checked only for an empty `ByteString`. However as soon there is a strict entity it means we are done already and no data are expected anymore.

mkurz:
@jrudolph If you have a few minutes, it would be nice if you could let me know if more changes are needed. It would even nicer if that patch could make it into the upcoming Play 2.8.8 release...

ignasi35 (approved):
LGTM

akka-http-core/src/test/scala/akka/http/impl/engine/server/HttpServerSpec.scala
```diff
@@ -648,6 +648,121 @@ class HttpServerSpec extends AkkaSpec(
...
+          |Content-Type: text/plain; charset=UTF-8
+          |Content-Length: 4
+          |
+          |Yeah""")
+
+      netIn.sendComplete()
+      netOut.expectComplete()
+    })
+
+    "not produce a `100 Continue` response if `Strict` entity is empty because `Content-Length` header is missing" in assertAllStagesStopped(new TestSetup {
```
jrudolph, line 743: :+1:

akka-http-core/src/test/scala/akka/http/impl/engine/server/HttpServerSpec.scala
```diff
@@ -648,6 +648,121 @@ class HttpServerSpec extends AkkaSpec(
...
+             |0123456789ABCDEF""") // only 16 bytes get send at first
+      inside(expectRequest()) {
+        case HttpRequest(POST, _, _, Default(ContentType(`application/octet-stream`, None), 24, data), _) =>
+          val dataProbe = TestSubscriber.manualProbe[ByteString]()
+          data.to(Sink.fromSubscriber(dataProbe)).run()
+          val dataSub = dataProbe.expectSubscription()
+          netOut.expectNoBytes(50.millis.dilated)
+          dataSub.request(1) // triggers `100 Continue` response
+          expectResponseWithWipedDate(
+            """HTTP/1.1 100 Continue
```
jrudolph, line 666: Ok, a server is not required to send a `100 Continue` response once it has seen some data but let's not change it right now.

jrudolph (approved):
LGTM, thanks @mkurz.

jrudolph:
> @jrudolph If you have a few minutes, it would be nice if you could let me know if more changes are needed. 

Great, thanks for the additional tests.

> It would even nicer if that patch could make it into the upcoming Play 2.8.8 release...

We probably won't be able to release any time soon.

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
jrudolph: :+1: thanks for the source comments.

jrudolph:
> > It would even nicer if that patch could make it into the upcoming Play 2.8.8 release...
> 
> We probably won't be able to release any time soon.

Seems we might after all, discussions are still ongoing :)

mkurz:
> Seems we might after all, discussions are still ongoing :)

Any news?
I plan to release Play 2.8.8 this week, maybe even tomorrow. Actually we are good to go, just some last tests going on. In case you are going to release akka-http 10.1.15, when would that be? :wink: Personally, I don't need this fix here to be in Play 2.8.8, I am totally fine to release 2.8.8 without it - I am not going to wait for a week or whatever just to get this fix in. If its 2 days, OK...

jrudolph:
Seems we don't release for now. We thought about releasing akka-http for other reasons but we won't do it right now.

mkurz:
@jrudolph That's OK for me.

PR #3792, opened 2021-03-31
build: publish to sonatype directly

jrudolph:
That seems to work for snapshots as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,6 +33,10 @@ cache:
     - $HOME/.jabba/jdk
     - $HOME/.cache/coursier
 
+env:
+  # override default (release to maven central) to only close and go to staging repo
+  - CI_SONATYPE_RELEASE="; sonatypePrepare; sonatypeBundleUpload; sonatypeClose"
+
 script:
   - sbt -jvm-opts .jvmopts-travis "$CMD"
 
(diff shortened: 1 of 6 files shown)


andreaTP (approved):

jrudolph:
Publishing still fails because we need to override `sonatypeProfileName`, however since the default is not marked a Global setting in sbt-sonatype, it seems it cannot be simply overridden (so that it will work deterministically).

jrudolph:
I fixed the remaining issues and also adapted the release script. Let's merge this after validation and see that it at least works for snapshots. Then when we do the next release, we probably be extra cautious to see if everything still works as expected.

ignasi35:
> I fixed the remaining issues and also adapted the release script. Let's merge this after validation and see that it at least works for snapshots.

How does it work for snapshots on Sonatype?

ignasi35:
> > I fixed the remaining issues and also adapted the release script. Let's merge this after validation and see that it at least works for snapshots.
> 
> How does it work for snapshots on Sonatype?

I wonder if the `-SNAPSHOT` suffix is managed by `sbt-ci-release`.

jrudolph:
> > > I fixed the remaining issues and also adapted the release script. Let's merge this after validation and see that it at least works for snapshots.
> > 
> > 
> > How does it work for snapshots on Sonatype?
> 
> I wonder if the `-SNAPSHOT` suffix is managed by `sbt-ci-release`.

I guess so, nothing had to be adjusted for that to work.

mkurz:
The docs still point to bintray: https://github.com/akka/akka-http/blob/master/docs/src/main/paradox/contributing.md

PR #3793, opened 2021-04-06
Include port in HTTP/2 binding warning

raboof:
This makes it easier to diagnose what's going on when you find this
message in your logs - for example, akka-management currently binds
using a flow (https://github.com/akka/akka-management/issues/866)
causing this message to be logged. This change would make it more
obvious that this warning is about the Akka Management port and not
the 'main' application port.


ignasi35 (approved):
Nice!

PR #3804, opened 2021-04-25
Allowed extension of nodeSeqMediaTypes for ScalaXmlSupport

umbreak:
Having `nodeSeqMediaTypes` in the trait `ScalaXmlSupport` allows clients to override it if they want to support some more mediaTypes related to xml (for example application/rdf+xml)


jrudolph:
OK TO TEST

jrudolph:
Thanks, @umbreak. I'm not opposed to allowing the override as you suggested here. A slightly more explicit alternative that already works right now is to define your own implicits with

```scala
implicit val xyzMarshaller = ScalaXmlSupport.nodeSeqMarshaller(mediaTypeToSupport)
// similar for unmarshaller
```

umbreak:
Yes you are right. I was following the approach taken by Circe for example.

johanandren:
Closing this due to inactivity and as Johannes suggested the fact that it already is possible.

Issue #3806, opened 2021-04-26
Akka-http 10.2.4: UnexpectedConnectionClosureException after KeepAliveTimeout expires on backend


marcinkierus:
We have akka-http in version 10.2.4 which acts as a dispatcher of requests to different backend systems. It also puts responses together and server it to the frontend.
The backend systems are behind F5 loadbalancer which is also SSL terminator. As web server we are using apache 2.4.

Occasionally we see 503 errors on the frontend side. When debug is enabled, an exception occurs at the time of the 503 error.

```
[DEBUG] [04/22/2021 10:14:41.466] [integration-akka.actor.default-dispatcher-11] [integration/Pool(shared->https://***:443)] [0 (Idle)]Before event [onNewRequest] In state [Idle] for [5139 ms]
[DEBUG] [04/22/2021 10:14:41.466] [integration-akka.actor.default-dispatcher-11] [integration/Pool(shared->https://***:443)] [0 (PushingRequestToConnection)]After event [onNewRequest] State change [Idle] -> [PushingRequestToConnection]
[DEBUG] [04/22/2021 10:14:41.466] [integration-akka.actor.default-dispatcher-11] [integration/Pool(shared->https://***:443)] [0 (PushingRequestToConnection)]Before event [onRequestDispatched] In state [PushingRequestToConnection] for [0 ms]
[DEBUG] [04/22/2021 10:14:41.466] [integration-akka.actor.default-dispatcher-11] [integration/Pool(shared->https://***:443)] [0 (WaitingForResponse)]After event [onRequestDispatched] State change [PushingRequestToConnection] -> [WaitingForResponse]
[DEBUG] [04/22/2021 10:14:41.474] [integration-akka.actor.default-dispatcher-11] [integration/Pool(shared->https://***:443)] [0 (WaitingForResponse)]Connection failed
[DEBUG] [04/22/2021 10:14:41.474] [integration-akka.actor.default-dispatcher-11] [integration/Pool(shared->https://***:443)] [0 (WaitingForResponse)]Before event [onConnectionFailed] In state [WaitingForResponse] for [7 ms]
[DEBUG] [04/22/2021 10:14:41.474] [integration-akka.actor.default-dispatcher-11] [integration/Pool(shared->https://***:443)] [0 (WaitingForResponse)]Ongoing request [GET /***/*** Empty] is failed because of [connection failure]: [The http server closed the connection unexpectedly before delivering responses for 1 outstanding requests]
[...]

Issue #3759, opened 2021-02-12
UnexpectedConnectionClosureException with upgrade to 10.2.2


shayangz:
Team,

We have a micro-services architecture where we make cross service calls through binary POST HTTP calls on back of akka-http (both client and server side). When service A wants to send a message to service B, it sends a message to an actor A_B which puts that message into a stream that processes the message via `Flow[(HttpRequest, ActorRef), (Try[HttpResponse], ActorRef), Http.HostConnectionPool]` and the response gets translated into a message to the original sender.

When we attempted to upgrade from akka-http `10.1.11` to `10.2.2` we started seeing these exceptions randomly sprinkled in logs. 

```
akka.http.impl.engine.client.OutgoingConnectionBlueprint$UnexpectedConnectionClosureException: The http server closed the connection unexpectedly before delivering responses for 1 outstanding requests
```

Nothing else about the infrastructure has changed. We reverted all the services back to `10.1.11` and the exceptions have disappeared.

We upgraded only one of the services to `10.2.2` leaving the rest of infrastructure on `10.1.11` and we still see the issue. In this scenario, the client is running `10.2.2` and calling to a bunch of `10.1.11` servers. When we turn on debugging on the client where we see the exceptions, we catch situations like this

```
2021-02-12 01:10:54,890 [serviceX-akka.actor.default-dispatcher-9] DEBUG akka.http.impl.engine.client.PoolId - Creating pool.
2021-02-12 01:10:54,893 [serviceX-akka.actor.default-dispatcher-39] DEBUG akka.http.impl.engine.client.PoolId - Dispatching request [POST /api/internal Strict(92 bytes)] to pool
2021-02-12 01:10:54,894 [serviceX-akka.actor.default-dispatcher-6] DEBUG akka.http.impl.engine.client.PoolId - [0 (Idle)]Dispatching request [POST /api/internal Strict(92 bytes)]
2021-02-12 01:10:54,894 [serviceX-akka.actor.default-dispatcher-6] DEBUG akka.http.impl.engine.client.PoolId - [0 (Idle)]Before event [onNewRequest] In state [Idle] for [3287 ms]
2021-02-12 01:10:54,894 [serviceX-akka.actor.default-dispatcher-6] DEBUG akka.http.impl.engine.client.PoolId - [0 (PushingRequestToConnection)]After event [onNewRequest] State change [Idle] -> [PushingRequestToConnection]
[...]

Issue #3481, opened 2020-09-25
akka.http.impl.engine.client.OutgoingConnectionBlueprint$UnexpectedConnectionClosureException: The http server closed the connection unexpectedly before delivering responses for 1 outstanding requests


MartinMajor:
Hello Hakkers,

Our stack:
- Scala 2.12
- Akka 2.6.8
- Akka-http 10.2.0

After upgrading from akka-http 10.1.11 we started to randomly get `akka.http.impl.engine.client.OutgoingConnectionBlueprint$UnexpectedConnectionClosureException: The http server closed the connection unexpectedly before delivering responses for 1 outstanding requests` exceptions. It happens on average 1x per 5000 requests. It is unlikely that it has anything to do with the target server, because exception is usually returned 1ms after calling that http request therefore there is no time to even ping target server.

According to logs it usually happens when some request ends and another follows directly after the previous one.

I've read [this issue](https://github.com/akka/akka-http/issues/1459) so I tried to enable DEBUG for connection pool and this is one occurrence:

```
2020-09-25T13:24:15.077Z DEBUG app/Pool(shared->https://domain:443) Dispatching request [POST /url Strict(415 bytes)] to pool
2020-09-25T13:24:15.077Z DEBUG app/Pool(shared->https://domain:443) [2 (WaitingForResponse)]Received response
2020-09-25T13:24:15.090Z DEBUG app/Pool(shared->https://domain:443) [2 (WaitingForResponse)]Before event [onResponseReceived] In state [WaitingForResponse] for [344 ms]
2020-09-25T13:24:15.091Z DEBUG app/Pool(shared->https://domain:443) [2 (WaitingForResponse)]onResponseReceived in WaitingForResponse with false
2020-09-25T13:24:15.091Z DEBUG app/Pool(shared->https://domain:443) [2 (WaitingForResponseDispatch)]After event [onResponseReceived] State change [WaitingForResponse] -> [WaitingForResponseDispatch]
2020-09-25T13:24:15.091Z DEBUG app/Pool(shared->https://domain:443) [2 (WaitingForResponseDispatch)]Before event [onResponseDispatchable] In state [WaitingForResponseDispatch] for [0 ms]
2020-09-25T13:24:15.091Z DEBUG app/Pool(shared->https://domain:443) [2 (WaitingForResponseEntitySubscription)]After event [onResponseDispatchable] State change [WaitingForResponseDispatch] -> [WaitingForResponseEntitySubscription]
[...]

PR #3816, opened 2021-05-21
core: add pool keep-alive-timeout

octonato:
References #3806, #3759 and #3481

Adds an extra timeout following the ideas discussed in https://github.com/akka/akka-http/issues/3806#issuecomment-829137002

### DIFF
--- a/akka-http-core/src/main/resources/reference.conf
+++ b/akka-http-core/src/main/resources/reference.conf
@@ -556,20 +556,19 @@ akka.http {
     # will automatically terminate itself. Set to `infinite` to completely disable idle timeouts.
     idle-timeout = 30 s
 
-    # The time after which an idle connection (without pending requests)
-    # will automatically terminate itself. Set to `infinite` to completely disable idle timeouts.
+    # When a server uses a `Connection: keep-alive` header to allow using the same
+    # connection to send a subsequent request, close the connection when no such
+    # request appears within the timeout.
+    # Set to `infinite` to allow the connection to remain open indefinitely
+    # (or be closed by the more general idle-timeout).
     #
-    # Note that this is different than the idle-timeout which is about the whole pool.
+    # This settings affects each single connection independently.
     #
-    # This settings afects each single connection independently.
-    # In that context, idle means the connection is sitting unused in the pool and is not associated any ongoing request.
-    #
-    # A common scenario where this setting can be useful is when a host closes a connection according to its keep-alive
-    # timeout and at the same time the client tries to sent a request on a now invalid connection.
-    # Such a call will fail with UnexpectedConnectionClosureException.
-    # This can be prevented (mitigated) by configuring this timeout with a value lower than the host keep-alive timeout.
-    # As such, you minimise the risk of using a connection that is likely to be closed by the host.
-    max-connection-idle-timeout = infinite
+    # A common scenario where this setting can be useful is when a server (or reverse-proxy) closes a connection according to its keep-alive
+    # timeout and at the same time the client tries to send a request on the now-invalid connection.
+    # Such a call will fail with `UnexpectedConnectionClosureException`.
+    # This can be avoided by configuring this timeout with a value lower than the server-side keep-alive timeout.
+    keep-alive-timeout = infinite
 
     # The pool implementation will fail a connection early and clear the slot if a response entity was not
     # subscribed during the given time period after the response was dispatched. In busy systems the timeout might be
--- a/akka-http-core/src/main/resources/reference.conf
+++ b/akka-http-core/src/main/resources/reference.conf
@@ -562,7 +562,7 @@ akka.http {
     #
     # A common scenario where this setting is useful is to prevent a race-condition inherent in HTTP: in most cases, a server
     # or reverse-proxy closes a persistent (kept-alive) connection after some time. HTTP does not define a protocol between
-    # client and server to negotiate a graceful tear down of an idle persistent connection. Therefore, it can happen that a server decides to
+    # client and server to negotiate a graceful teardown of an idle persistent connection. Therefore, it can happen that a server decides to
     # close a connection at the same time that a client decides to send a new request. In that case, the request will fail to be processed,
     # but the client cannot determine for which reason the server closed the connection and whether the request was (partly) processed or not.
     # Such a condition can be observed when a request fails with an `UnexpectedConnectionClosureException`.
--- a/akka-http-core/src/main/resources/reference.conf
+++ b/akka-http-core/src/main/resources/reference.conf
@@ -556,18 +556,21 @@ akka.http {
     # will automatically terminate itself. Set to `infinite` to completely disable idle timeouts.
     idle-timeout = 30 s
 
-    # When a server uses a `Connection: keep-alive` header to allow using the same
-    # connection to send a subsequent request, close the connection when no such
-    # request appears within the timeout.
-    # Set to `infinite` to allow the connection to remain open indefinitely
-    # (or be closed by the more general idle-timeout).
+    # HTTP connections are commonly used for multiple requests, that is, they are kept alive between requests. The
+    # `akka.http.host-connection-pool.keep-alive-timeout` setting configures how long a pool keeps a connection alive between
+    # requests before it closes the connection (and eventually reestablishes it).
     #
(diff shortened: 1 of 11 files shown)


octonato:
@jrudolph, I still want to add tests for this. 

Current tests are not passing on my machine, even without the changes in this PR. 🤷 

Still on draft because missing tests.

jrudolph:
> Current tests are not passing on my machine, even without the changes in this PR. shrug

What tests are broken and how?

akka-http-core/src/main/resources/reference.conf
```diff
@@ -556,6 +556,21 @@ akka.http {
     # will automatically terminate itself. Set to `infinite` to completely disable idle timeouts.
     idle-timeout = 30 s
 
+    # The time after which an idle connection (without pending requests)
+    # will automatically terminate itself. Set to `infinite` to completely disable idle timeouts.
+    #
+    # Note that this is different than the idle-timeout which is about the whole pool.
+    #
+    # This settings afects each single connection independently.
```
jrudolph, line 564:
```suggestion
    # This settings affects each single connection independently.
```

akka-http-core/src/main/resources/reference.conf
```diff
@@ -556,6 +556,21 @@ akka.http {
...
+    # In that context, idle means the connection is sitting unused in the pool and is not associated any ongoing request.
```
jrudolph, line 565:
```suggestion
    # In that context, idle means the connection is sitting unused in the pool and is not associated with any ongoing request.
```

akka-http-core/src/main/resources/reference.conf
```diff
@@ -556,6 +556,21 @@ akka.http {
...
+    #
+    # A common scenario where this setting can be useful is when a host closes a connection according to its keep-alive
```
jrudolph, line 567:
```suggestion
    # A common scenario where this setting can be useful is when a server (or reverse-proxy) closes a connection according to its keep-alive
```

akka-http-core/src/main/resources/reference.conf
```diff
@@ -556,6 +556,21 @@ akka.http {
...
+    # timeout and at the same time the client tries to sent a request on a now invalid connection.
+    # Such a call will fail with UnexpectedConnectionClosureException.
+    # This can be prevented (mitigated) by configuring this timeout with a value lower than the host keep-alive timeout.
+    # As such, you minimise the risk of using a connection that is likely to be closed by the host.
```
jrudolph, line 571: IIRC we use American style usually

```suggestion
    # As such, you minimize the risk of using a connection that is likely to be closed by the host.
```

akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/SlotState.scala
```diff
@@ -191,10 +191,11 @@ private[pool] object SlotState {
   private[pool] case object ToBeClosed extends ShouldCloseConnectionState(None)
   private[pool] case class Failed(cause: Throwable) extends ShouldCloseConnectionState(Some(cause))
 
-  private[pool] case object Idle extends ConnectedState with IdleState {
+  private[pool] final case class Idle(override val stateTimeout: Duration) extends ConnectedState with IdleState {
```
jrudolph, line 194: Let's introduce a member called "maxConnectionIdleTimeout` to make it more explicit what this particular timeout is about.

```suggestion
  private[pool] final case class Idle(maxConnectionIdleTimeout: Duration) extends ConnectedState with IdleState {
    override def stateTimeout: Duration = maxConnectionIdleTimeout
```

akka-http-core/src/main/scala/akka/http/impl/engine/client/pool/SlotState.scala
```diff
@@ -211,7 +212,7 @@ private[pool] object SlotState {
   private[pool] case object PreConnecting extends ConnectedState with IdleState {
     override def onConnectionAttemptSucceeded(ctx: SlotContext, outgoingConnection: Http.OutgoingConnection): SlotState = {
       ctx.debug("Slot connection was (pre-)established")
-      Idle
+      Idle(ctx.settings.maxConnectionIdleTimeout)
```
jrudolph, line 215: Maybe create a helper in SlotState for that repeated call.

jrudolph (commented):
Looks good from a first glance but tests will tell...

raboof:
'connection-idle' seems a confusing name since this timeout does not count when the connection is 'idle' (no bytes are currently exchanged) during an in-flight request.

Since this timeout only comes into play for 'Connection: keep-alive' requests, perhaps we should call this one 'keep-alive-timeout'?

jrudolph:
> Since this timeout only comes into play for 'Connection: keep-alive' requests, perhaps we should call this one 'keep-alive-timeout'?

:+1:

docs/src/main/paradox/common/timeouts.md
```diff
@@ -23,10 +23,6 @@ akka.http.host-connection-pool.idle-timeout
 akka.http.host-connection-pool.client.idle-timeout
 ```
 
-@@@ note
-For the client side connection pool, the idle period is counted only when the pool has no pending requests waiting.
```
raboof, line 27: I guess this was true for the old client pool, but hasn't been the case for a while?

johanandren (approved):
Looks reasonable

docs/src/main/paradox/common/timeouts.md
jrudolph: Actually, it should still apply but only for the `host-connection-pool.idle-timeout` which is a different timeout than the other ones listed here. The other ones are all connection-level timeouts, while the pool timeout is for the pool. Maybe the paragraph could be broken down into two?

akka-http-core/src/test/scala/akka/http/impl/engine/client/HostConnectionPoolSpec.scala
```diff
@@ -345,6 +345,31 @@ class HostConnectionPoolSpec extends AkkaSpecWithMaterializer(
...
+          pushRequest(HttpRequest(uri = "/simple"))
+
+          val conn1 = expectNextConnection()
+          val req = conn1.expectRequest()
+          conn1.pushResponse(HttpResponse(headers = headers.Connection("keep-alive") :: Nil, entity = req.uri.path.toString))
+
+          expectResponseEntityAsString() shouldEqual "/simple"
+          val receivedFirstResponse = System.nanoTime()
+
+          conn1.serverRequests.expectComplete()
```
jrudolph, line 360:
```suggestion
          // wait until pool closes connection because of keep-alive-timeout
          conn1.serverRequests.expectComplete()
```

akka-http-core/src/test/scala/akka/http/impl/engine/client/HostConnectionPoolSpec.scala
```diff
@@ -345,6 +345,31 @@ class HostConnectionPoolSpec extends AkkaSpecWithMaterializer(
...
+          conn1.serverResponses.sendComplete()
+
+          val complete = System.nanoTime()
+          (System.nanoTime() - receivedFirstResponse).nanos should be >= 800.millis
+          println("XXX", System.nanoTime() - receivedFirstResponse)
```
jrudolph, line 365: Should probably also check some upper limit.

jrudolph:
Added some documentation changes. Will now finish the testing bits.

docs/src/main/paradox/common/timeouts.md
```diff
@@ -7,26 +7,22 @@ are left to the streaming APIs and are easily implementable as patterns in user-
...
-independently for each of those using the following keys:
+This setting should be used as a last-resort safeguard to prevent unused or stuck connections from consuming resources for
+an indefinite time.
+
+The setting works the same way for server and client connections and it is configurable independently using the following keys:
 
 ```
 akka.http.server.idle-timeout
 akka.http.client.idle-timeout
-akka.http.host-connection-pool.idle-timeout
```
jrudolph, line 22: Moved that one down to the pool settings section.

jrudolph (approved):
Implementation LGTM. I worked on documentation and tests.

johanandren (approved):
LGTM!

Issue #3815, opened 2021-05-17
Investigate HTTP/2 performance improvements


jrudolph:
HTTP/2 processing currently does not benefit from strict entities to the same degree as HTTP/1.1. For that reason even simple requests are quite expensive.

Incoming and outgoing sides need to be distinguished in that regard:

 * On the outgoing side, strict entities are supported but if you need trailers (as in gRPC), you have to use `Chunked` entities. To avoid that, we could provide an attribute on a strict response to carry immediate trailer headers (to be interpreted by the HTTP/2 infrastructure)
 * On the incoming side, we never get to parse headers together with entity data because those are delivered in different frames. In contrast, with HTTP/1.1 you get the next slice of the stream buffer and if we happen to see the full entity we create a `Strict` entity. To avoid creating a streaming entity after seeing the `HEADERS` frame, we could introduce a setting to collect data up to a certain size before giving out a streamed entity. In many simple cases that would mean to just wait for the next frame containing the entity payload.

(Observed in https://discuss.lightbend.com/t/akka-grpc-performance-in-benchmarks/8236/7)

PR #3820, opened 2021-05-28
Allow trailing HTTP/2 headers on strict responses

raboof:
Through an attribute.

First step of the first bullet in #3815.

Still to do:
* [x] benchmark
* [x] Java API
* [x] documentation

It looks like there's more opportunity for optimization of strict HTTP/2 responses, but perhaps that should be a separate PR.

### DIFF
--- a/akka-http-bench-jmh/src/main/scala/akka/http/impl/engine/http2/H2ClientServerBenchmark.scala
+++ b/akka-http-bench-jmh/src/main/scala/akka/http/impl/engine/http2/H2ClientServerBenchmark.scala
@@ -46,7 +46,7 @@ class H2ClientServerBenchmark extends CommonBenchmark {
   val numRequests = 1000
 
   @Benchmark
-  @OperationsPerInvocation(10000) // should be same as numRequest
+  @OperationsPerInvocation(1000) // should be same as numRequest
   def benchRequestProcessing(): Unit = {
     implicit val ec: ExecutionContext = system.dispatcher
 
--- a/akka-http-bench-jmh/src/main/scala/akka/http/impl/engine/http2/H2ClientServerBenchmark.scala
+++ b/akka-http-bench-jmh/src/main/scala/akka/http/impl/engine/http2/H2ClientServerBenchmark.scala
@@ -26,6 +26,10 @@ import scala.collection.immutable
 import scala.concurrent.duration._
 import scala.concurrent.{ Await, ExecutionContext, Future }
 
+/**
+ * Test converting a HttpRequest to bytes at the client and back to a request at the server, and vice-versa
+ * for the response. Does not include the network.
+ */
 class H2ClientServerBenchmark extends CommonBenchmark {
   // Obtained by converting the input request bytes from curl with --http2-prior-knowledge
   def request(streamId: Int) =
--- a/akka-http-bench-jmh/src/main/scala/akka/http/impl/engine/http2/H2ClientServerBenchmark.scala
+++ b/akka-http-bench-jmh/src/main/scala/akka/http/impl/engine/http2/H2ClientServerBenchmark.scala
@@ -4,56 +4,59 @@
 
 package akka.http.impl.engine.http2
 
+import akka.NotUsed
 import akka.actor.ActorSystem
 import akka.http.CommonBenchmark
 import akka.http.impl.engine.http2.FrameEvent.HeadersFrame
 import akka.http.impl.engine.http2.framing.FrameRenderer
-import akka.http.scaladsl.model.{ HttpResponse }
-import akka.http.scaladsl.settings.ServerSettings
+import akka.http.scaladsl.client.RequestBuilding.Get
+import akka.http.scaladsl.model.HttpEntity.LastChunk
+import akka.http.scaladsl.model.headers.RawHeader
+import akka.http.scaladsl.model.{ ContentTypes, HttpEntity, HttpRequest, HttpResponse }
+import akka.http.scaladsl.settings.{ ClientConnectionSettings, ServerSettings }
 import akka.stream.ActorMaterializer
 import akka.stream.TLSProtocol.{ SslTlsInbound, SslTlsOutbound }
-import akka.stream.scaladsl.{ Flow, Keep, Sink, Source }
+import akka.stream.scaladsl.{ BidiFlow, Flow, Keep, Sink, Source }
 import akka.util.ByteString
 import com.typesafe.config.ConfigFactory
 import org.openjdk.jmh.annotations._
 
 import java.util.concurrent.{ CountDownLatch, TimeUnit }
-import scala.concurrent.{ Await, Future }
+import scala.collection.immutable
 import scala.concurrent.duration._
+import scala.concurrent.{ Await, ExecutionContext, Future }
 
-class H2ServerSwitchBenchmark extends CommonBenchmark {
+class H2ClientServerBenchmark extends CommonBenchmark {
   // Obtained by converting the input request bytes from curl with --http2-prior-knowledge
   def request(streamId: Int) =
     FrameRenderer.render(HeadersFrame(streamId, endStream = true, endHeaders = true, HPackSpecExamples.C41FirstRequestWithHuffman, None))
 
-  var response: HttpResponse = HPackSpecExamples.FirstResponse
+  @Param(Array("closedelimited", "chunked"))
+  var responsetype: String = _
 
-  var httpFlow: Flow[ByteString, ByteString, Any] = _
+  var response: HttpResponse = _
+  var httpFlow: Flow[HttpRequest, HttpResponse, Any] = _
   implicit var system: ActorSystem = _
   implicit var mat: ActorMaterializer = _
 
-  val packedResponse = ByteString(-62, -63, -64, -65, -66)
-
-  val numRequests = 10000
+  val numRequests = 1000
 
   @Benchmark
   @OperationsPerInvocation(10000) // should be same as numRequest
   def benchRequestProcessing(): Unit = {
+    implicit val ec: ExecutionContext = system.dispatcher
+
     val latch = new CountDownLatch(numRequests)
 
     val requests =
-      Source(Http2Protocol.ClientConnectionPreface +: Range(0, numRequests).map(i => request(1 + 2 * i)))
+      Source.repeat(Get("/")).take(numRequests)
         .concatMat(Source.maybe)(Keep.right)
 
     val (in, done) =
       requests
         .viaMat(httpFlow)(Keep.left)
         .toMat(Sink.foreach(res => {
-          // Skip headers/settings frames etc
-          if (res.containsSlice(HPackSpecExamples.C61FirstResponseWithHuffman)
-            || res.containsSlice(packedResponse)) {
-            latch.countDown()
-          }
+          res.discardEntityBytes().future.onComplete(_ => latch.countDown())
         }))(Keep.both)
         .run()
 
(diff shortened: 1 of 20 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/http2/HttpMessageRendering.scala
```diff
@@ -75,9 +75,15 @@ private[http2] sealed abstract class MessageRendering[R <: HttpMessage] extends
     HttpMessageRendering.addContentHeaders(headerPairs, r.entity)
     HttpMessageRendering.renderHeaders(r.headers, headerPairs, peerIdHeader, log, isServer = r.isResponse)
 
-    val headersFrame = ParsedHeadersFrame(nextStreamId(r), endStream = r.entity.isKnownEmpty, headerPairs.result(), None)
+    val streamId = nextStreamId(r)
+    val headersFrame = ParsedHeadersFrame(streamId, endStream = r.entity.isKnownEmpty, headerPairs.result(), None)
+    val trailingHeadersFrame: Option[ParsedHeadersFrame] =
+      r.attribute(AttributeKeys.trailer) match {
+        case Some(trailer) => Some(ParsedHeadersFrame(streamId, true, trailer.headers.map(header => (header.name, header.value)), None))
```
raboof, line 82: Since we're converting to pairs here anyway, perhaps we should store the headers as pairs already in `Trailer`, and allow creating the `Trailer` without going through `HttpHeader`. Since we're not currently providing access to the trailing headers for incoming messages anyway, perhaps it's OK to not have public API's to read the headers from the trailer for now anyway.

akka-http-bench-jmh/src/main/scala/akka/http/impl/engine/http2/H2ServerSwitchBenchmark.scala
```diff
@@ -0,0 +1,97 @@
...
+import akka.stream.scaladsl.{ Flow, Keep, Sink, Source }
+import akka.util.ByteString
+import com.typesafe.config.ConfigFactory
+import org.openjdk.jmh.annotations._
+
+import java.util.concurrent.{ CountDownLatch, TimeUnit }
+import scala.concurrent.{ Await, Future }
+import scala.concurrent.duration._
+
+class H2ServerSwitchBenchmark extends CommonBenchmark {
```
jrudolph, line 24: Can you add a small description what this is about (compared to the other benchmark)?

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
```diff
@@ -30,8 +30,11 @@ import scala.collection.immutable
  */
 @InternalApi
 private[http2] case class Http2SubStream(
-  initialHeaders:        ParsedHeadersFrame,
-  data:                  Source[Any, Any],
+  initialHeaders: ParsedHeadersFrame,
+  // outgoing response trailing headers can either be passed in eagerly via an attribute
+  // or streaming as the LastChunk of a chunked data stream
+  trailingHeaders:       Option[ParsedHeadersFrame],
```
jrudolph, line 36: `OptionVal`?

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -686,6 +686,7 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
         case newData: ByteString          => buffer ++= newData
         case HttpEntity.Chunk(newData, _) => buffer ++= newData
         case HttpEntity.LastChunk(_, headers) =>
+          // TODO we should probably consider it 'invalid' when we see both 'eager' trailing headers and a LastChunk. Should we fail/log in that case?
```
jrudolph, line 689: Yes, let's log a warning in that case, it's a developer error.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/HttpMessageRendering.scala
jrudolph: Yes, maybe let's strip down the public API completely for now so that you can only create a `Trailer` from `Seq[HttpHeader]` or pairs but not access the values. So, we can choose whatever representation we want in the implementation.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -237,6 +238,21 @@ class Http2ServerSpec extends AkkaSpecWithMaterializer("""
...
+        user.expectRequest()
+        val response =
+          HttpResponse(StatusCodes.OK, entity = HttpEntity.Strict(ContentTypes.`application/octet-stream`, ByteString("Hello")))
+            .addAttribute(AttributeKeys.trailer, Trailer(immutable.Seq[HttpHeader](RawHeader("Status", "grpc-status 10"))))
+        user.emitResponse(streamId, response)
+
+        network.expectHeaderBlock(streamId, endStream = false)
+        network.expectDATA(streamId, endStream = false, ByteString("Hello"))
+        val trailingResponseHeaders = network.expectDecodedResponseHEADERSPairs(streamId)
+        println(trailingResponseHeaders)
```
jrudolph, line 253: Needs an assertion.

docs/src/main/paradox/server-side/http2.md
```diff
@@ -75,6 +75,23 @@ For the reason this approach is known as h2c with
 [Prior Knowledge](https://httpwg.org/specs/rfc7540.html#known-http) of HTTP/2
 support.
 
+## Trailing headers
```
jrudolph, line 78: :+1:

akka-http-bench-jmh/src/main/scala/akka/http/impl/engine/http2/H2ServerProcessingBenchmark.scala
```diff
@@ -5,55 +5,53 @@
...
   @Benchmark
   @OperationsPerInvocation(10000) // should be same as numRequest
   def benchRequestProcessing(): Unit = {
+    implicit val ec = system.dispatcher
+
     val latch = new CountDownLatch(numRequests)
 
     val requests =
-      Source(Http2Protocol.ClientConnectionPreface +: Range(0, numRequests).map(i => request(1 + 2 * i)))
+      Source.repeat(Get("/")).take(numRequests)
```
jrudolph, line 45: Does this benchmark really work? It seems to spend all it's time in the materialization inside the bench method.

jrudolph (commented):
A few comments

akka-http-bench-jmh/src/main/scala/akka/http/impl/engine/http2/H2ServerProcessingBenchmark.scala
```diff
@@ -77,18 +81,10 @@ class H2ServerProcessingBenchmark extends CommonBenchmark {
...
-    val http1 = Flow[SslTlsInbound].mapAsync(1)(_ => {
-      Future.failed[SslTlsOutbound](new IllegalStateException("Failed h2 detection"))
-    })
-    val http2 =
-      Http2Blueprint.handleWithStreamIdHeader(1)(req => {
-        req.discardEntityBytes().future.map(_ => response)
-      })(system.dispatcher)
-        .join(Http2Blueprint.serverStackTls(settings, log, NoOpTelemetry))
-    httpFlow = Http2.priorKnowledge(http1, http2)
+    httpFlow = Http2Blueprint.handleWithStreamIdHeader(1)(req => {
```
jrudolph, line 85: `handleWithStreamIdHeader` doesn't do anything, so you basically removed the whole HTTP/2 pipeline here. Was that intended?

raboof: oops, no :S

akka-http-core/src/main/scala/akka/http/scaladsl/model/Trailer.scala
```diff
@@ -0,0 +1,36 @@
...
+
+import scala.collection.immutable
+
+class Trailer(@ApiMayChange val headers: immutable.Seq[(String, String)]) extends jm.Trailer {
+  /**
+   * Java API
+   */
+  override def addHeader(header: jm.HttpHeader): Trailer = {
+    val sheader = header.asInstanceOf[HttpHeader]
+    new Trailer((sheader.name, sheader.value) +: headers)
```
jrudolph, line 18: Why to the front?

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -237,6 +238,22 @@ class Http2ServerSpec extends AkkaSpecWithMaterializer("""
...
+        user.expectRequest()
+        val response =
+          HttpResponse(StatusCodes.OK, entity = HttpEntity.Strict(ContentTypes.`application/octet-stream`, ByteString("Hello")))
+            .addAttribute(AttributeKeys.trailer, Trailer(RawHeader("Status", "grpc-status 10")))
+        user.emitResponse(streamId, response)
+
+        network.expectHeaderBlock(streamId, endStream = false)
+        network.expectDATA(streamId, endStream = false, ByteString("Hello"))
+        val trailingResponseHeaders = network.expectDecodedResponseHEADERSPairs(streamId)
+        trailingResponseHeaders.size should be(1)
```
jrudolph, line 253: :+1:

akka-http-bench-jmh/src/main/scala/akka/http/impl/engine/http2/H2ServerProcessingBenchmark.scala
```diff
@@ -22,6 +22,10 @@ import akka.util.ByteString
 import com.typesafe.config.ConfigFactory
 import org.openjdk.jmh.annotations._
 
+/**
+ * Emulates request bytes coming in, selecting HTTP/2 over HTTP/1, sending a static response back,
+ * validating the response bytes come out again.
+ */
```
jrudolph, line 28: In my branch, I added more params / features to this benchmark to avoid measuring the client as well. I'm fine with also adding the client/server version.

akka-http-bench-jmh/src/main/scala/akka/http/impl/engine/http2/H2ClientServerBenchmark.scala
```diff
@@ -0,0 +1,112 @@
...
+
+  var response: HttpResponse = _
+  var httpFlow: Flow[HttpRequest, HttpResponse, Any] = _
+  implicit var system: ActorSystem = _
+  implicit var mat: ActorMaterializer = _
+
+  val numRequests = 1000
+
+  @Benchmark
+  @OperationsPerInvocation(10000) // should be same as numRequest
```
jrudolph, line 49:
```suggestion
  @OperationsPerInvocation(1000) // should be same as numRequest
```

jrudolph (approved):
LGTM, thanks a lot @raboof. One notable thing missing here, is that avoiding streams currently only works for empty responses, not for strict responses in general. I'm already working at allowing strict data in `Http2Substream`, so we can avoid creating streams for any kind of strict responses.

akka-http-bench-jmh/src/main/scala/akka/http/impl/engine/http2/H2ServerProcessingBenchmark.scala
raboof: Ah, cool, good to have both probably (but for this particular case indeed better to avoid measuring the client as well)

akka-http-core/src/main/scala/akka/http/scaladsl/model/Trailer.scala
raboof: No real reason - habit because it is faster for lists, but that likely won't be relevant here. I'll change it to append.

raboof:
> One notable thing missing here, is that avoiding streams currently only works for empty responses, not for strict responses in general.

Correct, I focused on allowing the user to specify the reply-with-trailing-headers without constructing a stream there, but didn't change any of the streaming further inside the implementation.

> I'm already working at allowing strict data in `Http2Substream`, so we can avoid creating streams for any kind of strict responses.

:+1: great!

PR #3822, opened 2021-05-31
Stack of potential HTTP/2 improvements

jrudolph:
So far this is somewhat of a grab bag. Will create separate PRs later.

Planned further optimizations:
 * incoming data to Strict conversion
 * import and adapt HPACK library to avoid header creation / allocation for already seen headers

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
@@ -18,8 +18,10 @@ import akka.http.impl.util.LogByteStringTools.logTLSBidiBySetting
 import akka.http.impl.util.StreamUtils
 import akka.http.scaladsl.model._
 import akka.http.scaladsl.settings.{ ClientConnectionSettings, Http2ClientSettings, Http2ServerSettings, ParserSettings, ServerSettings }
+import akka.stream.{ Attributes, FlowShape, Inlet, Outlet }
 import akka.stream.TLSProtocol._
 import akka.stream.scaladsl.{ BidiFlow, Flow, Source }
+import akka.stream.stage.{ GraphStage, GraphStageLogic, InHandler, OutHandler }
 import akka.util.{ ByteString, OptionVal }
 
 import scala.concurrent.duration.{ Duration, FiniteDuration }
@@ -106,6 +108,7 @@ private[http] object Http2Blueprint {
       FrameLogger.logFramesIfEnabled(settings.http2Settings.logFrames) atop // enable for debugging
       hpackCoding(masterHttpHeaderParser) atop
       framing(log) atop
+      coalesceWrites atop
       idleTimeoutIfConfigured(settings.idleTimeout)
   }
 
@@ -123,6 +126,7 @@ private[http] object Http2Blueprint {
       FrameLogger.logFramesIfEnabled(settings.http2Settings.logFrames) atop // enable for debugging
       hpackCoding(masterHttpHeaderParser) atop
       framingClient(log) atop
+      coalesceWrites atop
       idleTimeoutIfConfigured(settings.idleTimeout)
   }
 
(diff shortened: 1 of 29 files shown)


akka-http-bench-jmh/src/main/scala/akka/http/impl/engine/http2/H2ClientServerBenchmark.scala
```diff
@@ -1,112 +0,0 @@
...
-    val server: Flow[ByteString, ByteString, NotUsed] = Http2.priorKnowledge(http1, http2)
-    val client: BidiFlow[HttpRequest, ByteString, ByteString, HttpResponse, NotUsed] = Http2Blueprint.clientStack(ClientConnectionSettings(system), log, NoOpTelemetry)
-    httpFlow = client.join(server)
-  }
-
-  @TearDown
-  def tearDown(): Unit = {
-    system.terminate()
-  }
-}
```
raboof, line 112: did you mean to delete this one, or did it move?

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -596,10 +602,15 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
...
-      info.registerIncomingData(subIn)
-      sub.data.runWith(subIn.sink)(subFusingMaterializer)
+      sub.data match {
+        case Right(data) =>
+          val subIn = new SubSinkInlet[Any](s"substream-in-${sub.streamId}")
+          info.registerIncomingData(subIn)
+          data.runWith(subIn.sink)(subFusingMaterializer)
+        case Left(data) =>
+          info.addAllData(data)
+      }
```
raboof, line 613: :tada:

akka-http-core/src/main/scala/akka/http/impl/engine/http2/hpack/HeaderCompression.scala
```diff
@@ -44,7 +44,7 @@ private[http2] object HeaderCompression extends GraphStage[FlowShape[FrameEvent,
         case ParsedHeadersFrame(streamId, endStream, kvs, prioInfo) =>
           kvs.foreach {
             case (key, value) =>
-              encoder.encodeHeader(os, key.getBytes(HeaderDecompression.UTF8), value.getBytes(HeaderDecompression.UTF8), false)
+              encoder.encodeHeader(os, key.getBytes(HeaderDecompression.US_ASCII), value.getBytes(HeaderDecompression.US_ASCII), false)
```
patriknw, line 47: for ascii it's possible to take it one step further by using this in Akka:
https://github.com/akka/akka/blob/206dafa01d81224e5f93ddc6e6ec7fa702a03ce3/akka-actor/src/main/scala/akka/util/Unsafe.java#L129

akka-http-bench-jmh/src/main/scala/akka/http/impl/engine/http2/H2ClientServerBenchmark.scala
jrudolph: Thanks probably broken merge.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/hpack/HeaderCompression.scala
jrudolph: Ah, interesting, I think we have a similar helper already also in akka-http as well.

jrudolph:
Added the reverse optimization as well by adding a new option `min-collect-strict-entity-size`, which will postpone giving out Http messages until at least this many bytes have been seen or the stream has been closed. A downside to that approach is that a request or response can now be delayed indefinitely if no entity bytes are received. We might have to add an extra option to create a streaming message anyway if data is not received after a timeout eventually.

jrudolph:
I had a bit of time yesterday and pushed for a few more optimizations that turned up to be pretty significant:

 * Parsing headers is a significant part of parsing HTTP and akka-http does more work than other frameworks to validate and parse headers into a structured format. HTTP/2 uses HPACK for header compression which means that repeated headers between HTTP messages are represented by a simple index. Embracing that logic we can store already parsed headers inside the HPACK decompressor, so in typical scenarios (and especially in benchmarks), we can avoid parsing headers for most headers after the first request.
   * To get there, I in-sourced HPACK and made minimal changes to store parsed data inside of the compression tables.
 * A big chunk of remaining cost is caused by excessive system calls for doing network IO because we usually render out frames separately. I added an experimental stage that collects outgoing data for some time (1 ms in code, but in fact, 10 ms due to default scheduler granularity). Voluntarily adding 10 ms of latency won't do as an actual optimization (while 1 ms might be acceptable), but it demonstrates that this increases throughput a lot.

Taking all these optimizations together, throughput in grpc-bench running on a single core, increased from 7k RPS to 30k RPS on my machine which sounds really promising.

(Tests are failing because I didn't integrate the HPACK optimization with the client side yet and also because the write coalescing messes with some test expectation.)

jrudolph:
1 core:

```
-----------------------------------------------------------------------------------------------------------------------------------------
| name                        |   req/s |   avg. latency |        90 % in |        95 % in |        99 % in | avg. cpu |   avg. memory |
-----------------------------------------------------------------------------------------------------------------------------------------
| scala_akka                  |   26803 |       37.11 ms |       70.09 ms |       86.45 ms |      104.58 ms |   101.2% |    252.53 MiB |
| java_hotspot_grpc_pgc       |   20875 |       47.69 ms |       71.17 ms |       77.71 ms |       97.99 ms |  100.26% |    194.43 MiB |
-----------------------------------------------------------------------------------------------------------------------------------------
```

3 core:

```
-----------------------------------------------------------------------------------------------------------------------------------------
| name                        |   req/s |   avg. latency |        90 % in |        95 % in |        99 % in | avg. cpu |   avg. memory |
-----------------------------------------------------------------------------------------------------------------------------------------
| scala_akka                  |   58008 |       16.85 ms |       30.89 ms |       37.33 ms |       50.52 ms |  219.01% |     236.4 MiB |
| java_hotspot_grpc_pgc       |   55564 |       17.45 ms |       35.80 ms |       47.67 ms |       72.78 ms |  302.58% |     202.1 MiB |
-----------------------------------------------------------------------------------------------------------------------------------------
```

(Though, as said before, I don't quite trust those benchmark numbers)

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
```diff
@@ -179,6 +191,49 @@ private[http] object Http2Blueprint {
   def clientDemux(settings: Http2CommonSettings, masterHttpHeaderParser: HttpHeaderParser): BidiFlow[Http2SubStream, FrameEvent, FrameEvent, Http2SubStream, NotUsed] =
     BidiFlow.fromGraph(new Http2ClientDemux(settings, masterHttpHeaderParser))
 
+  val coalesceWrites: BidiFlow[ByteString, ByteString, ByteString, ByteString, NotUsed] =
```
patriknw, line 194: This idea is great, I just want to understand some more about if it's supposed to make a difference for our current throughput benchmarks. In those each "virtual user" emits unary calls sequentially and waits for response for each request before emitting next. We use rather many virtual users (up to 1500). Each virtual user has its own gRPC connection. The request and response messages are small.

jrudolph: Right now we send frames one-by-one. A gRPC usually needs at least 3 frames (initial headers, data, trailer headers), so we can reduce the number of system calls by a factor of three if we can emit those in one chunk.

There's a commented out variant that just schedules an AsyncInput which will collect all frames that are the result of the current run (i.e. a response should be sent). That version wouldn't introduce a fixed delay.

The wait time will help if there are multiple responses that should be sent out in about the same time frame but over multiple activations of the GraphInterpreter (responses come in from the user asynchronously from a `mapAsyncUnordered`.

I'll also try the alternative of batching together all pending frames from the demuxer directly without the extra stage.

patriknw: I see, headers, data, trailer explains why we probably saw an improvement also for the "single-threaded unary call clients"

jrudolph: > This idea is great, I just want to understand some more about if it's supposed to make a difference for our current throughput benchmarks. In those each "virtual user" emits unary calls sequentially and waits for response for each request before emitting next. We use rather many virtual users (up to 1500). Each virtual user has its own gRPC connection. The request and response messages are small.

I tested again emulating that particular behavior (i.e. concurrent requests == num connections), and indeed in that case, it doesn't seem as if coalescing writes will make a significant difference.

That said, we leave a lot of performance on the table if we don't use a load balancer that bundles up incoming user requests into a smaller number of load-balancer->backend connections.

jrudolph:
Most of the interesting parts from here have been merged in the meantime.

PR #3823, opened 2021-06-01
http2: avoid extra allocations in frame renderer

jrudolph:
A quick rewrite of the FrameRenderer to use preallocated arrays of exactly the right size. Split out of #3822.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/framing/FrameRenderer.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/framing/FrameRenderer.scala
@@ -5,16 +5,13 @@
 package akka.http.impl.engine.http2
 package framing
 
-import java.nio.ByteOrder
-
-import akka.util.ByteString.ByteString1C
-import akka.util.{ ByteString, ByteStringBuilder }
-import Http2Protocol.FrameType
 import akka.annotation.InternalApi
+import akka.http.impl.engine.http2.FrameEvent._
+import akka.http.impl.engine.http2.Http2Protocol.FrameType
+import akka.util.ByteString
 
+import java.nio.ByteOrder
 import scala.annotation.tailrec
-
-import FrameEvent._
 
 /** INTERNAL API */
 @InternalApi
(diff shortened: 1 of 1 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/http2/framing/FrameRenderer.scala
```diff
@@ -24,141 +21,200 @@ private[http2] object FrameRenderer {
...
-
-        renderFrame(
-          Http2Protocol.FrameType.HEADERS,
-          Http2Protocol.Flags.END_STREAM.ifSet(endStream) |
-            Http2Protocol.Flags.END_HEADERS.ifSet(endHeaders) |
-            Http2Protocol.Flags.PRIORITY.ifSet(prioInfo.isDefined),
-          streamId,
-          renderedPrioInfo ++ headerBlockFragment
-        )
+        HeaderBuilder(headerBlockFragment.size + (if (prioInfo.isDefined) 5 else 0))
```
raboof, line 47: Perhaps follow the order of the fields

```suggestion
        HeaderBuilder((if (prioInfo.isDefined) 5 else 0) + headerBlockFragment.size)
```

akka-http-core/src/main/scala/akka/http/impl/engine/http2/framing/FrameRenderer.scala
```diff
@@ -24,141 +21,200 @@ private[http2] object FrameRenderer {
...
-    headerBytes(7) = (streamId >> 8).toByte
-    headerBytes(8) = (streamId >> 0).toByte
-
-    ByteString1C(headerBytes) ++ payload
+  private class HeaderBuilder(payloadSize: Int) {
+    private val targetSize = 9 + payloadSize
+    private val buffer = new Array[Byte](targetSize)
+    private var pos = 0
+
+    def putHeader(tpe: FrameType, flags: ByteFlag, streamId: Int): HeaderBuilder = {
```
raboof, line 156: Since there is always a header, maybe pass these in in the constructor rather than via a `putHeader`?

raboof (approved):

akka-http-core/src/main/scala/akka/http/impl/engine/http2/framing/FrameRenderer.scala
jrudolph: :+1: done

akka-http-core/src/main/scala/akka/http/impl/engine/http2/framing/FrameRenderer.scala
jrudolph: :+1: done

PR #3829, opened 2021-06-07
http2: add setting to automatically collect (request) entity data into strict entities

jrudolph:
As prototyped in #3822.

It is currently only support for request entities. I'll create an issue to also support that functionality for the client side (which should be somewhat easy to add but needs tests).

### DIFF
--- a/akka-http-core/src/main/resources/reference.conf
+++ b/akka-http-core/src/main/resources/reference.conf
@@ -260,6 +260,19 @@ akka.http {
       # be increased for high bandwidth-delay-product connections.
       incoming-stream-level-buffer-size = 512kB
 
+      # For incoming requests, the infrastructure collects at least the given number of bytes before dispatching a HttpRequest.
+      # If all request data is received before or when the threshold is reached, the entity data is dispatched as a strict entity
+      # which allows more efficient processing of the request data without involving streams.
+      #
+      # If the given value is 0, the request is immediately dispatched and always carries a stream of data.
+      # You can use value `1` to create a strict entity if the request contains at most a single data frame, and a streamed
+      # entity otherwise.
+      #
+      # Note, that if enabled, requests that expect data (endStream = false on HEADERS) but never received a DATA frame will never be
+      # dispatched. In a regular HTTP context, this is uncommon. It would require that a client would expect data from the server first
+      # before sending any own data.
+      min-collect-strict-entity-size = 0
+
       # The maximum number of outgoing control frames to buffer when the peer does not read from its TCP connection before
       # backpressuring incoming frames.
       #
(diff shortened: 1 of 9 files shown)


akka-http-core/src/main/resources/reference.conf
```diff
@@ -260,6 +260,21 @@ akka.http {
...
+      # If all request data is received before or when the threshold is reached, the entity data is dispatched as a strict entity
+      # which allows more efficient processing of the request data without involving streams.
+      #
+      # If the given value is 0, the request is immediately dispatched and always carries a stream of data.
+      # You can use value `1` to create a strict entity if the request contains at most a single data frame, and a streamed
+      # entity otherwise.
+      #
+      # If a strict entity is dispatched, the data size will be at least the amount given here.
+      #
+      # Note, that if enabled, requests that expect data (endStream = false on HEADERS) but never received a DATA frame will never be
```
jrudolph, line 273: @raboof WDYT about this? I think this is no constraint at all for gRPC because every request needs to have a payload AFAIU.

But even for usual HTTP traffic this seems hardly a concern. A pattern that doesn't work when enabled would be a kind of long-polling where the client is still allowed to send some control messages but doesn't do so at the start of the request.

A DoS concern could be that an attacker never sends over an expected request entities. In that case, the behavior with the feature enabled would be better than before since we would allocate even less resources than with the feature disabled.

In any case, it would make sense to guard against these kinds of attacks with more timeouts, e.g. `max-time-to-first-entity-byte`.

raboof: > I think this is no constraint at all for gRPC because every request needs to have a payload AFAIU.

I think so too. That payload might be a `google.protobuf.Empty` which serializes to an empty bytearray, but it'd still be wrapped in 2 bytes of gRPC framing, so there would be a DATA frame after all.

patriknw: we can't make this enabled by default?

patriknw (approved):
LGTM, with a question...

akka-http-core/src/main/resources/reference.conf
```diff
@@ -260,6 +260,21 @@ akka.http {
...
 
+      # For incoming requests, the infrastructure collects at least the given number of bytes before dispatching a HttpRequest.
+      # If all request data is received before or when the threshold is reached, the entity data is dispatched as a strict entity
+      # which allows more efficient processing of the request data without involving streams.
+      #
+      # If the given value is 0, the request is immediately dispatched and always carries a stream of data.
+      # You can use value `1` to create a strict entity if the request contains at most a single data frame, and a streamed
+      # entity otherwise.
+      #
+      # If a strict entity is dispatched, the data size will be at least the amount given here.
```
raboof, line 271: If the HEADERS has `endStream` set, don't we dispatch an empty 'strict' entity?

raboof (approved):

akka-http-core/src/main/resources/reference.conf
jrudolph: Yes, this is misleading, I just removed the sentence.

akka-http-core/src/main/resources/reference.conf
jrudolph: Yes, it's because of this note that I wouldn't want to enable it by default right now. We should enable it in akka-grpc, though.

johanandren (approved):
LGTM!

PR #3830, opened 2021-06-07
core: cache ContentType rendering

jrudolph:


akka-http-core/src/main/scala/akka/http/scaladsl/model/ContentType.scala
```diff
@@ -48,6 +48,9 @@ sealed trait ContentType extends jm.ContentType with ValueRenderable {
   def mediaType: MediaType
   def charsetOption: Option[HttpCharset]
 
+  private[this] lazy val _value: String = super.value
+  override def value: String = _value
```
patriknw, line 52: if it's only a cache we could skip the `lazy`:
```
  // cached value, visibility not important
  private[this] val _value: String = null
  override def value: String = {
    if (_value eq null)
      _value = super.value
    _value
  }
```

PR #3832, opened 2021-06-08
http: only render date-header once per second

jrudolph:
Previously we cached the rendering but rendering actually was not the most expensive part but calling `System.currentTimeMillis` for every response. In this scheme, we only update once a second with the trade-off that reported times may be up to 1 second in the past.

It applies both to HTTP/2 and HTTP/1.1.

Before:

```
[info] Benchmark                                          Mode  Cnt       Score       Error  Units
[info] ServerProcessingBenchmark.benchRequestProcessing  thrpt    3  209478.673 ± 48778.234  ops/s
```

After:

```
[info] Benchmark                                          Mode  Cnt       Score       Error  Units
[info] ServerProcessingBenchmark.benchRequestProcessing  thrpt    3  270904.245 ± 29405.551  ops/s
```

which shows that this was a big issue for the HTTP/1.1 pipeline. The benchmark doesn't include networking so that improvement will be much less prominent for more real-world benchmarks.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/rendering/DateHeaderRendering.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/rendering/DateHeaderRendering.scala
@@ -35,17 +35,8 @@ import scala.concurrent.ExecutionContext
     case object Idle extends DateState
     case class AutoUpdated(value: String) extends DateState {
       var wasUsed: Boolean = false
-      private[this] var _pair: (String, String) = _
-      def headerPair: (String, String) = {
-        if (_pair eq null) _pair = "date" -> value
-        _pair
-      }
-      private[this] var _bytes: Array[Byte] = _
-      def headerBytes: Array[Byte] = {
-        if (_bytes eq null)
-          _bytes = (new ByteArrayRendering(48) ~~ headers.Date ~~ value ~~ CrLf).get
-        _bytes
-      }
+      val headerPair: (String, String) = "date" -> value
+      val headerBytes: Array[Byte] = (new ByteArrayRendering(48) ~~ headers.Date ~~ value ~~ CrLf).get
     }
     val dateState = new AtomicReference[DateState](Idle)
     val updateInterval = 1.second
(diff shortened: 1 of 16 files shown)


raboof (approved):
Nice!

akka-http-core/src/main/scala/akka/http/impl/engine/rendering/DateHeaderRendering.scala
```diff
@@ -0,0 +1,100 @@
...
+          a
+        case s @ Idle =>
+          val r = if (rendered ne null) rendered else renderValue()
+          val newValue = AutoUpdated(r)
+          newValue.wasUsed = true
+          // use CAS to avoid that multiple accessing threads schedule multiple timers
+          if (!dateState.compareAndSet(s, newValue)) get(rendered)
+          else {
+            import scala.concurrent.duration._
+            scheduler.scheduleOnce(1.second)(autoUpdate())
```
patriknw, line 83: define `1.second` as a constant

patriknw (approved):
LGTM

akka-http-core/src/main/scala/akka/http/impl/engine/rendering/DateHeaderRendering.scala
```diff
@@ -0,0 +1,102 @@
...
+      var wasUsed: Boolean = false
+      private[this] var _pair: (String, String) = _
+      def headerPair: (String, String) = {
+        if (_pair eq null) _pair = "date" -> value
+        _pair
+      }
+      private[this] var _bytes: Array[Byte] = _
+      def headerBytes: Array[Byte] = {
+        if (_bytes eq null)
+          _bytes = (new ByteArrayRendering(48) ~~ headers.Date ~~ value ~~ CrLf).get
```
jrudolph, line 46: I wonder if this is really safe. This might be initialized by multiple threads at the same time. `ByteArrayRendering.get` uses `java.util.Arrays.copyOfRange` to fill the result array which is not an atomic operation that establishes a happens-before relationship (in contrast, to creating the tuple above which actually should be safe IIUC).

So, we should either prepopulate the value or make it `@volatile`.

johanandren: Isn't prepopulating fine if we anyway toggle to idle when there is no load? Creating the full header once per second worst case can't be very expensive?

jrudolph: Yep, I thought it might make sense because because we often only use one of `headerBytes` or `headerPair`, but shouldn't make a difference whether to render it once per second or not, indeed.

PR #3838, opened 2021-06-15
http2: remove Http2FrameRendering stage

jrudolph:
The whole reason of the stage was to split DataFrames into chunks fitting
the current SETTINGS_MAX_FRAME_SIZE but that invariant has now for some
time been enforced by the multiplexer directly.


patriknw (approved):
LGTM, nice

PR #3841, opened 2021-06-16
http2: never create Default entity

jrudolph:
This was supposed to be a small optimization to avoid having to wrap data elements into `Chunks` when it was unlikely that we expected a trailing header. The heuristic was whether we have seen a `content-length` header or not. However, with more of our optimizations applied, gRPC may actually send a
content-length header on the server side which means we miss out on trailers on the client side.

Let's revert this optimization for now, so that we can receive trailing headers in all cases again.

This should help unbreak tests at https://github.com/akka/akka-grpc/pull/1365


patriknw (approved):
LGTM

jrudolph:
The tests are failing because of a more fundamental issue we need to discuss. I don't think it's critical, so I'll just disable the failing tests pending a better solution of #3843.

PR #3842, opened 2021-06-16
http2: don't recordStateChange when debug logging is disabled

jrudolph:


akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Multiplexer.scala
```diff
@@ -286,21 +286,21 @@ private[http2] trait Http2MultiplexerSupport { logic: GraphStageLogic with Stage
...
     }
 
     /** Logs DEBUG level timing data for the output side of the multiplexer*/
-    def reportTimings(): Unit = {
+    def reportTimings(): Unit = debug {
       val timingsReport = timings.toSeq.sortBy(_._1).map {
         case (name, nanos) => f"${nanos / 1000000}%5d ms $name"
       }.mkString("\n")
-      debug(s"Timing data for connection\n$timingsReport")
+      s"Timing data for connection\n$timingsReport"
```
patriknw, line 303: it's not nice to have line breaks in logs, but that is another matter

patriknw (approved):
LGTM

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Multiplexer.scala
jrudolph: That part is currently not even used anywhere :)

PR #3845, opened 2021-06-21
http2: convert stream handling map to LongMap

jrudolph:
To avoid the Map overhead.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
@@ -16,7 +16,7 @@ import akka.stream.scaladsl.{ Sink, Source }
 import akka.stream.stage.{ GraphStageLogic, InHandler, OutHandler }
 import akka.util.{ ByteString, OptionVal }
 
-import scala.collection.immutable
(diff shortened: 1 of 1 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -52,7 +51,7 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
    */
   def tryPullSubStreams(): Unit
 
-  private var streamStates = new immutable.TreeMap[Int, StreamState]
+  private val streamStates = new java.util.Hashtable[Int, StreamState](settings.maxConcurrentStreams)
```
patriknw, line 54: This will box the Int key. Would it be better to use `scala.collection.mutable.LonMap`?

jrudolph: I can try that.

jrudolph: It's hard to measure a difference but LongMap seems slightly faster, indeed.

patriknw (approved):
LGTM

PR #3846, opened 2021-06-21
core: cache ModeledHeader.value

jrudolph:
This prevents that cached header instances have to re-render their value every time.

The assumption is that the lazy value access / initialization cost is negligible in comparison with rendering.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
@@ -62,8 +62,11 @@ sealed trait ModeledHeader extends HttpHeader with Serializable {
   def renderInRequests: Boolean = false // default implementation
   def renderInResponses: Boolean = false // default implementation
   def name: String = companion.name
-  private lazy val _value = renderValue(new StringRendering).get
-  def value: String = _value
+  private var _value: String = _
+  def value: String = {
+    if (_value eq null) _value = renderValue(new StringRendering).get
+    _value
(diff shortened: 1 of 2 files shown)


patriknw (approved):
LGTM

raboof (commented):
And headers that are produced often are likely to be cached, and thus their value is likely to be determined often, making this likely to be helpful.

Should we benchmark to verify?

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
```diff
@@ -62,7 +62,11 @@ sealed trait ModeledHeader extends HttpHeader with Serializable {
   def renderInRequests: Boolean = false // default implementation
   def renderInResponses: Boolean = false // default implementation
   def name: String = companion.name
-  def value: String = renderValue(new StringRendering).get
+  private var _value: String = _
+  def value: String = {
+    if (_value eq null) _value = renderValue(new StringRendering).get
```
jrudolph, line 67: Changed to cheaper initialization method.

jrudolph:
> And headers that are produced often are likely to be cached, and thus their value is likely to be determined often, making this likely to be helpful.
> 
> Should we benchmark to verify?

It turned up in benchmarks that's why I'm adding it. The question is rather, whether the benchmark is a realistic scenario or not. This optimization only helps if concrete header values are reused. This may or may not be true for any usage. In any case, it gives the user the choice to cache header values to avoid the rendering overhead.

Issue #3844, opened 2021-06-21
Move `isStrict` method to base WebSocket message class


harpocrates:
### Short description

As far as I can tell, all WebSocket message types have a `def isStrict(): Boolean` method. I propose to move this method to the base class `akka.http.javadsl.model.ws.Message`.

### Details

Since the `def isStrict(): Boolean` is defined on `akka.http.javadsl.model.ws.BinaryMessage` and `akka.http.javadsl.model.ws.TextMessage` but not on `akka.http.javadsl.model.ws.Message`, I have to do an unnecessary pattern match in order to call `.isStrict` on a `akka.http.javadsl.model.ws.Message` (or a `akka.http.scaladsl.model.ws.Message`):

```scala
val isStrict: Boolean = message match {
  case b: BinaryMessage => b.isStrict
  case t: TextMessage => t.isStrict
}
```

It would be more ergonomic to be able to just do

```scala
val isStrict: Boolean = message.isStrict
```

johanandren:
I think you pretty much always will have to determine binary or text anyway, but I agree that it makes sense to have `isStrict` on `Message`. Would you be up for a PR adding it?

harpocrates:
> I think you pretty much always will have to determine binary or text anyway, but I agree that it makes sense to have isStrict on Message. Would you be up for a PR adding it?

I wanted `isStrict` in the context of detecting and debugging some entity subscription timeouts for non-strict payloads. The motivation is similar to `isText: Boolean`.

jrudolph:
Fixed in #3847.

PR #3847, opened 2021-06-21
core: move `isStrict` up a level to ws.Message

harpocrates:
All websocket message types already define a `def isStrict: Boolean`.
The change here is just to declare `def isStrict: Boolean` on the base
class.

References #3844

### DIFF
--- a/akka-http-core/src/main/mima-filters/10.2.4.backwards.excludes/websocket-isstrict.excludes
+++ b/akka-http-core/src/main/mima-filters/10.2.4.backwards.excludes/websocket-isstrict.excludes
@@ -1,0 +1,2 @@
+# move `isStrict` method to superclass
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.javadsl.model.ws.Message.isStrict")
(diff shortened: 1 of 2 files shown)


raboof:
OK TO TEST

jrudolph (approved):
LGTM (but needs mima exclusions)

raboof:
> LGTM (but needs mima exclusions)

Should we also explicitly add a `@DoNotInherit` marker to `akka.http.javadsl.model.ws.Message`?

jrudolph:
> Should we also explicitly add a `@DoNotInherit` marker to `akka.http.javadsl.model.ws.Message`?

Yes that would be good, especially for Java where we cannot seal effectively.

akka-http-core/src/main/mima-filters/10.2.4.backwards.excludes/websocket-isstrict.excludes
```diff
@@ -0,0 +1,2 @@
+# move `isStrict` method to superclass
```
jrudolph, line 1:
```suggestion
# Move `isStrict` method to superclass which does not support user extension
```

PR #3848, opened 2021-06-22
http2: avoid creating GenericOutlet for every out element

jrudolph:
A nicer solution would be to be able to use an implicit value class
which isn't possible as an inner class.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/BufferedOutletSupport.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/BufferedOutletSupport.scala
@@ -99,7 +99,7 @@ private[http2] class BufferedOutletExtended[T](outlet: GenericOutlet[T]) extends
  */
 @InternalApi
 private[http2] trait GenericOutletSupport { logic: GraphStageLogic =>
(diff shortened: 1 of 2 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/http2/BufferedOutletSupport.scala
```diff
@@ -99,15 +99,15 @@ private[http2] class BufferedOutletExtended[T](outlet: GenericOutlet[T]) extends
...
+  def fromSubSourceOutlet[T](subSourceOutlet: SubSourceOutlet[T]): GenericOutlet[T] =
     new GenericOutlet[T] {
       def setHandler(handler: OutHandler): Unit = subSourceOutlet.setHandler(handler)
       def push(elem: T): Unit = subSourceOutlet.push(elem)
       def complete(): Unit = subSourceOutlet.complete()
       def fail(cause: Throwable): Unit = subSourceOutlet.fail(cause)
       def canBePushed: Boolean = subSourceOutlet.isAvailable
     }
-  implicit def fromOutlet[T](outlet: Outlet[T]): GenericOutlet[T] =
+  def fromOutlet[T](outlet: Outlet[T]): GenericOutlet[T] =
```
jrudolph, line 110: No implicit conversion to avoid accidental usage.

patriknw (approved):
LGTM

PR #3849, opened 2021-06-22
http2: import shaded twitter hpack library (APL2)

jrudolph:
Imported from latest version at https://github.com/twitter/hpack/commit/713612a439aa8f9b22cf4840054f9e7b3a787c02

This is in preparation of some important optimization wrt header parsing.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -52,8 +52,6 @@ object Dependencies {
     // For akka-http-testkit-java
     val junit       = "junit"                         % "junit"                        % junitVersion  // Common Public License 1.0
 
-    val hpack       = "com.twitter"                   % "hpack"                        % "1.0.2"       // ApacheV2
-
     val caffeine    = "com.github.ben-manes.caffeine" % "caffeine"                     % "2.9.0"
 
     val scalafix    = "ch.epfl.scala"                 %% "scalafix-core"               % Dependencies.scalafixVersion // grab from plugin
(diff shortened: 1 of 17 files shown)


raboof (approved):

patriknw (approved):
LGTM

project/Dependencies.scala
```diff
@@ -52,8 +52,6 @@ object Dependencies {
     // For akka-http-testkit-java
     val junit       = "junit"                         % "junit"                        % junitVersion  // Common Public License 1.0
 
-    val hpack       = "com.twitter"                   % "hpack"                        % "1.0.2"       // ApacheV2
```
julienrf, line 55: This is in binary incompatible change: users of akka-http used to have hpack on the classpath. Now, if they upgrade, that dependency will be missing.

jrudolph: Strictly taken that's true but does it have practical relevance?

julienrf: Consider the following scenario.

There is a module, `foo`, which depends on akka-http:10.2.4. That module can call the hpack API because it is brought by akka-http (although I consider this is bad practice — see https://github.com/cb372/sbt-explicit-dependencies, that could happen).

Now, consider another module, `bar`, which depends on `foo`, and on akka-http:10.2.5. The build tool resolves a conflict between akka-http:10.2.4 and akka-http:10.2.5. It picks akka-http:10.2.5 since it assumes it’s backward binary compatible with 10.2.4. However, by picking 10.2.5, hpack is not on the classpath anymore, and `foo` will get a `NoClassDefFoundError` when it will try to use hpack.

Does that make sense?

jrudolph: Thanks for the explanation. I understand it from a technical perspective but I'm not sure if it is a significant problem warranting a fix. Is this a theoretical question or have you run into it?

julienrf: I don't remember running into this issue in practice, so maybe you're right and this is not a so common situation...

PR #3850, opened 2021-06-22
http2: cache expensive `isDebugEnabled` call

jrudolph:
For Slf4j `isDebugEnabled` calls the Slf4jLoggingFilter which needs to look up
the concrete logger with `SLFLoggerFactory.getLogger` for each invocation.

In HTTP/2 we log a lot with debug so caching the value here can make a difference.

The disadvantage is that a dynamic logger level change will not directly apply
to already existing connections but only to new ones which seems like a
useful trade-off.

### DIFF
--- a/akka-parsing/src/main/scala/akka/macros/LogHelper.scala
+++ b/akka-parsing/src/main/scala/akka/macros/LogHelper.scala
@@ -18,6 +18,9 @@ import scala.reflect.macros.blackbox
 @InternalApi
 private[akka] trait LogHelper {
   def log: LoggingAdapter
+  def isDebugEnabled: Boolean = log.isDebugEnabled
+  def isInfoEnabled: Boolean = log.isInfoEnabled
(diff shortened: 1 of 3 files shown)


akka-parsing/src/main/scala/akka/macros/LogHelper.scala
```diff
@@ -37,7 +40,7 @@ private[akka] object LogHelper {
       {
         val logHelper = ctx.prefix.splice
         val log = logHelper.log
```
patriknw, line 42: `log` can be removed now, only used once below

patriknw (approved):
LGTM, just an additional cleanup...

akka-parsing/src/main/scala/akka/macros/LogHelper.scala
jrudolph: :+1:

PR #3851, opened 2021-06-22
core: cache lowercaseName in `ModeledCustomHeaderCompanion`

jrudolph:


patriknw (approved):
LGTM

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
```diff
@@ -92,7 +92,11 @@ abstract class CustomHeader extends jm.headers.CustomHeader {
  */
 abstract class ModeledCustomHeaderCompanion[H <: ModeledCustomHeader[H]] {
   def name: String
-  def lowercaseName: String = name.toRootLowerCase
+  private var _lowercaseName: String = _
```
jrudolph, line 95: I wondered if it was ok to just initialize from `name` (because it's not a trait) but it still contains an initialization order problem when `name` is only initialized from the constructor of the concrete implementation, like

```scala
// ...
object XHeader extends ModeledCustomHeaderCompanion[XHeader] {
  val name = "X-Header"
}
```

So, we have to leave it like this.

PR #3853, opened 2021-06-24
http2: use ByteString.length instead of ByteString.size

jrudolph:
ByteString.size is implemented by SeqOps which introduces a few more intermediate frames and indirection.

Turned up with ~ 1% in profiles.

(A better generic solution would be to be able to provide a concrete implementation in `ByteString` which isn't possible because `SeqOps.size` is marked final.)


patriknw:
Oh, could you take a look at usages in akka/akka also?

jrudolph:
> Oh, could you take a look at usages in akka/akka also?

I created https://github.com/akka/akka/issues/30342 to track this.

johanandren (approved):

PR #3854, opened 2021-06-24
docs: add Scaladex badge to ReadMe to show Scala version support

rtyley:
This badge summarises which versions of Scala are supported by Akka HTTP (and what the latest Akka HTTP version is for each of those Scala versions):

[akka-http-core Scala version support](https://index.scala-lang.org/akka/akka-http/akka-http-core)

More details on the badge format: https://github.com/scalacenter/scaladex/pull/660

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 Akka HTTP
 =========
 
-[![Latest version](https://index.scala-lang.org/akka/akka-http/akka-http-core/latest.svg)](https://index.scala-lang.org/akka/akka-http)
+[![akka-http-core Scala version support](https://index.scala-lang.org/akka/akka-http/akka-http-core/latest-by-scala-version.svg)](https://index.scala-lang.org/akka/akka-http/akka-http-core)
 
 The Akka HTTP modules implement a full server- and client-side HTTP stack on top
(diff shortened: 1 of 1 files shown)


raboof:
OK TO TEST

raboof:
> This badge summarises which versions of Scala are supported by Akka HTTP (and what the latest Akka HTTP version is for each of those Scala versions):
> 
> [akka-http-core Scala version support](https://index.scala-lang.org/akka/akka-http/akka-http-core)

That's neat! At some point it'd be nice to be able to remove particularly old versions from the list, but for now this is nice ;).

raboof (approved):

rtyley:
> At some point it'd be nice to be able to remove particularly old versions from the list

Yeah, that totally makes sense!

jrudolph:
Cool, thanks @rtyley.

jrudolph:
(It's a bit misleading for Scala 3, though, which should be fully supported even if we don't build against it).

raboof:
Maybe we should invent a 'should work with scala3' badge linking to documentation on how to add scala2 dependencies to scala3 projects?

Issue #3741, opened 2021-01-18
HTTP2: backoff behavior for persistent HTTP/2 connections on connection failures


jrudolph:
https://github.com/akka/akka-http/pull/3737 added the possibility to reconnect a maximum number of times. This functionality prevents a loop if the server is not available but does not solve the case of a transient failure of the backend.

We still want to provide backoff behavior if connection attempts fail. This will help with transient problems with the backend connection.

We can copy / extract the existing backoff logic from the existing HTTP/1 host connection pool, which provides the settings `base-connection-backoff` and `max-connection-backoff`.

jrudolph:
Fixed in #3857.

PR #3857, opened 2021-06-28
http2: back off on client persistent connection failure

johanandren:
References #3741

Pretty much copied the backoff logic from the NewConnectionPool as is.

### DIFF
--- a/akka-http-core/src/main/resources/reference.conf
+++ b/akka-http-core/src/main/resources/reference.conf
@@ -476,11 +476,11 @@ akka.http {
       max-persistent-attempts = 0
 
       # Starting backoff before reconnecting when a persistent HTTP/2 client connection fails
-      # see `akka.http.host-connection-pool.base-connection-backoff` for details on the backoff
+      # see `akka.http.host-connection-pool.base-connection-backoff` for details on the backoff mechanism.
       base-connection-backoff = ${akka.http.host-connection-pool.base-connection-backoff}
 
       # Max backoff between reconnect attempts when a persistent HTTP/2 client connection fails
-      # see `akka.http.host-connection-pool.max-connection-backoff` for details on the backoff
+      # see `akka.http.host-connection-pool.max-connection-backoff` for details on the backoff mechanism.
       max-connection-backoff = ${akka.http.host-connection-pool.max-connection-backoff}
 
       # When gracefully closing the HTTP/2 client, await at most `completion-timeout` for in-flight
--- a/akka-http-core/src/main/resources/reference.conf
+++ b/akka-http-core/src/main/resources/reference.conf
@@ -475,6 +475,14 @@ akka.http {
       # Set to zero to retry indefinitely.
       max-persistent-attempts = 0
 
+      # Starting backoff before reconnecting when a persistent HTTP/2 client connection fails
+      # see `akka.http.host-connection-pool.base-connection-backoff` for details on the backoff
+      base-connection-backoff = ${akka.http.host-connection-pool.base-connection-backoff}
+
+      # Max backoff between reconnect attempts when a persistent HTTP/2 client connection fails
+      # see `akka.http.host-connection-pool.max-connection-backoff` for details on the backoff
+      max-connection-backoff = ${akka.http.host-connection-pool.max-connection-backoff}
+
       # When gracefully closing the HTTP/2 client, await at most `completion-timeout` for in-flight
       # requests to complete.
       completion-timeout = 3s
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/OutgoingConnectionBuilderImpl.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/OutgoingConnectionBuilderImpl.scala
@@ -81,7 +81,11 @@ private[akka] object OutgoingConnectionBuilderImpl {
     }
 
     override def managedPersistentHttp2(): Flow[HttpRequest, HttpResponse, NotUsed] =
-      PersistentConnection.managedConnection(http2(), http2Settings.maxPersistentAttempts)
+      PersistentConnection.managedConnection(
+        http2(),
+        http2Settings.maxPersistentAttempts,
+        http2Settings.baseConnectionBackoff,
+        http2Settings.maxConnectionBackoff)
 
     override def http2WithPriorKnowledge(): Flow[HttpRequest, HttpResponse, Future[OutgoingConnection]] = {
       // http/2 prior knowledge plaintext
@@ -89,7 +93,11 @@ private[akka] object OutgoingConnectionBuilderImpl {
     }
 
     override def managedPersistentHttp2WithPriorKnowledge(): Flow[HttpRequest, HttpResponse, NotUsed] =
-      PersistentConnection.managedConnection(http2WithPriorKnowledge(), http2Settings.maxPersistentAttempts)
+      PersistentConnection.managedConnection(
+        http2WithPriorKnowledge(),
+        http2Settings.maxPersistentAttempts,
+        http2Settings.baseConnectionBackoff,
+        http2Settings.maxConnectionBackoff)
 
     override private[akka] def toJava: JOutgoingConnectionBuilder = new JavaAdapter(this)
   }
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/OutgoingConnectionBuilderImpl.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/OutgoingConnectionBuilderImpl.scala
@@ -83,9 +83,7 @@ private[akka] object OutgoingConnectionBuilderImpl {
     override def managedPersistentHttp2(): Flow[HttpRequest, HttpResponse, NotUsed] =
       PersistentConnection.managedConnection(
         http2(),
-        http2Settings.maxPersistentAttempts,
-        http2Settings.baseConnectionBackoff,
-        http2Settings.maxConnectionBackoff)
+        http2Settings)
 
     override def http2WithPriorKnowledge(): Flow[HttpRequest, HttpResponse, Future[OutgoingConnection]] = {
       // http/2 prior knowledge plaintext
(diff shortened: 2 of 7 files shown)


akka-http-core/src/main/resources/reference.conf
```diff
@@ -475,6 +475,14 @@ akka.http {
       # Set to zero to retry indefinitely.
       max-persistent-attempts = 0
 
+      # Starting backoff before reconnecting when a persistent HTTP/2 client connection fails
+      # see `akka.http.host-connection-pool.base-connection-backoff` for details on the backoff
```
jrudolph, line 479:
```suggestion
      # see `akka.http.host-connection-pool.base-connection-backoff` for details on the backoff mechanism.
```

akka-http-core/src/main/resources/reference.conf
```diff
@@ -475,6 +475,14 @@ akka.http {
...
+      base-connection-backoff = ${akka.http.host-connection-pool.base-connection-backoff}
+
+      # Max backoff between reconnect attempts when a persistent HTTP/2 client connection fails
+      # see `akka.http.host-connection-pool.max-connection-backoff` for details on the backoff
```
jrudolph, line 483:
```suggestion
      # see `akka.http.host-connection-pool.max-connection-backoff` for details on the backoff mechanism.
```

akka-http-core/src/main/scala/akka/http/impl/engine/http2/OutgoingConnectionBuilderImpl.scala
```diff
@@ -81,15 +81,23 @@ private[akka] object OutgoingConnectionBuilderImpl {
     }
 
     override def managedPersistentHttp2(): Flow[HttpRequest, HttpResponse, NotUsed] =
-      PersistentConnection.managedConnection(http2(), http2Settings.maxPersistentAttempts)
+      PersistentConnection.managedConnection(
```
jrudolph, line 84: Maybe just pass the whole settings here?

akka-http-core/src/main/scala/akka/http/impl/engine/http2/client/PersistentConnection.scala
```diff
@@ -128,12 +139,32 @@ private[http2] object PersistentConnection {
             failStage(new RuntimeException(s"Connection failed after $maxAttempts attempts", cause))
           } else {
             setHandler(requestIn, Unconnected)
-            log.info(s"Connection attempt failed: ${cause.getMessage}. Trying to connect again${connectsLeft.map(n => s" ($n attempts left)").getOrElse("")}.")
-            connect(connectsLeft)
+            val nextEmbargo = embargo match {
+              case Duration.Zero => baseEmbargo
+              case otherValue    => (otherValue * 2).min(maxBaseEmbargo)
+            }
+            if (nextEmbargo == Duration.Zero) {
```
jrudolph, line 146: This case is when the setting is disabled?

akka-http-core/src/main/scala/akka/http/impl/engine/http2/client/PersistentConnection.scala
```diff
@@ -128,12 +139,32 @@ private[http2] object PersistentConnection {
...
+              log.info(s"Connection attempt failed: ${cause.getMessage}. Trying to connect again${connectsLeft.map(n => s" ($n attempts left)").getOrElse("")}.")
+              connect(connectsLeft, nextEmbargo)
+            } else {
+              // FIXME is it fine to just stay in Connecting state here while backing off?
```
jrudolph, line 150: I guess, it's fine, the states basically describe the state in relation to outside communication, so we could see the backoff as a delayed "connecting" phase.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/client/PersistentConnection.scala
```diff
@@ -84,7 +94,8 @@ private[http2] object PersistentConnection {
         connected:    Future[OutgoingConnection],
         requestOut:   SubSourceOutlet[HttpRequest],
         responseIn:   SubSinkInlet[HttpResponse],
-        connectsLeft: Option[Int]
+        connectsLeft: Option[Int],
+        embargo:      FiniteDuration
```
jrudolph, line 98: I find the naming slightly confusing, maybe `embargo` -> `lastEmbargo`, `nextEmbargo` -> `embargo`?

jrudolph (commented):
Looking good (but needs some tests).

akka-http-core/src/main/scala/akka/http/impl/engine/http2/client/PersistentConnection.scala
johanandren: Yep, I think perhaps the logic could be made a little bit simpler.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/client/PersistentConnection.scala
johanandren: I was wondering if there is any up/downstream stop/failure handling that would need to be specific to the backoff, but maybe not.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/client/PersistentConnection.scala
```diff
@@ -49,30 +59,31 @@ private[http2] object PersistentConnection {
       StatusCodes.BadGateway,
       entity = "The server closed the connection before delivering a response.")
 
-  private class Stage(connectionFlow: Flow[HttpRequest, HttpResponse, Future[OutgoingConnection]], maxAttempts: Option[Int]) extends GraphStage[FlowShape[HttpRequest, HttpResponse]] {
+  private class Stage(connectionFlow: Flow[HttpRequest, HttpResponse, Future[OutgoingConnection]], maxAttempts: Option[Int], baseEmbargo: FiniteDuration, _maxBackoff: FiniteDuration) extends GraphStage[FlowShape[HttpRequest, HttpResponse]] {
```
jrudolph, line 62: Maybe settings here as well?

johanandren:
WDYT @jrudolph is the adapted existing test cases enough or should I add some explicit tests verifying that there actually is a backoff as well (I saw that through printlns/logging but that doesn't help much against regression I guess)?

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2PersistentClientSpec.scala
```diff
@@ -78,181 +79,188 @@ abstract class Http2PersistentClientSpec(tls: Boolean) extends AkkaSpecWithMater
...
-      "when connecting keeps failing" inAssertAllStagesStopped new TestSetup(tls) {
-        override def clientSettings = super.clientSettings
-          .withTransport(ClientTransport.withCustomResolver((_, _) => {
-            Future.successful(new InetSocketAddress("example.invalid", 80))
-          }))
+    (("with no backoff" -> ((settings: Http2ClientSettings) => settings.withBaseConnectionBackoff(Duration.Zero))) ::
+      ("with backoff" -> ((settings: Http2ClientSettings) => settings.withBaseConnectionBackoff(300.millis).withMaxConnectionBackoff(800.millis))) ::
+      Nil).foreach {
+        case (desc, changeSettings) =>
+          s"transparently reconnect when connection is closed ($desc)" should {
```
jrudolph, line 86: Usually we just create a function which contains the example and call it with different parameters.

```scala
def reconnectionTest(withBackoff: Boolean) = ...

reconnectionTest(true)
reconnectionTest(false)
```

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2PersistentClientSpec.scala
```diff
@@ -78,181 +79,188 @@ abstract class Http2PersistentClientSpec(tls: Boolean) extends AkkaSpecWithMater
...
-      }
-    }
-
-    "eventually fail" should {
-      "when connecting keeps failing" inAssertAllStagesStopped new TestSetup(tls) {
-        override def clientSettings = super.clientSettings
-          .withTransport(ClientTransport.withCustomResolver((_, _) => {
-            Future.successful(new InetSocketAddress("example.invalid", 80))
-          }))
+    (("with no backoff" -> ((settings: Http2ClientSettings) => settings.withBaseConnectionBackoff(Duration.Zero))) ::
```
jrudolph, line 82: Ok, so that tests at least that it still works in general. Would be good to have some test that tests that we don't do excessive reconnects when the setting is enabled.

jrudolph (commented):
Some comments

jrudolph:
> WDYT @jrudolph is the adapted existing test cases enough or should I add some explicit tests verifying that there actually is a backoff as well (I saw that through printlns/logging but that doesn't help much against regression I guess)?

I guess the actually doesn't have to be tested but it would be good to check that the new function avoids a connect loop. We already have the test for `maxAttempts` maybe you can add a check that verifies that it doesn't fail immediately but only after some time.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2PersistentClientSpec.scala
```diff
@@ -78,181 +79,188 @@ abstract class Http2PersistentClientSpec(tls: Boolean) extends AkkaSpecWithMater
...
+              client.sendRequest(
+                HttpRequest(
+                  method = HttpMethods.POST,
+                  entity = "ping",
+                  headers = headers.`Accept-Encoding`(HttpEncodings.gzip) :: Nil
+                )
+                  .addAttribute(requestIdAttr, RequestId("request-1"))
+              )
+              // need some demand on response side, otherwise, no requests will be pulled in
+              client.responsesIn.request(1)
```
jrudolph, line 259: It seems this test was missing the right assertion. There should be an `client.responsesIn.expectError` here. Without backoff it should be immediate. With backoff enabled you should be able to add an `client.responsesIn.expectNoMessage` with an appropriate timeout before.

akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2PersistentClientSpec.scala
johanandren: Alright, that would be a bit cleaner, will fix.

jrudolph (approved):
LGTM, great to have that feature for HTTP/2 as well!

PR #3861, opened 2021-06-29
Optimize HTTP/2 header parsing by caching some header models directly in HPACK tables

jrudolph:
Aside from the write coalescing work (which is now continued in https://github.com/akka/akka/pull/30334) this is the biggest chunk of performance improvements that I have queued.

Tests should be fine but some cleanups are still pending.

### DIFF
--- a/akka-http-core/src/main/java/akka/http/shaded/com/twitter/hpack/Decoder.java
+++ b/akka-http-core/src/main/java/akka/http/shaded/com/twitter/hpack/Decoder.java
@@ -19,6 +19,7 @@ package akka.http.shaded.com.twitter.hpack;
 import java.io.IOException;
 import java.io.InputStream;
 
+import akka.http.impl.util.StringTools;
 import akka.http.shaded.com.twitter.hpack.HpackUtil.IndexType;
 
 import static akka.http.shaded.com.twitter.hpack.HeaderField.HEADER_ENTRY_OVERHEAD;
@@ -530,7 +531,7 @@ public final class Decoder {
     } else {
       result = buf;
     }
-    return new String(result, 0,0, result.length);
+    return StringTools.asciiStringFromBytes(result);
   }
 
   // Unsigned Little Endian Base 128 Variable-Length Integer Encoding
--- a/akka-http-core/src/main/java/akka/http/shaded/com/twitter/hpack/Decoder.java
+++ b/akka-http-core/src/main/java/akka/http/shaded/com/twitter/hpack/Decoder.java
@@ -1,4 +1,10 @@
 /*
+ * Copyright (C) 2009-2021 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+/*
+ * Adapted from github.com/twitter/hpack with this license:
+ *
  * Copyright 2014 Twitter, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
--- a/akka-http-core/src/main/java/akka/http/shaded/com/twitter/hpack/Decoder.java
+++ b/akka-http-core/src/main/java/akka/http/shaded/com/twitter/hpack/Decoder.java
@@ -464,17 +464,19 @@ public final class Decoder {
   private void indexHeader(int index, HeaderListener headerListener) throws IOException {
     if (index <= StaticTable.length) {
       HeaderField headerField = StaticTable.getEntry(index);
-      addHeader(headerListener, headerField.name, headerField.value, false);
+      Object parsed = addHeader(headerListener, headerField.name, headerField.value, headerField.parsedValue, false);
+      headerField.parsedValue = parsed;
     } else if (index - StaticTable.length <= dynamicTable.length()) {
       HeaderField headerField = dynamicTable.getEntry(index - StaticTable.length);
-      addHeader(headerListener, headerField.name, headerField.value, false);
+      Object parsed = addHeader(headerListener, headerField.name, headerField.value, headerField.parsedValue, false);
+      headerField.parsedValue = parsed;
     } else {
       throw ILLEGAL_INDEX_VALUE;
     }
   }
 
   private void insertHeader(HeaderListener headerListener, byte[] name, byte[] value, IndexType indexType) {
-    addHeader(headerListener, name, value, indexType == IndexType.NEVER);
+    Object parsedValue = addHeader(headerListener, name, value, null, indexType == IndexType.NEVER);
 
     switch (indexType) {
       case NONE:
@@ -482,7 +484,7 @@ public final class Decoder {
         break;
 
       case INCREMENTAL:
-        dynamicTable.add(new HeaderField(name, value));
+        dynamicTable.add(new HeaderField(name, value, parsedValue));
         break;
 
       default:
@@ -490,18 +492,19 @@ public final class Decoder {
     }
   }
 
-  private void addHeader(HeaderListener headerListener, byte[] name, byte[] value, boolean sensitive) {
+  private Object addHeader(HeaderListener headerListener, byte[] name, byte[] value, Object parsedValue, boolean sensitive) {
     if (name.length == 0) {
       throw new AssertionError("name is empty");
     }
     long newSize = headerSize + name.length + value.length;
     if (newSize <= maxHeaderSize) {
-      headerListener.addHeader(name, value, sensitive);
+      parsedValue = headerListener.addHeader(name, value, parsedValue, sensitive);
       headerSize = (int) newSize;
     } else {
       // truncation will be reported during endHeaderBlock
       headerSize = maxHeaderSize + 1;
     }
+    return parsedValue;
   }
 
   private boolean exceedsMaxHeaderSize(long size) {
--- a/akka-http-core/src/main/java/akka/http/shaded/com/twitter/hpack/Decoder.java
+++ b/akka-http-core/src/main/java/akka/http/shaded/com/twitter/hpack/Decoder.java
@@ -34,7 +34,7 @@ public final class Decoder {
   private static final IOException MAX_DYNAMIC_TABLE_SIZE_CHANGE_REQUIRED =
       new IOException("max dynamic table size change required");
 
-  private static final byte[] EMPTY = {};
+  private static final String EMPTY = "";
 
   private final DynamicTable dynamicTable;
 
@@ -51,7 +51,7 @@ public final class Decoder {
   private int skipLength;
   private int nameLength;
   private int valueLength;
-  private byte[] name;
+  private String name;
 
   private enum State {
     READ_HEADER_REPRESENTATION,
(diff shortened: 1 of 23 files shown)


jrudolph (commented):
Some notes

akka-http-core/src/main/java/akka/http/shaded/com/twitter/hpack/Decoder.java
```diff
@@ -464,44 +464,47 @@ private void readName(int index) throws IOException {
...
+      Object parsed = addHeader(headerListener, headerField.name, headerField.value, headerField.parsedValue, false);
+      headerField.parsedValue = parsed;
     } else {
       throw ILLEGAL_INDEX_VALUE;
     }
   }
 
-  private void insertHeader(HeaderListener headerListener, byte[] name, byte[] value, IndexType indexType) {
-    addHeader(headerListener, name, value, indexType == IndexType.NEVER);
+  private void insertHeader(HeaderListener headerListener, String name, String value, IndexType indexType) {
```
jrudolph, line 478: There was little value in the `byte[]`-based processing pipeline since ours is based on strings (and also HPACK did some ad-hoc conversions in unexpected places).

akka-http-core/src/main/scala/akka/http/impl/engine/http2/hpack/HeaderDecompression.scala
```diff
@@ -42,11 +47,34 @@ private[http2] object HeaderDecompression extends GraphStage[FlowShape[FrameEven
...
+            headers += name -> parsed
+            parsed
+          } else {
+            import Http2HeaderParsing._
+            def handle(parsed: AnyRef): AnyRef = {
+              headers += name -> parsed
+              parsed
+            }
+
+            name match {
```
jrudolph, line 63: We match by name instead of using a more dynamic structure (using the list of defined parsers in Http2HeaderParsing), because we expect the generated match being faster than a real HashMap (or similar). In fact, I looked at what Scala 2.13 generates for pure String matching. It basically uses a lookup switch on hashcodes, so that's basically a code-generated HashMap. Also this way, we avoid the overhead of dynamic dispatching the `parse()` invocations.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/client/ResponseParsing.scala
```diff
@@ -66,18 +66,24 @@ private[http2] object ResponseParsing {
         }
 
       } else remainingHeaders.head match {
-        case (":status", statusCodeValue) =>
+        case (":status", statusCodeValue: String) =>
```
jrudolph, line 69: Would make sense to convert this client-side match to a similar structure as on the server side. Let's postpone that to another PR.

jrudolph:
This is now pretty complete and ready for review.

ktoso:
Oh boy, awesome! It’s been a while I’ll give it a read just for fun 🤩

akka-http-core/src/main/scala/akka/http/impl/engine/http2/client/ResponseParsing.scala
```diff
@@ -66,18 +66,24 @@ private[http2] object ResponseParsing {
...
           checkUniquePseudoHeader(":status", status)
           checkNoRegularHeadersBeforePseudoHeader(":status", seenRegularHeader)
           rec(remainingHeaders.tail, statusCodeValue.toInt, contentType, contentLength, seenRegularHeader, headers)
 
-        case ("content-type", ct) =>
+        case ("content-type", contentTypeValue: ContentType) =>
```
johanandren, line 74: Could it be worth matching both value types into one top level case and then match the type to cut one case it needs to check for every header?

Or perhaps even benefit from the hashcode/lookup-switch mentioned below if we _never_ matched on value in these? Maybe worth looking into in a separate PR.

Edit: Maybe that is exactly what you were saying in the comment above.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/hpack/HeaderCompression.scala
```diff
@@ -44,7 +42,7 @@ private[http2] object HeaderCompression extends GraphStage[FlowShape[FrameEvent,
         case ParsedHeadersFrame(streamId, endStream, kvs, prioInfo) =>
           kvs.foreach {
             case (key, value) =>
-              encoder.encodeHeader(os, key.getBytes(HeaderDecompression.UTF8), value.getBytes(HeaderDecompression.UTF8), false)
+              encoder.encodeHeader(os, key, value.asInstanceOf[String], false)
```
johanandren, line 45: `case (key, value: String)` ?  (Also, why is this cast even needed, can the value be something else than a string here?)

akka-http-core/src/main/scala/akka/http/impl/engine/http2/hpack/HeaderDecompression.scala
```diff
@@ -4,35 +4,40 @@
...
-import akka.http.impl.engine.http2.Http2Compliance.Http2ProtocolException
 
 /**
  * INTERNAL API
  *
  * Can be used on server and client side.
  */
 @InternalApi
-private[http2] object HeaderDecompression extends GraphStage[FlowShape[FrameEvent, FrameEvent]] {
+private[http2] class HeaderDecompression(masterHeaderParser: HttpHeaderParser, parserSettings: ParserSettings) extends GraphStage[FlowShape[FrameEvent, FrameEvent]] {
```
johanandren, line 30:
```suggestion
private[http2] final class HeaderDecompression(masterHeaderParser: HttpHeaderParser, parserSettings: ParserSettings) extends GraphStage[FlowShape[FrameEvent, FrameEvent]] {
```

akka-http-core/src/main/scala/akka/http/impl/engine/http2/client/ResponseParsing.scala
```diff
@@ -87,7 +93,11 @@ private[http2] object ResponseParsing {
         case (name, _) if name.startsWith(":") =>
           malformedRequest(s"Unexpected pseudo-header '$name' in response")
 
-        case (name, value) =>
+        case (name, httpHeader: HttpHeader) =>
+          // FIXME: move validation to HeaderDecompression validateHeader(httpHeader)
+          rec(remainingHeaders.tail, status, contentType, contentLength, seenRegularHeader = true, headers += httpHeader)
```
johanandren, line 98: This fixme is about the entire block of checking rather than this last case?

johanandren (approved):
LGTM with some questions and suggestions

akka-http-core/src/main/java/akka/http/shaded/com/twitter/hpack/Encoder.java
```diff
@@ -171,21 +172,24 @@ private static void encodeInteger(OutputStream out, int mask, int n, int i) thro
   /**
    * Encode string literal according to Section 5.2.
    */
-  private void encodeStringLiteral(OutputStream out, byte[] string) throws IOException {
-    int huffmanLength = Huffman.ENCODER.getEncodedLength(string);
-    if ((huffmanLength < string.length && !forceHuffmanOff) || forceHuffmanOn) {
+  private void encodeStringLiteral(OutputStream out, String string) throws IOException {
+    byte[] stringBytes = new byte[string.length()];
+    Unsafe.copyUSAsciiStrToBytes(string, stringBytes);
```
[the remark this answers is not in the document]
jrudolph: That's a good point we could operate either using `charAt` or extracting the underlying `char[]` and using that.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/client/ResponseParsing.scala
jrudolph: > Edit: Maybe that is exactly what you were saying in the comment above.

Yep ;)

akka-http-core/src/main/scala/akka/http/impl/engine/http2/hpack/HeaderCompression.scala
jrudolph: > case (key, value: String)

:1+:

>  (Also, why is this cast even needed, can the value be something else than a string here?)

Because we have only model in `Http2Substream` for both incoming and outgoing streams. The parsing optimization here requires that we can put `AnyRef` into the map. It isn't used like that on the rendering side.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/client/ResponseParsing.scala
jrudolph: It's about this particular case and it was indeed missing here. I moved it to `HeaderDecompression` and also changed `RequestParsing` not to re-validate there.

The advantage of doing it in `HeaderDecompression` is that we only need to validate once per header that has once been successfully parsed into a model and then stored in the HPACK tables.

akka-http-core/src/main/java/akka/http/shaded/com/twitter/hpack/Decoder.java
```diff
@@ -515,17 +518,19 @@ private boolean exceedsMaxHeaderSize(long size) {
...
+    final byte[] result;
 
     if (huffmanEncoded) {
-      return Huffman.DECODER.decode(buf);
+      result = Huffman.DECODER.decode(buf);
     } else {
-      return buf;
+      result = buf;
     }
+    return new String(result, 0,0, result.length);
```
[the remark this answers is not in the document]
jrudolph: It is deprecated but it's also unfortunately the fastest way to create a string object from ASCII character bytes. It still exists in JDK 17, so I'm not super worried right now (but might make sense to move it to a helper, so it is easy to change out later).

johanandren (approved):

akka-http-core/src/main/java/akka/http/shaded/com/twitter/hpack/Decoder.java
```diff
@@ -34,7 +34,7 @@
   private static final IOException MAX_DYNAMIC_TABLE_SIZE_CHANGE_REQUIRED =
       new IOException("max dynamic table size change required");
 
-  private static final byte[] EMPTY = {};
+  private static final String EMPTY = "";
```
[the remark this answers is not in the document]
jrudolph: Added now a double license header with a note that we adapted the code.

akka-http-core/src/main/java/akka/http/shaded/com/twitter/hpack/Encoder.java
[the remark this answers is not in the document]
jrudolph: Now changed, so that the extra copy is only needed if huffman coding cannot be used.

PR #3862, opened 2021-07-01
core: don't include RequestTimeoutSupport if disabled

jrudolph:
This is a small optimization for HTTP/1.1.

The documentation of `request-timeout` already hints that you will get
no dynamic request-timeout support if you turn off the feature by setting
it to `infinite`. For some reason never made use of that flexibility.

It turns out that request timeout support is expensive even if it is enabled,
so this will help HTTP/1.1 performance a bit.

### DIFF
--- a/akka-http-core/src/main/resources/reference.conf
+++ b/akka-http-core/src/main/resources/reference.conf
@@ -43,7 +43,7 @@ akka.http {
     # the idle-timeout will kick in first and reset the TCP connection
     # without a response.
     #
-    # If this setting is not `infinite` the HTTP server layer attaches a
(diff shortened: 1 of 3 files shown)


johanandren (approved):

jrudolph:
Changed the value to turn the feature off to `off` because we had some tests in the docs that assumed that `infinite` would still allow setting request timeouts dynamically.

jrudolph:
Other test failure was a test instability recently introduced. Fixed here: https://github.com/akka/akka-http/pull/3864

johanandren (approved):

PR #3866, opened 2021-07-06
Use Github Actions for PR validation

jrudolph:


jrudolph:
This is only the initial version that should eventually replace the Jenkins PR validation tasks.

Extra features of Jenkins not covered here yet:

 * automatically format code and push to branch
 * give overview over tests
 * comment failures with reports

Extra features that are currently covered by Travis (not covered here):
 * whitesource checking / updates
 * publish snapshots and releases
 * publish docs

I'll add these in subsequent PRs.

jrudolph:
HttpAppSpec is failing when it reuses the ActorSystem of the spec. I'm pretty sure that's something I've seen before but never was able to track down.

.github/workflows/validate-and-test.yml
```diff
@@ -0,0 +1,73 @@
...
+        uses: actions/cache@v2
+        with:
+          path: project/**/target
+          key: build-target-${{ hashFiles('**/*.sbt', 'project/build.properties', 'project/**/*.scala') }}
+
+      - name: Autoformat
+        run: sbt +headerCreateAll +scalariformFormat +test:scalariformFormat
+
+      - name: Check for missing formatting
+        run: git diff --exit-code --color || { echo "[error] Unformatted code found. Please run 'sbt +headerCreate +test:headerCreate +scalariformFormat +test:scalariformFormat' and commit the reformatted code."; false; }
```
raboof, line 35: would be slightly easier to follow if we combined these into 1 step - but this is OK by me

raboof (approved):
Adding the diagnostic logging to `HttpAppSpec` so we have more to go on when it happens again seems fine to me.

.github/workflows/validate-and-test.yml
```diff
@@ -0,0 +1,73 @@
...
+
+jobs:
+  formatting-check:
+    name: Check for missing formatting
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Set up JDK ${{ matrix.JABBA_JDK }}
```
marcospereira, line 17: `matrix.JABBA_JDK` won't be available here.

marcospereira:
```suggestion
      - name: Set up JDK 8
```

.github/workflows/validate-and-test.yml
```diff
@@ -0,0 +1,73 @@
...
+          key: build-target-${{ hashFiles('**/*.sbt', 'project/build.properties', 'project/**/*.scala') }}
+
+      - name: Autoformat
+        run: sbt +headerCreateAll +scalariformFormat +test:scalariformFormat
+
+      - name: Check for missing formatting
+        run: git diff --exit-code --color || { echo "[error] Unformatted code found. Please run 'sbt +headerCreate +test:headerCreate +scalariformFormat +test:scalariformFormat' and commit the reformatted code."; false; }
+
+  compile-and-test:
+    name: Compile and test
```
marcospereira, line 38:
```suggestion
    name: Test / Scala ${{ matrix.SCALA_VERSION }} / Java ${{ matrix.JABBA_JDK }}
```

Just to make it slightly easier to read checks on Github UI.

.github/workflows/validate-and-test.yml
```diff
@@ -0,0 +1,73 @@
...
+        uses: coursier/cache-action@v6.2
+
+      - name: Cache Build Target
+        uses: actions/cache@v2
+        with:
+          path: project/**/target
+          key: build-target-${{ hashFiles('**/*.sbt', 'project/build.properties', 'project/**/*.scala') }}
+
+      # Quick testing for PR validation
+      - name: Validate pull request for JDK ${{ matrix.JABBA_JDK }}, Scala ${{ matrix.SCALA_VERSION }}
```
marcospereira, line 66:
```suggestion
      - name: Validate pull request
```

Removing since these will be in the job name.

.github/workflows/validate-and-test.yml
```diff
@@ -0,0 +1,73 @@
...
+        if: ${{ github.event_name == 'pull_request' }}
+        run: sbt ++${{ matrix.SCALA_VERSION }} test:compile validatePullRequest
```
marcospereira, line 68:
```suggestion
        run: sbt ++${{ matrix.SCALA_VERSION }} Test/compile validatePullRequest
```

To make sbt 1.5 happy.

.github/workflows/validate-and-test.yml
```diff
@@ -0,0 +1,73 @@
...
+
+      # Full testing for pushes
+      - name: Run all tests JDK ${{ matrix.JABBA_JDK }}, Scala ${{ matrix.SCALA_VERSION }}
```
marcospereira, line 71:
```suggestion
      - name: Run all tests
```

Removing since these will be in the job name.

.github/workflows/validate-and-test.yml
```diff
@@ -0,0 +1,73 @@
...
+        if: ${{ github.event_name == 'push' }}
+        run: sbt ++${{ matrix.SCALA_VERSION }} test:compile test mimaReportBinaryIssues
```
marcospereira, line 73:
```suggestion
        run: sbt ++${{ matrix.SCALA_VERSION }} Test/compile test mimaReportBinaryIssues
```

To make sbt 1.5 happy.

marcospereira, line 73: I would move Mima way up to the chain. It seems more efficient to fail it before running all the tests. Also, should it be part of pull requests validation?

marcospereira (commented):
Some suggestions.

.github/workflows/validate-and-test.yml
jrudolph: Great, thanks, these were left-over from a previous version...

.github/workflows/validate-and-test.yml
jrudolph: There will be more steps for that later for the automatic committing of the changes, so I'll guess I'll have another look then.

.github/workflows/validate-and-test.yml
jrudolph: The Github UI automatically adds annotations for matrix parameters, so this would only add clutter (also not sure if it would work at the job level first of all).

.github/workflows/validate-and-test.yml
jrudolph: I like the repetition here, it's not cluttering anything, and when you navigate to an error you will see immediately what configuration went wrong.

.github/workflows/validate-and-test.yml
jrudolph: Same as above.

.github/workflows/validate-and-test.yml
jrudolph: Mima is part of the `validatePullRequest` task, so it is part of PR validation.

For `master` pushes it doesn't really matter when the build fails, as long as it does fail and someone is notified.

jrudolph:
```suggestion
        run: sbt ++${{ matrix.SCALA_VERSION }} mimaReportBinaryIssues Test/compile test
```

jrudolph:
I added a small final change to not execute tests in parallel in Github Actions to try to avoid timeout-related test failures.

jrudolph:
> I added a small final change to not execute tests in parallel in Github Actions to try to avoid timeout-related test failures.

Which doesn't seem to affect build times after all...

PR #3869, opened 2021-07-08
http2: avoid calling LoggingAdapter.isDebugEnabled in multiplexer

jrudolph:
This was previously optimized in Http2Demux but missed here as well.


patriknw (approved):
LGTM

PR #3872, opened 2021-07-13
http2: a few simple microoptimizations that turned up in profiling

jrudolph:


akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Demux.scala
```diff
@@ -378,7 +378,7 @@ private[http2] abstract class Http2Demux(http2Settings: Http2CommonSettings, ini
 
       private def completeIfDone(): Unit = {
         val noMoreOutgoingStreams = isClosed(substreamIn) && activeStreamCount() == 0
-        val allOutgoingDataFlushed = isClosed(frameOut) || multiplexer.hasFlushedAllData
+        def allOutgoingDataFlushed = isClosed(frameOut) || multiplexer.hasFlushedAllData
```
jrudolph, line 381: Call will be avoided in next line if `!noMoreOutgoingStreams`

akka-http-core/src/main/scala/akka/http/impl/engine/http2/HttpMessageRendering.scala
```diff
@@ -96,7 +96,7 @@ private[http2] object HttpMessageRendering {
    * Mutates `headerPairs` adding headers related to content (type and length).
    */
   def addContentHeaders(headerPairs: VectorBuilder[(String, String)], entity: HttpEntity): Unit = {
-    if (entity.contentType != ContentTypes.NoContentType)
+    if (entity.contentType ne ContentTypes.NoContentType)
```
jrudolph, line 99: Condition that usually evaluates to `false`.

RayRoestenburg (approved):
LGTM

PR #3873, opened 2021-07-13
http2: use mutable.Queue for state in multiplexer

jrudolph:
To avoid expensive frequent updates to persistent collections like Set or Vector. Beneficial side-effect: states can all be constant values.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Multiplexer.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Multiplexer.scala
@@ -12,8 +12,9 @@ import akka.macros.LogHelper
 import akka.stream.stage.GraphStageLogic
 import akka.stream.stage.OutHandler
 import akka.stream.stage.StageLogging
+import com.github.ghik.silencer.silent
 
-import scala.collection.immutable
+import scala.collection.mutable
 
 /**
  * INTERNAL API
@@ -41,11 +42,10 @@ private[http2] trait Http2Multiplexer {
 }
 
 @InternalApi
-private[http2] sealed abstract class PullFrameResult {
-  def frame: DataFrame
-}
+private[http2] sealed abstract class PullFrameResult
 @InternalApi
 private[http2] object PullFrameResult {
+  final case object NothingToSend extends PullFrameResult
   final case class SendFrame(frame: DataFrame, hasMore: Boolean) extends PullFrameResult
   final case class SendFrameAndTrailer(frame: DataFrame, trailer: FrameEvent) extends PullFrameResult
 }
@@ -130,7 +130,10 @@ private[http2] trait Http2MultiplexerSupport { logic: GraphStageLogic with Stage
       private var _state: MultiplexerState = Idle
 
       def hasFlushedAllData: Boolean = allDataFlushed(_state)
-      private def allDataFlushed(state: MultiplexerState): Boolean = (state == WaitingForData || state == Idle)
+      private def allDataFlushed(state: MultiplexerState): Boolean = (state eq WaitingForData) || (state eq Idle)
+
+      private val controlFrameBuffer: mutable.Queue[FrameEvent] = new mutable.Queue[FrameEvent]
+      private val sendableOutstreams: mutable.Queue[Int] = new mutable.Queue[Int]
 
       private def updateState(transition: MultiplexerState => MultiplexerState): Unit = {
         val oldState = _state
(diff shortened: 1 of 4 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Multiplexer.scala
```diff
@@ -130,7 +130,10 @@ private[http2] trait Http2MultiplexerSupport { logic: GraphStageLogic with Stage
       private var _state: MultiplexerState = Idle
 
       def hasFlushedAllData: Boolean = allDataFlushed(_state)
-      private def allDataFlushed(state: MultiplexerState): Boolean = (state == WaitingForData || state == Idle)
+      private def allDataFlushed(state: MultiplexerState): Boolean = (state eq WaitingForData) || (state eq Idle)
+
+      private val controlFrameBuffer: mutable.Queue[FrameEvent] = new mutable.Queue[FrameEvent]
```
jrudolph, line 135: The implementation of `Queue` are wildly different between 2.12 and 2.13 but in 2.13, it's backed by a quite efficient `ArrayDequeue`.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Multiplexer.scala
```diff
@@ -139,34 +142,45 @@ private[http2] trait Http2MultiplexerSupport { logic: GraphStageLogic with Stage
...
+                if (sendableOutstreams.isEmpty) Idle
+                else WaitingForNetworkToSendData
               }
-            case PullFrameResult.SendFrameAndTrailer(_, trailer) =>
-              WaitingForNetworkToSendControlFrames(Vector(trailer), sendableOutstreams - streamId)
+            case PullFrameResult.SendFrameAndTrailer(frame, trailer) =>
+              send(frame)
+              controlFrameBuffer += trailer
+              WaitingForNetworkToSendControlFrames
+            case PullFrameResult.NothingToSend =>
```
jrudolph, line 180: We try to avoid removing elements from the queue to avoid scanning in `closeStream` below, so we just leave it in there and added a new return value of `NothingToSend` for `pullNextFrame` to be able to skip those entries here.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Multiplexer.scala
```diff
@@ -194,99 +214,89 @@ private[http2] trait Http2MultiplexerSupport { logic: GraphStageLogic with Stage
...
+          if (connectionWindowLeft == 0) {
+            sendableOutstreams += streamId
+            WaitingForConnectionWindow
+          } else sendDataFrame(streamId)
         def closeStream(streamId: Int): MultiplexerState = this
       }
 
       /** Not yet pulled but data waiting to be sent */
-      private[http2] case class WaitingForNetworkToSendControlFrames(controlFrameBuffer: immutable.Vector[FrameEvent], sendableOutstreams: immutable.Set[Int]) extends MultiplexerState {
-        require(controlFrameBuffer.nonEmpty)
```
jrudolph, line 204: We have a set of invariants that were either checked like this or inherent in the data model that we now don't check any more (e.g. in `Idle` both queues should be empty). Maybe we could add them back explicitly in `updateState` but maybe only enable them when `DEBUG` logging is enabled.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Multiplexer.scala
```diff
@@ -194,99 +214,89 @@ private[http2] trait Http2MultiplexerSupport { logic: GraphStageLogic with Stage
...
-        def withSendableOutstreams(sendableOutStreams: immutable.Set[Int]): WithSendableOutStreams
-
-        protected def sendNext(): MultiplexerState = {
-          val chosenId = prioritizer.chooseSubstream(sendableOutstreams)
-          sendDataFrame(chosenId, sendableOutstreams)
-        }
+      abstract class WithSendableOutStreams extends MultiplexerState {
+        protected def sendNext(): MultiplexerState =
+          if (prioritizer eq StreamPrioritizer.First)
+            sendDataFrame(sendableOutstreams.dequeue())
```
jrudolph, line 252: New fast-path for the currently only used `StreamPrioritizer` to avoid having to remove from the middle of the queue as in the general case below.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Multiplexer.scala
```diff
@@ -194,99 +214,89 @@ private[http2] trait Http2MultiplexerSupport { logic: GraphStageLogic with Stage
...
+        }
 
-        def closeStream(streamId: Int): MultiplexerState =
-          if (sendableOutstreams.contains(streamId)) {
-            val sendableExceptClosed = sendableOutstreams - streamId
-            copy(sendableOutstreams = sendableExceptClosed)
-          } else
-            this
+        def closeStream(streamId: Int): MultiplexerState = {
+          // leave stream in sendableOutstreams, to be skipped in sendNextFrame
```
raboof, line 244: I guess you mean in `sendDataFrame`?
```suggestion
          // leave stream in sendableOutstreams, to be skipped in sendDataFrame
```

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Multiplexer.scala
```diff
@@ -194,99 +214,89 @@ private[http2] trait Http2MultiplexerSupport { logic: GraphStageLogic with Stage
...
         def closeStream(streamId: Int): MultiplexerState =
-          if (sendableOutstreams.contains(streamId)) {
-            val sendableExceptClosed = sendableOutstreams - streamId
-
-            if (sendableExceptClosed.isEmpty)
-              if (pulled) WaitingForData else Idle
-            else withSendableOutstreams(sendableExceptClosed)
-          } else
-            this
+          // leave stream in sendableOutstreams, to be skipped in sendNextFrame
```
raboof, line 263:
```suggestion
          // leave stream in sendableOutstreams, to be skipped in sendDataFrame
```

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Multiplexer.scala
raboof: So the motivation is that looking for the stream in `closeStream` is more expensive than just keeping it in there until we encounter it here and go to `WaitingForData` / `WaitingForNetworkToSendData.onPull()`. That's not super obvious to me but I believe you :).

raboof (approved):

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Multiplexer.scala
jrudolph: Indeed, thanks.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Multiplexer.scala
jrudolph: > So the motivation is that looking for the stream in `closeStream` is more expensive than just keeping it in there until we encounter it here and go to `WaitingForData` / `WaitingForNetworkToSendData.onPull()`

Yes.

I guess it depends on how often `closeStream` is called. The intuition would say, probably not so often, as it's only needed if a stream is cancelled and that should be exceptional. So, removing single elements in `O(n)` would be fine as long as not a whole bunch of streams out of many is removed and we run into a `O(n^2)` situation.

Maybe I'll run another benchmark and see if that would happen in normal situations.

jrudolph: I found another reason for the current logic:

With the queues we allow duplicates in the queue to be able to keep `enqueueOutStream` idempotent (without having to scan the queue for existing elements). We currently might add duplicates to the queue e.g. in `OutStream.increaseWindow`.

To get rid of `NothingToSend`, we would have to ensure that a stream is never put into the queue twice by keeping a flag on the `OutStream` side whether we are currently enqueued or not. Not sure if that's worth it right now.

jrudolph: I added #3882 to make an attempt to revisit this complexity.

PR #3876, opened 2021-07-19
build: use Github Actions for publishing snapshots and releases

jrudolph:


.github/workflows/publish.yml
```diff
@@ -0,0 +1,62 @@
...
+
+      - name: Set up JDK 8
+        uses: olafurpg/setup-scala@v10
+        with:
+          java-version: adopt@1.8
+
+      - name: Cache Coursier cache
+        uses: coursier/cache-action@v6.2
+
+      - name: Publish Documentation
```
marcospereira, line 59: Maybe just add this as a step in the `publish-artifacts` job?

marcospereira (commented):
Minor suggestion, but LGTM.

.github/workflows/publish.yml
jrudolph: I had those together before but thought to speed it up a bit using paralellization this way. True, we lose a bit of work because everything needs to be compiled another time, but unidoc is still slow enough overall that the parallelization cuts off a few minutes.

PR #3878, opened 2021-07-20
build: dilate timeouts on Github Actions to try to avoid false positive test failures

jrudolph:


raboof (approved):

jrudolph:
I wondered why the PR validation task didn't fail the consistent failure seen on master: it was a change to the workflow which doesn't trigger a full build with `validatePullRequest`... Fix in #3881.

PR #3879, opened 2021-07-20
build: cancel outdated builds

jrudolph:
To avoid that a bunch of workflows is run when you push to a PR in quick
succession.


jrudolph:
Seems to work, see https://github.com/akka/akka-http/actions/runs/1048388724

raboof (approved):

PR #3880, opened 2021-07-20
tests: fix MessageSpec with higher test.timefactor values

jrudolph:
pushInput uses Probe.expectNext which uses the default remaining
timeout which is diluted as well. However, the eventually loop is
suppossed to fill up buffers but under dilation it is not guaranteed to
actual make enough process to fail within the time limit.

This failed the GH actions master job consistently since dilation was added in #3878.


jrudolph:
Going ahead with merge to fix master build and upcoming PRs.

PR #3907, opened 2021-10-01
docs: various documentation fixes

mattkohl:
No particular issue is relevant, just a documentation review.

Thanks for all your work on this toolkit!

### DIFF
--- a/docs/src/main/paradox/common/sse-support.md
+++ b/docs/src/main/paradox/common/sse-support.md
@@ -1,7 +1,7 @@
 # Server-Sent Events Support
 
 Server-Sent Events (SSE) is a lightweight and [standardized](https://www.w3.org/TR/eventsource/)
-protocol for pushing notifications from a HTTP server to a client. In contrast to WebSocket, which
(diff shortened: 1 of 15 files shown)


docs/src/main/paradox/common/sse-support.md
```diff
@@ -1,7 +1,7 @@
 # Server-Sent Events Support
 
 Server-Sent Events (SSE) is a lightweight and [standardized](https://www.w3.org/TR/eventsource/)
-protocol for pushing notifications from a HTTP server to a client. In contrast to WebSocket, which
+protocol for pushing notifications from an HTTP server to a client. In contrast to WebSocket, which
```
jrudolph, line 4: Good to be consistent about that!

(Incidentally, I prefer using "an HTTP" but once tried to confirm my usage with native English speakers but didn't get a conclusive answer... But maybe it depends on where people come from since some native speakers also seem to pronounce the letter "h" as "haitch" so that "a HTTP" would be more appropriate...)

jrudolph (approved):
LGTM, great, it seems you went over a lot of the docs! Much appreciated.

PR #3909, opened 2021-10-04
build: add fossa checks

jrudolph:
The check will only run after being merged.


raboof (approved):

jrudolph:
For some reason it built now for the first time and failed (because `olafurpg/setup-scala` seems to use the wrong `sbt` runner script?).

PR #3913, opened 2021-10-12
End HTTP/2 stream with empty DATA rather than empty HEADERS

raboof:
When ending the stream without any payload, use a DATA frame rather than
a HEADERS frame to work around https://github.com/golang/go/issues/47851.

Tested with https://github.com/raboof/http2-from-go-to-akka-http


jrudolph:
> Restarted the actions, since "Check for missing formatting" failed for an unknown reason.

Ah, I renamed the GH Action job recently and didn't update the required branch checks. Interestingly enough there's also a step with the same name which also fails but doesn't fail the build...

PR #3917, opened 2021-10-14
build: simplify check for missing formatting

### LINKED ISSUE #3915: Fix "Check for missing formatting" step in GH Action PR validator
For some reason that step currently always silently fails.

image

```
Error:  diff --exit-code --color || { echo "[error] Unformatted code found. Please run 'sbt +headerCreate +test:headerCreate +scalariformFormat +test:scalariformFormat' and commit the reformatted code."; false; }
```

jrudolph:
Fixes #3915


jrudolph:
I just removed the extra error message for now and the simple `git diff` invocation seems to work better.

PR #3922, opened 2021-10-25
core: discard outgoing headers containing line breaks

jrudolph:
Refs #3717

To avoid accidental response splitting when applications put unsanitized request
data into outgoing responses.

We previously discussed introducing a check in the header models which could avoid
overhead during rendering. There are a lot of header models, however, which
all had to be checked one by one for potential issues. This would require a significant
effort now to check all the existing headers and would also make it easy to
reintroduce problems later on when headers are added or changed.

The approach here requires that all headers are rendered using the new overload of
`Rendering.~~`. When that method is used, we render the header and check afterwards
that the rendered data does not contain any CR or LF characters.

The performance overhead seems negligible, the new method did not turn up in profiling.
An explanation why going over the rendered header data directly after it has been rendered
is so cheap could be that
 - it has good locality because the data has just been written
 - it has good branch prediction because test will almost never fail
 - even if it has linear cost for big headers, those big headers will also have significant
   memory allocation cost which might dwarf the extra checking

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/rendering/BodyPartRenderer.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/rendering/BodyPartRenderer.scala
@@ -130,7 +130,7 @@ private[http] object BodyPartRenderer {
     case x: RawHeader if (x is "content-type") || (x is "content-length") =>
       suppressionWarning(log, x, "illegal RawHeader")
 
-    case x => r ~~ x ~~ CrLf
+    case x => r ~~ x
   }
 
   /**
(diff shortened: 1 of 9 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/rendering/BodyPartRenderer.scala
```diff
@@ -130,7 +130,7 @@ private[http] object BodyPartRenderer {
     case x: RawHeader if (x is "content-type") || (x is "content-length") =>
       suppressionWarning(log, x, "illegal RawHeader")
 
-    case x => r ~~ x ~~ CrLf
+    case x => r ~~ x
```
raboof, line 133: OK, so the `~~ CrLf` can be skipped here (and in similar places) because it uses the new `~~` that checks the header and then adds the newline itself.

raboof (approved):
Agreed perhaps it should fail more loudly, otherwise LGTM

jrudolph:
> Agreed perhaps it should fail more loudly, otherwise LGTM

Now added a noisy warning when such a header is occurred.

PR #3933, opened 2021-11-04
build: remove silencer for good

jrudolph:


jrudolph:
Fixed 2.12 compilation problem (only one annotation of a type allowed per class).

raboof (approved):
Nice

PR #3943, opened 2021-11-18
[10.1 backport] core: discard outgoing headers containing line breaks (#3922)

jrudolph:
Refs akka#3717

To avoid accidental response splitting when applications put unsanitized request
data into outgoing responses.

We previously discussed introducing a check in the header models which could avoid
overhead during rendering. There are a lot of header models, however, which
all had to be checked one by one for potential issues. This would require a significant
effort now to check all the existing headers and would also make it easy to
reintroduce problems later on when headers are added or changed.

The approach here requires that all headers are rendered using the new overload of
`Rendering.~~`. When that method is used, we render the header and check afterwards
that the rendered data does not contain any CR or LF characters.

The performance overhead seems negligible, the new method did not turn up in profiling.
An explanation why going over the rendered header data directly after it has been rendered
is so cheap could be that
 - it has good locality because the data has just been written
 - it has good branch prediction because test will almost never fail
 - even if it has linear cost for big headers, those big headers will also have significant
   memory allocation cost which might dwarf the extra checking

(cherry picked from commit afecb315c4b8b09e8edcee6613ee68cb5f938b52)

### DIFF
--- a/akka-http-core/src/main/mima-filters/10.1.14.backwards.excludes/safe-header-rendering.excludes
+++ b/akka-http-core/src/main/mima-filters/10.1.14.backwards.excludes/safe-header-rendering.excludes
@@ -1,0 +1,5 @@
+# Add new methods to @InternalApi Rendering trait
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.impl.util.Rendering.mark")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.impl.util.Rendering.check")
+# for Scala 2.11
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.impl.util.Rendering.~~")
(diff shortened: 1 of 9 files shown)


akka-http-core/src/main/mima-filters/10.1.14.backwards.excludes/safe-header-rendering.excludes
```diff
@@ -0,0 +1,4 @@
+# Add new methods to @InternalApi Rendering trait
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.impl.util.Rendering.mark")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.impl.util.Rendering.check")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.http.impl.util.Rendering.~~")
```
jrudolph, line 4: Last line needed for Scala 2.11 support.

johanandren (approved):

PR #3944, opened 2021-11-18
[10.1] build: some conservative dependency upgrades

jrudolph:


project/Dependencies.scala
```diff
@@ -11,8 +11,8 @@ import scala.language.implicitConversions
 object Dependencies {
   import DependencyHelpers._
 
-  val jacksonVersion = "2.10.2"
-  val junitVersion = "4.13"
+  val jacksonVersion = "2.10.5.1"
```
jrudolph, line 14: Fossa caught a problem with the old version here: https://github.com/akka/akka-http/runs/4253494683?check_suite_focus=true

johanandren (approved):

PR #3949, opened 2021-11-18
[10.1] build: use latest genjavadoc

jrudolph:


jrudolph:
Merged directly as it will only affect the documentation.

PR #3950, opened 2021-11-18
build: don't publish snapshot docs from release-10.1 branch

jrudolph:
To avoid overwriting main snapshot docs...


ennru (approved):
LGTM.

PR #3957, opened 2021-12-01
build: validate on Scala 2.13.7

raboof:
Since that's what we publish with as well


raboof:
I like #3958 better ;)

PR #3958, opened 2021-12-01
build: keep Scala version in one place

raboof:
rather than both in GHA config and project/Dependencies.scala


jrudolph (approved):
LGTM, yep, that will help in the Future...

jrudolph:
Ah that now also helps to fix the required branch checks to `2.13` instead of having to update them all the time...

PR #3970, opened 2021-12-14
build: use fuzzy `+~` to switch Scala versions

raboof:
Keeping consistent with Akka (https://github.com/akka/akka/pull/30981)


jrudolph (approved):
LGTM

Issue #3378, opened 2020-07-20
Describe on how a logger can be customised in the Route-Chain (Java) / modify MDC


domdorn:
I need to customise the logger that gets returned by `extractLog( log -> () ) ` so I can attach the current correlationId / userId / other fields into the MDC. However, I'm unable to accomplish this with the documentation I found. This is what I got so far: 

```
  public static RouteAdapter extractCorrelationId(Function<String, Route> inner) {
    return optionalHeaderValueByName("X-CorrelationId", (corr) -> extractLog(log -> {
      String correlationId = corr.orElseGet(() -> {
        String c = UUID.randomUUID().toString();
        log.debug("request does not have a upstream correlationId, assigning " + c);
        return c;
      });
      return provide(correlationId, inner);
    }));
  }
```

and I was planning to decorate my root-route as follows: 
```
  public Route createRoute() {
    return extractCorrelationId(corr -> extractLog(log -> {
      LoggingAdapter newLog = log;
      var newMDC = newLog.mdc().$plus(Tuple2.apply("correlationId", corr)); // this is a MDC and not a LoggingAdapter :( 
      // how to bring the newMDC into the newLog ?  
      return withLog(newLog, () -> allRoutes());
    }));
  }

```

jrudolph:
You will have to create a new LoggingAdapter in the first place because at least that part of LoggingAdapters is not thread-safe and use `log.setMdc` or `log.mdc(map)` to set the mdc.

The other option is to use the slf4j APIs directly but then you'll have to make to set the MDC whenever you run a chunk of code in a new thread (or use the [automatic MDC propagation](https://developer.lightbend.com/docs/telemetry/current/extensions/mdc.html) in commercial Lightbend Telemetry).

I keep this ticket open to create documentation around this particular pattern.

alexklibisz:
I recently found a way to do this using the existing loggers. There are basically three directives:
- a directive that takes the RouteContext and provides it with a new DiagnosticMarkerBusLoggingAdapter
- a directive that extracts the DiagnosticMarkerBusLoggingAdapter from the RouteContext
- a directive that takes a key/value pair and appends it to the MDC map on the DiagnosticMarkerBusLoggingAdapter. 

As far as I can tell, this is enough to make it thread-safe: each request gets a new logger, and, AFAIK, each request's directives are executed serially, so the map updates must be serialized. In any case I also have tests that hit the routes in parallel to test the thread-safety. 

[...]

PR #3974, opened 2021-12-21
MdcLoggingDirectives (withMdcLogging, withMdcEntry, withMdcEntries, extractMarkerLog)

alexklibisz:
This is a first pass at four new directives to simplify MDC logging.

It didn't require any internal changes, rather just a matter of using existing APIs to expose a more convenient API for attaching MDC entries to a request's logger. 

My team has used this in production for a few months, so we figured it would be nice to contribute back. 

I'm happy to iterate on the API and implementation as needed. Some notes:

1. For documentation, I just mimicked some of the existing directive docs. I was able to get them rendered locally w/ paradox.
2. I left out the Java implementation and examples for now. Once we are happy with the API, I can go back and do the Java implementation and examples.

Related to #3378


.gitignore
```diff
@@ -10,3 +10,5 @@ target
 .*.swp
 *.vim
 .ensime*
+.bsp
```
alexklibisz, line 13: Intellij creates this build server protocol directory when importing projects with newer versions of SBT.

.gitignore
```diff
@@ -10,3 +10,5 @@ target
...
+.tool-versions
```
alexklibisz, line 14: Used by the [asdf version manager](https://github.com/asdf-vm/asdf) to allow easily toggling Java 8, Java 11, and Java 17. Specifically, I noticed the akka-http project would only build with Java 8, and I'm using 11 and 17 in all other projects.

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/MdcLoggingDirectivesSpec.scala
```diff
@@ -0,0 +1,117 @@
+package akka.http.scaladsl.server.directives
+
+import akka.event.{ DiagnosticMarkerBusLoggingAdapter, MarkerLoggingAdapter }
+import akka.http.scaladsl.model.StatusCodes
+import akka.http.scaladsl.server.RoutingSpec
+import scala.jdk.CollectionConverters._
+
+class MdcLoggingDirectivesSpec extends RoutingSpec {
```
alexklibisz, line 8: I'm wondering if there is a way to actually test logging output without making the tests project depend directly on logback-classic and akka-slf4j w/ a specific encoder? Maybe there's some mock testing logger that would let us inspect logs without actually going to the console?

akka-http/src/main/scala/akka/http/scaladsl/server/directives/MdcLoggingDirectives.scala
```diff
@@ -0,0 +1,63 @@
...
+  def withMdcLogging: Directive0 =
+    extractActorSystem
+      .flatMap { sys =>
+        mapRequestContext { ctx =>
+          ctx.log match {
+            // If it's already a DiagnosticMarkerBusLoggingAdapter, this is a no-op.
+            case _: DiagnosticMarkerBusLoggingAdapter => ctx
+            // Otherwise, we need to give the context a DiagnosticLoggingAdapter.
+            case _ =>
+              val (str, cls) = LogSource.fromAnyRef(sys, sys)
```
alexklibisz, line 27: LL 27 to 35 were purely type tetris on my part. They work, but I'm not 100% sure I setup the logger the right way. Very open to input here.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/MdcLoggingDirectives.scala
```diff
@@ -0,0 +1,63 @@
...
+}
+
+object MdcLoggingDirectives extends MdcLoggingDirectives {
+  // This is private because the only advantage of a DiagnosticLoggingAdapter is that you can mutate the MDC entries,
+  // which is better done through the withMdcEntry and withMdcEntries directive.
+  private val extractDiagnosticMarkerLog: Directive1[DiagnosticMarkerBusLoggingAdapter] =
+    withMdcLogging.tflatMap { _ =>
+      extractLog.flatMap { log =>
+        // This asInstanceOf is tested and virtually guaranteed to work, but is there any way to do it without casting?
+        val dmbla = log.asInstanceOf[DiagnosticMarkerBusLoggingAdapter]
```
Marcus-Rosti, line 59: Personally, I think you should match on the type and then either 1) create one if you can, or 2) be willing to fail

alexklibisz: Yeah, I think I can actually re-arrange things a bit to get rid of the asInstanceOf. Trying now.

alexklibisz: Ok, I was able to get rid of the casting in my last commit.

alexklibisz:
Hi @jrudolph, just pinging to see if you or the other akka folks have had a chance to look this over. Thanks.

jrudolph (changes requested):
Thanks for the contribution (and sorry for the delay).

I think the idea is nice to have directives to add MDC values.

I see two issues with the implementation:
 * `DiagnosticMarkerBusLoggingAdapter` is not thread-safe so the directive will probably overwrite MDCs randomly when requests are handled concurrently but the log is shared. On one hand, in most cases, `extractDiagnosticMarkerLog` will probably recreate a new log, so hopefully that one wouldn't be shared (however, you cannot be sure `ctx.log` could be a shared LoggingAdapter). On the other hand, even the recreated one is not safe to be used in a route because routes can executed concurrently and there may be race conditions and visibility issues even with the dedicated log.
 * the directives don't ensure that MDC values are removed when a request is not handled by an inner route, so the values would still apply even if the request is handled by a sibling route.

To me it seems that Akka's DiagnosticLoggingAdapter is not up for the job here. In the best case, we would need an immutable variant of DiagnosticLoggingAdapter that can safely be shared. Ultimately, this would be nice to have in Akka itself but we can start with building it here if you like.

alexklibisz:
Thanks for the thorough review @jrudolph. 

To your notes on thread-safety: I had assumed that once a request is "inside" of a directive, akka-http will evaluate everything serially. Sure, you can kick off your own Futures, etc., and those could concurrently access the logging adapter. But akka-http will serially execute nested directives like `withMdcEntry` for a single request. It seems like that might be an incorrect assumption? 

If you happen to know of any examples floating around that demonstrate concurrency issues, that would be very helpful. 
We have a nearly identical version of these directives deployed internally, along with some more aggressive tests for parallel invocations, and we haven't noticed any concurrency-related confusion (though it could be elusive).

I'll have to do some more research on how it would look to add an immutable DiagnosticLoggingAdapter, either here or in akka. I'm definitely interested in getting this correct, even if it takes a bit longer -- no major rush on my end.

.gitignore
alexklibisz: Done

alexklibisz:
I pushed up a couple small suggested changes.

I want to make sure I understand the thread-safety concerns here. It sounds like we're specifically concerned about a case like this:

1. Instantiate a singleton `DiagnosticMarkerBusLoggingAdapter`.
2. Use an existing directive like `mapRequestContext` to provide that singleton as the `log` for every request.
3. Since `log` is already a `DiagnosticMarkerBusLoggingAdapter`, any routes that use `extractDiagnosticMarkerLog` will not create a new instance of `DiagnosticMarkerBusLoggingAdapter`.
4. Any routes that use `withMdcEntries`, if called concurrently, will concurrently update the MDC map inside of the singleton. This is the race condition.

Right now I can't think of any other way to induce a race condition here.

And while the scenario above seems possible, it also seems like it is _already_ entirely possible (and also a bad idea) to instantiate a singleton  `DiagnosticMarkerBusLoggingAdapter` and share it among requests. The docs for `DiagnosticMarkerBusLoggingAdapter` already indicate it's not thread-safe. The new directives don't make it any easier to make this mistake.

If my premises above are valid, maybe a simple solution here would be to just create a new `DiagnosticMarkerBusLoggingAdapter` on every call to `extractDiagnosticMarkerLog`? Something like this:

```scala
private val extractDiagnosticMarkerLog: Directive1[DiagnosticMarkerBusLoggingAdapter] =
    extractActorSystem
      .flatMap { sys =>
        extractRequestContext
          .flatMap { ctx =>
            provide {
              ctx.log match {
                // If it's already a DiagnosticMarkerBusLoggingAdapter, create a new one, copying the old MDC.
                case oldDmbla: DiagnosticMarkerBusLoggingAdapter =>
                  import oldDmbla._
                  val newDmbla = new DiagnosticMarkerBusLoggingAdapter(bus, logSource, logClass, loggingFilterWithMarker)
                  newDmbla.mdc(oldDmbla.mdc)
                  newDmbla
                // Otherwise, we need to construct a DiagnosticLoggingAdapter.
                case _ =>
                  val (str, cls) = LogSource.fromAnyRef(sys, sys)
                  val filter = new DefaultLoggingFilter(sys.settings, sys.eventStream)
                  new DiagnosticMarkerBusLoggingAdapter(sys.eventStream, str, cls, filter)
              }
            }
          }
      }
```

jrudolph:
> If my premises above are valid, maybe a simple solution here would be to just create a new `DiagnosticMarkerBusLoggingAdapter` on every call to `extractDiagnosticMarkerLog`? Something like this:

I think that would be a good idea. I'd say it would be acceptable to use `DiagnosticMarkerBusLoggingAdapter` when it is never explicitly exposed and no one is likely to modify the MDC settings after creation.

Without that change something like

```scala
    withMdcEntry("user_id", "1234") {
      extractLog { log =>
           val fut = loadFromDB(log) // log used asynchronously here and may include mdc entries only defined below

           withMdcEntry("request_id", "abcd") {
              // and here again
              log.info("completing request")
              complete(StatusCodes.OK)
          }
        }
      }
    }
```

might go wrong.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/MdcLoggingDirectives.scala
```diff
@@ -0,0 +1,57 @@
...
+
+  def withMdcLogging: Directive0 =
+    extractDiagnosticMarkerLog
+      .flatMap { dmbla: DiagnosticMarkerBusLoggingAdapter =>
+        mapRequestContext { ctx =>
+          ctx.withLog(dmbla)
+        }
+      }
+
+  def extractMarkerLog: Directive1[MarkerLoggingAdapter] =
```
jrudolph, line 28: Do we need that, aren't all logger also `MarkerLoggingAdapter`s anyway even without `withMdcLogging` enabled?

akka-http/src/main/scala/akka/http/scaladsl/server/directives/MdcLoggingDirectives.scala
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * @groupname mdc Mdc logging directives
+ * @groupprio mdc 240
+ */
+trait MdcLoggingDirectives {
+
+  import MdcLoggingDirectives.extractDiagnosticMarkerLog
+
+  def withMdcLogging: Directive0 =
```
jrudolph, line 20: Can this be internal? Or maybe it's not needed at all when every `withMdc*` invocation creates a new logger?

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/MdcLoggingDirectivesSpec.scala
jrudolph: There's `EventFilter` in akka-testkit that can be used to intercept logging messages.

alexklibisz:
Thanks for the feedback @jrudolph . I'll pick back up on this in the next few days for another pass.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/MdcLoggingDirectives.scala
alexklibisz: >  aren't all logger also MarkerLoggingAdapters anyway even without withMdcLogging enabled?

I don't think so. I'm assuming you're implying we could just use `extractLog` instead of having this.

`extractLog` returns a `LoggingAdapter`: https://github.com/akka/akka-http/blob/90b5ecbc6f41b13af02153f3f61accb0458cf88e/akka-http/src/main/scala/akka/http/scaladsl/server/directives/BasicDirectives.scala#L279

`MarkerLoggingAdapter` has a few extra methods, e.g., https://github.com/akka/akka/blob/main/akka-actor/src/main/scala/akka/event/Logging.scala#L1848-L1849

Maybe it's true that the `LoggingAdapter` that gets returned from `extractLog` happens to always also extend `MarkerLoggingAdapter`, but that's not specified in the return type, so we can't access those methods.

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/MdcLoggingDirectivesSpec.scala
alexklibisz: 👍 I added a test using EventFilter in my last commit.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/MdcLoggingDirectives.scala
alexklibisz: I ended up merging this directive with the `extractMarkerLog` directive into a new directive called `withMarkerLoggingAdapter` that both sets and returns the new MDC adapter.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/MdcLoggingDirectives.scala
```diff
@@ -0,0 +1,64 @@
...
+                case existingAdapter: DiagnosticMarkerBusLoggingAdapter =>
+                  import existingAdapter._
+                  val newAdapter = new DiagnosticMarkerBusLoggingAdapter(bus, logSource, logClass, loggingFilterWithMarker)
+                  newAdapter.mdc(existingAdapter.mdc ++ entries)
+                  newAdapter
+                // If the logger is not already a DiagnosticLoggingAdapter, we create a new one with the given entries.
+                case _ =>
+                  val (str, cls) = LogSource.fromAnyRef(sys, sys)
+                  val filter = new DefaultLoggingFilter(sys.settings, sys.eventStream)
+                  val newAdapter = new DiagnosticMarkerBusLoggingAdapter(sys.eventStream, str, cls, filter)
```
alexklibisz, line 57: @jrudolph I'm curious if you know if there is a way to maintain the logger name and akka source properties here? For example, if someone had instantiated a logger with a specific logger name and akka source, then, as written, this code will reset those properties.

alexklibisz: I'll try that, thanks

akka-http/src/main/scala/akka/http/scaladsl/server/directives/MdcLoggingDirectives.scala
```diff
@@ -0,0 +1,64 @@
...
+      .flatMap { dmbla: DiagnosticMarkerBusLoggingAdapter =>
+        mapRequestContext { ctx =>
+          ctx.withLog(dmbla)
+        } & provide {
+          dmbla: MarkerLoggingAdapter
+        }
+      }
+  }
+
+  def withMdcEntries(mdc: (String, Any)*): Directive0 =
```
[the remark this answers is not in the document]
alexklibisz: Hmm, it replaces the entries at least in some sense. For example, if I currently have ("foo" -> 1) in the MDC map and then I withMdcEntries(("foo" -> 2)), I've replaced "foo".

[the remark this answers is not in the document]
johanandren: Ah, true, if they are present they are "overridden", ok, let's keep as is then.

johanandren:
Sorry for the silence here, @alexklibisz , would you be interested in completing this work with the new license?

alexklibisz:
> Sorry for the silence here, @alexklibisz , would you be interested in completing this work with the new license?

I'm not really sure how contribution has changed since the license change. I've been putting off this PR primarily due to that potential complexity. I won't be offended if you close the PR. I also won't be offended if someone else takes over this PR.

johanandren:
The only difference is that it would end up being released under the new (source available) BSL rather than the previous (open source) Apache License.

lightbend-cla-validator:
Hi @alexklibisz,

We see that you have signed the Lightbend Contributors License Agreement before, however, the CLA has changed since you last signed it.
Please review the new CLA and sign it before we proceed with reviewing this pull request:

akka-http/src/main/scala/akka/http/scaladsl/server/directives/MdcLoggingDirectives.scala
alexklibisz: Applied your suggestion in d0ac46e9dd03478e75c0c0d30b03100b59ef6e7c

alexklibisz:
I merged in main and applied the suggestion from this thread: https://github.com/akka/akka-http/pull/3974#discussion_r875025973

alexklibisz:
I am pretty confused by the CLA. What exactly changed since the last CLA? Obviously the overall project license changed. But is there a diff of the two CLAs somewhere so I can see what I'm re-agreeing to?

akka-http/src/main/scala/akka/http/scaladsl/server/directives/MdcLoggingDirectives.scala
```diff
@@ -0,0 +1,70 @@
...
+            provide {
+              // Convert the existing ctx.log LoggingAdapter to a DiagnosticMarkerBusLoggingAdapter.
+              // This requires various amounts of effort depending on the type of the current LoggingAdapter.
+              // In all cases, we create a _new_ DiagnosticMarkerBusLoggingAdapter in order to avoid race conditions,
+              // where multiple requests could add/remove entries from the same MDC map.
+              ctx.log match {
+                case existingAdapter: DiagnosticMarkerBusLoggingAdapter =>
+                  import existingAdapter._
+                  val newAdapter = new DiagnosticMarkerBusLoggingAdapter(bus, logSource, logClass, loggingFilterWithMarker)
+                  newAdapter.mdc(existingAdapter.mdc ++ entries)
```
alexklibisz, line 53: This part is concerning me a bit. Assume, for whatever reason, the logger is always already a singleton instance of `DiagnosticMarkerBusLoggingAdapter` (maybe defined once in `main` and provided to every request via `mapRequest` or similar). Then it seems the mdc map will never be cleared out, leading to larger and larger maps, i.e., a memory leak. There probably needs to be some case where we clear out this map.

johanandren: The original logging adapter is not mutated here though, so when `newAdapter` is GC:d the mdc map is as well. (If we would set the combined mdc on the original adapter without cleanup that would be a leak)

johanandren:
Very late retply, sorry: I _think_ the CLA does not differ, AFAICS it does not mention the license, you still keep your copyright etc., that the CLA checker requires re-sign is more of a technical inconvenience with our CLA checker/database.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/MdcLoggingDirectives.scala
alexklibisz: I think you're right. We're always copying the existing MDC entries and then adding values. We're not just perpetually adding to the original map.

alexklibisz:
> Hi @alexklibisz,
> 
> Thank you for your contribution! We really value the time you've taken to put this together.
> 
> We see that you have signed the Lightbend Contributors License Agreement before, however, the CLA has changed since you last signed it. Please review the new CLA and sign it before we proceed with reviewing this pull request:
> 
> https://www.lightbend.com/contribute/cla

Still need to clarify something regarding the CLA.

alexklibisz:
Tests should be passing on 2.12 and 2.13 now. Still need to clarify something regarding the CLA.

lightbend-cla-validator:
Hi @alexklibisz,

We see that you have signed the Lightbend Contributors License Agreement before, however, the CLA has changed since you last signed it.
Please review the new CLA and sign it before we proceed with reviewing this pull request:

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/MdcLoggingDirectivesSpec.scala
```diff
@@ -0,0 +1,95 @@
...
+        @volatile
+        var buf = Vector.empty[Logging.Info2]
+        override protected def matches(event: Logging.LogEvent): Boolean = event match {
+          case e: Logging.Info2 =>
+            buf :+= e
+            true
+          case _ => false
+        }
+      }
+      val filter = new EF(3)
```
johanandren, line 66: I'd recommend nesting 3 `EventFilter.custom` instead of the volatile field scheme, looking at the mdc values in the filters instead of after the fact. First log event will go innermost. I think it can match on exact log message and fail explicitly with a scalatest `fail` if the mdc isn't what it should be to provide extra detail.

PR #3976, opened 2022-01-04
core: add request timeout support to client-side pool APIs #42

rafal2nowak:
Reference [#42](https://github.com/akka/akka-http/issues/42)

### DIFF
--- a/akka-http-core/src/main/resources/reference.conf
+++ b/akka-http-core/src/main/resources/reference.conf
@@ -612,6 +612,12 @@ akka.http {
     # Set to `infinite` to allow the connection to remain open indefinitely (or be closed by the more general `idle-timeout`).
     keep-alive-timeout = infinite
 
+    # The maximum duration for a client to receive response from server for a given request. If there is no response within specified duration
+    # request fails with `RequestTimeoutException`. Later on if response arrives data is discarded regardless of HTTP status code.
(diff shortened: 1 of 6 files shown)


lightbend-cla-validator:
Hi @rafal2nowak,

rafal2nowak:
Hi,
I just signed the CLA.

jrudolph (commented):
Thanks a lot @rafal2nowak at taking a stab at this.

I'm not yet convinced that this is the approach we want to take (which is also one reason we never went ahead and implemented something so far).

This kind of request-timeout only takes into account the time a request was handed to the pool until a response is dispatched back to the user.

What people often want instead is a full-roundtrip timeout from dispatching a request to the pool (via `Http.singleRequest`) until the response is dispatched back. The other really important feature is that requests that are already timed out are cancelled as soon as possible. This means that these kinds of requests should not be dispatched if they have timed out in the queue. We might also want to give the option to be even more aggressive and reset pool slots if a response does not arrive in time.

So I guess, we are a bit suffering from sacrificing the good for the perfect solution here but if we accept this as a preliminary solution we need to be sure that we don't spent the `request-timeout` setting in a way that would prevent us from making it more aggressive later on.

rafal2nowak:
@jrudolph thank you for explanation. Given that I am closing this PR.

PR #3980, opened 2022-01-10
build: hide test jar with complicated name from scala-steward

jrudolph:
Scala steward is broken since a while and it seems the reason is this
jar file with the unicode name.

See https://gitter.im/scala-steward-org/scala-steward?at=6183bb66d78911028a1b7cd8

Probably, the scala-steward environment is not set up correctly to support unicode
file names completely.

/cc @fthomas


jrudolph:
Let's go with it and see if it fixes things for tonight's scheduled updates...

Issue #3662, opened 2020-11-25
HTTP client pool stops processing requests with "Exceeded configured max-open-requests value"


vilunov:
We use sttp v2 with single-request API (no streams). Some configs:
```hocon
akka.http.host-connection-pool {
  max-connections = 64
  max-open-requests = 128 // Must be a power of 2
}

akka.http.client {
  idle-timeout = 5 seconds
  connecting-timeout = 1 second  // this is ignored when used with sttp, but we introduced a workaround
}
```

At some point of time this happens:

![Screenshot_20201125_111803](https://user-images.githubusercontent.com/16502164/100200636-f36db400-2f0f-11eb-9e88-dccda2b6a312.png)

Here the requests to `typing` stopped working at 17:16, shortly after the server started at 17:11. Such freeze can happen at any point of the instance's life though, even days after the start.

Here are some hypotheses that we considered:
- The result stream is not consumed somehow and clogs the request pool -- stream consumption is handled by the sttp adapter.
- The freeze happens due to the pool's being overloaded -- we introduced timeouts and even with them the pools still freeze until the instance is restarted.
- Threads in the threadpool are being blocked -- in the heapdump most of the threads are either RUNNING or WAITING for new work.

Error log:
```
[...]

PR #3981, opened 2022-01-10
core: fail pool slot if connection is canceled before it is failed

jrudolph:
Refs #3662

In the issue, it was observed that a connection might only cancel a pool
connection but would not fail it (which was the implicit assumption of the
existing code). In this case a pool might get more and more unresponsive
with slots hanging after connections being canceled but not failed.

The exact condition could not be reproduced but let's be proactive here
and fail the slot if we see unexpected cancellation.


raboof (approved):
always tricky to add code fixing a problem we can't quite explain or reliably reproduce, but logically makes sense so let's see if it indeed has the desired effect.

jrudolph:
> always tricky to add code fixing a problem we can't quite explain or reliably reproduce, but logically makes sense so let's see if it indeed has the desired effect.

Yeah, might even be a 10.1.x-only problem, but it still seems valid to harden things a bit more.

Issue #3903, opened 2021-09-13
Allow parsing.max-content-length = 0


strelec:
java.lang.IllegalArgumentException: requirement failed: if set max-content-length must be > 0

This should simply require all requests to have empty bodies (for example, a server that only serves GET requests.)

jrudolph:
Thanks, @strelec, that might make sense (but there might be some reasons it was done the way it is). Would you like to create a PR for that feature?

GKuldeepak-Knoldus:
Hi Everyone, 
Has the issue been resolved? Also, lemme know if I'm wrong, we just need to change the value of `akka.http.client.parsing.max-content-length` to `0` from `20m` and if the issue is still open so, I would like to pick this issue.

jrudolph:
Thanks, @GKuldeepak-Knoldus. Yes, this issue is still up for grabs, so please go ahead!

GKuldeepak-Knoldus:
Thanks @jrudolph ,
I have raised a PR for the same. I have set the value to 0 for now, please lemme know, if it required anything further.

soo-underground:
Hi Everyone!
Since the linked PR has been rejected, can I try to fix this issue?

strelec:
Sure, have a crack at it.

PR #3982, opened 2022-01-10
Set the value of parsing.max-content-length = 0

GKuldeepak-Knoldus:
The value of `akka.http.client.parsing.max-content-length` has been set to `0` from `20m`

References #3903


jrudolph (changes requested):
That's not what we wanted to do actually :)

The default should be left untouched. The ticket is about allowing to set it to 0 and adding tests that setting it to 0 works correctly.

GKuldeepak-Knoldus:
Ok Got It, I Believe that this could be done by declaring the environment variables for the same. Plz lemme know If I can proceed with this, or something else is required!

jrudolph:
Actually not much to say here than what I wrote above:

> The ticket is about *allowing* to set `akka.http.client.parsing.max-content-length` to 0 and adding tests that setting it to 0 works correctly.

So the requirement needs to be adapted from `_ > 0` to `_ >=0` and then a test needs to be added that checks that if the setting is 0 no bodies at all are accepted. `ClientServerSpec` could be a good place for that.

I'm closing this one for now.

PR #4004, opened 2022-01-13
fixes #4003 ensure PoolMaster.shutdown(…) does not hang

### LINKED ISSUE #4003: `PoolMaster.shutdown` hangs indefinitely
Hello, 

I found a bug that leads to shutdown process hanging indefinitely.

`PoolMaster.shutdown(…)` call does not complete `shutdownCompletedPromise` for connection pool that has been shutdown because of `idle-timeout`

```
case Shutdown(poolId, shutdownCompletedPromise) =>
      statusById.get(poolId).foreach {
```

is never executed, because nothing is found in the `map`

t3hnar:
ensure PoolMaster.shutdown(…) does not hang

References #4003

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
@@ -164,18 +164,18 @@ private[http] final class PoolMasterActor extends Actor with ActorLogging {
 
     // Shutdown a pool and signal its termination.
     case Shutdown(poolId, shutdownCompletedPromise) =>
-      statusById.get(poolId).foreach {
-        case PoolInterfaceRunning(pool) =>
+      statusById.get(poolId) match {
+        case Some(PoolInterfaceRunning(pool)) =>
           // Ask the pool to shutdown itself. Queued connections will be resent here
           // to this actor by the pool actor, they will be retried once the shutdown
           // has completed.
           val completed = pool.shutdown()(context.dispatcher)
           shutdownCompletedPromise.tryCompleteWith(completed.map(_ => Done)(ExecutionContexts.sameThreadExecutionContext))
           statusById += poolId -> PoolInterfaceShuttingDown(shutdownCompletedPromise)
-        case PoolInterfaceShuttingDown(formerPromise) =>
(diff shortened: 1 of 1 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
```diff
@@ -164,15 +164,15 @@ private[http] final class PoolMasterActor extends Actor with ActorLogging {
...
           // to this actor by the pool actor, they will be retried once the shutdown
           // has completed.
           val completed = pool.shutdown()(context.dispatcher)
           shutdownCompletedPromise.tryCompleteWith(completed.map(_ => Done)(ExecutionContexts.sameThreadExecutionContext))
           statusById += poolId -> PoolInterfaceShuttingDown(shutdownCompletedPromise)
-        case PoolInterfaceShuttingDown(formerPromise) =>
+        case Some(PoolInterfaceShuttingDown(formerPromise)) =>
           // Pool is already shutting down, mirror the existing promise.
           shutdownCompletedPromise.tryCompleteWith(formerPromise.future)
         case _ =>
```
jrudolph, line 178: Maybe spell it out as `None` here.

jrudolph (approved):
LGTM, seems like the intention was there to make it work but maybe it was lost in the meantime?

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
t3hnar: I cannot - as we are covering more than just `None` case.
That line was not changed :)

jrudolph: Thanks!

t3hnar:
@jrudolph are you going to include this into the next minor version?

jrudolph:
> @jrudolph are you going to include this into the next minor version?

Yes, though there isn't anything planned soon. In the meantime, you can try out the snapshot (https://doc.akka.io/docs/akka-http/current/contributing.html) after it has been [built](https://github.com/akka/akka-http/actions/runs/1712737255).

PR #4012, opened 2022-01-30
fix warnings. `formatted` => `format`

xuwei-k:
- https://github.com/scala/scala/pull/9783
- https://github.com/scala/bug/issues/12471

```
warn] /home/runner/work/akka-http/akka-http/akka-http-core/src/main/scala/akka/http/impl/engine/http2/FrameLogger.scala:53:16: method formatted in class StringFormat is deprecated (since 2.12.16): Use `formatString.format(value)` instead of `value.formatted(formatString)`,
[warn] or use the `f""` string interpolator. In Java 15 and later, `formatted` resolves to the new method in String which has reversed parameters.
[warn]         .map(_ formatted "%02x")
[warn]                ^
```


jrudolph (approved):
LGTM, thanks good catch!

PR #4014, opened 2022-01-30
core: cleanup unused imports

nitikagarw:
References #xxxx

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Multiplexer.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Multiplexer.scala
@@ -4,8 +4,6 @@
 
 package akka.http.impl.engine.http2
 
-import akka.http.ccompat._
-
 import akka.annotation.InternalApi
 import akka.event.LoggingAdapter
 import akka.http.impl.engine.http2.FrameEvent._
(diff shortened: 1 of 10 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Multiplexer.scala
```diff
@@ -4,8 +4,6 @@
 
 package akka.http.impl.engine.http2
 
-import akka.http.ccompat._
```
jrudolph, line 7: All the ccompat ones are actually needed but only for Scala 2.12.

Marcus-Rosti: I think cats and any other library that uses implicits heavily has a lot of issues with ignoring unused imports.

jrudolph (approved):
LGTM (I removed the changes to files that shouldn't have been changed)

nitikagarw:
Thanks, @jrudolph. The aim of this PR was to mostly reduce the compile warnings and have cleaner logs on the terminal.

jrudolph:
> The aim of this PR was to mostly reduce the compile warnings and have cleaner logs on the terminal.

Yes, unfortunately, the `ccompat` imports are needed for 2.12 builds, so we need a different mechanism to remove the warning.

PR #4018, opened 2022-02-03
http2: implement graceful termination for HTTP/2 servers

### LINKED ISSUE #3580: Graceful shutdown for HTTP/2 servers
Graceful shutdown is currently only enabled for HTTP/1.1 servers. A potential strategy for HTTP/2 servers could look like this:

 * immediately send MAX_CONCURRENT_STREAMS = 0 to notify client that no new streams are accepted
 * refuse incoming streams in the meantime
 * during a grace period wait for running requests to be completed
 * afterwards close connection with a GOAWAY

All this functionality could be enabled with a setting and implemented in the demux when the user request stream is cancelled (with Akka 2.6 we could use cancellation causes to distinguish between graceful and hard shutdown dynamically).

jrudolph:
The functionality is built directly into the Http2Demux which
has the state of all streams readily available. It supports the
same interface as the HTTP/1 version, though.

Fixes #3580

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2.scala
@@ -233,7 +233,6 @@ private[http] final class Http2Ext(private val config: Config)(implicit val syst
       TLS(createEngine _, closing = TLSClosing.eagerClose)
 
     stack.joinMat(clientConnectionSettings.transport.connectTo(host, port, clientConnectionSettings)(system.classicSystem))(Keep.right)
-      .addAttributes(Http.cancellationStrategyAttributeForDelay(clientConnectionSettings.streamCancellationDelay))
   }
 
   def outgoingConnectionPriorKnowledge(host: String, port: Int, clientConnectionSettings: ClientConnectionSettings, log: LoggingAdapter): Flow[HttpRequest, HttpResponse, Future[OutgoingConnection]] = {
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2.scala
@@ -9,7 +9,7 @@ import akka.annotation.InternalApi
 import akka.dispatch.ExecutionContexts
 import akka.event.LoggingAdapter
 import akka.http.impl.engine.HttpConnectionIdleTimeoutBidi
-import akka.http.impl.engine.server.{ MasterServerTerminator, UpgradeToOtherProtocolResponseHeader }
+import akka.http.impl.engine.server.{ GracefulTerminatorStage, MasterServerTerminator, ServerTerminator, UpgradeToOtherProtocolResponseHeader }
 import akka.http.impl.util.LogByteStringTools
 import akka.http.scaladsl.Http.OutgoingConnection
 import akka.http.scaladsl.{ ConnectionContext, Http, HttpsConnectionContext }
@@ -26,7 +26,7 @@ import akka.stream.impl.io.TlsUtils
 import akka.stream.scaladsl.{ Flow, Keep, Sink, Source, TLS, TLSPlacebo, Tcp }
 import akka.stream.{ IgnoreComplete, Materializer }
 import akka.util.ByteString
-import akka.{ Done, NotUsed }
+import akka.Done
 import com.typesafe.config.Config
 
 import javax.net.ssl.SSLEngine
@@ -70,8 +70,12 @@ private[http] final class Http2Ext(private val config: Config)(implicit val syst
       else if (connectionContext.isSecure) settings.defaultHttpsPort
       else settings.defaultHttpPort
 
-    val http1 = Flow[HttpRequest].mapAsync(settings.pipeliningLimit)(handleUpgradeRequests(handler, settings, log)).join(http.serverLayer(settings, log = log))
-    val http2 = Http2Blueprint.handleWithStreamIdHeader(settings.http2Settings.maxConcurrentStreams)(handler)(system.dispatcher).join(Http2Blueprint.serverStackTls(settings, log, telemetry, Http().dateHeaderRendering))
+    val http1: HttpImplementation =
+      Flow[HttpRequest].mapAsync(settings.pipeliningLimit)(handleUpgradeRequests(handler, settings, log))
+        .joinMat(GracefulTerminatorStage(system, settings) atop http.serverLayer(settings, log = log))(Keep.right)
+    val http2: HttpImplementation =
+      Http2Blueprint.handleWithStreamIdHeader(settings.http2Settings.maxConcurrentStreams)(handler)(system.dispatcher)
+        .joinMat(Http2Blueprint.serverStackTls(settings, log, telemetry, Http().dateHeaderRendering))(Keep.right)
 
     val masterTerminator = new MasterServerTerminator(log)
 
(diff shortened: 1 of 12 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
```diff
@@ -158,12 +158,11 @@ private[http] object Http2Blueprint {
           FrameRenderer.render(GoAwayFrame(0, ex.errorCode))
         case ex: StreamTcpException => throw ex // TCP connection is probably broken: just forward exception
         case NonFatal(ex) =>
-          log.error(ex, s"HTTP2 connection failed with error [${ex.getMessage}]. Sending INTERNAL_ERROR and closing connection.")
+          log.debug(s"HTTP2 connection failed with error [${ex.getMessage}]. Sending INTERNAL_ERROR and closing connection.")
```
jrudolph, line 161: I have to look further into this. Not sure if the previous reporting was really useful.

jrudolph:
I'm still trying to get GracefulTerminationSpec also working with HTTP/2. I've still got a few errors to fix there.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2.scala
```diff
@@ -81,8 +85,13 @@ private[http] final class Http2Ext(private val config: Config)(implicit val syst
         incoming: Tcp.IncomingConnection =>
           try {
             httpPlusSwitching(http1, http2).addAttributes(prepareServerAttributes(settings, incoming))
-              .watchTermination()(Keep.right)
+              .watchTermination()(Keep.both)
               .join(HttpConnectionIdleTimeoutBidi(settings.idleTimeout, Some(incoming.remoteAddress)) join incoming.flow)
+              .mapMaterializedValue {
+                case (connectionTerminator, future) =>
+                  connectionTerminator.foreach(masterTerminator.registerConnection(_)(fm.executionContext))((fm.executionContext))
+                  future // drop the terminator matValue, we already registered is which is all we need to do here
```
raboof, line 93:
```suggestion
                  future // drop the terminator matValue, we already registered it which is all we need to do here
```

jrudolph:
There are lots of flaky tests but I don't think they are relatd.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/ProtocolSwitch.scala
```diff
@@ -4,28 +4,29 @@
...
+import akka.stream.stage.{ GraphStageLogic, GraphStageWithMaterializedValue, InHandler, OutHandler }
 
 import javax.net.ssl.SSLException
+import scala.concurrent.{ Future, Promise }
 
 /** INTERNAL API */
 @InternalApi
 private[http] object ProtocolSwitch {
-  type HttpServerFlow = Flow[SslTlsInbound, SslTlsOutbound, NotUsed]
+  type HttpServerFlow = Flow[SslTlsInbound, SslTlsOutbound, Future[ServerTerminator]]
```
raboof, line 22: (since it's now only used in only one place anymore, perhaps we should remove this `type` and just type it out below)

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Demux.scala
```diff
@@ -211,22 +211,45 @@ private[http2] abstract class Http2Demux(http2Settings: Http2CommonSettings, ini
...
+      override def terminate(deadline: FiniteDuration)(implicit ex: ExecutionContext): Future[Http.HttpTerminated] = {
+        terminateCallback.invoke(deadline)
+        terminationPromise.future
+      }
+      private def triggerTermination(deadline: FiniteDuration): Unit = {
+        // check if we are already terminating, otherwise start termination
+        if (!terminating) {
+          log.debug("Termination now starting...")
+          terminating = true
+          pushGOAWAY(ErrorCode.NO_ERROR, "")
```
raboof, line 246: I wonder if we should introduce a `debug` parameter to `triggerTermination`, might be helpful in debugging later... but perhaps not important for now.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Demux.scala
```diff
@@ -378,14 +408,19 @@ private[http2] abstract class Http2Demux(http2Settings: Http2CommonSettings, ini
...
+        val noMoreOutgoingStreams = (terminating || isClosed(substreamIn)) && activeStreamCount() == 0
         def allOutgoingDataFlushed = isClosed(frameOut) || multiplexer.hasFlushedAllData
         if (noMoreOutgoingStreams && allOutgoingDataFlushed) {
-          cancel(frameIn)
-          complete(frameOut)
-          // Using complete here (instead of a simpler `completeStage`) will make sure the buffer can be
-          // drained by the user before completely shutting down the stage finally.
-          bufferedSubStreamOutput.complete()
+          if (isServer)
+            completeStage()
```
raboof, line 415: so the server never has pending things in `bufferedSubStreamOutput` here? That's not obvious to me.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
raboof: Ok. Strange that this would change with this PR?

raboof (commented):
Very neat! One thing is not obvious to me (the `bufferedSubStreamOutput` thing)

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Demux.scala
jrudolph: You are right, that's probably a bit optimistic.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
jrudolph: Indeed, that's not needed yet. I had already added it to make GracfulTerminationSpec tests pass for HTTP/2. I've decided not to add those tests for HTTP/2 right now.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Demux.scala
jrudolph: I checked again. On the server side, it would mean that a user handler would still have a *request* buffered. However, in that case `activeStreamCount` would be > 0. 

This is different on the client side, where all network traffic is done but the user might not have read final responses.

raboof: Aah gotcha :+1: - might warrant a comment?

jrudolph: Yep, adding one.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Demux.scala
jrudolph: This seems like an implementation detail that is not really relevant for the other side, is it? After all, what we do here is close the connection because we want to which is a completely valid thing to do...

jrudolph: I now called it a "Voluntary connection close.". Whatever that will mean for the other side...

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
jrudolph: I was wrong. Without it Http2PersistentConnectionSpec will now fail.

jrudolph: Ok, found the culprit, I also added a cancellation propagation delay to the HTTP2 client side which now changed the error reporting behavior. I'm going to remove that change from this PR.

jrudolph:
@raboof do you want to add anything or can we merge?

raboof (approved):
Great!

Issue #2419, opened 2019-02-09
WebSocket stream not closed when there is a network disconnection (and while using keep-alive)


mbuccini:
Hello everyone,

We are experiencing an issue with akka-http 10.1.5 that doesn't terminate the websocket stream when there is a network disconnection.

Our server configuration is the following:
```
akka{
  http {
      server {
        idle-timeout = 60 seconds
        websocket{
          periodic-keep-alive-mode = ping
          periodic-keep-alive-max-idle = 30 seconds
        }
      }
      client {
        idle-timeout = infinite
      }
  }
}
```

The ping/pong mechanism seems to work as expected (and documented), meaning: I see in Wireshark the data being sent from the server every 30 seconds and the relative ACK from the client back. 

If I stop the client manually, the server (deployed on AWS) recognizes that there is a disconnection, and the websocket flow terminates from the server side.

If I drop the WIFI connection (where the client resides, while the server is deployed on AWS), the server does not receive any termination (or at least it's not propagated?). 

Did I misunderstand how the keep-alive mechanism is supposed to work? I could bet that some weeks/months ago this mechanism (with akka-http 10.1.1) used to work, however, I am not 100% sure about it - I just remember having tried it and it used to be OK, the automatic termination happened. 

I can't share any code, unfortunately, however, I might try to set up a little PoC, if needed.  

Thanks in advance.

jrudolph:
> If I drop the WIFI connection (where the client resides, while the server is deployed on AWS), the server does not receive any termination (or at least it's not propagated?).

This is basically the worst case for an idle TCP connection. All of the IP-level routing still works but one of the endpoint has gone away. This can only be detected once one of the sides sends another TCP packet which will eventually run into a timeout.

It seems we only implemented the sending of PINGs but do not react when there are no PONGs coming back.

jrudolph:
Also we should add ping/pong support to the client.

mbuccini:
Thanks for answering.

> Also we should add ping/pong support to the client.

I think that this is already included (see: https://doc.akka.io/docs/akka-http/current/configuration.html).

The issue is really when the connection drops suddenly without coming back and without giving the client the possibility to terminate the stream.

jrudolph:
> I think that this is already included (see: [doc.akka.io/docs/akka-http/current/configuration.html](https://doc.akka.io/docs/akka-http/current/configuration.html)).

Right, sorry, I missed where it was added.

mbuccini:
[...]

PR #4021, opened 2022-02-04
close websocket connection when idle configured timeout is reached 

kkonyshev:
References #2419

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/ws/WebSocket.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/ws/WebSocket.scala
@@ -41,6 +41,7 @@ private[http] object WebSocket {
     log:               LoggingAdapter): BidiFlow[FrameEvent, Message, Message, FrameEvent, NotUsed] =
     masking(serverSide, websocketSettings.randomFactory) atop
       FrameLogger.logFramesIfEnabled(websocketSettings.logFrames) atop
+      connectionIdeTimout(websocketSettings.connectionIdleTimout) atop
       frameHandling(serverSide, closeTimeout, log) atop
       periodicKeepAlive(websocketSettings) atop
       messageAPI(serverSide, closeTimeout)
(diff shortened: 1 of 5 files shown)


lightbend-cla-validator:

jrudolph (commented):
Thanks, @kkonyshev.

I wonder if that fixes anything, have you tested it? We already have idle-timeouts in place, do you know of a reason why they wouldn't apply?

kkonyshev:
No problem, @jrudolph.

Indeed, I tested it. 
The simplest way to verify the fix is to set 'periodic-keep-alive-max-idle' > 'connection-idle-timeout` (with enabled 'periodic-keep-alive-mode').

I didn't check why other idle-timeouts don't apply. Will have a look.

kkonyshev:
@jrudolph , existing idle timeouts apply to http2 configuration only.

kkonyshev:
@jrudolph , you are right, I will reuse the existing property and fix the PR. Thanks for the feedback.

lightbend-cla-validator:

lightbend-cla-validator:

lightbend-cla-validator:
At least one commit author (konstantin.konyshev@restore.energy, konstantin.konyshev@centrica.com) is not linked to a user. See https://help.github.com/en/articles/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user

lightbend-cla-validator:
At least one commit author (konyshev.konstantin@gmail.com) is not linked to a user. See https://help.github.com/en/articles/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user

plzakj (approved):

plzakj:
@jrudolph is it possible to proceed with this PR and also backport it to 10.1? We have verified that the changes also solve the issue #2419 on 10.1 in our case

jrudolph:
Hi @kkonyshev, we might have misunderstood before. It's not clear to me why the expected functionality has not already been covered by the existing connection-level idle timeout. 

Is this fix for client or server side?

As I said before we might have been missing to apply the idle timeout for the `websocketClient` functionality but for the server side it should have worked before (and it seems this is also what the original reporter at #2419 mentioned?).

It would help if we could start with a test case that shows the unexpected behavior (e.g. in WebSocketIntegrationSpec) so we can verify that something is actually fixed.

jrudolph:
Btw. 10.1.x is old by now and will only receive updates in rare cases.

kkonyshev:
@jrudolph, I believe I got your point last time and made changes accordingly. The Server-side didn't implement timeout handling, hence, the existing connection-level idle timeout wasn't applied at all. I added missing BidiFlow foon the server-side (it will use the exiting connection-level server timeout settings).
I will check why tests have failed.

plzakj:
@kkonyshev @jrudolph I was testing the changes on our side again and I was struggling a bit. I think I misunderstood as well. The changes from this PR apply bidirectionalIdleTimeout on the bidiflow stack which means that if there is e.g. ping being sent every 500ms and idle timeout set to 2s, then it will never timeout in the case pong responses are being lost.
Setting 'periodic-keep-alive-max-idle' > 'connection-idle-timeout` does not make sense to me.

In order to close the WS when pongs are being lost, I added a idleTimeout on the input flow (between masking.out and frameHandling.in). This causes timeout exception in the case there are no incoming frames during the timeout. Default is infinite. I'll run a separate pull request for this change.

@jrudolph I think the problem affects both server and client. When looking to HttpConnectionIdleTimeoutBidi - it applies bidirectionalIdleTimeout() which means that will not run into timeout as long as pings are being sent out?

johanandren:
Closing this due to inactivity, please re-open if you still want to contribute a fix for https://github.com/akka/akka-http/issues/2419 (under the new license)

PR #4024, opened 2022-02-08
build: try to report test results in GHA

jrudolph:


raboof (approved):
curious to see what this looks like ;)

jrudolph:
Doesn't work yet, since it doesn't manage to pick up any files so far...

jrudolph:
> Doesn't work yet, since it doesn't manage to pick up any files so far...

Ok, we didn't run any tests because ValidatePullRequest decided it's not necessary ;)

jrudolph:
The upload seems to work after all, there just weren't any test results to pick up before... Going ahead with merging, the report workflow will only work once merged.

jrudolph:
Here's the first result of the main build: https://github.com/akka/akka-http/runs/5108127753#r7

raboof:
> Here's the first result of the main build: https://github.com/akka/akka-http/runs/5108127753#r7

Cool! I noticed that you can find it from https://github.com/akka/akka-http/runs/5108014106?check_suite_focus=true by chopping off `?check_suite_focus=true` (not sure if there's a neater way).

It does look like the report gets truncated though ("Report exceeded GitHub limit of 65535 bytes and has been trimmed")

Issue #4017, opened 2022-02-02
Client-side SSE (Server-Sent Events) source does not emit server-side's default heartbeats


imRentable:
Hello, 

I am using Akka HTTP `v10.2.7` and trying to consume Server-Sent Events with an Akka HTTP client that are emitted by an Akka HTTP server similar to [this official example](https://doc.akka.io/docs/akka-http/10.0/sse-support.html). This works fine for events that have a non-empty string as value in their data field but not for events whose data field's value is an empty string. This happens to be the case when the default heartbeat provided by Akka HTTP is used: `akka.http.scaladsl.model.sse.ServerSentEvent#heartbeat`. The client source simply won't emit anything. Is this supposed to be like this? According to the examples given in the [specification](https://html.spec.whatwg.org/multipage/server-sent-events.html), a data field that has no value should trigger a client source to emit an event.

AFAIU, the cause for this behaviour lies within the `akka.http.scaladsl.unmarshalling.sse.ServerSentEventParser` in line 115 where data must not be empty. Otherwise, nothing is appended and a push won't happen in line 108.

There is also a comment stating that "Events without data are ignored according to the spec" but I have not been able to verify this within the aforementioned specification.

I wrote a minimal [client/server example](https://gist.github.com/imRentable/6be8ca9e8b2ee555090411bfef244257) that shows the aforementioned behaviour.

I am afraid that this could cause other client implementations to act weirdly since I think it does not work exactly as specified. That being said, please correct me if I am mistaken.

jrudolph:
Thanks for the report, @imRentable.

I'd say in section [9.2.6 Interpreting an event stream](https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation), the list of steps can be interpreted as skipping empty data events:

> 2. If the data buffer is an empty string, set the data buffer and the event type buffer to the empty string and return.

so that the processing never gets to actually dispatching the event.

Have you verified that browsers would show a different behavior?

imRentable:
@jrudolph: I agree, it is not very clear.

I tested it with Google Chrome `Version 97.0.4692.71 (Official Build) (x86_64)` and there I can see heartbeats in the form of `data:<empty string>`.

claasic:
There are three things I am reading differently.

1. The specification also lists an example with events that only consist of data fields with empty strings. These events are described as firing:
> The following stream fires two events:
> 
> data
> 
> data
> data
> 
> data:
[...]

PR #4043, opened 2022-02-24
http: hide empty SSE events behind flag

jrudolph:
Refs #4017

### DIFF
--- a/akka-http/src/main/resources/reference.conf
+++ b/akka-http/src/main/resources/reference.conf
@@ -55,5 +55,11 @@ akka.http {
 
     # The maximum size for parsing lines of a server-sent event.
     max-line-size = 4096
+
+    # Empty events are used by the akka-http SSE infrastructure for heartbeats. Received heartbeats are ignored by default
+    # and not delivered in the user stream.
+    #
+    # This setting can be enabled to pass those empty events to the application for explicitly handling.
+    emit-empty-events = off
   }
 }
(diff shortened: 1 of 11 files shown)


akka-http/src/main/resources/reference.conf
```diff
@@ -55,5 +55,11 @@ akka.http {
 
     # The maximum size for parsing lines of a server-sent event.
     max-line-size = 4096
+
+    # Empty events are used by the akka-http SSE infrastructure for heartbeats. Received heartbeats are ignored by default
+    # and not delivered in the user stream.
+    #
+    # This setting can be enabled to pass those empty events to the application for explicitly handling.
+    emit-empty-events = off
```
raboof, line 63: hmm, so if this is 'off' they are still emitted over the wire, just not passed to the application? That's not super clear from the name, but I'm not sure I have a better suggestion either.

akka-http/src/main/scala/akka/http/javadsl/settings/ServerSentEventSettings.scala
```diff
@@ -27,9 +27,15 @@ abstract class ServerSentEventSettings private[akka] () { self: ServerSentEventS
    */
   def maxLineSize: Int
 
+  /**
+   * Should events with empty data field be passed to the application.
```
raboof, line 31: :+1: to clarify that again here

raboof (approved):

akka-http/src/main/resources/reference.conf
jrudolph: Yeah, I hoped that `emit` was a stream term vs. network terms.

jrudolph: Yeah, I hoped that `emit` was a stream term vs. network terms.

jrudolph: Github was lagging...

PR #4044, opened 2022-02-24
http2: silence noisy error when HTTP/2 connection is closed with idle-timeout

### LINKED ISSUE #3900: HTTP idle-timeout encountered, no bytes passed in the last 60 seconds after updating to 10.2.6
After updating to 10.2.6 version I started seeing the following error in my application for every request
`HTTP2 connection failed with error [HTTP idle-timeout encountered, no bytes passed in the last 60 seconds. This is configurable by akka.http.[server|client].idle-timeout.]. Sending INTERNAL_ERROR and closing connection.`
I believe it comes from
this method https://github.com/akka/akka-http/blob/1088de3468cb3e488a8a4334226e8562cd3f10c5/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala#L161

I’m using the default value 60s for `akka.http.server.idle-timeout`. Is there a way to mute/handle this exception?

jrudolph:
Fixes #3900

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Blueprint.scala
@@ -7,7 +7,7 @@ package akka.http.impl.engine.http2
 import akka.NotUsed
 import akka.annotation.InternalApi
 import akka.event.LoggingAdapter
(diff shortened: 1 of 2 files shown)


raboof (approved):
yes, makes sense not to be loud about this.

PR #4047, opened 2022-03-01
 http2: unregister connection from MasterTermination when done

### LINKED ISSUE #4046: Memory leak with 10.2.8 when http2 is enabled
Can be reproduced by running `akka-http2-support/Test/runMain akka.http.scaladsl.Http2ServerTest` and accessing https://localhost:9000.

Seems to happen also if just HTTP/1.1 is used, so a likely culprit is the `ProtocolSwitch`.

jrudolph:
Fixes #4046

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2.scala
@@ -84,13 +84,15 @@ private[http] final class Http2Ext(implicit val system: ActorSystem)
         incoming: Tcp.IncomingConnection =>
           try {
             httpPlusSwitching(http1, http2).addAttributes(prepareServerAttributes(settings, incoming))
-              .watchTermination()(Keep.both)
-              .join(HttpConnectionIdleTimeoutBidi(settings.idleTimeout, Some(incoming.remoteAddress)) join incoming.flow)
-              .mapMaterializedValue {
-                case (connectionTerminator, future) =>
-                  connectionTerminator.foreach(masterTerminator.registerConnection(_)(fm.executionContext))((fm.executionContext))
+              .watchTermination() {
+                case (connectionTerminatorF, future) =>
+                  connectionTerminatorF.foreach { connectionTerminator =>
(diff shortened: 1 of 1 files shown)


jrudolph:
I tried to find a good test but I would only be able to assert on internals, i.e. on the amount of connections the MasterTerminator for a binding has registered but even that would require changes to the `ServerBinding` API.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2.scala
```diff
@@ -84,13 +84,15 @@ private[http] final class Http2Ext(implicit val system: ActorSystem)
...
-              .watchTermination()(Keep.both)
-              .join(HttpConnectionIdleTimeoutBidi(settings.idleTimeout, Some(incoming.remoteAddress)) join incoming.flow)
-              .mapMaterializedValue {
-                case (connectionTerminator, future) =>
-                  connectionTerminator.foreach(masterTerminator.registerConnection(_)(fm.executionContext))((fm.executionContext))
+              .watchTermination() {
+                case (connectionTerminatorF, future) =>
+                  connectionTerminatorF.foreach { connectionTerminator =>
+                    masterTerminator.registerConnection(connectionTerminator)(fm.executionContext)
+                    future.onComplete(_ => masterTerminator.removeConnection(connectionTerminator))(fm.executionContext)
```
raboof, line 91: `onComplete` :+1: 

Perhaps we should add a comment to `registerConnection`, but would remain easy to miss...

raboof (approved):

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2.scala
jrudolph: Or we could use WeakReferences instead, though, not sure if that's really necessary.

jrudolph: Or a protocol between MasterServerTerminator and ServerTerminator, so that a ServerTerminator knows about the MasterServerTerminator and could unregister itself on `postStop`.

PR #4051, opened 2022-03-01
build: add nightly job to matrix of configurations

jrudolph:
New configs are JDK 17 and Akka master


jrudolph:
Merging to see if it works

Issue #4058, opened 2022-03-03
Failed: GracefulTerminationSpec against Akka 2.6.x


jrudolph:
E.g. https://github.com/akka/akka-http/runs/5401287745?check_suite_focus=true

The problem seems to be that the last response before graceful termination closes the connection can be broken with Akka 2.6.x. The reason is probably that the `GracefulTerminatorStage` is using `failStage` to tear down a connection eventually. When that happens the error might overtake sending out the last response (can happen both for the automatic 503 response or the last regular response).

The reason is probably the write coalescing feature in akka-streams in 2.6.x. which makes it more likely that an error can overtake the latest outgoing bytes.

Part of a potential quick fix could be to use `completeStage` in `GracefulTerminatorStage`, instead. However, that does not work correctly because the stage still doesn't track response streams correctly, in which case, `completeStage` is not actually able to tear down the connection fully (in many cases that might not be an actual problem as the JVM is likely to be terminated afterwards anyway, however, in other cases users might rely on `binding.terminate` to return accurately that all connections have been terminated).

PR #4061, opened 2022-03-03
core: make GracefulTerminationSpec less strict

jrudolph:
Refs #4058

GracefulTerminationHandling has several race conditions that play out
differently in Akka 2.6. Most likely the reason is that write coalescing makes
it more likely that the error sent out when the termination timeout triggers
is able to overtake the sent out response.


akka-http-core/src/test/scala/akka/http/scaladsl/GracefulTerminationSpec.scala
```diff
@@ -171,6 +174,9 @@ class GracefulTerminationSpec
 
       ensureServerDeliveredRequest(r1) // we want the request to be in the server user's hands before we cause termination
       serverBinding.terminate(hardDeadline = time)
+      // avoid race condition between termination and sending out response
+      // FIXME: https://github.com/akka/akka-http/issues/4060
```
raboof, line 178: :+1:

raboof (approved):

PR #4069, opened 2022-03-07
Try to make GHA testing more reliable

jrudolph:
By reducing concurrency in sbt. Previous runs have shown that compilation and tests could still run in parallel otherwise.

(Not completely sure to me how that can happen, it seems that fine-grained `concurrentRestrictions` are broken under aggregation. Also `Test/compile test` doesn't seem to guarantee sequential execution. We could use `;Test/compile;test` instead but that still seems to show the problem with aggregation, i.e. even if the project is fully compiled, just running `test` in the root project still executes test suites from multiple submodules concurrently even with `Test / parallelExecution := false`.)

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -41,8 +41,11 @@ jobs:
           path: project/**/target
           key: build-target-${{ hashFiles('**/*.sbt', 'project/build.properties', 'project/**/*.scala') }}
 
+      - name: Compile everything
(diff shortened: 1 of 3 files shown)


raboof (approved):
> it seems that fine-grained `concurrentRestrictions` are broken under aggregation. Also `Test/compile test` doesn't seem to guarantee sequential execution

I think the `+~` might be messing this up somehow... :+1:

jrudolph:
> I think the `+~` might be messing this up somehow... +1

I think I even see that behavior just running `test` on the root project, so probably it is something else.

PR #4070, opened 2022-03-07
build: compile against right akka version in nightlies

jrudolph:


raboof (approved):
oops, missed that earlier

Issue #3069, opened 2020-04-16
Require Akka 2.7


raboof:
At some point we'd like to start requiring Akka 2.6.

For 10.2.0 we plan to do only #2886, so requiring Akka 2.6 would probably be for 10.3.0.

patriknw:
10.4.0 will require Akka 2.7

PR #4071, opened 2022-03-07
build: compile against right akka version in nightlies

jrudolph:
Can be fixed properly when moving to 2.6


docs/src/test/scala/docs/http/scaladsl/HttpClientExampleSpec.scala
```diff
@@ -13,7 +13,10 @@ import docs.CompileOnlySpec
 import org.scalatest.matchers.should.Matchers
 import org.scalatest.wordspec.AnyWordSpec
 
-@nowarn("msg=will not be a runnable program")
+@nowarn("msg=will not be a runnable program") //
+// OverflowStrategy.dropNew has been deprecated in latest Akka versions
+// FIXME: replace with 2.6 queue when 2.5 support is dropped
```
raboof, line 18: maybe tag with #3069 so we can easily search for it when we get there?

raboof (approved):

PR #4078, opened 2022-03-21
core: avoid loop in JavaQuery.render(HttpCharset, CharPredicate)

jrudolph:


akka-http-core/src/main/scala/akka/http/impl/model/JavaQuery.scala
```diff
@@ -33,5 +33,6 @@ final private[http] case class JavaQuery(query: sm.Uri.Query) extends jm.Query {
   override def withParam(key: String, value: String): jm.Query = jm.Query.create(query.map(_.asJava) :+ Pair(key, value): _*)
   override def render(charset: HttpCharset): String =
     UriRendering.renderQuery(new StringRendering, query, charset.nioCharset, CharacterClasses.unreserved).get
-  override def render(charset: HttpCharset, keep: CharPredicate): String = render(charset, keep)
```
hepin1989, line 36: So it was an infinite recursive call?

PR #4086, opened 2022-03-23
Update copyright year to 2022

raboof:


jrudolph (approved):
LGTM

Didn't we change the headers in some projects to use a version without the current year?

Issue #4087, opened 2022-03-23
Finished streams lingering in `HalfClosedRemoteSendingData` state leading to `REFUSED_STREAM`


fredfp:
We might be hitting an issue in akka-http (10.2.7 and 10.2.9) when using akka-grpc: on the server side, some requests seem to never fully finish and all new requests are being refused (`REFUSED_STREAM` because `activeStreamCount() > settings.maxConcurrentStreams`).

All we have is a heap dump from a time where the issue was likely. Analysing the heap dump, we notice that all streams in `Http2StreamHandling#streamStates` are in state `HalfClosedRemoteSendingData`, and their respective `outStream` seem to indicate that everything is finished (`buffer` is empty, `upstreamClosed` is true, `endStreamSent` is true, `trailer` is set, `maybeinlet` is null).

Looking at the code I would expect those streams to have transitioned to `Closed` state and removed from the `streamStates` map, they are not and I don't understand why.

We didn't manage to reproduce in isolation, and I doubt we can afford to run in debug mode for days until the issues pops up again. What would you advise we do to understand the issue better?

fredfp:
Is it possible to go through the following states on the server side: `Idle -> OpenReceivingDataFirst -> Open -> HalfClosedRemoteSendingData`? If yes, can it happen that we land in `HalfClosedRemoteSendingData` with `outStream` finishing at the same time but the proper transition to `Closed` doesn't happen?

jrudolph:
Thanks for the great report, @fredfp. I'll have a look.

fredfp:
My pleasure, thanks for having a look! As mentioned on gitter, I can't share the heap dump, but I'd be happy to do some screensharing or further data extraction if that could help. Just let me know.

jrudolph:
What kind of gRPC requests are involved in that case? Are they streaming or request/response style?

jrudolph:
This kind of call stack could be a problem where we the state machine is entered twice on the same call stack so that the inner state change is lost.

![image](https://user-images.githubusercontent.com/9868/159894270-2683a092-d08b-4943-8fed-c2869789ca9b.png)

fredfp:
We mostly have request/response style, and a single reqeust with streaming response. From the heap-dump I assumed the issue was with a request/response style given that `maybeinlet` had not been set (and is never cleared in the code).

jrudolph:
Thanks!

[...]

PR #4088, opened 2022-03-24
http2: defer enqueing out streams when state machine is already running

jrudolph:
To avoid reentering the state machine which will break setting intermediate states.

Refs #4087

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
@@ -149,7 +149,11 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper 
   private def updateState(streamId: Int, handle: StreamState => StreamState, event: String, eventArg: AnyRef = null): Unit =
     updateStateAndReturn(streamId, x => (handle(x), ()), event, eventArg)
 
+  private var stateMachineRunning = false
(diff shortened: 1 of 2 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -149,7 +149,21 @@ private[http2] trait Http2StreamHandling { self: GraphStageLogic with LogHelper
...
+  private var deferredStreamToEnqueue: Int = -1
+  private def enqueueOutStream(streamId: Int): Unit = if (stateMachineRunning) {
+    require(deferredStreamToEnqueue == -1, "Only one stream can be enqueued during a single state change")
+    deferredStreamToEnqueue = streamId
+  } else
+    multiplexer.enqueueOutStream(streamId)
+
+  private var stateMachineRunning = false
   private def updateStateAndReturn[R](streamId: Int, handle: StreamState => (StreamState, R), event: String, eventArg: AnyRef = null): R = {
+    require(!stateMachineRunning, "State machine already running")
```
fredfp, line 164: Is there a possibility that calling `tryPullSubStreams()` below leads to reentering `updateStateAndReturn` while `stateMachineRunning` is still `true`?

jrudolph: I don't think so, `tryPullSubStreams` only calls methods of the streams infrastructure which will not execute directly on the stack.

johanandren (approved):
LGTM

PR #4089, opened 2022-03-30
docs: log non-negative request duration time

sebastian-alfers:


jrudolph (approved):
LGTM, probably best we can do using this directive.

The best bet for getting accurate results would be to provide a custom Route directly which has most accurate control over when things are executed.

PR #4094, opened 2022-04-04
build: properly sort snapshots so we will find last one after "+9"

jrudolph:
Nightlies stopped building against the true last snapshot last week... (and there seems to be a small regression since then)

### DIFF
--- a/project/AkkaDependency.scala
+++ b/project/AkkaDependency.scala
@@ -7,6 +7,8 @@ package akka
 import sbt._
 import Keys._
 
+import scala.util.matching.Regex.Groups
+
 object AkkaDependency {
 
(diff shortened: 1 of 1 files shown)


jrudolph:
Haha, even the first nightly build against snapshots was already stuck on `+99`: https://github.com/akka/akka-http/runs/5498729572?check_suite_focus=true

jrudolph:
> Haha, even the first nightly build against snapshots was already stuck on `+99`: [akka/akka-http/runs/5498729572?check_suite_focus=true](https://github.com/akka/akka-http/runs/5498729572?check_suite_focus=true)

which funny enough was about right when I started working on building against master (but was still failing).

PR #4097, opened 2022-04-08
Compile akka http core

hughsimpson:
Enables and passes all tests in akka-http-core for scala 3, except for `If-Match dispatching` in `HttpHeaderSpec`, which throws a terrifying compiler error if left uncommented


lightbend-cla-validator:
Hi @hughsimpson,

akka-parsing/src/main/scala-3/akka/parboiled2/support/TailSwitch.scala
```diff
@@ -74,6 +74,15 @@ object TailSwitch {
   type Aux[L <: HList, LI <: HList, T <: HList, TI <: HList, R <: HList, RI <: HList, Out0] =
     TailSwitch[L, T, R] { type Out = Out0 }
 
-  implicit def tailSwitch[L <: HList, T <: HList, R <: HList]
-      : TailSwitch[L, T, R] { type Out = TailSwitch0[L, L, T, T, R, HNil] } = `n/a`
+  implicit def tailSwitch[L <: _ :: _, T <: _ :: _, R <: HList]
+  : TailSwitch[L, T, R] { type Out = TailSwitch0[L, L, T, T, R, HNil] } = `n/a`
+  /// Optimisations to reduce compilation times to something tolerable
```
hughsimpson, line 79: I couldn't figure out solve the general issue, but having some special-case implicits to short-circuit logic in certain common patterns seems to be enough for the usages in akka-http to compile

hughsimpson:
Have signed CLA

akka-http-core/src/main/scala/akka/http/impl/model/parser/SimpleHeaders.scala
```diff
@@ -100,14 +100,14 @@ private[parser] trait SimpleHeaders { this: Parser with CommonRules with CommonA
 
   // http://tools.ietf.org/html/rfc7233#section-4.2
   def `content-range` = rule {
-    (`byte-content-range` | `other-content-range`) ~ EOI ~> (`Content-Range`(_, _))
+    (`byte-content-range` ~ EOI ~> (`Content-Range`(_, _)) | `other-content-range` ~ EOI ~> (`Content-Range`(_, _)))
```
hughsimpson, line 103: The original form is thrown off by the different subtypes, but this doesn't feel too bad an expansion...

akka-http-core/src/main/scala/akka/http/impl/model/parser/UriParser.scala
```diff
@@ -259,7 +259,7 @@ private[http] final class UriParser(
     // has a max value-stack depth of 3
     def keyValuePairsWithLimitedStackUse: Rule1[Query] = rule {
       keyValuePair ~> { (key, value) => Query.Cons(key, value, Query.Empty) } ~ {
-        zeroOrMore('&' ~ keyValuePair ~> { (prefix: Query, key, value) => Query.Cons(key, value, prefix) }) ~>
+        zeroOrMore('&' ~ keyValuePair ~> { (prefix: Query.Cons, key, value) => Query.Cons(key, value, prefix) }) ~>
```
hughsimpson, line 262: Original form was thrown off by `Query` != `Query.Cons`

akka-http-core/src/test/scala/akka/testkit/AkkaSpec.scala
```diff
@@ -72,7 +75,7 @@ abstract class AkkaSpec(_system: ActorSystem)
 
   def this() = this(ActorSystem(AkkaSpec.getCallerName(getClass), AkkaSpec.testConf))
 
-  val log: LoggingAdapter = Logging(system, this.getClass)
+  val log: LoggingAdapter = Logging(system, this.getClass.asInstanceOf[Class[Any]])
```
hughsimpson, line 78: I don't really understand this one TBQH...

akka-parsing/src/main/scala-3/akka/parboiled2/support/TailSwitch.scala
prolativ: I've been trying to solve this in a more general way but it looks like this would require some fixes/optimizations in the compiler itself.

hughsimpson: Yeah, I had the same feeling... I poked around a little but it looks like the main issue is that the compiler is 'surprisingly slow' at evaluating the compile-time type reduction. Guessing the culprit could be the type destructuring, but didn't try to add a test-case to the compiler itself...

project/Dependencies.scala
```diff
@@ -27,7 +27,7 @@ object Dependencies {
 
   val scala212Version = "2.12.15"
   val scala213Version = "2.13.8"
-  val scala3Version = "3.1.1"
+  val scala3Version = "3.1.2"
```
hughsimpson, line 30: I see some improvements in compile times with this, although not nearly enough to offset the need for the `tailSwitchN` optimisatins

akka-parsing/src/main/scala-3/akka/parboiled2/support/TailSwitch.scala
hughsimpson: Anyway apologies for stepping on your toes with this one. I got rather carried away testing that the workaround held good for all the cases we encounter in the parsers 😅

jrudolph:
Wow, great that you tackled this, @hughsimpson! I'm going to have a thorough look later.

akka-parsing/src/main/scala-3/akka/parboiled2/support/TailSwitch.scala
```diff
@@ -74,6 +74,12 @@ object TailSwitch {
...
 
-  implicit def tailSwitch[L <: HList, T <: HList, R <: HList]
+  implicit def tailSwitch[L <: _ :: _, T <: _ :: _, R <: HList]
       : TailSwitch[L, T, R] { type Out = TailSwitch0[L, L, T, T, R, HNil] } = `n/a`
+  /// Optimisations to reduce compilation times to something tolerable
+  implicit def tailSwitch0: TailSwitch[HNil, HNil, HNil] { type Out = HNil } = `n/a`
+  implicit def tailSwitch1[T <: _ :: _]: TailSwitch[HNil, T, HNil] { type Out = HNil } = `n/a`
+  implicit def tailSwitch2[L <: _ :: _, R <: _ :: _]: TailSwitch[L, HNil, R] { type Out = Prepend0[L, R] } = `n/a`
+  implicit def tailSwitch3[L <: _ :: _]: TailSwitch[L, HNil, HNil] { type Out = L } = `n/a`
+  implicit def tailSwitch4[T <: HList, R <: _ :: _]: TailSwitch[HNil, T, R] { type Out = R } = `n/a`
```
jrudolph, line 84: Interesting that this case-by-case style is necessary. You would think that you could at least combine `tailSwitch1` and `tailSwitch4` but even that doesn't seem to work.

akka-http-core/src/test/scala/akka/testkit/AkkaSpec.scala
jrudolph: I tried variations but also failed to come up something better. It seems some automatic conversions wrt to Java wildcard types don't work as well (or at least different) in Scala 3 than before.

jrudolph (commented):
Looking great. The only thing that probably needs fixing is the error message assertions in tests to make sure they work for both Scala 2 and Scala 3. I'll have a look what's going on there.

jrudolph (approved):
LGTM, no blockers, if anything, it's the sorry shape of testing in the Scala 3 branch...

Thanks a lot, @hughsimpson, of getting this important milestone in place!

prolativ:
@jrudolph @hughsimpson it looks I've found the core reason of the problem with `TailSwitch` https://github.com/lampepfl/dotty/issues/14903

jrudolph:
> @jrudolph @hughsimpson it looks I've found the core reason of the problem with `TailSwitch` [lampepfl/dotty#14903](https://github.com/lampepfl/dotty/issues/14903)

Great work. While the current solution works for akka-http, there are probably other parboiled2 parsers where the workaround still wouldn't be enough, so good to make progress on a more general solution!

akka-http-core/src/test/scala/akka/http/impl/model/parser/HttpHeaderSpec.scala
```diff
@@ -160,7 +160,7 @@ class HttpHeaderSpec extends AnyFreeSpec with Matchers {
         Authorization(GenericHttpCredentials("NoTokenScheme", ""))
       "Authorization: QVFJQzV3TTJMWTRTZmN3Zk=" =!=
         ErrorInfo(
-          "Illegal HTTP header 'Authorization': Invalid input '=', expected auth-param, OWS, token68, 'EOI' or tchar (line 1, column 23)",
+          "Illegal HTTP header 'Authorization': Invalid input '=', expected tchar, OWS, auth-param, token68 or 'EOI' (line 1, column 23)",
```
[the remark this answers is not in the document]
hughsimpson: Yeah it's a funny one. I have been running tests against both 2.13.x and 3.1.x throughout so just figured 'meh. Something has changed but it's still correct:

akka-parsing/src/main/scala-3/akka/parboiled2/support/TailSwitch.scala
hughsimpson: Yeah that really threw me off. This should be three rules, I think, not five, but hey. It works.

PR #4098, opened 2022-04-10
core: avoid array copy for string entities

brharrington:
When creating an `HttpEntity` from a `String`, use
`ByteString.fromArrayUnsafe` to avoid creating an
unnecessary copy of the byte array.


raboof (approved):
`string.getBytes` [promises](https://docs.oracle.com/javase/7/docs/api/java/lang/String.html#getBytes(java.nio.charset.Charset)) it always returns a new byte array, so this should be :+1: AFAICS

jrudolph (approved):
LGTM, good idea.

PR #4093, opened 2022-04-04
cache: avoid locking cache during directive processing

jrudolph:
Threads competing to access the cache may block like this:

```
   java.lang.Thread.State: BLOCKED (on object monitor)
	at java.util.concurrent.ConcurrentHashMap.computeIfPresent(ConcurrentHashMap.java:1760)
	- waiting to lock <0x000000073cb6b320> (a java.util.concurrent.ConcurrentHashMap$Node)
	at com.github.benmanes.caffeine.cache.BoundedLocalCache.replace(BoundedLocalCache.java:2302)
	at com.github.benmanes.caffeine.cache.LocalAsyncCache.lambda$handleCompletion$7(LocalAsyncCache.java:210)
	at com.github.benmanes.caffeine.cache.LocalAsyncCache$$Lambda$105857/1233505965.accept(Unknown Source)
	at java.util.concurrent.CompletableFuture.uniWhenComplete(CompletableFuture.java:774)
	at java.util.concurrent.CompletableFuture$UniWhenComplete.tryFire(CompletableFuture.java:750)
	at java.util.concurrent.CompletableFuture$Completion.exec(CompletableFuture.java:457)
	at java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:289)
	at java.util.concurrent.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1056)
	at java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1692)
	at java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:175)
```

when another thread is updating the same cache node synchronously. We have an async cache
but we actually have to use it asynchronously. This is not the default for
directive processing which uses FastFuture by default for faster but synchronous processing.

Fixes #4092

Issue #4092, opened 2022-04-04
CachingDirectives may lock on cache if directive processing is slow


jrudolph:
If a route wrapped inside of a cache executes its directive processing slowly and synchronously (which is the default due to `FastFuture` behavior, the cache itself might lock up given enough concurrency.

This is because even using the async cache you have to run the compute function quickly and return a future as fast as possible. Our design will however run the directive synchronously and only run the route asynchronously. If you run some heavy work directly in directives (which can easily happen due to the nature of route construction in akka-http), you might not only block in those threads but also lock the cache itself (given enough concurrent accesses). You could say that running heavy work synchronously is a user bug in itself. However, even in that case the available concurrency should not restricted by the cache setup (number of nodes inside of the backing hash map) but by the amount of available CPU resources.

I found this in a real world scenario.

One thread is locking a node:

```
	at java.lang.Thread.sleep(Native Method)
    [...]
	at com.github.benmanes.caffeine.cache.LocalAsyncCache.lambda$get$2(LocalAsyncCache.java:94)
	at com.github.benmanes.caffeine.cache.LocalAsyncCache$$Lambda$105379/672290125.apply(Unknown Source)
	at com.github.benmanes.caffeine.cache.BoundedLocalCache.lambda$doComputeIfAbsent$14(BoundedLocalCache.java:2406)
	at com.github.benmanes.caffeine.cache.BoundedLocalCache$$Lambda$105385/540768935.apply(Unknown Source)
	at java.util.concurrent.ConcurrentHashMap.compute(ConcurrentHashMap.java:1892)
	- locked <0x000000073cb6b320> (a java.util.concurrent.ConcurrentHashMap$Node)
	at com.github.benmanes.caffeine.cache.BoundedLocalCache.doComputeIfAbsent(BoundedLocalCache.java:2404)
	at com.github.benmanes.caffeine.cache.BoundedLocalCache.computeIfAbsent(BoundedLocalCache.java:2387)
	at com.github.benmanes.caffeine.cache.LocalAsyncCache.get(LocalAsyncCache.java:93)
	at com.github.benmanes.caffeine.cache.LocalAsyncCache.get(LocalAsyncCache.java:84)
	at akka.http.caching.LfuCache.apply(LfuCache.scala:106)
	at akka.http.scaladsl.server.directives.CachingDirectives.$anonfun$alwaysCache$2(CachingDirectives.scala:53)
```

Another (multiple other) thread is blocked on accessing the same node:

```
[...]

PR #4099, opened 2022-04-11
caching: fix cached execution and test

jrudolph:
The `asyncRoute & ...` construction did not work as expected and also using
`check` in tests starved threads themselves.

Refs #4092, update to #4093


akka-http-caching/src/main/scala/akka/http/scaladsl/server/directives/CachingDirectives.scala
```diff
@@ -49,13 +49,13 @@ trait CachingDirectives {
     // Do directive processing asynchronously to avoid locking the cache accidentally (#4092)
     // This will be slightly slower, but the rational here is that caching is used for slower kind of processing
     // anyway so the performance hit should be acceptable.
-    CachingDirectives.asyncRoute &
```
jrudolph, line 52: Might have worked reversing the two directives. This seems simpler. Interesting that it only failed for Akka 2.6 and still not at all times... I reproduced it using Akka 2.6.18 and using `taskset -c 0` to limit CPU resources.

PR #4113, opened 2022-04-25
Scala 3: edit sources on-the-fly in sbt to support `@pre213` methods in shared source code

jrudolph:
(But hopefully get soon rid of it when we drop Scala 2.12 support)

~Only applied to akka-http-core so far.~

### DIFF
--- a/project/Pre213Preprocessor.scala
+++ b/project/Pre213Preprocessor.scala
@@ -1,0 +1,36 @@
+package akka.http.sbt
+
+import sbt._
+import Keys._
+
+object Pre213Preprocessor extends AutoPlugin {
+  val pre213Files = settingKey[Seq[String]]("files that should be edited for pre213 annotation")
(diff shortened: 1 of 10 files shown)


project/Pre213Preprocessor.scala
```diff
@@ -0,0 +1,36 @@
+package akka.http.sbt
+
+import sbt._
+import Keys._
+
+object Pre213Preprocessor extends AutoPlugin {
+  val pre213Files = settingKey[Seq[String]]("files that should be edited for pre213 annotation")
+
+  val pattern = """(?s)@pre213.*?@since213""".r
```
jrudolph, line 9: Yes, we've come a far way from (only) running regex on URLs...

jrudolph:
Let's merge this even if this pattern will ultimately be removed for 10.3.x.

PR #4079, opened 2022-03-21
Ongoing Scala 3 migration effort

jrudolph:
Projects can already use Akka HTTP with Scala 3 by using [CrossVersion.for3Use2_13](https://scala-lang.org/blog/2021/04/08/scala-3-in-sbt.html#using-scala-213-libraries-in-scala-3), but for libraries that is not a good solution. This PR is about providing 'native' Scala 3 artifacts for Akka HTTP.

Commits are supposed to stay, so let's try to use reasonable commits. Changes can be PR'd against this branch.

Here are some guidelines about what we want to achieve:

 * Scala 3 support for all modules
 * No significant code forks (version-dependent directories) of general purpose code. We can maintain some forks (e.g. in parsers if not possible otherwise) but in general we cannot maintain significant forks for Scala 3 compatibility. I.e. for now we need to find solutions that work for all of Scala 2.12, 2.13, and 3. If Scala 2.12 support is too big a burden, we might consider targeting Akka HTTP 10.3.0 where we might want to drop Scala 2.12 support (and Akka 2.5).
 * Only Akka 2.6 is supported, so I changed the default Akka version. Since we still build against Akka 2.5 by default we will have to find a solution for publishing the Scala 3 version against 2.6.
 * Please add a comment when working on a task item to avoid stepping on each other's toes.

Tasks:
 * [x] Port parboiled2 to our private fork
 * [x] akka-http-core
   * [x] Make main sources compile without parsers
   * [x] Gradually make parsers compile for Scala 3 #4097
   * [x] Incrementally move tests back from `scala-2` to `scala` directory, fixing issues in the process #4097
   * [x] Setup Scala 3 compiler plugin to elide methods marked with `@pre213` annotation (or wait until Scala 2.12 is dropped) #4113
 * [x] akka-http2-support (only has tests)
 * [x] akka-http
   * [x] Enable `akka-http` module for Scala 3 (currently disabled in Scala 3) and make main code compile @jchyb
 * [x] akka-http-testkit @jrudolph #4101
 * [x] akka-http-tests #4107
   * [x] Make tests compile #4107
   * [x] Make tests pass #4107
 * [x] Enable akka-caching @jrudolph #4101
 * [x] Enable akka-http-spray-json #4107
 * [x] Enable akka-http-scala-xml #4107
 * [x] Enable akka-http-jackson #4126
 * [ ] akka-http-bench-jmh
	 * [ ] Run some benchmarks to find most obvious issues
 * [x] docs module #4126
[...]

PR #4114, opened 2022-04-25
Scala 3: Remove 3.0-migration compiler option from all modules

jchyb:
References #4079, task `Make compilation work without 3.0-migration compiler option`

### DIFF
--- a/akka-http-core/src/main/java/akka/http/javadsl/model/Query.java
+++ b/akka-http-core/src/main/java/akka/http/javadsl/model/Query.java
@@ -7,6 +7,7 @@ package akka.http.javadsl.model;
 import akka.http.impl.model.JavaQuery;
 import akka.http.impl.model.UriJavaAccessor;
 import akka.http.scaladsl.model.*;
+import akka.http.javadsl.model.HttpCharset;
 import akka.japi.Pair;
 import akka.parboiled2.CharPredicate;
 import akka.parboiled2.ParserInput$;
--- a/akka-http-core/src/main/scala/akka/http/javadsl/Http.scala
+++ b/akka-http-core/src/main/scala/akka/http/javadsl/Http.scala
@@ -33,7 +33,7 @@ import akka.stream.scaladsl.Keep
 object Http extends ExtensionId[Http] with ExtensionIdProvider {
   override def get(system: ActorSystem): Http = super.get(system)
   override def get(system: ClassicActorSystemProvider): Http = super.get(system)
-  def lookup() = Http
+  def lookup = Http
   def createExtension(system: ExtendedActorSystem): Http = new Http(system)
 }
 
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/settings/ServerSettings.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/settings/ServerSettings.scala
@@ -59,32 +59,32 @@ abstract class ServerSettings private[akka] () extends akka.http.javadsl.setting
 
   /* Java APIs */
 
-  override def getBacklog = backlog
-  override def getPreviewServerSettings: akka.http.javadsl.settings.PreviewServerSettings = previewServerSettings
-  override def getDefaultHostHeader = defaultHostHeader.asJava
-  override def getPipeliningLimit = pipeliningLimit
-  override def getParserSettings: js.ParserSettings = parserSettings
-  override def getMaxConnections = maxConnections
-  override def getTransparentHeadRequests = transparentHeadRequests
-  override def getResponseHeaderSizeHint = responseHeaderSizeHint
-  override def getVerboseErrorMessages = verboseErrorMessages
-  override def getSocketOptions = socketOptions.asJava
-  override def getServerHeader = OptionConverters.toJava(serverHeader.map(_.asJava))
-  override def getTimeouts = timeouts
-  override def getRawRequestUriHeader = rawRequestUriHeader
-  override def getRemoteAddressHeader = remoteAddressHeader
-  override def getRemoteAddressAttribute: Boolean = remoteAddressAttribute
-  override def getLogUnencryptedNetworkBytes = OptionConverters.toJava(logUnencryptedNetworkBytes)
+  override def getBacklog = this.backlog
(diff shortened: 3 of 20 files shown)


akka-http-core/src/main/java/akka/http/javadsl/model/Query.java
```diff
@@ -7,6 +7,7 @@
 import akka.http.impl.model.JavaQuery;
 import akka.http.impl.model.UriJavaAccessor;
 import akka.http.scaladsl.model.*;
+import akka.http.javadsl.model.HttpCharset;
```
jchyb, line 10: This is interesting - before adding this import Scala 3 incorrectly resolved javadsl.model.HttpCharset as scaladsl.model.HttpCharset, causing problems in classes that inherit Query.

akka-http-core/src/test/scala/akka/http/impl/engine/client/HostConnectionPoolSpec.scala
```diff
@@ -542,7 +542,7 @@ class HostConnectionPoolSpec extends AkkaSpecWithMaterializer(
         lazy val requestIn = TestPublisher.probe[RequestContext]()
         lazy val responseOut = TestSubscriber.probe[ResponseContext]()
 
-        protected val server: Flow[HttpRequest, HttpResponse, Future[Http.OutgoingConnection]]
+        protected def server: Flow[HttpRequest, HttpResponse, Future[Http.OutgoingConnection]]
```
jchyb, line 545: In Scala 2 lazy val cannot be in an abstract class, and in Scala 3 lazy val cannot override regular val. No problems with overriding def in either, it seems

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/FutureDirectivesSpec.scala
```diff
@@ -123,7 +123,7 @@ class FutureDirectivesSpec extends RoutingSpec with Inside with TestKitBase {
       occurrences = 1,
       message = BasicRouteSpecs.defaultExnHandler500Error("XXX")
     ).intercept {
-        Get() ~> onSuccess(Future.failed(TestException)) { echoComplete } ~> check {
+        Get() ~> onSuccess(Future.failed[Int](TestException)) { echoComplete } ~> check {
```
jchyb, line 126: I recall here being some issues with implicit value not being found for onSuccess if a type parameter was not specified (unsure why). I was hoping to use Unit everywhere (as that seemed natural with the exception), but this specific one demanded actual value (again don't remember what the error message was, I'll try to reproduce it and comment)

akka-http-core/src/main/scala/akka/http/scaladsl/settings/ServerSettings.scala
```diff
@@ -59,32 +59,32 @@ abstract class ServerSettings private[akka] () extends akka.http.javadsl.setting
...
-  override def getResponseHeaderSizeHint = responseHeaderSizeHint
-  override def getVerboseErrorMessages = verboseErrorMessages
-  override def getSocketOptions = socketOptions.asJava
-  override def getServerHeader = OptionConverters.toJava(serverHeader.map(_.asJava))
-  override def getTimeouts = timeouts
-  override def getRawRequestUriHeader = rawRequestUriHeader
-  override def getRemoteAddressHeader = remoteAddressHeader
-  override def getRemoteAddressAttribute: Boolean = remoteAddressAttribute
-  override def getLogUnencryptedNetworkBytes = OptionConverters.toJava(logUnencryptedNetworkBytes)
+  override def getBacklog = this.backlog
```
jrudolph, line 62: So annoying, thanks for fixing :)

akka-http-core/src/main/java/akka/http/javadsl/model/Query.java
jrudolph: That sounds like a bug in the Java support of Scala 3, doesn't it? Wouldn't it be a problem if javac and scalac disagree about how names are resolved in Java source files?

akka-http-core/src/main/scala/akka/http/javadsl/Http.scala
```diff
@@ -33,7 +33,7 @@ import akka.stream.scaladsl.Keep
 object Http extends ExtensionId[Http] with ExtensionIdProvider {
   override def get(system: ActorSystem): Http = super.get(system)
   override def get(system: ClassicActorSystemProvider): Http = super.get(system)
-  def lookup() = Http
+  def lookup = Http
```
jrudolph, line 36: Technically this method is public, so maybe Mima will complain. Or maybe it shouldn't because it's an implementation and the final shape has already been determined by the trait it implements even in Scala 2?

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/FutureDirectivesSpec.scala
jrudolph: `Unit` doesn't work because the `Tupler` machinery then expects a nullary function instead of `echoComplete`. So, that's fine and as expected.

Having to the type annotation here is also fine, in more realistic use cases the type will be inferred anyway.

jrudolph (commented):
This is great! Thanks, @jchyb. The only thing which might make sense is moving the compat things to `akka.http.ccompat`.

akka-http-core/src/main/java/akka/http/javadsl/model/Query.java
jchyb: I agree, It also seems like the root of the previous problems with scaladsl and javadsl type inference. Before this I couldn't find a minimization for that issue, but now it seems clear. I'll try making one and file an issue (hopefully also today).

akka-http-core/src/main/scala/akka/http/javadsl/Http.scala
jrudolph: Mima doesn't complain, so :+1:

jrudolph:
Nice, thanks!

PR #4117, opened 2022-04-26
scala 3: enable cross building for 2.12 again

jrudolph:
To verify that we have nothing broken so far.


jrudolph:
Only works after #4113, so adding it there instead.

Issue #4120, opened 2022-05-09
Allow a custom caffeine cache loader


Marcus-Rosti:
I'd love to be able to plugin a specific caffeine implementation rather than just the [default](https://github.com/akka/akka-http/blob/main/akka-http-caching/src/main/scala/akka/http/caching/LfuCache.scala#L61-L67). For instance with an oauth token that requires being "refreshed" via an http request. I'd love to be able to use https://github.com/ben-manes/caffeine/wiki/Refresh to do it

jrudolph:
I think that would be useful. Could it be as simple as a function which creates an akka-http `Cache[T]` from an arbitrary caffeine `AsyncCache[T]`?

PR #4121, opened 2022-05-09
Implements a LoadingCache

Marcus-Rosti:
References #4120

Implements an AsyncLoadingCache for akka-http-caching. This cache allows loaded values to be asynchronously reloaded after a given interval. Should the value not be read for a given interval the cache will also evict that value.


akka-http-caching/src/main/scala/akka/http/caching/RefreshingCache.scala
```diff
@@ -0,0 +1,130 @@
...
+  def toJavaMappingFunction[K, V](genValue: () => Future[V]): BiFunction[K, Executor, CompletableFuture[V]] =
+    asJavaBiFunction[K, Executor, CompletableFuture[V]]((k, e) => genValue().toJava.toCompletableFuture)
+
+  def toJavaMappingFunction[K, V](loadValue: K => Future[V]): BiFunction[K, Executor, CompletableFuture[V]] =
+    asJavaBiFunction[K, Executor, CompletableFuture[V]]((k, e) => loadValue(k).toJava.toCompletableFuture)
+}
+
+/** INTERNAL API */
+@InternalApi
+private[caching] class RefreshingCache[K, V](val store: AsyncLoadingCache[K, V]) extends Cache[K, V] {
```
jrudolph, line 97: Is it even necessary to duplicate most of the implementation here? What kind of API changes are needed for this feature?

Marcus-Rosti: I think the big thing is that AsyncLoadingCache will require the caller to pass a loading function in on construction otherwise it will have some dumb default:
```scala

    implicit val dummyLoader: AsyncCacheLoader[K, V] =
      (_: K, _: Executor) => Future.failed(new NoSuchElementException("")).toJava.toCompletableFuture
```

So it requires an AsyncCacheLoader to be passed in on construction to the store, if you're okay with that api change here?

Marcus-Rosti: but yeah otherwise, I think the C can actually stay the same and just be common between the too.

so all of
```scala
private[caching] class LfuCache[K, V](val store: AsyncCache[K, V]) extends Cache[K, V] {
```
could probably just be
```scala
private[caching] class CacheImpl[K, V](val store: AsyncCache[K, V]) extends Cache[K, V] {
```

and then AsyncCacheLoader adheres to AsyncCache in the caffeine code other than a few default function changes

akka-http-caching/src/test/scala/akka/http/caching/ExpiringRefreshingCacheSpec.scala
```diff
@@ -0,0 +1,105 @@
...
+import org.scalatest.matchers.should.Matchers
+import org.scalatest.wordspec.AnyWordSpec
+
+class ExpiringRefreshingCacheSpec extends AnyWordSpec with Matchers with BeforeAndAfterAll {
+  implicit val system: ActorSystem = ActorSystem()
+  import system.dispatcher
+
+  "A RefreshingCache" should {
+    "refresh values after the refresh-after-write" in {
+      val wait = 1.second
```
johanandren, line 26: 4s for 4 test cases is a lot in the whole test suite.

Would be good to tune this down quite a bit and share a single wait time for all tests and the config rather than repeat. I think something like 200ms should be fine.

akka-http-caching/src/main/scala/akka/http/caching/impl/LoadingCacheImpl.scala
```diff
@@ -0,0 +1,37 @@
...
+  override def load(key: K): Future[V] = futureToScala(store.get(key))
+
+  override def loadAll(keys: Set[K]): Future[Map[K, V]] = {
+
+    val f: CompletableFuture[util.Map[K, V]] = store.getAll(keys.asJava)
+
+    val convert: util.Map[K, V] => Map[K, V] = _.asScala.toMap
+
+    val g: CompletableFuture[Map[K, V]] = f.thenApply(asJavaFunction(convert))
+    futureToScala(g)
```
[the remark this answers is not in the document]
johanandren: You may have to add explicit types/type hints if the inference doesn't work out, but I think this is fine.

akka-http-caching/src/main/scala/akka/http/caching/javadsl/RefreshingCacheSettings.scala
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Public API but not intended for subclassing
+ */
+@DoNotInherit
+abstract class RefreshingCacheSettings private[http] () { self: RefreshingCacheSettingsImpl =>
+
+  /* JAVA APIs */
+  def getMaxCapacity: Int
+  def getRefreshAfterWrite: Duration
+  def getExpireAfterWrite: Duration
```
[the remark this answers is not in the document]
Marcus-Rosti: @johanandren are you sure about this? The api for LfuCachineSettings accepts and returns scala Durations?

Marcus-Rosti:
@johanandren thanks so much for the review!

akka-http-caching/src/main/scala/akka/http/caching/javadsl/LoadingCacheSettings.scala
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Public API but not intended for subclassing
+ */
+@DoNotInherit
+abstract class LoadingCacheSettings private[http] () { self: LoadingCacheSettingsImpl =>
+
+  /* JAVA APIs */
+  def getMaxCapacity: Int
+  def getRefreshAfterWrite: Duration
+  def getExpireAfterWrite: Duration
```
johanandren, line 23: For Java should return and accept `java.time.Duration` rather than `scala.concurrent.duration.Duration` (but for Scala we want the Scala duration ofc)

johanandren: Ah, sorry, missed your comment, yes, I'm sure, that must have been an oversight for the original APIs (if you could add there as well it would be great)

Marcus-Rosti: cool, I can fix that! But would that be a breaking api change for people to transition?

johanandren: Good point, we have the APIs marked with `ApiMayChange` which signals that we do not guarantee compatibility quite yet, but it's still not super nice to break things.

 I think constructing Scala Duration from Java is not very pleasant/obvious either so maybe the risk is smallish and Java devs may accept it since it improves their lives.

Marcus-Rosti: 👍  alright I'll fix it!

Marcus-Rosti:
@johanandren hey I don't think I'll be updating this further as I won't be able to get any benefit from it in my code base. We can no longer upgrade akka versions

He-Pin:
> @johanandren hey I don't think I'll be updating this further as I won't be able to get any benefit from it in my code base. We can no longer upgrade akka versions

@Marcus-Rosti I think you can ,just not as soon as a new version is released.

johanandren:
Sorry to hear that but I understand, let's close this PR for now then.

PR #4123, opened 2022-05-13
=sbt Update Scala 3 version to 3.1.3

He-Pin:
https://github.com/lampepfl/dotty/releases/tag/3.1.3-RC3
More aggressive reduction of type selection (fixes parboiled2) https://github.com/lampepfl/dotty/pull/14987
Relax avoidance checks more for match type reduction https://github.com/lampepfl/dotty/pull/15036

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -27,7 +27,7 @@ object Dependencies {
 
   val scala212Version = "2.12.15"
   val scala213Version = "2.13.8"
-  val scala3Version = "3.1.2"
+  val scala3Version = "3.1.3"
   val allScalaVersions =
     // FIXME: can be simplified when Akka 2.5 is dropped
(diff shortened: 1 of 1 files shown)


raboof (commented):
Even though our Scala 3 support is still experimental, I don't think we should use -RC versions if we can help it, since I think you can use libraries compiled with 3.1.2 in a project using 3.1.3-RC3, but possibly not vice-versa (?).

jrudolph:
> Even though our Scala 3 support is still experimental, I don't think we should use -RC versions if we can help it, since I think you can use libraries compiled with 3.1.2 in a project using 3.1.3-RC3, but possibly not vice-versa (?).

Yep, good point, let's keep here until the final. Still good to have this PR testing against the latest version, as the previous versions of 3.1.3-RCs had the regression that broke everything...

PR #4126, opened 2022-05-27
scala 3: compile and test marshalling, JMH and docs subprojects

raboof:
On top of #4079

Three errors remaining around creating case classes from parameters

### DIFF
--- a/docs/src/test/scala/docs/http/scaladsl/HttpServerWithActorInteraction.scala
+++ b/docs/src/test/scala/docs/http/scaladsl/HttpServerWithActorInteraction.scala
@@ -70,7 +70,7 @@ object HttpServerWithActorInteraction {
             implicit val timeout: Timeout = 5.seconds
 
             // query the actor for the current auction state
-            val bids: Future[Bids] = (auction ? (replyTo => GetBids(replyTo))).mapTo[Bids]
+            val bids: Future[Bids] = auction.ask(GetBids(_))
             complete(bids)
           }
         )
--- a/docs/src/test/scala/docs/http/scaladsl/HttpServerWithActorInteraction.scala
+++ b/docs/src/test/scala/docs/http/scaladsl/HttpServerWithActorInteraction.scala
@@ -13,6 +13,7 @@ import akka.http.scaladsl.model.StatusCodes
 import akka.http.scaladsl.server.Directives._
 import akka.util.Timeout
 import spray.json.DefaultJsonProtocol._
+import spray.json.RootJsonFormat
 
 import scala.concurrent.duration._
 import scala.concurrent.{ ExecutionContext, Future }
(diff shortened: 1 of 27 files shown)


project/Common.scala
```diff
@@ -16,19 +16,15 @@ object Common extends AutoPlugin {
...
+      "-release", "8",
       "-unchecked",
-      "-Xlint",
       "-Ywarn-dead-code",
       // Silence deprecation notices for changes introduced in Scala 2.12
       // Can be removed when we drop support for Scala 2.12:
       "-Wconf:msg=object JavaConverters in package collection is deprecated:s",
       "-Wconf:msg=is deprecated \\(since 2\\.13\\.:s",
     ),
-    // '-release' parameter is restricted to 'Compile, compile' scope because
```
raboof, line 28: (tested, no longer seems needed)

docs/src/test/scala/docs/http/scaladsl/HttpServerWithActorInteraction.scala
```diff
@@ -69,7 +70,7 @@ object HttpServerWithActorInteraction {
             implicit val timeout: Timeout = 5.seconds
 
             // query the actor for the current auction state
-            val bids: Future[Bids] = (auction ? GetBids).mapTo[Bids]
+            val bids: Future[Bids] = (auction ? (replyTo => GetBids(replyTo))).mapTo[Bids]
```
raboof, line 73: Hmm, a bit unfortunate

project/Common.scala
jrudolph: Cool, thanks!

docs/src/test/scala/docs/http/scaladsl/server/directives/MiscDirectivesExamplesSpec.scala
```diff
@@ -143,7 +144,7 @@ class MiscDirectivesExamplesSpec extends RoutingSpec with CompileOnlySpec {
 
     // tests:
     def entityOfSize(size: Int) =
-      HttpEntity(ContentTypes.`text/plain(UTF-8)`, "0" * size)
+      HttpEntity(ContentTypes.`text/plain(UTF-8)`, List.fill(size)('0').mkString)
```
eshu, line 147: It does not look more readable...

docs/src/test/scala/docs/http/scaladsl/HttpServerWithActorInteraction.scala
johanandren: With a typed actor I'd expected being able to do `val bids: Future[Bids] = action ? (GetBids(_))` (or `ask` instead of `?` which makes the param list parenthesis's more obvious), without needing the `mapTo` at all.

johanandren (approved):

PR #4129, opened 2022-06-03
docs: replace abandoned swakka with actively maintained swagger-akka-http

leviramsey:
Swakka appears to be abandoned (no maintainer activity in 3 years) and there's an [outstanding issue around 10.2.x support](https://github.com/jtownson/swakka/issues/9).  Additionally, it's not built for Scala 2.13.

This PR proposes to recommend `swagger-akka-http` in its place, which is more actively maintained.  Alternatively, we can decide to not endorse that extension, in which case I think there's an even weaker case for endorsing swakka.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ target
 .*.swp
 *.vim
 .ensime*
+.bsp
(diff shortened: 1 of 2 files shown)


jrudolph (approved):
LGTM, good point. `swagger-akka-http` looks useful enough to add in here.

Issue #4128, opened 2022-06-02
Can not recover from `java.net.ConnectException: Connection refused`


sebastian-alfers:
This ticket is a break-out bug report / investigation based on https://github.com/akka/akka-http/issues/4127.

---

There are three scenarios that behave differently:
- 1: making one `Http().singleRequest`
- 2: making 10 `Http().singleRequest`
- 3: making 50`Http().singleRequest`

Underlying config:

```
akka {
  http {
    host-connection-pool {
      max-connections = 32
      max-open-requests = 16
    }
    client {
      idle-timeout = 1s
      connecting-timeout = 1s
    }
  }
}

```

---

## 1:
Doing this:
```
  implicit val actorSystem: ActorSystem = ActorSystem()
  implicit val mat = actorSystem.dispatcher
  Http().singleRequest(HttpRequest(uri = "http://localhost:9093")).recover {
    case ex => println(s"error: ${ex.getMessage}")
  }
```

logs:
```
[info] [2022-06-02 10:51:12,533] [INFO] [akka.actor.LocalActorRef] [default-akka.actor.default-dispatcher-5] [akka://default/system/IO-TCP/selectors/$a/0] - Message [akka.actor.ReceiveTimeout$] to Actor[akka://default/system/IO-TCP/selectors/$a/0#2008377474] was not delivered. [1] dead letters encountered. If this is not an expected behavior then Actor[akka://default/system/IO-TCP/selectors/$a/0#2008377474] may have terminated unexpectedly. This logging can be turned off or adjusted with configuration settings 'akka.log-dead-letters' and 'akka.log-dead-letters-during-shutdown'.
[info] [2022-06-02 10:51:12,674] [INFO] [akka.actor.LocalActorRef] [default-akka.actor.default-dispatcher-17] [akka://default/system/IO-TCP/selectors/$a/1] - Message [akka.actor.ReceiveTimeout$] to Actor[akka://default/system/IO-TCP/selectors/$a/1#-2005600569] was not delivered. [2] dead letters encountered. If this is not an expected behavior then Actor[akka://default/system/IO-TCP/selectors/$a/1#-2005600569] may have terminated unexpectedly. This logging can be turned off or adjusted with configuration settings 'akka.log-dead-letters' and 'akka.log-dead-letters-during-shutdown'.
[...]

PR #4133, opened 2022-06-07
core: use LIFO ordering in retry buffer of connection pool

jrudolph:
During connection problems, all slots will back off connection attempts,
so that effectively only a single request can make a new attempt (out of
potentially `max-connections` ones already pulled into the pool). However,
the previous order of putting requests back into the retry buffer means that
all requests in the pool get equal chances to run again. This way all requests
will fail approximately at the same time only after all requests have almost
timed out.

Using a LIFO buffer instead, makes it more likely that only a single request
at a time will get the chance to make another try, so it can be failed earlier,
then the next, etc. It does not change the overall throughput of failing requests
but decreases the average latency.

Refs #4128


johanandren (approved):
LGTM

Issue #4134, opened 2022-06-14
`toStrictEntity` throws an exception on invalid chunked requests


mpdn:
When using `toStrictEntity`, sending a chunked request with invalid framing causes an internal error, that is eventually propagated to the exception handler/log. As this is a bad request I would expect `toStrictEntity` to reject it as such, similar to how `entity` does, although it's not totally clear to me whether this is a bug or a footgun.

Repro code:

```
import akka.actor.ActorSystem
import akka.http.scaladsl.Http
import akka.http.scaladsl.model.StatusCodes
import akka.http.scaladsl.server.Directives._
import akka.http.scaladsl.server.ExceptionHandler
import java.net.Socket
import java.nio.charset.StandardCharsets
import org.scalatest.funsuite.AnyFunSuite
import scala.concurrent.duration.{ Duration, MINUTES, SECONDS }
import scala.concurrent.{ Await, ExecutionContext, Promise, TimeoutException }
import scala.util.Using

class AkkaTest extends AnyFunSuite {
  test("Partial request results in") {
    implicit val actorSystem: ActorSystem = ActorSystem()
    implicit val dispatcher: ExecutionContext = actorSystem.dispatcher

    val exception = Promise[Throwable]

    val exceptionHandler = ExceptionHandler { case e: Throwable =>
      exception.failure(e)
      complete(StatusCodes.InternalServerError)
    }

    val route = handleExceptions(exceptionHandler) {
      toStrictEntity(Duration(1, MINUTES)) {
        complete(StatusCodes.OK)
      }
    }

    val bindingFuture = Http()
      .newServerAt("127.0.0.1", 0)
      .bind(route)

    val binding = Await.result(bindingFuture, Duration.Inf)
    try {
      Using.resource(new Socket(null: String, binding.localAddress.getPort)) { socket =>
        val content =
          """POST /foo HTTP/1.1
            |Host: www.example.com
            |Transfer-Encoding: chunked
            |
            |2
            |a
            |"""

        socket.getOutputStream.write(content
          .stripMargin
          .replace("\n", "\r\n")
          .getBytes(StandardCharsets.UTF_8)
        )
      }

      assertThrows[TimeoutException] {
        Await.result(exception.future, Duration(10, SECONDS))
      }
    } finally {
      Await.result(binding.unbind(), Duration.Inf)
    }
  }
}
```

Causes this exception:

```
[...]

PR #4135, opened 2022-06-15
Make `toStrictEntity` fail with bad request in case of entity errors

mpdn:
Fixes https://github.com/akka/akka-http/issues/4134

Unsure what exactly to add to the docstring, if anything. Could be something like `Fails the route with an [IllegalRequestException] if the request framing is invalid` - thoughts?

This rethrows `EntityStreamException` as `IllegalRequestException`. This matches more things than the above issue - eg. situations such as exceeding the max entity size. Turns out, exceeding the max entity size also resulted in an internal server error previously and due to this change now results in a bad request - it really ought to be a 413 entity too large, but I can't really distinguish these errors as they currently use the same underlying exception types. But at least 400 Bad Request is better than 500 internal error.

### DIFF
--- a/akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/BasicDirectivesSpec.scala
+++ b/akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/BasicDirectivesSpec.scala
@@ -206,17 +206,19 @@ class BasicDirectivesSpec extends RoutingSpec {
           complete("content")
         }
       } ~> check {
-        entityAs[String] shouldEqual s"Request too large"
+        entityAs[String] shouldEqual "Request too large"
         status shouldEqual StatusCodes.BadRequest
       }
     }
 
     "return 400 Bad request on EntityStreamException" in {
       val errorMessage = "An EntityStreamException error"
-      val failingEntity = HttpEntity(ContentTypes.`text/plain(UTF-8)`, Source.failed(EntityStreamException(errorMessage)))
+      val errorDetail = "The internal details of the error"
+      val entity = Source.failed(EntityStreamException(errorMessage, errorDetail))
+      val request = HttpEntity(ContentTypes.`text/plain(UTF-8)`, entity)
       val timeout = 10.milliseconds
 
-      Post("/abc", failingEntity) ~> {
+      Post("/abc", request) ~> {
         extractStrictEntity(timeout) { _ =>
           complete("content")
         }
--- a/akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/BasicDirectivesSpec.scala
+++ b/akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/BasicDirectivesSpec.scala
@@ -197,6 +197,21 @@ class BasicDirectivesSpec extends RoutingSpec {
       }
     }
 
+    "return 400 Bad request on EntityStreamException" in {
+      val errorMessage = "An EntityStreamException error"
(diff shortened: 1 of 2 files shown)


akka-http/src/main/scala/akka/http/scaladsl/server/directives/BasicDirectives.scala
```diff
@@ -402,6 +402,8 @@ trait BasicDirectives {
           throw IllegalRequestException(
             StatusCodes.RequestTimeout,
             ErrorInfo(s"Request timed out after $timeout while waiting for entity data", "Consider increasing the timeout for toStrict"))
+        case EntityStreamException(info) =>
+          throw IllegalRequestException(StatusCodes.BadRequest, info)
```
raboof, line 406: `info` is already split into a `summary` that may be public and a `detail` that should not be included in the response by default, so that's :+1:

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/BasicDirectivesSpec.scala
```diff
@@ -197,6 +197,35 @@ class BasicDirectivesSpec extends RoutingSpec {
...
+        }
+      } ~> check {
+        entityAs[String] shouldEqual s"Request too large"
+        status shouldEqual StatusCodes.BadRequest
+      }
+    }
+
+    "return 400 Bad request on EntityStreamException" in {
+      val errorMessage = "An EntityStreamException error"
+      val failingEntity = HttpEntity(ContentTypes.`text/plain(UTF-8)`, Source.failed(EntityStreamException(errorMessage)))
```
raboof, line 216: Might be good to also pass error details, so the test verifies those do not show up in the response.

raboof (approved):
> This matches more things than the above issue - eg. situations such as exceeding the max entity size

That seems like an overall improvement to me

> it really ought to be a 413 entity too large

400 is already an improvement over 500, but I agree 413 would be even better. We actually do have an `EntityStreamSizeException` - I wonder why we don't throw that one in `toStrict` instead of a generic `EntityStreamException`. That might be worth a look.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/BasicDirectives.scala
mpdn: Good idea - detail added.

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/BasicDirectivesSpec.scala
mpdn: Good idea - detail added.

mpdn:
> We actually do have an `EntityStreamSizeException` - I wonder why we don't throw that one in `toStrict` instead of a generic `EntityStreamException`.

Yeah, I was wondering about that as well. `EntityStreamSizeException` does not extend `EntityStreamException` though, and does not have an `ErrorInfo` so it's fairly different to the exception it currently throws.

raboof:
> > We actually do have an `EntityStreamSizeException` - I wonder why we don't throw that one in `toStrict` instead of a generic `EntityStreamException`.
> 
> Yeah, I was wondering about that as well. `EntityStreamSizeException` does not extend `EntityStreamException` though, and does not have an `ErrorInfo` so it's fairly different to the exception it currently throws.

Ah right - so it's useful to use `EntityStreamException` to get the split between public and private info, and `EntityStreamSizeException` won't get that until we finish https://github.com/akka/akka-http/pull/3439 . :+1:

johanandren (approved):

PR #4144, opened 2022-08-03
document ByteRange suffix

nkconnor:
I wasn't sure what the behavior of Suffix was. I had to dig into the source to see if it was applicable for my request (like `bytes=-100`). This should make it more clear.


johanandren (approved):
LGTM, thanks!

raboof (approved):

PR #4158, opened 2022-09-16
=core Fold InHandler and OutHandler to GraphStageLogic.

He-Pin:
References https://github.com/akka/akka/issues/19752

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/ws/FrameEventRenderer.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/ws/FrameEventRenderer.scala
@@ -33,7 +33,7 @@ private[http] final class FrameEventRenderer extends GraphStage[FlowShape[FrameE
 
           push(out, renderStart(start))
 
-        case f: FrameData =>
+        case _: FrameData =>
           fail(out, new IllegalStateException("unexpected FrameData (need FrameStart first)"))
       }
     }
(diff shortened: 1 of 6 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/ws/FrameEventRenderer.scala
```diff
@@ -22,21 +22,19 @@ private[http] final class FrameEventRenderer extends GraphStage[FlowShape[FrameE
   val out = Outlet[ByteString]("FrameEventRenderer.out")
   override val shape = FlowShape(in, out)
 
-  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new GraphStageLogic(shape) {
-
-    val Initial = new InHandler {
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
```
johanandren, line 25: I think we should leave this one as is because it makes the state machine a bit easier to follow (Initial -> subsequent render states)

He-Pin: okay：）

johanandren (approved):
LGTM, thanks!

johanandren:
Some formatting is off though

johanandren (approved):
LGTM, thanks!

PR #4167, opened 2022-10-13
Update to Jackson databind 2.13.4.1

patriknw:
* Fixes https://nvd.nist.gov/vuln/detail/CVE-2022-42003 (even if we were not really affected out of the box).


patriknw:
fail is only Fossa that doesn't know about the patch yet

johanandren (approved):

PR #4180, opened 2022-10-25
Update release issue template

patriknw:


docs/release-train-issue-template.md
```diff
@@ -0,0 +1,73 @@
...
+- [ ] Create a new milestone for the [next version](https://github.com/akka/akka-http/milestones)
+- [ ] Close the [$VERSION$ milestone](https://github.com/akka/akka-http/milestones?direction=asc&sort=due_date)
+- [ ] Make sure all important PRs have been merged
+- [ ] Wait until [main build finished](https://github.com/akka/akka-http/actions) after merging the latest PR
+- [ ] Update the [draft release](https://github.com/akka/akka-http/releases) with the next tag version `v$VERSION$`, title and release description. Use the `Publish release` button, which will create the tag.
+- [ ] Check that GitHub Actions release build has executed successfully (GitHub Actions will start a [CI build](https://github.com/akka/akka-http/actions) for the new tag and publish artifacts to Maven central via Sonatype)
+- [ ] Run a test against the staging repository to make sure the release went well, for example by using https://github.com/akka/akka-http-quickstart-scala.g8 and adding the sonatype staging repo with `resolvers += "Staging Repo" at "https://oss.sonatype.org/content/repositories/staging"`
```
patriknw, line 28: Staging is different from all other repositories. Worth keeping?

docs/release-train-issue-template.md
```diff
@@ -0,0 +1,73 @@
...
+- [ ] Announce on [Gitter akka/akka](https://gitter.im/akka/akka)
+- [ ] Announce internally (with links to Tweet, discuss)
+
+For minor or major releases:
+
+- [ ] Include noteworthy features and improvements in Akka umbrella release announcement at akka.io. Coordinate with PM and marketing.
+
+### Afterwards
+
+- [ ] Add the released version to `project/MiMa.scala` to the `mimaPreviousArtifacts` key *of all current compatible branches*.
```
patriknw, line 69: The mima step should probably be included in other repos as well, I'll iterate.

docs/release-train-issue-template.md
```diff
@@ -0,0 +1,73 @@
...
+
+Key links:
+  - akka/akka-http milestone: https://github.com/akka/akka-http/milestone/?
+-->
+
+### Cutting the release
+
+- [ ] Check that open PRs and issues assigned to the milestone are reasonable
+- [ ] For minor or major versions, update the Change date in the LICENSE file and update the `licenses` url in the build.
+- [ ] Create a PR to add a release notes entry in docs/src/main/paradox/release-notes/. As a helper run `scripts/commits-for-release-notes.sh <last-version-tag>` which will output a list of commits grouped by submodule, and the closed issues for this milestone
```
johanandren, line 21: So we keep this special case for Akka HTTP for now rather than completely align cross projects?

scripts/release-train-issue-template.md
```diff
@@ -1,91 +0,0 @@
-Release Akka HTTP $VERSION$
```
patriknw, line 1: this might be a better place for the template, but all other repos have it in docs

docs/release-train-issue-template.md
johanandren: I think what Raboof used to do, running the Akka gRPC test suite against the artifacts before releasing, always felt a little bit reassuring, but not so sure of the actual value of it.

Perhaps better to have such tests continuously running than to defer them to release.

docs/release-train-issue-template.md
johanandren: I agree, it feels like a cumbersome step to me.

docs/release-train-issue-template.md
patriknw: That's not needed in other repos (other than akka/akka) because in those we only validate against latest previous, which is good enough for those.

docs/release-train-issue-template.md
patriknw: let's keep it for now

docs/release-train-issue-template.md
```diff
@@ -0,0 +1,76 @@
...
+
+Key links:
+  - akka/akka-http milestone: https://github.com/akka/akka-http/milestone/?
+-->
+
+### Cutting the release
+
+- [ ] Check that open PRs and issues assigned to the milestone are reasonable
+- [ ] For minor or major versions, update the Change date in the LICENSE file and update the `licenses` url in the build.
+- [ ] For minor or major versions, add a release notes entry in `docs/src/main/paradox/release-notes/`.
```
patriknw, line 21: @johanandren I adjusted this according to https://github.com/akka/akka-http/pull/4183

PR #4181, opened 2022-10-27
Update MiMa for 10.4.0

patriknw:


johanandren:
I think the second one is safe to remove as well as it does no longer compile with Akka HTTP 10.2

johanandren (approved):

patriknw:
thanks

PR #4182, opened 2022-10-31
chore: Silence or fix most compilation warnings

johanandren:
Mainly
 * Deprecated ActorMaterializer usage, fixed or silenced
 * Scala 2.13 completeness check, most fixed, test unnapplies left
 * IOResul deprecation in Akka 2.6, special handling removed
 * Misc others silenced with annotations
 * Switching from deprecated akka-stream methods to the new ones (since we no longer support Akka 2.5)

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Demux.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Demux.scala
@@ -27,6 +27,7 @@ import akka.stream.scaladsl.Source
 import akka.stream.stage.{ GraphStageLogic, GraphStageWithMaterializedValue, InHandler, OutHandler, StageLogging, TimerGraphStageLogic }
 import akka.util.{ ByteString, OptionVal }
 
+import scala.annotation.nowarn
 import scala.collection.immutable
 import scala.concurrent.{ ExecutionContext, Future, Promise }
 import scala.concurrent.duration.Duration
@@ -289,7 +290,7 @@ private[http2] abstract class Http2Demux(http2Settings: Http2CommonSettings, ini
 
         pingState.tickInterval().foreach(interval =>
           // to limit overhead rather than constantly rescheduling a timer and looking at system time we use a constant timer
-          schedulePeriodically(ConfigurablePing.Tick, interval)
+          scheduleWithFixedDelay(ConfigurablePing.Tick, interval, interval)
         )
       }
 
@@ -371,6 +372,7 @@ private[http2] abstract class Http2Demux(http2Settings: Http2CommonSettings, ini
           pullFrameIn()
         }
 
+        @nowarn("msg=deprecated")
         override def onUpstreamFailure(ex: Throwable): Unit = {
           ex match {
             // every IllegalHttp2StreamIdException will be a GOAWAY with PROTOCOL_ERROR
@@ -392,6 +394,9 @@ private[http2] abstract class Http2Demux(http2Settings: Http2CommonSettings, ini
             // handle every unhandled exception
             case NonFatal(e) =>
               super.onUpstreamFailure(e)
+
+            case other =>
+              throw other // compiler completeness check pleaser
           }
         }
       })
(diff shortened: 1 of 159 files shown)


johanandren:
Failure is fossa not knowing the license of jackson.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Demux.scala
```diff
@@ -289,7 +290,7 @@ private[http2] abstract class Http2Demux(http2Settings: Http2CommonSettings, ini
 
         pingState.tickInterval().foreach(interval =>
           // to limit overhead rather than constantly rescheduling a timer and looking at system time we use a constant timer
-          schedulePeriodically(ConfigurablePing.Tick, interval)
+          scheduleWithFixedDelay(ConfigurablePing.Tick, interval, interval)
```
patriknw, line 293: Note that `schedulePeriodically` is the same as `scheduleAtFixedRate` so this could change the rate of the ticks slightly. I don't know if that would matter here but in general `scheduleWithFixedDelay` is more safe since it doesn't generate bursts after pauses.

```
   * Warning: `scheduleAtFixedRate` can result in bursts of scheduled tasks after long
   * garbage collection pauses, which may in worst case cause undesired load on the system.
   * Therefore `scheduleWithFixedDelay` is often preferred.
```

akka-http-core/src/main/scala/akka/http/impl/engine/http2/framing/FrameRenderer.scala
```diff
@@ -91,7 +91,8 @@ private[http2] object FrameRenderer {
               b.putInt32(value)
 
               renderNext(remaining)
-            case Nil =>
+            case Nil   =>
+            case other => throw new IllegalArgumentException(s"Unexpected remaining: $other") // compiler completeness check pleaser
```
patriknw, line 95: Even though this shouldn't happen, could this result in that we toString on unknown stuff and possibly include sensitive (or too much) in logs? Similar for other places where the compiler completeness was added.

patriknw (commented):
Great cleanup, two questions...

akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2Demux.scala
johanandren: This one is used to tick itself, and check if it should keep a HTTP2 ping or has waited for too long for a pong and should timeout the connection, so worst case with a delayed tick is a connection was kept alive longer than the timeout before it actually timed out. I think that is better than bursts here.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/framing/FrameRenderer.scala
johanandren: Good thinking, I mostly thought that it would be good with details in case the should never happen actually happens, but better safe, at least on parse/render side of things. I'll take another pass through those and see.

johanandren: Done in 8d14ad6f4

akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpHeaderParser.scala
```diff
@@ -130,7 +131,9 @@ private[engine] final class HttpHeaderParser private (
         values(valueIx) match {
           case branch: ValueBranch            => parseHeaderValue(input, cursor, branch)()
           case valueParser: HeaderValueParser => startValueBranch(valueIx, valueParser) // no header yet of this type
-          case EmptyHeader                    => resultHeader = EmptyHeader; cursor
+          case EmptyHeader =>
+            resultHeader = EmptyHeader; cursor
```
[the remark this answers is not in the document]
johanandren: Accidental I think (if it wasn't scalafmt)

PR #4188, opened 2022-11-10
build: Fix release script typo

johanandren:


patriknw (approved):
ouch, might be copied to other places

PR #4192, opened 2022-11-15
ci: remove name from JVM selection again

ennru:
The filenames for uploads may not contain `:`, splitting out the JVM name.

References #4191

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -16,7 +16,14 @@ jobs:
       fail-fast: false
       matrix:
         SCALA_VERSION: [2.12, 2.13, 3.1]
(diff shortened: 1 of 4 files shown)


patriknw:
Note that you must update the branch protection rules when changing the names. Currently it doesn't include all:
Screenshot 2022-11-15 at 13 26 01

ennru:
Should even 3.1 builds be required now? For all JVMs?

johanandren (approved):

ennru:
3.1 builds are not required for all JVM versions.

patriknw:
> Should even 3.1 builds be required now? For all JVMs?

I don't think we should have an optional jobs in PR validation. That is just very confusing. However, we don't need to run all combinations in PR validation. We can defer some of them to nightly (or main merge) builds.

PR #4196, opened 2022-11-21
ci: More memory for JDK8 nightlies

ennru:
Nightlies started failing with out of heap errors on JDK8 last week. Not obvious why it shows now on the renovated workflows. The JDK version was Temurin 8u352b08 even before switching to coursier/setup-action.


johanandren (approved):
Let's see if this is enough heap. 👍

johanandren:
Looks like it may be needed for the PR validation JDK 8 build as well, if this fixes it.

PR #4199, opened 2022-11-21
Set max heap for JDK 8 in PR validation

johanandren:
Following up #4196


ennru (approved):
Always remember the context for where things go 💯 
LGTM.

PR #4200, opened 2022-11-21
Even more memory to CI JDK 8

johanandren:


ennru (approved):
Bump it!

akka-http-core/src/main/scala/akka/http/ParsingErrorHandler.scala
```diff
@@ -9,6 +9,8 @@ import akka.http.javadsl.{ model => jm }
 import akka.http.scaladsl.model.{ ErrorInfo, HttpResponse, StatusCode }
 import akka.http.scaladsl.settings.ServerSettings
 
+// FIXME trigger PR validation
```
patriknw, line 12: @ennru accidental commit?

johanandren: Removed in #4201

johanandren:
I was thinking it is JDK 8 ergonomics changing between JDK 8 versions, making it (sometimes) guess bad limits in a container in the version we switched to. I don't have much evidence for that though. 

Note that it was OOM:ing something like 40-50% of the test runs on JDK 8, not consistently.

PR #4214, opened 2023-01-12
fix #4213: Content-Length when entity is unallowed

### LINKED ISSUE #4213: Empty entity should set Content-Length header
When a Play application running on akka-http returns status 205 Reset Content, some clients hang with this response and just wait for it to complete, even though the server considers it is complete and will never send any further data. I have encountered the issue with the following clients: `play-ws-standalone` (with underlying Netty-based AHC), Apache Commons `http-client`, and `curl` in command line. Some clients do not hang and immediately consider the request complete, such as Chrome or Firefox browsers.

It has to do with the fact that the response has no body (which is correct, as mandated by RFC for status 205), and no indication of content length to explicitly say that there is no body. That last part is incorrect behavior according to RFC 7231 section 6.3.6:

> Since the 205 status code implies that no additional content will be provided, a server MUST NOT generate a payload in a 205 response.  In other words, a server MUST do one of the following for a 205 response: a) indicate a zero-length body for the response by including a Content-Length header field with a value of 0; b) indicate a zero-length payload for the response by including a Transfer-Encoding header field with a value of chunked and a message body consisting of a single chunk of zero-length; or, c) close the connection immediately after sending the blank line terminating the header section.

It seems the HTTP clients I've identified, do rely on this requirement stated by RFC. Testing with servers that add the `Content-Length: 0` header, they work as expected. Also, they may eventually work if the server reaches its connection timeout before the client does, and thus closes the connection (which is a valid way to complete the response as seen from the client).

I found an earlier mention of this issue in 2021 on [Gitter](https://gitter.im/playframework/playframework?at=6017e61232e01b4f71692e76) but it seems to not have been followed through.

Credit to Ben Fradet for identifying the same issue with Tomcat in 2017; I shamelessly copy-pasted half of [his issue on the Apache tracker](https://bz.apache.org/bugzilla/show_bug.cgi?id=61086) over here.

bursauxa:
Adds `Content-Length: 0` header on responses that do not allow a response body (204, 205 and 304).

I did not add a specific test, as there was already a test for rendering a 304 response - I simply made that test more correct :slightly_smiling_face:

Fixes #4213

### DIFF
--- a/akka-http-core/src/main/java/akka/http/javadsl/model/HttpMethods.java
+++ b/akka-http-core/src/main/java/akka/http/javadsl/model/HttpMethods.java
@@ -27,12 +27,24 @@ public final class HttpMethods {
 
     /**
      * Create a custom method type.
+     * @deprecated The created method will compute the presence of Content-Length headers based on deprecated logic (before issue #4213).
      */
+    @Deprecated
     public static HttpMethod custom(String value, boolean safe, boolean idempotent, akka.http.javadsl.model.RequestEntityAcceptance requestEntityAcceptance) {
         //This cast is safe as implementation of RequestEntityAcceptance only exists in Scala
         akka.http.scaladsl.model.RequestEntityAcceptance scalaRequestEntityAcceptance
           = (akka.http.scaladsl.model.RequestEntityAcceptance) requestEntityAcceptance;
         return akka.http.scaladsl.model.HttpMethod.custom(value, safe, idempotent, scalaRequestEntityAcceptance);
+    }
+
+    /**
+     * Create a custom method type.
+     */
+    public static HttpMethod custom(String value, boolean safe, boolean idempotent, akka.http.javadsl.model.RequestEntityAcceptance requestEntityAcceptance, boolean contentLengthAllowed) {
+        //This cast is safe as implementation of RequestEntityAcceptance only exists in Scala
+        akka.http.scaladsl.model.RequestEntityAcceptance scalaRequestEntityAcceptance
+          = (akka.http.scaladsl.model.RequestEntityAcceptance) requestEntityAcceptance;
+        return akka.http.scaladsl.model.HttpMethod.custom(value, safe, idempotent, scalaRequestEntityAcceptance, contentLengthAllowed);
     }
 
     /**
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/rendering/HttpResponseRendererFactory.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/rendering/HttpResponseRendererFactory.scala
@@ -234,7 +234,7 @@ private[http] class HttpResponseRendererFactory(
           }
 
           def renderContentLengthHeader(contentLength: Long) =
-            if (status.allowsEntity) r ~~ ContentLengthBytes ~~ contentLength ~~ CrLf else r
+            if (ctx.requestMethod.contentLengthAllowed(status)) r ~~ ContentLengthBytes ~~ contentLength ~~ CrLf else r
 
           def headersAndEntity(entityBytes: => Source[ByteString, Any]): StrictOrStreamed =
             if (noEntity) {
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMethod.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMethod.scala
@@ -9,6 +9,7 @@ import java.util.Locale
 import akka.http.impl.util._
 import akka.http.javadsl.{ model => jm }
 import akka.http.scaladsl.model.RequestEntityAcceptance._
+import akka.util.ConstantFun
 
 sealed trait RequestEntityAcceptance extends jm.RequestEntityAcceptance {
   def isEntityAccepted: Boolean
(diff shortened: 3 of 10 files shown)


lightbend-cla-validator:
At least one commit author (antoine.bursaux@kelkoogroup.com) is not linked to a user. See https://help.github.com/en/articles/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user

bursauxa:
Thanks @johanandren for pointing that out. You are right, it will require some case-by-case handling, I will proceed with the changes.

johanandren:
Wdyt about the explicit MUST NOT emit content length for 204, that seems like it contradicts what the clients expect if it is as you describe in the issue?

bursauxa:
> Wdyt about the explicit MUST NOT emit content length for 204, that seems like it contradicts what the clients expect if it is as you describe in the issue?

I specifically encountered the issue with a 205 code. It was mentioned in the linked issue, but I oversimplified in this pull request. Testing with `curl` shows that indeed, a 204 code without `Content-Length` header works fine - the client does not hang.

It turns out to be very specific to the 205 code... The previous condition on `status.allowsEntity` matches all codes where `Content-Length` must not be used, with the exception of 205 that does not allow an entity but should still provide a `Content-Length`.

There are also cases with HEAD and CONNECT methods that I did not encounter, I do not even know to what degree they would realistically cause issues to clients (a response body seems very unlikely for these methods), but might as well align them with the RFC while working on it.

bursauxa:
Not entirely related, but you made me think @johanandren ... And I stumbled upon these tests with HEAD requests containing bodies (`ResponseRendererSpec.scala`, around line 150). This is not valid according to the same [RFC 9110](https://httpwg.org/specs/rfc9110.html#HEAD) you mentioned earlier:

> The HEAD method is identical to GET except that the server MUST NOT send content in the response.

akka-http-core/src/main/scala/akka/http/impl/engine/rendering/HttpResponseRendererFactory.scala
```diff
@@ -233,8 +233,21 @@ private[http] class HttpResponseRendererFactory(
...
+            // - for HEAD it is technically allowed, but must match the content-length of hypothetical GET request, which is not handled here
+            // - for 304 (Not Modified) it must similarly match the content-length of hypothetical 200-accepted request, which is not handled here
+            // - for 1xx (Informational) or 204 (No Content) it is explicitly not allowed
+            // - for 2xx (Successful) it is explicitly not allowed when the method is CONNECT
+            val isContentLengthForbidden =
+              (ctx.requestMethod eq HttpMethods.HEAD) ||
+              (status eq StatusCodes.NotModified) ||
+              (100 to 199 contains status.intValue) ||
+              (status eq StatusCodes.NoContent) ||
+              ((ctx.requestMethod eq HttpMethods.CONNECT) && (200 to 299 contains status.intValue))
```
johanandren, line 247: Since this is called on every request-response we need to be careful to make it as cheap as possible for the most common cases (and if for all cases even better). Comparing with a range allocates a `scala.util.Range` each time for example.

I wonder if we could push it to a call on `StatusCode` and maybe `HttpMethod` as well (something like `Method#contentLengthAllowed(forStatus: StatusCode)`) instead, making it cheaper.

johanandren:
I think we can leave HEAD returning bodies alone for now, as you say it is against spec, but in case anyone is relying on doing that already.

akka-http-core/src/main/scala/akka/http/impl/engine/rendering/HttpResponseRendererFactory.scala
bursauxa: Sounds good, I added it to HttpMethod :+1:

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMethod.scala
```diff
@@ -32,41 +32,52 @@ object RequestEntityAcceptance {
...
 
+  // define requirements for content-length according to https://httpwg.org/specs/rfc9110.html#field.content-length
+  // for CONNECT it is explicitly not allowed in the 2xx (Successful) range
+  private def contentLengthAllowedForConnect(forStatus: StatusCode): Boolean = forStatus.intValue < 200 || forStatus.intValue >= 300
+  // for HEAD it is technically allowed, but must match the content-length of hypothetical GET request, so can not be anticipated
+  private def contentLengthAllowedForHead(forStatus: StatusCode): Boolean = false
+  // for other methods there are common rules:
+  // - for 304 (Not Modified) it must match the content-length of hypothetical 200-accepted request, so can not be anticipated
+  // - for 1xx (Informational) or 204 (No Content) it is explicitly not allowed
+  private def contentLengthAllowedCommon(forStatus: StatusCode): Boolean = forStatus.intValue >= 200 && !(forStatus eq StatusCodes.NoContent) && !(forStatus eq StatusCodes.NotModified)
```
johanandren, line 69: Possible tiny optimization comparing only ints:

```
val code = forStatus.intValue
(code >= 200 && code != 204 && code != 304)
// or perhaps even more readable
!(code < 200 || code == 204 || code == 304)
```

akka-http-core/src/main/java/akka/http/javadsl/model/HttpMethods.java
```diff
@@ -28,11 +28,11 @@ private HttpMethods() {}
     /**
      * Create a custom method type.
      */
-    public static HttpMethod custom(String value, boolean safe, boolean idempotent, akka.http.javadsl.model.RequestEntityAcceptance requestEntityAcceptance) {
+    public static HttpMethod custom(String value, boolean safe, boolean idempotent, akka.http.javadsl.model.RequestEntityAcceptance requestEntityAcceptance, boolean contentLengthAllowed) {
```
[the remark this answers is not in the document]
bursauxa: Which way should I go to mark the older method deprecated? Standard Java `@Deprecated` or is there something specific?

johanandren:
Some related test failures, and some I'm not entirely sure about, to look into.

Also, failures from MiMa which is our binary incompatibility detection, I think all of them can now safely be filtered as described here: https://github.com/akka/akka-http/blob/main/CONTRIBUTING.md#binary-compatibility

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMethod.scala
```diff
@@ -32,41 +32,77 @@ object RequestEntityAcceptance {
...
   def custom(name: String, safe: Boolean, idempotent: Boolean, requestEntityAcceptance: RequestEntityAcceptance): HttpMethod = {
     require(name.nonEmpty, "value must be non-empty")
     require(!safe || idempotent, "An HTTP method cannot be safe without being idempotent")
-    apply(name, safe, idempotent, requestEntityAcceptance)
+    apply(name, safe, idempotent, requestEntityAcceptance, oldContentLengthCondition)
+  }
+
+  // declare constant functions so custom HttpMethod instances are equal (case class equality) when built with the same parameters
+  private val constantTrue: StatusCode => Boolean = (_) => true
+  private val constantFalse: StatusCode => Boolean = (_) => false
```
johanandren, line 62: You can use `akka.util.ConstantFun#anyToTrue` and `akka.util.ConstantFun#anyToFalse` instead of defining these.

akka-http-core/src/test/scala/akka/http/impl/engine/client/HostConnectionPoolSpec.scala
```diff
@@ -200,7 +200,7 @@ class HostConnectionPoolSpec extends AkkaSpecWithMaterializer(
 
           conn1.pushResponse(HttpResponse(entity = HttpEntity.Default(ContentTypes.`application/octet-stream`, 100, Source.empty)))
           val res = expectResponse()
-          res.entity.contentLengthOption.get shouldEqual 100
+          res.entity.contentLengthOption.get shouldEqual 0
```
johanandren, line 203: Shouldn't that option be None in these two HEAD scenarios rather than exist and have 0 length?

bursauxa: I thought the same, but the way it is implemented, it points to the entity so there is indeed a length defined, at least for `Strict` and `Default` entities (unless I am mistaken, here it is `EmptyEntity` which is a special case of `Strict`). The difference in output is performed in the renderer.

Maybe it could be improved by also making changes at the model level, but I feel like this PR is already "large" for such a small change on the functional surface, so I am quite reluctant about it. What is your opinion?

johanandren: Yeah, let's not touch that.

bursauxa:
Tests were good with previous run, this one should take care of binary incompatibilities.

Issue #2394, opened 2019-01-19
Custom HTTP method closes connection unexpectedly in HTTP/2


miracle2121:
When Akka HTTP server is running with HTTP/2, it closes connection unexpectedly when sending a custom HTTP method. It works for HTTP 1.1 which returns a 501. 
Also, it doesn't seem to honor the custom HTTP methods passed into `ParsetSettings` when creating the server.

```
rli@rli-mn1 $ curl -i -X FOO --http1.1 https://rli-mn1.linkedin.biz:9443/play-skeleton-mt/admin
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100    23  100    23    0     0    658      0 --:--:-- --:--:-- --:--:--   676
HTTP/1.1 501 Not Implemented
Date: Sat, 19 Jan 2019 00:10:19 GMT
Connection: close
Content-Type: text/plain; charset=UTF-8
Content-Length: 23

Unsupported HTTP method
rli@rli-mn1 $ curl -i -X FOO https://rli-mn1.linkedin.biz:9443/play-skeleton-mt/admin
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (56) Unexpected EOF
```

Following is the stacktrace:
```
2019/01/18 16:10:25.066 [^[[31merror^[[0m] [a.a.RepointableActorRef] [play-dev-mode-akka.actor.default-dispatcher-38] [play-skeleton-mt] [] Error in stage [ExposeAttributes(<function1>)]: Malformed request: Unknown HTTP method: 'FOO'
java.lang.RuntimeException: Malformed request: Unknown HTTP method: 'FOO'
        at akka.http.impl.engine.http2.RequestParsing$.malformedRequest(RequestParsing.scala:163)
        at akka.http.impl.engine.http2.RequestParsing$$anonfun$parseRequest$1$$anonfun$3.apply(RequestParsing.scala:94)
        at akka.http.impl.engine.http2.RequestParsing$$anonfun$parseRequest$1$$anonfun$3.apply(RequestParsing.scala:94)
        at scala.Option.getOrElse(Option.scala:121)
[...]

Issue #4226, opened 2023-02-15
HTTP/2 error when parsing percent-encoded query string


mkurz:
Very simple reproducer with explanation: https://github.com/mkurz/akka-http-percent-encoding-bug

Copy and pasted README (in case I every delete that repo):

akka-http's experimental http2 support does not correctly handle URL parsing errors.
Specially when passing an invalid percent-encoded character in the path and/or query string (here `%_D`):

As reproducer the example from the akka docs can be used, but with HTTP/2 enabled:

* https://doc.akka.io/docs/akka-http/current/introduction.html#low-level-http-server-apis
* https://doc.akka.io/docs/akka-http/current/server-side/http2.html#enable-http-2-support

```sh
# Start the server
sbt run
```

```sh
# HTTP/1 works:
$ curl --http1.1 -v http://localhost:8080/?param=%_D
*   Trying 127.0.0.1:8080...
* Connected to localhost (127.0.0.1) port 8080 (#0)
> GET /?param=%_D HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.87.0
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 400 Bad Request
< Server: akka-http/10.2.10
< Date: Wed, 15 Feb 2023 16:29:18 GMT
< Connection: close
< Content-Type: text/plain; charset=UTF-8
< Content-Length: 78
< 
* Closing connection 0
Illegal request-target: Invalid input '_', expected HEXDIG (line 1, column 10)
```

```sh
# HTTP/2 gives error:
$ curl --http2-prior-knowledge -v http://localhost:8080/?param=%_D
*   Trying 127.0.0.1:8080...
* Connected to localhost (127.0.0.1) port 8080 (#0)
* Using HTTP2, server supports multiplexing
* Copying HTTP/2 data in stream buffer to connection buffer after upgrade: len=0
* h2h3 [:method: GET]
* h2h3 [:path: /?param=%_D]
* h2h3 [:scheme: http]
* h2h3 [:authority: localhost:8080]
* h2h3 [user-agent: curl/7.87.0]
* h2h3 [accept: */*]
* Using Stream ID: 1 (easy handle 0xaaabd90f0dc0)
> GET /?param=%_D HTTP/2
> Host: localhost:8080
> user-agent: curl/7.87.0
> accept: */*
> 
* Connection state changed (MAX_CONCURRENT_STREAMS == 256)!
* Closing connection 0
[...]

PR #4227, opened 2023-02-17
fix: Invalid HTTP2 request headers should lead to bad request responses

johanandren:
References #4226
Also fixes https://github.com/akka/akka-http/issues/2394


patriknw (approved):
LGTM

Issue #4228, opened 2023-02-17
The backwards compatibility for http2 enable doesn't work


johanandren:
I thought I covered it with tests but I have now noticed it does not pick up being enabled via the old config path anymore.

PR #4229, opened 2023-02-17
fix: http2-enable flag was not picking up the old one

johanandren:
References #4228


patriknw (approved):
LGTM. Strange. Back in the old days config resolve didn't include application conf override but that was very long since that was changed. Probably some resolve or fallback setup that is making this. Anyway, this new way is safer.

PR #4232, opened 2023-02-21
chore: Change date and release notes for 10.5.0

johanandren:
References #4231


LICENSE.txt
```diff
@@ -15,7 +15,7 @@ Additional Use Grant:
     Connecting to a Play Framework websocket and/or Play Framework
     request/response bodies for server and play-ws client.
 
-Change Date:          TBD ( 3 years after 10.5.0 release )
+Change Date:          2026-02-21 ( 3 years after 10.5.0 release )
```
patriknw, line 18: I think the original instruction was to just have the date here, see https://github.com/akka/akka/commit/ca7052179650aec65ef9f5e73276f79be6542263

LICENSE.txt
```diff
@@ -3,7 +3,7 @@ Business Source License 1.1
 Parameters
 
 Licensor:             Lightbend, Inc.
-Licensed Work:        Akka HTTP 10.5.x
+Licensed Work:        Akka HTTP 10.5.0
```
patriknw, line 6: 👍

patriknw:
here is one more place https://github.com/akka/akka-http/blob/main/build.sbt#L32
it will end up in the pom.xml
Didn't we automate that somewhere else, taking the tag from the build?

LICENSE.txt
johanandren: 👍

patriknw (approved):
LGTM

PR #4234, opened 2023-02-21
build: Don't change build just before releasing, breaks release version

johanandren:


ennru (approved):
Not needed when publishing. LGTM.

Issue #4241, opened 2023-03-09
Http/2 server handler throwing closes connection


johanandren:
If the user handler throws an exception this leads to an error log, HTTP/2 protocol INTERNAL_ERROR returned and connection close.

For a client this ends up as a framing/protocol error, for example with curl `Error in the HTTP2 framing layer`. 

Not entirely sure this is a problem, it is perhaps up to the user to catch errors and turn them into HTTP 500 internal error responses, but it would also be nice if they didn't have to do that and exceptions thrown from user logic was turned into responses instead.

PR #4242, opened 2023-03-10
fix: Recover handler errors, turn into HTTP 500 response 

johanandren:
References #4241

### DIFF
--- a/akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2HighlevelServerSpec.scala
+++ b/akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2HighlevelServerSpec.scala
@@ -1,0 +1,138 @@
+/*
+ * Copyright (C) 2020-2023 Lightbend Inc. <https://www.lightbend.com>
+ */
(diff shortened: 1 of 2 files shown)


akka-http2-support/src/test/scala/akka/http/impl/engine/http2/Http2HighlevelServerSpec.scala
```diff
@@ -0,0 +1,138 @@
...
+      sendClientRequest()
+      val serverRequest2 = expectServerRequest()
+      serverRequest2.promise.success(HttpResponse())
+      val response2 = expectClientResponse()
+      response2.status should be(StatusCodes.OK)
+
+    }
+  }
+
+  case class ServerRequest(request: HttpRequest, promise: Promise[HttpResponse]) {
```
johanandren, line 70: Below is mostly copy pasta from Http2ClientServerSpec but that one has log capturing fail on errors logged so had to introduce new test file for this case.

efgpinto (approved):
Looks good to me!

Issue #4243, opened 2023-03-22
Failed: Http2PersistentClientTlsSpec when waiting for a response


johanandren:
https://github.com/akka/akka-http/actions/runs/4474855377/jobs/7863714930#step:7:1976

```
2023-03-21T02:22:42.2809273Z [info]   - should when waiting for a response *** FAILED *** (306 milliseconds)
2023-03-21T02:22:42.2840797Z [info]     java.lang.AssertionError: No severe log messages should be emitted during test run but got [1] warnings and [0] errors (see marked lines above)
2023-03-21T02:22:42.2841595Z [info]     at akka.http.impl.util.WithLogCapturing.withFixture(WithLogCapturing.scala:77)
```

[http2persistentclientfail.txt](https://github.com/akka/akka-http/files/11039501/http2persistentclientfail.txt)

octonato:
https://github.com/akka/akka-http/actions/runs/4580764617/jobs/8089764590#step:7:2190

octonato:
**Compile and test (3.2, 1.8, main)**
https://github.com/akka/akka-http/actions/runs/4943244959/jobs/8837533393#step:7:2364

**Compile and test (3.2, 1.7, main)**
https://github.com/akka/akka-http/actions/runs/4943244959/jobs/8837533762#step:7:2275

octonato:
**Compile and test (3.2, 1.8, main)**
https://github.com/akka/akka/actions/runs/4953727716/jobs/8861506809#step:5:2673

Failed on same job as above, but on a different build.

octonato:
https://github.com/akka/akka-http/actions/runs/5029876815/jobs/9021892632#step:7:2239

johanandren:
I have a fix waiting for review in https://github.com/akka/akka-http/pull/4260

octonato:
fixing in #4260

Issue #4244, opened 2023-03-24
failed: Http2PersistentClientSpec - HTTP 2 managed persistent connection (tls = true) should not leak any stages if completed should when waiting for a response


patriknw:
https://github.com/akka/akka-http/actions/runs/4496373106/jobs/7910961387#step:7:2011

patriknw:
@johanandren what shall we do about this test? Failing frequently, and we have 3 fail tickets https://github.com/akka/akka-http/issues?q=is%3Aissue+is%3Aopen+Http2PersistentClientSpec

johanandren:
We should probably spend some time to figure out what the flakiness is about.

johanandren:
https://github.com/akka/akka-http/actions/runs/4921593158/jobs/8791571934#step:7:2025

Issue #4225, opened 2023-02-15
Failed: Http2PersistentClientTlsSpec when some requests are waiting for a response


johanandren:
https://github.com/akka/akka-http/actions/runs/4182819742/jobs/7246397762#step:8:2110

```
--> [HTTP 2 managed persistent connection (tls = true) should transparently reconnect when connection is closed (with backoff: false) should when some requests are waiting for a response] Start of log messages of test that [Succeeded]
[1938](https://github.com/akka/akka-http/actions/runs/4182819742/jobs/7246397762#step:8:1939)
| [DEBUG] [02/15/2023 10:24:45.006] [pool-1-thread-1-ScalaTest-running-Http2PersistentClientTlsSpec] [WithLogCapturing(akka://Http2PersistentClientSpec)] Logging started for test [HTTP 2 managed persistent connection (tls = true) should transparently reconnect when connection is closed (with backoff: false) should when some requests are waiting for a response]
[1939](https://github.com/akka/akka-http/actions/runs/4182819742/jobs/7246397762#step:8:1940)
| [DEBUG] [02/15/2023 10:24:45.011] [pool-1-thread-1-ScalaTest-running-Http2PersistentClientTlsSpec] [akka.actor.ActorSystemImpl(Http2PersistentClientSpec)] Binding server using HTTP/2
[1940](https://github.com/akka/akka-http/actions/runs/4182819742/jobs/7246397762#step:8:1941)
| [DEBUG] [02/15/2023 10:24:45.013] [Http2PersistentClientSpec-akka.actor.internal-dispatcher-7] [akka://Http2PersistentClientSpec/system/IO-TCP/selectors/$a/8] Successfully bound to /127.0.0.1:40803
[1941](https://github.com/akka/akka-http/actions/runs/4182819742/jobs/7246397762#step:8:1942)
| [DEBUG] [02/15/2023 10:24:45.035] [Http2PersistentClientSpec-akka.actor.default-dispatcher-9] [Http2ClientDemux(akka://Http2PersistentClientSpec)] Changing state from Idle to WaitingForNetworkToSendControlFrames
[1942](https://github.com/akka/akka-http/actions/runs/4182819742/jobs/7246397762#step:8:1943)
[...]

Issue #4104, opened 2022-04-12
Failed: HTTP 2 managed persistent connection (tls = true) should transparently reconnect when connection is closed (with backoff: false) should when the first connection fails to materialize


jrudolph:
https://github.com/akka/akka-http/runs/5983063661?check_suite_focus=true

```
[info]   - should when the first connection fails to materialize *** FAILED *** (3 seconds, 55 milliseconds)
[info]     java.lang.AssertionError: assertion failed: timeout (3 seconds) during expectMsgClass waiting for class akka.http.impl.engine.http2.Http2PersistentClientSpec$ServerRequest
[info]     at scala.Predef$.assert(Predef.scala:279)
[info]     at akka.testkit.TestKitBase.expectMsgClass_internal(TestKit.scala:571)
[info]     at akka.testkit.TestKitBase.expectMsgType(TestKit.scala:543)
[info]     at akka.testkit.TestKitBase.expectMsgType$(TestKit.scala:542)
[info]     at akka.testkit.TestKit.expectMsgType(TestKit.scala:973)
[info]     at akka.http.impl.engine.http2.Http2PersistentClientSpec$TestSetup$server$.expectRequest(Http2PersistentClientSpec.scala:386)
[info]     at akka.http.impl.engine.http2.Http2PersistentClientSpec$$anon$4.<init>(Http2PersistentClientSpec.scala:230)
```

patriknw:
https://github.com/akka/akka-http/actions/runs/4760734610/jobs/8461324301#step:7:2151

patriknw:
https://github.com/akka/akka-http/actions/runs/8013747351/job/21891240918#step:7:2388

PR #4260, opened 2023-05-04
chore: Ignore test case tear down race condition warning

johanandren:
References #4244, #4225, #4104


octonato (approved):

octonato:
@johanandren, I closed https://github.com/akka/akka-http/issues/4243 because it was clear that this was related, but the issues you referenced in the description are less obvious to me if they are related or not.

Issue #4264, opened 2023-05-05
Produce RFC 8288 compliant Link headers


bblfish:
Currently, the Link header implementation produces RFC 5988 (2010) compliant Link header. But that has been superseded by RFC 8288 (2017). And as I show below, the current Akka implementation does not produce headers compliant for that later RFC, as it does not quote URLs. I discovered that by finding that the Link headers produced by Akka were not parsed correctly by http4s. See https://github.com/http4s/http4s/issues/7101

Below is the Reasoning I followed to work out what the situation is using extracts from the various RFCs.

## RFC Reasoning

In [RFC 8288](https://www.rfc-editor.org/rfc/rfc8288.html#section-3) defines the value of the Link header via the following simple ABNF:

```ABNF
Link       = #link-value
link-value = "<" URI-Reference ">" *( OWS ";" OWS link-param )
link-param = token BWS [ "=" BWS ( token / quoted-string ) ]
```

The problem we are having is with `token` and in particular the character `:` .
`token` is defined in [rfc 7230](https://www.rfc-editor.org/rfc/rfc7230#section-3.2.6) with the ABNF

```abnf
token         = 1*tchar
tchar          = "!" / "#" / "$" / "%" / "&" / "'" / "*"
                    / "+" / "-" / "." / "^" / "_" / "`" / "|" / "~"
                    / DIGIT / ALPHA
                    ; any VCHAR, except delimiters
```

which as I see does not contain the `:` character. 
That seems to confirm the implementation in http4s.

The problem is that the predecessor of rfc 7230, namely [rfc 5988](https://www.rfc-editor.org/rfc/rfc5988#section-4.2)

```abnf
link-value       = "<" URI-Reference ">" *( ";" link-param )
link-param     = ( ( "rel" "=" relation-types )
                 | ( "anchor" "=" <"> URI-Reference <"> )
                 | ( "rev" "=" relation-types )
                 | ( "hreflang" "=" Language-Tag )
                 | ( "media" "=" ( MediaDesc | ( <"> MediaDesc <"> ) ) )
                 | ( "title" "=" quoted-string )
                 | ( "title*" "=" ext-value )
                 | ( "type" "=" ( media-type | quoted-mt ) )
                 | ( link-extension ) )
   ;[snip]
relation-type  = reg-rel-type | ext-rel-type
reg-rel-type   = LOALPHA *( LOALPHA | DIGIT | "." | "-" )
ext-rel-type   = URI
```
[...]

PR #4267, opened 2023-05-08
fix: Align better with RFC 8288 by quoting strings with colon 

johanandren:
Refs #4264

Instead of the suggested always quoting I went for additional quoting when colon in string, keeping it closer to original behavior and also trading one more char check for two additional bytes in each Link/rel entry.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/LinkValue.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/LinkValue.scala
@@ -40,7 +40,9 @@ object LinkParam {
 }
 
 object LinkParams {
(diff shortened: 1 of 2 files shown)


bblfish:
That looks right. I guess you have a test that checks that the old style of Link headers still parses, 
```http
Link: <>; rel=https://www.w3.org/ns/auth/acl#accessControl
```
so that you maintain compatibility with older versions of Akka :-)

johanandren:
Had coverage but lost it when messing around with the original idea of always quoting. I'll make sure to re-add.

bblfish (approved):
This looks like the right fix.

octonato (approved):

PR #4281, opened 2023-06-12
fix: Use NIO for creating tmp files

mkurz:
Backports #4261

Can you please cut a 10.2.11 release with this merged?


patriknw:
There is currently no decision by Lightbend to backport this fix. I'll bring it up to product management.

jtownley:
This bug was not deemed critical and will not be back ported to 2.6. 

I am happy to chat about no cost licensing for Play that extends beyond the usage grant in the current license if required.

PR #4289, opened 2023-07-28
Websocket stream failure on receive/send timeout (network lost detection)

gawronA:
Hello all,

I would like to contribute to Akka project by providing a feature that allows to catch websocket disconnections due to network unavailability.
I have introduced two configuration parameters for websocket client:
```
akka.http.client.websocket.receive-idle-timeout
akka.http.client.websocket.send-idle-timeout
```
and server:
```
akka.http.server.websocket.receive-idle-timeout
akka.http.server.websocket.send-idle-timeout
```
that can be configured together with `akka.http.{client, server}.periodic-keep-alive-max-idle` to detect lost network connection.

The working principle is simple: if elements are not send or received (including ping/pong frames), the streams are failed with
`akka.stream.StreamIdleTimeoutException`

The default config values are `infinite` which means that this feature is disabled.

I have provided two new test modules: `WebSocketServerReceiveIdleTimeoutSpec.scala` and `WebSocketServerSendIdleTimeoutSpec.scala` because I couldn't find out how config overrides can be injected upon server creation (they are always taken from ActorSystem settings). If anyone have a better idea how this can be resolved, I'm open to proposals.

Thanks!

References #2419

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/ws/WebSocket.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/ws/WebSocket.scala
@@ -90,19 +90,20 @@ private[http] object WebSocket {
   private[this] final val PongFullFrame: FrameStart = FrameEvent.fullFrame(Opcode.Pong, None, ByteString.empty, fin = true)
   private[this] final val mkDirectAnswerPong = () => DirectAnswer(PongFullFrame)
 
-  /** The layer that transparently monitors data flow in both directions and fails the flow
+  /**
+   * The layer that transparently monitors data flow in both directions and fails the flow
    *  with [[akka.stream.StreamIdleTimeoutException]] if elements are not passed within configured time.
-   *  */
+   */
   def idleTimeout(settings: WebSocketSettings): BidiFlow[FrameHandler.Output, FrameHandler.Output, FrameOutHandler.Input, FrameOutHandler.Input, NotUsed] = {
     if (settings.receiveIdleTimeout == Duration.Inf && settings.sendIdleTimeout == Duration.Inf) BidiFlow.identity
     else {
       val receiveIdleTimeoutFlow = settings.receiveIdleTimeout match {
         case receiveIdleTimeout: FiniteDuration => Flow[FrameHandler.Output].idleTimeout(receiveIdleTimeout)
-        case _ => Flow[FrameHandler.Output]
+        case _                                  => Flow[FrameHandler.Output]
       }
       val sendIdleTimeoutFlow = settings.sendIdleTimeout match {
         case sendIdleTimeout: FiniteDuration => Flow[Input].idleTimeout(sendIdleTimeout)
-        case _ => Flow[Input]
+        case _                               => Flow[Input]
       }
 
       BidiFlow.fromFlows(receiveIdleTimeoutFlow, sendIdleTimeoutFlow)
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/ws/WebSocket.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/ws/WebSocket.scala
@@ -42,6 +42,7 @@ private[http] object WebSocket {
     masking(serverSide, websocketSettings.randomFactory) atop
       FrameLogger.logFramesIfEnabled(websocketSettings.logFrames) atop
       frameHandling(serverSide, closeTimeout, log) atop
+      idleTimeout(websocketSettings) atop
       periodicKeepAlive(websocketSettings) atop
       messageAPI(serverSide, closeTimeout)
 
@@ -89,6 +90,21 @@ private[http] object WebSocket {
   private[this] final val PongFullFrame: FrameStart = FrameEvent.fullFrame(Opcode.Pong, None, ByteString.empty, fin = true)
   private[this] final val mkDirectAnswerPong = () => DirectAnswer(PongFullFrame)
 
+  /** The layer that transparently monitors data flow in both directions and fails the flow
+   *  with [[akka.stream.StreamIdleTimeoutException]] if elements are not passed within configured time.
+   *  */
+  def idleTimeout(settings: WebSocketSettings): BidiFlow[FrameHandler.Output, FrameHandler.Output, FrameOutHandler.Input, FrameOutHandler.Input, NotUsed] = {
+    val receiveIdleTimeoutFlow = settings.receiveIdleTimeout match {
+      case receiveIdleTimeout: FiniteDuration => Flow[FrameHandler.Output].idleTimeout(receiveIdleTimeout)
+      case _ => Flow[FrameHandler.Output].map(identity)
+    }
+    val sendIdleTimeoutFlow = settings.sendIdleTimeout match {
+      case sendIdleTimeout: FiniteDuration => Flow[Input].idleTimeout(sendIdleTimeout)
+      case _ => Flow[Input].map(identity)
+    }
+
+    BidiFlow.fromFlows(receiveIdleTimeoutFlow, sendIdleTimeoutFlow)
+  }
   /**
    * The layer that implements all low-level frame handling, like handling control frames, collecting messages
    * from frames, decoding text messages, close handling, etc.
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/ws/WebSocket.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/ws/WebSocket.scala
@@ -94,16 +94,19 @@ private[http] object WebSocket {
    *  with [[akka.stream.StreamIdleTimeoutException]] if elements are not passed within configured time.
    *  */
   def idleTimeout(settings: WebSocketSettings): BidiFlow[FrameHandler.Output, FrameHandler.Output, FrameOutHandler.Input, FrameOutHandler.Input, NotUsed] = {
-    val receiveIdleTimeoutFlow = settings.receiveIdleTimeout match {
-      case receiveIdleTimeout: FiniteDuration => Flow[FrameHandler.Output].idleTimeout(receiveIdleTimeout)
-      case _ => Flow[FrameHandler.Output].map(identity)
-    }
-    val sendIdleTimeoutFlow = settings.sendIdleTimeout match {
-      case sendIdleTimeout: FiniteDuration => Flow[Input].idleTimeout(sendIdleTimeout)
-      case _ => Flow[Input].map(identity)
-    }
+    if (settings.receiveIdleTimeout == Duration.Inf && settings.sendIdleTimeout == Duration.Inf) BidiFlow.identity
+    else {
+      val receiveIdleTimeoutFlow = settings.receiveIdleTimeout match {
+        case receiveIdleTimeout: FiniteDuration => Flow[FrameHandler.Output].idleTimeout(receiveIdleTimeout)
+        case _ => Flow[FrameHandler.Output]
+      }
+      val sendIdleTimeoutFlow = settings.sendIdleTimeout match {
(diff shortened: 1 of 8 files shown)


gawronA:
CLA signed.

akka-http-core/src/main/scala/akka/http/impl/engine/ws/WebSocket.scala
```diff
@@ -89,6 +90,21 @@ private[http] object WebSocket {
...
+    val receiveIdleTimeoutFlow = settings.receiveIdleTimeout match {
+      case receiveIdleTimeout: FiniteDuration => Flow[FrameHandler.Output].idleTimeout(receiveIdleTimeout)
+      case _ => Flow[FrameHandler.Output].map(identity)
+    }
+    val sendIdleTimeoutFlow = settings.sendIdleTimeout match {
+      case sendIdleTimeout: FiniteDuration => Flow[Input].idleTimeout(sendIdleTimeout)
+      case _ => Flow[Input].map(identity)
+    }
+
+    BidiFlow.fromFlows(receiveIdleTimeoutFlow, sendIdleTimeoutFlow)
```
johanandren, line 106: It think we have no current optimzation getting rid of this BidiFlow when both are identity flows, I think it could be worth special case the default of both idle timeouts disabled, something like: 

```
if (settings.receiveIdleTimeout == Duration.Inf && settings.sendIdleTimeout == Duration.Inf) BidiFlow.identity
else { current logic }
```

akka-http-core/src/main/scala/akka/http/impl/engine/ws/WebSocket.scala
```diff
@@ -89,6 +90,21 @@ private[http] object WebSocket {
   private[this] final val PongFullFrame: FrameStart = FrameEvent.fullFrame(Opcode.Pong, None, ByteString.empty, fin = true)
   private[this] final val mkDirectAnswerPong = () => DirectAnswer(PongFullFrame)
 
+  /** The layer that transparently monitors data flow in both directions and fails the flow
+   *  with [[akka.stream.StreamIdleTimeoutException]] if elements are not passed within configured time.
+   *  */
+  def idleTimeout(settings: WebSocketSettings): BidiFlow[FrameHandler.Output, FrameHandler.Output, FrameOutHandler.Input, FrameOutHandler.Input, NotUsed] = {
+    val receiveIdleTimeoutFlow = settings.receiveIdleTimeout match {
+      case receiveIdleTimeout: FiniteDuration => Flow[FrameHandler.Output].idleTimeout(receiveIdleTimeout)
+      case _ => Flow[FrameHandler.Output].map(identity)
```
johanandren, line 99: Just `Flow[FrameHandler.Output]` already is an identity flow, so need for the additional map, same for the input one when disabled.

akka-http-core/src/test/scala/akka/http/impl/engine/ws/WebSocketServerReceiveIdleTimeoutSpec.scala
```diff
@@ -0,0 +1,57 @@
...
+            TextMessage("dummy")
+          }
+          }, None)
+        })
+      val binding = Await.result(bindingFuture, 3.seconds.dilated)
+      val myPort = binding.localAddress.getPort
+
+      Source.maybe.via(Http().webSocketClientFlow(WebSocketRequest("ws://127.0.01:" + myPort))).to(Sink.ignore).run()
+
+      handlerTermination.future.failed.futureValue
```
johanandren, line 53: What is the type of exception you get?

akka-http-core/src/test/scala/akka/http/impl/engine/ws/WebSocketServerReceiveIdleTimeoutSpec.scala
```diff
@@ -0,0 +1,57 @@
...
+
+      val bindingFuture = Http().newServerAt("localhost", 0)
+        .bindSync({
+          _.attribute(webSocketUpgrade).get.handleMessages(handler.recover { case ex => {
+            handlerTermination.failure(ex)
+            TextMessage("dummy")
+          }
+          }, None)
+        })
+      val binding = Await.result(bindingFuture, 3.seconds.dilated)
```
johanandren, line 48: or `binding.futureValue(timeout(3.seconds.dilated))` to be consistent

johanandren (commented):
Seems like a nice/useful feature, impl looks pretty good, I left some suggestions.

akka-http-core/src/test/scala/akka/http/impl/engine/ws/WebSocketServerReceiveIdleTimeoutSpec.scala
gawronA: `akka.stream.StreamIdleTimeoutException`
I've added exception class matchers in tests

gawronA:
@johanandren thank you for the review. I applied your suggestions.
Meanwhile I've found one issue with a case where `akka.http.client.websocket.send-idle-timeout` is set to non-infinite value and timeout is issued. Then the exception class that we get is `akka.stream.scaladsl.TcpIdleTimeoutException` instead of `akka.stream.StreamIdleTimeoutException`. This is caused by [Akka stream tcp](https://github.com/akka/akka/blob/main/akka-stream/src/main/scala/akka/stream/impl/io/TcpStages.scala#L601) stage where any exceptions derived from `java.util.concurrent.TimeoutException` (including `akka.stream.StreamIdleTimeoutException`) are mapped to `TcpIdleTimeoutException`.
My proposals to resolve it:
1. Intentionally disable `idle-timeout` handling by tcp layer and prepare implementation on http layer [like in http2 server](https://github.com/akka/akka-http/blob/main/akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2.scala#L98)
2. Leave it like that and properly document the case that when `akka.http.client.websocket.send-idle-timeout` is configured, the handler stream will be failed with `akka.stream.scaladsl.TcpIdleTimeoutException`
3. Remove the `akka.http.client.websocket.send-idle-timeout` config parameter and related functionality

Also feel free to directly push whatever headers are needed to `WebSocketServerReceiveIdleTimeoutSpec.scala` and `WebScoketServerSendIdleTimeoutSpec.scala`.

akka-http-core/src/test/scala/akka/http/impl/engine/ws/WebSocketIntegrationSpec.scala
```diff
@@ -244,6 +245,61 @@ class WebSocketIntegrationSpec extends AkkaSpecWithMaterializer(
...
+      val myPort = binding.localAddress.getPort
+
+      import system.dispatcher
+      val future = Source(1 to 10).map(_ => TextMessage("dummy"))
+        .map(x => {
+          Await.result(Future {
+            Thread.sleep(200)
+          }, Duration.apply(3, TimeUnit.SECONDS))
+          x
+        })
```
johanandren, line 263: Replace the in-stream-blocking map+sleep in the tests with `.delay(200.millis)`

johanandren (commented):
Do a `sbt "headerCreateAll; compile"` to get the headers added and the code formatted and check in the result.

I think the TCP stage turning any stream idle timeout to TCP timeout is unfortunate but something that we could possibly fix in Akka core instead of doing anything about here.

gawronA:
In the tests I want to simulate constant flow of messages in one direction and ensure that we still get timeout when messages are not sent in opposite direction.
I managed to replace the in-stream-blocking map with `.delay(200.mililis)` in `send-idle-timeout` test, but when I also changed it in `receive-idle-timeout` test it failed cause the stream wasn't time-outed. This is something I didn't expect to happen and investigated that this piece of code:
```
val future = Source(1 to 10).map(_ => TextMessage("dummy"))
  .map(x => {
    Await.result(Future {
      Thread.sleep(200)
    }, 3.seconds)
    x
  })
  .via(Http().webSocketClientFlow(
    WebSocketRequest("ws://127.0.01:" + myPort),
    settings = ClientConnectionSettings("akka.http.client.websocket.receive-idle-timeout = 1s"))
  ).toMat(Sink.foreach(println))(Keep.right).run()
```
produces no output - expected. However, this piece of code
```
val future = Source(1 to 10).map(_ => TextMessage("dummy"))
  .delay(300.millis, DelayOverflowStrategy.backpressure).addAttributes(Attributes.inputBuffer(1, 1))
  .via(Http().webSocketClientFlow(
    WebSocketRequest("ws://127.0.01:" + myPort),
    settings = ClientConnectionSettings("akka.http.client.websocket.receive-idle-timeout = 1s"))
  ).toMat(Sink.foreach(println))(Keep.right).run()
```
produces:
```
TextMessage.Strict(dummy)
TextMessage.Strict(dummy)
TextMessage.Strict(dummy)
TextMessage.Strict(dummy)
TextMessage.Strict(dummy)
TextMessage.Strict(dummy)
TextMessage.Strict(dummy)
TextMessage.Strict(dummy)
TextMessage.Strict(dummy)
TextMessage.Strict(dummy)
```
which is quite unexpected as the `.delay` operator somehow passes elements through the websocket flow (as if they were received from server) and therefore the timeout don't happen.

I may be wrong but in my opinion the `.delay` operator or `inputBuffer` attribute somehow interfere websocket flow.

akka-http-core/src/test/scala/akka/http/impl/engine/ws/WebSocketIntegrationSpec.scala
```diff
@@ -244,6 +245,57 @@ class WebSocketIntegrationSpec extends AkkaSpecWithMaterializer(
         resTry.failed.get.getMessage should ===("Connection failed.")
       }
     }
+
+    "fail the materialized future with akka.stream.StreamIdleTimeoutException if elements are not received within receive-idle-timeout" in {
+      val bindingFuture = Http().newServerAt("localhost", 0).bindSync({
+        _.attribute(webSocketUpgrade).get.handleMessages(Flow.apply, None)
```
johanandren, line 251: Flow.apply == Flow.identity so this is an echo-server, the test expects it to be a server that does not emit anything, something like `Flow.fromSinkAndSourceCoupled(Sink.ignore, Source.maybe[Message])` could work in a test (in real app you'd need to deal with text/binary, and also consume streaming but here we know it will be strict)

gawronA:
Ahh of course `Flow.apply` was the cause. I completely overlooked that one. Thank you for pointing it out :)

> Other would be collecting those TextMessages in a buffer for example. There is however no async boundary between source and delay, so there is no buffer introduced and `.addAttributes(Attributes.inputBuffer(1, 1))` does not actually do anything.

I would like to leave `.addAttributes(Attributes.inputBuffer(1, 1))` as this causes the source to emit each element with a delay of 200ms. Removing it results in emitting all elements after the delay. The first case better reflects real world scenario.

johanandren:
Right, delay has an internal buffer, so it is not about the async boundary buffers, I didn't remember that from the top of my head. Fine to keep.

akka-http-core/src/test/scala/akka/http/impl/engine/ws/WebSocketServerSendIdleTimeoutSpec.scala
```diff
@@ -0,0 +1,65 @@
...
+          _.attribute(webSocketUpgrade).get.handleMessages(handler.recover {
+            case ex =>
+              handlerTermination.failure(ex)
+              TextMessage("dummy")
+          }, None)
+        }).futureValue(timeout(3.seconds.dilated))
+      val myPort = binding.localAddress.getPort
+
+      Source(1 to 10).map(_ => {
+        Await.result(Future(Thread.sleep(200)), Duration(3, TimeUnit.SECONDS))
```
johanandren, line 56: A final blocking sleep to get rid of, then I think this PR is good to go

johanandren (approved):
LGTM, thanks, good work!

Issue #4295, opened 2023-08-15
Internal server error triggered by “Accept-Charset” header with unexpected value


abjrnxaxa:
Hello!
The following is a (partial) repost of a question I asked in the [Lightbend forum](https://discuss.lightbend.com/t/internal-server-error-triggered-by-accept-charset-header-with-unexpected-value/10427), where I was asked to open up an issue here.

In the [minimal example](https://github.com/akka/akka-http/blob/v10.5.2/docs/src/test/scala/docs/http/scaladsl/HttpServerRoutingMinimal.scala) listed under "Routing DSL for HTTP servers" within the Akka HTTP documentation, an internal server error (HTTP response code 500) is triggered when an HTTP request is made with an unexpected value in the “Accept-Charset” header. This behavior can be replicated, for instance, using curl as follows:
`curl http://localhost:8080/hello -H 'Accept-Charset: asd'`
I tried this in Akka HTTP 10.5.2 and some earlier versions.

Based on the recommendations provided in the [mdm web docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Charset), I would expect the header to be ignored by default. (Alternatively, a `NotAcceptible` response was suggested in the initial thread.)

The current behavior becomes particularly problematic as malicious actors frequently try to inject code through this header; each attempt appears in my server logs.

johanandren:
I wonder if disabling `akka.http.parsing.illegal-header-warnings` could be a workaround?

johanandren:
Alternatively a more specific entry for `Accept-Charset` in `akka.http.parsing.ignore-illegal-header-for`

johanandren:
Probably not, the weird thing is the internal server error, we need to figure out why that happens.

johanandren:
The problem lies in the logic for deciding how the response is encoded, `Marshal.selectMarshallingForContentType` via `WithOpenCharset` via `PredefinedToEntityMarshallers.stringMarshaller`  which passes the parsed but broken charset to `HttpEntity.apply` which in turn touches it triggering throwing of the NIO charset exception (which was actually captured during request parsing, making it a bit tricky to debug/figure out). 

I'm not entirely sure yet how it is best fixed, I think there are two options: 
* by choosing a default charset if the charset is invalid in the `stringMarshaller` (ignoring the header as you suggested)
* detecting in `stringMarshaller` and turning it into an `IllegalRequestException` that will then properly handled by the regular exception handling infra.

PR #4296, opened 2023-08-16
fix: Fail-over to UTF-8 if Accept-Charset is invalid

johanandren:
References #4295

Follows the suggestion in https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Charset and falls back to UTF-8 when the requested Accept-Charset is not a valid charset (instead of failing with a NIO exception and logging noise)

### DIFF
--- a/akka-http/src/main/scala/akka/http/scaladsl/marshalling/PredefinedToEntityMarshallers.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/marshalling/PredefinedToEntityMarshallers.scala
@@ -43,7 +43,18 @@ trait PredefinedToEntityMarshallers extends MultipartMarshallers {
 
   implicit val StringMarshaller: ToEntityMarshaller[String] = stringMarshaller(`text/plain`)
   def stringMarshaller(mediaType: MediaType.WithOpenCharset): ToEntityMarshaller[String] =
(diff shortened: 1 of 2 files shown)


akka-http/src/main/scala/akka/http/scaladsl/marshalling/PredefinedToEntityMarshallers.scala
```diff
@@ -43,7 +43,18 @@ trait PredefinedToEntityMarshallers extends MultipartMarshallers {
...
   def stringMarshaller(mediaType: MediaType.WithOpenCharset): ToEntityMarshaller[String] =
-    Marshaller.withOpenCharset(mediaType) { (s, cs) => HttpEntity(mediaType withCharset cs, s) }
+    Marshaller.withOpenCharset(mediaType) { (s, cs) =>
+      val charset = try {
+        cs.nioCharset // throws if value is invalid
+        cs
+      } catch {
+        case _: java.nio.charset.UnsupportedCharsetException =>
+          // request contained an invalid charset name, fall back to UTF-8
+          HttpCharsets.`UTF-8`
```
patriknw, line 53: ok, so this would be the same as if the header was not specified? effectively ignoring the header if it's invalid, which is along the lines of the recommendation in https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Charset

patriknw (approved):
LGTM

PR #4302, opened 2023-08-30
=core Make use of `statefulMap` instead of `statefulMapConcat`.

He-Pin:
Motivation:
Change to use `statefulMap` instead when not concating.


He-Pin:
I ran a benchmark locally:
image

And I think this is OK.

PR #4303, opened 2023-08-30
chore: Dropping more internal related to the removed SSLConfig

johanandren:
Noticed in nightlies failing after we dropped SSLConfig from Akka


patriknw (approved):
LGTM

PR #4315, opened 2023-09-22
chore: api may change removals for 10.6

johanandren:
Some APIs that has been around long enough that they should be marked stable and some that are used internally and will not change.


patriknw (approved):
LGTM

Issue #3892, opened 2021-08-03
Ongoing request not failed when HTTP/2 PersistentConnection ultimately fails


jrudolph:
That's because we don't pull the pending request out of its slot in https://github.com/akka/akka-http/blob/c0a05b8706aa137426d577d10900fe12501ce68a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/client/PersistentConnection.scala#L140.

PR #4322, opened 2023-10-07
fix: Emit a failed response for the triggering request on connection failure

johanandren:
References #3892 

Superseedes #4299


akka-http-core/src/main/scala/akka/http/impl/engine/http2/client/PersistentConnection.scala
```diff
@@ -139,7 +139,19 @@ private[http2] object PersistentConnection {
           requestOut.fail(new StreamTcpException("connection broken"))
 
           if (connectsLeft.contains(0)) {
-            failStage(new RuntimeException(s"Connection failed after $maxAttempts attempts", cause))
+            if (isAvailable(requestIn)) {
```
johanandren, line 142: This is probably always true, but for good measure

patriknw (approved):
LGTM

Issue #4323, opened 2023-10-12
CVE-2023-44487  


He-Pin:
Description
A client might overload the server by issue frequent RST frames. This can cause a massive amount of load on the remote system and so cause a DDOS attack.

References
https://www.cve.org/CVERecord?id=CVE-2023-44487
https://blog.cloudflare.com/technical-breakdown-http2-rapid-reset-ddos-attack/
https://cloud.google.com/blog/products/identity-security/google-cloud-mitigated-largest-ddos-attack-peaking-above-398-million-rps/

I just read this, not sure if akka-http is affected.
Netty's fix is https://github.com/netty/netty/commit/58f75f665aa81a8cbcf6ffa74820042a285c5e61

patriknw:
Thanks for the notice. We are aware of this and have started to look into it.

This is a well know potential issue for all http2 libraries no worries, but otherwise we would prefer that security related issues are reported according to https://akka.io/security/

utkuaydn:
Hi @patriknw, will this fix will also be released for 10.2.x?

johanandren:
We will not do any further backports of fixes to Akka HTTP 10.2

PR #4324, opened 2023-10-13
fix: Rate limit protection against rapid resets

johanandren:
References #4323

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/framing/RSTFrameLimit.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/framing/RSTFrameLimit.scala
@@ -12,8 +12,6 @@ import akka.http.scaladsl.settings.Http2ServerSettings
 import akka.stream.{ Attributes, FlowShape, Inlet, Outlet }
 import akka.stream.stage.{ GraphStage, GraphStageLogic, InHandler, OutHandler }
 
-import scala.concurrent.duration._
-
 /**
  * INTERNAL API
  */
@@ -28,6 +26,7 @@ private[akka] final class RSTFrameLimit(http2ServerSettings: Http2ServerSettings
   val shape = FlowShape(in, out)
 
   override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new GraphStageLogic(shape) with InHandler with OutHandler {
+    private var rstSeen = false
     private var rstCount = 0
     private var rstSpanStartNanos = 0L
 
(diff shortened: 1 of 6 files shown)


akka-http2-tests/src/test/scala/akka/http/impl/engine/http2/Http2ServerSpec.scala
```diff
@@ -1686,6 +1688,31 @@ class Http2ServerSpec extends AkkaSpecWithMaterializer("""
...
+      val request = HttpRequest(protocol = HttpProtocols.`HTTP/2.0`, uri = "/foo")
+      val error = intercept[AssertionError] {
+        for (streamId <- 1 to 300 by 2) {
+          network.sendBytes(
+            FrameRenderer.render(HeadersFrame(streamId, true, true, network.encodeRequestHeaders(request), None))
+              ++ FrameRenderer.render(RstStreamFrame(streamId, ErrorCode.CANCEL))
+          )
+        }
+      }
+      error.getMessage should include("Too many RST frames per second for this connection.")
```
johanandren, line 1713: Didn't figure out a good way to verify the connection close when I don't quite know how many `sendBytes` will succeed before failing because the connection is closed. Hopefully this is good enough.

akka-http-core/src/main/scala/akka/http/impl/engine/http2/framing/RSTFrameLimit.scala
```diff
@@ -0,0 +1,66 @@
...
+    private var rstSpanStartNanos = 0L
+
+    setHandlers(in, out, this)
+
+    override def onPush(): Unit = {
+      grab(in) match {
+        case frame: RstStreamFrame =>
+          rstCount += 1
+          val now = System.nanoTime()
+          if (rstSpanStartNanos == 0L) {
```
patriknw, line 41: 0 is also a valid value for `System.nanoTime`. Doesn't really matter I guess.
We don't want to `rstSpanStartNanos = System.nanoTime` in the constructor, because there might not be any Rst?

akka-http-core/src/main/scala/akka/http/impl/engine/http2/framing/RSTFrameLimit.scala
```diff
@@ -0,0 +1,66 @@
...
+            rstSpanStartNanos = now
+            push(out, frame)
+          } else if ((now - rstSpanStartNanos) <= maxResetsIntervalNanos) {
+            if (rstCount > maxResets) {
+              failStage(new Http2Compliance.Http2ProtocolException(
+                ErrorCode.ENHANCE_YOUR_CALM,
+                s"Too many RST frames per second for this connection. (Configured limit ${maxResets}/${maxResetsIntervalNanos.nanos.toSeconds} s)"))
```
patriknw, line 48: or use `http2ServerSettings.maxResetsInterval.toCoarsest`

patriknw (commented):
looking good

akka-http-core/src/main/scala/akka/http/impl/engine/http2/framing/RSTFrameLimit.scala
johanandren: That could be fine I guess, it is only once per connection in that case, so other things will be way more expensive anyway.

johanandren: Added a boolean

patriknw (approved):
LGTM

PR #4333, opened 2023-10-31
doc: Simpler old-version-warning

johanandren:
Copied from Alpakka, uses an injected project version from the build to determine its own version, and picks up https://doc.akka.io/docs/akka/current/paradox.json (also generated at build) to determine current latest version, so no longer depends on updating the http://akka.io/versions.json for each release.


patriknw (approved):
LGTM

PR #4335, opened 2023-11-02
build: Publish snapshots

patriknw:
* and publish docs with new ssh key

### DIFF
--- a/project/AkkaDependency.scala
+++ b/project/AkkaDependency.scala
@@ -31,11 +31,7 @@ object AkkaDependency {
       case None =>
         Option(System.getProperty("akka.http.build.akka.version")) match {
           case Some("main") | Some("master") => mainSnapshot
-          case Some("release-2.6") =>
-            // Don't 'downgrade' building even if akka.sources asks for it
-            // (typically for the docs that require 2.6)
-            if (defaultVersion.startsWith("2.6")) Artifact(determineLatestSnapshot("2.6"), true)
-            else Artifact(defaultVersion)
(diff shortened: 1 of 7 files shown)


project/AkkaDependency.scala
```diff
@@ -93,6 +93,8 @@ object AkkaDependency {
 
     val snapshotVersionR = """href=".*/((\d+)\.(\d+)\.(\d+)\+(\d+)-[0-9a-f]+-SNAPSHOT)/"""".r
 
+    // FIXME what shall we do?
```
patriknw, line 96: we don't have directory listing enabled there

project/AkkaDependency.scala
```diff
@@ -80,34 +76,23 @@ object AkkaDependency {
...
-        .toVector.sortBy(_._1)
-    allVersions.last._2
+    val versionFile = "https://doc.akka.io/docs/akka/snapshot/paradox.json"
+    val body = Await.result(http.run(url(versionFile)), 10.seconds).bodyAsString
+    val versionProperty = """"version" : """"
+    val i = body.indexOf(versionProperty)
+    if (i == -1)
+      throw new RuntimeException(s"No version in $versionFile")
+    body.substring(i).split('"')(3)
   }
```
patriknw, line 97: @johanandren I added the snapshot retrieval. Tried it locally and seems to work, but there is a risk that docs were published but artifacts failed. That is currently the case. Hopefully that will be rare when "that problem" has been solved.

johanandren: I think that's ok

johanandren (approved):

PR #4343, opened 2024-01-16
feat: CORS-support

johanandren:
Based on https://github.com/lomigmegard/akka-http-cors but simplified here and there with less API surface, also a few optimisations.


akka-http/src/main/scala/akka/http/scaladsl/server/directives/CorsDirectives.scala
```diff
@@ -0,0 +1,131 @@
...
+
+  // FIXME should we keep this, fold it into the default rejection handler?
+  def corsRejectionHandler: RejectionHandler =
+    RejectionHandler
+      .newBuilder()
+      .handleAll[CorsRejection] { rejections =>
+        val causes = rejections.map(_.description).mkString(", ")
+        complete((StatusCodes.BadRequest, s"CORS: $causes"))
+      }
+      .result()
```
johanandren, line 109: This doesn't work for Java right now

akka-http/src/main/scala/akka/http/scaladsl/server/Directives.scala
```diff
@@ -38,6 +38,7 @@ trait Directives extends RouteConcatenation
   with WebSocketDirectives
   with FramedEntityStreamingDirectives
   with AttributeDirectives
+  with CorsDirectives
```
johanandren, line 41: This will cause problems for users having the original cors module on classpath because names will collide. Maybe that is fine though and the recommendation is to move over to the built in cors support?

johanandren:
The overhead for regular CORS requests is a bit more expensive than I would have hoped, up to 10%, but couldn't find a way to optimise it further (it needs to copy/filter headers).

akka-http-bench-jmh/src/main/scala/akka/http/CorsBenchmark.scala
johanandren, line 1: This would be better just benching the directive/route than an end to end server test, but might be good enough for now.

akka-http/src/main/resources/reference.conf
```diff
@@ -62,4 +62,56 @@ akka.http {
     # This setting can be enabled to pass those empty events to the application for explicit handling.
     emit-empty-events = off
   }
+
+  # Configuration for the cors directive, does not apply unless the directive is used
+  #cors
+  cors {
```
patriknw, line 68: both package names and config are different from akka-http-cors, so there shouldn't be any problems for users that still have akka-http-cors in classpath, right?

patriknw (approved):
LGTM, maybe we should start a migration-guide page informing about this addition and how to migrate?

akka-http/src/main/resources/reference.conf
johanandren: The problem is what I mentioned in https://github.com/akka/akka-http/pull/4343#discussion_r1453711606 

We have an import-all directives-object/baseclass, now including `cors()`, if the original library is also in scope that will clash and not compile, not entirely sure what happens for the scenario of bumping without recompiling (but the original library does not support newer Akka versions so maybe that is a non issue?)

johanandren:
I agree, a migration page of some sort would be good to add. Will do.

ennru (approved):
LGTM.

PR #4348, opened 2024-02-14
Fail on attempt to look up custom modelled headers

jroper:
See #1102.

Throws an `IllegalArgumentException` if an attempt is used to lookup a modelled custom header by class from an `HttpRequest`.

While ideally, the best thing would be to support registering modelled headers, until that gets implemented, I think it's important to make sure users know when they try to lookup a custom header using an unsupported method that it's not supported. This will save a lot of heartache.

Things I'm not sure about this change:

* Performance - it adds an `isAssignableFrom` to every `HttpRequest.header[T]` call when the header is not present.
* Whether there are any places where we depend on the request returning `None`, eg, if we have methods that try this first, and then fallback to using a method that is supported for custom headers. I couldn't find anything.

It's also important to be aware that this could introduce a regression in situations where not returning the header doesn't matter - eg if the header is an unused feature, or a feature where that isn't important and no one cares that it's not working. Or, in the case of https://github.com/akka/akka-grpc/issues/1897, it turns a bug (gRPC compression is never negotiated, but requests still succeed) into a much more serious bug (all requests fail).

### DIFF
--- a/akka-http-core/src/main/java/akka/http/javadsl/model/HttpMessage.java
+++ b/akka-http-core/src/main/java/akka/http/javadsl/model/HttpMessage.java
@@ -57,6 +57,8 @@ public interface HttpMessage {
     /**
      * Try to find the first header of the given class and return
      * Optional.of(header), otherwise this method returns an empty Optional.
+     *
+     * @throws IllegalArgumentException if headerClass is a custom header.
      */
     <T extends HttpHeader> Optional<T> getHeader(Class<T> headerClass);
 
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpHeader.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpHeader.scala
@@ -10,6 +10,7 @@ import scala.util.{ Failure, Success }
 import akka.parboiled2.ParseError
 import akka.http.impl.util.ToStringRenderable
 import akka.http.impl.model.parser.{ CharacterClasses, HeaderParser }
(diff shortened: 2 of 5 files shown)


johanandren:
I think we should hold off with the fail fast here, for the risk of a regression and performance issue. But we should at least amend docs and API-docs with a warning about that it is not possible.

johanandren:
Thinking about this some more I have changed my mind, let's add this fail fast after all. Regressions will prove an incorrect assumption/bug in user apps and if we add support for preparsed modeled headers we would still have this possible problem if a user forgot to register their modeled header. 

Would be good to know about the overhead though.

jroper:
I found [this](https://stackoverflow.com/questions/496928/what-is-the-difference-between-instanceof-and-class-isassignablefrom/12106524#12106524) benchmark here, that showed about 360 ops per microsecond, so about 3ns per invocation. That was for a positive test where the class being tested was a direct subclass of the class it was being invoked on. So, looking like the cost of one or two L1 cache reads, which sounds intuitive for how I imagine the check would be implemented. For a negative test, which is what matters here, presumably it would take longer, as I assume it has to walk up the superclass tree until it finds `Object`. We are testing using a superclass, as opposed to an interface, an interface would assumbly be slower. Our modelled headers have a superclass hierarchy typically 4-5 superclasses deep, so I would expect a negative test to be 10-20ns, assuming the hierarchy is in L1 cache.

Anyway, I think that's fine for this context.

johanandren (approved):
I agree, that definitely seems worth it. LGTM

PR #4358, opened 2024-02-28
build: More foolproof way of picking up version for native image test

johanandren:


johanandren:
Green test run in https://github.com/akka/akka-http/actions/runs/8082049078

PR #4360, opened 2024-03-05
feat: Client TLS identity extraction and assertion directives

johanandren:


johanandren:
@jroper would be great if you could take a peek at the directives and see if this is something that you envisioned, or tell me if I'm completely off.

johanandren:
After thinking about it a bit: I guess a single "require" directive that matches on CN and SAN would maybe make more sense than the separate ones I've created now. And maybe also extractors the values for arbitrary logic, and pulling those back to utilities that we can use from Akka gRPC as well.

jroper:
It looks good to me. It's really hard to say what we should provide in terms of matching the CN vs SAN. Common Names have been deprecated for HTTPS server certificates for almost 25 years. But for client certificates, there's just no standard out there as to what a client certificate should use to identify itself. I think most common would be some sort of DNS value, but should that be in the CN or the SAN?

I think a simple API that matches both CN and SAN may satisfy 95+% of use cases, and a lower level power API will deal with the rest - the directive to extract the certificate is probably enough.

johanandren:
Ready for final review

docs/src/main/paradox/routing-dsl/directives/tls-directives/extractClientCertificate.md
```diff
@@ -0,0 +1,28 @@
...
+
+@@signature [TlsDirectives.scala](/akka-http/src/main/scala/akka/http/scaladsl/server/directives/TlsDirectives.scala) { #extractClientCertificate }
+
+@@@
+
+## Description
+
+This directive extracts the client certificate for the client mTLS connection where the request was made.
+
+If there is no client trusted certificate present (can only happen with `setWantClientAuth(true)`) or reject the request with a @apidoc[TlsClientUnverifiedRejection].
```
efgpinto, line 15:
```suggestion
If there is no client trusted certificate present (can only happen with `setWantClientAuth(true)`) or reject the request with a @apidoc[TlsClientUnverifiedRejection].
```

Having trouble understading this sentence.. what happens if no client trusted certificate present? It's rejected right? Is the "or" ok there?

efgpinto (approved):
Not very knowledgeable on this, but LGTM. Small note on docs.

docs/src/main/paradox/routing-dsl/directives/tls-directives/extractClientCertificate.md
johanandren: Heh, yeah that is incomprehensible. If there is no trusted cert present the request is rejected, I'll look through all of them and update to something more clear.

PR #4372, opened 2024-04-02
docs: remove docs referencing external CORS lib

sebastian-alfers:
Add-on to https://github.com/akka/akka-http/pull/4343


johanandren (commented):
Drop the actual dependency as well?

Issue #2903, opened 2020-01-21
Split packages in akka.http.core and akka.http


atooni:
Akka HTTP Version 10.1.11:

The following packages currently occur in akka.http.core and akka.http with disjunct classes:

- akka.http.impl.settings
- akka.http.javadsl.settings
- akka.http.scaladsl.settings

We have various options here:
* Move all classes to the same package into the same artifact
* Rename one of the packages to avoid the split-package issue, preferably choosing a package name which shares the artifact name as it's prefix
* With a very good reason (is that even possible?) I think we can keep the split package, but it is the most painful and dirty solution

Related to https://github.com/woq-blended/akka-osgi/issues/2.
Comments of https://github.com/akka/akka-http/issues/2902 apply.

mikea:
This issue makes it impossible to use akka http with modules enabled:

` error: module XXX reads package akka.http.scaladsl.settings from both akka.http.core and akka.http`

not sure what is the workaround here besides not using jigsaw/akka-http

PR #4377, opened 2024-04-08
feat: add JWT and claim validation directives

efgpinto:
With this PR, we want to:
- add directive for JWT validation (should validate issue, expiration, signature, etc.)
- allow for claim validation and assertions

Opening as an early version for feedback.


akka-http/src/main/scala/akka/http/scaladsl/server/util/JwtSprayJson.scala
```diff
@@ -0,0 +1,70 @@
...
+import java.time.Clock
+import pdi.jwt.exceptions.JwtNonStringException
+import spray.json._
+
+/**
+ * Implementation of `JwtCore` using `JsObject` from spray-json.
+ *
+ * This class originally came from jwt-spray-json,
+ * but was removed in https://github.com/jwt-scala/jwt-scala/commit/bf1131ce02480103c0b953b97da001105a3ee038
+ */
```
johanandren, line 18: We'll need a mention of original author/license/copyright if it was taken from another project (jwt-scala is Apache 2)

akka-http/src/main/scala/akka/http/scaladsl/server/util/JwtKeyLoader.scala
```diff
@@ -0,0 +1,101 @@
...
+import pdi.jwt.{JwtAlgorithm, algorithms}
+
+import java.io.File
+import java.nio.file.Files
+import java.security.spec.{PKCS8EncodedKeySpec, X509EncodedKeySpec}
+import java.security.{KeyFactory, KeyPair}
+import javax.crypto.spec.SecretKeySpec
+import akka.pki.pem.{DERPrivateKeyLoader, PEMDecoder, PEMLoadingException}
+
+object JwtKeyLoader {
```
johanandren, line 17: Once we are done with the API we should go through and mark as much as possible as internal API, possibly just the directives should be user/public API

akka-http/src/main/scala/akka/http/scaladsl/server/util/JwtSupport.scala
```diff
@@ -0,0 +1,210 @@
...
+
+  def validate(token: String): Either[Exception, JsObject]
+
+}
+
+object JwtSupport {
+
+  private val NoValidationOptions = JwtOptions.DEFAULT.copy(signature = false, expiration = false, notBefore = false)
+
+  def fromConfig(config: Config): JwtSupport = {
```
johanandren, line 32: I think this should all go in the `JwtSettings` and be loaded through `SettingsCompanionImpl#apply(ActorSystem)` by default, since that caches the instance rather than re-parses all config/reloads key in this case for every incoming request.

Especially important since it does blocking read of keypair from filesystem.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/JwtDirectives.scala
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  private def jwt(settings: Config): Directive1[JwtClaims] = {
+    // FIXME: Should we use this or just manually export the header value? Is there a clever way to extract a realm that makes sense? path?
+    authenticateOAuth2("realm", bearerTokenAuthenticator(settings)).flatMap { claims =>
+      extractRequestContext.flatMap { ctx =>
+        provide(new JwtClaims(ctx, claims))
+      }
+    }
+  }
```
johanandren, line 41: I think these two should be `jwt(): Directive1[JwtClaims]` and a public `jw(JwtSettings): Directive1[JwtClaims]` for users that want to programatically configure

akka-http/src/main/scala/akka/http/scaladsl/server/directives/JwtDirectives.scala
```diff
@@ -0,0 +1,98 @@
...
+// JwtClaims provides utilities to easily assert and extract claims from the JWT token
+class JwtClaims(ctx: RequestContext, claims: JsObject) {
+
+  import akka.http.scaladsl.unmarshalling.{ FromStringUnmarshaller => FSU }
+  import ctx.executionContext
+  import ctx.materializer
+
+  def hasClaim(name: String): Boolean = claims.fields.contains(name)
+
+  def get[T](name: String)(implicit fsu: FSU[T]): Future[T] = getClaim[T](name)
```
johanandren, line 64: I think we should skip the integration with `FromStringUnmarshaller` here, having to deal with `Future` is completely superficial, just having a few concrete methods for `getInt`, `getString`, `getBoolean` that all returns `Option` is enough and easily discoverable, handling default values is already built into `Option`/`Optional` so no need for an additional method for that.

Maybe name them `intClaim` for Scala and `getIntClaim` for Java - that way there could be one interface per language but a common implementation.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/JwtDirectives.scala
```diff
@@ -0,0 +1,98 @@
...
+
+  def jwt(): Directive1[JwtClaims] = {
+    extractActorSystem.flatMap { system =>
+      jwt(system.settings.config)
+    }
+  }
+
+  private def jwt(settings: Config): Directive1[JwtClaims] = {
+    // FIXME: Should we use this or just manually export the header value? Is there a clever way to extract a realm that makes sense? path?
+    authenticateOAuth2("realm", bearerTokenAuthenticator(settings)).flatMap { claims =>
```
johanandren, line 36: Extract it to the JwtSettings and put a default realm in config?

project/Dependencies.scala
```diff
@@ -100,7 +101,9 @@ object Dependencies {
   )
 
   lazy val httpCore = l ++= Seq(
-    Test.sprayJson, // for WS Autobahn test metadata
+    Compile.jwtScala,
+    Compile.sprayJson,
```
johanandren, line 105: I'm not convinced the convenience is worth it vs everyone using akka-http now getting these two transitive dependencies. I think we should make it a separate `akka-http-jwt` module that users opt in for using instead.

akka-http/src/main/scala/akka/http/scaladsl/server/util/JwtSupport.scala
```diff
@@ -0,0 +1,210 @@
...
+      .getConfigList("secrets")
+      .asScala
+      .map { secretConfig =>
+        val keyId = secretConfig.getString("key-id")
+        val algorithmSecret: JwtAlgorithmSecret = secretConfig.getString("algorithm") match {
+          case "none" => JwtNoneAlgorithmSecret
+          case alg =>
+            val algorithm = JwtAlgorithm.fromString(alg)
+            if (secretConfig.hasPath("secret-dir")) {
+              val dir = new File(secretConfig.getString("secret-dir"))
```
johanandren, line 48: Pick up all/both the discrete paths from config rather than directory from config with hardcoded magic filenames.

akka-http/src/main/scala/akka/http/scaladsl/server/util/JwtSupport.scala
```diff
@@ -0,0 +1,210 @@
...
+              }
+              JwtKeyLoader.loadKey(keyId, algorithm, dir)
+            } else if (secretConfig.hasPath("secret")) {
+              algorithm match {
+                case symmetric: algorithms.JwtHmacAlgorithm =>
+                  val base64Secret = secretConfig.getString("secret")
+                  val secretBytes = Base64.getDecoder.decode(base64Secret)
+                  JwtSymmetricAlgorithmSecret(symmetric, new SecretKeySpec(secretBytes, algorithm.fullName))
+                case _ => // FIXME exception
+                  throw new RuntimeException("Jwt.secretLiteralNotSupportedWithAsymmetricAlgorithm(keyId, algorithm.name)")
```
johanandren, line 61: Just IllegalArgument is fine here IMO

akka-http/src/main/scala/akka/http/scaladsl/server/directives/JwtDirectives.scala
efgpinto: tks, see https://github.com/akka/akka-http/pull/4377/commits/f99ace74ee4bc95853c5fb481f4074b536665f86

akka-http/src/main/scala/akka/http/scaladsl/server/directives/JwtDirectives.scala
efgpinto: Thanks. removed `FromStringUnmarshaller` now, still lacking a common interface. Will add when implementing the java api.

akka-http/src/main/scala/akka/http/scaladsl/server/util/JwtSupport.scala
efgpinto: thanks. done in https://github.com/akka/akka-http/pull/4377/commits/c02d9ba4a853d4a5bdb952614c0bebe813d7d0dd

akka-http/src/main/scala/akka/http/scaladsl/server/util/JwtSprayJson.scala
efgpinto: added now [here](https://github.com/akka/akka-http/pull/4377/commits/0022b48394a6bffec66ef572c81c58664526d44a#diff-7fa80f952d823dc275b3af34fb322b27e4b30d2ab369e80c1fc72cd631e884d5). please check if it's good.

akka-http-jwt/src/main/scala/akka/http/jwt/impl/settings/JwtSettingsImpl.scala
johanandren, line 1: For the files where everything is internal API I think we should have an `internal` or `impl` package, we've traditionally used `impl` in Akka HTTP and `internal` in newer Akka so maybe go with `internal`

akka-http-jwt/src/main/scala/akka/http/jwt/scaladsl/JwtSettings.scala
```diff
@@ -0,0 +1,19 @@
+package akka.http.jwt.scaladsl
+
+import akka.actor.ClassicActorSystemProvider
+import akka.annotation.{ ApiMayChange, DoNotInherit }
+import akka.http.jwt.impl.settings.{ JwtSettingsImpl, JwtSupport }
+import com.typesafe.config.Config
+
+@ApiMayChange @DoNotInherit
+trait JwtSettings {
+  def jwtSupport: JwtSupport
```
johanandren, line 10: Since the actual type is internal:

```suggestion
  /** INTERNAL API */
  @InternalApi private[akka] def jwtSupport: JwtSupport
```

akka-http-jwt/src/main/scala/akka/http/scaladsl/server/directives/JwtDirectives.scala
```diff
@@ -0,0 +1,61 @@
...
+      jwtSupport.validate(token) match {
+        case Right(claims) => Some(claims)
+        case Left(ex) =>
+          log.debug("The token was rejected: {}", ex.getMessage)
+          None // FIXME: should we propagate anything else further?
+      }
+  }
+
+  // JwtClaims provides utilities to easily assert and extract claims from the JWT token
+  class JwtClaims(claims: JsObject) {
```
johanandren, line 41: This type is now tied to the JwtDirectives trait instance, (path dependent type), which is maybe a bit strange even if it probably doesn't surface as a problem, would probably be fine to put it directly in the directives package, or in the companion object.

akka-http-jwt/src/main/resources/reference.conf
```diff
@@ -0,0 +1,33 @@
...
+  # secrets: [
+  #             {
+  #               key-id: my-key
+  #               issuer: my-issuer
+  #               algorithm: HS256
+  #               # The secret can be set via an environment variable or loaded from a file.
+  #               # To load the secret from an environment variables use:
+  #               secret: ${MY_PRECIOUS_SECRET}
+  #               # To load the secret from a file use (and remove the above secret setting):
+  #               secret-dir: /path/to/secrets
```
efgpinto, line 28: @johanandren picking up from your comment [here](https://github.com/akka/akka-http/pull/4377#discussion_r1556080356):
> Pick up all/both the discrete paths from config rather than directory from config with hardcoded magic filenames.

If I’m understanding correctly.. you’re suggesting that instead of having a config for `secret-dir` we could have one where the complete file path is provided.. say: `secret-file=/path/to/secret.key` right?

But we would need different settings depending if its symmetric or asymmetric, right? For asymmetric keys one it would be something like:
`public-key=/path/to/public.key`
`private-key=/path/to/private.key`

Is this what you had in mind?

akka-http-jwt/src/main/scala/akka/http/scaladsl/server/directives/JwtDirectives.scala
johanandren, line 1: I think there should be a separate package namespace, Java modules doesn't really like that multiple artefacts adds classes to the same packages (even if we are tied with to that with other things in akka-http already https://github.com/akka/akka-http/issues/2903, let's not add more headache to that), `akka.http.jwt` seems like a reasonable base package for the module.

akka-http-jwt/src/main/resources/reference.conf
```diff
@@ -0,0 +1,33 @@
...
+  # Enables or disables the JWT signature validation.
+  # This is useful for development and testing purposes
+  # where you can still assert presence of claims without validating the signature.
+  dev = off
+
+  # The realm to use in the WWW-Authenticate header when a token is missing or invalid.
+  realm = "akka-http-jwt"
+
+  # Allows configuration for the JWT secrets used to verify tokens.
+  # An example config would be:
```
johanandren, line 18: I think the original code had some note about order of secrets being significant? Good to add that here.

akka-http-jwt/src/main/scala/akka/http/jwt/impl/settings/JwtSprayJson.scala
johanandren, line 1: This is all internal, see previous comment about separate package for that

akka-http-jwt/src/main/scala/akka/http/scaladsl/server/directives/JwtDirectives.scala
```diff
@@ -0,0 +1,61 @@
...
+
+    def longClaim(name: String): Option[Long] = claims.fields.get(name).collect { case JsNumber(value) => value.toLong }
+
+    def doubleClaim(name: String): Option[Double] = claims.fields.get(name).collect { case JsNumber(value) => value.toDouble }
+
+    def stringClaim(name: String): Option[String] = claims.fields.get(name).collect { case JsString(value) => value }
+
+    def booleanClaim(name: String): Option[Boolean] = claims.fields.get(name).collect { case JsBoolean(value) => value }
+
+    def toJson: String = claims.toString()
```
johanandren, line 55: This is our escape hatch for more complex JWT structure? `rawJsonString` or something like that since it's not a JSON ADT?

akka-http-jwt/src/main/scala/akka/http/scaladsl/server/directives/JwtDirectives.scala
```diff
@@ -0,0 +1,61 @@
...
+
+    def intClaim(name: String): Option[Int] = claims.fields.get(name).collect { case JsNumber(value) => value.toInt }
+
+    def longClaim(name: String): Option[Long] = claims.fields.get(name).collect { case JsNumber(value) => value.toLong }
+
+    def doubleClaim(name: String): Option[Double] = claims.fields.get(name).collect { case JsNumber(value) => value.toDouble }
+
+    def stringClaim(name: String): Option[String] = claims.fields.get(name).collect { case JsString(value) => value }
+
+    def booleanClaim(name: String): Option[Boolean] = claims.fields.get(name).collect { case JsBoolean(value) => value }
```
johanandren, line 53: These claim extractors looks straight forward enough to me 👍

johanandren (commented):
Looking good!

akka-http-jwt/src/main/resources/reference.conf
johanandren: Yes, I think depending on type we can just have different entries and read what is needed.

efgpinto: tks, see https://github.com/akka/akka-http/pull/4377/commits/b2576e272027d54cd4e41dc2d7a2784aa44b96fd

akka-http-jwt/src/main/scala/akka/http/jwt/impl/settings/JwtSettingsImpl.scala
efgpinto: moved to internal in https://github.com/akka/akka-http/pull/4377/commits/a23c92bac6784e0e37b7e3f13c5cd8fb75521b80

akka-http-jwt/src/main/scala/akka/http/scaladsl/server/directives/JwtDirectives.scala
efgpinto: moved to independent class directives package

akka-http-jwt/src/main/scala/akka/http/scaladsl/server/directives/JwtDirectives.scala
efgpinto: Indeed. I even added a `rawClaim(name: String): String` now.. I think that is the real escape hatch. I think since this is actually a JWT and the claims are JSON objects in reality, having a way to get the raw claim as a JSON string can be useful and actually makes sense in my mind. What are your thoughts?

I'm not entirely sure if we should keep the `toJson`, as in, a way for the user to see the whole real claim (i.e. set of claims).

akka-http-jwt/src/test/resources/my-private.key
```diff
@@ -0,0 +1,28 @@
+-----BEGIN PRIVATE KEY-----
```
johanandren, line 1: Add a README.md to this directory with commands/instructions about how to generate keypair and token to make it easy for ourselves in the future

akka-http-jwt/src/main/scala/akka/http/scaladsl/server/directives/JwtDirectives.scala
johanandren: Maybe we could have spray JsValue as API on the Scala side but json string on the Java side.

> I'm not entirely sure if we should keep...

Yeah, let's skip it for now and add the day it turns out someone stuffs more complex claims in there.

johanandren (commented):
Shaping up nicely, left a few smaller things except for those I think it's time to add a section to the docs.

akka-http-jwt/src/main/scala/akka/http/jwt/scaladsl/server/directives/JwtClaims.scala
```diff
@@ -4,22 +4,76 @@
 
 package akka.http.jwt.scaladsl.server.directives
 
-// JwtClaims provides utilities to easily assert and extract claims from the JWT token
+import spray.json.JsValue
+
+/**
+ * JwtClaims provides a utility to access claims extracted from a JWT token.
+ */
 trait JwtClaims {
```
johanandren, line 12:
```suggestion
 * Not for user extension
 */
@DoNotInherit
trait JwtClaims {
```

akka-http-jwt/src/main/scala/akka/http/jwt/javadsl/server/directives/JwtClaims.scala
```diff
@@ -6,21 +6,73 @@ package akka.http.jwt.javadsl.server.directives
 
 import java.util.Optional
 
-abstract class JwtClaims {
+/**
+ * JwtClaims provides a utility to access claims extracted from a JWT token.
+ */
+trait JwtClaims {
```
johanandren, line 12:
```suggestion
 * Not for user extension
 */
@DoNotInherit
trait JwtClaims {
```

akka-http-jwt/src/main/scala/akka/http/jwt/scaladsl/server/directives/JwtDirectives.scala
```diff
@@ -13,18 +13,35 @@ import akka.http.scaladsl.server.directives.Credentials
 import akka.http.scaladsl.server.directives.SecurityDirectives.{ Authenticator, authenticateOAuth2 }
 import spray.json.JsObject
 
+/**
+ * JwtDirectives provides utilities to easily assert and extract claims from a JSON Web Token (JWT).
```
johanandren, line 17:
```suggestion
 * JwtDirectives provides utilities to assert and extract claims from a JSON Web Token (JWT).
```
We try to avoid "easily" and "simply" in docs since it can feel condescending for a struggling user.

akka-http-jwt/src/main/resources/reference.conf
```diff
@@ -0,0 +1,40 @@
...
+akka.http.jwt {
+  # Enables or disables the JWT signature validation.
+  # This is useful for development and testing purposes
+  # where you can still assert presence of claims without validating the signature.
+  dev = off
+
+  # The realm to use in the WWW-Authenticate header when a token is missing or invalid.
+  realm = "akka-http-jwt"
+
+  # Allows configuration for the JWT secrets used to verify tokens.
```
johanandren, line 17: Should we have more of an explicit doc listing of the supported algorithms and what config keys each require or do you think an example section covering all like now is enough?

I'm thinking something like
```suggestion
  # Allows configuration for the JWT secrets used to verify tokens.
  # Supported algorithms and their required key configs:
  #  - "HS256" - requires either a secret in 'secret' or a filesystem path with a secret via 'secret-path', if both are present xxx happens
```

akka-http-jwt/src/test/scala/akka/http/jwt/scaladsl/server/directives/JwtDirectivesSpec.scala
```diff
@@ -0,0 +1,309 @@
...
+
+  def jwtHeader(claims: Map[String, JsValue] = basicClaims, secret: String = secret): Authorization = {
+    val token = JwtSprayJson.encode(JsObject("alg" -> JsString("HS256")), JsObject(claims), secret)
+    Authorization(OAuth2BearerToken(token))
+  }
+
+  def configTemplate(secret: String) = s"""
+          akka.loglevel = DEBUG
+          akka.http.jwt {
+            dev = off
```
johanandren, line 62: Would be good with one test case for the dev issuer as well

akka-http-jwt/src/main/resources/reference.conf
efgpinto: Yeah, I think it makes sense. Added now, not exactly in that format.. but have a look and let me know: https://github.com/akka/akka-http/pull/4377/commits/51f09bfe5c1eb04f807bb1887a140592e95eb0aa

akka-http-jwt/src/test/scala/akka/http/jwt/scaladsl/server/directives/JwtDirectivesSpec.scala
efgpinto: Added unit tests for when `dev = on` .. I've removed the `dev-issuer` config since that is not used unless for signing (which we don't support).

https://github.com/akka/akka-http/pull/4377/commits/7d4fe8725a8ff77157f7defb6b5c9603d12362f5

docs/src/main/paradox/routing-dsl/directives/jwt-directives/jwt.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Signature
+
+@@signature [JwtDirectives.scala](/akka-http-jwt/src/main/scala/akka/http/jwt/scaladsl/server/directives/JwtDirectives.scala) { #jwt }
+
+@@@
+
+## Description
+
+This directive provides a way to validate a JSON Web Token (JWT) from a request and extracts its claims for further processing. For details on how what a valid JWT is, see [jwt.io](https://jwt.io/) or consult [RFC 7519](https://datatracker.ietf.org/doc/html/rfc7519).
```
johanandren, line 13:
```suggestion
This directive provides a way to validate a JSON Web Token (JWT) from a request and extracts its claims for further processing. For details on what a valid JWT is, see [jwt.io](https://jwt.io/) or consult [RFC 7519](https://datatracker.ietf.org/doc/html/rfc7519).
```

akka-http-jwt/src/main/scala/akka/http/jwt/javadsl/server/directives/JwtDirectives.scala
```diff
@@ -0,0 +1,45 @@
...
+ */
+abstract class JwtDirectives {
+
+  /**
+   * Wraps its inner route with support for the JWT mechanism, enabling JWT token validation.
+   * JWT token validation is done automatically extracting the token from the Authorization header.
+   * If the token is valid, the inner route is executed and provided the set of claims as [[JwtClaims]],
+   * otherwise a 401 Unauthorized response is returned.
+   */
+  def jwt(inner: JFunction[JwtClaims, Route]): Route = RouteAdapter {
```
efgpinto, line 28: So, `jwt` as name is it okay? instead of something like `extractJwtClaims`? Had this question awhile back but forgot to raise..

johanandren (approved):
LGTM with one tiny fix to docs

docs/src/test/scala/docs/http/scaladsl/server/directives/JwtDirectivesExamplesSpec.scala
```diff
@@ -0,0 +1,37 @@
...
+        }
+      }
+
+    // tests:
+
+    // regular request
+
+    // manually injected valid JWT for test purposes with a claim "role" -> "admin"
+    val jwtToken = Authorization(OAuth2BearerToken(
+      "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIn0.Gfx6VO9tcxwk6xqx9yYzSfebfeakZp5JYIgP_edcw_A"))
```
efgpinto, line 29: Should we use an hardcoded token here? I started with building the token in-place but the solution to build one with signature end up a bit clunky (unless I exposed the utility we use ourselves internally which didn't seem like a good option)... ended up leaving it like this but still wondering..

akka-http-jwt/src/main/scala/akka/http/jwt/internal/JwtSettingsImpl.scala
```diff
@@ -0,0 +1,244 @@
...
+  private val NoValidationOptions = JwtOptions.DEFAULT.copy(signature = false, expiration = false, notBefore = false)
+
+  private[jwt] val PrivateKeyConfig = "private-key"
+  private[jwt] val PublicKeyConfig = "public-key"
+
+  private case class JwtValidationException(msg: String) extends RuntimeException(msg)
+
+  def fromConfig(jwtConfig: Config): JwtSupport = {
+    val devSecret = if (jwtConfig.getBoolean("dev")) {
+      Some(JwtSecret("dev", None, JwtNoneAlgorithmSecret))
```
efgpinto, line 71: I wonder if we should print a log warning about `dev` being on if that's the case? Trying to avoid people having it on in production without noticing..

efgpinto (commented):
I think I've addressed all feedback now. Don't have any outstanding items on my list but left a few questions for final review/considerations. Feel free to send any additional feedback of course.

efgpinto:
@johanandren do you have a clue on what might be missing to make CI happy? Had a look at the failing pipelines but not obvious for me.

akka-http-jwt/src/main/scala/akka/http/jwt/javadsl/server/directives/JwtDirectives.scala
johanandren: I think just `jwt` is fine when we don't have a set of different directives that needs to be distinct from each other.

docs/src/test/scala/docs/http/scaladsl/server/directives/JwtDirectivesExamplesSpec.scala
johanandren: I think this is fine

johanandren:
One thing is a test failure:

```
--> [jwt] Start of log messages of test that [Failed(org.scalatest.exceptions.TestFailedException: Request was rejected with rejection AuthenticationFailedRejection(CredentialsRejected,***"akka-http-jwt"))]
| [DEBUG] [04/15/2024 15:27:38.107] [pool-8-thread-12-ScalaTest-running-JwtDirectivesExamplesSpec] [WithLogCapturing(akka://docs-http-scaladsl-server-directives-JwtDirectivesExamplesSpec)] Logging started for test [jwt]
| [DEBUG] [04/15/2024 15:27:38.176] [pool-8-thread-12-ScalaTest-running-JwtDirectivesExamplesSpec] [akka.actor.ActorSystemImpl(docs-http-scaladsl-server-directives-JwtDirectivesExamplesSpec)] The token was rejected: Failed to verify JWT token due to unsupported algorithm
| [DEBUG] [04/15/2024 15:27:38.181] [pool-8-thread-12-ScalaTest-running-JwtDirectivesExamplesSpec] [WithLogCapturing(akka://docs-http-scaladsl-server-directives-JwtDirectivesExamplesSpec)] Logging finished for test [jwt]
<-- [jwt] End of log messages of test that [Failed(org.scalatest.exceptions.TestFailedException: Request was rejected with rejection AuthenticationFailedRejection(CredentialsRejected,***"akka-http-jwt"))]
[info] JwtDirectivesExamplesSpec:
[info] - jwt *** FAILED *** (77 milliseconds)
[info]   Request was rejected with rejection AuthenticationFailedRejection(CredentialsRejected,***"akka-http-jwt") (JwtDirectivesExamplesSpec.scala:32)
[info]   org.scalatest.exceptions.TestFailedException:
[info]   at akka.http.scaladsl.testkit.TestFrameworkInterface$Scalatest.failTest(TestFrameworkInterface.scala:27)
[info]   at akka.http.scaladsl.testkit.TestFrameworkInterface$Scalatest.failTest$(TestFrameworkInterface.scala:27)
[info]   at akka.http.scaladsl.server.RoutingSpec.failTest(RoutingSpec.scala:25)
[info]   at akka.http.scaladsl.testkit.RouteTestResultComponent$RouteTestResult.rawResponse(RouteTestResultComponent.scala:62)
[info]   at akka.http.scaladsl.testkit.RouteTestResultComponent$RouteTestResult.response(RouteTestResultComponent.scala:39)
[info]   at akka.http.scaladsl.testkit.RouteTest.response(RouteTest.scala:66)
[info]   at akka.http.scaladsl.testkit.RouteTest.response$(RouteTest.scala:66)
[info]   at akka.http.scaladsl.server.RoutingSpec.response(RoutingSpec.scala:25)
[info]   at akka.http.scaladsl.testkit.RouteTest.responseAs(RouteTest.scala:77)
[info]   at akka.http.scaladsl.testkit.RouteTest.responseAs$(RouteTest.scala:75)
[info]   at akka.http.scaladsl.server.RoutingSpec.responseAs(RoutingSpec.scala:25)
[info]   at docs.http.scaladsl.server.directives.JwtDirectivesExamplesSpec.$anonfun$new$4(JwtDirectivesExamplesSpec.scala:32)
```

the other is a check that any api annotated with `@ApiMayChange` is listed in the `compatibility-guidelines.md` doc page, so you need to add the public may-change JWT api:s there

akka-http-jwt/src/main/scala/akka/http/jwt/internal/JwtSettingsImpl.scala
johanandren: Yes, sounds like a good idea. 

Somewhat tricky to implement though, we wouldn't want to log it for each request, and not sure we want to put a logger in the settings class. So in the directive with a flag to only log it one or a few times (flag would need to be volatile because cross-thread access, but if can be setting first so it is only ever touched if dev enabled)?

efgpinto: Added with https://github.com/akka/akka-http/pull/4377/commits/c65785d57511a95d796ea6bc4104e09a01221e21

Let me know what you think. It's logging once every minute.

efgpinto:
Green at last. Thanks Johan for the tips. See what you think about https://github.com/akka/akka-http/commit/c65785d57511a95d796ea6bc4104e09a01221e21 .. otherwise I think this is good to go.

johanandren (approved):

PR #4381, opened 2024-04-18
doc: Include quickstart in docs

patriknw:
* with zip files to download

### DIFF
--- a/docs/src/main/paradox/quickstart-scala/index.md
+++ b/docs/src/main/paradox/quickstart-scala/index.md
@@ -11,7 +11,6 @@ This guide highlights Akka HTTP fundamentals in the context of a simple user reg
 
 * Download the quickstart project and run the app.
 * Follow this guide to explore the code.
-* Integrate the project into IntelliJ. If you prefer a different IDE, the principles should be similar, as long as it supports Scala.
 
 ## Prerequisite
 


docs/src/main/paradox/quickstart-java/images/idea-open-project.png
johanandren, line 1: Not sure why we have a screenshot for opening the Java project as a sbt project (and with JDK 8), maybe we should drop this?

docs/src/main/paradox/quickstart-scala/images/idea-open-project.png
johanandren, line 1: Makes more sense here but since we don't support JDK 8 anymore maybe confusing?

build.sbt
```diff
@@ -483,6 +483,10 @@ lazy val docs = project("docs")
         writer = new Writer(serializerPlugins = Writer.defaultPlugins(paradoxDirectives.value)))
     },
     paradoxGroups := Map("Language" -> Seq("Scala", "Java")),
+    paradoxRoots := List(
+      "index.html",
+      "quickstart-java/index.html",
+      "quickstart-scala/index.html"),
```
johanandren, line 489: What does this lead to, that we don't list the quickstarts in the regular navigation tree but have separate navs for those?

johanandren (approved):
LGTM but something existing we may want to improve

docs/src/main/paradox/quickstart-scala/index.md
```diff
@@ -0,0 +1,204 @@
...
+The Akka HTTP example for Scala is a zipped project that includes a distribution of the [sbt](http://www.scala-sbt.org) build tool. Download and unzip the example as follows:
+
+1. Download the project [zip file](../attachments/akka-http-quickstart-scala.zip).
+2. Extract the zip file to a convenient location.
+
+* On Linux and MacOS systems, open a terminal and use the command `unzip` akka-quickstart-scala.zip. Note: On OSX, if you unzip using Archiver, you also have to make the sbt files executable by entering the following two commands:
+
+```
+$ chmod u+x ./sbt
+$ chmod u+x ./sbt-dist/bin/sbt
```
patriknw, line 35: We don't really have those for sbt included. I think we should require a local installation of sbt.

build.sbt
patriknw: It's pretty good to not have them included in ToC because then they become "standalone" without distraction of other things.

docs/src/main/paradox/quickstart-java/images/idea-open-project.png
patriknw: yeah, I don't know why we should promote IntelliJ. I'll remove the page.

build.sbt
johanandren: Was more looking for confirmation that separate ToC trees is what it means, wasn't sure.

johanandren (approved):

PR #4402, opened 2024-05-08
build: Split out docs publish workflow to allow post-release docs publish

johanandren:
As we have it in other Akka repos.


ennru (approved):
LGTM.

Issue #4346, opened 2024-01-28
Remove dependency on `sbt-dependency-graph` to avoid repo.scala-sbt.org


mkurz:
I started to block [repo.scala-sbt.org](http://repo.scala-sbt.org/) and [repo.typesafe.com](http://repo.typesafe.com/) in my `/etc/hosts`. 

sbt-dependency-graph never made it to maven central.

Turns out, this line

https://github.com/akka/akka-http/blob/67c704111b755274e5daa423c2c923cbb017e8f9/project/plugins.sbt#L16

is not needed anymore since sbt 1.4, because it's build in now.
See https://www.scala-sbt.org/1.x/docs/sbt-1.4-Release-Notes.html#sbt-dependency-graph+is+in-sourced

Actually for advanced usage you can replace it with 
```
addDependencyTreePlugin
```

However it's not that easy, because you make use of apis here:
https://github.com/akka/akka-http/blob/67c704111b755274e5daa423c2c923cbb017e8f9/project/ValidatePullRequest.scala#L10-L11

which in sbt are now [internal](https://github.com/sbt/sbt/blob/v1.9.8/main/src/main/scala/sbt/internal/graph/model.scala#L67)...
(Maybe it's easy to fix, but didn't try, leave that up to you)

mkurz:
Same story here:
- https://github.com/lightbend/sbt-paradox-dependencies/issues/34

PR #4404, opened 2024-05-17
build: Build gardening

johanandren:
Refs #4346

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2,7 +2,6 @@ import akka._
 import AkkaDependency._
 import Dependencies.{h2specExe, h2specName}
 import com.typesafe.sbt.MultiJvmPlugin.autoImport.MultiJvm
-import com.typesafe.sbt.SbtScalariform.ScalariformKeys
 import java.nio.file.Files
 import java.nio.file.attribute.{PosixFileAttributeView, PosixFilePermission}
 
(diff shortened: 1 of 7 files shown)


project/ValidatePullRequest.scala
```diff
@@ -29,580 +30,89 @@ import scala.collection.immutable
 import scala.sys.process._
 import scala.util.Try
 import scala.util.matching.Regex
+ */
```
ennru, line 33: TODO?

build.sbt
```diff
@@ -83,7 +80,7 @@ lazy val root = Project(
     id = "akka-http-root",
     base = file(".")
   )
-  .enablePlugins(UnidocRoot, NoPublish, PublishRsyncPlugin, AggregatePRValidation, NoScala3)
+  .enablePlugins(UnidocRoot, NoPublish, PublishRsyncPlugin, NoScala3) // AggregatePRValidation
```
ennru, line 83: Should it stay as comment?

project/CopyrightHeader.scala
```diff
@@ -27,8 +28,7 @@ object CopyrightHeader extends AutoPlugin {
         )
       )
     },
-    ValidatePR / additionalTasks += Compile / headerCheck,
-    ValidatePR / additionalTasks += Test / headerCheck
+    Def.settings(additionalTasks += Compile / headerCheck, additionalTasks += Test / headerCheck),
```
ennru, line 31: It might be better to even here be explicit and use `AkkaHttpValidatePullRequest.additionalTasks`

project/ValidatePullRequest.scala
```diff
@@ -29,580 +30,89 @@ import scala.collection.immutable
...
-                logger,
-                name.value)
-            }
-        }
+/**
+ * This autoplugin adds Paradox doc generation to validatePullRequest task,
+ * when a project has ParadoxPlugin autoplugin enabled.
+ */
+object ParadoxWithPrValidation extends AutoPlugin {
+  import AkkaHttpValidatePullRequest._
```
ennru, line 106: It might be better to even here be explicit and use `AkkaHttpValidatePullRequest.additionalTasks`

build.sbt
johanandren: I should remove, that was the old ad-hoc pr validation as far as I understand.

Issue #4417, opened 2024-09-04
Failed: compiling against Akka main because scala java 8 compat dropped 


johanandren:
Fall out from https://github.com/akka/akka/pull/32435

Failing nightlies Akka HTTP was relying on a transitive dependency.

Fyi @sebastian-alfers this would be nice to fix next since CI is failing.

sebastian-alfers:
Taking a look.

sebastian-alfers:
https://github.com/akka/akka-http/pull/4418

PR #4418, opened 2024-09-05
chore: Drop scala java8 compat

sebastian-alfers:
Fixes https://github.com/akka/akka-http/issues/4417

Imports seem inconsistent, so its a best-effort. Should maybe auto-alignt in a separate pr?

Is there maybe a better way to avoid the "map to object", like this one?
```
.map(length -> (Object)length));
```

### DIFF
--- a/akka-http-core/src/main/java/akka/http/javadsl/model/ContentRange.java
+++ b/akka-http-core/src/main/java/akka/http/javadsl/model/ContentRange.java
@@ -9,7 +9,7 @@ import akka.http.scaladsl.model.ContentRange$;
 import java.util.Optional;
 import java.util.OptionalLong;
 
-import scala.jdk.javaapi.OptionConverters;
+import scala.Option;
 
 public abstract class ContentRange {
     public abstract boolean isByteContentRange();
@@ -31,7 +31,7 @@ public abstract class ContentRange {
     }
     @SuppressWarnings("unchecked")
     public static ContentRange create(long first, long last, OptionalLong instanceLength) {
-        return ContentRange$.MODULE$.apply(first, last, OptionConverters.toScala(instanceLength).map(length -> (Object)length));
+        return ContentRange$.MODULE$.apply(first, last, (instanceLength.isPresent() ? Option.apply(instanceLength.getAsLong()) : Option.empty()));
     }
     public static ContentRange createUnsatisfiable(long length) {
         return new akka.http.scaladsl.model.ContentRange.Unsatisfiable(length);
--- a/akka-http-core/src/main/java/akka/http/javadsl/model/ContentRange.java
+++ b/akka-http-core/src/main/java/akka/http/javadsl/model/ContentRange.java
@@ -8,7 +8,8 @@ import akka.http.scaladsl.model.ContentRange$;
 
 import java.util.Optional;
 import java.util.OptionalLong;
-import scala.compat.java8.OptionConverters;
+
+import scala.jdk.javaapi.OptionConverters;
 
 public abstract class ContentRange {
     public abstract boolean isByteContentRange();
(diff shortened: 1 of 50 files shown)


sebastian-alfers:
Currently fails with:
```
[info] - should Work for usage when running HTTPS server *** FAILED *** (42 milliseconds)
[info]   The future returned an exception of type: javax.net.ssl.SSLHandshakeException, with message: NotAfter: Tue Sep 03 13:18:30 CEST 2024. (SSLContextFactorySpec.scala:105)
```

But seems also to fail in `main`, so taking a look what is going on there.

sebastian-alfers:
Depends on https://github.com/akka/akka-http/pull/4419.

akka-http-core/src/main/java/akka/http/javadsl/model/ContentRange.java
```diff
@@ -30,7 +31,7 @@ public static ContentRange create(long first, long last, long instanceLength) {
     }
     @SuppressWarnings("unchecked")
     public static ContentRange create(long first, long last, OptionalLong instanceLength) {
-        return ContentRange$.MODULE$.apply(first, last, OptionConverters.toScala(instanceLength));
+        return ContentRange$.MODULE$.apply(first, last, OptionConverters.toScala(instanceLength).map(length -> (Object)length));
```
johanandren, line 34: `OptionConverters.toScala(OptionalLong)` returns `Option[java.lang.Long]` so this is just Java not understanding variance (as expected), however, it is surprising that it thinks that `Option[Long]` in Scala is `Option<Object>`.

Actually mapping involves allocation and a lambda that will actually be invoked, we don't want that, just force-cast is probably ok `(Option<Object>)(Object) OptionConverters.toScala(instanceLength)` even if it feels a bit dirty.

akka-http-core/src/main/java/akka/http/javadsl/model/headers/CacheDirectives.java
```diff
@@ -19,7 +19,7 @@ public static CacheDirective MAX_STALE() {
         return new akka.http.scaladsl.model.headers.CacheDirectives.max$minusstale(OptionConverters.toScala(Optional.empty()));
     }
     public static CacheDirective MAX_STALE(long deltaSeconds) {
-        return new akka.http.scaladsl.model.headers.CacheDirectives.max$minusstale(OptionConverters.toScala(OptionalLong.of(deltaSeconds)));
+        return new akka.http.scaladsl.model.headers.CacheDirectives.max$minusstale(OptionConverters.toScala(OptionalLong.of(deltaSeconds)).map(delta -> (Object)delta));
```
johanandren, line 22: This seems kinda unnecessary, can't we just call `new Some<>(deltaSeconds)` and skip the whole conversion?

akka-http-core/src/main/java/akka/http/javadsl/model/ContentRange.java
johanandren: Or if we want to keep it slightly cleaner, this should compile fine `(instanceLength.isPresent() ? Option.apply(instanceLength.getAsLong()) : Option.empty())`

akka-http-core/src/main/java/akka/http/javadsl/model/sse/ServerSentEvent.java
```diff
@@ -18,18 +18,19 @@
...
  * Representation of a server-sent event. According to the specification, an empty data field
  * designates an event which is to be ignored which is useful for heartbeats.
  */
 public abstract class ServerSentEvent {
 
-    private static final Option<String> stringNone =  toScala(Optional.empty());
+    private static final Option<String> stringNone = OptionConverters.toScala(Optional.empty());
 
-    private static final Option<Object> intNone = toScala(OptionalInt.empty());
+    private static final Option<Object> intNone = OptionConverters.toScala(OptionalInt.empty()).map(emptyIntOption -> (Object)emptyIntOption);
```
johanandren, line 33: These two are really weird to start with, `Option.empty()` should be fine for both?

johanandren (approved):

PR #4461, opened 2025-01-08
Validate custom HTTP method names

### LINKED ISSUE #4460: Validate custom HTTP methods
Currently, we don't validate the input to `HttpMethod.custom()`. A user can pass a method with newlines into it, for example, and use that to make HTTP client requests with headers injected from the passed in method name. In theory, this is a security vulnerability, however that would require an application that made requests using a string method name received from an untrusted source - I don't believe any application should be making a request using an HTTP method name that came from an untrusted source, that's inviting a whole host of security vulnerabilities beyond just this one, that validating the method name would not mitigate.

Nevertheless, we should validate the method passed to it.

jroper:
Fixes #4460

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/model/parser/CommonRules.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/model/parser/CommonRules.scala
@@ -6,11 +6,11 @@ package akka.http.impl.model.parser
 
 import scala.collection.immutable
 import scala.collection.immutable.TreeMap
-
 import akka.http.scaladsl.model._
 import akka.http.scaladsl.model.headers._
 import akka.parboiled2._
 import akka.parboiled2.support.hlist._
+import akka.util.ConstantFun
 
 private[parser] trait CommonRules { this: Parser with StringBuilding =>
   protected def maxCommentParsingDepth: Int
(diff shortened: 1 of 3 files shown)


akka-http-core/src/main/scala/akka/http/impl/model/parser/CommonRules.scala
```diff
@@ -465,7 +465,7 @@ private[parser] trait CommonRules { this: Parser with StringBuilding =>
     token ~> { s =>
       HttpMethods.getForKey(s) match {
         case Some(m) => m
-        case None    => HttpMethod.custom(s)
+        case None    => HttpMethod(s, isSafe = false, isIdempotent = false, requestEntityAcceptance = RequestEntityAcceptance.Expected, ConstantFun.anyToTrue)
```
jroper, line 468: For custom headers received by the server, there is no need to validate them because they have already been parsed according to the spec, so skip the validation in `HttpMethod.custom`.

akka-http-core/src/test/scala/akka/http/scaladsl/model/HttpMethodSpec.scala
```diff
@@ -0,0 +1,21 @@
...
+
+package akka.http.scaladsl.model
+
+import org.scalatest.matchers.should.Matchers
+import org.scalatest.wordspec.AnyWordSpec
+
+class HttpMethodSpec extends AnyWordSpec with Matchers {
+  "HttpMethod.custom()" should {
+    "accept a valid name" in {
+      HttpMethod.custom("Yes.Thi$_is_1~'VAL|D`_me+h*d-^ame!#%&")
```
johanandren, line 13: The most reasonable custom method name I ever saw :D

johanandren (approved):
LGTM

akka-http-core/src/test/scala/akka/http/scaladsl/model/HttpMethodSpec.scala
jroper: It's specified in RFC666 - Gaslighting HTTP resources.

PR #4477, opened 2025-03-27
bump: sbt-scalafix that works with Scala 2.13.15

johanandren:
Without this the project does not load locally for me (how does it pass CI checks?)


ennru (approved):
LGTM.

PR #4504, opened 2025-07-03
feat: Option to not emit content length for strict entities

johanandren:


johanandren:
Scala 3 bincomp for moving from case class to regular class is trickier, would need separate sources for Scala 2.13 and 3 because case classes extend other things there, not available in Scala 2 stdlib (`scala.deriving.Mirror$Sum,scala.deriving.Mirror`)

patriknw (approved):
LGTM

PR #4512, opened 2025-10-18
chore: Fix resolvers

JustinPihony:


johanandren:
Hmmm, why isn't CI running here?

johanandren:
Was because `actions/checkout@v4` not allowed by repo security, would be better with more specific, or preferrably hash but we can follow up with that, added v4 for now.

JustinPihony:
Shall I use this for the rest of my changes? 

`uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11`

PR #4537, opened 2026-04-14
build: bump h2spec and make possible to run on non x86

johanandren:
Previously it has been impossible to run all the HTTP2 tests because hardcoded to only do x86.

This PR
  - Upgrades h2spec from 1.5.0 to 2.6.0 (latest)
  - enable a few h2spec test cases that now pass
  - Makes h2spec tests runnable on non-x86 platforms (e.g. Apple Silicon) by checking for a locally installed h2spec binary before falling back to the downloaded amd64
   binary
  - Tests skip gracefully when h2spec is not available instead of failing
  - Adapts test CLI flags for h2spec v2.x (http2/<section> positional arg instead of -s)
  - Adds install instructions for h2spec to CONTRIBUTING.md

On amd64 (CI), behavior is unchanged — the binary is downloaded automatically. On other architectures, developers can install h2spec via `go install github.com/summerwind/h2spec/cmd/h2spec@latest`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -366,6 +366,19 @@ the plaintext traffic in Wireshark: you can set its `InterceptorPort` to
 the same value as the `ReturnPort` and then sniff the loopback interface and
 filter on that port.
 
+### h2spec (HTTP/2 conformance testing)
+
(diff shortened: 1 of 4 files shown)


ennru (approved):
LGTM.

PR #4548, opened 2026-08-18
fix: HTTP/2 connection unusable after a bad request response

johanandren:
A header that fails validation, such as `Connection` or an unknown `:method`, aborted decoding of the whole block. The remaining headers never entered our HPACK dynamic table while the peer's encoder had already added them, so every later header block on the connection resolved its indices against the wrong entries. Header parse errors are now collected in the header listener so the block is always decoded to completion.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/http2/hpack/HeaderDecompression.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/http2/hpack/HeaderDecompression.scala
@@ -98,8 +98,8 @@ private[http2] final class HeaderDecompression(masterHeaderParser: HttpHeaderPar
           headerLimitExceeded(s"Decoded header list for stream $streamId exceeded configured maximum of ${http2Settings.maxHeaderListSize} bytes")
         else parseError match {
           // push details further and let RequestErrorFlow handle responding with bad request
-          case Some(info) => push(eventsOut, ParsedHeadersFrame(streamId, endStream, Seq.empty, prioInfo, Some(info)))
-          case None       => push(eventsOut, ParsedHeadersFrame(streamId, endStream, headers.result(), prioInfo, None))
+          case Some(_) => push(eventsOut, ParsedHeadersFrame(streamId, endStream, Seq.empty, prioInfo, parseError))
+          case None    => push(eventsOut, ParsedHeadersFrame(streamId, endStream, headers.result(), prioInfo, None))
         }
       } catch {
         case _: IOException =>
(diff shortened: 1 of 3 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/http2/hpack/HeaderDecompression.scala
```diff
@@ -87,13 +96,12 @@ private[http2] final class HeaderDecompression(masterHeaderParser: HttpHeaderPar
...
           headerLimitExceeded(s"Decoded header list for stream $streamId exceeded configured maximum of ${http2Settings.maxHeaderListSize} bytes")
-        else
-          push(eventsOut, ParsedHeadersFrame(streamId, endStream, headers.result(), prioInfo, None))
-      } catch {
-        case ex: ParsingException =>
-          decoder.endHeaderBlock() // decoding was aborted midway, reset the accumulated header list size
+        else parseError match {
           // push details further and let RequestErrorFlow handle responding with bad request
-          push(eventsOut, ParsedHeadersFrame(streamId, endStream, Seq.empty, prioInfo, Some(ex.info)))
+          case Some(info) => push(eventsOut, ParsedHeadersFrame(streamId, endStream, Seq.empty, prioInfo, Some(info)))
```
ennru:
```suggestion
          case Some(_) => push(eventsOut, ParsedHeadersFrame(streamId, endStream, Seq.empty, prioInfo, parseError))
```
More obvious to pass the `parseError`.

akka-http2-tests/src/test/scala/akka/http/impl/engine/http2/hpack/HeaderDecompressionSpec.scala
```diff
@@ -10,19 +10,51 @@ import akka.http.impl.engine.parsing.HttpHeaderParser
...
+      case (block, idx) => HeadersFrame(1 + idx * 2, endStream = true, endHeaders = true, block, None)
+    }.toList)
+      .via(new HeaderDecompression(HttpHeaderParser(parserSettings, log), parserSettings, settings.http2Settings))
+      .runWith(Sink.seq).futureValue
+      .map(_.asInstanceOf[ParsedHeadersFrame])
+  }
+
   "HeaderDecompression" should {
 
+    "keep the HPACK dynamic table in sync when a header block is rejected" in {
```
ennru, line 33: Possibly add a test case that exercises two invalid headers?

ennru (approved):
LGTM.

PR #4554, opened 2026-08-19
fix: reject WINDOW_UPDATE that would overflow the flow-control window

johanandren:


akka-http-core/src/main/scala/akka/http/impl/engine/http2/Http2StreamHandling.scala
```diff
@@ -383,7 +404,7 @@ private[http2] trait Http2StreamHandling extends GraphStageLogic with LogHelper
     }
 
     def increaseWindow(delta: Int): StreamState = {
-      outStream.increaseWindow(delta)
+      outStream.increaseWindow(delta) // used for SETTINGS_INITIAL_WINDOW_SIZE distribution; overflow ignored here, same as before
```
ennru:
```suggestion
      outStream.increaseWindow(delta) // used for SETTINGS_INITIAL_WINDOW_SIZE distribution
```

ennru (approved):

PR #4556, opened 2026-08-20
fix: end Inflater/Deflater on cancellation, not just normal stream completion

johanandren:
Gzip/deflate (de)compression stages held `java.util.zip.Inflater/Deflater` instances that wrap native zlib memory, but only released them (.end()) on normal stream completion — a client disconnecting mid-response, or any other cancellation/failure, leaked the native memory instead.

Added `postStop()`-based cleanup to the decompression stages (releasing every Inflater created, including for concatenated deflate blocks) and a `Compressor.close()` hook wired into the encoder's stream teardown, so resources are released on every termination path.


ennru (approved):

patriknw:
Anything that need fixing in Akka Streams gzip too, or is that completely different?

johanandren:
All upstream compression/deflation looks safe as far as I can see.

PR #4559, opened 2026-08-21
chore: consistently use Locale.ROOT for casing

johanandren:
Not a huge thing but we overlooked and used JVM selected locale in a few places, which could mess things up (turkish for example).


ennru (approved):

johanandren:
Taking a look in other library repos as well.

Issue #4563, opened 2026-08-26
Route.toFunction / ServerBuilder.bind(route) should allow custom ExecutionContext and Materializer (like deprecated asyncHandler)


vqngoacc:
Since 10.2.0 and https://github.com/akka/akka-http/pull/3115, Http().newServerAt(...).bind(route) converts a Route via the implicit RouteResult.routeToFunction → Route.toFunction(route), which hardcodes:

- ExecutionContext → system.classicSystem.dispatcher
- Materializer → SystemMaterializer(system).materializer

The deprecated Route.asyncHandler allowed passing a custom executionContext and materializer, with fallback to materializer.executionContext when EC was null.

For services that want to isolate some APIs on a dedicated dispatcher (separate from akka.actor.default-dispatcher), this matters because:

1. Route.createAsyncHandler maps Future[RouteResult] → HttpResponse on the EC passed into createAsyncHandler — currently always system.dispatcher from toFunction.
2. ServerBuilder.withMaterializer does not fix this — it only affects server bind/stream materialization (bindAndHandleAsyncImpl). The Route → handler conversion happens before bind, via toFunction(system), and never sees the builder’s materializer.

In testing we observed management endpoints (health, metrics) becoming unresponsive when akka.actor.default-dispatcher is saturated, even though these endpoints are served on a separate port with a custom Materializer configured via ServerBuilder.withMaterializer. The route still completed via Future callbacks on the default dispatcher.

PR #4564, opened 2026-09-01
feat: allow custom ExecutionContext/Materializer when converting Route to handler function

johanandren:
References #4563

`Route.toFunction`/`toFlow` (Scala) and `Route.handler` (Java) gained overloads that accept an explicit `ExecutionContext`/`Materializer` instead of always defaulting to
`system.dispatcher`/`SystemMaterializer`, restoring the flexibility the deprecated asyncHandler used to offer, so routes served on a custom dispatcher no longer complete on the shared default dispatcher regardless of `ServerBuilder.withMaterializer`.

Also fixes a related bug: `RouteAdapter.flow(system, materializer)` (Java) was silently ignoring the given materializer and always using `SystemMaterializer`.

`ServerBuilder.bind()` itself is unchanged (adding an overload there isn't possible, `ServerBuilder` lives in akka-http-core, which `Route` can't be referenced from, and there'd also be source/bin compat problems with that).

Users needing isolation now call `.bind(Route.toFunction(route, ec, mat))` or `.bind(route.handler(system, ec, mat)::apply)` explicitly.


sebastian-alfers (approved):
LGTM

